PR #633, opened 2016-04-13
compiling of static files with webpack hotloading

dorey:


dorey:
This is now part of #644

Issue #646, opened 2016-04-19
Create separate lists of KPI forms


tinok:
This is necessary for implementing the new UI. 

We need three lists: 
Deployed and active - "Active"
Non-deployed - "Forms"
Deployed and non-active - "Inactive"

dorey:
[ ] DRF changes to the asset serializer
[ ] splitting up the react view based on the DRF changes

tinok:
We need to include the number of submissions for each project as well.

jnm:
I've tracked the need for retrieving the number of submissions from KC in a separate issue: https://github.com/kobotoolbox/kpi/issues/689

PR #673, opened 2016-05-04
Towards #646

jnm:


jnm:
Superseded by https://github.com/kobotoolbox/kpi/pull/688

PR #692, opened 2016-05-11
Use Travis to build and test images.

### LINKED ISSUE #691: No Travis tests running.

esmail:
Closes #691.


dorey:
So this fails when the docker build fails?

I'm going to close this because it will conflict with the travis file in `new-ui` which runs python and js tests and uses the new `sudo: false` travis config

PR #711, opened 2016-05-19
Asset endpoint lists available Enketo URLs

### LINKED ISSUE #705: provide enketo URLs in asset api endpoint
similar to downloads, provide a list of the available enketo links

jnm:
Closes #705. **NOTE**: `ENKETO_API_TOKEN` now must be specified in the environment.

Available only on the detail view (not the list view). Returns all available Enketo survey URLs per http://apidocs.enketo.org/v2/#/post-survey-all. Some link types listed in the placeholder are not being returned by https://ee.kobotoolbox.org/api/v2/survey/all, and trying to request them specifically (e.g. https://ee.kobotoolbox.org/api/v2/survey/single) returns a 405. Perhaps an Enketo configuration change is required to enable more types of surveys.


dorey:
Already merged into `develop`

Issue #703, opened 2016-05-13
Display user_notification message


tinok:
A sitewide_message > user_notification should be displayed in the KPI notification box. 

Users should be notified by the new notification by seeing the message expanded. The user can then close the message but see it again by clicking on the notification icon in the top bar.

PR #760, opened 2016-06-17
Maintenance notification

### LINKED ISSUE #729: Automatic question names are not truncated
Question names are supposed to be truncated but are currently not truncated at all, leading to broken XML / performance issues, and very long strings saved when transmitting XML to KC. 
[long_name.xlsx](https://github.com/kobotoolbox/kpi/files/286984/long_name.xlsx)

dorey:
Display a notification when the server is going down for maintenance. Closes #729


esmail:
Was this not supposed to close #703?

PR #774, opened 2016-06-23
Assigns a "$kuid" field with each row in asset.content

dorey:


dorey:
Depends on formpack#93 being merged

PR #779, opened 2016-06-27
Handle more error conditions when accessing Enketo

jnm:


esmail:
Note to self:
`All exceptions that Requests explicitly raises inherit from requests.exceptions.RequestException.`
http://docs.python-requests.org/en/master/user/quickstart/#errors-and-exceptions

PR #791, opened 2016-06-30
Fix #788 move empty library message into page body

### LINKED ISSUE #788: Move 'empty library' message into body
It is currently at the bottom of the page (only shows for users who have no library assets)

image

pmusaraj:


pmusaraj:
This PR is now no longer useful, its commits are now in develop-UI.

PR #823, opened 2016-07-15
`kc_reader` utility

dorey:
* adds a `kc_reader` utility To perform read-only queries on KC tables for xforms and submissions.
* models will only be accessed through getter methods such as [`instance_count`](https://github.com/kobotoolbox/kpi/commit/e7339314cc1d379399437ecb88cbda4ab5b7c72d#diff-05d181aad038e404fe4e7982db4949b4R65)

todo:
find a way to prevent django from auto-creating migrations

### DIFF
--- a/kpi/deployment_backends/kc_reader/shadow_models.py
+++ b/kpi/deployment_backends/kc_reader/shadow_models.py
@@ -1,0 +1,78 @@
+from django.contrib.auth.models import User
+from django.db import models
+from django.db import ProgrammingError
+
+
+class ReadOnlyModelError(ValueError):
+    pass
+
+
+class _ReadOnlyModel(models.Model):
+    class Meta:
+        abstract = True
+
+    def save(self, *args, **kwargs):
+        raise ReadOnlyModelError('Cannot save read-only-model')
+
+    def delete(self, *args, **kwargs):
+        raise ReadOnlyModelError('Cannot delete read-only-model')
+
+
+class LazyModelGroup:
+    @property
+    def XForm(self):
+        if not hasattr(self, '_XForm'):
+            self._define()
+        return self._XForm
+
+    @property
+    def Instance(self):
+        if not hasattr(self, '_Instance'):
+            self._define()
+        return self._Instance
+
+    def _define(self):
+        class _ReadOnlyXform(_ReadOnlyModel):
+            class Meta:
+                managed = False
+                db_table = 'logger_xform'
+                verbose_name = 'xform'
+                verbose_name_plural = 'xforms'
+
+            xml = models.TextField()
+            user = models.ForeignKey(User, null=True)
+            id_string = models.SlugField()
+            date_created = models.DateTimeField()
+            date_modified = models.DateTimeField()
+            uuid = models.CharField(max_length=249, default=u'')
+
+        class _ReadOnlyInstance(_ReadOnlyModel):
+            class Meta:
+                managed = False
+                db_table = 'logger_instance'
+                verbose_name = 'instance'
+                verbose_name_plural = 'instances'
+
+            xml = models.TextField()
+            user = models.ForeignKey(User, null=True)
+            xform = models.ForeignKey(_ReadOnlyXform, related_name='instances')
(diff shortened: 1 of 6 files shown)


kpi/deployment_backends/kc_reader/utils.py
```diff
@@ -0,0 +1,68 @@
+from django.contrib.auth.models import User
+from django.db import models
+from django.db import ProgrammingError
+
+
+class __ReadOnlyModelError(ValueError):
+    pass
+
+
+class __ReadOnlyModel:
```
jnm: There are other methods that write, e.g. `update()`, but this is a good safety net for most accidents

kpi/deployment_backends/kc_reader/utils.py
```diff
@@ -0,0 +1,68 @@
...
+            return func(*args, **kwargs)
+        except ProgrammingError as e:
+            raise ProgrammingError('kc_reader error accessing kobocat '
+                                   'tables: {}'.format(e.message))
+    return _wrapper
+
+
+@wrap_kc_reader_call
+def instance_count(xform_id_string, user_id, version_id=None):
+    return __ReadOnlyInstance.objects.filter(xform__user_id=user_id,
```
jnm: `Instance` has a foreign key to `XForm`; not sure if something like this would be faster, but guessing it might take advantage of a DB index: `return __ReadOnlyXform.objects.get(xform_id_string, user_id).instances.all()`.

What does `version_id` do?

kpi/deployment_backends/kobocat_backend.py
```diff
@@ -349,3 +349,13 @@ def get_enketo_survey_links(self):
             except KeyError:
                 pass
         return links
+
+    def _submission_count(self):
+        _deployment_data = self.asset._deployment_data
+        id_string = _deployment_data['backend_response']['id_string']
+        # avoid migrations from being created for kc_reader mocked models
+        # there should be a better way to do this, right?
```
jnm: KC uses a cache field on `XForm`: `num_of_submissions = models.IntegerField(default=0)`. Then, it has this method to get the submission count:
```
    def submission_count(self, force_update=False):
        if self.num_of_submissions == 0 or force_update:
            count = self.instances.filter(deleted_at__isnull=True).count()
            self.num_of_submissions = count
            self.save()
        return self.num_of_submissions
```
I can't find where this is actually incremented upon submission (I can see where it's updated on deletion, though), but maybe that's only because GitHub's search is bad.

kpi/deployment_backends/kc_reader/utils.py
dorey: * i timed it with the change and it came out a little bit slower with the changes
* `version_id` does nothing and has been removed
* i'll be adding `filter(deleted_at__isnull=False)` (good catch)

kpi/deployment_backends/kobocat_backend.py
dorey: KC keeps it updated [here](https://github.com/kobotoolbox/kobocat/blob/master/onadata/apps/logger/models/instance.py#L61). It ran pretty quickly when I tested it. I suspect an index or two were added since `num_of_submissions` was first needed

kpi/deployment_backends/kc_reader/utils.py
dorey: Yeah. I think making the models private_ish_ is the main safeguard. Overriding the methods is secondary.

kpi/deployment_backends/kc_reader/shadow_models.py
```diff
@@ -0,0 +1,78 @@
...
+        abstract = True
+
+    def save(self, *args, **kwargs):
+        raise ReadOnlyModelError('Cannot save read-only-model')
+
+    def delete(self, *args, **kwargs):
+        raise ReadOnlyModelError('Cannot delete read-only-model')
+
+
+class LazyModelGroup:
```
dorey: This is the best way I could find to prevent `manage.py makemigrations` from recognizing these models.

Issue #842, opened 2016-07-28
Project title should be mandatory


tinok:
When creating a new project, the Project Title field should be mandatory. If a user tries to create a project without specifying the title they should see an error "Please enter a title for your project".

PR #843, opened 2016-07-28
 Save project metadata and allow it to be edited

### LINKED ISSUE #822: Allow user to view and edit all project-specific metadata fields in KPI
Refers to the metadata fields created in #645. 

This should be displayed above the KC settings iframe on the Settings page.

jnm:
Closes #822, but still needs some styling help


tinok:
Also closes #842.

PR #844, opened 2016-07-29
Reports with mongo

dorey:
This contains some quick-fixes to get this branch demoable.

TODO: modify formpack to allow alternative row identifiers (other than the `name` field)


dorey:
Moved into #870

PR #870, opened 2016-08-17
Reports

pmusaraj:


hub/models.py
```diff
@@ -38,6 +38,8 @@ class ExtraUserDetail(models.Model):
     user = models.OneToOneField(settings.AUTH_USER_MODEL, related_name='extra_details')
     data = JSONField(default={})
 
+    # @jnm: Can we reassess how to do this and not break tests?
```
dorey: This would pass tests `KOBOCAT_URL` was falsey, but because it's a string (to prevent breaking links elsewhere) it breaks.

PR #896, opened 2016-08-29
Autoreport mock data test

dorey:
No changes to the public-facing code in this PR, but this should be OK to merge if tests pass.


dorey:
changing this PR to point to develop

PR #901, opened 2016-08-30
Move KC profile sync to serializer, management cmd

### LINKED ISSUE #821: Allow user to view and edit all profile settings (in KPI)
This includes 

* all the fields available through #550
* Reset password button/view
* metadata field (we can possibly do without this one)

This should include a migration of all user profile details the user previously specified in KC into KPI's database.

jnm:
* Fixes #821 
* Does not break tests
* Does not prevent KPI user creation when KC tables do not exist
* Assembled in the USA from foreign and domestic components :ok_hand:

### DIFF
--- a/hub/models.py
+++ b/hub/models.py
@@ -4,9 +4,6 @@ from django.conf import settings
 from markitup.fields import MarkupField
 from jsonfield import JSONField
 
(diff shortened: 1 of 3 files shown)


dorey:
🌠 🏆

PR #926, opened 2016-09-19
Codacy Python test coverage metrics.

esmail:


magicznyleszek:
@jnm is this PR going anywhere? Maybe we should drop it since 2 years passed and there are conflicts already

PR #962, opened 2016-10-10
Don't log an error when KC cannot find XLS

### LINKED ISSUE #945: sync_kobocat_xforms spams the log with 404 errors
`sync_kobocat_xforms`, which runs as a periodic Celery task, tries to sync KC forms whose XLS files are missing every time it is invoked. This clutters up logs.

Now that https://github.com/kobotoolbox/kobocat/pull/280 is merged, `sync_kobocat_xforms` should recognize a 404 when querying `/api/v1/forms/.../form.xls` as an irrecoverable error. The failure should be stored persistently and the form not retried again.

jnm:
Fixes #945

### DIFF
--- a/kpi/management/commands/sync_kobocat_xforms.py
+++ b/kpi/management/commands/sync_kobocat_xforms.py
@@ -306,8 +306,11 @@ class Command(BaseCommand):
                                 response.status_code)
                         ]
                         print_tabular(*error_information)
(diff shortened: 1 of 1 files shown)


dorey:
These forms do not have an associated xlsform in kobocat. This will not be necessary in the long run but for now, it unclutters the logs and helps us to address real issues.

PR #981, opened 2016-10-18
Allow injection of NPM cache to speed up and stabilize Docker builds.

esmail:
Closes kobotoolbox/tasks#103, kobotoolbox/tasks#100.


esmail:
Attempted workaround to the often-failing `npm install` step of Docker builds. Unfortunately, resulted in longer build times due to checking the cache into the Docker build context.

The [alternative workaround](https://github.com/kobotoolbox/kpi/pull/985), which attempts to clean up ungraceful PhantomJS installation failures, appears to be working well.

PR #989, opened 2016-10-21
Only test PR commits once in Travis.

esmail:
Seems like GitHub's status API has been down; trying this PR yet again...


esmail:
As mentioned in the code, details here: http://stackoverflow.com/a/31882307/1877326

PR #1010, opened 2016-11-02
Skipping content-type checking in the browser

dorey:
The goal was to prevent random (and large) files from being uploaded, but this `content-type` check has prevented a number of valid spreadsheet submissions out in the wild.


esmail:
Travis failure is expected without removal of `ttfautohint`.

PR #1014, opened 2016-11-02
Allow asset/collection lookup when `KPI_PREFIX='/'`.

esmail:
Technically, this was tested on top of #1010.


esmail:
Closes #1011.

Issue #1039, opened 2016-11-23
cachebuster prevents compilemessages from running


rhunwicks:
When I run `python manage.py compilemessages` using just the `kobo.settings` file, I get:

```
EnvironmentError: django-cachebuster could not find a '.git' directory in your project path
```

Django-Cachebuster was last updated in 2012 and its repo says *DJANGO 1.4+ USERS: You should probably look to use Django's included static template tag*. Also, kpi is already using WhiteNoise which handles cache attributes for static files.

Therefore, I think that we should remove django-cachebuster.

jnm:
'twas fixed in https://github.com/kobotoolbox/kpi/pull/1040!

PR #1040, opened 2016-11-25
Remove django-cachebuster - see #1039

rhunwicks:


dorey:
I think this is fine. I put this in there for cache busting and then found it might be helpful for support requests on different installations, but there are better ways to do this that don't use outdated python libraries.

dorey:
Tests passing; merging.

Thanks @rhunwicks !

PR #1061, opened 2016-12-09
Optionally treat proxied, non-uWSGI connections as secure.

esmail:
Closes kobotoolbox/tasks#105.


esmail:
Depends on `SECURE_PROXY_SSL_HEADER` e.g. in [`kobo-docker`](https://github.com/kobotoolbox/kobo-docker/blob/74436d/docker-compose.server.yml#L97)

PR #1086, opened 2017-01-12
Do not assume that all exceptions have a `response`

jnm:
If we encounter an unexpected exception, we want to [re-raise](https://github.com/kobotoolbox/kpi/blob/b87c49ca8c7fb95d3917135ca71683a21a89c9b9/kpi/deployment_backends/kobocat_backend.py#L293) that original exception, not accidentally raise an `AttributeError` instead :blush:

### DIFF
--- a/kpi/deployment_backends/kobocat_backend.py
+++ b/kpi/deployment_backends/kobocat_backend.py
@@ -286,7 +286,7 @@ class KobocatDeploymentBackend(BaseDeploymentBackend):
                 'version': self.asset.version_id,
             })
         except KobocatDeploymentException as e:
-            if e.response.status_code == 404:
+            if hasattr(e, 'response') and e.response.status_code == 404:
                 # Whoops, the KC project we thought we were going to overwrite
                 # is gone! Try a standard deployment instead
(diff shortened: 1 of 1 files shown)


kpi/deployment_backends/kobocat_backend.py
```diff
@@ -286,7 +286,7 @@ def redeploy(self, active=None):
                 'version': self.asset.version_id,
             })
         except KobocatDeploymentException as e:
-            if e.response.status_code == 404:
+            if hasattr(e, 'response') and e.response.status_code == 404:
```
jnm: Much cleaner, but it's `response.status_code` that contains the `404`, not `response` itself. This pattern of getting `first.second.third` while dodging an `AttributeError` if `second` doesn't exist seems really common; maybe there's a more Pythonic way than what I wrote?

esmail: :facepalm: My bad!

And `if getattr(e, 'response', object()).__dict__.get('status_code') == 404` somehow doesn't have the same ring (or resolution of `@property`/`@getter`) to it :)

PR #1092, opened 2017-01-17
Archive and delete asset from list

pmusaraj:


pmusaraj:
Ok, will do that and post this for QA.

Issue #1087, opened 2017-01-12
OtherFormBuilderRedirectMiddleware causing redirect loop


jnm:
Observed on OCHA `mp03`:

* `KPI_PREFIX` set to `/forms`
* Logged in as user who prefers KPI
* [`if prefix != preferred_prefix`](https://github.com/kobotoolbox/kpi/blob/fd4e3f6c0e87b49a741205d7aea841a6fa793b9b/hub/middleware.py#L21) evaluated to `True` resulting in redirect, regardless of whether request URL ended with `forms` or `forms/`
* Logged values of `prefix` and `preferred_prefix`: found that `prefix` was `/forms` and `preferred_prefix` was `/forms/`
* Changing `KPI_PREFIX` to `/forms/` did not help
* Calling `strip('/')` on both prefixes seems a suitable solution

esmail:
I'm sure you already noticed, but for posterity, `KPI_PREFIX`, and [therefore](https://github.com/kobotoolbox/kpi/blob/fd4e3f6c0e87b49a741205d7aea841a6fa793b9b/hub/middleware.py#L12) `preferred_prefix`, are [forced](https://github.com/kobotoolbox/kpi/blob/fd4e3f6c0e87b49a741205d7aea841a6fa793b9b/kobo/settings.py#L181) to begin and end with single `/` characters.

jnm:
@esmail, I actually didn't notice the [`settings.py` change](https://github.com/kobotoolbox/kpi/pull/1042). It appears to be causing another problem: https://github.com/kobotoolbox/kpi/issues/1091

Issue #1041, opened 2016-11-25
Running in a subdirectory using KPI_PREFIX breaks Whitenoise


rhunwicks:
KPI uses Whitenoise to serve static files, and it works out whether to attempt to serve the file itself by comparing the requested url to `settings.STATIC_URL`.

At the moment `kobo/settings.py` sets this value using:

```
# KPI_PREFIX should be set in the environment when running in a subdirectory
if KPI_PREFIX and KPI_PREFIX != '/':
    STATIC_URL = '{}/{}'.format(KPI_PREFIX, STATIC_URL)
    from django.conf.global_settings import LOGIN_URL
    LOGIN_URL = '{}/{}'.format(KPI_PREFIX, LOGIN_URL)
```

`STATIC_URL` is already set to `/static/`

Therefore, if `KPI_PREFIX` is set to `/forms` then the resulting `STATIC_URL` is `/forms//static`.

When a request comes in for `/forms/static/js/somefile.js` Whitenoise compares the request to `STATIC_URL` and doesn't find a match, and so doesn't serve the file.

The way that Kobocat updates settings based on `KOBOCAT_ROOT_URI_PREFIX` is much more reliable - see `onadata/settings/kc_environ.py`.

I will prepare a PR that implements that approach for KPI.

jnm:
I believe this was closed by https://github.com/kobotoolbox/kpi/pull/1042

PR #1093, opened 2017-01-17
Strip trailing / from KPI_PREFIX and DKOBO_PREFIX

### LINKED ISSUE #1091: Replacing a form from XLS returns Resolver404 error
Console shows import status as error: 

    {"status":"error","uid":"iYmu8VQY4XdSibgAiTCYTC","messages":{"error_type":"Resolver404","error":{"path":"assets/adypZWmaQXEMzrESqtU33z/"}},"date_created":"2017-01-17T16:25:52.794872Z"}

UI shows:
image

jnm:
This was the norm (but wasn't enforced) prior to df4e058, which then rewrote
KPI_PREFIX so it always began **and ended** with a single slash. That change
precipitated two bugs (#1087 and #1091). This commit fixes #1091 while
attempting to keep the fix for #1041 intact.

### DIFF
--- a/kobo/settings.py
+++ b/kobo/settings.py
@@ -174,15 +174,17 @@ STATIC_URL = '/static/'
 # Following the uWSGI mountpoint convention, this should have a leading slash
 # but no trailing slash
 KPI_PREFIX = os.environ.get('KPI_PREFIX', 'False')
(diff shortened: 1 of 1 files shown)


esmail:
Thanks, @jnm. Was thinking of the same solution.

PR #1094, opened 2017-01-17
Specify the invalid username message

m4tx:
The original message does not indicate a reason for rejecting e.g. a one letter username.


jnm:
Thanks!

PR #1100, opened 2017-01-21
Fix duplicated usernames in Sharing Permissions

m4tx:
When a user had edit permissions they appeared twice in the form Settings in Sharing Permissions panel.


dorey:
Good catch. Thanks for fixing this.

Issue #1102, opened 2017-01-24
Form builder UI issues with library questions drag-and-drop


pmusaraj:
Two issues reported by a user:

### 1. Sometimes dragging and dropping items from the sidebar doesn't work. And sometimes it only works when adding above another question, and not below it. 

_Proposed solution_: better demarcate the drop area and add padding to the last list element, so that items can be dropped after it. This is better explained using a screencast
[screencast 2017-01-24 15-26-08.mp4.zip](https://github.com/kobotoolbox/kpi/files/727940/screencast.2017-01-24.15-26-08.mp4.zip)

### 2. When there is an error while dragging and dropping to the form builder, no error message is shown to user. 

_Proposed solution_: include the error message using alertify.

Pedro719:
I agree with your solution tom (1) better demarcating the drop area and adding padding so items can be dropped after it. I think we can get a better understanding of how to execute this by looking at Trello when the user is moving cards from one board to another.

<img width="348" alt="screen shot 2017-01-24 at 4 45 19 pm" src="https://cloud.githubusercontent.com/assets/25228823/22270059/fdbfa102-e25b-11e6-8b95-a204c7cd900f.png">



[trello 2.zip](https://github.com/kobotoolbox/kpi/files/728146/trello.2.zip)

Pedro719:
When dragging block of questions to form builder I received an error as expected.
[Test 1102.zip](https://github.com/kobotoolbox/kpi/files/742588/Test.1102.zip)

Issue #782, opened 2016-06-28
Standardize all kpi error modals


tinok:
Deployment failure: 

![image](https://cloud.githubusercontent.com/assets/673511/16402235/fcf9631e-3cb8-11e6-941c-37cd0e6cf2bb.png)

Formbuilder preview error: 

![image](https://cloud.githubusercontent.com/assets/673511/16402262/1a8678ea-3cb9-11e6-94fe-6e91ada735a3.png)

Import failure: 

![image](https://cloud.githubusercontent.com/assets/673511/16402269/2ce37ab0-3cb9-11e6-93af-4796d9828474.png)

Deletion and cloning modals: 

![image](https://cloud.githubusercontent.com/assets/673511/16402360/02347ade-3cba-11e6-8ea7-483be4377c7e.png)

![image](https://cloud.githubusercontent.com/assets/673511/16402375/1c7b8cd4-3cba-11e6-954d-41c99a6a0f10.png)

southisack:
Here are some comments and changes that I noticed. 
[782.zip](https://github.com/kobotoolbox/kpi/files/742784/782.zip)

southisack:
Here are my comments on the dialog

![782](https://cloud.githubusercontent.com/assets/4983734/22481402/1cd4d73a-e7c3-11e6-818d-e052ed461bd2.jpg)

pmusaraj:
Hmm, these are all valid, but I think we should handle them at a second stage, when we approve the whole visual interface. (It would make little sense to updated fonts and form styles only for these dialogs...)

Pedro719:
Dialogs are clear, I was wondering for the "unable to deploy" with code beneath it should we give the user a button or a link that allows them to report the issue from that instance if they want?

pmusaraj:
The "report issue" would fall under the "new feature" category, to be handled separately from this issue.

PR #1125, opened 2017-01-31
Issues #1102, #782

pmusaraj:


jsapp/js/actions.es6
```diff
@@ -400,31 +400,31 @@ actions.resources.deployAsset.failed.listen(function(data, dialog_or_alert){
...
     var msg;
     if (data.status == 500 && data.responseJSON && data.responseJSON.error) {
-      msg = `<code><pre>${data.responseJSON.error}</pre></code>`;
+      msg = `<pre>${data.responseJSON.error}</pre>`;
     } else if (data.status == 500 && data.responseText) {
-      msg = `<code><pre>${data.responseText}</pre></code>`;
+      msg = `<pre>${data.responseText}</pre>`;
     } else {
-      msg = t('please check your connection and try again.');
+      msg = t('Please check your connection and try again.');
```
dorey: Minor changes like these should go into the [`msgstr` value of the djangojs.po file](https://github.com/kobotoolbox/form-builder-translations/blob/master/en/LC_MESSAGES/djangojs.po#L40). The idea behind having lower cased strings is to keep the coded strong as simple as possible and make spelling and grammar changes to the expanded `msgstr` values. (This is as an alternative to using IDs as the msgids, like `please_check_connection`.)

We do have a mix between capitalized and lower cased strings. But if we want to go through and change them all then it would be a bit of an ordeal with transifex and the existing translations. So, I'd suggest when they are one way, we keep it that way.

pmusaraj: Ah, I wasn't aware of the djangojs.po process. I can revert this (and a few other similar changes) and make them in the djangojs.po file.

PR #1128, opened 2017-01-31
Tasks/114  login lockout

esmail:
Closes kobotoolbox/tasks#114.


esmail:
Still need to verify on a server that [`AXES_REVERSE_PROXY_HEADER`](http://django-axes.readthedocs.io/en/latest/configuration.html) is configured correctly.

PR #1134, opened 2017-02-06
closes #1072

### LINKED ISSUE #1072: reports show autonamed questions as "untitled"
Reports feature shows labels as "untitled" on questions with no explicit names.

dorey:
by using the asset's "$autoname" key first, this will associate the label with the correct identifier.


jnm:
Tested on https://kf.du.kbtdev.org

PR #1137, opened 2017-02-07
Import `LOGIN_URL` at beginning of `settings.py`

### LINKED ISSUE #1136: Login without `?next=` redirected to `/forms/accounts/profile/`
1. Visit https://kf.kobotoolbox.org/forms/accounts/login/. **NB:** there's no `?next=` in this URL;
1. Log in;
1. Notice that you are redirected to https://kf.kobotoolbox.org/forms/accounts/profile/, which displays a Not Found error.

jnm:
Avoids accidentally overwriting a valid setting with whatever default is defined in `django.conf.global_settings`. Removes the `LOGIN_REDIRECT_URL` import outright, since that setting [is defined manually](https://github.com/kobotoolbox/kpi/blob/462a2a88ca1f622ea86191deaa761d10e2406359/kobo/settings.py#L59). Fixes #1136

### DIFF
--- a/kobo/settings.py
+++ b/kobo/settings.py
@@ -15,6 +15,7 @@ import subprocess
 
 from django.conf import global_settings
 from django.conf.global_settings import LANGUAGES as _available_langs
+from django.conf.global_settings import LOGIN_URL
 from django.utils.translation import get_language_info
 import dj_database_url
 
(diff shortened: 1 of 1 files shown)


jnm:
Now running at https://kf.du.kbtdev.org/accounts/login/. For testing, compare this URL's behavior to that of https://kf.kobotoolbox.org/forms/accounts/login/ as described in https://github.com/kobotoolbox/kpi/issues/1136.

PR #1140, opened 2017-02-11
Display full form title in Preview

m4tx:
~~This requires merging kobotoolbox/formpack#121 (I will update requirements.* after this, as it requires commit hash, the checks should pass then).~~ (both done)


dorey:
👍

PR #1141, opened 2017-02-11
Show form logo in the preview

m4tx:
This first requires kobotoolbox/enketo-express#643 to be merged.

### DIFF
--- a/kpi/views.py
+++ b/kpi/views.py
@@ -514,6 +514,12 @@ class AssetSnapshotViewSet(NoUpdateModelViewSet):
                                       request=request,
                                       ),
                             )
+            deployment_data = snapshot.asset._deployment_data
+            if 'backend_response' in deployment_data:
+                preview_url += '{}/xformsManifest/{}'.format(
+                    settings.KOBOCAT_URL,
+                    deployment_data['backend_response']['metadata'][0]['xform'])
+
             return HttpResponseRedirect(preview_url)
         else:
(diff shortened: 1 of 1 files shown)


dorey:
Generally speaking, we don't want to add dependencies into complex parts of the `_deployment_data` structure. If we were to do it, i would suggest putting it into a method of the asset model. That being said, I don't quite understand how this works. It appends the deployed form's `xformsManifest` URL onto the existing URL?

So wouldn't it try to redirect to something like this?

`"https://enketo/webform/preview?form=http://kpi/asset_snapshots/s56789.xmlhttp://kc/username/forms/31415/form.xml"`

If the URL that came out of it was valid, I believe this would just preview the most recently deployed form, which is not necessarily the latest version that's in the draft stage of KPI.

We are planning to implement media files on the KPI side so we will probably want to link to a logo that is attached to the draft.

kpi/views.py
```diff
@@ -514,6 +514,12 @@ def preview(self, request, *args, **kwargs):
                                       request=request,
                                       ),
                             )
+            deployment_data = snapshot.asset._deployment_data
+            if 'backend_response' in deployment_data:
+                preview_url += '{}/xformsManifest/{}'.format(
```
dorey: Based on enketo-express#643, I'm guessing this is supposed to pass `&manifestUrl=` in the preview_url value

Issue #1054, opened 2016-12-05
XLS imports don’t work on Beta with Arabic forms


emilykferguson:

magicznyleszek:
I was able to import arabic xls file without problem on https://kf.master.kbtdev.org - probably fixed by https://github.com/kobotoolbox/kpi/pull/1913 or earlier

Issue #949, opened 2016-09-30
multi-language forms are not editable in the form builder


dorey:
Languages are concatenated into one. (Which is the behavior one would expect with the recent changes to the format of asset.content)

The fix:
- pass a flattened structure to the form builder instead of the expanded content.
- long term: restructure the form builder to handle translations properly

PR #1143, opened 2017-02-13
Formbuilder handles translations

### LINKED ISSUE #787: Forms with translations can't be previewed from the form builder
(in 729 branch)

dorey:
closes #787


dorey:
There is a test here that is not being run by coveralls at the moment which would prevent the % from decreasing by this PR. I would like to run all the django tests with `pytest`. But for now, they are run separately by `py.test -sx --ignore=node_modules`

dorey:
Related (and mostly closing) #1144 #949 and possibly #1054

PR #1168, opened 2017-02-24
`AssetVersion` sharing permissions.

esmail:


kpi/views.py
```diff
@@ -525,27 +521,25 @@ def preview(self, request, *args, **kwargs):
...
             return AssetVersionListSerializer
         else:
             return AssetVersionSerializer
 
-    def get_queryset(self):
-        _asset_uid = self.get_parents_query_dict()['asset']
-        _deployed = self.request.query_params.get('deployed', None)
-        _queryset = self.model.objects.filter(asset__uid=_asset_uid)
-        if _deployed is not None:
-            _queryset = _queryset.filter(deployed=_deployed)
```
jnm: @dorey, what was the original intent of having `deployed` in the query string? Was it just to toggle between showing all versions and just those versions that are deployed? Or should it provide for displaying only the undeployed versions as well? It looks like `deployed` won't be recognized at all after merging this PR; maybe you worked that out with @esmail in the office?

dorey: Thanks for noting that. Yes, the `deployed` query string was supposed to filter them down (I don't see a situation where we would want to see un-deployed versions.) When @esmail and I left off, we were going to go over this with you on Monday.

Adding `donotmerge`, for now.

dorey: Since the `deployed` query param is not used in any requests or tests I'll remote this `donotmerge` so we can get this in `master`.

PR #1190, opened 2017-03-10
Moving webpack configs into `webpack` directory

dorey:


esmail:
@dorey, I can't figure out what the cause of [`Error: Cannot find module '/srv/src/kpi/webpack/prod.config.js'`](https://jenkins.kobotoolbox.org/job/KPI%20Docker%20build/572/console) was (another Docker mystery, @jnm?), but [`cd "${KPI_SRC_DIR}"`](https://github.com/kobotoolbox/kpi/pull/1190/commits/75cccf4fa140fc2559a2bbca8573b6323a9ac98e) is redundant due to [`WORKDIR ${KPI_SRC_DIR}/`](https://github.com/kobotoolbox/kpi/blob/2b5e99a3d36424f6f15feb84d06b4fd3e3eaacf7/Dockerfile.koboform_base#L46), so I've reverted it.

dorey:
@esmail , I think we should have kept this open since there's an open question about why it failed a number of times. I am happy to see it building successfully

Though now that it's merged and closed it let's keep it closed and hope the problem does not come back.

PR #1193, opened 2017-03-15
Relax `view_func` comparison from `is` to `==`

### LINKED ISSUE #1192: New Relic breaks `switch_builder` view
With New Relic installed, [`view_func is switch_builder`](https://github.com/kobotoolbox/kpi/blob/3734b68eb3a16f62de3271ee99b6bdb4d0727e4f/hub/middleware.py#L34) always evaluates to `False`, even when the view function really _is_ `switch_builder`. New Relic must wrap each view function for profiling purposes. Example of the comparison failing:

| | `repr()` | `type()` | `id()` |
| - | - | - | -
`view_func` | <function switch_builder at 0x7f42764bb7d0> | <class 'newrelic.common.object_wrapper.FunctionWrapper'> | 139923318464928
`switch_builder` | <function switch_builder at 0x7f42764bb7d0> | <type 'function'> | 139923429242832

jnm:
Fixes #1192

### DIFF
--- a/hub/middleware.py
+++ b/hub/middleware.py
@@ -31,7 +31,7 @@ class OtherFormBuilderRedirectMiddleware(object):
         ''' Using process_view instead of process_request allows the resolver
         to run and return 404 when appropriate, instead of blindly returning
         302 for all requests '''
-        if view_func is switch_builder:
+        if view_func == switch_builder:
             # Never redirect the view that changes form builder preference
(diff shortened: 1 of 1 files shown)


esmail:
Interesting that this works. So I guess the New Relic function wrapper throws us a bone and overrides `__eq__`.

PR #1195, opened 2017-03-21
Rudimentiary permissions sync from KC

### LINKED ISSUE #1161: `sync_kobocat_xforms` only considers owned KC forms
1. User Jill shares form Survey with user Bob in KC;
1. User Bob logs into KC and sees form Survey;
1. User Bob logs into KPI and form Survey never appears.

One way to deal with this would be to grant `view_asset` in KPI to anyone who has some level of access to a KC `XForm`. However, that leaves an unsolved problem about what happens when Jill stops sharing the form Survey with Bob in KC:

1. Jill gives Bob view access to an XForm in KC;
1. `sync_kobocat_xforms` finds Jill's corresponding KPI Asset and gives Bob view access to it;
1. Jill removes Bob's view access in KC;
1. `sync_kobocat_xforms` cannot tell whether the KPI `view_asset` permission was added by a user (and thus should be left alone) or was added by `sync_kobocat_xforms` itself (and thus should be removed to match KC)

jnm:
Enabled with `--grant-kpi-view-permission` argument to `sync_kobocat_xforms` management command or `SYNC_KOBOCAT_XFORMS_GRANT_KPI_VIEW_PERMISSION`, which governs the periodic Celery task's behavior.

WARNING: this is a one-way operation! Users will retain their `view_asset` permission in KPI even if their KC access is revoked.

Currently testing on https://kf.tri.kbtdev.org. Closes #1161

### DIFF
--- a/kobo/settings.py
+++ b/kobo/settings.py
@@ -318,6 +318,15 @@ if 'KOBOCAT_URL' in os.environ:
     # Create/update KPI assets to match KC forms
     SYNC_KOBOCAT_XFORMS_PERIOD_MINUTES = int(os.environ.get('SYNC_KOBOCAT_XFORMS_PERIOD_MINUTES',
                                                             '30'))
+    # Create view permission in KPI when KC form is shared? Default is False.
+    # WARNING: this is a one-way operation! Users will retain their
+    # `view_asset` permission in KPI even if their KC access is revoked.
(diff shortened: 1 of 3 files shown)


jnm:
Likely to be superseded soon

jnm:
Superseded by #1210

PR #1224, opened 2017-04-21
Remove double slash added to ENKETO_SERVER

jeverling:
A trailing slash is added to `ENKETO_SERVER` in the settings module. Another one is introduced between the domain and path, resulting in a double slash.


jeverling:
fixes #1225

PR #1226, opened 2017-04-21
Deprecate Material Design Lite

pmusaraj:


jsapp/js/components/accountSettings.es6
```diff
@@ -144,6 +152,9 @@ export var AccountSettings = React.createClass({
       background: `#${stringToColor(accountName)}`
     };
 
+    let bHeight = $('textarea#bio').prop('scrollHeight');
```
dorey: Note: In general, this kind of call to the DOM should be moved out of the render function

jsapp/js/components/accountSettings.es6
```diff
@@ -235,7 +246,11 @@ export var AccountSettings = React.createClass({
               <bem.AccountSettings__item>
                 <label>
                   {t('Bio')}
-                  <textarea value={this.state.bio}
+                  <textarea 
+                    value={this.state.bio}
+                    rows="1"
+                    id="bio"
+                    style={{height: bioHeight + 'px'}}
```
dorey: I think `+ 'px'` is unnecessary here because it is implied.

jsapp/js/components/drawer.es6
```diff
@@ -180,9 +179,6 @@ var Drawer = React.createClass({
         </div>
       </bem.Drawer>
       );
-  },
-  componentDidUpdate() {
-    mdl.upgradeDom();
```
dorey: Another reason this PR needs to be merged... no more calls to `mdl.upgradeDom()` _every render_ !

jsapp/js/components/formLanding.es6
```diff
@@ -170,7 +169,7 @@ var FormLanding = React.createClass({
                   {/*<bem.FormView__label m='questions'></bem.FormView__label>*/}
                   <bem.FormView__label m='clone'>
                       <bem.FormView__link m='clone'
-                          data-version-id={item.version_id}
+                          data-version-id={item.uid}
```
dorey: Should we rename this prop to `data-form-uid`? Or is the `item` a "version" object?

jsapp/js/components/sharingForm.es6
```diff
@@ -22,8 +22,8 @@ var UserPermDiv = React.createClass({
   ],
   PermOnChange(perm) {
     var cans = this.props.can;
-    var permName = perm.value;
-    if (permName != '') {
+    if (perm != null) {
```
dorey: equivalent to `if (perm)`

pmusaraj: Ah yes, thanks! (Fixed.)

jsapp/js/components/formLanding.es6
pmusaraj: This is the list of form versions, so yes, `item` is a "version" object. And `data-version-id` matches what `saveCloneAs` expects in `mixins.es6`. I think it makes sense to leave it as version id.

jsapp/js/components/accountSettings.es6
pmusaraj: I replaced this with the `react-autosize-textarea` component. Works well, and allows for much cleaner code.

PR #1233, opened 2017-04-26
Notify users when survey save has failed

### LINKED ISSUE #1109: Form builder does not fail when user tries to save but server is down

dorey:
* shows an updated button and a message telling the user not to close the tab.
* displays a brief alertify message if the survey is scrolled.

* Note: falls short of saving the changes (e.g. to localStorage) which could prevent data loss but could also lead to privacy leaks of things stored in the browser.

Closes #1109


dorey:
This probably needs to be part of a greater design discussion. But the 3 methods implemented show some of the ways we can notify the user of the failed save.

pmusaraj:
I added a commit with the UI following our current conventions for error messages. Can you review?

PR #1236, opened 2017-04-26
fixes #1234

### LINKED ISSUE #1234: Cloning question in Form Builder fails
It adds an empty question with a console error: `TypeError: undefined is not an object (evaluating 'this.collection.getSurvey')`. Error is present in our production, OCHA as well as the `develop` branch.

dorey:
Assigns the parent collection to a newly cloned ChoiceList


jnm:
Tested on https://kf.du.kbtdev.org

Issue #1240, opened 2017-04-29
Question created in the form builder interface should be required=true


dorey:
This has been the default behavior, because requiring questions leads to complete responses.

_However_ , we need to undo the way we did this because it has caused `undefined` values or questions with no specified `required` to default to true.

So, in fixing #636, we are introducing this bug. which can be fixed in the `view.row.coffee` file, rather than in the model row's defaults.

magicznyleszek:
I couldn't reproduce this by testing, so I'm assuming this is fixed.

PR #1241, opened 2017-04-29
closes #636, but opens #1240

### LINKED ISSUE #636: formbuilder marks questions as required if setting was NULL in XLS
To reproduce: 
1. Import a form from XLS with three questions, marked in `required` as `true`, `false`, and `''` (blank)
2. Open formbuilder, save, close
3. Download form

Expected: values in 'required' column were the same as in imported file

Instead: Question that was `''`' (blank) is now required.

dorey:
* When no `required` value is set, it will default to
  "false".

### DIFF
--- a/jsapp/xlform/src/model.base.coffee
+++ b/jsapp/xlform/src/model.base.coffee
@@ -82,6 +82,14 @@ module.exports = do ->
         parent = parent._parent
       parent
 
+  _innerValue = (val)->
+    # occasionally, the value passed to rowDetail
+    # is an object, which evaluates to true
+    if _.isObject(val)
+      val.value
+    else
+      val
+
   class base.RowDetail extends base.BaseModel
     idAttribute: "name"
     validation: () =>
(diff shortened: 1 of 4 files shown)


jsapp/xlform/src/model.configs.coffee
```diff
@@ -218,7 +218,7 @@ module.exports = do ->
       value: ""
       _hideUnlessChanged: true
     required:
```
dorey: Biggest change in this PR is this (and corresponding test changes)

New questions will no longer have `"required=true"` when the model is initialized.

jsapp/xlform/src/model.base.coffee
```diff
@@ -107,7 +115,16 @@ module.exports = do ->
       # In the quick fix, this is done in the view for 'required' rowDetails
       # (grep: XLF.configs.truthyValues)
 
-      if value not in [undefined, false, null]
```
dorey: `_innerValue(...)` was necessary because in certain cirtumstances, value is set to `{"value":false, "_hideUnlessChanged": true}`, which would evaluate to `false`.

I also found that changing the order of the "if" conditions caught an edge case, and didn't change behavior negatively.

PR #1249, opened 2017-05-03
Upgrade New Relic monitor

jnm:
The New Relic issue of consequence was fixed by https://github.com/kobotoolbox/kpi/pull/1193, but I'm upgrading the monitor because we said we would (https://humanitarian.atlassian.net/browse/OPS-2518).


jnm:
PS (@esmail): `make --always-make pip_compile ARGS='--upgrade-package=newrelic'` worked really well

PR #1267, opened 2017-05-08
Form landing UI

pmusaraj:
Update form landing screen UI based on designs. Based off `deprecate-mdl`.


jsapp/js/components/formLanding.es6
```diff
@@ -106,55 +110,53 @@ var FormLanding = React.createClass({
...
-          <div className="question-count">{this.state.summary.row_count}</div>
-        </bem.FormView__cell>
-        <bem.FormView__cell m={['question-list']}>
-          {survey.map((s, i)=>{
-            if (s.label == undefined) return false;
-            var icon = icons._byId[s.type];
-            if (!icon) {
-              return false;
-            }
+  // renderQuestionsSummary () {
```
dorey: I'm not attached to this code if we want to just delete it that's fine.

jsapp/js/components/formLanding.es6
```diff
@@ -181,16 +181,156 @@ var FormLanding = React.createClass({
...
             </button>
           </bem.FormView__cell>
         }
       </bem.FormView__row>
       );
   },
+  renderCollectData () {
+    var deployment__links = this.state.deployment__links;
+
+    var available__links = {
```
dorey: The order of these `available__links` is not necessarily preserved. The best way to do this would be a `es6-map` which is equivalent to python's `OrderedDict`

dorey: fixed in 11e39f

PR #1287, opened 2017-05-19
Backend changes for the image gallery

preisman:
Here are the backend changes for the image gallery thus far. They cover:

- Returning a list of attachments, optionally filtered by media type.
- Returning the details for a specific attachment.
- Pagination based on the total number of attachments. Can use either limit/offset pagination or page/page_size pagination. Default limit/page_size is 10.
- Loading raw images inside KPI when the filename is supplied in the detail link.
- Image resizing on demand, including fallback options for images existing in local KPI storage (if they haven't been uploaded to AWS). This was tested extensively by enabling/disabling AWS storage and mounting/unmounting the KC image directory inside KPI (in kobo-docker).
- Grouping results by question or submission in the list view. These groupings return the index number and various labels (or timestamps) for each question or submission, as well as attachments.count and attachments.results for the attachments associated with the specified question or submission.
- Pagination for for record submissions now occurs based on the total number of submissions, with a default offset of 5. Pagination for questions now occurs based on the (nested) number of attachments per question, with a default offset of 5.
- Unit and Integration test coverage, using Mocking as needed.


pmusaraj:
Superseded by #1333

PR #1291, opened 2017-05-24
New Gallery and Gallery Modal UI

yahyazini:
This pull request contains all the back end changes to KPI in addition to the gallery's new UI


dorey:
* ❗️Commit 383bda7c20ea1458b781ac9231c682fc5d6d9f36 needs to be reverted because it deletes the form builder.

kpi/filters.py
```diff
@@ -219,3 +220,45 @@ def filter_queryset(self, request, queryset, view):
                 ).values_list('pk', flat=True)
             )
         return queryset.filter(pk__in=object_permission_ids)
+
+
+class AttachmentFilter(filters.BaseFilterBackend):
+    def filter_queryset(self, request, queryset, view):
+        type = request.query_params.get('type')
```
dorey: we should avoid this variable name since it overlaps with python's builtin `type()`. Python convention seems to be to add a trailing underscore; `type_`.

pmusaraj:
Superseded by #1333

Issue #1302, opened 2017-06-04
`sync_kobocat_xforms` creates duplicate `AssetVersion`s


jnm:
[When there's a hash mismatch](https://github.com/kobotoolbox/kpi/blob/6df89ce6104576e318961addd8b08f3c898c1a35/kpi/management/commands/sync_kobocat_xforms.py#L199), we correctly [update KPI with the form content from KC](https://github.com/kobotoolbox/kpi/blob/6df89ce6104576e318961addd8b08f3c898c1a35/kpi/management/commands/sync_kobocat_xforms.py#L200), but we never update the hash on KPI's side! [_sync_form_metadata()](https://github.com/kobotoolbox/kpi/blob/6df89ce6104576e318961addd8b08f3c898c1a35/kpi/management/commands/sync_kobocat_xforms.py#L282) does it properly.

PR #1315, opened 2017-06-07
Add management command to remove duplicate `AssetVersion`s

### LINKED ISSUE #1305: Management command to clean duplicate `AssetVersion`s
https://github.com/kobotoolbox/kpi/issues/1302 has created, in some cases, several thousand duplicate versions per `Asset`. These versions all have identical content with the exception of differing `$kuid`s. These should be purged, and only the oldest duplicate should be retained.

jnm:
Fixes #1305. Related to #1302. Built on top of PR #1307.

### DIFF
--- a/kpi/management/commands/remove_duplicate_assetversions.py
+++ b/kpi/management/commands/remove_duplicate_assetversions.py
@@ -1,0 +1,208 @@
+import json
+from collections import defaultdict
+from hashlib import md5
+from optparse import make_option
+
+from django.contrib.auth.models import User
+from django.core.management.base import BaseCommand
+from django.db import transaction
+from ...models import Asset, AssetVersion
+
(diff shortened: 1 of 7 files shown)


kpi/management/commands/remove_duplicate_assetversions.py
```diff
@@ -0,0 +1,208 @@
...
+                            pass
+                except KeyError:
+                    continue
+            serialized = json.dumps((
+                version.deployed_content,
+                version.name,
+                version._deployment_data,
+                version.version_content,
+                version.deployed
+            ))
```
dorey: `json.dumps([...], sort_keys=True)` will fix the issue where md5 hashes of equivalent dicts:

* `{"type": "text", "name": "abc"}`
* `{"name": "abc", "type": "text"}`

would not match. (also available in `formpack.utils.json_hash`)

PR #1327, opened 2017-06-12
Deployment fixes

jnm:
* Replace stray `deployed_versions.length` with `deployed_versions.count`;
* Fix Fabric `legacy.py`


jnm:
Self-merge to fix non-working `master`

PR #1333, opened 2017-06-12
Form gallery v2

yahyazini:


magicznyleszek:
@magicznyleszek note to self: see how much is usable from here; try to run it and see how it looks

magicznyleszek:
We will not follow this PR any further - there is a new one for Gallery now: #1959

PR #1338, opened 2017-06-16
Logging err fix

dorey:
Fixes issue with #1337 which causes logging.error to throw an error
Removes contenttype for AssetDeployment on initial migration


dorey:
failing test is because of (already-fixed) `gulpfile.js` error

PR #1411, opened 2017-07-31
Get timezone from env variable

starsep:
Get timezone from env variable

Related PRs:
https://github.com/kobotoolbox/kobo-docker/pull/129
https://github.com/kobotoolbox/kobocat/pull/374


dorey:
Should always be UTC to avoid timezone confusion

PR #1443, opened 2017-09-05
Add support for `edit` link; closes #1433 (again)

### LINKED ISSUE #1433: Add proxy endpoint to KoboCat data API
For the table and map views of a form (and potentially also gallery), I would like to set up a proxy from KPI to KC.  

The KC API call is `GET /api/v1/data/{pk}` where `pk` is the numeric form ID. The proxy should be able to convert the form's `uid` to the `pk` expected by KC. And it should also pass any other URL parameters. 

The KPI API should accept `GET` requests at the `/assets/{uid}/submissions` base. 
Some examples: 

* `/assets/{uid}/submissions` should proxy to `/api/v1/data/{pk}`
* `/assets/{uid}/submissions/1234` should proxy to `/api/v1/data/{pk}/1234` (where `1234` is the submission id)
* `/assets/{uid}/submissions?limit=200&start=0` should proxy to `/api/v1/data/{pk}?limit=200&start=0`
* `/assets/{uid}/submissions?query={"field":"value"}` should proxy to `/api/v1/data/{pk}?query={"field":"value"}` (for filtered results)
* `/assets/{uid}/submissions?sort={"field":1}` should proxy to `/api/v1/data/{pk}?sort={"field":1}` (for sorted results)
* `/assets/{uid}/submissions/1234/enketo` should proxy to `/api/v1/data/{pk}/1234/enketo` (for the enketo edit link)

To be able to fully migrate the data table views, we would need to add an endpoint to KPI for `DELETE /api/v1/data/{pk}/{dataid}`, i.e. deleting a specific submission instance. (I'm not sure whether this merits its own issue.)

jnm:


pmusaraj:
I merged this with my `data-map` branch, and it works well, but it fails the Travis build at `bower install` (which is odd...). Any ideas @jnm?

jnm:
It's punishing me (j/k) for naming it `redux`, which was bone-headed given the framework of the same name...

Issue #1447, opened 2017-09-20
When adding a question to the library, it adds only the question without the choices


emilykferguson:

pmusaraj:
Ready, please test in `kf.du`.

PR #1450, opened 2017-09-27
Fix bug of missing choices when adding to library

pmusaraj:
See #1447


jnm:
Tested on kf.du.kbtdev.org and QA'd by @emilykferguson

PR #1455, opened 2017-10-10
Added documentation on available superuser reports

tinok:


README.md
```diff
@@ -50,3 +50,9 @@ Construct a string using the [Whoosh query language](http://whoosh.readthedocs.i
 When the `q` parameter contains a search term without a specified field, e.g. `/tags/?q=health`, that term is matched against the search "document" (the `text` field).
 
 \* Implemented by Haystack as [a Whoosh TEXT field using the StemmingAnalyzer](https://github.com/django-haystack/django-haystack/blob/ad90028a22b4274b8df1f4698dd59ac0643f03d5/haystack/backends/whoosh_backend.py#L174). Unsuitable for exact matching.
+
+Admin reports
+-------------
+There are several types of data reports available to superusers. 
+* Full list of users including their details provided during signup, number of deployed projects (XForm count), number of submissions, date joined, and last login: `<kpi server domain>/forms/superuser_stats/user_report/`. File being created is a CSV, so don't download immediately to wait for server to be finished writing to the file (it will download even if incomplete).
```
jnm: `<kpi server domain>/forms/` only applies to people who had dkobo running at `<kpi server domain>`. I was thinking we could change to `<kpi base url>/superuser_stats/user_report/`, but that might be confusing. Any ideas?

README.md
```diff
@@ -50,3 +50,9 @@ Construct a string using the [Whoosh query language](http://whoosh.readthedocs.i
...
+* Monthly aggregate figures for number of forms, deployed projects, and submissions (from kobocat): `<kc server domain>/kobo/superuser_stats/`
```
jnm: `kobo` in the URL assumes there's a user named `kobo`. KC has that weird issue where usernames can collide with other paths (e.g. what happens if someone signs up as `admin`). I'd replace with `<kc server domain>/<superuser username>/superuser_stats`

tinok: It's actually very nice in that it will work regardless of which superuser you used to sign with...

README.md
tinok: Given that any modern kobo-docker installation doesn't include dkobo, anyone who would see this would not need the `forms/` part in the URL. Either way the URL you suggested works.

README.md
jnm: Right, but it depends on a user named `kobo` in the database. It'd be nice if `<kc server domain>/superuser_stats/` could be used, without having to worry that someone might register the username `superuser_stats` and cause problems.

PR #1461, opened 2017-10-23
Direct KPI-formpack exports

jnm:
Important: this PR adds a media file storage requirement to KPI. Deployment environments will need to be adjusted accordingly. See https://github.com/kobotoolbox/kobo-docker/issues/141.

Please note that `kpi/models/import_task.py` was renamed to `kpi/models/import_export_task.py`. The existing `ImportTask` and the new `ExportTask` both inherit from a common `ImportExportTask`.

Commit https://github.com/kobotoolbox/kpi/commit/f728f9b26804fc1cd1556222561fb62922cb50ff is intended for temporary use. It is no worse than the existing, synchronous formpack export system in KC, but it ignores the asynchronous capability provided by the backend code in this PR.


jnm:
Now running on https://kf.du.kbtdev.org/. See https://github.com/kobotoolbox/kobo-docker/pull/142 for Docker storage-related changes

KeynesYouDigIt:
@tinok  CI error in travis is similar to what I was getting locally, Ill check again today

dependencies/pip/dev_requirements.txt
```diff
@@ -5,13 +5,15 @@
...
 amqp==2.1.4
 anyjson==0.3.3
 begins==0.9
 billiard==3.5.0.2
+boto3==1.4.7
 boto==2.45.0
+botocore==1.7.33          # via boto3, s3transfer
 celery==4.0.2
 cffi==1.9.1               # via cryptography
 cryptography==1.7.2       # via paramiko, pyopenssl
```
KeynesYouDigIt: Can we make this 1.8 at some point? makes install go much easier

jnm: Is that a dependency of `boto3` added here by `pip-compile`? If so, what version of `boto3` do we need to use to pull in `botocore` 1.8?

jnm: @KeynesYouDigIt, I'll try `boto3` [1.5.8](https://github.com/boto/boto3/blob/1.5.8/setup.py#L17)

PR #1492, opened 2017-11-27
Update jquery-UI version to fix security vulnerability

pmusaraj:
Also updates jquery to most recent version of the 2.x branch. 

Removes `less` and `less-loader`, not used.


jnm:
Deployed this to `unu` and did a quick clickthrough :blush:

PR #1553, opened 2018-02-06
Dot notation in question names table fix

pmusaraj:
Modify react-table column accessor to accomodate dot notation in the row's object property name. Should go together with https://github.com/kobotoolbox/kobocat/issues/420


jnm:
@pmusaraj for a field named `dotty.dot.dot`, the column header appears (and is decoded properly), but no data shows up in the rows:
image

XLS export looks okay:
image

Issue #1552, opened 2018-02-06
Exclude white backgrounds from user avatar


pmusaraj:
Since text is white, light shades of white shouldn't be allowed.

jnm:
`cdcvte` is the tricky username

Issue #1466, opened 2017-11-02
Add single-submission option to Collect Data dropdown in Form Landing


pmusaraj:
Due to several support requests with `returnURL` parameter (which requires a single-submission URL in Enketo), we should re-add this to the Form Landing screen. Description should also mention that the single submission screen can be used with `returnURL` as a parameter.

Issue #620, opened 2016-04-08
Form title is not displayed in preview


tinok:
The form title displayed in kpi formbuilder preview is always shown as `Untitled`. See screenshot: ![image](https://cloud.githubusercontent.com/assets/673511/14396709/22de626a-fda7-11e5-8dd3-9859c5bd16fb.png)

dorey:
* preview from form landing page displays title correctly
* preview from within formbuilder currently displays default value: "Snapshot XML" (!!)

PR #1555, opened 2018-02-07
Misc. UI fixes

pmusaraj:
Fixes the following small issues: 

#620 
#1466 
#1547 
#1552


jnm:
Testing on tri

- [ ] #620
- [ ] #1466
- [ ] #1547
- [ ] #1552

Issue #1238, opened 2017-04-27
allow users to access reports when granted permission


dorey:
Connected to #936 , but also different.

* Currently, reports can only be accessed by the project owner.
* When a user has been granted permission to view data of a project, they should be able to view the report
* Should be implemented as a DRF permission filter. E.g. in a `kobo/apps/reports/filters.py`

jnm:
We'll probably just modify [`get_queryset()`](https://github.com/kobotoolbox/kpi/blob/3e41c9a167aeea3b51fbccb661dd3fcc5edf8bf9/kobo/apps/reports/views.py#L19) in the interest of time. Reference https://github.com/kobotoolbox/kpi/blob/c84210b9b80f26429c143b1429b812e85e2d9983/kpi/models/object_permission.py#L43

Discussion: https://www.flowdock.com/app/kobotoolbox/kobo/threads/CBlaZsNMBs0ICDhRQQSwUu18L_0?filter=search&query=reports

jnm:
Fixed by https://github.com/kobotoolbox/kpi/pull/1557

PR #1557, opened 2018-02-09
Give users access to shared projects reports when granted permission - fix 

noliveleger:
Fixed issue discussed in #1238.


jnm:
@noliveleger, FYI: https://help.github.com/articles/closing-issues-using-keywords/

PR #1558, opened 2018-02-09
darken user avatar backgrounds, to avoid having white on white

### LINKED ISSUE #1552: Exclude white backgrounds from user avatar

pmusaraj:
fixes #1552

see screenshot below for the difference, especially for user `cdcvte`

cdcvte

### DIFF
--- a/jsapp/js/utils.es6
+++ b/jsapp/js/utils.es6
@@ -169,7 +169,7 @@ export function isLibrary(router) {
 
 export function stringToColor(str, prc) {
   // Higher prc = lighter color, lower = darker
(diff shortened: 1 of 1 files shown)


jsapp/js/utils.es6
```diff
@@ -169,7 +169,7 @@ export function isLibrary(router) {
 
 export function stringToColor(str, prc) {
   // Higher prc = lighter color, lower = darker
-  var prc = typeof prc === 'number' ? prc : 0.1;
+  var prc = typeof prc === 'number' ? prc : -15;
```
jnm: I love this

PR #1576, opened 2018-02-15
Try to resolve item_path before removing prefix - fixes #1530

### LINKED ISSUE #1530: Make _resolve_url_to_asset_or_collection work with settings.KPI_PREFIX used in urls.py
[_resolve_url_to_asset_or_collection](https://github.com/kobotoolbox/kpi/blob/master/kpi/models/import_export_task.py#L32) tries to resolve the URL of an asset or collection to determine the type and instance. The code currently checks if `settings.KPI_PREFIX` is used and removes it before trying to match.
If the application is mounted under a path in uwsgi, this probably works fine.
However, we are doing it a bit differently, we use a urlpattern in our `ROOT_URLCONF` like this:

`url(r'kpi/', include('kobo.urls'))`

This allows us to register additional urlpatterns without having to modify upstream, among other things. But it means that all Django urlpatterns contain the prefix and the code in `_resolve_url_to_asset_or_collection` doesn't resolve the URL after the prefix has been removed.

jeverling:
This fixes #1530

### DIFF
--- a/kpi/models/import_export_task.py
+++ b/kpi/models/import_export_task.py
@@ -1,41 +1,46 @@
 # FIXME: clean up these imports!
 import base64
-import pytz
-from io import BytesIO
 import datetime
-import dateutil.parser
-import re
 import logging
 import posixpath
+import re
 import tempfile
-from os.path import splitext
 from collections import defaultdict
-from django.db import models, transaction
+from io import BytesIO
+from os.path import splitext
+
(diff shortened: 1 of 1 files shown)


jnm:
Merged manually with some changes: https://github.com/kobotoolbox/kpi/commit/041c986bfc64478cfb7ca493e337642bbeaccee7

PR #1578, opened 2018-02-20
Fix #1577: fix issue where sharing form in UI fails to load if asset …

### LINKED ISSUE #1577: Sharing modal fails to load if asset has `from_kc_only` permission
(And possibly any additional permission not in the KPI definition.)

Frontend code should simply ignore these additional permissions and continue to display the sharing form with only the permissions it recognizes.

pmusaraj:
…includes a user with "from_kc_only" permission


jnm:
I opened a separate PR to fix the Travis failure: https://github.com/kobotoolbox/kpi/pull/1579

PR #1616, opened 2018-03-07
Latest-version-only exports

### LINKED ISSUE #1609: Add latest-version-only checkbox to export options
People have written integrations that don't look at exported column headers but instead depend on the specific positions of fields in exported data. Now that we include all columns from all deployed versions, these integrations are broken. Previously, only the latest deployed version was used for exports, so we should add an option that allows people to retain that behavior if they desire.

### LINKED ISSUE #1611: labels/values incorrect in export filenames
When downloading a report that just contains label the resulting file name is [formname] - values - [timestamp].xls. Instead it should be [formname] - labels - [timestamp].xls

See https://www.flowdock.com/app/kobotoolbox/ux-ui/threads/iWAVAag_yzsnkxmsBA-LYtZB1bW.

Probably needs a change in https://github.com/kobotoolbox/kpi/blob/828cb2be342d17c85bb2498e897f7e3917fac077/kpi/models/import_export_task.py#L343-L347

jnm:
Closes #1609
Fixes #1611

### DIFF
--- a/jsapp/js/components/formEditors.es6
+++ b/jsapp/js/components/formEditors.es6
@@ -258,6 +258,10 @@ export class ProjectDownloads extends React.Component {
       lang: '_default',
       hierInLabels: false,
       groupSep: '/',
+      // If there's only one version, the resulting file will be the same
+      // regardless of whether this is true or false, but we'll use this to
+      // report if the export was "multi-versioned" later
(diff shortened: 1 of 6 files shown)


jnm:
Available for testing on https://kf.unu.kbtdev.org/

PR #1630, opened 2018-03-15
Include fix for kobotoolbox/formpack#159, the

jnm:
column-ordering bug


noliveleger:
@jnm, because we fetch last commit of FormPack, we should remove this temporary test `test_temporary_test_of_formpack_issue_163_patch` from KPI. 
https://github.com/kobotoolbox/kpi/blob/master/kpi/tests/test_mock_data_exports.py#L471

It's part of FormPack now.
https://github.com/kobotoolbox/formpack/blob/master/tests/test_formpack_internals.py#L198

Issue #1569, opened 2018-02-14
Allow filtering GPS points on map based on selected response


tinok:
A user should be able to click on one or more responses within a question and only see GPS points that correspond to those. 

This functionality is available in the old KC map. 

![image](https://user-images.githubusercontent.com/673511/36207477-111d2214-1164-11e8-9c7d-6cac490a85c6.png)

PR #1700, opened 2018-04-04
Map colors

### LINKED ISSUE #1660: New map view disaggregation uses red for many responses
When disaggregating by a response that has a lot of options, many appear to be displayed in red, making it hard to distinguish them. Rather, they should use the full gamut of available colors as much as possible. 

image

pmusaraj:
Fixes #1660, #1652 and #1569.


jsapp/js/components/map.es6
```diff
@@ -150,15 +156,21 @@ export class FormMap extends React.Component {
...
+      var mM = [], lbl = undefined;
+      let choices = this.props.asset.content.choices,
+          survey = this.props.asset.content.survey;
+
+      let q = survey.find(z => z.name === viewby || z.$autoname === viewby);
 
       Object.keys(mapMarkers).map(function(m, i) {
-        var lbl = choices.find(o => o.name === m || o.$autoname == m);
+        if (q && q.type == 'select_one') {
+          lbl = choices.find(o => o.list_name === q.select_from_list_name && (o.name === m || o.$autoname === m));
```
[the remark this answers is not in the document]
smutnyleszek: It also applies to all other single letters and abbreviations in this file - but maybe it should be fixed in another issue?

jsapp/js/components/map.es6
```diff
@@ -150,15 +156,21 @@ export class FormMap extends React.Component {
...
+        }
         mM.push({
           count: mapMarkers[m].count,
           id: mapMarkers[m].id,
-          labels: lbl ? lbl.label : undefined
+          labels: lbl ? lbl.label : undefined,
+          value: m != "undefined" ? m : undefined
```
smutnyleszek: Does it mean that `m` can actually be a string "undefined"? Wouldn't that cause errors in few previous lines? Maybe we should return function on top if `m` has bad value?

jsapp/js/components/map.es6
```diff
@@ -180,20 +192,46 @@ export class FormMap extends React.Component {
             iconSize: [20, 20],
           });
         }
-        prepPoints.push(L.marker(item._geolocation, {icon: icon, sId: item._id}));
+        prepPoints.push(
+          L.marker(item._geolocation, {
+            icon: icon, 
+            sId: item._id, 
+            typeId: viewby && mapMarkers ? mapMarkers[itemId].id : null
```
smutnyleszek: IIUC this mirrors the if statement from line 175(187), so if the mentioned line will change, this could potentially be a bug

jsapp/js/components/map.es6
```diff
@@ -322,7 +361,35 @@ export class FormMap extends React.Component {
...
+      else
+        layer._icon.classList.remove("unselected");
+    });
+  }
+
+  resetFilterByMarker() {
+    let markers = this.state.markers;
+    this.setState({filteredByMarker: false});
+    markers.eachLayer( function(layer) {
+      layer._icon.classList.remove("unselected");
```
smutnyleszek: "unselected" is used few times, I think it would be safer as a declared variable?

jsapp/scss/components/_kobo.navigation.scss
```diff
@@ -136,6 +136,7 @@ $headerAccountTextColor: #dbedf7;
 .popover-menu--account-menu {
   .popover-menu__content {
     min-width: 240px;
+    z-index: 1001;
```
smutnyleszek: Would it be easy to create zindexes.scss file for keeping all important `z-indexes` as variables there? We wouldn't need those magical numbers then :)

jsapp/js/components/map.es6
```diff
@@ -180,20 +192,46 @@ export class FormMap extends React.Component {
...
         var markers = L.featureGroup(prepPoints);
       } else {
-        var markers = L.markerClusterGroup({maxClusterRadius: this.calculateClusterRadius, disableClusteringAtZoom: 16});
+        var markers = L.markerClusterGroup({
+          maxClusterRadius: this.calculateClusterRadius, 
+          disableClusteringAtZoom: 16,
+          iconCreateFunction: function(cluster) {
+            var childCount = cluster.getChildCount();
+
+            var c = ' marker-cluster-';
```
smutnyleszek: Maybe it would be more obvious why space is here if you change it to `'marker-cluster marker-cluster-'`

jsapp/scss/components/_kobo.map.scss
```diff
@@ -166,6 +167,49 @@ body .leaflet-control-layers,
 
   .map-marker-item {
     padding: 4px 6px;
+    position: relative;
+
+    &:after {
```
smutnyleszek: minor cssnazi: `::after` (+ other occurences in this file)

jsapp/scss/components/_kobo.map.scss
```diff
@@ -166,6 +167,49 @@ body .leaflet-control-layers,
...
+      &:after {
+        content: "\f111";
+      }
+    }
+
+    &.unselected {
+      opacity: 0.6;
+
+      &:after {
+        content: "\f10c";
```
smutnyleszek: Maybe add a small comment here and in previous one about which FontAwesome shape you're using here?

jsapp/scss/components/_kobo.map.scss
```diff
@@ -230,95 +275,104 @@ body .leaflet-control-layers,
...
 
-  &-3, &-3:after {
-    background: rgba(252, 74, 124, 0.4);
-  }
+$color-collection: rgba(255, 50, 0, 0.4), rgba(125, 245, 46, 0.4), rgba(86, 0, 178, 0.4), rgba(0, 236, 118, 0.4), rgba(202, 0, 188, 0.4), rgba(104, 155, 0, 0.4), rgba(161, 107, 255, 0.4), rgba(191, 184, 0, 0.4), rgba(0, 66, 187, 0.4), rgba(197, 231, 113, 0.4), rgba(239, 0, 161, 0.4), rgba(2, 231, 203, 0.4), rgba(239, 0, 69, 0.4), rgba(2, 153, 148, 0.4), rgba(43, 130, 255, 0.4), rgba(183, 142, 0, 0.4), rgba(60, 46, 126, 0.4), rgba(35, 106, 0, 0.4), rgba(175, 158, 255, 0.4), rgba(198, 116, 0, 0.4), rgba(0, 140, 218, 0.4), rgba(255, 109, 60, 0.4), rgba(1, 76, 140, 0.4), rgba(184, 69, 0, 0.4), rgba(165, 228, 255, 0.4), rgba(112, 31, 9, 0.4), rgba(163, 231, 229, 0.4), rgba(133, 0, 75, 0.4), rgba(167, 234, 188, 0.4), rgba(96, 33, 97, 0.4), rgba(21, 70, 0, 0.4), rgba(255, 110, 148, 0.4), rgba(0, 87, 84, 0.4), rgba(255, 157, 210, 0.4), rgba(51, 64, 37, 0.4), rgba(255, 153, 153, 0.4), rgba(12, 64, 93, 0.4), rgba(255, 194, 186, 0.4), rgba(96, 42, 58, 0.4), rgba(134, 0, 40, 0.4);
```
smutnyleszek: Are those some generated colors - if yes, maybe add source url?

jsapp/scss/components/_kobo.map.scss
```diff
@@ -230,95 +275,104 @@ body .leaflet-control-layers,
...
+$colorLength: length($color-collection);
 
-  &-6, &-6:after {
-    background: rgba(33, 231, 184, 0.4);
+@for $i from 1 through $colorLength {
+  .map-marker.#{$class-slug}-#{$i},
+  .map-marker.#{$class-slug}-#{$i + $colorLength},
+  .map-marker.#{$class-slug}-#{$i + $colorLength * 2},
+  .map-marker.#{$class-slug}-#{$i + $colorLength * 3},
+  .map-marker.#{$class-slug}-#{$i + $colorLength * 4} {
```
smutnyleszek: Would it be possible to do this with `nth-child`? We would avoid a problem with this list not having `* 5`, `* 6`, etc.

jsapp/scss/components/_kobo.map.scss
```diff
@@ -230,95 +275,104 @@ body .leaflet-control-layers,
...
-  }
+.marker-cluster-small {
+  background-color: rgba(253, 187, 132, 0.4);
+}
 
-  &-13, &-13:after {
-    background: rgba(122, 187, 200, 0.4);
-  }
+.marker-cluster-small div {
+  background-color: rgba(253, 187, 132, 0.5);
```
smutnyleszek: Wouldn't `background-color: inherit` be good enough? :)

jsapp/scss/components/_kobo.map.scss
pmusaraj: I think autoprefixer covers this.

jsapp/scss/components/_kobo.navigation.scss
pmusaraj: Sure, but let's note it for future work. It's beyond this PR's scope.

jsapp/js/components/map.es6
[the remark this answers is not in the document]
pmusaraj: Yes, this is a little too compact. I'll rework it.

jsapp/js/components/map.es6
pmusaraj: No, it wouldn't cause errors, but it would print "undefined" to users when rendering the list. But the "undefined" string can still be a property in the previous lines (this comes from the API call).

jsapp/scss/components/_kobo.map.scss
pmusaraj: I added a link to the generator. This will soon be replaced with a configurable list of colors though, similar to reports.

jsapp/scss/components/_kobo.map.scss
pmusaraj: I'm not sure how `nth-child` could do this. We can discuss in near future.

jsapp/scss/components/_kobo.map.scss
pmusaraj: It wouldn't have the subtle opacity difference.

jsapp/scss/components/_kobo.navigation.scss
smutnyleszek: Noted it down here: https://github.com/orgs/kobotoolbox/projects/6#card-8762792

jsapp/scss/components/_kobo.map.scss
smutnyleszek: It doesn't: https://github.com/postcss/autoprefixer/issues/114 :(

pmusaraj: Right. I wasn't sure what the difference between `:after` and `::after` was. I've looked it up, and we really don't need to worry about this.

smutnyleszek:
@penar Can I get somewhere a file with ready form with translations that I could just import? So I can finish testing this? :)

pmusaraj:
This is an example of a translated form with a geopoint: https://drive.google.com/open?id=10MxaphtTEyKPdVwIFqNHiPuDaLhbVeqI

smutnyleszek:
@pmusaraj I've tested the map colors issue from #1660 and now it's much better + other things seem to work well too

PR #1710, opened 2018-04-05
remove duplicate eslint rules

smutnyleszek:
@pmusaraj please check if the removed properties aren't the preferred ones


pmusaraj:
Hmm, the Travis build fails, because of `react/no-did-mount-set-state`.

pmusaraj:
@smutnyleszek I disabled the `camelCase` rule, because it stops the build from completing, and we have lots of non-camelcased variables. I did clean up some other items raised by ESLint, and it now passes tests. 

We can revisit camelcasing in the near future, but it's not a huge priority in my book.

PR #1716, opened 2018-04-09
form editor header icon back click navigation

### LINKED ISSUE #1688: Cannot find a way to get back to the form list when inside the form editor
As raised by @jnm.

smutnyleszek:
closes #1688 

@pmusaraj Q1: I noticed that you can go back by clicking "x" on the right side - maybe instead of adding click event to the icon on the left, we want to improve the visibility of "x"-close button?

Q2: I can add an left-pointing arrow to appear on hover to emphasise the fact it's "go back" button. What do you think?


smutnyleszek:
@pmusaraj I've changed it to link to the list (deduplicated it with shared function), plus added the tooltip (had to create `left-tooltip` and change `right-tooltip` to make it clearer)

PR #1719, opened 2018-04-09
Add editorconfig file

smutnyleszek:
I think I was able to identify all different 4-spaces, but I might have missed something. We can improve it later :)

Read more about it: http://editorconfig.org/


pmusaraj:
I'm ok with this, but I do see one issue: as soon as I make a change to a file (in my test, an .es6 file), the `diff` for the commit becomes quite long, because of previous inconsistencies. 

I suggest this PR should include all formatting changes based on the EditorConfig, to make sure future PRs don't mix formatting changes with functional changes.

@jnm thoughts?

smutnyleszek:
I'm ok for going through all files and applying indentation rules. And as this will touch all the files - it means we will have an option to change the current "unofficial" indentation rules. What do you think about changing everything to 2 spaces? @pmusaraj @jnm

pmusaraj:
No, @smutnyleszek let's not change the indentation rules for Python files. In fact, is there a way to limit the scope of the `editorconfig` for now to frontend JS/CSS files only?

smutnyleszek:
@pmusaraj I changed the config to only apply to `*.{es6,js,coffee,scss,css,html}` files :) There seems to be a way to autofix all the files (https://stackoverflow.com/questions/39343031/editorconfig-how-to-autofix-all-files-in-a-project) and then going through all of the changes before commiting

smutnyleszek:
@pmusaraj I commited all files autofix fixed (I reviewed all changes to make sure it doens't brake anything). 

Unfortunately it didn't caught up the multiline html tag definitions:

```
<bem.FormBuilderHeader__button m={['save', {
                    savepending: this.state.asset_updated === update_states.PENDING_UPDATE,
                    savecomplete: this.state.asset_updated === update_states.UP_TO_DATE,
                    savefailed: this.state.asset_updated === update_states.SAVE_FAILED,
                    saveneeded: this.state.asset_updated === update_states.UNSAVED_CHANGES,
                  }]} onClick={this.saveForm} className="disabled"
                  disabled={!this.state.surveyAppRendered || !!this.state.surveyLoadError}>
```

I see sometimes we try to follow something like this:

```
<sometag one="1"
                  two="2">
```

and sometimes:

```
<sometag one="1"
  two="2">
```

- do we want to agree on something? (I personally can tell that the first one is worse, because changeing element tag name will require to change multiple lines in code)

Issue #1691, opened 2018-03-28
Allow imports to specify a url instead of base64Encoded


jnm:
i.e. provide an API to import XLSForms via URL, leveraging logic used by the [`import_from_url` management command](https://github.com/kobotoolbox/kpi/blob/7a306680a7c7f34490d5eb95e6d0c1646fc02a3b/kpi/management/commands/import_from_url.py).

pmusaraj:
This looks and works well. One possible improvement:  
API should return an error message when submitted URL is not an XLSForm. It currently returns an error if the URL is a 404, but if I input a URL like `https://www.google.com` the frontend doesn't receive an error.

jnm:
@jnm don't forget to write unit tests...

jnm:
@penar, oh boy, were you ever right. Writing tests for this has revealed a real mess. I also think the backend silently fails to process XLS files. Argh.

PR #1727, opened 2018-04-18
Let users import XLSForms via URL

pmusaraj:
This new feature also includes the following changes: 

- when clicking "New" from the sidebar, users will see the "Create New Project" modal directly (the dropdown with buttons Project and Upload has been removed), and after they create the project, they will have three choices: **Design in Form Builder**, **Upload an XLSForm** or **Import an XLSForm via URL**. 
- when replacing a form via XLS, they will see two choices: **Upload an XLSForm** and **Import an XLSForm via URL**.

@mroswell once this is merged, tests may need to be adjusted.

The Library workflow for creating new questions/blocks/collections remains unchanged.


pmusaraj:
@tinok fixed all the issues mentioned in your previous comment. Please see the help article here: http://help.kobotoolbox.org/creating-forms/importing-an-xlsform-via-url 

The message regarding uploads is also more prominent, see https://monosnap.com/file/sJ25widK7EhBzo4msJFfZnVnkF8I3U and I added (step 1 of 2) to the modal window title. 

I agree that we should mention this in Intercom.

jnm:
@pmusaraj, I rewrote the backend component and deployed to https://kf.du.kbtdev.org/ for testing. I think the error handling [that you mentioned](https://github.com/kobotoolbox/kpi/issues/1691#issuecomment-380921147) is better now.

pmusaraj:
- [ ] when the URL has a completely invalid hostname, the UI outputs `undefined undefined` for the errors.

PR #1731, opened 2018-04-20
Fix #1730: do not refresh asset list when deleting an asset

### LINKED ISSUE #1730: Avoid reloading all assets after deleting an asset from list
This is a small performance issue that has recently bugged me a lot: a full refresh of the asset list is triggered when deleting an asset from the list. This is sometimes slow, especially on accounts with 100+ assets. 

It's not necessary to refresh the list of assets, especially since we already have a method to filter out deleted assets from the search store. We should fix `onDeleteAssetCompleted` so that it clears both the non-categorized `defaultQueryResultsList` and the `defaultQueryCategorizedResultsLists`.

pmusaraj:
Instead, update the list of categorized results in the search store. Fixes #1730

### DIFF
--- a/jsapp/js/searches.es6
+++ b/jsapp/js/searches.es6
@@ -80,7 +80,6 @@ function SearchContext(opts={}) {
     },
     onDeleteAssetCompleted (asset) {
       var filterOutDeletedAsset = ({listName}) => {
-        // TODO: look into why sometimes this.state[listName] is not defined
         if (this.state[listName] != undefined) {
           let uid = asset.uid;
           let listLength = this.state[listName].length;
@@ -94,7 +93,22 @@ function SearchContext(opts={}) {
           }
         }
       };
+      var filterOutDeletedAssetFromCategorizedList = () => {
+        let list = this.state.defaultQueryCategorizedResultsLists;
+        if (list != undefined) {
+          var l = {};
+          for (var category in list) {
+            l[category] = list[category].filter(function(result){
+              return result.uid !== asset.uid;
+            });
+          }
+          let o = {};
+          o.defaultQueryCategorizedResultsLists = l;
+          this.update(o);
+        }
+      };
       filterOutDeletedAsset({listName: 'defaultQueryResultsList'});
+      filterOutDeletedAssetFromCategorizedList();
       if (this.state.searchResultsList && this.state.searchResultsList.length > 0) {
         filterOutDeletedAsset({listName: 'searchResultsList'});
       }
--- a/jsapp/js/searches.es6
+++ b/jsapp/js/searches.es6
@@ -95,7 +95,7 @@ function SearchContext(opts={}) {
       };
       var filterOutDeletedAssetFromCategorizedList = () => {
         let list = this.state.defaultQueryCategorizedResultsLists;
-        if (list != undefined) {
+        if (list) {
           var l = {};
           for (var category in list) {
             l[category] = list[category].filter(function(result){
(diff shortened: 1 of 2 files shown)


smutnyleszek:
@pmusaraj would you like me to review this?

jsapp/js/searches.es6
```diff
@@ -94,7 +93,22 @@ function SearchContext(opts={}) {
           }
         }
       };
+      var filterOutDeletedAssetFromCategorizedList = () => {
+        let list = this.state.defaultQueryCategorizedResultsLists;
+        if (list != undefined) {
```
smutnyleszek: Why not `!==`?

jsapp/js/searches.es6
```diff
@@ -94,7 +93,22 @@ function SearchContext(opts={}) {
           }
         }
       };
+      var filterOutDeletedAssetFromCategorizedList = () => {
+        let list = this.state.defaultQueryCategorizedResultsLists;
```
smutnyleszek: Do we have any rules on `const` vs `let`? There is an eslint rule about that: https://eslint.org/docs/rules/prefer-const

jsapp/js/searches.es6
```diff
@@ -80,7 +80,6 @@ function SearchContext(opts={}) {
     },
     onDeleteAssetCompleted (asset) {
       var filterOutDeletedAsset = ({listName}) => {
-        // TODO: look into why sometimes this.state[listName] is not defined
         if (this.state[listName] != undefined) {
```
smutnyleszek: Do you think removing this `// TODO` should leave an explanation comment?

jsapp/js/searches.es6
```diff
@@ -94,7 +93,22 @@ function SearchContext(opts={}) {
           }
         }
       };
+      var filterOutDeletedAssetFromCategorizedList = () => {
+        let list = this.state.defaultQueryCategorizedResultsLists;
+        if (list != undefined) {
+          var l = {};
+          for (var category in list) {
```
smutnyleszek: Why `var` not `let` or `const`?

jsapp/js/searches.es6
```diff
@@ -94,7 +93,22 @@ function SearchContext(opts={}) {
...
+            l[category] = list[category].filter(function(result){
+              return result.uid !== asset.uid;
+            });
+          }
+          let o = {};
+          o.defaultQueryCategorizedResultsLists = l;
```
smutnyleszek: Nitpicking: what is `l` and `o`? I would be happier with longer names ;-)

jsapp/js/searches.es6
pmusaraj: We currently don't. I'd rather not change this now, let's note it for a future formatting/code cleanup update.

jsapp/js/searches.es6
pmusaraj: No need. FYI, the search should get an overhaul soon-ish.

jsapp/js/searches.es6
pmusaraj: `filterOutDeletedAssetFromCategorizedList` closely emulates `filterOutDeletedAsset`, and I kept the `l` and `o` variables. They're basically temporary holders of the list data, post-filtering.

jsapp/js/searches.es6
smutnyleszek: Ok

jsapp/js/searches.es6
smutnyleszek: ok :)

smutnyleszek:
@pmusaraj :ship: :it:

PR #1743, opened 2018-04-24
UI updates for registration/login screens

### LINKED ISSUE #1739: Missing link on expired account activation
- I created a selenium_test2 account 20 days ago.
- I discovered today I could not log in.
- I found the email in my account that offered a link to activate my new account.
- The text on that link landing page reads as follows: 
```
Sorry, we couldn't activate your account. The link we sent you has expired.
Click here to create a new account.
```
But the "Click here to create a new account" text doesn't link to anything. 
- The href appears in the html above the text, but it's empty. 
```<a href="/accounts/register/"></a>```

screen shot 2018-04-23 at 5 44 51 pm

pmusaraj:
This PR has some UI changes to our login/registration screens. It also fixes #1739 (thanks @mroswell for raising that ticket). 

Screenshots below of the subtle changes in the registration/login screens. Left is before, right is after. 

login-before-after

registration-before-after

(Please ignore differences in the text on the right column of the screen below. That is content set differently for each Django instance.)
account-created-before-after

activation-failed-before-after

forgotpassword-before-after


jnm:
I think these changes are definitely an improvement overall, and I'd be happy merging this, but I'll raise a few minor things first:

* The `/accounts/register/` input fields all being the same width (old style) looks cleaner to me than the new widths (but I'm guessing you intentionally changed it and have a good reason to prefer the new way?);
* The right-aligned "Reset Password" button looks lonely without the "Cancel" link. Could it be expanded to match the width of the email input field, the way "Login" matches the username and password fields?

pmusaraj:
Forgot to metion that the "Reset Password" button width now matches the field above it (like Login).

PR #1746, opened 2018-04-26
Issue 1683 frontend part

smutnyleszek:
frontend part of the Copy Permissions

some polishing to do:

- [x] display current asset title in the modal
- [x] filter out current asset in copy-from select
- [x] hide whole thing if user don’t have multiple forms


pmusaraj:
At first look this is great, a few minor notes: 

- [ ] please add a loading indicator after submitting the form. You could do this similar to the redeploying form, where the submit button is disabled while the action is taking place. 
- [ ] after updating sharing preferences for a form, if i open the sharing preferences for a second form, the "copy permissions" section is expanded. it should be collapsed. 
- [ ] thanks for displaying the title of the project in the sharing modal, but it to be a bit more prominent, in terms of style. I'll post a design here shortly.

PR #1763, opened 2018-05-04
escape & characters in form title, fixes #1761

### LINKED ISSUE #1761: Inserting an & into the form_title breaks Enketo
In general using **&** in the `form_title` is not a problem. However, if a **&** character is inserted into the title in the formbuilder, it will not be escaped properly and the form will not load correctly in Enketo. The error messages are something like
`{code: 500, message: "xmlParseEntityRef: no name"}`
or
`{code: 500, message: "EntityRef: expecting ';'"}`
depending on whether there is a character following the **&**.

It doesn't matter whether the form_title is changed using the form edit view in the formbuilder or the *Project title* input:

image

When a XLSForm is uploaded that contains a **&** character in the `form_title` cell it works just fine.

pmusaraj:


jnm:
Thanks for working on this @pmusaraj. I'm going to close the PR per the discussion in https://github.com/kobotoolbox/kpi/issues/1761

PR #1769, opened 2018-05-09
stylelintrc config

magicznyleszek:
I based the file on https://github.com/stylelint/stylelint-config-recommended and https://github.com/stylelint/stylelint-config-standard - just removed few and changed most to warnings

@pmusaraj should I apply autofix on all our files in this PR?


smutnyleszek:
Dropping this in favor of #1778

magicznyleszek:
@jnm we decided not to use stylelint in #1778 but then reverted this decision in #1790 - so I want to bring this back, as writing neatier css code is always a plus. I touched a few files to remove errored lines, but am not doing any autofix all, so it will be easier to merge it. Plus now stylelint will be run with test/build scripts, so no bad stuff allowed in production code anymore.

magicznyleszek:
@Eric-Dilcher I assigned this to you too, as you probably have something to say in this matter :)

jsapp/scss/components/_kobo.collection-asset-list.scss
```diff
@@ -143,8 +143,6 @@
     position: relative;
     overflow: visible;
 
-    &--visible {}
```
jnm: I would guess that someone left these empty blocks as a hint for future developers. You're the one working on it, though, so if you want to get rid of them, that's fine with me :)

PR #1773, opened 2018-05-12
Include formpack fix for large exports

jnm:
See kobotoolbox/formpack#171. Also includes a trivial update to pyxform, which
resolves a lxml version conflict


jnm:
This accidentally removes `jsapp/fonts/.gitignore`. Not the best day today.

PR #1778, opened 2018-05-14
add prettier as dependency and precommit script

magicznyleszek:
@pmusaraj I already have prettier installed as a plugin in my editor. The plugin has an option to apply prettier on saving file, and it worked great for me since I started using it. Now I see the prettier folk recommend adding a precommit script that prettify staged files - do you think we would like to try that, or better to have some control on the commits?

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,1 @@
+singleQuote: true


smutnyleszek:
@pmusaraj I've found out you can have `.prettierignore` so it will not fix automatically some files (like `package.json` for example. I've added what I think should be there, but feel free to improve it :)

Issue #1751, opened 2018-04-28
Table UI improvements


pmusaraj:
- [x] add Enketo edit link (icon) to each table row. upon clicking the link, make a call to the EE API to receive said link, and open it in a new window
- [x] post Enketo editing, refresh current table view
- [x] add pagination to submission modal window (record x of y),~@jnm some backend help may be needed here~ no backend needed, actually
- [x] make sure prev/next pagination goes through all records, (now it stops at end of current table view pagination)
- [x] (somewhat related to table) identify unlabelled groups as such in submission modal (currently says "Unlabelled", should say "Unlabelled group")

PR #1779, opened 2018-05-15
Table UI fixes and improvements

### LINKED ISSUE #1663: Form owner can hide columns, set primary column in table view
1. A (project owner) user should be able to hide any number of columns 
1. User should be able to choose one column to be displayed in the first column space. 
1. The user should be able to save these settings to the server to become the default view of the table. Any user with whom the project is shared would see the view with the same settings. But other users can change the view though the changes can not be stored except by the owner. 
1. [Nice to have] Ideally, the backend request should exclude hidden columns.

Originally
> Owner of the project will assign the default field, but users can change and a session will be store in the browser, and the user session will override the owner's choice

pmusaraj:
Fixes #1663 and #1751.


pmusaraj:
Safari bug resolved.

pmusaraj:
Replaced by #1788.

PR #1790, opened 2018-05-23
Revert prettier code-formatting

pmusaraj:
Reverts https://github.com/kobotoolbox/kpi/pull/1778. 

Prettier's pre-commit hook makes formatting changes to whole files (as soon as anything on the file is edited). That means that we would quickly lose the line-by-line `git blame` history (or make it extremely complicated to access). 

After discussions with John, we decided to revert Prettier. We may revisit this in the future, but we would certainly look for something that's less drastic.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,5 +1,0 @@
-locale
-node_modules
-static
-package.json
-package-lock.json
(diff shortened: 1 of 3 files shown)


jnm:
It'd be great to have a tool like Prettier that operates only on changed lines, but alas...

Issue #1794, opened 2018-05-28
API Endpoint: create a new template from a project


magicznyleszek:
It will be called by the frontend from the list of projects.

It should take the asset id as a parameter, and create a new asset of type template with the contents of the initial asset. 

Question: should permissions be carried over to the template as well? (My take would be that they shouldn't, especially considering that the template can be shared and made public in the near future.)

Issue #1793, opened 2018-05-28
API Endpoint: create/reset project from a template


magicznyleszek:
While creating a new project, users can choose to create it from a template. Endpoint should copy all template properties, but should not override project name, project settings (description, sector, country, etc.) and project permissions. 

Parameters: template assetid, project assetid

Note: in the frontend, there will always be a current project asset id, but it would be nice to have the API be able to create a new project when there is no project assetid.

PR #1806, opened 2018-06-01
Add new type of asset: `template`

noliveleger:
Covers #1793 , #1794 

and backend part of issues #1634, #1633


kpi/deployment_backends/mixin.py
```diff
@@ -18,11 +18,15 @@ def connect_deployment(self, **kwargs):
...
+        if self.can_be_deployed:
+            if not self.has_deployment:
+                self.connect_deployment(backend=backend, active=active)
+            else:
+                self.deployment.redeploy(active=active)
+            self._mark_latest_version_as_deployed()
         else:
-            self.deployment.redeploy(active=active)
-        self._mark_latest_version_as_deployed()
+            raise BadAssetTypeException("A form template can't be deployed")
```
jnm: This message should probably match whatever you choose for https://github.com/kobotoolbox/kpi/pull/1806/files#diff-43986855ae1015d5975b107cf1ee9869R691 (which I have suggested changing since only `survey`s can be deployed)

kpi/views.py
```diff
@@ -848,37 +1042,44 @@ def deployment(self, request, uid):
...
-                data=request.data,
-                context={'asset': asset}
-            )
-            serializer.is_valid(raise_exception=True)
-            serializer.save()
-            # TODO: Understand why this 404s when `serializer.data` is not
-            # coerced to a dict
-            return Response(dict(serializer.data))
+            if not asset.can_be_deployed:
+                raise BadAssetTypeException(u"A form template can't be deployed")
```
jnm: (sorry, same thing over and over) This message should probably match https://github.com/kobotoolbox/kpi/pull/1806/files#diff-43986855ae1015d5975b107cf1ee9869R691

kpi/views.py
```diff
@@ -848,37 +1042,44 @@ def deployment(self, request, uid):
...
+                # coerced to a dict
+                return Response(dict(serializer.data))
+
         elif request.method == 'PATCH':
-            if not asset.has_deployment:
-                raise exceptions.MethodNotAllowed(
-                    method=request.method,
-                    detail='Use POST to create a new deployment'
+            if not asset.can_be_deployed:
+                raise BadAssetTypeException(u"A form template can't be deployed")
```
jnm: Again, match https://github.com/kobotoolbox/kpi/pull/1806/files#diff-43986855ae1015d5975b107cf1ee9869R691

kpi/views.py
```diff
@@ -907,6 +1108,18 @@ def perform_create(self, serializer):
             user = get_anonymous_user()
         serializer.save(owner=user)
 
+    def partial_update(self, request, *args, **kwargs):
+        instance = self.get_object()
+
+        if CLONE_ARG_NAME in request.data:
+            serializer= self._get_clone_serializer(instance)
```
jnm: whitespace 🤣 sorry, couldn't resist!

jsapp/fonts/.gitignore
```diff
@@ -1,11 +0,0 @@
-*.eot
```
jnm: Watch out for this. We should probably fix the `Dockerfile` so that it doesn't delete this file. Removing this file may have broken your build.

kpi/tests/test_cloning.py
```diff
@@ -16,6 +16,8 @@
 )
 from .kpi_test_case import KpiTestCase
 from .test_assets import AssetsTestCase
+from kpi.constants import *
```
jnm: :grin: If importing [lots of constants](https://github.com/kobotoolbox/kpi/pull/1806/files#diff-7efb57e207546d622f02cddb15ffa899R99) is onerous, how about `from kpi import constants` and then using references like `constants.ASSET_TYPE_ARG_NAME` afterwards?

kpi/views.py
```diff
@@ -742,13 +871,78 @@ def _get_clone_serializer(self):
...
+                cloned_data["asset_type"] = self.request.data.get(ASSET_TYPE_ARG_NAME, original_asset.asset_type)
+
+            cloned_asset_type = cloned_data.get("asset_type")
+            # Settings are: Country, Description, Sector and Share-metadata
+            # Copy settings only when original_asset is `survey` or `template`
+            # and `asset_type` property of `cloned_data` is `survey` or `template`
+            # or None (partial_update)
+            if (original_asset.asset_type == ASSET_TYPE_TEMPLATE and cloned_asset_type == ASSET_TYPE_SURVEY) or \
+                (original_asset.asset_type == ASSET_TYPE_SURVEY and cloned_asset_type == ASSET_TYPE_TEMPLATE) or \
+                (cloned_asset_type is None and original_asset.asset_type in [ASSET_TYPE_TEMPLATE, ASSET_TYPE_SURVEY]):
```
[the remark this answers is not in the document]
noliveleger: Sure we can. 
I understood that `Asset.settings` should not be copied from `survey` to `survey`.  
I may have misunderstood with `Asset.content.settings` I guess.

kpi/models/asset.py
```diff
@@ -685,7 +687,7 @@ def _snapshot(self, regenerate=True):
         asset_version = self.asset_versions.first()
 
         _note = None
-        if self.asset_type in ['question', 'block']:
+        if self.asset_type in [ASSET_TYPE_QUESTION, ASSET_TYPE_BLOCK]:
```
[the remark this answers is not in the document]
noliveleger: Right. I would use the property `can_be_deployed` instead of the condition `asset_type == ASSET_TYPE_SURVEY`

kpi/models/asset.py
```diff
@@ -685,7 +687,7 @@ def _snapshot(self, regenerate=True):
...
             _note = ('Note: This item is a {} and must be included in '
```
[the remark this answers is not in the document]
noliveleger: ok

kpi/deployment_backends/mixin.py
```diff
@@ -45,3 +49,7 @@ def deployment(self):
             return DEPLOYMENT_BACKENDS[backend](self)
         except KeyError, e:
             raise KeyError('cannot retrieve asset backend: {}'.format(backend))
+
+    @property
+    def can_be_deployed(self):
+        return self.asset_type and self.asset_type != 'template'
```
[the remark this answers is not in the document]
noliveleger: You're totally right.

kpi/views.py
noliveleger: Fair enough :-D. I can't believe I missed that one. Bad copy&paste :P

kpi/tests/test_cloning.py
noliveleger: I agree with you. I forgot to specify the variables after refactoring the code. I will import the needed constants. I don't want to use `constants.VARIABLE`

kpi/views.py
```diff
@@ -742,13 +871,78 @@ def _get_clone_serializer(self):
...
+        """
+        Some business rules must be applied when cloning an asset to another with a different type.
+        It prepares the data to be cloned accordingly.
+
+        :param original_asset: Asset
+        :param source_version: AssetVersion
+        :param partial_update: Boolean
+        :return: dict|None
+        """
+        if self._validate_destination_type(original_asset):
```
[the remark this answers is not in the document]
noliveleger: It's just a personal preference when reading the code, but your way is perfectly fine for me too. 
I will update the code to match this discussion. 

So method `_get_clone_serializer` should look like that:

```
    def _get_clone_serializer(self, current_asset=None):
        ...
        cloned_data = self._prepare_cloned_data(original_asset, source_version, partial_update)
        if partial_update:
            return self.get_serializer(current_asset, data=cloned_data, partial=True)
        else:
           return self.get_serializer(data=cloned_data)
```

and 

```
    def _prepare_cloned_data(self, original_asset, source_version, partial_update):
       ...
        if self._validate_destination_type(original_asset):
            ...
        else:
            raise BadAssetTypeException("...")
```

kpi/models/asset.py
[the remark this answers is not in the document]
noliveleger: Actually, I'm wondering why this condition is here. It does nothing except assigning `_note` which is not used anywhere else.

kpi/models/asset.py
```diff
@@ -685,9 +687,9 @@ def _snapshot(self, regenerate=True):
         asset_version = self.asset_versions.first()
 
         _note = None
```
jnm: @noliveleger, it doesn't seem like this is passed along to anything else, and [it never was](https://github.com/kobotoolbox/kpi/commit/a9c94881745b031118ec598c32116b51238f072c#diff-43986855ae1015d5975b107cf1ee9869R496) to begin with. Am I right?

noliveleger: @jnm, that was my question in other comments

> Actually, I'm wondering if this condition is useful. It does nothing except assigning _note which is not used anywhere else.

magicznyleszek:
@jnm @noliveleger shouldn't `form-template` branch be deleted after merge? I can change #1813 to point to `master`

PR #1807, opened 2018-06-04
don't display unused button as it obstructs other UI elements

magicznyleszek:
On smaller screens (`< 999px`) we applied styles to the button that was not visible, but was covering other UI elements (esp. "Projects" menu item). We already set `display: none` to it on default, so I guess someone forgot about the `@media` styles? 

See screenshot:

screen shot 2018-06-04 at 10 50 26

### DIFF
--- a/jsapp/js/components/drawer.es6
+++ b/jsapp/js/components/drawer.es6
@@ -137,9 +137,6 @@ class Drawer extends Reflux.Component {
       stores.serverEnvironment,
     ];
   }
-  toggleFixedDrawer() {
-    stores.pageState.toggleFixedDrawer();
-  }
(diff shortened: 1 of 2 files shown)


pmusaraj:
@magicznyleszek from a quick overview, it looks like that button isn't being displayed at all. If that is the case, please completely remove the button and the styles linked to it. 

And we need to do a review of the app in smaller screen sizes (< 999px but also < 768px). I'll make a separate ticket for that though.

magicznyleszek:
@pmusaraj thanks for pointing it out, as obviously it is always `display: none` so no point having it 🤣 Added a removal commit.

PR #1808, opened 2018-06-04
Upgrade to Webpack 4, React 16.4.0, Node 8

pmusaraj:
This branch includes several under-the-hood improvements: 

- Webpack updated to version 4 (and Webpack config files have been updated to match)
- React updated to 16.4.0 (current version)
- `react-hotkey` is no longer up to date, using `react-shortcuts` instead
- `babel` and `eslint` libraries have also been updated
- `mocha-phantomjs` has been repalced by `mocha-chrome`
- some syntax and webpack config scripts cleanups
- remove `Open Sans` font, it was only used in multiselect choices in form builder (use Roboto instead)


package.json
```diff
@@ -49,53 +45,54 @@
     "lodash.merge": "^3.3.2",
     "lodash.restparam": "^3.6.1",
     "mime-types": "^2.1.8",
-    "mocha": "^2.3.4",
-    "mocha-phantomjs": "^4.0.2",
+    "mocha": "^5.2.0",
+    "mocha-chrome": "^1.1.0",
```
magicznyleszek: Have you considered adding https://github.com/GoogleChrome/puppeteer to not be dependant on the installed (or not installed) Chrome?

pmusaraj: I'm not familiar with Puppeteer but it sounds like it's overkill for what we need at this point. `mocha-chrome` runs headless Chrome for the tests. I updated the Travis config to include Chrome and that seems to be sufficient.

magicznyleszek: Yeah Puppeteer needs to download ~80 MB with every `npm install` :-P

jsapp/js/keymap.js
```diff
@@ -0,0 +1,6 @@
+export default {
+  APP_SHORTCUTS: {
+    EDGE: 'alt+e',
```
magicznyleszek: What is this for? When I use it I see "Active Projects" and "Deleted (#)" show up, but both of these does nothing when clicked

jsapp/scss/stylesheets/partials/form_builder/_card_multichoice.scss
```diff
@@ -21,133 +21,146 @@
...
-        color: #D08B8C;
-        font-size: 16px;
-        cursor: pointer;
+
+    .editable-buttons {
+      display: none;
+
+      // TEMPORARY
+
     }
```
magicznyleszek: Maybe instead of:

```scss
.editable-buttons {
  display: none;

  // TEMPORARY

}
```

this would be more clear:
```scss
// TEMPORARY
.editable-buttons {display: none;}
// ENDTEMPORARY
```

magicznyleszek:
@pmusaraj that's all from my side, feel free to merge after checking my review 👍

jsapp/js/app.es6
```diff
@@ -89,18 +91,31 @@ class App extends React.Component {
...
-        <div className="mdl-wrapper">
+        <Shortcuts
+          name='APP_SHORTCUTS'
+          handler={this._handleShortcuts}
+          className="mdl-wrapper"
+          global>
+
           { !this.isFormBuilder() && !this.state.pageState.headerHidden &&
-            <div className="k-header__bar"></div>
+            <div className="k-header__bar" />
```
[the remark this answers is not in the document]
pmusaraj: Yes, in fact it's now enforced (I think by `eslint-plugin-react`)

jsapp/js/keymap.js
pmusaraj: This is a way for us to build features but not expose them to users. It just toggles a class on the body (`is-edge`). For example, under Data, there is a tab for legacy reports. (Some of the things it enables don't do anything though, like the "Deleted (#)", we need to clean that up.)

magicznyleszek: Oh, so `is-edge` means "is bleeding edge"?

pmusaraj: Yes (I guess it could be confused with IE Edge).

PR #1811, opened 2018-06-05
change text in tags input

magicznyleszek:


jsapp/js/components/tagInput.es6
```diff
@@ -32,7 +32,7 @@ class TagInput extends React.Component {
 
   render() {
     var inputProps = {
-      placeholder: t('add tag')
+      placeholder: t('Add tag(s)')
```
magicznyleszek: @pmusaraj it is probably nitpicking, but is it possible to add multiple tags at once with that input? Because if not, then "Add tags" is never true 😛

Issue #1812, opened 2018-06-06
Question skip logic doesn't render properly


magicznyleszek:
## The issue

Steps to reproduce:

1. Create New Project → Form Builder opens
2. Add "Select One" question as first
3. Add "Text" question as second
4. Open "Settings > Skip Logic" for second question
5. "Add a condition" → select first question
6. Close second question "Settings"
7. Open "Settings > Skip Logic" for second question
8. Observe that previously configured condition is broken: 
  a. Selected "expression" is not displayed
  b. Selected "response" is not displayed
  c. You can change the question to "Select question from list" causing a JS error (propsed fix: #1814)

<img width="551" alt="screen shot 2018-06-06 at 10 20 05" src="https://user-images.githubusercontent.com/2521888/41025821-35bab64c-6973-11e8-9f88-c12e6147b313.png">

The error from console:
```js
Uncaught TypeError: this._get_question(...).get_type is not a function
    at SkipLogicCriterion.rowDetailsSkipLogic.SkipLogicCriterion.SkipLogicCriterion.change_question (app-274216cf7456a29f2ab3.js:89411)
    at SkipLogicPresenter.module.exports.skipLogicHelpers.SkipLogicPresenter.SkipLogicPresenter.change_question (app-274216cf7456a29f2ab3.js:88673)
    at HTMLSelectElement.<anonymous> (app-274216cf7456a29f2ab3.js:89864)
    at HTMLSelectElement.dispatch (app-274216cf7456a29f2ab3.js:24553)
    at HTMLSelectElement.elemData.handle (app-274216cf7456a29f2ab3.js:24361)
    at Object.trigger (app-274216cf7456a29f2ab3.js:27619)
    at HTMLSelectElement.<anonymous> (app-274216cf7456a29f2ab3.js:27697)
    at Function.each (app-274216cf7456a29f2ab3.js:19724)
    at jQuery.fn.init.each (app-274216cf7456a29f2ab3.js:19559)
    at jQuery.fn.init.trigger (app-274216cf7456a29f2ab3.js:27696)
    at constructor.triggerChange (app-274216cf7456a29f2ab3.js:97014)
    at constructor.onSelect (app-274216cf7456a29f2ab3.js:98311)
    at constructor.selectHighlighted (app-274216cf7456a29f2ab3.js:97716)
    at constructor.<anonymous> (app-274216cf7456a29f2ab3.js:96653)
[...]

PR #1814, opened 2018-06-07
Fix display issues with skip logic

magicznyleszek:
This fixes one of the problems from #1812 (the `8.c. You can change the question to "Select question from list" causing a JS error`). Now you can't select the placeholder value, which was possible in some cases before.

screen shot 2018-06-07 at 19 55 08


pmusaraj:
The last commit also adds fixes (workarounds, really) for the remaining issues of ticket #1812.

PR #1823, opened 2018-06-14
Fix minor styling issues

### LINKED ISSUE #1818: Opening tags moves project details up
Minor one, but irritating - after opening tags editor, the project details (name, date, etc.) move up by few pixels; plus the amount of whitespace between name and tags editor is too big:

screen shot 2018-06-09 at 15 40 47

screen shot 2018-06-09 at 15 41 01

PS. While doing this, verify if this function could be deleted - from `tagInput.es6`:

```es6
handleChange(tags) {
  var transformed = tags.map(function(tag) {
    // Behavior should match KpiTaggableManager.add()
    return tag.trim().replace(/ /g, '-');
  });
  this.setState({tags: transformed});

  var uid = this.props.uid;
  actions.resources.updateAsset(uid, {
    tag_string: transformed.join(',')
  });

}
```

pmusaraj:
Fixes #1818. 

Also improves button display when hovering over an asset row: 

- less button spacing
- background color only on buttons, fixes a poor-looking overlap on the Created columns (for certain screen sizes)
- align "more actions" menu better with the button

### DIFF
--- a/jsapp/scss/components/_kobo.asset-row.scss
+++ b/jsapp/scss/components/_kobo.asset-row.scss
@@ -3,12 +3,13 @@ $hover-asset-row: #F2F2F3;
 .asset-row {
   list-style-type: none;
   border-top: 1px solid $divider-color;
-  padding: 4px 12px;
+  padding: 6px 12px;
   white-space: nowrap;
   transition: max-height 0.5s ease, opacity 0.5s ease;
-  min-height: 55px;
+  min-height: 52px;
   opacity: 1;
   position: relative;
+  align-items: flex-start;
 
   &:hover {
     background: $hover-asset-row;
@@ -106,7 +107,6 @@ $hover-asset-row: #F2F2F3;
     position: absolute;
     top: 8px;
     right: 0px;
-    background: $hover-asset-row;
     width: 0;
     overflow: hidden;
     opacity: 0;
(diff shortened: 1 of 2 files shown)


pmusaraj:
@magicznyleszek these aren't the same buttons as what you've linked to, those are in a modal window, if I'm not mistaken. 

Feel free to try the padding change, I'd be happy to review that.

jsapp/scss/components/_kobo.asset-row.scss
```diff
@@ -105,20 +109,13 @@ $hover-asset-row: #F2F2F3;
 
   > .asset-row__buttons {
     position: absolute;
-    top: 8px;
-    right: 0px;
-    width: 0;
-    overflow: hidden;
-    opacity: 0;
+    top: ($s-asset-row - $s-asset-row-icon) * 0.5;
```
pmusaraj: I'm not sure why you changed this from a simple pixel value to a calculation. From what I can tell, the calculation results in 8px too...

jsapp/scss/components/_kobo.asset-row.scss
```diff
@@ -163,9 +167,17 @@ $hover-asset-row: #F2F2F3;
...
+      transition: $hover-transition;
+      // make them appear moment after items behind starts to disappear
+      transition-delay: 0.1s;
+    }
+
+    // hide all columns but first one
+    .asset-row__cell--asset-details > .asset-row__cell:not(:first-of-type) {
+      opacity: 0;
+      visibility: hidden;
+      transition: $hover-transition;
```
pmusaraj: This is not the desired behaviour. In the previous commit, I specifically removed the width of `.asset-row__buttons` so that we could see the `Created` or `Modified` value of that row even on hover.

jsapp/scss/components/_kobo.asset-row.scss
magicznyleszek: I did it because these values are connected - but I changed this top to work for rows with description (wasn't centered) and removed those unnecessary variables. I guess premature optimization :P

jsapp/scss/components/_kobo.asset-row.scss
magicznyleszek: Oh, ok; removed it :) (modified is actually being hidden by some code that was already there)

magicznyleszek:
@pmusaraj I applied your comments

Issue #1832, opened 2018-06-21
Better error message instead of "failed to update profile"


jeverling:
When I edit my profile, and enter invalid data, e.g. an invalid email or no email, *kpi* is just diplaying `failed to update profile`. The proper error message that *kpi* responds with, e.g. `{"email":["Enter a valid email address."]}` is not displayed to the user.

pmusaraj:
Please take a look at this @magicznyleszek

PR #1836, opened 2018-06-25
Show better error message for update profile

magicznyleszek:
fixes https://github.com/kobotoolbox/kpi/issues/1832


magicznyleszek:
Notes from meeting:

* implement some DRY inline text fields error and use it in settings
* after implementing it create issue to apply those elsewhere (e.g. new project form name)
* use `{"field name": ["error", "error"]}` errors to display inline messages for user

magicznyleszek:
@pmusaraj this is now ready to be reviewed :)

PR #1842, opened 2018-06-29
This PR exists only to test Travis CI

jnm:


jnm:
👋
goodbye, but don't you cry, i'll be back again someday
☃

PR #1867, opened 2018-07-13
*._isSelectQuestion undefined fix

magicznyleszek:
Reproduction steps:
1. in form builder:
2. add Select One question
3. go to question Settings > Skip Logic > + Add a condition
4. close Settings
5. expand queston
6. select questino (blue border thingy)
7. click delete icon on one of the options

Currently there is an error, and it should no longer be there :)

Fixes http://sentry.kbtdev.org/kobo/kpi-frontend/issues/8446/


pmusaraj:
Great, thanks for tracking this down!

PR #1870, opened 2018-07-15
Fix bad setCollectMethod dataset reading

### LINKED ISSUE #1869: Form landing reloads when choosing Collect data method (sometimes)
Reproduction steps:
1. Open form landing of a project
2. Scroll down to `Collect data` selector
3. Open selector and on any option click the area that is not a text (see image below)

screen shot 2018-07-15 at 09 39 08

Sentry issue `KPI-FRONTEND-MZ` http://sentry.kbtdev.org/kobo/kpi-frontend/issues/8441/ (`Cannot read property 'label' of undefined`)

magicznyleszek:
fixes #1869


pmusaraj:
Thanks @magicznyleszek, great fix. We should use `currentTarget.dataset` elsewhere as well, there are a few places in the code where we use data attributes, and use jquery to get to the data attributes, and it's inefficient.

PR #1875, opened 2018-07-19
Fix #1422 by optimizing performance of API asset lists

### LINKED ISSUE #1422: Slow API call to asset list when limit set to 200
API calls to `assets/?q=asset_type%3Asurvey&limit=200` in an account with 200 forms take ~15 secs on a test server. 

Introduced after the temporary switch to 200 records per API call to fix #1380.

jnm:
Response time for a user with 137 surveys (comprising about 184M of data) decreases from ~20 seconds to ~2 seconds.
```
In [10]: aa = Asset.objects.filter(asset_type='survey', owner=u)

In [11]: aa.count()
Out[11]: 137

In [12]: len(pickle.dumps(aa)) / 1024 / 1024
Out[12]: 184
```
image

### DIFF
--- a/kpi/deployment_backends/kc_access/shadow_models.py
+++ b/kpi/deployment_backends/kc_access/shadow_models.py
@@ -77,6 +77,7 @@ class LazyModelGroup:
             date_modified = models.DateTimeField()
             uuid = models.CharField(max_length=32, default=u'')
             last_submission_time = models.DateTimeField(blank=True, null=True)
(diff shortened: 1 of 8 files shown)


jnm:
Need to look at this before merging: https://sentry.kbtdev.org/kobo/kpi/issues/13722/

jnm:
`filters.py` is about to change in `master` (5fe9dea) and will require revising 6dff086

PR #1813, opened 2018-06-06
Feature: Form Templates

magicznyleszek:
There will not be much (if any) changes to FE as for now, so I think @pmusaraj you can take a look :) Plus this merge goes to the `form-template` main branch, so further FE changes can be done there.

fixes #1816 #1810 #1634 #1633

PR #1886, opened 2018-07-24
Cleanup old coffee code, fix browser tests

### LINKED ISSUE #1879: Remove unused .sub-header-bar code
Whole `.sub-header-bar` is set to `display: none`, so none of the code that touches it is being used by it (it could be used by different things though):

screen shot 2018-07-23 at 17 06 50

Try following and investigating all the code-breadcrumbs before removing `.sub-header-bar`.

pmusaraj:
- Removes old UI code in `.coffee` files in the form builder, fixes #1879
- Fixes broken `/browser_tests/` endpoint, now you can visit that page to see the tests running correctly
- PR is built on top of `form-template-frontend`, please only merge after #1813 is merged

Minor note: I added the `tests` bundle to the production environment too, and tests run, but for some reason, some tests fail in the production environment. I will look into this further.


jsapp/scss/components/_kobo.form-builder.scss
```diff
@@ -218,7 +214,9 @@ $t-form-builder-aside-open: 200ms;
 
       button {
         color: $cool-gray;
-        overflow: visible; // for tooltip
+        overflow: visible;
+
+        // for tooltip
```
magicznyleszek: something weird is happening with comments and this is second time I see it in your commit - is this some autofix? `// for tooltip` refers to `overflow: visible;`, so if not on same line, it should be above it :)

pmusaraj: Oh, yeah, I use a sass beautifier that messes these up.

PR #1889, opened 2018-07-25
Feature: allow replacing project with template

### LINKED ISSUE #1810: A user can choose a template when replacing an existing project
When replacing a form, users currently can choose to replace from an XLSForm file (on their device or using a URL). 

User should also be able to replace the form with a template (provided there are templates on his/her account).

Note also that the label for the replace option is currently "Replace with XLS". I suggest we update it to simply "Replace".

### LINKED ISSUE #1887: Enable replacing project with template and fix replacing questions
A separate issue from `form-template-frontend` branch.

"Choose template" option is currently hidden for "Replace project" flow. It mostly works - the big issue is that when replacing project with a template, the questions are not overwritten.

magicznyleszek:
## Description

Replacing project with template implementation. FE uses PATCH on asset with `clone_from` parameter (there is a bug, see related issues).

## Related issues

fixes #1887 
fixes #1810
~~blocked by #1892~~


pmusaraj:
Something isn't right here. After replacing a project from a template, I get an error when trying to preview in Enketo. 

Also, sometimes if I go to the form builder immediately after replacing with a template, I see the old form questions in the builder. I would need to refresh to see the new questions.

magicznyleszek:
@pmusaraj ready to be tested :)

jsapp/js/actions.es6
```diff
@@ -383,17 +383,21 @@ actions.resources.listTags.completed.listen(function(results){
   }
 });
 
-actions.resources.updateAsset.listen(function(uid, values, params={}) {
+actions.resources.updateAsset.listen(function(uid, values, callbacks={}) {
```
pmusaraj: We shouldn't change `params` to `callbacks` here, and we shouldn't change `onComplete` to `onCompleted` below. It's not necessary to make these changes to nomenclature, it is longer to review and it's also prone to merge conflicts (the `searchStore` branch makes changes in the same lines, roughly).

jsapp/js/actions.es6
```diff
@@ -383,17 +383,21 @@ actions.resources.listTags.completed.listen(function(results){
...
-        params.onComplete(asset);
+      if (callbacks.onCompleted) {
+        callbacks.onCompleted(asset);
       }
       notify(t('successfully updated'));
     })
-    .fail(function(resp){
+    .fail((resp) => {
       actions.resources.updateAsset.failed(resp);
+      if (callbacks.onFailed) {
```
pmusaraj: none of these `onFailed` callbacks are called, maybe we should just not have them. Makes for cleaner code.

pmusaraj: I see where they are used, but shouldn't we use the `actions.resources.updateAsset.failed` listener here?

pmusaraj:
Naming changes look good.

magicznyleszek:
Currently blocked by BE issue: `dictionary update sequence element #0 has length 1; 2 is required`

magicznyleszek:
@pmusaraj ping :)

pmusaraj:
Tested this on `unu` @magicznyleszek and I still have an issue: after replacing a project with a template, when I go to the form builder, I see the project's original questions. I should see the new questions, from the template.

magicznyleszek:
@pmusaraj fond the problem - Form Builder uses allAssets store, which unlike searches wasn't being updated on updateAsset action. Deployed on `unu` for retest :)

magicznyleszek:
@pmusaraj ping ;-)

jsapp/js/components/modalForms/projectSettings.es6
```diff
@@ -507,12 +553,10 @@ class ProjectSettings extends React.Component {
...
 
-          {this.props.context !== PROJECT_SETTINGS_CONTEXTS.REPLACE &&
-            <button onClick={this.displayStep.bind(this, this.STEPS.CHOOSE_TEMPLATE)}>
-              <i className='k-icon-template' />
-              {t('Use a template')}
-            </button>
-          }
+          <button onClick={this.displayStep.bind(this, this.STEPS.CHOOSE_TEMPLATE)}>
+            <i className='k-icon-template' />
+            {t('Use a template')}
```
jnm: I don't think this should appear before the XLSForm options when replacing a project, since those are likely to be used far more frequently.

jnm:
I tested this and it seems to work well. Penar may not have time to do any more reviews. I'd be happy to merge once the order is changed as I indicated above.

jsapp/js/components/modalForms/projectSettings.es6
magicznyleszek: @jnm would it be ok if template option would be last when replacing project and also when creating project?

magicznyleszek: @jnm I am displaying the template as last for replace project, but as second for new project :)

PR #1895, opened 2018-07-28
add issue and PR template, move contributing to same directory

magicznyleszek:
I used https://github.com/stevemao/github-issue-templates as inspiration

Idea 1: We could have prefixes for all things, not only "Feature:", similar to https://seesparkbox.com/foundry/semantic_commit_messages


magicznyleszek:
@pmusaraj moved it back to root 👍

PR #1899, opened 2018-07-29
Enhancement: add browserslist configuration

### LINKED ISSUE #1944: Have supported browsers set explicitly
## Description

We need to explicitly set the browsers we support, so our CSSs and JSs are built properly. It will stope some Sentry errors from happening. Plus I'm guessing some people would like to know if their browser supports KPI.

magicznyleszek:
## Description

I'm not sure if `browserslistrc` has the config we want (I was inspired by default one, just changed last 1 to last 2 versions), but now we will have control over supported browsers <3

@pmusaraj @jnm do you know/remember anything about browsers that we explicitly want to support?

## Related issues

Fixes #1944


magicznyleszek:
Note: add IE11

jnm:
We could consider matching Enketo: https://enketo.org/faq/#browsers

magicznyleszek:
@jnm @pmusaraj I updated the list, to see what browsers it concerns, see http://browserl.ist/?q=last+1+major+versions%2C+%3E+1%25%2C+not+dead%2C+IE+11

.browserslistrc
```diff
@@ -0,0 +1,6 @@
+# Browsers that we support
+
+last 1 major versions
```
jnm: In your original description of the PR, you mentioned `last 2 major versions`. Any reason in particular for the change?

We should also support Firefox ESR. When answering support requests, I've seen multiple folks running that in institutional and corporate environments.

magicznyleszek: I did 2 versions

PR #1904, opened 2018-07-31
Misc ui fixes

### LINKED ISSUE #1903: Miscellaneous UI issues
Some regressions have shown up in the UI, likely due to the many recent changes and merges. A short list below: 

- [x] submission modal, fix close icon positioning (it stays visible even when scrolling, at least in Safari)
- [x] several modals have a horizontal scrollbar, map settings for example, or reports (in Safari, at least)
- [x] map settings marker colors screen, save button is missing styling: https://monosnap.com/file/zRIBIOhaSdpMT37pYA3Bv0bufAoFT9
- [x] Library sidebar in FormBuilder, has an empty tag, https://monosnap.com/file/ILFcwe1ZBuHf6zu7GFfjzGq1GC72Vt 
- [x] in asset list, when toggling tags, buttons are vertically aligned in the row. they shouldn't be. 

And, not a new issue, but it's good to fix: 
- [x] Add favicon for the app (it returns a 404 on Safari currently)

magicznyleszek:
fixes #1903 - check isssue for a list of bugs

### DIFF
--- a/jsapp/scss/libs/_mdl.scss
+++ b/jsapp/scss/libs/_mdl.scss
@@ -7,7 +7,7 @@ $cool-gray: lighten($black, 30%);
 $cool-silver: lighten($black, 40%);
 $cool-lightgray: lighten($black, 60%);
 $divider-color: lighten($black, 63%);
-$cool-whitegray: lighten($black, 68%);
+$cool-whitegray: lighten($black, 67%);
 $cool-blue: #2095f3;
 $cool-green: #52c5d0;
 $warm-green: #52ce7b;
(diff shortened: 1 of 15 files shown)


jsapp/scss/libs/_mdl.scss
```diff
@@ -1,10 +1,13 @@
 // Theming variables, mixins and includes
 
 $black: #414453;
+$cool-charcoal: lighten($black, 10%);
```
pmusaraj: I'm not sure that these variables are better than using the SCSS mixins directly... maybe it's just a personal preference.

PR #1921, opened 2018-08-12
Bugfix: block survey routes from non-surveys

### LINKED ISSUE #1943: Survey routes shouldn't be accessible for non-surveys
## Description

Currently you can access every survey related route with other asset types by replacing uid in url. We need to block these routes for non-surveys.

## Steps to Reproduce

1. Open any library item and copy it's `uid` from url
2. Open any form's landing view
3. Replace form's `uid` with copied library item `uid` in url and wait for page load

## Expected behavior 

The route should not be accessible.

## Actual behavior 

There is a DEPLOY button (broken) and some other things visible.

magicznyleszek:
## Description

Currently you can access every survey related route with other asset types by replacing `uid` in url. We need to block these routes for non-surveys.

@pmusaraj this is not a perfect solution - I override rendering in all components that are being used in forbidden routes. 

A better and future-proof solution would be to add this logic in Router, but it would require: 
a) rewriting Router
b) needing to know `asset_type` at Router level

Currently we only have `uid` there, so I would need to check if we already have given asset in a store and check the type or make a call to BE for the data. It's doable, but more time consuming - what do you think?

## Related issues

Fixes #1943


magicznyleszek:
@jnm  what do you think?

magicznyleszek:
As we work with #2293 it seems our direction would be to allow displaying asset routes for all asset types - we would limit the available elements depending on type. Thus this is no longer needed.

PR #1924, opened 2018-08-13
better error message for import failure

magicznyleszek:
## Description

Use error message from backend instead of showing general one.

@jnm btw. should I check the FE code and do similar thing for other error handling? I could do it for all BE calls FE makes - the only problem is that the responses differ: sometimes error message is in `responseJSON.detail`, sometimes in `messages.error`, and sometimes error response is html code - created an issue for this idea: #1926

### DIFF
--- a/jsapp/js/components/formEditors.es6
+++ b/jsapp/js/components/formEditors.es6
@@ -455,8 +455,12 @@ export class ProjectSettings extends React.Component {
                 alertify.error(t('Failed to reload project after upload!'));
               });
             },
-            () => {
-              alertify.error(t('Could not initialize XLSForm upload!'));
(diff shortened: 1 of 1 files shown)


joeflack4:
Any idea when this update will enter production?

We are currently experiencing confusing "import failure" errors.

invalid xlsform import error

In our case, this was due to an invalid xlsform (attached). There were certain rows which had invalid syntax under "type", and the row was empty otherwise. This is because what I uploaded was accidentally one of our custom XlsForm templates, not an actual, valid form.

I'll open up a new issue "Better error message for import failure (part II)" (#1962), just in case this issue still exists in the codebase.

PR #1927, opened 2018-08-14
Enhancement: display frontend build commit in dev env

### LINKED ISSUE #1925: Enhancement: include git commit hash in webpack build
## Description

So it's easier to see what "version" the built frontend is in.

## Additional details

Some idea: https://stackoverflow.com/questions/38400314/including-git-commit-hash-and-date-in-webpack-build

magicznyleszek:
## Description

Added more info to gitRevInfo and moved it to top right corner, out of the way. I changed it to red (so it doesn't look like part of UI as much) and added opacity change on hover.

NOTE: if you run `npm run watch` and switch commits, the commit hash will not update

@pmusaraj @jnm a question: would we want the build commit on production too? E.g. it could be accessible through `window.kpi.frontendCommit`.

screen shot 2018-08-14 at 17 16 46
screen shot 2018-08-14 at 17 17 02
screen shot 2018-08-14 at 17 17 12

## Related issues

Fixes #1925


jnm:
@magicznyleszek, I'm not seeing the `FE commit` when building locally or deploying to `kvar`. I do see your other changes, though. Is there something I have to enable during the build process?

magicznyleszek:
@jnm to summarize what I wrote on Flowdock: during deploy we clone shallow copy of KPI repository, so webpack is not able to ask for git commit. I added a safe check for git repository - `FE commit` will not show in this case.

magicznyleszek:
@jnm IIRC this should be dropped?

Issue #1928, opened 2018-08-14
Question Matrix data is not displayed correctly in the auto-report


magicznyleszek:
e.g. question names do not appear, just the first letter.

magicznyleszek:
I fixed one small problem with #1929, but further work will require team planning

PR #1929, opened 2018-08-14
display more than single letter for matrix in reports

magicznyleszek:
## Description

A small but useful change: for Question Matrixes in reports, user will see `group_UID_FIRST_QUESTION_SECOND_QUESTION` instead of `g`. More changes in the related issue will require BE + FE changes.

## Related issues

Part of #1928

### DIFF
--- a/jsapp/js/components/reports.es6
+++ b/jsapp/js/components/reports.es6
@@ -530,7 +530,12 @@ class ReportContents extends React.Component {
       <div>
         {
           reportData.map((rowContent, i)=>{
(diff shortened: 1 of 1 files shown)


pmusaraj:
This is a good improvement, thanks @magicznyleszek! 

In the future, though, we should try to display the labels of the matrix cells (rowlabel / columnlabel). Right now this displays something like this: `group_df76n42_row_column` where 

group_df76n42 = the matrix group's data column name 
row = data row prefix
column = data column suffix

PR #1936, opened 2018-08-15
Enhancement: Make long table column names viewable

### LINKED ISSUE #1935: Table view: Allow user to show full column headers (e.g. wrap them with new lines, smaller font)
## Description

Currently when you have a column with long name, it get's stripped with ellipsis and there is no way to see it whole.

PS. @tinok I changed your idea board note into issue :)

magicznyleszek:
## Description

Override react-table styles and allow column name wrapping. I decided to additional break words by letters (if by whitespace is not possible), because we often have very_long_snake_cased column names.

## Screenshots

Before:
screen shot 2018-08-15 at 18 03 15

After:
screen shot 2018-08-15 at 18 03 35

## Related issues

Fixes #1935

### DIFF
--- a/jsapp/js/components/table.es6
+++ b/jsapp/js/components/table.es6
@@ -329,7 +329,10 @@ export class DataTable extends React.Component {
 
       columns.push({
         Header: h => {
-          return _this.getColumnLabel(key, q, qParentG);
+          const columnName = _this.getColumnLabel(key, q, qParentG);
+          return (
+            <span title={columnName}>{columnName}</span>
+          );
         },
         id: key,
         accessor: row => row[key],
--- a/jsapp/scss/components/_kobo.table.scss
+++ b/jsapp/scss/components/_kobo.table.scss
@@ -53,16 +53,6 @@
     }
 
     .rt-thead {
-      // allow wrapping labels for long column names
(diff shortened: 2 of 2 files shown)


pmusaraj:
This needs some before/after screenshots.

pmusaraj:
I’m afraid rhis would cover too much screen space for longer questions.

magicznyleszek:
@pmusaraj actually you're right, the higher column name row makes current table bug happen more often. 

So previously - data inaccessible at viewport smaller than 350px height:
screen shot 2018-08-16 at 14 36 54

After proposed changes - data inaccessible at viewport smaller than 600px height (in this particular scenario):
screen shot 2018-08-16 at 14 27 34

Thanks for pointing this out - I will try introducing some tooltip for column names (maybe `title` attribute for lighter alternative to `data-tip`). Plus I will take a look into making the data accessible on smaller vertical viewports

magicznyleszek:
@pmusaraj now at least 2 rows of body will always be visible and when you hover over the column name, system tooltip appears:
screen shot 2018-08-16 at 14 55 13

The scrollbar situation is far from perfect, but at least the data can be viewed:
screen shot 2018-08-16 at 14 57 18

PR #1938, opened 2018-08-16
Header name fixes

### LINKED ISSUE #1756: Enter key and Edit Name/Description
1. For this name input at the top of the page: 
```
XPATH, "//input[@type='text' and @placeholder='Project title']"
```
It would be help to add an Enter key to save an edit. Currently, I think I needed a Tab key. Would be great to have both.

2. For this name input on the Settings page:
```
XPATH, //input[@type='text' and @placeholder='Enter title of project here']
```
It might be helpful to also add an Enter. For now it saves the form, which may also be a desired behavior(?)

3. For the description input on the Settings page, the Enter key starts a new line.
That's probably an appropriate behavior.

In essence, 
- on the Name input at the top of the form, Enter seems to do nothing. 
- On the Settings page in the Name input Enter submits the form. 
- On the Settings page in the Name input Enter starts a new line. 

Three different behaviors. I think the main ask here would be to change the behavior of item 1 above, and then potentially leave the other two as-is. (Another option for the Description is to have the Enter key do a Save, and expect that people won't use a newline in a Description (or accept a Shift-Enter as a new line.) In that case, all three would consistently save the new content via Enter.

### LINKED ISSUE #1883: Remove form name editing in app header
screen shot 2018-07-24 at 15 28 04

This change promotes more focused name editing flow for users - too many ways could be confusing.

magicznyleszek:
## Description

Too many places where you can change the name of the form are confusing.

Things in this PR:
- ~~header form name is no longer an editable input~~
- long form names are displayed in smaller font to fit them
- :bug: don't show old asset name while loading new one (especially visible when going from form to list view and to other form) - ~~now there is a spinner while loading the asset~~
- :bug: fixed asset row long text ellipsis-trimming (title, description and dates)
- some cleanups and dead code removals
- immediate title update with ENTER key

## Related issues

Fixes #1883
Closes #1756


magicznyleszek:
Note: undo editing header changes, keep bugfixes

magicznyleszek:
@pmusaraj I undid unwanted changes

PR #1939, opened 2018-08-16
Bugfix: Fix broken Add Question to Library button

### LINKED ISSUE #1859: Blank "no name" question added to library instead of my actual question
## Description
When adding some questions to the question library, rather than the actual question being added, an empty "no name" question with essentially no data is added.

In addition to screenshots of this, I've added an export of the "no name" question, as well as the original form where this issue occurred.

## Screenshots
1. Adding a question

2. The most recent entry appearing in question library after adding

3. What the question looks like in the editor

---
**Attachments**
1. [no_name form downloaded.xls.zip](https://github.com/kobotoolbox/kobocat/files/2180670/no_name.form.downloaded.xls.zip)
2. [full form.xls.zip](https://github.com/kobotoolbox/kobocat/files/2180690/full.form.xls.zip)

magicznyleszek:
## Description

The button was broken basically for all types other than `select_one` or `select_multiple`.

Interestingly it was broken for over half a year: https://github.com/kobotoolbox/kpi/commit/5d5360fb9 🕵️ 

## Related issues

Fixes #1859

### DIFF
--- a/jsapp/js/models/surveyScope.es6
+++ b/jsapp/js/models/surveyScope.es6
@@ -11,21 +11,22 @@ class SurveyScope {
   add_row_to_question_library (row) {
     if (row.constructor.kls === 'Row') {
       var rowJSON = row.toJSON2();
+      let content;
       if (rowJSON.type === 'select_one' || rowJSON.type === 'select_multiple') {
         var surv = this.survey.toFlatJSON();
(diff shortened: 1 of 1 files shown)


joeflack4:
> Interestingly it was broken for over half a year
Well gollee. 

Thanks for the fix.

Issue #1690, opened 2018-03-27
Only mark form as needing redeployment if form contents changes 


pmusaraj:
Any change to the project's settings (including name, description, or other details) marks the form as requiring to be redeployed -- even though the form itself hasn't changed. 

Expected behavior: Unless the form itself is changed, don't prompt the user to redeploy.

Supersedes #1469.

pmusaraj:
Feel free to have a look at this Leszek, it's a little tricky but may be interesting.

pmusaraj:
See also https://github.com/kobotoolbox/kpi/issues/443

smutnyleszek:
I will read the code and see how it works :)

alnf:
It would be nice to have this implemented, since the message about redeploy is very misleading.

jnm:
`<name>` is actually included in the XForm generated by KC, so to get a name change to appear in Collect or Enekto, a redeployment _is_ required. Changes in the other metadata have no bearing on the deployed XForm, so the message should be suppressed for them.

If we decide what we want to consider redeployment-worthy content, then we could hash that and have the backend report the hash in the API representation. The frontend could compare the latest content hash (instead of version ID) to the deployed content hash and display the redeployment message only if the hashes differ. I'd suggest including `AssetVersion.name` and `AssetVersion.version_content` in the hash.

There's also an `AssetVersion.deployed_content`, but it seems mostly unused:
```
>>> AssetVersion.objects.exclude(deployed_content=None).order_by('date_modified').first().date_modified
datetime.datetime(2016, 10, 3, 6, 41, 44, 4000, tzinfo=<UTC>)
>>> AssetVersion.objects.exclude(deployed_content=None).order_by('date_modified').last().date_modified
datetime.datetime(2016, 10, 18, 6, 52, 40, 822997, tzinfo=<UTC>)
>>> AssetVersion.objects.exclude(deployed_content=None).count()
42
>>> AssetVersion.objects.count()
1039699
```

pmusaraj:
Comparing hashes sounds good to me. And I agree with including `AssetVersion.name` and `AssetVersion.version_content`.

jnm:
Blocked by https://github.com/kobotoolbox/kpi/issues/1833

magicznyleszek:
@jnm I see that #1833 is closed - does it mean it's not blocked anymore?

magicznyleszek:
Waiting for #1941 and frontend work can start

[...]

PR #1941, opened 2018-08-16
Implement version content hashing per #1690

jnm:
Back-end changes only!

### DIFF
--- a/kpi/models/asset_version.py
+++ b/kpi/models/asset_version.py
@@ -58,7 +58,7 @@ class AssetVersion(models.Model):
     def content_hash(self):
         # used to determine changes in the content from version to version
         # not saved, only compared with other asset_versions
-        _json_string = json.dumps(self.version_content,sort_keys=True)
+        _json_string = json.dumps(self.version_content, sort_keys=True)
         return hashlib.sha1(_json_string).hexdigest()
 
     def __unicode__(self):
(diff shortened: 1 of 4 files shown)


kpi/models/asset_version.py
```diff
@@ -54,14 +54,11 @@ def to_formpack_schema(self):
...
+    def content_hash(self):
         # used to determine changes in the content from version to version
-        # not saved, only compared with other asset_versions (in tests and
-        # migration scripts, initially)
-        _json_string = json.dumps({'version_content': self.version_content,
-                                   'deployed_content': self.deployed_content,
-                                   'deployed': self.deployed,
-                                   }, sort_keys=True)
+        # not saved, only compared with other asset_versions
+        _json_string = json.dumps(self.version_content,sort_keys=True)
```
noliveleger: what about adding a lovely space after the comma something like `version_content,  sort_keys=true)`

magicznyleszek: @jnm @noliveleger have you considered using https://www.pylint.org to not bother about such problems (plus avoid some coding errors)?

noliveleger: @magicznyleszek, not a bad idea even if I already have a linter for python integrated in my IDE. 
Maybe when we upgrade to python3/django 2.x

PR #1950, opened 2018-08-18
Update react-select

### LINKED ISSUE #1949: Graphical issue; height of "Sharing Permissions" modal for sharing a collection
## Description
When opening the View/Edit selector menu of the "Sharing Permissions" modal for sharing a collection, the menu that opens becomes clipped off at the bottom.

A scroll bar is visible, so a user could potentially scroll down to click the option that they are looking for. But of course that is poor UX. It would be better design to simply change the positioning or height of the modal itself or its inside elements to prevent this issue from occurring.

## Steps to Reproduce

1. Log into KoBo Toolbox and go to the library module.
2. Go to a collection that you own / have administrative rights to.
3. Click the "..." to open a menu and click the option to share.
4. In the "invite collaborators" section of the modal, click the permissions (View/Edit) dropdown menu.
5. The graphical issue should now be visible.

## Expected behavior 
No graphical issue when opening the "Sharing Permissions" modal for sharing a collection.

## Actual behavior 
Graphical issue appears; the menu that opens becomes clipped off at the bottom.

share-library-collection graphical-bug

magicznyleszek:
## Description

Current version of `react-select` was missing `menuPlacement` option needed for the fix, so I took the opportunity to update it to latest version. I was also able to do some cleanups in our custom styles.

## Related issues

Fixes #1949


pmusaraj:
I like the new arrow down icon for selects, but the UI is now a little inconsistent: 

- [ ] in submission modal, the Validation status select no longer has a placeholder (it should have "Select..."
- [ ] in submission modal, dropdown option text should be left-aligned
- [ ] in table view, placeholder "Select..." is missing
- [ ] in account settings, dropdown font size is smaller than it used to be
- [ ] in account settings, city/country are vertically misaligned
- [ ] in account settings, maybe sector and gender should be 50/50 columns, like city/country
- [ ] in all Selects, the selected item's blue background is too bright. we should use the Kobo blue here. 
- [ ] in form builder sidebar, the hover effects are inconsistent with our color scheme

pmusaraj:
@magicznyleszek I just noticed that this branch also breaks the "frozen column" feature in the table display.

magicznyleszek:
Ok I will fix those issues :)

magicznyleszek:
@pmusaraj fixed all issues

magicznyleszek:
@pmusaraj I don't experience any of these locally or on unu, maybe it depends on a form? On unu I created a new one with map question, locally I checked on old one.

PR #1973, opened 2018-08-29
Feature: endpoint for hash of all assets

### LINKED ISSUE #1894: Add endpoint that returns hash of all assets
## Description

Frontend needs this to know if it makes sense to request the big list of all assets. We currently work on better asset data storage on FE #1885 and this is required to take it one step further.

## Related issues

Blocks #1960

noliveleger:
Adds endpoint at `assets/hash` with hash of all assets, based on `version_id`. Frontend will consume this endpoint and decide whether to request a refresh of the asset list or not. 

Fixes #1894.

### DIFF
--- a/kpi/fixtures/test_data.json
+++ b/kpi/fixtures/test_data.json
@@ -106,22 +106,6 @@
         }
     },
     {
-        "model": "kpi.asset",
-        "pk": 2,
-        "collection": 1,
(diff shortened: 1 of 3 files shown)


noliveleger:
@pmusaraj, I think you can review it. Otherwise @jnm may take a look at it.

PR #1913, opened 2018-08-02
UI for form languages and translations 

pmusaraj:
## Description

This PR adds UI that lets users add and update translations on a form. See screenshots below for some details. When testing, please carefully review the changes in the form builder, the PR strips out a good amount of translation-related UI there. 

## Screenshots 

screen shot 2018-08-02 at 11 49 07 am

screen shot 2018-08-02 at 11 52 25 am

screen shot 2018-08-02 at 11 52 36 am

screen shot 2018-08-02 at 11 53 07 am

## Issues

I didn't have time to do one last feature here: allow users to either: 

- reorder languages or
- set the default language (the one that's loaded in the form builder)

@magicznyleszek please look into the possibility of adding this feature (might require some backend work too)

Note: the PR also reorganizes the modal forms, by putting them all in a subfolder. 

## Related issues

Closes #1948
Closes #1273 
Closes #1175
Closes #949
Closes #1166
Closes #1144
Closes #1381 
Closes #327

PR #1975, opened 2018-08-29
Make sure the `default_language`, if any, is first

### LINKED ISSUE #1942: Reorder translations by default_language setting
## Description

PATCHing asset with `content.settings.default_language` set, should reorder languages/translations in such way that default_language is first.

## Additional details

Needed for #1913

jnm:
Closes #1942. Marking as high priority since we accidentally deployed a half-baked version of #1913 and need this to finish the job properly.

### DIFF
--- a/kpi/models/asset.py
+++ b/kpi/models/asset.py
@@ -218,14 +218,18 @@ class FormpackXLSFormUtils(object):
     def _remove_empty_expressions(self, content):
         remove_empty_expressions_in_place(content)
 
-    def _adjust_active_translation(self, content):
-        # to get around the form builder's way of handling translations where
-        # the interface focuses on the "null translation" and shows other ones
-        # in advanced settings, we allow the builder to attach a parameter
-        # which says what to name the null translation.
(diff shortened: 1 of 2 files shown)


noliveleger:
@jnm, I could not directly comment on these lines in `asset.py`
```
L297    def _prioritize_translation(self, content, translation_name, is_new=False):
L298        _translations = content.get('translations')
L299        _translated = content.get('translated', [])
L300        if is_new and (translation_name in _translations):
L301            raise ValueError('cannot add existing translation')
L302        elif (not is_new) and (translation_name not in _translations):
```
I don't remember if content always has a `translations` key but in case it does not, `L300` and `L302` should crash because of None iteration. 
Like `L299`, I would set a default value (empty list or dict) to `content.get('translations')` on like `L298`

jnm:
@noliveleger, will `assert` that the content has `translations` per our discussion.

jnm:
@noliveleger, what do you think about this approach?

PR #1980, opened 2018-08-30
Enhancement: use searchStore as much as possible when displaying asset list

### LINKED ISSUE #1960: Don't fetch all assets if data did not change
## Description

Getting all assets is a network heavy call, as assets JSON can easily weight few MBs. To avoid that we should call all assets hash endpoint first and then decide if we want to call for all of them.

## Related issues

Blocked by #1894

pmusaraj:
## Description

Following #1885 and #1973 (please review this after 1973), this PR does the following: 

- when user returns to the Projects list, it displays the list of assets from the searchStore
- in the background, it checks whether BE asset list hash matches FE asset list hash, and only relaunches a network request (i.e. a new search) if there is a mismatch

This should result in improved performance for users, because now the full projects list will only be requested should there be a hash mismatch.  

Fixes #1960 

## Testing notes

When testing this in the UI, please note the following: 

- after this PR, returning to the Projects list should be faster for users, even when they make changes to a project (you should not see a "loading" placeholder when returning to the Projects screen)
- if a second user shares a project with the first user, you should see this new project in the user's Projects list after returning to it from another screen (say, library, form builder, another project, etc.)

## Dev notes

A few small code cleanups are also includes, PR removes some unused methods in searches.es6 and dataInterface.es6.

### DIFF
--- a/jsapp/js/searches.es6
+++ b/jsapp/js/searches.es6
@@ -6,9 +6,11 @@
 import _ from 'underscore';
 import Reflux from 'reflux';
 import $ from 'jquery';
+import SparkMD5 from 'spark-md5';
 
 import stores from './stores';
 import actions from './actions';
+import {dataInterface} from './dataInterface';
 import {assign} from './utils';
 import assetParserUtils from './assetParserUtils';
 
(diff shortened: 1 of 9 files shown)


jsapp/js/searches.es6
```diff
@@ -377,6 +391,28 @@ function SearchContext(opts={}) {
...
+      } else {
+        searchStore.update({defaultQueryState: 'done'});
+
+        dataInterface.assetsHash()
+          .done((data) => {
+            if (data.hash && data.hash === assetsHash(searchStore.state.defaultQueryResultsList)) {
+              // if hashes match, do nothing (current searchStore contents is accurate)
+            } else {
+              this.searchDefault();
+            }
```
magicznyleszek: maybe just do negative if without else and reverse the comment meaning?

pmusaraj:
@magicznyleszek your comment's been addressed, and this is deployed on `unu` if you want to do a test run.

magicznyleszek:
@penar forgot to comment: feel free to merge it :)

PR #1991, opened 2018-09-12
Translations further fixes

### LINKED ISSUE #1988: Feature: Minor enhancements to translations UI
## Description

Following the translations PR #1913, a few improvements: 

- [x] When a form has an unnamed language, when opening the form builder, we currently show an error message: "There is an unnamed translation in your form definition. Please give a name to all translations in your form." It would be nice to point the user in the direction of where they can fix this issue (i.e. the translations/languages modal.) _note: current message is also used in tests._
- [x] When changing the default language, we should warn the user about the action (even though, it is easily reversible). Message should be: "Are you sure you would like to set ##LANG## as the default language for this form?"
- [ ] When hovering over any of the buttons inside the modal, the modal jumps down by about a pixel, and reverts when removing the mouse from the hover position
- [ ] The 'update translation' button is the same as for KC (legacy) in the left sidebar, which could be confusing.

magicznyleszek:
## Description

Fixing minor UI problems

## Related issues

Fixes #1988

### DIFF
--- a/jsapp/xlform/src/model.inputParser.coffee
+++ b/jsapp/xlform/src/model.inputParser.coffee
@@ -100,7 +100,11 @@ module.exports = do ->
     translations = o.translations
     if translations
       if translations.length > 1 and translations.indexOf(null) isnt -1
-        throw new Error("There is an unnamed translation in your form definition. Please give a name to all translations in your form.")
+        throw new Error("""
+          There is an unnamed translation in your form definition.
+          Please give a name to all translations in your form.
+          Use "Manage Translations" modal accessible from form landing page.
+        """)
 
       ###
       TRANSLATIONS HACK (Part 1/2):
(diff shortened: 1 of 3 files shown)


jsapp/xlform/src/model.inputParser.coffee
```diff
@@ -100,7 +100,11 @@ module.exports = do ->
     translations = o.translations
     if translations
       if translations.length > 1 and translations.indexOf(null) isnt -1
-        throw new Error("There is an unnamed translation in your form definition. Please give a name to all translations in your form.")
+        throw new Error("""
+          There is an unnamed translation in your form definition.
+          Please give a name to all translations in your form.
+          Use "Manage Translations" modal accessible from form landing page.
```
pmusaraj: Regular users won't know what a "modal" is. The term is too technical. 

But, anyhow, in the future, there should be a link displayed that loads the Translations modal window. Or a button in the header of the form builder.

magicznyleszek: Changed it on the main branch to `Use "Manage Translations" option from form landing page.`

PR #1999, opened 2018-09-21
Fix submission displaying of grouped repeatable groups

### LINKED ISSUE #1994: Repeated groups inside of groups not being displayed in submission detail view page
## Description

If we create a repeated group inside of another group then while viewing the data, the data under repeated group is not being displayed. However it is displayed it is not nested under any group.

## Steps to Reproduce

1. Create a form with which has some questions in a group
2. Inside that group create another group of questions and mark "Repeat this group if necessary" in the group setting
3. Add some data
4. View the submission in detail popup page

## Expected behavior 
screen shot 2018-09-18 at 11 41 06 am

The group Degree is repeated and it is showing as expected, this is the case if it is not under any group.

## Actual behavior 
screen shot 2018-09-18 at 11 43 12 am

Here the group Movie Name is repeated group which is inside of two groups A and AB, here the data is not displayed.

## Additional details
This is similar issue to this #1931 which was fixed but it did not look into repeated groups.

magicznyleszek:
## Description

Tested on this form: [aSt6SL2KYGD5PRG5XMBFaD.xls.zip](https://github.com/kobotoolbox/kpi/files/2404525/aSt6SL2KYGD5PRG5XMBFaD.xls.zip)

This is a second thing I'm fixing in submission displaying and I am starting to think that we need to redo this thing a bit, because a) there are probably more bugs here (like repeatable group inside repeatable group and possibly other crazy combinations) b) code is hard to grasp, and c) it could look better, e.g. groups endings are not obvious:

screen shot 2018-09-21 at 10 32 32

## Related issues

Fixes #1994


magicznyleszek:
@pmusaraj fixed - it was a leading slash issue in `getGroupedName`

PR #2008, opened 2018-09-27
Overloads logging to avoid use logging.getLogger

noliveleger:
Before messing, REST-UI with this solution. 
I've created a separated branch to see what are the changes for logging


jnm:
This is nice. I do still wonder if [the way I've used Raven/Sentry](https://github.com/kobotoolbox/kpi/blob/7c99d63472073a765e41a47d0347b05765df9755/kobo/settings.py#L540) is wrong and has led to this being necessary in the first place, but this PR will make it easy to change if we ever find such a mistake.

PR #2007, opened 2018-09-27
Warn before closing Translations table modal

magicznyleszek:
## Description

Works on all: Escape key, backdrop click and "X" button click. Code allows for adding same functionality in other modals.

## Related issues

Fixes #2006
Fixes #2011

PR #2019, opened 2018-10-02
Fix crash on share button

### LINKED ISSUE #2018: Crash on share button
## Description

There is a bug in the updated react branch (and possibly master): when clicking on the share button from a list (projects or library), there is a white screen:

image

magicznyleszek:
## Description

One liner

## Related issues

Fixes #2018


jnm:
I think this is fixed already by https://github.com/kobotoolbox/kpi/pull/2007/files#diff-090a4263a189d134ef1ea609a77d3549L231

magicznyleszek:
Right :-D

PR #2032, opened 2018-10-17
Update checkboxes and radios

### LINKED ISSUE #1839: Use single bem styled checkbox and radio everywhere
Currently we have no styles for radios (they are implemented on `form-template-frontend` branch) and styles for checkboxes are very aggressive: `[type="checkbox"]` which makes really hard to do custom checkbox if needed. Plus there are some `input + label` rules that shouldn't be as global as they are.

Proposition:
- [x] Wait for https://github.com/kobotoolbox/kpi/pull/1864 to be merged to master or split the changes out of the PR
- [x] Change `bem.FormModal__radio` and `bem.FormModal__checkbox` to something global outside `FormModal` namespace
- [x] Replace all instances of checkboxes and radios with new bem components
- [x] Remove global aggressive styles

It will be easier to add a modifier to the bem component, than overriding the global rules, when we need a custom checkbox/radio.

magicznyleszek:
## Description

Changes here:
- fix bug with translation tab content not displaying while groupby is missing (form without "Select One" question)
- fix bug with REST Form crashing when missing subset fields
- fix bug in "Table display options": if you unselect "Show group names in table headers"
- cleanup radio and checkbox components
- use radio and checkbox globally
- remove aggressive checkbox and radio styles

## Related issues

Fixes #1839

### DIFF
--- a/jsapp/js/components/modalForms/projectSettings.es6
+++ b/jsapp/js/components/modalForms/projectSettings.es6
@@ -8,6 +8,7 @@ import alertify from 'alertifyjs';
 import Select from 'react-select';
 import Dropzone from 'react-dropzone';
 import TextBox from 'js/components/textBox';
+import Checkbox from 'js/components/checkbox';
 import bem from 'js/bem';
 import TextareaAutosize from 'react-autosize-textarea';
 import stores from 'js/stores';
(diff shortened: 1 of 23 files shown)


jsapp/js/components/modalForms/projectSettings.es6
```diff
@@ -711,17 +712,11 @@ class ProjectSettings extends React.Component {
           </bem.FormModal__item>
 
           <bem.FormModal__item m='metadata-share'>
-            <input
-              type='checkbox'
-              id='share-metadata'
+            <Checkbox
               checked={this.state['share-metadata']}
               onChange={this.onShareMetadataChange}
+              label={t('Help KoboToolbox improve this product by sharing the sector and country where this project will be deployed.') + ' ' + t('All the information is submitted anonymously, and will not include the project name or description listed above.')}
```
jnm: Even the old lines were too long to fit in GitHub's code view without scrolling horizontally, but if you're going to keep `') + ' ' + t('`, why not split it across two lines?

PR #2037, opened 2018-10-22
Make export constants consistent with FormPack - fix

### LINKED ISSUE #2030: Make export constants consistent with formpack
Per my review of https://github.com/kobotoolbox/formpack/pull/179, if we understand that:
* `lang = formpack.constants.UNSPECIFIED_TRANSLATION` means export XML names;
* `lang = formpack.constants.UNTRANSLATED` means export the `null` (same as default?) translation;

- [x] then the documentation for `ExportTask` should be corrected:
https://github.com/kobotoolbox/kpi/blob/78e5ab0f238c0175db1565ff23dfc371bec73f3a/kpi/models/import_export_task.py#L345-L347

- [ ] ~~Handling `_default` (as a synonym for `formpack.constants.UNTRANSLATED`) should be moved out of `ExportTask` and into the [view code](https://github.com/kobotoolbox/kpi/blob/78e5ab0f238c0175db1565ff23dfc371bec73f3a/kpi/views.py#L554):
https://github.com/kobotoolbox/kpi/blob/78e5ab0f238c0175db1565ff23dfc371bec73f3a/kpi/models/import_export_task.py#L448-L449~~

- [ ] ~~The [view code](https://github.com/kobotoolbox/kpi/blob/78e5ab0f238c0175db1565ff23dfc371bec73f3a/kpi/views.py#L554) should substitute `UNSPECIFIED_TRANSLATION` for `xml`.~~

`xml` comes from the frontend at https://github.com/kobotoolbox/kpi/blob/f34f568317127c35719c3885abb666589b02affc/jsapp/js/components/formEditors.es6#L235

**Update**: I changed my mind a lot after seeing this in practice at https://github.com/kobotoolbox/kpi/pull/2037. I don't think we should have API clients having to deal with `lang` being `false` or `null`.

noliveleger:
Fixed #2030.

### DIFF
--- a/jsapp/js/components/formEditors.es6
+++ b/jsapp/js/components/formEditors.es6
@@ -159,10 +159,33 @@ export class ProjectDownloads extends React.Component {
     dataInterface.getAssetExports(this.props.asset.uid).done((data)=>{
       if (data.count > 0) {
         data.results.reverse();
-        this.setState({exports: data.results});
+        let results = data.results.map(result => {
+          if (result.data) {
+            let lang = result.data.lang,
+                langDescription = lang;
+
+            // Backend now changes `_default` to `FormPack.constant.UNTRANSLATED` in `ExportTaskViewSet`.
+            // `FormPack.constant.UNTRANSLATED` constant equals `None`.
+            // Old exports can still have `_default` as value for `lang` though.
+            if (lang === '_default' || lang === null) {
+              langDescription = t('Default');
+            }
+            // Because `xml` can be a code for an existing language (https://en.wikipedia.org/wiki/Malaysian_Sign_Language),
+            // `_xml` is now sent to backend instead of `xml`.
+            // Backend also changes `_xml` to `FormPack.constant.UNSPECIFIED_TRANSLATION`
+            // `FormPack.constant.UNSPECIFIED_TRANSLATION` constant equals `False`.
+            else if (lang === '_xml' || lang === false) {
+              langDescription = t('XML values and headers');
+            }
+            result.data['lang_desc'] = langDescription;
+          }
+          return result;
+        });
+
+        this.setState({exports: results});
 
         // Start a polling Interval if there is at least one export is not yet complete
-        data.results.every((item) => {
+        results.every((item) => {
           if(item.status === 'created' || item.status === 'processing'){
             this.pollingInterval = setInterval(this.refreshExport, 4000, item.url);
             return false;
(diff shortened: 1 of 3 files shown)


jsapp/js/components/formEditors.es6
```diff
@@ -232,7 +255,7 @@ export class ProjectDownloads extends React.Component {
                           <label htmlFor='lang'>{t('Value and header format')}</label>
                           <select name='lang' value={this.state.lang}
                               onChange={this.langChange}>
-                            <option value='xml'>{t('XML values and headers')}</option>
+                            <option value='_xml'>{t('XML values and headers')}</option>
```
jnm: Admirable, but this wraps onto a second line:
image

Also, for every language it's the `values and headers`, but we probably wouldn't want to say `English (en) values and headers` either.

kpi/models/import_export_task.py
```diff
@@ -342,9 +343,11 @@ class ExportTask(ImportExportTask):
     `data` attribute to a dictionary with the following keys:
     * `type`: required; `xls` or `csv`
     * `source`: required; URL of a deployed `Asset`
-    * `lang`: optional; `xml` for XML names or the name of the language to be
-              used for labels. Leave unset, or use `_default` or `None`, for
-              labels in the default language
+    * `lang`: optional; It can be:
+        - unset (`None`) or `formpack.constants.UNTRANSLATED` for labels in the default language (default: `None`).
```
jnm: I try to keep lines under 80 characters for readability. It can certainly be a pain sometimes.

Is there a docstring parser you like to use that doesn't handle line breaks within these parameter descriptions well?

jsapp/js/components/formEditors.es6
```diff
@@ -159,10 +159,33 @@ export class ProjectDownloads extends React.Component {
...
+            // Old exports can still have `_default` as value for `lang` though.
+            if (lang === '_default' || lang === null) {
+              langDescription = t('Default');
+            }
+            // Because `xml` can be a code for an existing language (https://en.wikipedia.org/wiki/Malaysian_Sign_Language),
+            // `_xml` is now sent to backend instead of `xml`.
+            // Backend also changes `_xml` to `FormPack.constant.UNSPECIFIED_TRANSLATION`
+            // `FormPack.constant.UNSPECIFIED_TRANSLATION` constant equals `False`.
+            else if (lang === '_xml' || lang === false) {
+              langDescription = t('XML values and headers');
```
jnm: Admirable, but this wraps onto a second line:
image

Also, for every language it's the `values and headers`, but we probably wouldn't want to say `English (en) values and headers` either.

jnm:
The build is failing because the formpack requirement wasn't updated.

Anyway, I decided I was wrong with what I was asking for in #2030. I'm closing this in favor of https://github.com/kobotoolbox/kpi/pull/2038.

jsapp/js/components/formEditors.es6
noliveleger: What's the suggestion? Only `XML` in caps?

PR #2042, opened 2018-10-23
Allows webpack to work with `kobo-install` when it doesn't run from the host OS.

noliveleger:
With `kobo-install` ip address is not used anymore. It only uses a fake domain name which is added to `/etc/hosts`. 
I'm running docker in `linux` VM and using the browser from `macOS`, so `webpack` sees `localhost` as the `host` OS, not the virtual machine. 
This PR allows `webpack` to bind all ips on port 3000 (not only `localhost`) and points the its web server to good IP thanks to the fake domain name.

@magicznyleszek :
it should work in `kobo-docker` from `master` or `kobo-install` branches. Do you see any potential issues to do this? Can you test it?

@jnm:
I added you too as a reviewer because of a small python change in configuration where RAVEN is not enabled.

### DIFF
--- a/kobo/settings.py
+++ b/kobo/settings.py
@@ -534,7 +534,12 @@ LOGGING = {
             'handlers': ['console'],
             'level': 'DEBUG',
             'propagate': True
-        }
+        },
+        'django.db.backends': {
+            'level': 'ERROR',
+            'handlers': ['console'],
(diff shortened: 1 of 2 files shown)


noliveleger:
@magicznyleszek, thanks :-)

Issue #916, opened 2016-09-12
Implement Password Strength criteria


emilykferguson:
Require users to create/update their passwords to fit our password strength criteria.

magicznyleszek:
My two cents: as for password strengths I think we should follow the work Dropbox team did in 2012: https://blogs.dropbox.com/tech/2012/04/zxcvbn-realistic-password-strength-estimation/

jnm:
Looks like there's also a Python implementation:
https://github.com/dwolfhub/zxcvbn-python

Registration is handled by (an [old version](https://github.com/kobotoolbox/kpi/blob/e0842b5ba7bdcdcf29627f489fecd0ba74158a74/dependencies/pip/requirements.txt#L41) of) https://github.com/macropin/django-registration. @magicznyleszek, have a look at a frontend-only implementation and we'll see if the backend can follow.

magicznyleszek:
@jnm do we want to do anything else besides #2043 here? I don't remember why, but this PR wasn't set to close this… maybe you do :)

jnm:
Yeah, thanks for mentioning this. We need to have the backend require a minimum level of password strength, if the sysadmin decides that's appropriate.

PR #2043, opened 2018-10-29
Password strength in registration form

magicznyleszek:
## Description

Adds visual password strength helper:

screenshot 2018-10-29 at 13 24 31

It uses the awesome https://github.com/dropbox/zxcvbn library. Mind you, it only gives a clue to the user, it doesn't block bad passwords from being used

## Related issues

Part of #916


magicznyleszek:
@jnm I've checked and unfortunately it doesn't :( There was an attempt to enable overriding messages but PR was closed https://github.com/dropbox/zxcvbn/pull/124 I ~~could find~~ can take [the messages](https://github.com/dropbox/zxcvbn/blob/master/src/feedback.coffee) and display them with `t()` function? But then we would need to put them somewhere so it's clear they should be translated

PR #2046, opened 2018-10-30
New management commands to shrink big tables

noliveleger:
Based on `reversion`.`deleterevisions` management commands. I've added two commands:

- `delete_asset_snapshots`
- `delete_import_tasks`

4 options can customized: 
- `days`: Records older than `days` will be deleted
- `chunks`: Number of records per deletion.  
- `vacuum`: Vacuum after deletion
- `vacuum-full`: Full Vacuum after deletion

### DIFF
--- a/kpi/management/commands/delete_assets_snapshots.py
+++ b/kpi/management/commands/delete_assets_snapshots.py
@@ -6,11 +6,11 @@ from datetime import timedelta
 from django.db.models import Max
 from django.utils import timezone
 
-from .delete_base_command import BaseCommand
+from .delete_base_command import DeleteBaseCommand
 from kpi.models import AssetSnapshot
 
 
-class Command(BaseCommand):
+class Command(DeleteBaseCommand):
 
     help = "Deletes assets snapshots"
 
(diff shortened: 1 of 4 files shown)


kpi/management/commands/delete_assets_snapshots.py
```diff
@@ -0,0 +1,33 @@
...
+
+class Command(BaseCommand):
+
+    help = "Deletes assets snapshots"
+
+    def _prepare_delete_queryset(self, **options):
+        days = options["days"]
+        self._model = AssetSnapshot
+
+        # Retrieve Snapshots linked to assets' latest versions.
```
jnm: (Sorry, I can't figure out how to highlight multiple lines for a comment, so I'm just highlighting the first one. This applies to lines 21 through 28)

Nice, but you can get the database to do all the work for you if you want:
```python
latest_version_ids = list(
    AssetSnapshot.objects.exclude(asset_version=None)
    .values('asset_id')
    .annotate(latest_version_id=Max('asset_version_id'))
    .values_list('latest_version_id', flat=True)
    .distinct()
)
```

kpi/management/commands/delete_base_command.py
```diff
@@ -0,0 +1,99 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+import sys
+
+from django.core.management.base import BaseCommand as DjangoBaseCommand
+from django.db import transaction, connection
+
+
+class BaseCommand(DjangoBaseCommand):
```
jnm: :heart: the reusable code, but :broken_heart: using the name of a Django built-in :grin:

Issue #2051, opened 2018-11-05
Add support in the form builder for `range` question type 


jnm:
See http://xlsform.org/en/#range. This one may be trickier than #2050 due to the extra parameters, which should appear as soon as the question is added (and not require clicking on the settings gear).

Per our discussion (doesn't that sound offical): let's use semicolons to separate items in the `parameters` column.

For reference: https://github.com/kobotoolbox/tasks/issues/276

magicznyleszek:
@tinok `range` question type has three important parameters: `start`, `end` and `step` - I wonder how should I display them:
1. Always visible, below question name
2. Openable in similar vain to `select_one`
3. In settings with the rest
Note that `range` has default values (from 0 to 10, step by 1)

I'm leaning towards 2, as these are too important for settings, but we don't have anything similar to 1 in the UI yet.

PR #2076, opened 2018-11-14
Enable file and range question types (simple)

### LINKED ISSUE #2050: Add support in the form builder for `file` question type
Hopefully it will be similar to `image`, `audio`, and `video`. We _could_ (not mandatory if it's a pain) add a UI (in the settings area) for `body::accept` as described in https://github.com/XLSForm/pyxform/issues/194#issuecomment-384339583. Sample XLSForm (which should be tested to see if it actually works :grin:)

type | name | label | body::accept
-- | -- | -- | --
file | uploaded_doc | Upload your dissertation | .pdf,.doc,.odt,.docx

Also, you get to design an icon for this :)

For reference: https://github.com/kobotoolbox/tasks/issues/276

magicznyleszek:
## Description

Enable `file` and `range` question types in configuration for form builder. `file` is almost there, as you can add and edit this question type, separate PR will add `body::accept` in `settings`. As for `range` - it is possible to edit existing one (`parameters` text field in `settings`), but until full support is added in separate PR, I disabled possibility of adding this type.

## Related issues

Part of #2051
Part of #2050

### DIFF
--- a/jsapp/xlform/src/model.configs.coffee
+++ b/jsapp/xlform/src/model.configs.coffee
@@ -155,7 +155,7 @@ module.exports = do ->
         value: "Enter a date"
     range:
       label:
-        value: "Choose a range"
+        value: "Enter a number within a specified range"
     calculate:
       calculation:
         value: ""
--- a/jsapp/xlform/src/model.configs.coffee
+++ b/jsapp/xlform/src/model.configs.coffee
@@ -132,6 +132,9 @@ module.exports = do ->
     audio:
       label:
         value: "Use the camera's microphone to record a sound"
+    file:
+      label:
+        value: "Upload a file"
     note:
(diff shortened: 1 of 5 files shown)


jsapp/xlform/src/model.configs.coffee
```diff
@@ -150,6 +153,9 @@ module.exports = do ->
     date:
       label:
         value: "Enter a date"
+    range:
+      label:
+        value: "Choose a range"
```
jnm: Let's call this "Enter a number within a specified range".
TODO: [figure out](https://www.flowdock.com/app/kobotoolbox/kobo/threads/IroY1P_E_Bfz0my8UzU9qOY_27L) where this actually appears ;)

PR #2077, opened 2018-11-14
Avoid imports error when default_language is present but no translations

### LINKED ISSUE #2073: `ValueError: translation cannot be found` when importing XLSForm
From https://www.flowdock.com/app/kobotoolbox/support/threads/yuUp1PuC4QFvtJh1U9wyhJZRVR5:
> in the past 12hrs we've had 6-7 users have asked for help through intercom about receiving this error when trying to upload an xlsform -
image
> [ACTIVITYATTENDANCESHEET.xls.zip](https://github.com/kobotoolbox/kpi/files/2577214/ACTIVITYATTENDANCESHEET.xls.zip)
> one user said they figured it out -
image
> but I looked at another users xlsform and they did not have a translation column in their form

noliveleger:
fixes #2073

### DIFF
--- a/kpi/models/asset.py
+++ b/kpi/models/asset.py
@@ -302,7 +302,20 @@ class FormpackXLSFormUtils(object):
         if is_new and (translation_name in _translations):
             raise ValueError('cannot add existing translation')
         elif (not is_new) and (translation_name not in _translations):
-            raise ValueError('translation cannot be found')
+            # if there are no translations available, don't try to prioritize,
+            # just ignore the translation `translation_name`
+            if len(_translations) == 1 and _translations[0] is None:
+                return
+            else:  # Otherwise raise an error.
+                # Remove None from translations we want to display to users
+                valid_translations = [t for t in _translations if t is not None]
+                raise ValueError("Translation `{translation_name}` could not be found in your form. "
+                                 "We found the translations: `{translations}`. " \
+                                 "Please select one of these with the exact same spelling.".format(
+                    translation_name=translation_name,
+                    translations="`, `".join(valid_translations)
+                ))
+
         _tindex = -1 if is_new else _translations.index(translation_name)
         if is_new or (_tindex > 0):
             for sheet_name in 'survey', 'choices':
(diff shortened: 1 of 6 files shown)


kpi/models/asset.py
```diff
@@ -302,7 +302,20 @@ def _prioritize_translation(self, content, translation_name, is_new=False):
...
         elif (not is_new) and (translation_name not in _translations):
-            raise ValueError('translation cannot be found')
+            # if there are no translations available, don't try to prioritize,
+            # just ignore the translation `translation_name`
+            if len(_translations) == 1 and _translations[0] is None:
+                return
+            else:  # Otherwise raise an error.
+                # Remove None from translations we want to display to users
+                valid_translations = [t for t in _translations if t is not None]
+                raise ValueError("Translation `{translation_name}` could not be found in your form. "
```
jnm: I think we should tell the user specifically that their default language setting is the problem. Can we move this logic to `_make_default_translation_first()`? I realize `_translations = content['translations']` or similar will have to be duplicated. Then, we could make the error message something like:
```
`{translation_name}` is specified as the default language, but only these translations are present in the form: `{translations}`.
```
I like the "Please select..." language, but let's keep the message concise at the expense of courtesy :)

kpi/models/asset.py
```diff
@@ -302,7 +302,20 @@ def _prioritize_translation(self, content, translation_name, is_new=False):
         if is_new and (translation_name in _translations):
             raise ValueError('cannot add existing translation')
         elif (not is_new) and (translation_name not in _translations):
-            raise ValueError('translation cannot be found')
+            # if there are no translations available, don't try to prioritize,
+            # just ignore the translation `translation_name`
+            if len(_translations) == 1 and _translations[0] is None:
+                return
```
jnm: :+1:

kpi/models/asset.py
noliveleger: Done!

kpi/utils/log.py
```diff
@@ -1,10 +1,13 @@
...
+
 def get_default_logger():
     return stdlib_logging.getLogger("console_logger")
 
+
 # logging does not print to stdout when RAVEN/sentry is activated.
 # To avoid to call logging.getLogger everywhere, we overload logging
 logging = get_default_logger()
 
+
```
jnm: PEP-8 really requires `\n\n\n` at the end of a file? I guess I don't like the BDFL's taste in whitespace :astonished:
(just commentary, not a request for any change to the code)

noliveleger: No, I made the mistake. 
There are too many. There should be only 2.

PR #2082, opened 2018-11-16
Update favicon(s)

### LINKED ISSUE #2081: Update favicon
kpi still uses the old favicon before the logo change in 2016. It should use 
[New KoBo Logo.zip](https://github.com/kobotoolbox/kpi/files/2587523/New.KoBo.Logo.zip)

magicznyleszek:
## Description

I updated the favicon and added some other popular ones (we only had `favicon.ico`) from generator

## Related issues

Fixes #2081


static/safari-pinned-tab.svg
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" standalone="no"?>
```
jnm: This has some unexpectedly sharp edges:
image
Here are color and B&W SVGs that I cut from the larger KoBoToolbox logo:
[kobo-svg.zip](https://github.com/kobotoolbox/kpi/files/2659518/kobo-svg.zip)
Also, I'm not a Safari user, but should the interior negative space be transparent? I'd think it should remain white (and that's what I did in the attached SVGs).

magicznyleszek: I'm also not a Safari user 😛 According to [docs](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/pinnedTabs/pinnedTabs.html) the icon should be single color, as they recolor all nontransparent layers. Also they require a single path (I made a compound path).

PR #2083, opened 2018-11-16
Check if anonymousUser has permissions for collection icon

### LINKED ISSUE #2074: Public collection icon does not always reflect permissions
## Description

Sometimes a public collection—public, in the sense that `AnonymousUser` has view access—does not show the globe badge on its folder icon:
image
I think the opposite can happen as well, but this is not as concerning.

## Steps to Reproduce

1. Question Library, New, Collection
2. Click three dots next to collection name
3. Click "Make Public"
4. Notice the appearance of the globe badge:
    screen shot 2018-11-13 at 18 52 05
5. (nerd alert) `PATCH` the collection with `{"discoverable_when_public": false}`
6. Notice the disappearance of the globe badge:
screen shot 2018-11-13 at 18 55 59

## Alternative Steps to Reproduce

1. Question Library, New, Collection
2. Click three dots next to collection name
3. Click Share
4. For "Enter a username", input `AnonymousUser`
5. Select `View` permission
6. Notice the globe badge does not appear, even after reload.

## Expected behavior 

If a collection is available for access by another user (whether "discoverable" or not), the interface must inform the user somehow.

## Actual behavior 

The globe badge is shown if `discoverable_when_public` is `true`, regardless of permissions granted to `AnonymousUser`

## Why does this matter?

"Make Private" in the UI makes two separate requests to the backend:
image
If the `PATCH` succeeds but the `DELETE` fails, then anyone who has subscribed to this collection could still access it, even as the UI shows the owner that it's no longer public.

:eyes:
https://github.com/kobotoolbox/kpi/blob/4eef7047227222a89ee6f84c8a21975a37d825b1/jsapp/js/components/librarySidebar.es6#L305-L306

magicznyleszek:
## Description

See issue description. Fix turned out to be quite simple. For better results in the future, we would need a new BE endpoint, to not make two separate calls

## Related issues

Fixes #2074


jsapp/js/components/librarySidebar.es6
```diff
@@ -210,6 +210,15 @@ class LibrarySidebar extends Reflux.Component {
       assetid: collectionUid
     });
   }
+  isCollectionPublic(collection) {
+    let isPublic = false;
+    collection.permissions.forEach((perm) => {
```
jnm: Any reason not to use [`getAnonymousUserPermission()`](https://github.com/kobotoolbox/kpi/blob/7255202c0729a884a3cecb3ea4378fd1c448029c/jsapp/js/utils.es6#L35-L42)? There's an example of it being called in this file, [a few lines below](https://github.com/kobotoolbox/kpi/blob/7255202c0729a884a3cecb3ea4378fd1c448029c/jsapp/js/components/librarySidebar.es6#L225).

magicznyleszek: The reason is me not knowing what this function does 👎 Fixed

jnm: 👎🔄👍

jnm:
How I tested this (because it's silly):

1. Made a collection;
1. Tried "Make Public" and saw the icon change correctly;
1. Tried "Make Private" and saw the icon change correctly;
1. Went into sharing settings for the collection and gave view access to AnonymousUser;
    image
1. Noticed no change in the permissions modal (but no error, either);
1. Closed the modal and reloaded the page;
1. Saw the publicly-shared icon appear next to the collection :heavy_check_mark:

PR #2096, opened 2018-11-28
Allow deleted inherited permissions - fix

### LINKED ISSUE #512: Cannot remove inherited permissions
Either the front end needs to create a `deny` permission, or we need to have the back end translate naive requests and delete or deny as appropriate.

Reproduce this bug:
1. Create or import a collection;
2. Create at least one question or block inside the collection;
3. Share the collection with another user;
4. Attempt to remove that user's access to one of the assets inside the collection.
5. Note that the back end returns a 405 with `{detail: "Cannot delete inherited permissions."}`

noliveleger:
Fixes #512 
I finally decided to keep the endpoint response as it was.  @magicznyleszek, so there's nothing to change in FE.
It only returns effective permissions as suggested by @jnm 

@jnm, sorry this PR also contains several PEP-8 compliance fixes (missing white-spaces).


kpi/serializers.py
```diff
@@ -529,6 +530,28 @@ class AssetSerializer(serializers.HyperlinkedModelSerializer):
     # Only add link instead of hooks list to avoid multiple access to DB.
     hooks_link = serializers.SerializerMethodField()
 
+    def get_permissions(self, obj):
+        queries = []
+        for user_id, perm_id in list(obj.get_effective_perms()):
+            queries.append(Q(permission_id=perm_id, user_id=user_id))
```
jnm: I don't think this is going to be workable in a production environment, sadly :(
Using a test account in my local environment, Django Debug Toolbar shows the number of queries jumping from 31 to 553; the SQL execution time increases from 41 to 366 ms (~9x), and the overall CPU time goes up about 4.7x.

I'm going to look at the permission code overall and see if we can't avoid calling `get_effective_perms()` on every API request.

image

PS: Getting this info while running inside a Docker container requires setting `INTERNAL_IPS` in `settings.py`; in my case, I used `INTERNAL_IPS=['172.17.0.1']`, which corresponds to the IP assigned to the `docker0` interface on my host:
```
linux-host$ ip addr show docker0
6: docker0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default 
    link/ether 02:42:12:30:0f:ac brd ff:ff:ff:ff:ff:ff
    inet 172.17.0.1/16 brd 172.17.255.255 scope global docker0
       valid_lft forever preferred_lft forever
    inet6 fe80::42:12ff:fe30:fac/64 scope link 
       valid_lft forever preferred_lft forever
```

kpi/views.py
```diff
@@ -169,13 +169,6 @@ def perform_create(self, serializer):
             serializer.save()
 
     def perform_destroy(self, instance):
-        # Only directly-applied permissions may be modified; forbid deleting
```
jnm: Some permissions marked as inherited really can't be removed, like the permissions assigned to the owner. I'm going to play around and see how `remove_perm()` (on line 179 below) handles those cases.

jnm:
@noliveleger, I cherry-picked your `Added unittest for effective perms after deleting a child of collection` commit on top of `master`, and it failed (as expected). But then, I cherry-picked `Allow to DELETE method on inherited permissions`, and it passed! See https://github.com/kobotoolbox/kpi/commits/2096-alternative-1.

Then, I went further and added a small change to exclude `deny=True` permissions from what's returned by the nested permission serializer: https://github.com/kobotoolbox/kpi/commits/2096-alternative-2. Would this be enough to fix #512?

kpi/serializers.py
jnm: Update: the new caching improved the situation, but I'm surprised that it's still twice as slow as before. Using `master`, on a local account with 56 assets:
before

Using this branch, with the new caching commits:
after

kpi/utils/cache.py
```diff
@@ -0,0 +1,29 @@
...
+from django_request_cache import get_request_cache
+
+
+def void_cache_for_request(keys):
+    """
+    Decorator that removes keys from to current request cache
+    Useful to void cache if content has to be refreshed within
+    the same request (e.g. on save() or delete())
+
+    :param keys: tuple. substrings to be searched for.
```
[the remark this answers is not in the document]
noliveleger: I used `tuple` because (I think) `.startswith` needs a tuple to do the check on multiple strings at a time. 
You are right, they can be only prefixes of keys.

kpi/models/object_permission.py
```diff
@@ -303,12 +315,80 @@ def _filter_anonymous_perms(self, unfiltered_set):
...
+        This method is cached per request because it can be called several times
+        in a row in the same request.
+
+        :param content_type: ContentType
+        :param permission_codename: str
+        :param startswith: bool. Optional. Default is True
+        :param first: bool. Optional. Default is False
+        :return: list<tuple> | <tuple>
+        """
+        # Everyone with change_ should also get share_
```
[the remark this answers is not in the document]
noliveleger: Not at all. I'll fix that.

kpi/serializers.py
noliveleger: Yep. I know. 
AFAIR  there are lots of repeating queries (because of permissions) that are deeply hidden in Django framework. (retrieving same users several time).
We might be able to optimize it a little bit more but this will be deprecated with `v2`.

jnm:
I think we decided to use https://github.com/kobotoolbox/kpi/commits/2096-alternative-2 but to keep this branch as an example of per-request caching, which we will probably need in the future.

noliveleger:
Closing the PR because https://github.com/kobotoolbox/kpi/commits/2096-alternative-2 
 is now part of PR #2417.

PR #2108, opened 2018-12-05
Display error notification on authentication errors

### LINKED ISSUE #506: Trigger useful error message when user is logged out

magicznyleszek:
## Description

I hooked up to ajaxError callback and display error message every time I get `403` or `401` (I haven't seent `401`s from our BE, but it's better to have it for future).

## Related issues

Fixes #506


jsapp/js/dataInterface.es6
```diff
@@ -25,6 +27,16 @@ var dataInterface;
   })();
   this.rootUrl = rootUrl;
 
+  $(document).ajaxError((event, request, settings) => {
+    if (request.status === 403 || request.status === 401) {
+      let errorMessage = t("It seems you're not logged in anymore. Try reloading the page")
+      if (request.responseJSON && request.responseJSON.detail) {
+        errorMessage = request.responseJSON.detail;
+      }
+      notify(errorMessage, 'error');
```
jnm: If the user is really logged out (see previous comment), can we display this in a modal? With `notify()`, it seems awfully easy for the user disregard the message and continue navigating around, mucking things up (e.g. they can create a new library item and add a bunch of questions; the list views will incorrectly start showing `Let's get started by creating your first...`). At first, I thought a savvy user could open another tab and log in again to salvage their session, but doing that just results in `CSRF Failed: CSRF token missing or incorrect.` As far as I can tell, once the user has logged out, it's impossible to do anything useful in the app without reloading. Maybe you could allow the modal to be closed, so that if someone needed to, they could scroll around the page to take note of their work before refreshing.

magicznyleszek: Check out if this is what you were thinking about :) Maybe the alert modal should have "reload" button?

jnm: Reload button sounds good, but I'd be happy to merge without. What does still need some improvement is the handling of `404`s. Sometimes, when a client is not authenticated, we don't want to return a `403` because we don't want to reveal the existence of a resource—so we return `404` instead. See this screencast: https://www.flowdock.com/rest/files/58517/WEmcyjLFs9Mu34AXCAaSBw/authentication%20errors%20-%20PR%202018.mp4

magicznyleszek: I already do `dataInterface.selfProfile` check for `403`s and `401`s, I've added `404` to the code too :)

PR #2117, opened 2018-12-10
Fallback for unlabelled select_one options for table view

### LINKED ISSUE #2116: Table view crashes for labels with display:none
## Description

I'm not exactly sure why this happens, but I was able to reproduce it.

## Steps to Reproduce

1. Use form from https://www.flowdock.com/app/kobotoolbox/support/threads/BGZQkbTd3i3hqW1z2CTvRVf2G4_
2. Enter a submission with weekdays table filled in
3. Go to table view and experience deadly error

## Expected behavior 

It should definitely not crash.

## Actual behavior 

KPI is 100% unusable!

## Additional details

See investigation at https://www.flowdock.com/app/kobotoolbox/support/threads/BGZQkbTd3i3hqW1z2CTvRVf2G4_

magicznyleszek:
## Description

A simple safecheck. Will display "Unlabelled" in place of missing label.

## Related issues

Fixes #2116

### DIFF
--- a/jsapp/js/components/table.es6
+++ b/jsapp/js/components/table.es6
@@ -442,8 +442,12 @@ export class DataTable extends React.Component {
                   value={filter ? filter.value : ''}>
             <option value=''>Show All</option>
             {choices.filter(c => c.list_name === col.question.select_from_list_name).map((item, n) => {
+              let displayLabel = t('Unlabelled');
+              if (item.label) {
+                displayLabel = item.label[0];
+              }
               return (
-                <option value={item.name} key={n}>{item.label[0]}</option>
+                <option value={item.name} key={n}>{displayLabel}</option>
               );
             })}
           </select>;


jsapp/js/components/table.es6
```diff
@@ -442,8 +442,12 @@ export class DataTable extends React.Component {
                   value={filter ? filter.value : ''}>
             <option value=''>Show All</option>
             {choices.filter(c => c.list_name === col.question.select_from_list_name).map((item, n) => {
+              let displayLabel = t('Unlabelled');
+              if (item.label) {
+                displayLabel = item.label[0];
+              }
               return (
-                <option value={item.name} key={n}>{item.label[0]}</option>
+                <option value={item.name} key={n}>{displayLabel}</option>
```
jnm: Let's show the `item.name` if there's no `item.label`. You can see that behavior already in the form builder, e.g. if you remove an option/choice from one version to the next:
image

magicznyleszek: I also added fallback to `$autoname` :)

PR #2121, opened 2018-12-11
Fix extra whitespace in New Project modal

### LINKED ISSUE #2112: Extra whitespace in "Create project" modal
The viewport needs to be fairly wide to see this:
image

In a narrower window, all is fine:
image

magicznyleszek:
## Description

Simple media query added

## Related issues

Fixes #2112


magicznyleszek:
@jnm I fixed two things: templates list not being 100% width and the funny thing. The funny thing is all radios and checkboxes appearing wrong, because of my webpack config fix from today - I added a commit that brings back `autoprefixer` plugin there (it is in the `postcss.config.js`, and the HACK empty array plugins was disabling it)

PR #2128, opened 2018-12-17
File question type extensions setting

### LINKED ISSUE #2050: Add support in the form builder for `file` question type

magicznyleszek:
## Description

Adds "Accepted Files" setting for `file` question type. Placeholder shows default value and serves as example what to write.

screenshot 2018-12-17 at 16 59 15

## Related issues

Fixes #2050


jsapp/xlform/src/view.acceptedFiles.coffee
```diff
@@ -0,0 +1,30 @@
...
+$baseView = require './view.pluggedIn.backboneView'
+$viewTemplates = require './view.templates'
+
+module.exports = do ->
+  class AcceptedFilesView extends $baseView
+    className: 'accepted-files card__settings__fields__field'
+    events: {
+      'input input': 'onChange'
+    }
+    placeholder: '.pdf,.doc,.odt,.docx'
```
jnm: This is very elegant as a placeholder and an informative example, but after testing, this being the default just isn't so. https://github.com/XLSForm/pyxform/issues/194 specifies `application/*` as the default, so one of us misread something :)

Have any ideas for placeholder text that's better than `Enter a comma-separated list of extensions`?

magicznyleszek: I tried finding the source of this default, but can't find it now 🤣 How about `e.g. ".pdf,.doc,.odt"` as an placeholder?

PR #2132, opened 2018-12-19
Display range parameters outside of settings

### LINKED ISSUE #2120: Display `range` question parameters at top level
- [ ] Return to this design:
  image
- [ ] When creating a new `range` from scratch, show default these default values in accordance with http://xlsform.org/en/#range:
  * start: 0
  * end: 10
  * step: 1

This builds on top of https://github.com/kobotoolbox/kpi/pull/2093. Discussion is at https://www.flowdock.com/app/kobotoolbox/kobo/threads/3E1iusXaaQE96z6y7fsW-memlBA.

magicznyleszek:
## Description

A special case for `range` question type: it's parameters are displayed now outside of settings. Also this introduces placeholders to parameters inputs that displays default values.

## Related issues

Fixes #2120

### DIFF
--- a/jsapp/xlform/src/view.params.templates.coffee
+++ b/jsapp/xlform/src/view.params.templates.coffee
@@ -1,12 +1,17 @@
 module.exports = do ->
-  numberParam = (label, number, defaultValue) ->
+  numberParam = (label, currentValue, defaultValue) ->
     if typeof defaultValue isnt 'undefined'
       defaultValueAttr = "placeholder='#{defaultValue}'"
+
+    if currentValue isnt ''
+      valueAttr = "value='#{currentValue}'"
+    else if typeof defaultValue isnt 'undefined'
+      valueAttr = "value='#{defaultValue}'"
 
     return """
     <label class='text-box'>
       <span class='text-box__label'>#{label}</span>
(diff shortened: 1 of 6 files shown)


jsapp/xlform/src/view.params.templates.coffee
```diff
@@ -1,9 +1,12 @@
 module.exports = do ->
-  numberParam = (label, number) ->
+  numberParam = (label, number, defaultValue) ->
+    if typeof defaultValue isnt 'undefined'
+      defaultValueAttr = "placeholder='#{defaultValue}'"
```
jnm: Could we make the default value actually a value instead of a placeholder? Right now, for example, if I change `start` and `step` but leave `end` at its default of `10`, I end up with `"parameters": "start=-2;step=3"` in the content and `start=-5;step=3` in the XLS. I don't think this is intuitive, and who knows if some future client or XLSForm converter won't follow the spec (or the spec will change :smiling_imp:)

magicznyleszek: I'm using default values when parameters are empty and the placeholder is still there, so it is more clear to user that after removing the input value and saving the form, it will still be `10`. It would probably be easies if you check how it feels live :)

PR #2143, opened 2018-12-27
Hide question asset name editing functionality

### LINKED ISSUE #663: When cloning a question don't prompt for a title
Library Questions don't have titles. So when a user clones an existing question they should not be prompted for a new title (as they are when cloning a block or form).

magicznyleszek:
## Description

Changes:

- for `question` asset type, don't display `name` input (as we automatically use question name)
- display helpful tooltip so users will know why no name field is displayed
- be more explicit when creating new Library item, i.e. the `NEW` menu has now "Block" option (this is not bulletproof, as when creating `question`, you can end up with `block` and vice versa - we can address this in future)
- thanks to being explicit we now always display the asset type in FB header

screenshot 2018-12-27 at 23 26 29

screenshot 2018-12-27 at 23 18 38

## Related issues

Fixes #663

### DIFF
--- a/jsapp/js/app.es6
+++ b/jsapp/js/app.es6
@@ -276,6 +276,8 @@ export var routes = (
 
     <Route path='library' >
       <Route path='new' component={AddToLibrary} />
+      <Route path='new/question' component={AddToLibrary} />
+      <Route path='new/block' component={AddToLibrary} />
       <Route path='new/template' component={AddToLibrary} />
       <Route path='/library/:assetid'>
         {/*<Route name="library-form-download" path="download" handler={FormDownload} />,*/}
--- a/jsapp/js/components/formEditors.es6
+++ b/jsapp/js/components/formEditors.es6
@@ -408,6 +408,14 @@ export class AddToLibrary extends React.Component {
       backRoute: '/library'
     };
 
+    if (this.props.location.pathname === '/library/new/question') {
+      this.state.desiredAssetType = ASSET_TYPES.question.id;
+    }
+
+    if (this.props.location.pathname === '/library/new/block') {
+      this.state.desiredAssetType = ASSET_TYPES.block.id;
+    }
+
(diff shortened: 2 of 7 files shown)


jnm:
1. for `question` asset type, don't display `name` input (as we automatically use question name)
    * Fine; great :+1: 
1. display helpful tooltip so users will know why no name field is displayed
    * Good intent, but as you [observed](https://github.com/kobotoolbox/kpi/issues/663#issuecomment-432239308), library `question`s do have titles. With the code in this PR, I note two problems:
        1. `document.title` for a `question` I create from scratch is always "Untitled", even after naming my question;
        1. If I upload single-question XLSForm that has a `title` in its `settings` worksheet into the library, the XLSForm `title` cannot be changed and is displayed in the list view regardless of how I name my question.
    **We don't want to** go down the road of adding special logic all over the place to deal with titles for `question`s, and I don't particularly like the UI discontinuity of disabling the title editor, even with the tooltip—sorry! Please see my **Proposal** below.
1. be more explicit when creating new Library item, i.e. the `NEW` menu has now "Block" option (this is not bulletproof, as when creating `question`, you can end up with `block` and vice versa - we can address this in future)
    * Why should we restrict people from adding two or more questions and creating a `block`, or from removing all but one and creating a `question`? The backend is [flexible with this](https://github.com/kobotoolbox/kpi/blob/81096021377371959c90437be0c1fdbf2ffc595b/kpi/models/asset.py#L601-L607) already.
1. thanks to being explicit we now always display the asset type in FB header
    * OK, but I think we won't end up needing this due to the fluidity of question/block identity.

Also, this doesn't fix #663, because when I clone a `question`, I'm still prompted for a name.

## Proposal

[redacted]

After thinking through this, I don't understand why #663 was opened in the first place. I'll discuss it with Tino.

magicznyleszek:
@jnm I should close this one amirite?

PR #2147, opened 2019-01-02
Add read_only as visible property

### LINKED ISSUE #1726: It should be possible to mark a question as read_only
Unless I'm overlooking something, it doesn't seem to be possible to create a [read_only](http://xlsform.org/#read-only) question in the formbuilder. I think this would be a useful addition.

magicznyleszek:
## Description

It's in the specs http://xlsform.org/en/#read-only

## Related issues

Fixes #1726

### DIFF
--- a/jsapp/xlform/src/view.rowDetail.coffee
+++ b/jsapp/xlform/src/view.rowDetail.coffee
@@ -380,7 +380,7 @@ module.exports = do ->
   viewRowDetail.DetailViewMixins.read_only =
     html: ->
       @$el.addClass("card__settings__fields--active")
-      viewRowDetail.Templates.checkbox @cid, @model.key, _t("Read Only")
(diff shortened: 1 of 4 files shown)


jsapp/xlform/src/view.rowDetail.coffee
```diff
@@ -377,6 +377,13 @@ module.exports = do ->
     afterRender: ->
       @listenForCheckboxChange()
 
+  viewRowDetail.DetailViewMixins.read_only =
+    html: ->
+      @$el.addClass("card__settings__fields--active")
+      viewRowDetail.Templates.checkbox @cid, @model.key, _t("Read Only")
```
jnm: It looks like the existing convention from neighboring lines (e.g. "Default response", "Mandatory response") is to capitalize the first word only. Please try to follow what the existing code is doing; of course, if there's no logical pattern, then use your best judgement :ok_hand:

PR #2149, opened 2019-01-03
Define shadow UserObjectPermission; fixes #2148

jnm:
Also removes django-guardian dependency (and its pesky warning!)

~~Edit: this doesn't work. Or, more accurately, it works only in threads where the shadow models have not been [defined](https://github.com/kobotoolbox/kpi/blob/1000a9e0f4b5e68c4894585361788969bd3b9e9c/kpi/deployment_backends/kc_access/shadow_models.py#L73) yet.~~ (Fixed!)

PR #2150, opened 2019-01-03
Add a kludgy 2-step user deletion process to the admin

jnm:
Please review https://github.com/kobotoolbox/kpi/pull/2149 first.

This closes kobotoolbox/tasks#32. It works from the Django admin user list only. Trying to delete an individual user from the admin change form will still raise a 500 error.

This isn't great, but let's put our energy toward having separate databases for KoBoCAT and KPI.

### DIFF
--- a/dependencies/pip/dev_requirements.txt
+++ b/dependencies/pip/dev_requirements.txt
@@ -29,7 +29,6 @@ django-celery-beat==1.1.1
 django-constance[database]==2.2.0
 django-debug-toolbar==1.6
 django-extensions==1.7.6
-django-guardian==1.4.1
 django-haystack==2.6.0
 django-jsonbfield==0.1.0
 django-loginas==0.2.3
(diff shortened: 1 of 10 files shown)


noliveleger:
@jnm. I know it's a `kludgy 2-step`. 
I'm wondering if we could (at least) deactivate `log in as user` after using step1. Or just adding a column which display users' current state of deletion. (e.g. Step 1 deletion - done).

jnm:
I think I understand why it's a bad workflow, but I'm not sure how deactivating `log in as user` would help. Could you clarify that, unless you think what I'm about to write would work well enough? :)

What if, after `Remove related objects...`, we redirected back to a filtered view of users that displayed only the ones whose related objects had just been removed? For example, if we checked off users whose PKs were 1 through 7, clicked `Removed related objects...`, and confirmed the action, we would then be taken to https://kf.kobotoolbox.org/admin/auth/user/?pk__in=1,2,3,4,5,6,7, which would display only those users. Maybe there's even a way we could have the checkboxes be automatically ticked.

PR #2163, opened 2019-01-17
Fix validation status filtering

### LINKED ISSUE #2162: Filtering and sorting by status of validation does not work in Table View
The name of the property and its value sent by the ajax call are not good ones. 
Property sent is `__ValidationStatus`, value is `{"$regex":"...", "options":"i"}}`.

The property should be `_validation_status.uid` and value doesn't need to use a regex anymore, it should the requested value. (i.e. `validation_status_approved` or `validation_status_not_approved` or `validation_status_on_hold`)
(Using nested properties has been fixed by https://github.com/kobotoolbox/kobocat/issues/420 and https://github.com/kobotoolbox/kobocat/issues/422)

For example, filtering by status `Approved`, current request looks like this: 
`http://[kpi-url]/assets/[asset-id]/submissions/?limit=30&start=0&sort={%22_id%22:-1}&query={%22__ValidationStatus%22:{%22$regex%22:%22validation_status_approved%22,%22$options%22:%22i%22}}`

it should be `http://[kpi-url]/assets/[asset-id]/submissions/?limit=30&start=0&sort={%22_id%22:-1}&query={%22_validation_status.uid%22:%22validation_status_approved%22}`

It's the same issue when sorting by validation status. The property should `_validation_status.uid`, not `__ValidationStatus`

magicznyleszek:
## Description

Uses proper query for validation status filtering now.

## Related issues

Fixes #2162

### DIFF
--- a/jsapp/js/components/modalForms/submission.es6
+++ b/jsapp/js/components/modalForms/submission.es6
@@ -15,7 +15,7 @@ import stores from 'js/stores';
 import ui from 'js/ui';
 import icons from '../../../xlform/src/view.icons';
 import {
-  VALIDATION_STATUSES,
+  VALIDATION_STATUSES_LIST,
   MODAL_TYPES
 } from 'js/constants';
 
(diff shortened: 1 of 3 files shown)


magicznyleszek:
@noliveleger fixed sorting :)

PR #2165, opened 2019-01-17
Display translations box for no languages

### LINKED ISSUE #2156: Always show translation box on form landing page
## Description

Because it is hard to find, and it's a new feature. So create an empty (no languages) state for this:

screenshot 2019-01-11 at 20 19 16

magicznyleszek:
## Description

When a project doesn't have any languages defined, we will display this:

screenshot 2019-01-17 at 09 51 59

And otherwise same as before:

screenshot 2019-01-17 at 09 52 11

## Related issues

Fixes #2156

### DIFF
--- a/jsapp/js/components/formLanding.es6
+++ b/jsapp/js/components/formLanding.es6
@@ -127,6 +127,9 @@ export class FormLanding extends React.Component {
   isFormRedeploymentNeeded() {
     return !this.isCurrentVersionDeployed() && this.userCan('change_asset', this.state);
   }
+  hasLanguagesDefined(translations) {
+    return translations && (translations.length > 1 || translations[0] !== null);
+  }
   showLanguagesModal (evt) {
     evt.preventDefault();
     stores.pageState.showModal({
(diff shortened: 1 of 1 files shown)


jsapp/js/components/formLanding.es6
```diff
@@ -423,23 +423,26 @@ export class FormLanding extends React.Component {
...
-            {translations.map((langString, n)=>{
-              return (
-                <li key={n}>
-                  {langString || t('Unnamed language')}
-                </li>
-              );
-            })}
-          </ul>
+          {!translations || translations.length < 2 &&
+            t('This project has no languages defined yet')
```
Eric-Dilcher: Am I correct in assuming that `t(...)` is somehow translating the text?

magicznyleszek: Yes 👍

jnm: I'd expect to see `This project has no languages defined yet` if the asset content contains this:
```json
"translations": [
    null
],
```
...and that works. :heavy_check_mark:
**However**, I don't think that this should show the same thing:
```json
"translations": [
    "English (en)"
],
```
It's jarring to see `This project has no languages defined yet` if I've just gone into "Manage Languages" and named my language.

magicznyleszek: @jnm good catch - fixed :)

PR #2166, opened 2019-01-17
Update to Django 1.8.19 and sync all requirements files according to production requirements

noliveleger:


noliveleger:
An error occurs when the container is launched for the first time after a fresh build.
```
DisallowedHost: Invalid HTTP_HOST header: 'kpi:8000'. You may need to add u'kpi' to ALLOWED_HOSTS.
```
Something changes between `1.8.15` and `1.8.16` as it appears only from versions `1.8.16`.
After making tests, it doesn't to be a problem  and may appear only when `runserver_plus` is used instead of `uWSGi`

PR #2167, opened 2019-01-17
Fix disappearing columns for no data

### LINKED ISSUE #2153: Clicking on table header to sort hides entire column if first 30 records have no data
## Steps to Reproduce

1. In [this project](https://kf.kobotoolbox.org/#/forms/aquYpKHzQfFdifEbYkAaQV/data/table), view html table
2. Click on Video URL column

## Expected behavior 

1. Table refreshes to sort by that column
2. Column still displays

## Actual behavior 

Table may sort by that column but column disappears until table is sorted by another one. 

## Additional details

I believe this is because columns are hidden if the current view has no data for it. The obvious downside is that for large projects, sorting by a column means showing the rows first where the field was blank, meaning that the column is hidden. The second downside is that this also hides other columns that happen to be blank in this view.

**As a user I'd like to always see all columns, not just the one where the current rows have data for them. **

magicznyleszek:
## Description

Ensures survey rows (for table columns) are always included - even if **all** current submissions have no values.

## Related issues

Fixes #2153

### DIFF
--- a/jsapp/js/components/table.es6
+++ b/jsapp/js/components/table.es6
@@ -150,15 +150,36 @@ export class DataTable extends React.Component {
       console.error(error);
     });
   }
-  _prepColumns(data) {
-    var excludes = ['_xform_id_string', '_attachments', '_notes', '_bamboo_dataset_id', '_status',
-                    'formhub/uuid', '_tags', '_geolocation', '_submitted_by', 'meta/instanceID', 'meta/deprecatedID', '_validation_status'];
 
-    var uniqueKeys = Object.keys(data.reduce(function(result, obj) {
+  _prepColumns(data) {
+    const excludedKeys = [
+      '_xform_id_string',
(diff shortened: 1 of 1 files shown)


jsapp/js/components/table.es6
```diff
@@ -150,15 +150,36 @@ export class DataTable extends React.Component {
...
 
-    uniqueKeys = uniqueKeys.filter((el, ind, arr) => excludes.includes(el) === false);
+    const surveyKeys = [];
+    this.props.asset.content.survey.forEach((row) => {
+      surveyKeys.push(row.$autoname);
+    });
+
+    // make sure the survey columns are displayed, even if current data's
+    // submissions doesn't have them
+    let uniqueKeys = [...new Set([...dataKeys, ...surveyKeys])];
```
jnm: :rainbow:

PR #2177, opened 2019-02-01
Use minutes as minutes not months as minutes

### LINKED ISSUE #2176: Oops, this `__unicode__()` method returns months instead of minutes
It's `year-month-day hour-month`:
https://github.com/kobotoolbox/kpi/blob/cfa495e2d23175818ae94a24aefa4ba43eb73e10/kpi/models/asset_version.py#L66

...but it should be `year-month-day hour-minute` (with zero padding, so that the year has four digits and the other values always have two digits). See http://strftime.org/.

I'm not sure that the output of `AssetVersion.__unicode__()` ever gets shown to the user, but it makes for frustrating debugging.

magicznyleszek:
## Description

Uses minutes with zero padding now :) #fullstack

## Related issues

Fixes #2176

### DIFF
--- a/kpi/models/asset_version.py
+++ b/kpi/models/asset_version.py
@@ -63,5 +63,5 @@ class AssetVersion(models.Model):
 
     def __unicode__(self):
         return '{}@{} T{}{}'.format(self.asset.uid, self.uid,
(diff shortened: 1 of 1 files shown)


jnm:
image

Issue #2179, opened 2019-02-04
The KC settings iframe has too much vertical whitespace


jnm:
This is in Settings -> General. The whitespace effectively hides the ARCHIVE PROJECT and DELETE PROJECT AND DATA buttons.

See https://github.com/kobotoolbox/kobocat-template/pull/92#pullrequestreview-197882670.

PR #2191, opened 2019-02-09
Add and configure iframe resizer

magicznyleszek:
## Description

KPI part of the fix.

## Related issues

Part of #2179

### DIFF
--- a/jsapp/js/components/modalForms/projectSettings.es6
+++ b/jsapp/js/components/modalForms/projectSettings.es6
@@ -751,6 +751,10 @@ class ProjectSettings extends React.Component {
     const sectors = session.currentAccount.available_sectors;
     const countries = session.currentAccount.available_countries;
 
+    if (this.props.context === PROJECT_SETTINGS_CONTEXTS.EXISTING && this.props.iframeUrl) {
+      $('iframe#kc-settings-iframe').iFrameResize({log: true});
+    }
+
     return (
       <bem.FormModal__form
         onSubmit={this.handleSubmit}
@@ -869,7 +873,12 @@ class ProjectSettings extends React.Component {
 
           {this.props.context === PROJECT_SETTINGS_CONTEXTS.EXISTING && this.props.iframeUrl &&
             <bem.FormView__cell m='iframe'>
-              <iframe src={this.props.iframeUrl} />
+              <iframe
+                id='kc-settings-iframe'
+                src={this.props.iframeUrl}
+                scrolling='no'
+                style={{width: '1px', minWidth: '100%'}}
+              />
             </bem.FormView__cell>
           }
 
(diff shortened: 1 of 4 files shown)


jsapp/js/components/modalForms/projectSettings.es6
```diff
@@ -751,6 +751,10 @@ class ProjectSettings extends React.Component {
     const sectors = session.currentAccount.available_sectors;
     const countries = session.currentAccount.available_countries;
 
+    if (this.props.context === PROJECT_SETTINGS_CONTEXTS.EXISTING && this.props.iframeUrl) {
+      $('iframe#kc-settings-iframe').iFrameResize({log: true});
```
Eric-Dilcher: the selector is unnecessarily specific.  Anyways, IDs are unique across applications.

jsapp/js/components/modalForms/projectSettings.es6
```diff
@@ -869,7 +873,12 @@ class ProjectSettings extends React.Component {
 
           {this.props.context === PROJECT_SETTINGS_CONTEXTS.EXISTING && this.props.iframeUrl &&
             <bem.FormView__cell m='iframe'>
-              <iframe src={this.props.iframeUrl} />
+              <iframe
+                id='kc-settings-iframe'
+                src={this.props.iframeUrl}
+                scrolling='no'
+                style={{width: '1px', minWidth: '100%'}}
```
Eric-Dilcher: why do you have this style?  Why isn't it in css?   I'm assuming it's needed for iframe-resizer to work.  Please add a comment that says that

package.json
```diff
@@ -26,6 +26,7 @@
     "eslint-plugin-react": "^7.10.0",
     "file-loader": "^2.0.0",
     "font-awesome": "^4.4.0",
+    "iframe-resizer": "^3.6.5",
```
Eric-Dilcher: looks like v4 was published not that long ago.  You should update to it :)

jsapp/js/components/modalForms/projectSettings.es6
Eric-Dilcher: OK, I see the note in the package.

> Note: Using min-width to set the width of the iFrame, works around an issue in iOS that can prevent the iFrame from sizing correctly.

Can you add this note to the css file when you move the style there?

jnm:
I can't make it work locally :sob: 
cry

Ah yes, if only I had read the console messages:
`Failed to execute 'postMessage' on 'DOMWindow': The target origin provided ('http://172.17.0.1:8001') does not match the recipient window's origin ('http://172.17.0.1:8000').`
image

jsapp/js/components/modalForms/projectSettings.es6
magicznyleszek: fixed :)

jsapp/js/components/modalForms/projectSettings.es6
magicznyleszek: fixed :)

package.json
magicznyleszek: updated here and on the other branch

magicznyleszek:
@jnm any idea how to test it locally? Can we deploy this branch to a test server and the `kobocat-template` branch too?

magicznyleszek:
Close in favor of #2238

PR #2196, opened 2019-02-13
Validation status fixes

### LINKED ISSUE #2188: When updating a validation status, choice is lost but saved in DB
### Steps to reproduce.
Update a validation status in the table view. For example:  `Approved`
After choosing it, It shows `Select...`  instead of `Approved`. 
If you reload the page, the value is the correct one.

### LINKED ISSUE #2189: Cannot remove validation status once set
Once `On Hold`, `Not Approved`, or `Approved` is chosen, it's not possible to go back to an empty validation status. This behavior is the same in the list view and submission modal.

Could you have the frontend allow us to revert to an empty status, pretty please? :kissing_heart: 

:wave: @mdestaubin

magicznyleszek:
## Description

Fixes two issues, but also this:

- when editing Validation Status in Submission Modal, it updates in UI after API call finishes
- when editing Validation Status in Submission Modal, the Table View Validation Status also refreshes

## Related issues

Fixes #2188 
Fixes #2189

# Warning
Do not merge before https://github.com/kobotoolbox/kobocat/pull/525 is reviewed and merged

### DIFF
--- a/jsapp/js/components/modalForms/submission.es6
+++ b/jsapp/js/components/modalForms/submission.es6
@@ -217,9 +217,12 @@ class Submission extends React.Component {
     });
   }
 
-  validationStatusChange(e) {
-    const data = {'validation_status.uid': e.value};
-    actions.resources.updateSubmissionValidationStatus(this.props.asset.uid, this.state.sid, data);
+  validationStatusChange(evt) {
+    if (evt.value === null) {
+      dataInterface.removeSubmissionValidationStatus(this.props.asset.uid, this.state.sid);
+    } else {
+      dataInterface.updateSubmissionValidationStatus(this.props.asset.uid, this.state.sid, {'validation_status.uid': evt.value});
+    }
   }
 
   languageChange(e) {
--- a/jsapp/js/components/modalForms/submission.es6
+++ b/jsapp/js/components/modalForms/submission.es6
@@ -52,14 +52,17 @@ class Submission extends React.Component {
   }
   componentDidMount() {
     this.getSubmission(this.props.asset.uid, this.state.sid);
-    this.listenTo(actions.resources.updateSubmissionValidationStatus.completed, this.refreshSubmission);
+    this.listenTo(actions.resources.updateSubmissionValidationStatus.completed, this.refreshSubmissionValidationStatus);
+    this.listenTo(actions.resources.removeSubmissionValidationStatus.completed, this.refreshSubmissionValidationStatus);
   }
 
-  refreshSubmission(result, sid) {
-    if (result.uid) {
+  refreshSubmissionValidationStatus(result, sid) {
+    if (result && result.uid) {
       this.state.submission._validation_status = result;
-      this.setState({submission: this.state.submission});
+    } else {
+      this.state.submission._validation_status = {};
     }
+    this.setState({submission: this.state.submission});
   }
 
   getSubmission(assetUid, sid) {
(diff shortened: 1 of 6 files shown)


jsapp/js/components/modalForms/submission.es6
```diff
@@ -52,14 +52,17 @@ class Submission extends React.Component {
   }
   componentDidMount() {
     this.getSubmission(this.props.asset.uid, this.state.sid);
-    this.listenTo(actions.resources.updateSubmissionValidationStatus.completed, this.refreshSubmission);
+    this.listenTo(actions.resources.updateSubmissionValidationStatus.completed, this.refreshSubmissionValidationStatus);
```
Eric-Dilcher: If I understand correctly, this component has access to `this.listenTo` (which is a redux thing) because at the bottom we attach a `Reflux.ListenerMixin`

What is the reason we use mixins, and not simply have the component extend `Reflux.Component`?

magicznyleszek: Your guess is correct :) I think mixins are "the old way" and I'm just following what's already in the code.

jsapp/js/components/table.es6
```diff
@@ -133,22 +133,31 @@ export class DataTable extends React.Component {
...
+      dataInterface.removeSubmissionValidationStatus(_this.props.asset.uid, sid).done((result) => {
+        _this.state.tableData[index]._validation_status = {};
         _this.setState({tableData: _this.state.tableData});
-      } else {
-        console.error('error updating validation status');
-      }
-    }).fail((error)=>{
-      console.error(error);
-    });
+      }).fail((error)=>{
```
Eric-Dilcher: since `console.error` is functionally identical to `(error)=>{ console.error(error); }`, you can just do 
`.fail(console.error)`

You can also do that below

jsapp/js/components/table.es6
```diff
@@ -766,54 +776,52 @@ export class DataTable extends React.Component {
...
-      d = {
-        submissions_ids: Object.keys(this.state.selectedRows),
-        'validation_status.uid': val
-      };
+  onBulkUpdateStatus(evt) {
+    const val = evt.target.getAttribute('data-value');
+    const selectAll = this.state.selectAll;
+    const data = {};
+    let selectedCount;
+    let apiFn;
```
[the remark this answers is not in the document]
magicznyleszek: I'm not a big fan of ternary operator, but here it looks kinda better :-*

jsapp/js/components/table.es6
magicznyleszek: nice! fixed

kpi/deployment_backends/mixin.py
```diff
@@ -239,25 +240,29 @@ def validation_status(self, kpi_request, pk, *args, **kwargs):
         http_method_params = {}
         if kpi_request.method == "PATCH":
             http_method_params = {"json": kpi_request.data}
-        else:
+        elif kpi_request.method == "GET":
```
Eric-Dilcher: This changes the semantics of the code. I think you should add a `else` block, and if it should never be run, then either throw an error, or log something.

jsapp/js/components/table.es6
```diff
@@ -766,54 +772,47 @@ export class DataTable extends React.Component {
...
-          'validation_status.uid': val
-        };
+        data.confirm = true;
+        data['validation_status.uid'] = val;
       }
+      selectedCount = this.state.resultsTotal;
+    } else {
+      data.submissions_ids = Object.keys(this.state.selectedRows);
+      data['validation_status.uid'] = val;
+      selectedCount = Object.keys(this.state.selectedRows).length;
```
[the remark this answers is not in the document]
magicznyleszek: fixed :)

kpi/deployment_backends/mixin.py
magicznyleszek: @noliveleger this is your change - what do you think? :)

noliveleger: IMHO, there is no needs to add an `else` because method can be only `PATCH`, `GET` or `DELETE`.
`if elseif` is only there to override `http_method_params`.

@Eric-Dilcher would you prefer something like this?
```
if kpi_request.method == "PATCH":
    http_method_params = {"json": kpi_request.data}	            http_method_params = {"json": kpi_request.data}
elif kpi_request.method == "GET":
    http_method_params = {"params": kpi_request.GET}
else:
    http_method_params = {}
```

jnm: This should be okay as-is because of the Django REST Framework `@detail_route` decorator. The `else` case is a deliberate no-op.

PR #2199, opened 2019-02-18
Make group label an editable input

### LINKED ISSUE #2197: Cannot edit group names in the form builder
1. Create a new form on kf.kobotoolbox.org;
2. Add a question;
3. Select the question;
4. Click "Create group with selected questions";
5. Try to change the name of the group from "Group" to something else;
6. Notice that "Group" cannot be edited.

magicznyleszek:
## Description

This bug was introduced by me: #2151 :ashamed: This should fix it :)

## Related issues

Fixes #2197

### DIFF
--- a/jsapp/xlform/src/view.row.templates.coffee
+++ b/jsapp/xlform/src/view.row.templates.coffee
@@ -80,16 +80,16 @@ module.exports = do ->
       #{expandingSpacerHtml}
       """
 
-  groupView = (g)->
+  groupView = ()->
     """
     <div class="survey__row__item survey__row__item--group group card js-select-row">
       <header class="group__header">
         <i class="group__caret js-toggle-group-expansion fa fa-fw fa-caret-down"></i>
-        <span class="group__label js-cancel-select-row js-cancel-sort">#{g.getValue('label')}</span>
-          <div class="group__header__buttons">
-            <span class="group__header__buttons__button group__header__buttons__button--settings js-toggle-card-settings"><i class="fa fa-cog"></i></span>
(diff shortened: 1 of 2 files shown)


jsapp/xlform/src/view.row.templates.coffee
```diff
@@ -80,16 +80,16 @@ module.exports = do ->
...
     """
     <div class="survey__row__item survey__row__item--group group card js-select-row">
       <header class="group__header">
         <i class="group__caret js-toggle-group-expansion fa fa-fw fa-caret-down"></i>
-        <span class="group__label js-cancel-select-row js-cancel-sort">#{g.getValue('label')}</span>
-          <div class="group__header__buttons">
-            <span class="group__header__buttons__button group__header__buttons__button--settings js-toggle-card-settings"><i class="fa fa-cog"></i></span>
-            <span class="group__header__buttons__button group__header__buttons__button--delete js-delete-group"><i class="fa fa-trash-o"></i></span>
-          </div>
+        <input type="text" placeholder="#{_t("Group title is required")}" class="card__header-title js-cancel-select-row js-cancel-sort">
```
jnm: It seems like these `placeholder`s (this one and the previously added [`Question title is required`](https://github.com/kobotoolbox/kpi/blob/d4ed96bf475440bf7fe37dd6859f0881ffa7e501/jsapp/xlform/src/view.row.templates.coffee#L109)) aren't really used for anything. I think we should either remove them, or, if they are used somewhere that I've missed, then change `title` to `label` to use the correct XLSForm terminology consistently.

jnm: Tracked at https://github.com/kobotoolbox/kpi/issues/2201

PR #2219, opened 2019-03-04
Prevent navigating out of form builder with unsaved changes

### LINKED ISSUE #2200: Pressing "delete" button while adding question in form builder closes the form without saving
## Description

Pressing "delete" button while adding question in form builder closes the form without saving

## Steps to Reproduce

1. Open form builder
2. Click the + button to add a question
3. Without entering any text in the field, click the "+ Add question" button". The list of question types appears
4. Press the computer's "delete" button

## Expected behavior 

Either nothing should happen (the "delete" button shouldn't have any impact) or KoBo should ask to "leave without saving?"

## Actual behavior 

The form editor closes, the user is taken back to the form's main screen. If there was unsaved edits, those edits are lost.

## Additional details

MacOS

magicznyleszek:
## Description

This covers hash navigation - i.e. the `react-router` one. So using browser history navigation will now display prompt for users, so they will not lose unsaved changes.

## Related issues

Fixes #2200

### DIFF
--- a/jsapp/js/editorMixins/editableForm.es6
+++ b/jsapp/js/editorMixins/editableForm.es6
@@ -161,6 +161,8 @@ const ASIDE_CACHE_NAME = 'kpi.editable-form.aside';
 
 export default assign({
   componentDidMount() {
+    this.props.router.setRouteLeaveHook(this.props.route, this.routerWillLeave)
+
     document.body.classList.add('hide-edge');
 
     this.loadAsideSettings();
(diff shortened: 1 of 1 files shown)


jsapp/js/editorMixins/editableForm.es6
```diff
@@ -249,12 +257,14 @@ export default assign({
   }, 200),
 
   preventClosingTab() {
+    this.setState({preventNavigatingOut: true});
     $(window).on('beforeunload.noclosetab', function(){
-      return t('you have unsaved changes');
+      return t('You have unsaved changes. Leave form without saving?');
```
jnm: I agree that this is an improvement, but every time a localized string like this changes, all of the translations have to be re-done (by a human).

PR #2221, opened 2019-03-07
Count next to 'My Library' now shows count of questions, blocks, and …

### LINKED ISSUE #1451: Library only counts collections (low priority)
Given this structure of assets in a user's library:

* Library Root
  * Independent Question
  * Independent Block
  * Collection
    * Question inside a Collection
    * Block inside a Collection

The sidebar count currently reports 1. Cause: only collections are being counted. After discussion with @pmusaraj, I propose we do the opposite and count all questions and blocks (i.e. count all the items that could be added to a form). The desired count for this user's library would then be 4.

Eric-Dilcher:
Fixes #1451

### DIFF
--- a/jsapp/js/components/librarySidebar.es6
+++ b/jsapp/js/components/librarySidebar.es6
@@ -260,7 +260,6 @@ class LibrarySidebar extends Reflux.Component {
     };
   }
   render () {
-    console.log("state", this.state.defaultQueryCount)
     return (
       <bem.CollectionsWrapper>
         <ui.PopoverMenu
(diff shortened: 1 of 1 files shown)


jsapp/js/components/librarySidebar.es6
```diff
@@ -255,6 +260,7 @@ class LibrarySidebar extends Reflux.Component {
     };
   }
   render () {
+    console.log("state", this.state.defaultQueryCount)
```
magicznyleszek: Cleanup this thing and ready to go :shipit:

PR #2223, opened 2019-03-11
New icon and label for NEW > Question

### LINKED ISSUE #2203: Enable the saving in the library of entire groups of questions
## Description

In many cases, the same groups of questions of questions are used across forms. When those groups include dozens of questions, and hundreds of potential answers, saving questions to the library isn't enough. Users should be able to save entire groups, so that they can be added at once into new forms.

magicznyleszek:
## Description

With this button you can end up with Question, but also a Block. I've changed the label and updated the icon:

Screenshot 2019-03-11 at 20 14 56

## Related issues

Fixes #2203


magicznyleszek:
Closing this because of new Public Collections changes coming.

PR #2224, opened 2019-03-13
Bumped redis version to 3.2.0

noliveleger:
`kpi` doesn't have the same issue as `kobocat` (https://github.com/kobotoolbox/kobocat/issues/532), but it's better to keep versions synchronized.

### DIFF
--- a/dependencies/pip/requirements.txt
+++ b/dependencies/pip/requirements.txt
@@ -92,7 +92,7 @@ python-dateutil==2.7.5
 python-digest==1.7
 pytz==2018.9
 pyxform==0.12.0
-redis==3.0.1              # via celery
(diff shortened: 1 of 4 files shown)


dependencies/pip/requirements.txt
```diff
@@ -92,7 +92,7 @@ python-dateutil==2.7.5
 python-digest==1.7
 pytz==2018.9
 pyxform==0.12.0
-redis==3.0.1              # via celery
+redis==3.2.0              # via celery
```
jnm: It seems like we're asking for a newer version than specified by `celery`. That's OK, but if so, we should pin `redis==3.2.0` in `requirements.in`.

PR #2225, opened 2019-03-19
Edit submission wrong url bug fix

### LINKED ISSUE #2217: EDIT button opens wrong submission
1. Create a form with at least one question whose responses will make submissions easy to distinguish;
1. Deploy;
1. Collect 2 or more submissions;
1. Open the table view;
1. View the first submission;
1. Click NEXT to view the second submission;
1. Click PREVIOUS to view first submission again;
1. Click EDIT;
1. Observe that the submission shown in Enketo is second(!) submission.

magicznyleszek:
## Description

This fixes two problems:

1. edit submission url wasn't cleared when switching submissions
2. edit submission url is produced by Enketo and a safe mechanism allows to get only one per minute (to avoid two people editing at the same time), which we failed to take into account in submission modal

I basically use the idea from table view in the submission modal - i.e. don't get edit submission url from Enketo on submission data load, but do it when user clicks "Edit" button.

## Related issues

Fixes #2217


jsapp/js/components/modalForms/submission.es6
```diff
@@ -500,18 +517,25 @@ class Submission extends React.Component {
...
-              <a href={this.state.enketoEditLink}
-                onClick={this.promptRefresh}
-                target='_blank'
-                className='mdl-button mdl-button--raised mdl-button--colored'>
-                {t('Edit')}
+            {this.userCan('change_submissions', this.props.asset) &&
+              <a
+                onClick={this.launchEditSubmission.bind(this)}
+                className='mdl-button mdl-button--raised mdl-button--colored'
+                {...({
```
jnm: :cool:, made me learn something :)
For tidiness, could this and the following two lines be replaced with `{...{disabled: this.isSubmissionEditable() ? null : true}}`?

magicznyleszek: I don't remember why I did it this way honestly :-D Changed to even simpler solution.

magicznyleszek:
@jnm I fixed it by introducing an `enketoHandler` component - to make it DRY :) It stores each url for 30 seconds (avoids complexity of watching the 404s), and opens either stored one or gets it and then opens.

jsapp/js/components/modalForms/submission.es6
jnm: :smile: I think I know why you did it: because HTML wants either `<tag>` or `<tag disabled>`, not something like `<tag disabled="true">`. I think React [handles this well](https://github.com/facebook/react/issues/961), though.

I was still puzzled in a different way because `<a>` is not supposed to recognize `disabled`, but yet it works here (and not in any simple HTML file I created). I see the reason why: https://github.com/kobotoolbox/kpi/blob/9db69cb7bcae52d8cb20092707e0b67df3a8e9cb/jsapp/scss/libs/_mdl.buttons.scss#L153.

jnm:
@magicznyleszek, thank you; I'm happy with this PR. I did make [a few small changes](https://github.com/kobotoolbox/kpi/pull/2225/commits/c49fc24944bf58e710c7bd597133cc5d9710bd0d), and I'd be happy to hear your thoughts on them.

@arsea, what do you think? Want to do the honors and merge this?

magicznyleszek:
@jnm I like your changes :) I wonder how I missed this bug with wrong parameters 😱

Issue #1592, opened 2018-02-27
Can't drag a question or block from the library into a group


tinok:
1. Create a new form. Add two questions and group them
1. Select a question or block from the library sidebar and try to drag it between Q1 and Q2 (inside the group)

Expected: The library question can be placed anywhere inside the group
Instead: It can only be added before or after the group

NB: It's not necessary to add two questions to create a group, and it should be possible (and is, with #2242 merged) to drag a library question into a single-question group. —@jnm

jnm:
Fixed by #2242

PR #2242, opened 2019-04-04
Make it possible to drag from Library into group in Form Builder

magicznyleszek:
## Description

Changes:

- some whitespace stuff for readability
- enabling dragging from Library list into groups
- fixed `inputParse.parse` bugs caused by messing with original object data
- I had a problem with dragging the same item multiple times (consecutive items were placed at bottom of form regardless where I dropped them) - after lots of uneventful debugging, my macOS needed restarting for updates and the bug was gone 🤷‍♂️❓ Tried to reproduce it with no success - my guess is that webpack hotreload did something funky or it stopped working and I didn't notice

## Related issues

#1592

## Fun fact

1592 AD had "The Ultimate Pi Day" on 3-14-1592, it was the largest correspondence between calendar dates and significant digits of pi (`3.141592…`) ever

### DIFF
--- a/jsapp/xlform/src/model.survey.coffee
+++ b/jsapp/xlform/src/model.survey.coffee
@@ -85,8 +85,8 @@ module.exports = do ->
       if !row.rows && rowlist = row.getList()
         @choices.add(name: rowlist.get("name"), options: rowlist.options.toJSON())
 
-    insertSurvey: (survey, index=-1)->
-      index = @rows.length  if index is -1
+    insertSurvey: (survey, index=-1, targetGroupId)->
+      index = @rows.length if index is -1
       for row, row_i in survey.rows.models
         index_incr = index + row_i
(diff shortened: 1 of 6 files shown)


jsapp/xlform/src/model.survey.coffee
```diff
@@ -85,23 +85,46 @@ module.exports = do ->
...
-          # its a group
           name_detail = row.get('name')
           name_detail.set 'value', name_detail.deduplicate(@)
-          @rows.add(row.toJSON(), at: index_incr)
-      ``
+          target.rows.add(
+            row.toJSON(),
+            at: index_incr
+          )
+      return
```
jnm: :ok_hand:

jnm: :ok_hand:

PR #2264, opened 2019-04-28
Password change causes logout

### LINKED ISSUE #2258: Log out user after updating password
## User story

1. As a survey admin, I want to make sure anyone else using my account is logged out as soon as I change my account password in order to protect collected data. 
2. I would also like to sign out all other users even if I don't change my password (similar to Gmail feature) 

## UX flow
1. Log in to User1 on computer 1 and computer 2 (or different browsers)
2. Go to profile, modify password
3. User on computer 2 should be redirected to login page (as soon as next request is made).

Relevant settings file: 
https://github.com/kobotoolbox/kpi/blob/7b5a1e708d5436d6d6348eb9066ae223507fc4ef/kobo/settings/base.py#L112-L114

## Background

If a user changes their password of the account (while being logged in), anyone logged in with the same account continues being able to use the account.

magicznyleszek:
## Description

Updating password will now cause a logout (added a warning note near button)

## Related issues

Fixes #2258

### DIFF
--- a/jsapp/js/actions.es6
+++ b/jsapp/js/actions.es6
@@ -666,6 +666,7 @@ actions.auth.changePassword.listen((currentPassword, newPassword) => {
 });
 actions.auth.changePassword.completed.listen(() => {
   notify(t('changed password successfully'));
+  actions.auth.logout();
 });
 actions.auth.changePassword.failed.listen(() => {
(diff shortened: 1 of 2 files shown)


magicznyleszek:
It turned out, this is not needed. A proper solution for the problem would be done on backend.

PR #2269, opened 2019-04-30
better eslint rules

magicznyleszek:
## Description

Linting code is fun and useful. Things I did:

- changed our tiny `.eslintrc` file to `.eslintrc.json` which extends the `recommended` settings
- downgraded few rules from `error` to `warning`
- fixed errors that weren't linted before


Eric-Dilcher:
@magicznyleszek this looks good 2 go

PR #2292, opened 2019-06-07
Less linter spam

magicznyleszek:
## Description

Disables linter warnings appearing in Dev Tools console :) Currently all of the warnings appear with every livereload change, and there are a lot of them - makes the console not very usable. Plus there is linting in the editor, which is the only place I use linter.


magicznyleszek:
Sorry for merging, but I was about to copy the changes to 3rd dev branch and it felt bad :-P

PR #2303, opened 2019-06-26
Bring back webpack console logging, drop stylelint plugin

magicznyleszek:
Undo silencing console output. Drop `stylelint-webpack-plugin`, as it can't do what I would like it to do. We will not have stylelint output while building, will stick to editor stylelint :+1:


jnm:
Related to #2304

Issue #2294, opened 2019-06-09
Display full project title everywhere


magicznyleszek:
Display full title in all the places where it is possible to do so. Currently there are a few where we still trim them to single line with ellipsis. 

Make sure to check: projects list, app header, project landing page.

arsea:
PR #2325 created

magicznyleszek:
Closed via #2325

PR #2325, opened 2019-07-19
Project title fully displayed in list and sidebar, title attr added

arsea:
## Description

The title should now be in full in the Projects list and the sidebar.
In the Form editor it now shows the tooltip on hover.

Fixes issue #2294


magicznyleszek:
@arsea btw, instead of "fixes issue #…" write "fixes #…", then GH will automatically close the issue :)

Issue #2328, opened 2019-07-29
Use API v2 in frontend calls.


noliveleger:
When `two-databases` branch is merged into `master`, let's try to use v2 endpoints. 

This a list of endpoints of `v2`. `FE` may not use all of them. 
Basically, `v2` adds the prefix `/api/v2` to `v1` urls:
 
This a list of endpoints of `v2`: 
- /api/v2/asset_snapshots/
- /api/v2/asset_snapshots/`<uid>`/
- /api/v2/asset_snapshots/`<uid>`/preview/
- /api/v2/asset_snapshots/`<uid>`/xform/
- /api/v2/assets/
- /api/v2/assets/`<parent_lookup_asset>`/data/
- /api/v2/assets/`<parent_lookup_asset>`/data/`<pk>`/
- /api/v2/assets/`<parent_lookup_asset>`/data/`<pk>`/edit/
- /api/v2/assets/`<parent_lookup_asset>`/data/`<pk>`/validation_status/
- /api/v2/assets/`<parent_lookup_asset>`/data/bulk/
- /api/v2/assets/`<parent_lookup_asset>`/data/validation_statuses/	
- /api/v2/assets/`<parent_lookup_asset>`/files/
- /api/v2/assets/`<parent_lookup_asset>`/files/`<uid>`/
- /api/v2/assets/`<parent_lookup_asset>`/files/`<uid>`/content/	
- /api/v2/assets/`<parent_lookup_asset>`/hook-signal/	
- /api/v2/assets/`<parent_lookup_asset>`/hooks/
- /api/v2/assets/`<parent_lookup_asset>`/hooks/`<parent_lookup_hook>`/logs/	
- /api/v2/assets/`<parent_lookup_asset>`/hooks/`<parent_lookup_hook>`/logs/`<uid>`/
- /api/v2/assets/`<parent_lookup_asset>`/hooks/`<parent_lookup_hook>`/logs/`<uid>`/retry/	
- /api/v2/assets/`<parent_lookup_asset>`/hooks/`<uid>`/	
- /api/v2/assets/`<parent_lookup_asset>`/hooks/`<uid>`/retry/	
- /api/v2/assets/`<parent_lookup_asset>`/permissions/	
- /api/v2/assets/`<parent_lookup_asset>`/permissions/`<uid>`/
- /api/v2/assets/`<parent_lookup_asset>`/permissions/bulk/	
- /api/v2/assets/`<parent_lookup_asset>`/permissions/clone/
[...]

PR #2331, opened 2019-07-29
Use pagination for Data ViewSet and Table View bug fixes on filtering

### LINKED ISSUE #2319: Filtering validation status doesn't refresh pagination info.
## Description

When you have several submissions (more than the number of rows allowed per page when data is paginated), if you want to narrow down results per their validation status, pagination doesn't change and stays the same likewise no filters were applied.

## Steps to Reproduce

1. Create a form
2. Submit data more than 10 times. (Let's say 12)
3. On Table view, display results with 10 rows per page.
4. Go to Page 2, update records to (for example) `Approved` status.
5. Go to Page 1, filter results by choosing `Approved` from the dropdown

## Expected behavior 

On header: Title should be: 1 - 10 of 10 results.
On footer: Page should be: 1 of 1 
Next button is disabled

## Actual behavior 

On header: Title is:  1 - 10 of 12 results.
On footer: Page is: 1 of 2
Next button is enabled

### LINKED ISSUE #2320: Update table view after updating validation statuses if a filter is applied
## Description

Related to #2319, but not a bug. The message I received after having updated filtered data confused me from `Approved` to `Not Approved`.
> The query did not return any results.

## Steps to Reproduce

1. Create a form
2. Submit data more than 10 times. (Let's say 12)
3. On Table view, display results with 10 rows per page.
4. Go to Page 2 and update records to (for example) `Approved` status.
5. Go to Page 1, filter results by choosing `Approved` from the dropdown
6. Check the main checkbox in the header row. 
7. Click `Select All` 
8. Click `Update Selected`, choose a different status. 

## Expected behavior 

- Table view should refresh as an empty table, still filtered.
- The Notification ("query didn't return results") should not be displayed.

## Actual behavior 

Nothing changes and message `The query did not return any results.` is displayed

## Additional details

arsea:
## Description

1 - When bulk updating all pre-filtered submissions, it refreshed to an empty table and no error message is displayed
2- When list is filtered, upper (1-10 results out of 29) and lower (Page 1 of 1 + Next/Previous) pagination is updated to reflect the new total results number.
3- When filtering and Clicking Select All, if we change the filter, the checkbox is unselected

## Related issues

Fixes #2319 
Fixes #2320 

**When those branches/PR are merged into master, don't forget to change base from `two-databases` to `master` to remove commits already reviewed**

Cover a part of #2328


jsapp/js/components/table.es6
```diff
@@ -79,37 +79,42 @@ export class DataTable extends React.Component {
...
+            loading: false,
+            selectedRows: {},
+            tableData: results
           });
           // TODO: debounce the queries and then enable this notification
-          alertify.warning(t('The query did not return any results.'));
+          // Block the warning if selectAll is true
+          if (!this.state.selectAll) {
+            alertify.warning(t('The query did not return any results.'));
+          }
```
magicznyleszek: Don't we want to get rid of this `alertify` call and display an empty message when there are zero results?

jsapp/js/components/table.es6
```diff
@@ -938,7 +1010,7 @@ export class DataTable extends React.Component {
           columns={selectedColumns || columns}
           defaultPageSize={defaultPageSize}
           pageSizeOptions={[10, 30, 50, 100, 200, 500]}
-          minRows={1}
+          minRows={0}
```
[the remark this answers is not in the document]
arsea: When no results are returned, it was still showing an empty row. This way, the table is completely empty.

jsapp/js/components/table.es6
arsea: I'm not sure I understand, an empty message?

jsapp/js/components/table.es6
```diff
@@ -816,6 +836,49 @@ export class DataTable extends React.Component {
...
+      selectedCount = this.state.resultsTotal;
+    } else {
+      data.submissions_ids = Object.keys(this.state.selectedRows);
+      selectedCount = data.submissions_ids.length;
+    }
+
+    const dialog = alertify.dialog('confirm');
+    const opts = {
+      title: t('Delete selected submissions'),
+      message: t('You have selected ## submissions. Are you sure you would like to delete them? This action is irreversible.').replace('##', selectedCount),
```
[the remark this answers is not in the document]
arsea: I will change it

jsapp/js/components/table.es6
```diff
@@ -824,35 +887,59 @@ export class DataTable extends React.Component {
...
+    }
+    const selectedLabel = t('## selected').replace('##', selectedCount);
+
     return (
       <bem.FormView__item m='table-meta'>
-        {showingResults}
-        {this.state.selectAll ?
+        <span>{showingResults}</span>
+
+        {this.state.selectAll &&
```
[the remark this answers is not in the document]
arsea: @magicznyleszek you mean having the option available as soon as more than one row has been selected?

jsapp/js/components/table.es6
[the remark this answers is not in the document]
magicznyleszek: Nice!

jsapp/js/components/table.es6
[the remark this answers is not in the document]
magicznyleszek: Yup :)

jsapp/js/components/table.es6
magicznyleszek: Instead of this:

Screenshot 2019-08-08 at 20 09 55

Do this:

Screenshot 2019-08-08 at 20 11 51

magicznyleszek: What do you think?

arsea: Sounds good I'll do that

jsapp/js/components/table.es6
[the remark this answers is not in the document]
arsea: Good idea!

jsapp/js/components/table.es6
tinok: Could we change the wording to `Your filters returned no submissions.`?

arsea: will do!

kpi/views/v2/data.py
```diff
@@ -20,7 +21,7 @@
 
 
 class DataViewSet(AssetNestedObjectViewsetMixin, NestedViewSetMixin,
```
jnm: @noliveleger, I'm getting ``AssertionError: 'DataViewSet' should either include a `queryset` attribute, or override the `get_queryset()` method.`` if I try to access the data endpoint in a browser without specifying a format, e.g. `http://kf.kobo.local:9000/api/v2/assets/aS8uoRSD5C9qNtkr35GUWd/data/?limit=30&start=0&sort={%22_id%22:-1}`.

Also, if I request XML format and still include `sort={%22_id%22:-1}`, I get `FieldError: Cannot resolve keyword u'_id' into field. Choices are: date_created, date_modified, deleted_at, id, status, user, user_id, uuid, xform, xform_id, xml`.

kpi/paginators.py
```diff
@@ -1,7 +1,84 @@
...
+)
 from rest_framework.reverse import reverse_lazy
+from rest_framework.utils.urls import remove_query_param, replace_query_param
+
+
+class DataPagination(PageNumberPagination):
+    """
+    Pagination class for submissions.
+    Because DataViewSet doesn't provide a real Django ORM QuerySet and pagination
+    is based on `start` and `limit` querystring parameters, some methods of
```
jnm: If we had the `v2` endpoint use `offset` instead of `start`, could we use DRF's `LimitOffsetPagination`? And, would that free us from having to re-implement as many things as you've had to already, or would it not save any work?

kpi/deployment_backends/kobocat_backend.py
```diff
@@ -488,6 +488,24 @@ def get_submission_validation_status_url(self, submission_pk):
...
+        :return: mixed. JSON or XML
+        """
+
+        if pk:
+            submissions = list(self.get_submissions(format_type, [int(pk)], **kwargs))
+            if len(submissions) > 0:
+                return submissions[0]
+            return None
+        else:
+            raise ValueError(_('Primary key must be provided'))
```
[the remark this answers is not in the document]
noliveleger: FYI: https://github.com/kobotoolbox/kpi/blob/a17ab11e81e13fdc16aad1c9d59b44bac70e5dc9/kobo/apps/hook/models/service_definition_interface.py#L33
is called from `HookSignalViewSet`.

kpi/paginators.py
```diff
@@ -1,7 +1,84 @@
...
+    PageNumberPagination had to be rewritten to support that case.
+    """
+    page_size_query_param = 'limit'
+    start_query_param = 'start'
+    max_page_size = 500
```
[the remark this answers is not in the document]
noliveleger: Yes and no :-D. 
When pagination is activated, maximum is `max_page_size`. Otherwise it's value on settings. 
But it makes non-sense. I have changed it.

kpi/deployment_backends/kobocat_backend.py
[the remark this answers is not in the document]
noliveleger: I think it's ok now. Validation takes place within the views as requested and it returns a 404 is `pk` is not valid.

kpi/deployment_backends/kobocat_backend.py
```diff
@@ -488,6 +488,24 @@ def get_submission_validation_status_url(self, submission_pk):
...
+        """
+        Returns only one occurrence.
+
+        :param pk: int. `Instance.id`
+        :param format_type: str.  INSTANCE_FORMAT_TYPE_JSON|INSTANCE_FORMAT_TYPE_XML
+        :param kwargs: dict. Filter params
+        :return: mixed. JSON or XML
+        """
+
+        if pk:
```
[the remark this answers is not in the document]
noliveleger: I've updated the docstring.

kpi/deployment_backends/kobocat_backend.py
```diff
@@ -488,6 +488,24 @@ def get_submission_validation_status_url(self, submission_pk):
...
+            submissions = list(self.get_submissions(format_type, [int(pk)], **kwargs))
+            if len(submissions) > 0:
```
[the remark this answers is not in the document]
noliveleger: The goal of this statement was not to test whether multiple submissions existed but only index 0.
I've refactored the code to be more `EAFP` stylish ;-)

kpi/views/v2/data.py
noliveleger: Hmm. Good catch. I always tested it with the API, KPI UI and unittests which all use JSONRenderer by default.

For the `XML` format, I have desactivated the sort. I've forced `sort` by id. 
I'll open an issue to restore sorting for `XML`.

kpi/paginators.py
noliveleger: Awesome, thank you. I had to specify Limit values, so I refactored the `DataPagination` class.
it's way better like this.

kpi/views/v2/data.py
```diff
@@ -192,6 +194,12 @@ def edit(self, request, pk, *args, **kwargs):
                                                            params=request.GET)
         return Response(**json_response)
 
+    def get_queryset(self):
+        # This method is needed when pagination is activated and renderer is
+        # `BrowsableAPIRenderer`. Because data comes from Mongo, `list()` and
+        # `retrieve()` don't need Django Queryset, we only need pass.
+        pass
```
jnm: `return None` is what happens implicitly here; it might be clearer to write that explicitly

kpi/deployment_backends/kobocat_backend.py
```diff
@@ -610,12 +618,16 @@ def __get_submissions_in_xml(self, instance_ids=[], **kwargs):
...
-        sort = {'id': 1}
         kwargs['instance_ids'] = instance_ids
 
         if 'fields' in kwargs:
             raise ValueError(_('`Fields` param is not supported with XML format'))
 
+        # FIXME. Use Mongo to sort data and ask PostgreSQL to follow the order.
+        # See. https://stackoverflow.com/a/867578
+        if 'sort' in kwargs:
+            raise ValueError(_('`sort` param is not supported with XML format'))
```
jnm: The problem with `ValueError` here is that it will return a 500 to the client, and with `DEBUG=False`, the explanatory message will be hidden.
image

kpi/views/v2/data.py
```diff
@@ -4,8 +4,10 @@
 from django.conf import settings
 from django.http import Http404
 from django.utils.translation import ugettext_lazy as _
-from rest_framework import exceptions, renderers, serializers, viewsets
+from rest_framework import renderers, viewsets
+from rest_framework.exceptions import ValidationError
```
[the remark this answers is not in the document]
noliveleger: I did that to avoid loading all the modules, but it really makes sense. 
I restored it as it was.

PR #2350, opened 2019-08-07
Fixes #2349 to remove settings iframe

### LINKED ISSUE #2349: Remove unnecessary iframe from general settings screen
The kc iframe is no longer needed in the general settings page. It now has its own section in Media:

image

tinok:
## Description

This removes the iframe from the general settings view and deletes the associated css needed to show a potentially large KC settings screen.

## Related issues

Fixes #2349

### DIFF
--- a/jsapp/js/components/formSubScreens.es6
+++ b/jsapp/js/components/formSubScreens.es6
@@ -82,9 +82,7 @@ export class FormSubScreens extends React.Component {
         case `/forms/${this.state.uid}/data/downloads`:
           return this.renderProjectDownloads();
         case `/forms/${this.state.uid}/settings`:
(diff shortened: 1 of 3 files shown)


arsea:
I think we should also remove [the whole scss block](https://github.com/kobotoolbox/kpi/blob/094fea0cfb04fe8f394b8a20c18d4bba1a359ce0/jsapp/scss/components/_kobo.form-view.scss#L478).

I'm not sure why the `height: 100%` is there and the bottom padding is ignored, making the "Delete project and Data" button stick to the bottom of the div.

What do you think @tinok ?

magicznyleszek:
@tinok @arsea I added some cleanup and merged this :)

PR #2360, opened 2019-08-14
Remove potential errors fixer that by itself caused errors

magicznyleszek:
Fixes Case A from #2363


arsea:
@magicznyleszek I still get errors. The first language stays intact, but the first time I tried, my second language became "Untitled" (no screenshot, sorry).
When I tried to replicate, I created 3 languages.
Screen Shot 2019-08-14 at 09 46 24
I then made changes to the form and saved. My third language disappeared
Screen Shot 2019-08-14 at 09 46 51
I then made other changes to the form but nothing changed within the languages then.

PR #2374, opened 2019-08-27
TableView: Bulk options don't change accordingly to users' permissions.

### LINKED ISSUE #2373: Bulk selection only appears for users with 'Change submissions' permission and does not display options according to their permissions
##  Steps to Reproduce

1. Create a form with User A
2. Deploy it before sharing (because of bugfix not merged yet: #2340)
2. Share this form with User B and gives User B 'change submissions' permission.
3. Share this form with User C and gives User C 'validate submissions' permission.
4. In another browser, log into User B's account
5. Open Table View. 
4. In another browser, log into User C's account
5. Open Table View. 

## Expected behavior 

- Bulk selection is not available for User B.
- Bulk selection is available with all options for User C. (`Validation status` and `Delete` options)

## Actual behavior 

- Bulk selection should be available for User B and only `Delete` option should be displayed in the `Bulk menu`.
- Bulk selection should be available for User C and only `Validation status` options should be displayed in the `Bulk menu`.

## Additional info
Based on #1970

arsea:
## Description

Conditional showing of options in the "X selected" dropdown menu depending on permissions.
Select multiple checkboxes available for users with Change AND Validation permissions.

## Related issues

Fixes #2373

### DIFF
--- a/jsapp/js/components/table.es6
+++ b/jsapp/js/components/table.es6
@@ -907,19 +907,17 @@ export class DataTable extends React.Component {
         {Object.keys(selected).length > 0 &&
           <ui.PopoverMenu type='bulkUpdate-menu' triggerLabel={selectedLabel} >
             {this.userCan('validate_submissions', this.props.asset) &&
-              <React.Fragment>
-                {VALIDATION_STATUSES_LIST.map((item, n) => {
-                  return (
-                    <bem.PopoverMenu__link
-                      onClick={this.onBulkUpdateStatus}
-                      data-value={item.value}
-                      key={n}
-                    >
-                      {t('Set status: ##status##').replace('##status##', item.label)}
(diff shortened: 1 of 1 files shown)


jsapp/js/components/table.es6
```diff
@@ -906,21 +906,27 @@ export class DataTable extends React.Component {
...
-                  onClick={this.onBulkUpdateStatus}
-                  data-value={item.value}
-                  key={n}
-                >
-                  {t('Set status: ##status##').replace('##status##', item.label)}
-                </bem.PopoverMenu__link>
-              );
-            })}
+            {this.userCan('validate_submissions', this.props.asset) &&
+              <React.Fragment>
```
magicznyleszek: This `<React.Fragment>` isn't necessary, as `.map` returns an array of elements, which is ok for React :)

PR #2386, opened 2019-09-01
Fixes for submission endpoint parameter validation

jnm:
As discussed at https://www.flowdock.com/app/kobotoolbox/kobo/threads/H0b6vTHuvJIseGOqslQb020YkCW


kpi/views.py
```diff
@@ -960,6 +960,12 @@ def list(self, request, *args, **kwargs):
         # Do not allow requests to retrieve more than `SUBMISSION_LIST_LIMIT`
         # submissions at one time
         limit = filters.get('limit', settings.SUBMISSION_LIST_LIMIT)
+        try:
+            limit = int(limit)
+        except ValueError:
+            raise exceptions.ValidationError(
+                {'limit': _('A valid integer is required')}
+            )
```
noliveleger: I would have refactored the code to use the same validator in both place (here and `BaseDeploymentBackend`)
I'll do it in one of the PRs for `two-databases`.

PR #2387, opened 2019-09-03
Show 'SUMMARY' & 'DATA' tabs (with filtered data) to users with 'partial_view_submissions' permission.

### LINKED ISSUE #2385: [two databases] Users with `partial submissions` don't have access to `DATA` tab.
Users with Partial Submissions should be able to see the data of users they supervise.

## Steps to Reproduce

1. Create a form
2. Assign `Partial Permissions` to User B
3. Log into `UserB` account.
4. Open form create at step1. 

## Expected behavior 

Tabs `FORM` and `DATA` should be available
Not sure about `SUMMARY`.

## Actual behavior 

Only tab `FORM` is available

noliveleger:
It covers `FE` & `BE` issues. 
Fixes #2385 

## Test procedure 
I followed these steps to ensure everything was working as expected.

- Go to `kpi` admin platform. 
- Create two new users `User A` and `User B`
- Log into `User A`'s account
- Go to `Accounts settings` and check `Require authentication to see forms and submit data`
- Create a new form 
- Assign `Partial View Submissions (for `User B`), Add Submissions` to `User B`. 
- Open the form in Enketo 
- Submit data once with `User A`'s credentials.
- Log into `User B`'s account (in `kpi`) with another browser.
- Open the form in Enteko. Be sure to log out if you are already logged in.
- Submit data once with `User B`'s credentials.

`User A` should see 2 submissions in:
- Summary
- Reports 
- Table View
- Exports

_Maps & Gallery depend on the type of questions you added in the form_

`User B` should see only 1 submission in:
- Summary
- Reports 
- Table View (Number of submissions is wrong: Read below `know issues`)
- Exports

_Maps depends on the type of questions you added in the form_
**Gallery and legacy exports can't work** because `kc` doesn't know anything about `partial_view_submission` permission. iframe shows `Not shared`. @jnm suggested to hide those options. 

N.B.: I've also follow the same procedure with the registration form and two other new users.

## Known issues

1. ~~Total of submissions is wrong in the header for users with `partial_view_submission` permission:~~
    ~~`FE` gets value from the asset's `deployment__submission_count` property which comes from `XForm`.num_of_submissions property.~~
    - ~~https://github.com/kobotoolbox/kpi/blob/2385-partial-submission-permissions-fixes/kpi/serializers/v2/asset.py#L256-L259~~
    - ~~https://github.com/kobotoolbox/kpi/blob/2385-partial-submission-permissions-fixes/kpi/deployment_backends/base_backend.py#L110-L112~~
    - ~~https://github.com/kobotoolbox/kpi/blob/2385-partial-submission-permissions-fixes/kpi/deployment_backends/kobocat_backend.py#L423-L430~~
    - ~~https://github.com/kobotoolbox/kpi/blob/2385-partial-submission-permissions-fixes/kpi/deployment_backends/kc_access/utils.py#L46-L53~~
    
    ~~It can be easily fixed in `kpi` by querying mongo with correct filters but it can be a performance issue if there is lots of instances.~~ 

     Fixed in this PR.
     Users can still read `kc` value from `kc`'s API.

2. Number of submissions is wrong in Table View. 
     It's fixed within PR #2331


jsapp/js/components/formSubScreens.es6
```diff
@@ -41,11 +41,14 @@ export class FormSubScreens extends React.Component {
     }
   }
   render () {
+    let permAccess = false;
+
+    (this.userCan('view_submissions', this.state) || this.userCan('partial_submissions', this.state)) ? permAccess = true : '';
```
magicznyleszek: I think you can make it a one liner:

```
let permAccess = this.userCan('view_submissions', this.state) || this.userCan('partial_submissions', this.state);
```

jsapp/js/components/formSummary.es6
```diff
@@ -319,6 +319,9 @@ class FormSummary extends React.Component {
   }
   render () {
     let docTitle = this.state.name || t('Untitled');
+    let permAccess = false;
+
+    (this.userCan('view_submissions', this.state) || this.userCan('partial_submissions', this.state)) ? permAccess = true : '';
```
magicznyleszek: I think you can make it a one liner:

```
let permAccess = this.userCan('view_submissions', this.state) || this.userCan('partial_submissions', this.state);
```

kobo/apps/reports/serializers.py
```diff
@@ -24,7 +25,11 @@ def to_representation(self, obj):
 
         split_by = request.query_params.get('split_by', None)
 
-        _list = report_data.data_by_identifiers(obj, vnames, split_by=split_by)
+        permission_filters = obj.get_filters_for_partial_perm(request.user.id)
+        submission_stream = obj.deployment.get_submissions(
```
jnm: I'm wondering if we should add a required parameter called something like `requesting_user` to `get_submissions()` and then have that method handle filtering for partial permissions. The goal would be to have the filtering logic in one place and to break (intentionally) any code that's retrieving submissions without making sure the requestor is entitled to see the full set.

kpi/models/import_export_task.py
```diff
@@ -566,12 +569,9 @@ def _run_task(self, messages):
...
         self.log_and_mark_stuck_as_errored(self.user, source_url)
 
-        if isinstance(source.deployment, MockDeploymentBackend):
-            # Currently used only for unit testing (`MockDeploymentBackend`)
-            # TODO: Have the KC backend also implement `_get_submissions()`?
-            submission_stream = source.deployment.get_submissions()
-        else:
-            submission_stream = None
+        permission_filters = source.get_filters_for_partial_perm(self.user.id)
+        submission_stream = source.deployment.get_submissions(
```
jnm: (Repeated comment) Consider consolidating partial permission filtering code inside `get_submissions()`

kobo/apps/reports/serializers.py
noliveleger: Indeed.

kpi/deployment_backends/base_backend.py
```diff
@@ -109,6 +109,15 @@ def validate_submission_list_params(cls, **kwargs):
 
         return params
 
+    def calculated_submission_count(self, **kwargs):
+        params = self.validate_submission_list_params(**kwargs)
+        # Remove useless property for count
+        params.pop('fields', None)
+        params.pop('start', None)
+        params.pop('start', None)
```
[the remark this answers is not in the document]
noliveleger: Exact ;-(

kpi/models/import_export_task.py
noliveleger: I'll do

kpi/tests/api/v2/test_api_assets.py
```diff
@@ -311,6 +317,42 @@ def test_asset_version_id_and_content_hash(self):
...
+        }
+        self.asset.assign_perm(anotheruser, PERM_PARTIAL_SUBMISSIONS,
+                               partial_perms=partial_perms)
+
+        response = self.client.get(self.asset_url, format='json')
+        self.assertEqual(response.data['deployment__submission_count'], 2)
+        self.client.logout()
+        self.client.login(username='anotheruser', password='anotheruser')
+        response = self.client.get(self.asset_url, format='json')
+        self.assertEqual(response.data['deployment__submission_count'], 1)
```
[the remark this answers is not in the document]
noliveleger: Ok,
The missing leading underscore is because of the `MockDeploymentBackend` which looks for `id`, not `_id`.
I will change it to be consistent with `Mongo` and `KobocatDeploymentBackend`.

kpi/tests/test_mock_data_exports.py
```diff
@@ -251,6 +300,18 @@ def test_csv_export_english_labels(self):
         ]
         self.run_csv_export_test(expected_lines, export_options)
 
+    def test_csv_export_english_labels_partial_submissions(self):
```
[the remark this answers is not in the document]
noliveleger: No reason. The partial permission test can be done only once on `csv` or `xls`.

Issue #2383, opened 2019-08-29
[two databases] Preview fails when opening it from landing page


noliveleger:
## Description

When open the preview of a form, the modal shows a blank page

## Steps to Reproduce

1. Create or use an existing form
2. Click on the eye to preview the form

## Expected behavior 

The preview shows the form

## Actual behavior 

The preview shows a blank page

## Additional details
It's because `FE` send to `BE` the new `v2` hyperlink identity value in its payload:
`asset: http://[kpi]/api/v2/assets/{asset_uid}/` when posting to `v1` `asset_snapshot` endpoint ->`http://[kpi]/asset_snapshots/`.
Either payload sends old value. (e.g `http://[kpi]/assets/{asset_uid}/`) to old endpoint, or 
payload sends new value to new endpoint (e.g `http://[kpi]/api/v2/asset_snapshot/`)

noliveleger:
As per discussed with @jnm, we are going to make a hotfix to use `v2` endpoint for `snapshot`. 
IMHO: as a temporary fix, I would look at the `data` object to see whether ` asset` url is based on `v2` or `v1` and call respective endpoint for `snapshot`

PR #2399, opened 2019-09-10
Use final snapshots fix

magicznyleszek:
## Description

Apply final fix as required PR was merged. Also moved both (very similar) methods next to each other.

## Related issues

Followup to #2388 (and #2383)


jsapp/js/dataInterface.es6
```diff
@@ -84,22 +84,21 @@ var dataInterface;
...
-      }
-
       return $ajax({
-        url: url,
+        url: `${ROOT_URL}/api/v2/asset_snapshots/`,
         method: 'POST',
         data: data
       });
     },
+    createTemporaryAssetSnapshot ({source}) {
```
noliveleger: Good move ;-)

PR #2402, opened 2019-09-11
Better UI of nested checkboxes

magicznyleszek:
## Description

Changes for "View submissions" and "Restrict…" checkboxes:

Screenshot 2019-09-11 at 10 52 35

### DIFF
--- a/jsapp/js/components/permissions/userAssetPermsEditor.es6
+++ b/jsapp/js/components/permissions/userAssetPermsEditor.es6
@@ -378,15 +378,10 @@ class UserAssetPermsEditor extends React.Component {
       formModifiers.push('pending');
     }
 
-    const formClassNames = classNames(
-      'user-permissions-editor',
-      isNew ? 'user-permissions-editor--new' : ''
-    );
-
     return (
       <bem.FormModal__form
         m={formModifiers}
-        className={formClassNames}
+        className='user-permissions-editor'
         onSubmit={this.submit}
       >
         {isNew &&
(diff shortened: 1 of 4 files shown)


jsapp/js/components/permissions/userAssetPermsEditor.es6
```diff
@@ -416,38 +411,34 @@ class UserAssetPermsEditor extends React.Component {
...
+                onChange={this.onCheckboxChange.bind(this, 'submissionsViewPartial')}
+                label={t('Restrict to submissions made by certain users')}
+              />
+
+              {this.state.submissionsViewPartial === true &&
+                <TagsInput
+                  value={this.state.submissionsViewPartialUsers}
+                  onChange={this.onSubmissionsViewPartialUsersChange}
+                  addOnBlur
+                  addKeys={[9, 13, 32]}
```
noliveleger: Not a blocker at all, Just a thought I'd like to share. 
I could not remember that key code `9` was for, so I was wondering what would you think about using constants instead of these key codes? 
Like
**constants.es6**:
```
KEY_CODES = {
  ENTER: 13,
  SPACE: 32,
  TAB: 9
}
```

and using them in the code
`addKeys={[KEY_CODES.TAB, KEY_CODES.ENTER, KEY_CODES.SPACE]}`

magicznyleszek: Good idea!

PR #2406, opened 2019-09-12
Bugfix: disaggregating map results by nested groups questions

### LINKED ISSUE #1476: Fix disaggregation for questions in a nested group in maps
See also #1467.

magicznyleszek:
## Description

The problem was because FE code ignored nested groups. I introduced a nice function that generates a list of survey questions with a flat path of each one (could be useful in other places). Added a test for it too.

## Related issues

Fixes #1476

### DIFF
--- a/jsapp/js/assetUtils.es6
+++ b/jsapp/js/assetUtils.es6
@@ -1,38 +1,35 @@
+import {QUESTION_TYPES} from 'js/constants';
+
 /**
- * @param {string} questionName
  * @param {object} survey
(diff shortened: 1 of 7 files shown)


magicznyleszek:
@jnm in the coffee code unfortunately there is not a single list with configs for question types. If you're referring to `jsapp/xlform/src/view.icons.coffee` - I can use es6 files there easily (i.e. `constants = require '../../js/constants'`). I can also build the other config-like places in coffee code using the constant :>

Issue #2316, opened 2019-07-12
Display `Asset` permission names differently depending on `asset_type`


jnm:
Use the type-dependent `label` returned by `/api/v2/assets/[asset uid]/permissions/` when displaying permissions in the frontend. See https://github.com/kobotoolbox/kpi/issues/2316#issuecomment-524524123. History: https://www.flowdock.com/app/kobotoolbox/kobo/threads/FnkA7h0HnqTZ4O8WPfcf5OU0oMX

jnm:
from @tinok, this is the per-`asset_type` naming we should use:

* asset_type = 'survey': Form
* asset_type = 'template': Template
* asset_type = 'block': Block
* asset_type = 'question': Question
* asset_type = 'collection': Collection

magicznyleszek:
I think that `'Can view {asset_type}'` is perfectly fine, would just vote for `'Can view ##asset_type##'`, as this is an existing convention on Frontend :)

jnm:
@magicznyleszek, I pushed [a branch](https://github.com/kobotoolbox/kpi/compare/two-databases...2316-permission-labels-by-asset-type?expand=1) where the API returns a type-dependent `label` for each _assigned_ permission, e.g. `/api/v2/assets/awu9kEjGa7AhHXaJbnJAmq/permissions/` returns something that includes:
```json
{
    "url": "http://kf.kobo.local:9000/api/v2/assets/awu9kEjGa7AhHXaJbnJAmq/permissions/pNzPwd383ypfpxjRMqkcCw/",
    "user": "http://kf.kobo.local:9000/api/v2/users/regular/",
    "permission": "http://kf.kobo.local:9000/api/v2/permissions/change_asset/",
    "label": "Change form"
},
```
Is this useful to you on the frontend? I haven't done anything with Collections yet.

magicznyleszek:
@jnm I confirm it would be useful :) Instead of getting the name from global config based on permission codename, we would just use it directly from permission. And we should probably remove permissions `name` from `${ROOT_URL}/api/v2/permissions/` while changing this?

jnm:
Great :) I'll make a PR for Oli. I left `name` in `${ROOT_URL}/api/v2/permissions/`, but it now returns what Django considers the "name," which we previously had in `description`. I removed `description` from the endpoint.

magicznyleszek:
@jnm so I should create a branch from `two-databases` after #2379 is done, right?

jnm:
[...]

PR #2409, opened 2019-09-12
Add `assignable_permissions` to v2 asset detail

jnm:
## Checklist

3. [x] Ensure the tests pass
5. [ ] If this is a big feature, make sure to prefix the title with `Feature:` and add a thorough description for non-dev folk

## Description

This allows clients to retrieve a per-asset list of assignable permissions, including appropriate labels based on the asset type.

## Related issues

Part of #2316


kpi/tests/api/v2/test_api_assets.py
```diff
@@ -311,6 +299,44 @@ def test_asset_version_id_and_content_hash(self):
...
+        self.assertListEqual(
+            response.data['assignable_permissions'],
+            [
+                {'label': u'View form',
+                 'url': 'http://testserver/api/v2/permissions/view_asset/'},
+                {'label': u'Add submissions',
+                 'url': 'http://testserver/api/v2/permissions/add_submissions/'},
+                {'label': u'View submissions',
+                 'url': 'http://testserver/api/v2/permissions/view_submissions/'},
+                {'label': u'Change submissions',
```
noliveleger: Does not pass the tests because the label is `Change and delete submissions` now.

kpi/tests/api/v2/test_api_assets.py
```diff
@@ -154,10 +140,12 @@ def test_restricted_access_to_version(self):
         self.assertEqual(resp2.data['detail'], 'Not found.')
 
 
-class AssetsDetailApiTests(BaseTestCase):
+class AssetsDetailApiTests(BaseAssetTestCase):
     fixtures = ['test_data']
 
-    URL_NAMESPACE = ROUTER_URL_NAMESPACE
+    def __init__(self, *args, **kwargs):
+        self.URL_NAMESPACE = ROUTER_URL_NAMESPACE
```
noliveleger: What is the need to add `URL_NAMESPACE` to instance's context?
Not that I'm against it but it's strange to have only this class using it this way. 
A wise man told me one day, we had to be consistent 😏

kpi/tests/api/v2/test_api_assets.py
jnm: Oops

kpi/tests/api/v2/test_api_assets.py
jnm: Aha, yes. I was hoping you'd comment on this. I knew either I was confused or something was off about the `URL_NAMESPACE` stuff overall; I'm glad it turned out to be the former :)

PR #2410, opened 2019-09-13
Permissions names (Frontend)

magicznyleszek:
## Description

Frontend part of the thing. 

## Related issues

Part of #2316

### DIFF
--- a/jsapp/js/components/permissions/userPermissionRow.es6
+++ b/jsapp/js/components/permissions/userPermissionRow.es6
@@ -120,7 +120,7 @@ class UserPermissionRow extends React.Component {
             {permName}
 
             {permUsers.length > 0 &&
-              String.fromCharCode(160) + '(' + permUsers.join(', ') + ')'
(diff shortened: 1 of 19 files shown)


jsapp/js/components/permissions/userPermissionRow.es6
```diff
@@ -100,14 +102,25 @@ class UserPermissionRow extends React.Component {
...
             title={perm.description}
-            key={perm.name}
+            key={permName}
           >
-            {perm.name}
+            {permName}
 
             {permUsers.length > 0 &&
-              ' (' + permUsers.join(', ') + ')'
+              String.fromCharCode(160) + '(' + permUsers.join(', ') + ')'
```
noliveleger: What about add it to `KEY_CODES' constant as well?

magicznyleszek: Good thinking!

PR #2412, opened 2019-09-15
Restore ability to uncheck public sharing boxes

### LINKED ISSUE #2407: [two-databases] Publicly shared options can't be unchecked
## Description

When a project is shared publicly, it's not possible to uncheck the checkboxes until the modal stays open.

## Steps to Reproduce

1. Open sharing permissions of a project
2. Check one on the checkboxes under `Share publicly by link`
3. Uncheck the same checkboxes

## Expected behavior 

Checkboxes become unchecked

## Actual behavior 

Checkboxes stay checked

## Additional details

Publicly access is correctly set on first attempt (If we close the modal and re-open it, the checkboxes are unchecked)

jnm:
Fixes #2407 and removes some unneeded front-end code


.eslintrc.json
```diff
@@ -30,6 +30,7 @@
     "no-case-declarations": 1,
     "no-confusing-arrow": 1,
     "no-console": 1,
+    "no-debugger": 1,
```
magicznyleszek: I didn't know `debugger` was a thing in JavaScript :-o

jsapp/js/components/permissions/userPermissionRow.es6
```diff
@@ -120,8 +120,13 @@ class UserPermissionRow extends React.Component {
...
             {permName}
 
-            {permUsers.length > 0 &&
+            { // KEY_CODES.NBSP doesn't work for me in Chrome or FF, but it's
+              // okay because (for now) we're not including this username list
+              // anyway
+              /*
+              permUsers.length > 0 &&
               String.fromCharCode(KEY_CODES.NBSP) + '(' + permUsers.join(', ') + ')'
+              */
```
magicznyleszek: Ouch! Maybe we should remove this code and `KEY_CODES.NBSP` instead of commenting out?

.eslintrc.json
jnm: :grin: I like it

jsapp/js/components/permissions/userPermissionRow.es6
jnm: As discussed, I tried showing a maximum of three usernames, with (`and ## more`) as a placeholder when the number of assigned users exceeds that. I'm not sure if I like the result or not:
image
I'll push the code anyway.

magicznyleszek: It's not perfect, I agree, but I think it's more than OK for initial version of row level perms :) Let's merge this PR?

PR #2417, opened 2019-09-18
Poor API v2 performance vs API v1

### LINKED ISSUE #2413: [two-databases] Poor performance from v1 and v2 asset list endpoints
https://www.flowdock.com/app/kobotoolbox/kobo/threads/kYWmgAyzErxc4NO7tAcsciig8sI

I think the gist of this is [if you use `prefetch_related()`](https://github.com/kobotoolbox/kpi/blob/830ab1713479be922748ce036dae7a2c84ca61c3/kpi/models/asset.py#L769-L776), you can get the results of the related queryset _in their entirety_ without additional database queries. However, `filter()`ing them will hit the database again:
https://github.com/kobotoolbox/kpi/blob/830ab1713479be922748ce036dae7a2c84ca61c3/kpi/utils/object_permission_helper.py#L43-L49

Illustration:
```
In [1]: from django.db.backends import utils as db_backends_utils
   ...: from debug_toolbar.management.commands.debugsqlshell import PrintQueryWrapper
   ...: 
   ...: db_backends_utils.CursorDebugWrapper = PrintQueryWrapper
   ...: 

In [2]: assets = Asset.objects.all().prefetch_related('permissions__user')

In [3]: asset = assets.first()
SELECT <snip> [0.96ms]

In [4]: [p.user_id for p in asset.permissions.all()] # we get this for free!
Out[4]: [2, 2, 2, 2, 2, 2]

In [5]: [p.user_id for p in asset.permissions.filter(user_id=2)]
SELECT "kpi_objectpermission"."id",
       "kpi_objectpermission"."user_id",
       "kpi_objectpermission"."permission_id",
       "kpi_objectpermission"."deny",
       "kpi_objectpermission"."inherited",
       "kpi_objectpermission"."object_id",
       "kpi_objectpermission"."content_type_id",
       "kpi_objectpermission"."uid"
FROM "kpi_objectpermission"
WHERE ("kpi_objectpermission"."object_id" = 11
       AND "kpi_objectpermission"."content_type_id" = 28
       AND "kpi_objectpermission"."user_id" = 2) [0.63ms]
Out[5]: [2, 2, 2, 2, 2, 2]
```

noliveleger:
It uses `django_cache_request` to avoid sending same queries multiple times in a row.

Closes #2413


kpi/models/object_permission.py
```diff
@@ -948,3 +962,194 @@ def _update_partial_permissions(self, user_id, perm, remove=False,
...
+                object_id=self.pk)
+            for user_id, object_permissions in all_object_permissions.items():
+                perms += build_dict(user_id, object_permissions)
+
+        return set(perms)
+
+    @staticmethod
+    @cache_for_request
+    def __get_permissions_for_content_type(content_type_id, codenames,
+                                           startswith=True, first=False):
```
jnm: tbh, writing `__get_permissions_for_content_type(…, first=True)`, or even `__get_permissions_for_content_type(…, True)`, when calling this method is longer than typing `__get_permissions_for_content_type(…)[0]` :wink:

kpi/models/asset.py
```diff
@@ -666,7 +666,8 @@ def get_filters_for_partial_perm(self, user_id, perm=PERM_VIEW_SUBMISSIONS):
         return None
 
     def get_grant_permissions(self):
-        return self.permissions.filter(deny=False)
+        return self.permissions.filter(deny=False).\
```
jnm: Let's move this method and `get_label_for_permission()` into `ObjectPermissionMixin`

kpi/serializers/v1/asset.py
```diff
@@ -49,12 +49,37 @@ class Meta(AssetSerializerV2.Meta):
                   'permissions',
                   'settings',)
 
+    # Restricting permissions with `get_grant_permissions` add extra queries
+    # TODO optimize this
```
jnm: Could we fix it by simply having `get_grant_permissions()` do the filtering in Python instead of calling `filter()`?

kpi/serializers/v2/asset.py
```diff
@@ -343,6 +343,41 @@ class Meta(AssetSerializer.Meta):
...
+                        assignment.user_id in filtered_user_ids):
+                    user_assignments.append(assignment)
+
+            context.update({'asset': obj})
+            context.update({'asset_uid': obj.uid})
+
+            return AssetPermissionAssignmentSerializer(user_assignments,
+                                                       many=True, read_only=True,
+                                                       context=context).data
+        except KeyError:
```
jnm: It looks like `self.context['object_permissions_per_object']` is the only lookup by key, so can we `except` right after that? If the two possible execution paths are

1. finding `'object_permissions_per_object'` in `self.context` and eventually returning an instance of `AssetPermissionAssignmentSerializer`, and
2. _not_ finding `'object_permissions_per_object'` and punting to the superclass' `get_permissions()`,

then we should be able to `return super(AssetListSerializer, self).get_permissions(obj)` in the `except` right after `try`ing `self.context['object_permissions_per_object']`.

It's not purely a stylistic thing: what if a `KeyError` was raised in `user_can_share()` or the constructor for `AssetPermissionAssignmentSerializer` due to a programming error? It would be discarded silently.

kpi/utils/object_permission_helper.py
```diff
@@ -43,11 +43,11 @@ def get_assignments_queryset(cls, affected_object, user):
...
         # because it's specific to `ObjectPermission`.
         if not user or user.is_anonymous():
-            queryset = queryset.filter(user_id=affected_object.owner.pk)
+            queryset = queryset.filter(user_id=affected_object.owner_id)
         elif not cls.user_can_share(affected_object, user):
             # Display only users' permissions if they are not allowed to modify
             # others' permissions
             queryset = queryset.filter(user_id__in=[user.pk,
-                                                    affected_object.owner.pk])
+                                                    affected_object.owner_id])
```
jnm: I'm totally in favor of this but am embarrassed to say that I can't find Django documentation to support my observation that `related_field_id` avoids the extra query required for `related_field.pk`. If you see any, please let me know.

kpi/views/v2/asset.py
```diff
@@ -203,6 +210,37 @@ def get_queryset(self, *args, **kwargs):
...
 
+    def get_serializer_context(self):
+        """
+        Extra context provided to the serializer class.
+        """
+
+        context_ = super(AssetViewSet, self).get_serializer_context()
+        if self.action == 'list':
+            # To avoid making a triple join-query for each asset in the list
+            # to retrieve object permissions. We make a big one at beginning
```
jnm: "To avoid…permissions, we make"

kpi/views/v2/asset.py
```diff
@@ -203,6 +210,37 @@ def get_queryset(self, *args, **kwargs):
...
+                select_related('user', 'permission')
+
+            object_permissions_per_object = defaultdict(list)
+
+            for op in object_permissions:
+                object_permissions_per_object[op.object_id].append(op)
+
+            context_.update({
+                'object_permissions_per_object': object_permissions_per_object
+            })
```
jnm: A quick glance at DRF makes me think `context_` is a `dict`; could we `context_['object_permissions_per_object'] = object_permissions_per_object`?

kpi/serializers/v2/asset.py
```diff
@@ -343,6 +343,41 @@ class Meta(AssetSerializer.Meta):
...
+                # others' permissions
+                filtered_user_ids = [obj.owner_id, user.pk]
+
+            for assignment in assignments:
+                if (filtered_user_ids is None or
+                        assignment.user_id in filtered_user_ids):
+                    user_assignments.append(assignment)
+
+            context.update({'asset': obj})
+            context.update({'asset_uid': obj.uid})
```
jnm: A quick glance at DRF makes me think `context` is a `dict`; could we `context[key] = value` instead of calling `update()`, or is `context` actually something fancier?

kpi/models/asset.py
jnm: Let's remove it altogether! :boom:

kpi/serializers/v1/asset.py
jnm: I feel good about this

kpi/models/object_permission.py
```diff
@@ -948,3 +962,194 @@ def _update_partial_permissions(self, user_id, perm, remove=False,
...
+                ],
+                '<user_id>': [
+                    (permission_id, permission_codename, deny),
+                    (permission_id, permission_codename, deny),
+                    ...
+                ]
+            }
+        """
+        records = ObjectPermission.objects. \
+            filter(content_type_id=content_type_id, object_id=object_id).\
```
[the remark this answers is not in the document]
noliveleger: I don't want space. It's a default setting of my IDE. I forget to change it. 
I'll update everywhere you noticed.

kpi/models/object_permission.py
```diff
@@ -948,3 +962,194 @@ def _update_partial_permissions(self, user_id, perm, remove=False,
...
+        Returns a set of user ids and object permission ids related to
+        object `self`.
+
+        Args:
+            is_denied (bool): If `True`, returns denied permissions
+            user (User)
+            codename (str)
+
+        Returns:
+            set: [(User's pk, Permission's pk)]
```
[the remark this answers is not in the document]
noliveleger: I'm in favor to use `.pk` everywhere we need to use a `Django` object's primary key.  (which I started to do since I read https://stackoverflow.com/questions/2165865/django-queries-id-vs-pk)
Even if I would keep `user_id` as a variable (or argument).

kpi/models/object_permission.py
```diff
@@ -948,3 +962,194 @@ def _update_partial_permissions(self, user_id, perm, remove=False,
...
+        if isinstance(codenames, string_types):
+            codenames = [codenames]
+
+        if startswith:
+            if len(codenames) > 1:
+                raise NotImplementedError('Can search for only one codename '
+                                          'when `startswith` is `True`')
+            filters.update({'codename__startswith': codenames[0]})
+        else:
+            filters.update({'codename__in': codenames})
```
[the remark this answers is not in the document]
noliveleger: `codename__in` is not needed anymore.  I removed it.

kpi/utils/object_permission_helper.py
noliveleger: Couldn't find any either.

jnm: 🤷

kpi/models/object_permission.py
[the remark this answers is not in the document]
jnm: Great, thanks for investigating

PR #2426, opened 2019-09-30
Display API token in settings

### LINKED ISSUE #2418: Display API token in user account settings
## Description

As a user I'd like to see the API token in my account settings rather than having to navigate to  https://[kpi-url]/token/?format=json. 

## Suggested UI implementation

The token could be displayed in clear text or hidden by default (e.g. clicking a 👁 button to toggle visibility). 

It could be placed under the Change Password button.

magicznyleszek:
## Description

API token is obfuscated until users clicks the button.

## Related issues

Fixes #2418


magicznyleszek:
@jnm this PR was wrongly set to be merged into `master` - it's not as big as you can see now :-P

PR #2429, opened 2019-10-02
Audit option in Form Builder

### LINKED ISSUE #2187: Add the audit meta question type to the formbuilder
## Description

`audit` is supported in xlsform and pyxform, so it should be added to the formbuilder as well. 

See http://support.kobotoolbox.org/creating-forms/audit-logging-meta-question-type

magicznyleszek:
## Description

Displays `audit` meta question in the "Layout & Settings" section, with editable `appearance` field.

## Related issues

Fixes #2187

### DIFF
--- a/jsapp/js/components/metadataEditor.es6
+++ b/jsapp/js/components/metadataEditor.es6
@@ -21,8 +21,7 @@ export default class MetadataEditor extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      metaProperties: [],
-      auditAppearance: null
+      metaProperties: []
     };
     autoBind(this);
   }
@@ -64,8 +63,8 @@ export default class MetadataEditor extends React.Component {
     }
   }
 
-  onAuditAppearanceChange(newVal) {
-    this.getSurveyDetail(META_QUESTION_TYPES.get('audit')).set('appearance', newVal);
+  onAuditParametersChange(newVal) {
+    this.getSurveyDetail(META_QUESTION_TYPES.get('audit')).set('parameters', newVal);
     this.rebuildState();
     if (typeof this.props.onChange === 'function') {
       this.props.onChange();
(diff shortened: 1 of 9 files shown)


jsapp/xlform/src/model.surveyDetail.coffee
```diff
@@ -6,16 +6,26 @@ module.exports = do ->
...
     toJSON: ()->
       if @get("value")
-        nameSlashType = @get("name")
-        name: nameSlashType
-        type: nameSlashType
+        out = {}
+
+        out.name = @get("name")
+        # type is same as name
+        out.type = out.name
```
jnm: Does it make sense to you that this always uses the name as the type? In https://github.com/kobotoolbox/kpi/blob/2187-audit/jsapp/xlform/src/model.configs.coffee, there's a `type` inside the `asJson` object for each setting, but I can't tell where that's used for anything except this one check: https://github.com/kobotoolbox/kpi/blob/f62a0c58a0b4c90c6e5a96182feb86a54a74dcfb/jsapp/xlform/src/model.surveyFragment.coffee#L346-L356

Of course, all the `name`s in the `asJson`s happen to be identical to the `type`s :slightly_smiling_face:

jsapp/js/components/metadataEditor.es6
```diff
@@ -0,0 +1,163 @@
...
+  onCheckboxChange(name, isChecked) {
+    this.getSurveyDetail(name).set('value', isChecked);
+    this.rebuildState();
+    if (typeof this.props.onChange === 'function') {
+      this.props.onChange();
+    }
+  }
+
+  onAuditAppearanceChange(newVal) {
+    this.getSurveyDetail(META_QUESTION_TYPES.get('audit')).set('appearance', newVal);
```
jnm: The spec actually requires this to be `parameters`, not `appearance` (the original issue description was mistaken). I already have a local commit to fix it.

jsapp/xlform/src/model.surveyDetail.coffee
jnm: https://github.com/kobotoolbox/kpi/issues/2467

PR #2443, opened 2019-10-08
use properer listener for bulk permissions call

magicznyleszek:
Kosher cleanup of #2442 👌


jsapp/js/components/permissions/sharingForm.es6
```diff
@@ -35,15 +35,16 @@ class SharingForm extends React.Component {
 
   componentDidMount () {
     this.listenTo(stores.asset, this.onAssetChange);
-    this.listenTo(actions.permissions.getAssetPermissions.completed, this.onGetAssetPermissionsCompleted);
+    this.listenTo(actions.permissions.bulkSetAssetPermissions.completed, this.onAssetPermissionsUpdated);
+    this.listenTo(actions.permissions.getAssetPermissions.completed, this.onAssetPermissionsUpdated);
     this.listenTo(actions.permissions.getCollectionPermissions.completed, this.onGetCollectionPermissionsCompleted);
```
noliveleger: IMHO, I would keep the same name structure for `Collection` too. 
What do you think?

PR #2471, opened 2019-10-23
Update and fix babel building

### LINKED ISSUE #2012: Update and fix babel building
## Description

Updating `babel-loader ^7.1.4 → ^8.0.4` causes build to break, find what's problem and what other things needs updating.

magicznyleszek:
## Description

New babel isn't so forgiving in exports, so it needed quite a few changes.

## Related issues

Fixes #2012


magicznyleszek:
@noliveleger I don't have an opinion on this, I just do what works :-P (:-1: me). After updating one of the dependencies build process started to complain about unnamed default exports, so I changed them. But then some other dependency also got updated and I think it doesn't complain anymore, as I was able to change keymap to unnamed default export 🤷‍♀ I read about some reasoning as to why it is potentially dangerous to use unnamed default exports, but don't remember what it was exactly - something about overwriting variables…

PR #2483, opened 2019-10-29
Back button in settings

### LINKED ISSUE #2479: Allow users to return to previous view from user settings
## Description

As a user, I would like to return to my previous view after viewing or making changes to my account settings. 

At the moment, after opening user settings there is no way to get back to the same view (other than hitting the browser Back button). 

## UX suggestion

1. User opens settings
2. User clicks on x button to close settings
3. User returns to same view as before (1)

## Design suggestion

image

magicznyleszek:
## Description

Close (back) button with safety confirm modal if changes were made - prevents navigating out of settings (back button clicked, closing tab, changing route).

## Related issues

Fixes #2479


noliveleger:
@magicznyleszek. Everything works as expected. 
On my browsers (chrome, FF & Safari) the cross is a little bit misaligned though (according to other views).

Screen Shot 2019-11-19 at 09 51 54

Either, the cross could be center horizontally by moving it to the right:
Screen Shot 2019-11-19 at 09 49 35

or the column could be narrowed

Screen Shot 2019-11-19 at 09 44 25

What do you think?

magicznyleszek:
@noliveleger you're right - I fixed the alignment and allowed myself to merge it :)

PR #2484, opened 2019-10-30
Introduce source maps and vendor splitting

### LINKED ISSUE #2469: Make source maps available (publicly?)
It would certainly be nice to see something like
```javascript
actions.resources.loadAsset.listen(function(params){
  var dispatchMethodName;
  if (params.url) {
    dispatchMethodName = params.url.indexOf('collections') === -1 ?
        'getAsset' : 'getCollection';
  } else {
    dispatchMethodName = {
      c: 'getCollection',
      a: 'getAsset'
    }[params.id[0]];
  }

  dataInterface[dispatchMethodName](params)
    .done(actions.resources.loadAsset.completed)
    .fail(actions.resources.loadAsset.failed);
});
```
in Sentry instead of what it currently displays, which is this variety of nasty:
```javascript
{snip} sset":"getCollection":{c:"getCollection",a:"getAsset"}[e.id[0]],s.dataInterface[t](e).done(m.resources.loadAsset.completed).fail(m.resources {snip}
```
It should be possible to get better stack traces by connecting Sentry to Webpack's source maps: https://docs.sentry.io/platforms/javascript/sourcemaps/#webpack.

While we're at it, I'm not sure why we shouldn't expose a source map to the public, on our production servers, to ease debugging and get more useful details from users' screenshots of errors in the developer console. [Webpack says](https://webpack.js.org/configuration/devtool/#production),

> You should configure your server to disallow access to the Source Map file for normal users!

…but why? We're an open-source project. What's the downside?

magicznyleszek:
## Description

This adds source map (for sentry errors and debugging on production). It also splits out vendor code from THE JS file - now we will have two: `app.js` and `vendors.js` :)

See the Sentry error example: https://sentry.kbtdev.org/kobo/kpi-frontend/issues/305702/?query=is:unresolved%20url:https://kf.kvar.kbtdev.org/

## Related issues

Fixes #2469

### DIFF
--- a/jsapp/js/components/modal.es6
+++ b/jsapp/js/components/modal.es6
@@ -66,6 +66,9 @@ class Modal extends React.Component {
 
       case MODAL_TYPES.NEW_FORM:
         // title is set by formEditors
+        if (window.Raven) {
+          window.Raven.captureException(new Error('Testing source maps!'));
+        }
         break;
 
       case MODAL_TYPES.ENKETO_PREVIEW:
--- a/jsapp/js/components/modal.es6
+++ b/jsapp/js/components/modal.es6
@@ -66,9 +66,6 @@ class Modal extends React.Component {
 
       case MODAL_TYPES.NEW_FORM:
         // title is set by formEditors
-        if (window.Raven) {
-          window.Raven.captureException(new Error('Testing source maps!'));
-        }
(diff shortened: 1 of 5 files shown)


jnm:
Quite nice :kissing_heart:

Issue #2332, opened 2019-07-30
Add `collection` as a new `asset_type`; remove `Collection` model


jnm:
I thought I had an issue opened for this already, but I can't find it! Needed for https://github.com/kobotoolbox/kpi/issues/2293.

jnm:
Closed by #2570

PR #2496, opened 2019-11-05
Validate backend perms

### LINKED ISSUE #2343: Show warning message if backend sends permissions that violate validity rules
Context: https://github.com/kobotoolbox/kpi/pull/2289#discussion_r307943838

> Then if data is invalid, I would display error notification to inform that BE has wrong data and to ask for user to either contact admin or remove and add permissions again?

Yes, this sounds good. :star: Please add something like "The stored permissions are invalid. Please assign them again. If this problem persists, contact help@kobotoolbox.org" (maybe using `replaceSupportEmail()`). I intentionally mangled stuff on the backend as a test; the backend has its own checks against contradictory permissions, so something like this should be rare. It's just a high-stakes kind of rare since there's the potential for unintentional access being granted, so I think having the warning is worthwhile.

> Messing with the applyValidityRules setting and disabling checkboxes would be confusing I think…
OK, let's not do that. The warning alone is fine.

> The real question here is - should FE be prepared for getting wrong data from BE? Because we don't do it anywhere else.

Permissions are more delicate than other things, and determining in other contexts whether data sent by the BE is actually valid may be difficult or impossible without duplicating lots of code. For less-sensitive places where the BE sends data so bad that the FE cannot render properly, throwing an exception (that's logged by Sentry) and displaying an error is enough.

> As for "Edit submissions" label unfortunately it is hardcoded in FE, as all the other labels in the UPDATE PERMISSIONS form. It has to be this way, because it is not possible to render the permissions 1:1 from BE data due to the partial permissions.

I'd have to re-read the partial permissions code to fully understand, but let's leave it for a future PR. Thanks for using "Change submissions".

_Originally posted by @jnm in https://github.com/_render_node/MDE3OlB1bGxSZXF1ZXN0UmV2aWV3MjY3NDI1NDcw/pull_request_reviews/more_threads_ (GitHub "Reference in new is…" feature doesn't work? This link is dead; added a manual link at the top.)

magicznyleszek:
## Description

Validates if permissions send from backend are valid with rules send from backend. I assume #2332 will be done soon, so I did not add a special code for handling collections.

## Related issues

Fixes #2343


jsapp/js/components/permissions/permValidator.es6
```diff
@@ -0,0 +1,70 @@
...
+  constructor(props){
+    super(props);
+    autoBind(this);
+  }
+
+  componentDidMount() {
+    this.listenTo(actions.permissions.getAssetPermissions.completed, this.validateBackendData);
+  }
+
+  validateBackendData(permissionAssignments) {
```
jnm: I'm not sure this is being done per user(?)
I tested by taking a form owned by user Apple and assigning user Zebra permission to edit the form. Then, I hacked the backend to remove Zebra's permission to view the form, but the `INVALID_PERMS_ERROR` didn't appear. At first glance, I think Apple's view form permission is satisfying the `hasAllImplied` check on Zebra's edit permission.

PR #2500, opened 2019-11-09
2173 show map with no data

### LINKED ISSUE #2173: Map shows error when no GPS data has been collected yet
## Description

A deployed project that has a GPS (`geopoint`) question but hasn't yet received any data for that question is causing an error to be displayed instead of the map. Identified [here](https://www.flowdock.com/app/kobotoolbox/kobo/threads/TgzQafedPzUyro6Q0SHNOPVRwF5).

## Steps to Reproduce

1. Create new form with `geopoint` question
2. Deploy and submit data (but don't add any gps points)
3. Open map

## Expected behavior 

Map is displayed but without any data on it. 

## Actual behavior 

Instead, an error is displayed: 
image

## Additional details

The map is also not displayed when the form doesn't have a `geopoint` question. It's a separate issue, but it might be nice to just always display the map either way.

duvld:
## Description

Added more verbose error messages if user clicks on 'map' in Data tab with no GeoPoint data. Also displays entire map regardless of data but dims the map and shows error text over the map.

## Related issues

Fixes #2173


jsapp/scss/components/_kobo.map.scss
```diff
@@ -399,6 +399,28 @@ body .leaflet-control-layers,
...
+  left: 50%;
+  transform: translate(-50%, -50%);
+  width:100%;
+}
+.map-transparent-background{
+  height: 100%;
+  position: absolute;
+  width: 100%;
+  background-color: grey;
+  z-index: 401;
```
magicznyleszek: I think this should be `z-index: 1001;` - so the message covers all the UI. If there is no data, there is no point in user interacting with it.

magicznyleszek: Then you will need to add `z-index: 0;` to `.form-view--map`, so the background is contained within the map div.

jsapp/scss/components/_kobo.map.scss
```diff
@@ -399,6 +399,28 @@ body .leaflet-control-layers,
   }
 }
 
+.map-no-geopoint{
+  text-align:center;
+  margin:auto;
+  padding:10px;
```
magicznyleszek: Both margin and padding are not needed here.

jsapp/scss/components/_kobo.map.scss
```diff
@@ -399,6 +399,28 @@ body .leaflet-control-layers,
...
+  font-weight:bold;
+}
+.map-no-geopoint-wrapper{
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
```
magicznyleszek: Here you can drop `left` and change `transform` to `transform: translateY(-50%);` - the text is already centered horizontally using `text-align: center`

PR #2504, opened 2019-11-13
Django 2 Upgrade (part 2): Drop Python 2 support

### LINKED ISSUE #2489: Drop Python 2 support

noliveleger:
It contains PR #2503 changes, so PR #2503 should reviewed first.

Closes #2489


kpi/fields/generic_hyperlinked_related.py
```diff
@@ -19,14 +20,14 @@ def __init__(self, **kwargs):
         # situation. We will override them dynamically.
         kwargs['view_name'] = '*'
         kwargs['queryset'] = ObjectPermission.objects.none()
-        return super(GenericHyperlinkedRelatedField, self).__init__(**kwargs)
+        # ToDo verify why return in __init__
+        super().__init__(**kwargs)
```
jnm: Looks like the interpreter is explicit in Python 2 and 3:
```
TypeError: __init__() should return None, not [blah]
```
This would fail if somehow the ancestor's initializer returned something other than `None`. We should remove `return`s from all `__init__()`s.

This was probably me not being attentive to the difference between [`__new__()`](https://docs.python.org/3/reference/datamodel.html#object.__new__) and [`__init__()`](https://docs.python.org/3/reference/datamodel.html#object.__init__).

hub/admin.py
```diff
@@ -29,7 +29,7 @@ def get_actions(self, request):
         almost always fail
         """
 
-        actions = super(UserDeleteKludgeAdmin, self).get_actions(request)
+        actions = super().get_actions(request)
```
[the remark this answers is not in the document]
jnm: https://github.com/kobotoolbox/kpi/issues/2509

kpi/fields/generic_hyperlinked_related.py
jnm: https://github.com/kobotoolbox/kpi/issues/2510

PR #2507, opened 2019-11-15
Django 2 Upgrade (part 3): Force explicitly "on_delete" for foreign keys.

noliveleger:
This PR contains only changes related to required argument `on_delete` for `ForeignKeys`. 
The default behaviour in Django 1.x was on `CASCADE`.

On top of #2504


jnm:
Great, so this just continues the same behavior that we had under Django 1.8: https://docs.djangoproject.com/en/1.8/ref/models/fields/#django.db.models.ForeignKey.on_delete
> When an object referenced by a ForeignKey is deleted, Django by default emulates the behavior of the SQL constraint ON DELETE CASCADE and also deletes the object containing the ForeignKey. This behavior can be overridden by specifying the on_delete argument.

PR #2511, opened 2019-11-18
Invalidate sessions on password change

### LINKED ISSUE #2258: Log out user after updating password

noliveleger:
It makes all (other) sessions invalid when users change their password. 

Closes #2258

### DIFF
--- a/kobo/settings/base.py
+++ b/kobo/settings/base.py
@@ -105,9 +105,6 @@ MIDDLEWARE = [
     'django.middleware.common.CommonMiddleware',
     'django.middleware.csrf.CsrfViewMiddleware',
     'django.contrib.auth.middleware.AuthenticationMiddleware',
-    # TODO: Uncomment this when interoperability with dkobo is no longer
-    # needed. See https://code.djangoproject.com/ticket/21649
-    # 'django.contrib.auth.middleware.SessionAuthenticationMiddleware',
     'django.contrib.messages.middleware.MessageMiddleware',
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
     'hub.middleware.UsernameInResponseHeaderMiddleware',
--- a/kpi/serializers/current_user.py
+++ b/kpi/serializers/current_user.py
@@ -2,6 +2,7 @@
 import datetime
 import pytz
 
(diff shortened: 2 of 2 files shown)


jnm (changes requested):
@noliveleger, how hard would it be to log out KC sessions when using a password reset email? ~~I'll send you a screencast on Flowdock since GitHub won't let me upload a video here :(~~

haha, it's 2021! have a video!

https://user-images.githubusercontent.com/2085013/103964666-02a86a00-512a-11eb-803e-f14c5aea1c56.mp4

noliveleger:
@jnm, good catch. I don't think it's hard. We may need to implement our own PasswordReset view to call `update_session_auth_hash()`

jnm:
@noliveleger, that's good, but I think it only handles regular password changes, not resets done with an emailed token. The emailing is handled by django-registration-redux (which I think has been renamed back to django-registration):
https://github.com/macropin/django-registration/blob/d0b4cce3f5f0e29a837e92fc86cc6866085b2668/registration/auth_urls.py#L55-L58
…which in turn uses `django.contrib.auth.PasswordResetConfirmView`:
https://github.com/django/django/blob/2.2/django/contrib/auth/views.py#L298-L303

Did Django miss a place where they should be calling `update_session_auth_hash()`? I'll have to test again, because I thought it _was_ logging out my other KPI sessions.

jnm:
Moved the above to https://github.com/kobotoolbox/kpi/issues/2595

PR #2516, opened 2019-11-23
Sync correctly KoBoCat permissions on bulk permission assignments

noliveleger:
KoBoCat internal permission assignment is also exclude from permission assignments list.

### DIFF
--- a/kpi/views/v2/asset_permission_assignment.py
+++ b/kpi/views/v2/asset_permission_assignment.py
@@ -182,13 +182,9 @@ class AssetPermissionAssignmentViewSet(AssetNestedObjectViewsetMixin,
             # If something fails later, this query should rollback
             perms_to_delete = self.asset.permissions.exclude(
                 user__username=self.asset.owner.username)
-
-            # Delete all related KoBoCat permissions as well
             for perm in perms_to_delete.all():
-                remove_applicable_kc_permissions(self.asset,
-                                                 perm.user,
-                                                 perm.permission.codename)
-            perms_to_delete.delete()
+                self.asset.remove_perm(perm.user,
+                                       perm.permission.codename)
 
             for assignment in assignments:
                 context_ = dict(self.get_serializer_context())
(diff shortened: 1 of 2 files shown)


kpi/views/v2/asset_permission_assignment.py
```diff
@@ -172,12 +180,22 @@ def bulk_assignments(self, request, *args, **kwargs):
...
-            self.asset.permissions.exclude(user__username=self.asset.owner.username).delete()
+            perms_to_delete = self.asset.permissions.exclude(
+                user__username=self.asset.owner.username)
+
+            # Delete all related KoBoCat permissions as well
+            for perm in perms_to_delete.all():
+                remove_applicable_kc_permissions(self.asset,
+                                                 perm.user,
+                                                 perm.permission.codename)
+            perms_to_delete.delete()
```
jnm: This seems fine, but I do wonder why `delete()` is used here when [AssetBulkInsertPermissionSerializer](https://github.com/kobotoolbox/kpi/blob/ac6ac5444cb978dc4131e5a70e18ecec7c061f0b/kpi/views/v2/asset_permission_assignment.py#L199-L202) uses `assign_perm()`:
https://github.com/kobotoolbox/kpi/blob/ac6ac5444cb978dc4131e5a70e18ecec7c061f0b/kpi/serializers/v2/asset_permission_assignment.py#L55

I realize that `delete()` should be faster, but if we need to iterate over the permissions before deleting them anyway, maybe it'd be better to let `remove_perm()` handle the job.

noliveleger: Exact, I used the `.delete()` to generate only one query instead of as many as there are permission assignments. 

Pros: 
- Only place to change code (inside `remove_perm()`) if we change the deployment backend. 

Cons:
- As I said, more generated queries instead of only one (regarding KPI `ObjectPermission` objects deletion)

Too bad that KoBoCat does not support bulk delete permissions.
I think that most of the time, the performance gain is not worth compared to code readibility. Let's change it.

jnm:
Cherry-picked these changes into new tag `two-databases_2.019.48`:
https://github.com/kobotoolbox/kpi/compare/two-databases_2.019.42...two-databases_2.019.48

PR #2519, opened 2019-11-26
[Feature] Add `delete_submissions` permission

### LINKED ISSUE #2282: Mirror KC behavior for delete-submission permission
In KoBoCAT, if you have permission to edit submissions, you can delete them as well. By contrast, KPI only assigns `delete_submissions` to the _owner_, as a calculated permission that is not stored in the database. **After** completing https://github.com/kobotoolbox/kpi/issues/2281, we should:

1. Change the calculation logic to include `delete_submissions` in the list of permissions for anyone who has `change_submissions`;
1. Change the KPI submissions API to check for `delete_submissions` instead of `change_submissions`. See here the test that needs to be updated: https://github.com/kobotoolbox/kpi/blob/4616238aeb1dea7a5ec47bba3af4e83a4dc4827e/kpi/tests/test_api_submissions.py#L184-L192

### LINKED ISSUE #2488: Create new 'edit without deleting' permission
As a survey admin, I  would like to allow one of my team members to edit submissions but not delete them so that I don't have to worry about accidental data loss. 

At the moment (after #2367) any user with 'edit submission' rights can also delete submissions.

noliveleger:
Closes #2488 
Closes #2282

Related to https://github.com/kobotoolbox/kobocat/pull/591 and must be deployed at the same time 

@jnm, @magicznyleszek, both of you are assigned as reviewers because it touches frontend and backend

Management command **assign_delete_submissions_to_existing_users_with_change_submissions** must be run on existing database


kpi/management/commands/assign_delete_submissions_to_existing_users_with_change_submissions.py
```diff
@@ -0,0 +1,72 @@
...
+            permission__codename=PERM_CHANGE_SUBMISSIONS,
+            date_created__lte=applied_date
+        ).select_related('user', 'permission')
+
+        records_count = query.count()
+        cpt = 1
+        if verbosity >= 1:
+            self.stdout.write(f'Found {records_count} permission assignments')
+
+        for record in query.all():
```
jnm, line 46: We don't need the manual batching stuff of yore, but I think we still need `.iterator()` to avoid loading all the ~~objects~~ results of the Postgres query into memory at once

kpi/management/commands/assign_delete_submissions_to_existing_users_with_change_submissions.py
```diff
@@ -0,0 +1,72 @@
...
+        except MigrationRecorder.Migration.DoesNotExist:
+            self.stdout.write(f'You must run migration `{migration_name}` first')
+            sys.exit(1)
+
+        applied_date = expected_migration.applied
+        asset_content_type_id = ContentType.objects.get_for_model(Asset).pk
+        query = ObjectPermission.objects.filter(
+            content_type_id=asset_content_type_id,
+            permission__codename=PERM_CHANGE_SUBMISSIONS,
+            date_created__lte=applied_date
```
jnm, line 38: I'm not really sure we need this—provided I'm understanding it correctly. If `PERM_CHANGE_SUBMISSIONS` has allowed deletion for years, the lag time between the migration being applied and the completion of this management command isn't going to make much of a difference

kpi/management/commands/assign_delete_submissions_to_existing_users_with_change_submissions.py
```diff
@@ -0,0 +1,72 @@
...
+            try:
+                asset = Asset.objects.get(pk=record.object_id)
+            except Asset.DoesNotExist:
+                asset_id = record.object_id
+                if verbosity >= 2:
+                    sys.stderr.write(f'\n\tAsset with id {asset_id} could not be found')
+                cpt += 1
+                continue
+
+            asset.assign_perm(record.user, PERM_DELETE_SUBMISSIONS)
```
jnm, line 56: It could be a denial. We may be better off copying `change_asset` to `delete_asset` at a lower level, using something like `bulk_create()`, and preserving the `inherited` and `deny` flags. That should save time, too, since we wouldn't need to instantiate each asset.

kpi/models/asset.py
```diff
@@ -596,23 +596,28 @@ class Meta:
...
 
     # Some permissions must be copied to KC
-    KC_PERMISSIONS_MAP = {  # keys are KC's codenames, values are KPI's
+    KC_PERMISSIONS_MAP = {  # keys are KPI's codenames, values are KC's
         PERM_CHANGE_SUBMISSIONS: 'change_xform',  # "Can Edit" in KC UI
         PERM_VIEW_SUBMISSIONS: 'view_xform',  # "Can View" in KC UI
         PERM_ADD_SUBMISSIONS: 'report_xform',  # "Can submit to" in KC UI
-        PERM_VALIDATE_SUBMISSIONS: 'validate_xform',  # "Can Validate" in KC UI
+        PERM_DELETE_SUBMISSIONS: 'delete_data_xform',  # "Can delete submissions" in KC UI
+        PERM_VALIDATE_SUBMISSIONS: 'validate_xform',  # "Can validate submissions" in KC UI
```
jnm, line 619: I think "Can Validate" was correct, and the delete permission is called "Can Delete Data" in the KoBoCAT UI:
https://github.com/kobotoolbox/kobocat/blob/02a331a15da3ee2ff00a697a19e54fd2587a51ff/onadata/apps/main/views.py#L391-L395

kpi/models/object_permission.py
```diff
@@ -830,7 +832,7 @@ def assign_perm(self, user_obj, perm, deny=False, defer_recalc=False,
         for implied_perm in implied_perms:
             self.assign_perm(
                 user_obj, implied_perm, deny=deny, defer_recalc=True)
-        # We might have been called by ourself to assign a related
+        # We might have been called by ourselves to assign a related
```
jnm, line 835: Well, yes, this was weirdly written, but: https://www.merriam-webster.com/dictionary/ourself
Then, from there, you can waste tons of time reading people argue about the use of "royal we" in technical and scientific writing :zany_face:

kpi/models/object_permission.py
```diff
@@ -918,7 +920,7 @@ def has_perm(self, user_obj, perm):
 
     @transaction.atomic
     def remove_perm(self, user_obj, perm, defer_recalc=False, skip_kc=False):
-        r"""
+        """
```
jnm, line 923: I don't know why this was a raw string if it didn't have any backslashes, but just in case your editor or linter is doing this blindly, `r"""` is allowed under some (different) circumstances by PEP 257:
> Use r"""raw triple double quotes""" if you use any backslashes in your docstrings.

kpi/serializers/v2/asset_permission_assignment.py
```diff
@@ -52,6 +52,11 @@ def create(self, validated_data):
                 'user': "Owner's permissions cannot be assigned explicitly"})
         permission = validated_data['permission']
         partial_permissions = validated_data.get('partial_permissions', None)
+
+        bulk = self.context.get('bulk', False)
+        if bulk is False and asset.has_deployment:
+            asset.deployment.remove_from_flag(user_id=user.pk)
```
jnm, line 58: I guess there's another place where the flag is removed for bulk assignments (verify)

jnm: OK, it's handled by `AssetPermissionAssignmentViewSet.bulk_assignments()`. May I request a comment here mentioning that :)

kpi/views/v2/asset_permission_assignment.py
```diff
@@ -178,7 +178,11 @@ def bulk_assignments(self, request, *args, **kwargs):
         # one assignment fails.
         with transaction.atomic():
 
-            # First delete all assignments before assigning new ones.
+            # First, delete deployment backend flags if any
+            if self.asset.has_deployment:
+                self.asset.deployment.remove_from_flag()
```
jnm, line 183: Wouldn't `self.asset.permissions.exclude(user__username=self.asset.owner.username)` below include `PERM_FROM_KC_ONLY`?

jnm (changes requested):

kpi/deployment_backends/base_backend.py
```diff
@@ -26,6 +26,9 @@ def store_data(self, vals=None):
     def delete(self):
         self.asset._deployment_data.clear()
 
+    def remove_from_backend_flag(self, user_id):
```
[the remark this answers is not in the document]
noliveleger: 🤦

kpi/deployment_backends/kobocat_backend.py
```diff
@@ -255,6 +261,28 @@ def connect(self, identifier=None, active=False):
...
+        #
+        # This method is here instead of `ObjectPermissionMixin` because
+        # it's specific to KoBoCat as backend.
+        filters = {
+            'permission__codename': PERM_FROM_KC_ONLY,
+            'object_id': self.asset.id,
+            'content_type': ContentType.objects.get_for_model(self.asset)
+        }
+        if user_id is not None:
+            filters['user_id'] = user_id
```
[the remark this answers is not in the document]
noliveleger: Yes! it's useful for bulk assignements.

kpi/management/commands/assign_delete_submissions_to_existing_users_with_change_submissions.py
```diff
@@ -0,0 +1,72 @@
...
+class Command(BaseCommand):
+
+    help = 'Assign `delete_submissions` permission to users ' \
+           'who already had `change_submissions` prior to migration ' \
+           '`0024_add_date_created_to_object_permission`'
+
+    def handle(self, *args, **options):
+
+        verbosity = options['verbosity']
+        self.stdout.write(str(verbosity))
```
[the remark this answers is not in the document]
noliveleger: 🤦 , no! Forgotten debug line!

kpi/management/commands/assign_delete_submissions_to_existing_users_with_change_submissions.py
noliveleger: I wanted the cover the case where: 
- owners see the "delete submission" permission and do not give it to some co-workers on purpose
- the person  in charge of KoBoToolbox upgrades on their server forgot to run the management command at the same time of the migration but runs 

But I hear you. It does not hurt but it may be overkill indeed. Should I remove it?

kpi/models/object_permission.py
noliveleger: My bad! I thought this comment was created by one of my fellow countrymen and I thought it was a mistake. 
I'll revert it back.

kpi/models/object_permission.py
noliveleger: 👍

kpi/deployment_backends/kobocat_backend.py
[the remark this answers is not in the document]
jnm: :ok_hand:

kpi/models/object_permission.py
jnm: I think there are just as many (or more) arguments for `ourself` being wrong here as right :joy:

kpi/views/v2/asset_permission_assignment.py
jnm: Yes, it would, but it would also make `remove_perm()` fail. We have to trash `from_kc_only` separately beforehand

kpi/migrations/0025_add_date_created_to_object_permission.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ('kpi', '0024_alter_jsonfield_to_jsonbfield'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='objectpermission',
```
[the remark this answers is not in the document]
jnm: Performance seems acceptable now using `bulk_create()` :crossed_fingers:

PR #2525, opened 2019-12-03
[Public collections] Add public collection related details to asset endpoints

### LINKED ISSUE #2520: Add public-collection-related details to asset list endpoint
To support #2293, we need the asset list to return:

* list of languages (I guess in the [`Full Name (standardized code)` format](http://xlsform.org/en/#multiple-language-support), but verify with Leszek)
* number of subscribers (can just be zero for non-collection assets)
* status - four possible values:
    * not shared
    * shared with specific users
    * shared with everyone (`AnonymousUser` has `view_asset`)
    * discoverable by everyone (`AnonymousUser` has `discover_asset`; implies that `AnonymousUser` also has `view_asset`)

This should be built on top of either https://github.com/kobotoolbox/kpi/tree/2332-collection-as-asset-type or Leszek's branch (ask him) https://github.com/kobotoolbox/kpi/tree/2293%2B2332-public-collections-as-asset.

image

### LINKED ISSUE #2541: Collection children count in assets list
## Description

The `/api/v2/assets/` endpoint should have a children count for `asset_type:collection`, to be displayed in a listing. I only need a number, not a list of children - if this is any help.

## Additional details

Requirement for #2293

### LINKED ISSUE #2544: Collection `access_type` has been lost after merging `Collection` model into `Asset`
Bring this back:
https://github.com/kobotoolbox/kpi/blob/1c2dec92faf4cbbda88092f5b7c97d5388dd0dcd/kpi/serializers/v2/collection.py#L59

It's missing from https://github.com/kobotoolbox/kpi/tree/2332-collection-as-asset-type.

This is needed so that the frontend can tell which collections have been subscribed to (part of https://github.com/kobotoolbox/kpi/issues/2293).

noliveleger:
Closes #2520 
Closes #2544 
Closes #2541 

**N.B.:**
A new management command has been created to update collection languages. It should be run against existing databases.
`./manage.py update_collection_languages`


kpi/management/commands/update_collection_languages.py
```diff
@@ -0,0 +1,53 @@
...
+        max_pk = 0
+        stop = False
+        cpt = 0
+
+        while not stop:
+            query = query.filter(pk__gt=max_pk).order_by('pk')
+            if query.count() == 0:
+                stop = True
+            else:
+                for collection in query.all()[:chunks]:
```
jnm, line 39: I think now that you've upgraded Django :bow: it uses server-side cursors by default, and this type of chunking isn't needed anymore. Just a note.

kpi/models/asset.py
```diff
@@ -912,17 +980,66 @@ def to_clone_dict(self, version_uid=None, version=None):
...
+        if children:
+            summaries = [child.summary for child in children]
+            languages = set(obj_languages)
+        else:
+            # ToDo validate whether all children should be included
+            # (e.g. archives?)
+            summaries = self.children.values_list('summary', flat=True)
+
+        for summary in summaries:
+            if isinstance(summary, str):
```
jnm, line 1010: we could add a TODO to simplify this once #2635 is done (or add a note in #2635 to remind us to come back here)

jnm: (is `summary__languages` allowed with JSONB?)

kpi/models/asset.py
```diff
@@ -912,17 +980,66 @@ def to_clone_dict(self, version_uid=None, version=None):
...
+                try:
+                    summary = json.loads(summary)
+                except ValueError:
+                    continue
+            child_languages = [language
+                               for language in summary.get('languages', [])
+                               if language is not None]
+
+            if child_languages:
+                languages = set(list(languages) + child_languages)
```
jnm, line 1020: could be simplified to `languages.update(child_languages)`

kpi/models/asset.py
```diff
@@ -912,17 +980,66 @@ def to_clone_dict(self, version_uid=None, version=None):
...
+
+        languages = list(languages)
+        # If languages are still the same, no needs to update the object
+        if obj_languages == languages:
```
jnm, line 1024: probably should sort both of these before comparing

kpi/serializers/v2/asset.py
```diff
@@ -304,6 +324,23 @@ def get_assignable_permissions(self, asset):
             }
             for codename in asset.ASSIGNABLE_PERMISSIONS_BY_TYPE[asset.asset_type]]
 
+    def get_subscribers_count(self, asset):
+        if asset.asset_type != ASSET_TYPE_COLLECTION:
+            return 0
+
+        return UserAssetSubscription.objects.filter(asset_id=asset.pk).count()
```
jnm, line 331: if this feature becomes popular, we might need to cache this, perhaps inside `summary`

kpi/serializers/v2/asset.py
```diff
@@ -357,33 +456,91 @@ class Meta(AssetSerializer.Meta):
...
-    def get_permissions(self, obj):
+    def get_children(self, asset):
+        if asset.asset_type != ASSET_TYPE_COLLECTION:
+            return {'count': 0}
+
+        try:
+            children_count = self.context['children_count_per_asset'].get(asset.pk, 0)
+        except KeyError:
+            # Maybe overkill, there are no reasons to enter here.
+            # in the list context, `children_count` should be always
```
jnm, line 473:
```
`children_count_per_asset`
```

kpi/signals.py
```diff
@@ -100,3 +100,11 @@ def post_delete_asset(sender, instance, **kwargs):
...
     ObjectPermission.objects.filter_for_object(instance).delete()
     # No recalculation is necessary since children will also be deleted
+
+    # Update parent's languages if this object is a child of another asset.
+    try:
+        if instance.parent:
+            instance.parent.update_languages()
+    except Asset.DoesNotExist:  # `parent` may exists in DJANGO models cache but not in DB
+        pass
+
```
jnm, line 110: idea to avoid accidentally catching an unexpected exception in `update_languages()`:
```
try:
    parent = instance.parent
except Asset.DoesNotExist:
    pass
else:
    parent.update_languages()
```

jnm: (unless there's a legitimate reason why the exception would not be raised _until_ `update_languages()` runs)

kpi/views/v2/asset.py
```diff
@@ -238,12 +246,37 @@ def get_serializer_context(self):
...
+
+            user_subscriptions_per_asset = defaultdict(int)
+            for record in subscriptions_queryset:
+                user_subscriptions_per_asset[record['asset_id']] = \
+                    record['user_subscriptions']
+
+            context_['user_subscriptions_per_asset'] = user_subscriptions_per_asset
+
+            # 4) Get children count per asset
+            # ToDo Verify if all children must be included in count (e.g. Archives)
```
jnm, line 268: not sure if i commented on this before (i know the same question appears in a different place in this PR), but only deployed assets can be archived, and none of those should be inside a collection (library assets cannot be deployed)

jnm (changes requested):

kpi/management/commands/update_collection_languages.py
noliveleger: I left a comment in the code  to update it with new Django `iterator()`

kpi/model_utils.py
```diff
@@ -79,13 +79,14 @@ def _load_library_content(structure):
                     content=scontent,
                     asset_type='question',
                     owner=structure['owner'],
-                    parent=collection
+                    parent=collection,
+                    update_parent_languages=False,
```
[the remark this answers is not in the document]
noliveleger: `_load_library_content()` is called there 
https://github.com/kobotoolbox/kpi/blob/c6f9e393a30fa7fcb1284212332146fbd9413b1c/kpi/model_utils.py#L131
`create_assets` is called there https://github.com/kobotoolbox/kpi/blob/28be63cab2b5a5aa474cf9b43e7409a49742ca4b/kpi/models/import_export_task.py#L222
which assigns `item._orm` which is saved at
https://github.com/kobotoolbox/kpi/blob/28be63cab2b5a5aa474cf9b43e7409a49742ca4b/kpi/models/import_export_task.py#L253

kpi/models/asset.py
noliveleger: @jnm, I've tried to modified the code with `summary__languages` but it seems to be a bit slower than the current code.

```
        if children:
            languages = set(obj_languages)
            children_languages = [child.summary.get('languages')
                                  for child in children
                                  if child.summary.get('languages')]
        else:
            children_languages = list(self.children
                                      .values_list('summary__languages',
                                                   flat=True)
                                      .exclude(Q(summary__languages=[]) |
                                               Q(summary__languages=[None]))
                                      .order_by())

        # Flatten `children_languages` to 1-dimension list.
        if children_languages:
            languages.update(reduce(add, children_languages))

        languages.discard(None)
```

Fetching a nested property of a `JSONBField` has a cost in performance. 
```
In [15]: Asset.objects.filter(asset_type='collection').values_list('summary', flat=True).all()
Out[15]: SELECT "kpi_asset"."summary"
  FROM "kpi_asset"
 WHERE "kpi_asset"."asset_type" = 'collection'

Execution time: 0.001931s [Database: default]
```

```
In [14]: Asset.objects.filter(asset_type='collection').exclude(Q(summary__languages=[]) | Q(summary__languages=[None])).values_list('summary', flat=True).all()
Out[14]: SELECT "kpi_asset"."summary"
  FROM "kpi_asset"
 WHERE ("kpi_asset"."asset_type" = 'collection' AND NOT ((("kpi_asset"."summary" -> 'languages') = '[]' OR ("kpi_asset"."summary" -> 'languages') = '[null]')))

Execution time: 0.002602s [Database: default]
```

noliveleger: I've run `./manage.py update_collection_languages` on my local env. 
See https://www.flowdock.com/rest/files/58517/chY2WGh8Vc8AB4YvMLe4Ew/update_languages.txt. 

I'll update the code in favor on this new version. It's clearer (IMHO) and the difference is insignificant.

jnm: Cool :sunglasses: thanks for the extensive analysis :muscle:

kpi/model_utils.py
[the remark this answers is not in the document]
jnm: OK! We discussed this and dealt with it in 5b6460b2

PR #2533, opened 2019-12-10
2173 show map with no data new

### LINKED ISSUE #2173: Map shows error when no GPS data has been collected yet

duvld:
New branch created due to the changes to how the frontend environment works, still fixes #2173, plus finished adding requested changes from PR #2500.


magicznyleszek:
Fixed some global `z-index`es conflicts :)

PR #2534, opened 2019-12-10
[Public collections] Replaced 'django.core.exceptions.ValidationError' with 'rest_framework.serializers.ValidationError'

noliveleger:
To return a 400 error instead of 500.


kpi/models/object_permission.py
```diff
@@ -91,9 +95,11 @@ def get_objects_for_user(user, perms, klass=None, all_perms_required=True):
         if '.' in perm:
             new_app_label, codename = perm.split('.', 1)
             if app_label is not None and app_label != new_app_label:
-                raise ValidationError("Given perms must have same app "
-                                      "label (%s != %s)" % (app_label,
-                                                            new_app_label))
+                raise serializers.ValidationError({
```
jnm: also a programmer's error

kpi/models/object_permission.py
```diff
@@ -103,14 +109,18 @@ def get_objects_for_user(user, perms, klass=None, all_perms_required=True):
             new_ctype = ContentType.objects.get(app_label=app_label,
                                                 permission__codename=codename)
             if ctype is not None and ctype != new_ctype:
-                raise ValidationError("Computed ContentTypes do not match "
-                                      "(%s != %s)" % (ctype, new_ctype))
+                raise serializers.ValidationError({
```
jnm: also a programmer's error

kpi/models/object_permission.py
```diff
@@ -103,14 +109,18 @@ def get_objects_for_user(user, perms, klass=None, all_perms_required=True):
...
+                    'permission': 'Computed ContentTypes do not match '
+                                  '(%s != %s)' % (ctype, new_ctype)
+                })
             else:
                 ctype = new_ctype
 
     # Compute queryset and ctype if still missing
     if ctype is None and klass is None:
-        raise ValidationError("Cannot determine content type")
+        raise serializers.ValidationError({
```
jnm: also a programmer's error

kpi/models/object_permission.py
```diff
@@ -119,8 +129,9 @@ def get_objects_for_user(user, perms, klass=None, all_perms_required=True):
     else:
         queryset = _get_queryset(klass)
         if ctype.model_class() != queryset.model:
-            raise ValidationError("Content type for given perms and "
-                                  "klass differs")
+            raise serializers.ValidationError({
```
jnm: also a programmer's error

kpi/models/object_permission.py
```diff
@@ -233,8 +244,10 @@ class Meta:
                                   '__get_all_user_permissions',))
     def save(self, *args, **kwargs):
         if self.permission.content_type_id is not self.content_type_id:
-            raise ValidationError('The content type of the permission does '
-                                  'not match that of the object.')
+            raise serializers.ValidationError({
```
jnm: also a programmer's error

kpi/serializers/current_user.py
```diff
@@ -106,9 +106,10 @@ def update(self, instance, validated_data):
                         'current_password': 'Incorrect current password.'
                     })
         elif any((current_password, new_password)):
-            raise serializers.ValidationError(
-                'current_password and new_password must both be sent ' \
-                'together; one or the other cannot be sent individually.'
-            )
+            raise serializers.ValidationError({
+                'password': 'current_password and new_password must both be '
```
jnm: unless i am not remembering the DRF documentation correctly, keys in this dictionary should be the names of fields. that's not really a perfect fit for this case, because the error deals with two fields at once, but i guess the lesser evil is to pick one field or the other to serve as the dict key

jnm: or move it to a validate method as you mentioned, @noliveleger?

kpi/utils/mongo_helper.py
```diff
@@ -260,7 +260,7 @@ def _get_cursor_and_count(cls, mongo_userform_id, hide_deleted=True,
             if ObjectId.is_valid(query.get('_uuid')):
                 query['_uuid'] = ObjectId(query.get('_uuid'))
             else:
-                raise ValidationError(_('Invalid _uuid specified'))
+                raise ValidationError({'_uuid': _('Invalid')})
```
jnm: would the field in this case be `q`? ideally the validation wouldn't be this deep, but then we'd have to move database-specific code up into the validator so it would know what a good uuid looks like :/

kpi/views/v2/data.py
```diff
@@ -215,8 +215,10 @@ def _get_deployment(self):
         Returns the deployment for the asset specified by the request
         """
         if not self.asset.has_deployment:
-            raise serializers.ValidationError(
-                _('The specified asset has not been deployed'))
+            raise serializers.ValidationError({
+                'deployment': _('The specified asset has not been deployed')
```
jnm: maybe `ValidationError` isn't the best choice? a 400 should be returned if a user tries to access data for a form that hasn't been deployed (i think? another possibility would be to return an empty list).

however, since the user did not POST or PATCH anything, i don't think it makes sense to have this dictionary with field names.

hub/models.py
```diff
@@ -99,9 +100,9 @@ def clean(self):
         try:
             self.user_matches(user, ignore_invalid_queries=False)
         except FieldError as e:
-            raise ValidationError({'user_queries': e.message})
+            raise serializers.ValidationError({'user_queries': e.message})
         except TypeError:
-            raise ValidationError(
+            raise serializers.ValidationError(
```
jnm: looks like the django admin interface won't tolerate this :( it returns a 500 instead of showing the validation message. this isn't (yet) used in any DRF view, so the old way should be ok.

kobo/apps/reports/report_data.py
```diff
@@ -130,10 +130,15 @@ def data_by_identifiers(asset, field_names=None, submission_stream=None,
     if field_names is None:
         field_names = fields_by_name.keys()
     if split_by and (split_by not in fields_by_name):
-        raise serializers.ValidationError(_("`split_by` field '{}' not found.").format(split_by))
+        raise serializers.ValidationError({
+            'data_by_identifiers': _("`split_by` field '{}' not found.").format(split_by)
```
jnm: i don't remember how this gets here, or even if this message would ever appear to a user. if it can't happen directly from user input, then maybe it should be a different exception that returns a 500, so that we're alerted to a programmer's mistake

kobo/apps/reports/report_data.py
```diff
@@ -130,10 +130,15 @@ def data_by_identifiers(asset, field_names=None, submission_stream=None,
...
+        })
     if split_by and (fields_by_name[split_by].data_type != 'select_one'):
-        raise serializers.ValidationError(_("`split_by` field '{}' is not a select one question.").
-                                          format(split_by))
+        raise serializers.ValidationError({
+            'data_by_identifiers':
```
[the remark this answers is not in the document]
noliveleger: Actually it comes from the user, `split_by` is the value passed in the `GET` request as query string. 
e.g : https://kf.kobotoolbox.org/reports/{asset_id}/?names=&split_by={question_name}

kobo/apps/reports/report_data.py
noliveleger: Actually it comes from the user, `split_by` is the value passed in the `GET` request as query string. 
e.g : https://kf.kobotoolbox.org/reports/{asset_id}/?names=&split_by={question_name}

`data_by_identifiers` is not the correct key though, it should be `split_by`

kpi/deployment_backends/kc_access/shadow_models.py
```diff
@@ -243,12 +243,12 @@ def save(self, *args, **kwargs):
         content_type = KobocatContentType.objects.get_for_model(
             self.content_object)
         if content_type != self.permission.content_type:
-            raise ValidationError(
-                "Cannot persist permission not designed for this "
-                "class (permission's type is %r and object's type is "
-                "%r)"
-                % (self.permission.content_type, content_type)
-            )
+            raise serializers.ValidationError({
```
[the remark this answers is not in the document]
noliveleger: I've created a new `BadContentTypeException` in `kpi.exceptions.py`

jnm (approved):

PR #2540, opened 2019-12-19
Added username specification to password reset form

### LINKED ISSUE #2473: Improve password reset screen
## Description

As a user with many user accounts, I would like to be able to reset the password for **just one account at a time**. 

At the moment, resetting a password when not signed in sends one email for each user account, so users with many accounts can receive dozens of emails. 

## Design suggestion

image

duvld:
Fixes #2473 and merges branch 2473-better-password-reset-screen branch. 

Changes:

Before, the password reset form has only an email field and would send an email for every account created under the same email. 

Now, there is an optional Username field in the password reset form that would send only one email regarding the password resetting of that account. If left empty the form would behave as it did before.


duvld:
Cleaning up old PRs, if we want to revisit this issue (maybe with designs) we have the code here

PR #2553, opened 2020-01-14
2523 media links in table view

### LINKED ISSUE #2523: Include media links in table view
## Description

As a survey administrator, I want to view my photos and listen to recordings directly from the web table view so I can analyze and verify collected data more quickly. 

At the moment KPI allows accessing media from the single submission view, but not the table view. 

## Steps to Reproduce

1. Create project with audio and photo questions
2. Collect sample data
3. Open the table view

## Expected behavior 

User can click on any of the file names and view/listen to/watch any attachments in a new tab.

## Additional details

This is already doable in the single submission view and should function the same way: 

image

duvld:
## Description

Fixed #2523, adds clickable link in table view for all media type files in form (images, audio, video).

Refactoring note: This change uses very similar code to https://github.com/kobotoolbox/kpi/blob/1c2dec92faf4cbbda88092f5b7c97d5388dd0dcd/jsapp/js/components/modalForms/submission.es6#L156-L183 inside of https://github.com/kobotoolbox/kpi/blob/5f97496c8e5a6c37d31ee4274f2002066df3c07e/jsapp/js/components/table.es6#L985-L1002 

The returns and little pieces of logic are different between the two functions, not sure if there's a clean way to refactor both of these out to an external file.

### DIFF
--- a/jsapp/js/components/table.es6
+++ b/jsapp/js/components/table.es6
@@ -753,7 +753,6 @@ export class DataTable extends React.Component {
   }
   bulkSelectAllRows(isChecked) {
     let s = this.state.selectedRows;
-    let ind = 0;
     this.state.tableData.forEach(function(r) {
       if (isChecked) {
         s[r._id] = true;
--- a/jsapp/js/components/table.es6
+++ b/jsapp/js/components/table.es6
@@ -413,6 +413,9 @@ export class DataTable extends React.Component {
 
               if (q.type == 'audio') {
                 var kc_server = document.createElement('a');
+                //var deployment_id = this.props.asset.deployment__identifier;
+                console.log('asset: ' + this.props.asset);
+                console.log('hello????');
                 //kc_server.href = this.props.asset.deployment__identifier;
                 // // if this has more components than /{username}/forms/{uid}, it's safe to assume kobocat is running under a
                 // // KOBOCAT_ROOT_URI_PREFIX
--- a/jsapp/js/components/table.es6
+++ b/jsapp/js/components/table.es6
@@ -984,8 +984,8 @@ export class DataTable extends React.Component {
   }
   getMediaDownloadLink(fileName) {
     var attachmentUrl = null;
-    this.state.tableData.some(function(a) {
-        a._attachments.some(function(b) {
+    this.state.tableData.forEach(function(a) {
+        a._attachments.forEach(function(b) {
           if (b.filename.includes(fileName)) {
             fileName = b.filename;
           }
--- a/jsapp/js/components/table.es6
+++ b/jsapp/js/components/table.es6
@@ -405,6 +405,25 @@ export class DataTable extends React.Component {
         filterable: false,
         Cell: row => {
             if (showLabels && q && q.type && row.value) {
+              console.log('q: ' + q + ' q.type: ' + q.type + ' row.value: ' + row.value);
+
+              if (q.type == 'image') {
+                return <a href="https://google.com">hello</a>;
+              }
+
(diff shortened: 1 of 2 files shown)


jsapp/js/components/table.es6
```diff
@@ -405,6 +404,10 @@ export class DataTable extends React.Component {
         filterable: false,
         Cell: row => {
             if (showLabels && q && q.type && row.value) {
+              if (q.type == 'image' || q.type == 'audio' || q.type == 'video') {
```
magicznyleszek: There is a list of question types in `constants.es6`, please use the values from there instead of strings (e.g. `q.type === QUESTION_TYPES.get('audio').id`. Also please use triple equals instead of double :)

jsapp/js/components/table.es6
```diff
@@ -729,6 +732,12 @@ export class DataTable extends React.Component {
     this.fetchData(this.state.fetchState, this.state.fetchInstance);
     this.setState({ promptRefresh: false });
   }
+
+  printAssets(){
```
magicznyleszek: Please delete this function

jsapp/js/components/table.es6
```diff
@@ -748,14 +757,19 @@ export class DataTable extends React.Component {
   }
   bulkSelectAllRows(isChecked) {
     let s = this.state.selectedRows;
-
+    let ind = 0;
```
magicznyleszek: Please delete this leftover line

jsapp/js/components/table.es6
```diff
@@ -748,14 +757,19 @@ export class DataTable extends React.Component {
...
     this.state.tableData.forEach(function(r) {
+          console.log(r);
```
magicznyleszek: Please delete this leftover line

jsapp/js/components/table.es6
```diff
@@ -748,14 +757,19 @@ export class DataTable extends React.Component {
...
       if (isChecked) {
         s[r._id] = true;
       } else {
         delete s[r._id];
       }
-    });
+    },
+
+    //console.log(this.state.tableData.valueOf())
+    console.log(this.state.tableData[ind]),
```
magicznyleszek: Please delete these leftover lines

jsapp/js/components/table.es6
```diff
@@ -774,6 +788,16 @@ export class DataTable extends React.Component {
     }
 
   }
+
+  addLinksToFileNames() {
```
magicznyleszek: Please delete this leftover function

jsapp/js/components/table.es6
duvld: Should I make the same changes here? https://github.com/kobotoolbox/kpi/blob/2d41d4f4f574313a4d18b8e3ff87aa45c52c230e/jsapp/js/components/table.es6#L412-L416

magicznyleszek: I think you can leave it as is - ideally we would want all the code to use the constant, but let's try to only update the parts of code we touch :)

jsapp/js/components/table.es6
```diff
@@ -958,6 +964,24 @@ export class DataTable extends React.Component {
       </bem.FormView__item>
     );
   }
+  getMediaDownloadLink(fileName) {
+    var attachmentUrl = null;
```
magicznyleszek: I think you can remove this line and in the end of function just return:

```
return `${kc_base}/attachment/original?media_file=${encodeURI(fileName)}`;
```

PR #2554, opened 2020-01-15
2549 remove geopoint limit on question filter

### LINKED ISSUE #2549: Pull all GPS points when filtering by response
## Description

As a user with a large project (>5000 points) I'd like to see all responses that correspond to a particular question response on the map. When clicking a question and one of the responses, the map should make a new API request for all data corresponding to the response. 

At the moment, the map only pulls the first 5000 points and then shows a subset of points within this subset when filtering. 

## To reproduce current behavior

User goal: Show all points in a particular province. (The below project has 15,781 responses.)

1. User opens the map. The feature pulls the first 5000 records `?limit=5000&start=0&sort={%22_id%22:-1}&fields=[%22_id%22,%22_geolocation%22]`
1. User clicks on the question filter and selects *A.04 Province/District* . Feature pulls new list of records, again limited by 5000: `?limit=5000&start=0&sort={%22_id%22:-1}&fields=[%22_id%22,%22_geolocation%22,%22A04%22]
1. User clicks on *Nord Kivu*. Map shows 1,854 points. 

image

## Expected behavior 

3. User sees ALL 6,236 points where A04="Nord-Kivu", even if that means displaying more than 5,000. Query in this example would be: `?fields=["_id","_geolocation","A04"]&query={"A04":%20"Nord-Kivu"}`

duvld:
## Description

Fixes #2549, queries to get data no longer has the 5000 limit if `viewby` exists.

### DIFF
--- a/jsapp/js/dataInterface.es6
+++ b/jsapp/js/dataInterface.es6
@@ -491,18 +491,13 @@ export var dataInterface;
       if (fields.length)
         f = `&fields=${JSON.stringify(fields)}`;
 
-      console.log('entire query: ' + query + s + f);
-      console.log('# of fields: ' + fields.length);
-
       if (fields.length > 2) {
         const query_no_limit = `&start=${page}`;
-        console.log('in the if');
         return $ajax({
           url: `${ROOT_URL}/api/v2/assets/${uid}/data/?${query_no_limit}${s}${f}${filter}`,
           method: 'GET'
         });
       } else {
-        console.log('in the else');
         return $ajax({
           url: `${ROOT_URL}/api/v2/assets/${uid}/data/?${query}${s}${f}${filter}`,
           method: 'GET'
--- a/jsapp/js/dataInterface.es6
+++ b/jsapp/js/dataInterface.es6
@@ -491,7 +491,7 @@ export var dataInterface;
       if (fields.length)
         f = `&fields=${JSON.stringify(fields)}`;
 
-      if (fields.length > 2) {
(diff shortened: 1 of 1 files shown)


jsapp/js/dataInterface.es6
```diff
@@ -491,10 +491,18 @@ export var dataInterface;
...
+        return $ajax({
+          url: `${ROOT_URL}/api/v2/assets/${uid}/data/?${query_no_limit}${s}${f}${filter}`,
+          method: 'GET'
+        });
+      } else {
+        return $ajax({
+          url: `${ROOT_URL}/api/v2/assets/${uid}/data/?${query}${s}${f}${filter}`,
+          method: 'GET'
+        });
+      }
```
magicznyleszek: If you pass `limit=0` in the url, it will return all submissions. I think that instead of putting this logic in here, it would be better to do it in `map.es6` where you call `dataInterface.getSubmissions(this.props.asset.uid, 5000, 0, sort, fq)`. Just add the `fields.length > 3` logic there and pass either `5000` or `0` as limit :)

magicznyleszek: In this way the map-related change will be in the map file and it will be easier to understand this code in the future :)

duvld: Already did in https://github.com/kobotoolbox/kpi/pull/2560 ! I tried to put `limit = 0` but it throws me an error saying `{"limit":"must be a positive integer"}`

PR #2556, opened 2020-01-18
Fix vendors split

magicznyleszek:
Postfix for #2484 plus fixing frontend tests


jnm:
Travis failure is a `pip` issue that has nothing to do with this PR

PR #2575, opened 2020-02-24
Parse search queries into ORM filters

### LINKED ISSUE #2514: Custom query parser to handle queries without Whoosh/Haystack
Custom query parser to handle asset queries without Whoosh/Haystack
https://github.com/kobotoolbox/kpi/compare/two-databases...2514-custom-query-parser

### Enhancements
- Searches are usually faster
- If you specify a field, it becomes a database query, and it's reliable.
- Obviate the need for the haystack search index, one less thing to break

### Some changes

- An empty query results in the full set, not an empty set.
- Some changes to [how you search](https://github.com/kobotoolbox/kpi/tree/two-databases#searching-assets-and-collections):
	- *Still documenting this*
	- Field querying is suitable for exact matches
	- *Whoosh-specific things: either implemented compatibly or won't be necessary*

### Implementation details

- Used [canopy](http://canopy.jcoglan.com/langs/python.html) to generate a parser for `&q=` parameter (`grammar.peg -> grammar.py`)
- Turn the parse result into a Django Q object (`query_parser.py`)
- Call it from Search Filter (`filters.py`)

### Steps before this is done

- [x] Improve exception handling for, e.g., grammar.ParseError and django.core.exceptions.FieldError
- [x] Add test coverage. (The old system didn't have tests either)
  - [ ] Identify common use cases (queries that should still work, new ways to do old queries).
  - [x] Update the readme around searching [https://github.com/kobotoolbox/kpi/tree/two-databases#searching-assets-and-collections](https://github.com/kobotoolbox/kpi/tree/two-databases#searching-assets-and-collections)
- [x] Repair anything this breaks in the frontend (query behavior differences)
- [x] Remove Haystack/Whoosh (unless we keep it for some kind of fallback/compatibility?

jnm:
…obviating the need for Whoosh and Haystack. One of the last stepping stones on the path to merging `two-databases` into `master`‽ We should revert https://github.com/kobotoolbox/kobo-install/commit/a73ac5394a5f144bfcec11562bcae418bae16658 after merging this. Also, when we have a chance, we should remove other Whoosh-related stuff from `kobo-install`, `kobo-docker`, and `kobo-deployments`—but that's lower priority, since leaving it alone won't break anything.

Closes #2514


kpi/tests/test_utils.py
```diff
@@ -160,3 +162,18 @@ def test_autovalue_choices_arabic(self):
         part1 = 'العربية'
         part2 = '_001'
         self.assertEqual(surv['choices'][1]['$autovalue'], part1 + part2)
+
+    def test_query_parser(self):
+        query_string = '''
+            (a:a OR b:b AND c:c) AND d:d OR (
+                snakes:🐍🐍 AND NOT alphabet:🍲soup
+            ) NOT 'in a house' NOT "with a mouse"
```
noliveleger: `AND` is not required when using `NOT`?

jnm: Correct! It's assumed to be the same as `AND NOT`

kpi/utils/query_parser/query_parser.py
```diff
@@ -0,0 +1,133 @@
...
+    * `owner__username=meg` traverses the `owner` relationship, returning
+        any object whose `owner` field matches an object whose `username` field
+        exactly matches `meg` (case sensitive)
+    * `color:orange NOT (type:fruit OR type:chair)` returns anything
+        whose color is orange so long as it is not a fruit or a chair. AND, OR,
+        and NOT operators are supported. They must be in ALL CAPS.
+
+Special notes:
+    * If no field is specified in the query,
+        `settings.SEARCH_DEFAULT_FIELD_LOOKUP` is assumed. For example, if that
```
noliveleger: Will it work whether `settings.SEARCH_DEFAULT_FIELD_LOOKUP` is `summary` and `summary` field is changed from JSONField to JSONBField?

jnm: I guess if `summary` becomes a JSONB, we'd want `summary__icontains` to match on the stringified JSON representation, e.g. `summary__icontains:blah` would hopefully find an object whose summary looked like `{"deeply": [{"nested": "thing"}, {"blah": "blah"}]}`. I don't know whether or not it would work like that, but I'll try a test.

In the future, we can look at https://docs.djangoproject.com/en/2.2/ref/contrib/postgres/search/ now that we're on a version of Django that supports it.

jnm: @noliveleger, it seems like `icontains` does search through the entire JSON string by using `WHERE UPPER("jsonb_column"::text) LIKE UPPER('%blah%')`. Here's an example on `PerUserSetting`, just because it was a convenient model that had a `JSONBField`:
```
>>> uset = PerUserSetting()
>>> uset.user_queries = {"deeply": [{"nested": "thing"}, {"blah": "blah"}]}
>>> uset.save()
INSERT INTO "hub_perusersetting" ("user_queries",
                                  "name",
                                  "value_when_matched",
                                  "value_when_not_matched")
VALUES ('{"deeply": [{"nested": "thing"}, {"blah": "blah"}]}', 'INTERCOM_APP_ID', '', '') RETURNING "hub_perusersetting"."id" [4.86ms]
>>> PerUserSetting.objects.filter(user_queries__icontains='blah')
SELECT "hub_perusersetting"."id",
       "hub_perusersetting"."user_queries",
       "hub_perusersetting"."name",
       "hub_perusersetting"."value_when_matched",
       "hub_perusersetting"."value_when_not_matched"
FROM "hub_perusersetting"
WHERE UPPER("hub_perusersetting"."user_queries"::text) LIKE UPPER('%blah%')
LIMIT 21 [1.35ms]
<QuerySet [<PerUserSetting: INTERCOM_APP_ID>]>
```

noliveleger: 👍

PR #2578, opened 2020-02-25
Use native `JSONBField`s everywhere (instead of `JSONField`s)

noliveleger:


kpi/models/import_export_task.py
```diff
@@ -82,8 +82,9 @@ class Meta:
     )
 
     user = models.ForeignKey('auth.User', on_delete=models.CASCADE)
-    data = JSONField()
-    messages = JSONField(default={})
+    # ToDo validate whether both fields below need to be changed to jsonbfield
```
jnm: `data` is definitely an improvement as shown in your `filter()` change below. as for `messages`, why not?

kpi/models/import_export_task.py
```diff
@@ -600,9 +601,8 @@ def _run_task(self, messages):
     def _filter_by_source_kludge(queryset, source):
         """
         A disposable way to filter a queryset by source URL.
-        TODO: make `data` a `JSONBField` and use proper filtering
         """
-        return queryset.filter(data__contains=source)
+        return queryset.filter(data__source=source)
```
jnm: great, but it seems like `_filter_by_source_kludge()` is used only in `kpi/tests/test_mock_data_exports.py`. maybe we trash this method and use `filter(data__source=source)` directly in the tests? (or at least remove `_kludge` from the method name :smile:)

kpi/views/v1/export_task.py
```diff
@@ -33,9 +33,7 @@ def get_queryset(self, *args, **kwargs):
             return queryset
         if q.startswith('source:'):
             q = remove_string_prefix(q, 'source:')
-            # This is exceedingly crude... but support for querying inside
-            # JSONField not available until Django 1.9
-            queryset = queryset.filter(data__contains=q)
+            queryset = queryset.filter(data__source=q)
```
jnm: awesome

kpi/models/import_export_task.py
noliveleger: indeed

kpi/migrations/0015_assetversion.py
```diff
@@ -69,7 +70,7 @@ class Migration(migrations.Migration):
         migrations.AlterField(
             model_name='asset',
             name='summary',
-            field=JSONBField(default=dict, null=True),
+            field=jsonfield.fields.JSONField(default=dict, null=True),
```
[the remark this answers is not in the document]
noliveleger: that's an oversight.

PR #2579, opened 2020-02-26
Removed red warning when group title field is empty

### LINKED ISSUE #2526: Don't show red warning for groups without labels
## Description

Currently, the formbuilder requires a label for all groups and shows a red error if a group has no label, whereas xlsform allows unlabeled groups. This makes it confusing for users to use the formbuilder when a form needs logical groups that shouldn't be visible in Collect or Enketo. 

For example: 
* a group around questions for a particular household member that should be displayed only the person is older than 30 
* a group around 4 questions to force a new row in Enketo's grid theme format

## Steps to Reproduce

1. Create questions
2. Create group around questions
3. Remove the default 'Group' label

## Expected behavior 

User sees no text in group title, and does not see any errors. 
image

## Actual behavior 

Form shows `Group title is required`. 

## Additional details

See https://github.com/kobotoolbox/kpi/issues/950.

duvld:
## Description
Removed placeholder in `formBuilder` that warns user of empty group title.
Now remains blank and no red warning.
## Related issues
Fixes #2526


jsapp/scss/stylesheets/partials/form_builder/_card.scss
```diff
@@ -68,11 +68,6 @@
             color: $black;
         }
 
-        // we use placeholder as error message, as empty title is not welcomed
-         &::placeholder {
-            color: $cool-red;
-            background-color: rgba($cool-red, 0.075);
-        }
```
magicznyleszek: Please bring this back - these styles are used to display placeholder error message (you can search for `"Question label is required"` to see where it is being used) when question label is empty

jnm:
I added the `donotmerge` label to this because I'm concerned that the names of unlabeled groups may be randomly renamed each time the form is saved. Has anyone tested that?

magicznyleszek:
So I created a group "Welcome" that got a generated name `group_wo8gr91`. Saved it. I went to Form Builder and removed the name and saved it again. It is still `group_wo8gr91`.

Then I created a group "" that got a generated name `group_sm32s67`. Went the same route as previously and got `group_sm32s67`. 

@jnm is this what you were worried about?

jnm:
@magicznyleszek, yes, indeed. Thanks!

PR #2580, opened 2020-03-02
Make collections a type of asset instead of a separate model

jnm:
Includes a migration to copy all existing collections, retaining permissions and parent-child relationships.

**N.B.:** (from #2525)
A new management command has been created to update collection languages. It should be run against existing databases.
`./manage.py update_collection_languages`

Idea: add a separate data migration that runs this management command. Large DBs can fake the migration and run the command manually to avoid downtime.


jnm:
@dorey observes that the created assets have `content` and thus asset versions :-1:

kpi/filters.py
```diff
@@ -177,22 +152,7 @@ def filter_queryset(self, request, queryset, view):
...
-            # Find all the objects associated with those permissions, and then
-            # find all the permissions applied to all of those objects
-            result |= ObjectPermission.objects.filter(
-                content_type=content_type,
-                object_id__in=permissions_assigned_to_user.values(
-                    'object_id'
-                ).distinct()
-            )
+        result = ObjectPermission.objects.filter(
+            asset__permissions__user=user
```
noliveleger, line 156: Something I'm wondering: Is Django is smart enough to not add a join on `auth_user`  because of this condition `asset_permissions__user=user` ? 
Should it be better to use user's PK directly. e.g. `asset_permissions_user_id=user.pk`

kpi/management/commands/sync_kobocat_xforms.py
```diff
@@ -367,8 +367,7 @@ def _sync_permissions(asset, xform):
     current_kpi_perms = defaultdict(set)
     for user, kpi_permission in ObjectPermission.objects.filter(
                 deny=False,
-                content_type=ASSET_CT,
-                object_id=asset.pk
+                asset=asset,
```
noliveleger, line 370: Same question? 
Should it be better to use asset's PK directly. e.g. `asset_id=asset.pk`

kpi/filters.py
jnm: @noliveleger, yes, Django is smart enough. Let's not clutter our code with `…_id=….pk` where it's not needed:
```
>>> ObjectPermission.objects.filter(asset__permissions__user=user).distinct().count()
SELECT COUNT(*)
  FROM (
        SELECT DISTINCT "kpi_objectpermission"."id" AS Col1,
               "kpi_objectpermission"."user_id" AS Col2,
               "kpi_objectpermission"."permission_id" AS Col3,
               "kpi_objectpermission"."deny" AS Col4,
               "kpi_objectpermission"."inherited" AS Col5,
               "kpi_objectpermission"."asset_id" AS Col6,
               "kpi_objectpermission"."uid" AS Col7
          FROM "kpi_objectpermission"
         INNER JOIN "kpi_asset"
            ON ("kpi_objectpermission"."asset_id" = "kpi_asset"."id")
         INNER JOIN "kpi_objectpermission" T3
            ON ("kpi_asset"."id" = T3."asset_id")
         WHERE T3."user_id" = 1
       ) subquery


Execution time: 0.006980s [Database: default]

888
```

There are definitely still places where `object.related_id` is the way to go in order to avoid an extra query:
```
# a = Asset.objects.last()
>>> a.owner_id
21
>>> a.owner.pk
SELECT "auth_user"."id",
       "auth_user"."password",
       "auth_user"."last_login",
       "auth_user"."is_superuser",
       "auth_user"."username",
       "auth_user"."first_name",
       "auth_user"."last_name",
       "auth_user"."email",
       "auth_user"."is_staff",
       "auth_user"."is_active",
       "auth_user"."date_joined"
  FROM "auth_user"
 WHERE "auth_user"."id" = 21


Execution time: 0.000461s [Database: default]

21
```

kpi/management/commands/sync_kobocat_xforms.py
jnm: https://github.com/kobotoolbox/kpi/pull/2580#discussion_r567208901

PR #2588, opened 2020-03-10
Read the CSRF token from the DOM instead of cookie

jnm:
Closes kobotoolbox/tasks#343
Closes kobotoolbox/tasks#344


jnm:
This seems to cause problems after (failed?) deployments. Double check it.

magicznyleszek:
@jnm everything works correctly on my machine and the changes look legit :+1: What do you mean by problems after deployment?

jnm:
@magicznyleszek, I think I did something silly and checked out a different branch but forgot to rebuild the front-end files. I meant deploying a form, but I don't think there's really a problem (or at least not the one I saw). This is okay to merge if it's working well for you. Thanks for testing it.

PR #1864, opened 2018-07-12
Feature: REST Services

magicznyleszek:
## Description

Designs with initial idea: https://app.atomic.io/d/PjCNF93XD8wm

**Screenshots** for general idea:

screen shot 2018-08-02 at 10 14 31
screen shot 2018-08-02 at 10 14 39
screen shot 2018-08-15 at 16 40 13
screen shot 2018-08-15 at 16 39 43

## Testing

Preparations:
- [x] You need to use `REST-UI` branch in **kpi** and `kpi_rest_services` branch in **kobocat**
- [x] Run migrations with `./manage.py makemigrations` and `./manage.py migrate` in both containers
- [x] Remember to `docker-compose build kpi` and `docker-compose build kobocat`
- [x] Run project

To cover all the changes few tests (copied from #1868):

### Add REST Service
1. Have at least one deployed project
2. Go to project landing, then `Settings` and `REST Services`
3. Use `REGISTER A NEW SERVICE` button. _You can use `http://mockbin.com/request` as endpoint for tests_
4. Fill up the form and `CREATE`
5. Verify that the service appears on the list

### Edit REST Service
1. Have at least one deployed project with a service added
2. Go to project landing, then `Settings` and `REST Services`
3. Use `Edit` button on a service
4. Change name in the form and `SAVE`
5. Verify that the service appears on the list with name changed

### Delete REST Service
1. Have at least one deployed project with a service added
2. Go to project landing, then `Settings` and `REST Services`
3. Use `Delete` button on a service
4. `CONFIRM` prompt
5. Verify that the service does not appear on the list

### Count number of REST Service submissions
1. Have at least one deployed project with a service added
2. Go to project landing, then `Settings` and `REST Services`
3. Remember the `Count` column number for the service
4. Open the form in `Enketo Express` and create a new submission in a new window or tab
5. Refresh `REST Services` view
6. Verify that the `Count` column number for the service changed by +1

### Retry failed REST Service submission
1. Have at least one deployed project with a service added
2. Go to project landing, then `Settings` and `REST Services`
3. Edit your `Rest Service` by changing its endpoint to something wrong such as `http://dummy.endpoint.local`. _You can still use [Mockbin.org](http://mockbin.org/bin/create) to create a bin with 500 responses_
4. Open the form in `Enketo Express` and create a new submission in a new window or tab
5. Open service logs view
6. Verify that the submission has failed (status code)
7. Revert back the endpoint of your service to one that works.
8. Use `Retry submission` button on failed submission
[...]

PR #2596, opened 2020-03-12
Protect hook endpoints against SSRF attacks

noliveleger:
Closes KPI part of https://github.com/kobotoolbox/tasks/issues/341
Related to https://github.com/kobotoolbox/kobocat/pull/600
Use new package: https://github.com/kobotoolbox/ssrf-protect

### DIFF
--- a/kobo/apps/hook/models/service_definition_interface.py
+++ b/kobo/apps/hook/models/service_definition_interface.py
@@ -6,6 +6,7 @@ from abc import ABCMeta, abstractmethod
 
 import constance
 import requests
+from ssrf_protect.ssrf_protect import SSRFProtect, SSRFProtectException
 
 from kpi.utils.log import logging
 from .hook import Hook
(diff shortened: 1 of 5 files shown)


kobo/apps/hook/models/service_definition_interface.py
```diff
@@ -113,7 +116,15 @@ def send(self):
                     text = response.text
                     status_code = response.status_code
                 self.save_log(status_code, text)
-
+            except SSRFProtectException as e:
```
jnm, line 119: I should've brought this up way back in #1864, but these very large `try` blocks and catch-call `except Exception as e` handlers should be refactored at some point. In this `send()` method, for example, I think the only calls that should be inside `try`s are:
* `SSRFProtect.validate(self._hook.endpoint)`
* `response.raise_for_status()`

I'm not sure why `except Exception as e` is better than letting the exception fall through to Sentry. Maybe I'm forgetting something.

PR #2608, opened 2020-04-01
Better deleted submission error and title for REST Services

### LINKED ISSUE #2470: Deleted submissions referenced by REST Service logs
## Description

Sometimes people delete a submission after it was posted to an external server via REST Services. When they do that and then click on the corresponding entry in the REST Services log, they see the JSON `{"detail":"Not found."}` message and potentially a funky count in the title area of the modal, e.g. `Submission Record (0 of 1)`.

## Steps to Reproduce

1. Set up REST Services
2. Make a submission
3. Check that the submission appears in the REST Services log
4. Delete the submission
5. Click on the submission in the REST Services log

## Expected behavior 

The modal says something like "The submission could not be found. It may have been deleted." Maybe the error text should including the submission ID number.

## Actual behavior 

The modal shows JSON (which is pretty descriptive, hence the low priority of this issue).
image

duvld:
# Description
Two changes:
* Made a better displayed error message if submission is deleted in REST service view
* Removed the `0 of 1` from submission title if there is no sequence of submissions to view. It seems like this happens regardless if the submission is viewed from the REST services table

## Related issues
Fixes #2470

### DIFF
--- a/jsapp/js/components/modal.es6
+++ b/jsapp/js/components/modal.es6
@@ -160,13 +160,18 @@ class Modal extends React.Component {
     let title = t('Submission Record'),
         p = props.params,
         sid = parseInt(p.sid);
+    console.dir(p);
 
     if (p.tableInfo) {
       let index = p.ids.indexOf(sid) + (p.tableInfo.pageSize * p.tableInfo.currentPage) + 1;
       title =  `${t('Submission Record')} (${index} ${t('of')} ${p.tableInfo.resultsTotal})`;
     } else {
       let index = p.ids.indexOf(sid);
-      title =  `${t('Submission Record')} (${index} ${t('of')} ${p.ids.length})`;
+      if (p.ids.length === 1) {
+          title = `${t('Submission Record')}`;
+      } else {
+          title = `${t('Submission Record')} (${index} ${t('of')} ${p.ids.length})`;
+      }
     }
 
     return title;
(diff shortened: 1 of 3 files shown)


jsapp/js/components/modal.es6
```diff
@@ -160,13 +160,18 @@ class Modal extends React.Component {
     let title = t('Submission Record'),
         p = props.params,
         sid = parseInt(p.sid);
+    console.dir(p);
```
magicznyleszek, line 163: Please remove this leftover console :)

jsapp/js/components/modalForms/submission.es6
```diff
@@ -105,12 +106,16 @@ class Submission extends React.Component {
...
     }).fail((error)=>{
-      if (error.responseText)
-        this.setState({error: error.responseText, loading: false});
-      else if (error.statusText)
-        this.setState({error: error.statusText, loading: false});
-      else
+      if (error.responseText) {
+        let error_message = error.responseText;
+        if (error_message === DETAIL_NOT_FOUND)
+          error_message = 'The submission could not be found. It may have been deleted. Submission ID: ' + sid;
```
magicznyleszek, line 112: Plese make it a translatable string: `error_message = t('The submission could not be found. It may have been deleted. Submission ID: ##id##').replace('##id##', sid);`

jsapp/js/components/modal.es6
```diff
@@ -166,7 +166,11 @@ class Modal extends React.Component {
       title =  `${t('Submission Record')} (${index} ${t('of')} ${p.tableInfo.resultsTotal})`;
     } else {
       let index = p.ids.indexOf(sid);
-      title =  `${t('Submission Record')} (${index} ${t('of')} ${p.ids.length})`;
+      if (p.ids.length === 1) {
+          title = `${t('Submission Record')}`;
```
magicznyleszek, line 170: Nice 👍

PR #2610, opened 2020-04-02
Fixed bug where adding translations to matrix would add another null …

### LINKED ISSUE #2604: Matrix is broken when using translations
## Description

Having a matrix question and translations generates an unnecessary `null` translation when saving in Form Builder.

## Steps to Reproduce

1. Create new form using Form Builder
2. Define a new question using "Question Matrix" (you can leave default values)
3. Save and close
4. In Manage Languages modal define default language (e.g. "English (en)") and close the modal
5. Open form in Form Builder
6. Don't edit anything, just hit Save and close
7. Observe Languages list is now "English (en), Unnamed language"

## Expected behavior 

The `null` language should not be added.

## Additional details

Observed on https://kf.kobotoolbox.org/.

My guess would be to check out the `nullifyTranslations`/`unnullifyTranslations` flow and what it does to the matrix type. Probably Matrix has some custom hack on its own.

duvld:
## Description

There was an error with the way `label`s were being set by `unnullifyTranslation`. The function was being fed no choices so it didn't bother setting all `label`s to something like `label::(Translation (tr))`

Changed the way `editableForm` saves forms with Matrices

## Related issues

Fixed #2604

### DIFF
--- a/jsapp/js/editorMixins/editableForm.es6
+++ b/jsapp/js/editorMixins/editableForm.es6
@@ -233,8 +233,9 @@ export default assign({
     }
 
     let surveyJSON = surveyToValidJson(this.app.survey);
+    let surveyJSONWithMatrix = koboMatrixParser({source: surveyJSON}).source;
     if (this.state.asset) {
-      surveyJSON = unnullifyTranslations(surveyJSON, this.state.asset.content);
+      surveyJSON = unnullifyTranslations(surveyJSONWithMatrix, this.state.asset.content);
     }
     let params = {content: surveyJSON};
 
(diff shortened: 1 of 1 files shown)


jsapp/js/editorMixins/editableForm.es6
```diff
@@ -233,8 +233,9 @@ export default assign({
     }
 
     let surveyJSON = surveyToValidJson(this.app.survey);
+    let surveyJSONWithMatrix = koboMatrixParser({source: surveyJSON}).source;
```
magicznyleszek, line 236: Please move definition of `surveyJSONWithMatrix` inside od `if` below, as it is not used anywhere else, and it could be confusing in the future why there are two `surveyJSON`s, if they stay next to each other at the same nesting level :)

magicznyleszek, line 236: Please move definition of `surveyJSONWithMatrix` inside od `if` below, as it is not used anywhere else, and it could be confusing in the future why there are two `surveyJSON`s, if they stay next to each other at the same nesting level :)

PR #2632, opened 2020-04-16
Migrate text-based JSON fields without locking tables

jnm:
* Partially reverts #2578 by returning to `jsonfield.JSONField`, which supports both `text` and `jsonb` column types, instead of using `django.contrib.postgres.fields.JSONField`, which supports only the `jsonb` column type.
* Adds a management command, `migrate_json_text_to_jsonb` that can be run _instead of_ migration `0024_alter_jsonfield_to_jsonbfield` to avoid downtime on large databases. It works by:
    1. Creating a new `jsonb` column for each `text` JSON column;
    1. Setting up a trigger to mirror any writes to the `text` column into the `jsonb` column;
    1. Running chunked `UPDATE`s so that the `jsonb` column is populated for all rows;
    1. Removing the original `text` column and renaming the `jsonb` column to match.
* Includes a workaround for Unicode searches of `summary`;
* Modifies `0024_alter_jsonfield_to_jsonbfield` to strip out `\u0000` (encoded null characters) in text JSON columns.

Test of `migrate_json_text_to_jsonb` on kf.master.kbtdev.org:
```
Adding columns and triggers..........
Updating kpi_asset: 1000 <snip> 604512
Updating kpi_assetsnapshot: 1000 <snip> 285552
Updating kpi_assetversion: 1000 <snip> 3662485
Updating kpi_exporttask: 1000 <snip> 263732
Updating kpi_importtask: 17
Setting new columns to NOT NULL..........
Replacing old columns with new columns and dropping triggers..........
Done!

real    33m58.444s
user    0m2.348s
sys     0m0.400s
```

After this is done and included in a release, #2635 must be completed.


jnm:
@dorey, I feel good about this trigger approach, but I can add in some extra checking logic from https://github.com/kobotoolbox/kpi/blob/7bf34a188214796b19c04ff39682719243173974/kpi/management/commands/populate_asset_jsonbfields.py if you'd like.

PR #2636, opened 2020-04-21
Edit asset page : Unwanted hints shown in other items

gushil:
## Description

Edit asset page : Unwanted hints shown in other items when user is pointing to an item icon.


magicznyleszek:
@gushil would you be able to write down how to reproduce this bug? I see the culprit element in the dev tools, but am not able to make the error happen, so it probably requires some particular context.

gushil:
@magicznyleszek 

image

I hope this image can help you.

Thanks.

PR #2639, opened 2020-04-24
Added a network check before username check

### LINKED ISSUE #2341: "User not found" message during server/network failure
When handling a failed `queryUserExistence()` request:
https://github.com/kobotoolbox/kpi/blob/57794d49f4b7a74fcb9472e55e80d095e76b7029/jsapp/js/stores.es6#L400
It would be nice to make a distinction between a network/server failure and bad user input. For example, if I disconnect from the internet and try to grant permissions, I see `User not found: [username]` for valid usernames. Since we now (#2289) display an alert (`notifyUnknownUser()`) instead of a more-ambiguous greyed-out button, it would be better to show `User not found` only when receiving a 404 from the backend.

duvld:
## Description
Added check for active internet connection before checking backend for username
When trying to add permissions during a network failure, no `User not found` notifications are shown

May be a better solution for a more general network failure handling, for example notifying that the connection has dropped (?)
## Related issues
Fixes #2341

### DIFF
--- a/jsapp/js/stores.es6
+++ b/jsapp/js/stores.es6
@@ -396,10 +396,8 @@ var selectedAssetStore = Reflux.createStore({
 var userExistsStore = Reflux.createStore({
   init () {
     this.checked = {};
-    if (navigator.onLine) {
-        this.listenTo(actions.misc.checkUsername.completed, this.usernameExists);
-        this.listenTo(actions.misc.checkUsername.failed, this.usernameDoesntExist);
-    }
+    this.listenTo(actions.misc.checkUsername.completed, this.usernameExists);
+    this.listenTo(actions.misc.checkUsername.failed, this.usernameDoesntExist);
   },
   checkUsername (username) {
     if (username in this.checked) {
--- a/jsapp/js/stores.es6
+++ b/jsapp/js/stores.es6
@@ -396,8 +396,10 @@ var selectedAssetStore = Reflux.createStore({
 var userExistsStore = Reflux.createStore({
   init () {
     this.checked = {};
-    this.listenTo(actions.misc.checkUsername.completed, this.usernameExists);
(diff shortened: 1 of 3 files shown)


jsapp/js/stores.es6
```diff
@@ -396,8 +396,10 @@ var selectedAssetStore = Reflux.createStore({
 var userExistsStore = Reflux.createStore({
   init () {
     this.checked = {};
-    this.listenTo(actions.misc.checkUsername.completed, this.usernameExists);
-    this.listenTo(actions.misc.checkUsername.failed, this.usernameDoesntExist);
+    if (navigator.onLine) {
```
magicznyleszek, line 399: By checking it here - during initialization - it would prevent it from working when the internet connection is regained. I think it would be best for now to add this check `if (navigator.onLine)` in the `notifyUnknownUser` function (in both `jsapp/js/components/permissions/userAssetPermsEditor.es6` and `jsapp/js/components/permissions/userCollectionPermsEditor.es6`) with a comment of explanation. We would need to think of a more general offline mode

duvld: I tried your solution and it works when the network goes offline, but after regaining connection entering an existing user prompts the `notifyUnkownUser` again, the user would have to refresh the page in order to get expected functionality again. Maybe we should leave this for a more general soltution?

magicznyleszek: I think we need to change `stores.userExists` to not store any information on checks if there is no internet. So both `usernameExists` and `usernameDoesntExist` callbacks should run `this.checked[username] = …` only if there is internet. This will make the user check be triggered and end up being `undefined` (so "unknown").

I did also create a general network failure handling issue https://github.com/kobotoolbox/kpi/issues/2685

PR #2640, opened 2020-04-25
Adding gradual pagination to version history

### LINKED ISSUE #1306: "Show full history" should support pagination
After fixing #1302, the `AssetSerializer` now returns paginated `deployed_versions`, with 100 results on the first page. The form landing page should allow users to view more `deployed_versions` after the first 100 by requesting more pages from the API.

duvld:
## Description

Added the following changes to how version history is displayed. I've made the changes limited to 2 items each to help paint the picture:

1306-pagination-1

As mentioned in #1306 , the `LOAD MORE` and `LOAD LESS` options would increase/decrease the amount of version displayed by 20 (unlike the 2 shown above). `HIDE FULL HISTORY` and `SHOW FULL HISTORY` still function the same and the amount of version displayed remains the same until refreshed

#### Further Improvements
We could also add a field to manually set how many versions to be displayed if an arbitrary number like 20 is too general

Should probably change the name of the button to something like `SHOW VERSION HISTORY`, this is left alone for now due to potential translation issues

## Related issues

Fixes #1306

### DIFF
--- a/jsapp/js/components/formLanding.es6
+++ b/jsapp/js/components/formLanding.es6
@@ -10,7 +10,10 @@ import ui from '../ui';
 import mixins from '../mixins';
 import DocumentTitle from 'react-document-title';
 import CopyToClipboard from 'react-copy-to-clipboard';
-import {MODAL_TYPES} from '../constants';
+import {
+    MODAL_TYPES,
+    DVCOUNT_LIMIT_MINIMUM
+} from '../constants';
 import {
   formatTime,
   t,
@@ -21,7 +24,8 @@ export class FormLanding extends React.Component {
   constructor(props){
     super(props);
     this.state = {
-      selectedCollectMethod: 'offline_url'
+      selectedCollectMethod: 'offline_url',
+      DVCOUNT_LIMIT: 2
     };
     autoBind(this);
   }
@@ -129,6 +133,16 @@ export class FormLanding extends React.Component {
       asset: this.state
     });
   }
+  increaseDVCOUNT_LIMIT() {
+    if (this.state.DVCOUNT_LIMIT + DVCOUNT_LIMIT_MINIMUM <= this.state.deployed_versions.count + DVCOUNT_LIMIT_MINIMUM) {
+      this.setState({ DVCOUNT_LIMIT : this.state.DVCOUNT_LIMIT + DVCOUNT_LIMIT_MINIMUM });
+    }
+  }
+  decreaseDVCOUNT_LIMIT() {
+    if (this.state.DVCOUNT_LIMIT - DVCOUNT_LIMIT_MINIMUM > 0) {
+      this.setState({ DVCOUNT_LIMIT : this.state.DVCOUNT_LIMIT -  DVCOUNT_LIMIT_MINIMUM });
+    }
+  }
   renderHistory () {
     var dvcount = this.state.deployed_versions.count;
     return (
@@ -147,7 +161,7 @@ export class FormLanding extends React.Component {
             </bem.FormView__group>
             {this.state.deployed_versions.results.map((item, n) => {
               return (
-                <bem.FormView__group m='items' key={n} >
+                <bem.FormView__group m='items' key={n} className={n >= this.state.DVCOUNT_LIMIT ? 'hidden' : ''} >
                   <bem.FormView__label m='version'>
                     {`v${dvcount - n}`}
                     {item.uid === this.state.deployed_version_id && this.state.deployment__active &&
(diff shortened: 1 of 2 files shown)


jsapp/js/constants.es6
```diff
@@ -347,6 +347,8 @@ new Set([
 
 export const NAME_MAX_LENGTH = 255;
 
+export const DVCOUNT_LIMIT_MINIMUM = 20;
```
magicznyleszek, line 350: Please only add constants to this file that are used in more than one file. In this situation `DVCOUNT_LIMIT_MINIMUM` should be defined in `formLanding.es6`

jsapp/js/components/formLanding.es6
```diff
@@ -177,6 +191,16 @@ export class FormLanding extends React.Component {
...
+            {this.state.historyExpanded &&
+              <button className='mdl-button mdl-button--colored' onClick={this.increaseDVCOUNT_LIMIT}>
+                {this.state.DVCOUNT_LIMIT < dvcount && t('Load more')}
+              </button>
+            }
+            {this.state.historyExpanded &&
+               <button className='mdl-button mdl-button--colored' onClick={this.decreaseDVCOUNT_LIMIT}>
+                  {this.state.DVCOUNT_LIMIT - DVCOUNT_LIMIT_MINIMUM >= DVCOUNT_LIMIT_MINIMUM && t('Load less')}
+              </button>
+            }
```
magicznyleszek, line 203: Is there really a point in having a "load less" button? I would just keep it simple and have two buttons.

jsapp/js/components/formLanding.es6
```diff
@@ -129,8 +134,31 @@ export class FormLanding extends React.Component {
       asset: this.state
     });
   }
+  increaseDVCOUNT_LIMIT() {
+    if (this.state.DVCOUNT_LIMIT + DVCOUNT_LIMIT_MINIMUM <= this.state.deployed_versions.count + DVCOUNT_LIMIT_MINIMUM) {
+      this.setState({ DVCOUNT_LIMIT : this.state.DVCOUNT_LIMIT + DVCOUNT_LIMIT_MINIMUM });
+    }
+    if (this.state.DVCOUNT_LIMIT >= 80) {
+      var curr = this.state.currentVersions;
+      if (this.state.nextPageUrl) {
```
magicznyleszek, line 143: To avoid setting `nextPageUrl` in render cycle, let's do it this way:

```javascript
let urlToLoad = null;
if(this.state.furtherPageUrl) {
  urlToLoad = this.state.furtherPageUrl;
} else if (this.state.deployed_versions.next) {
  urlToLoad = this.state.deployed_versions.next;
}
if (urlToLoad !== null) {
  dataInterface.loadNextPageUrl(urlToLoad).done((data) => {
    this.setState({furtherPageUrl: data.deployed_versions.next});
    Object.values(data.deployed_versions.results).forEach(function(ele){
        curr.push(ele);
    });
  });
}
```
This way you will either load first `.next` or second `.next` and further ones :)

jsapp/js/components/formLanding.es6
```diff
@@ -129,8 +134,31 @@ export class FormLanding extends React.Component {
...
+      this.setState({currentVersions: curr});
+    }
+  }
   renderHistory () {
     var dvcount = this.state.deployed_versions.count;
+    if (this.state.nextPageUrl === 'inital') {
+      this.setState({
+          nextPageUrl: this.state.deployed_versions.next,
+      });
+      this.state.currentVersions = this.state.deployed_versions.results;
```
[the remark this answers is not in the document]
duvld: I added it because I wanted a way to reference the return of loading the next page, and I could not reference `this` in the response to my `dataInterface` call. Although your code below solved this issue I'm still not entirely sure why. It appears like I'm not doing anything to the `curr` var anymore yet the code needs `curr.push(ele)` to work...

jsapp/js/components/formLanding.es6
```diff
@@ -129,8 +134,31 @@ export class FormLanding extends React.Component {
       asset: this.state
     });
   }
+  increaseDVCOUNT_LIMIT() {
+    if (this.state.DVCOUNT_LIMIT + DVCOUNT_LIMIT_MINIMUM <= this.state.deployed_versions.count + DVCOUNT_LIMIT_MINIMUM) {
+      this.setState({ DVCOUNT_LIMIT : this.state.DVCOUNT_LIMIT + DVCOUNT_LIMIT_MINIMUM });
+    }
+    if (this.state.DVCOUNT_LIMIT >= 80) {
```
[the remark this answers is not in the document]
duvld: Leftover from the first time I tried my solution, I needed to call `dataInterface` once the next time a user clicks it would need to load the next 100. Removed now

magicznyleszek:
@duvld while re-reviewing this, I realised my initial review didn't cover everything so I decided to just change it myself :-)

PR #2641, opened 2020-04-27
Dropped makeEditable from Choices, used input instead

### LINKED ISSUE #2152: Drop makeEditable from choices list
## Description

Choices list in Form Builder is the last place that uses the weird `makeEditable` function. Let's remove it and use simple `<input>`s. That way it would be more consistent with the rest of user editable text. Plus we would get tab-navigation as bonus.

## Additional details

See #2151 for details on how `makeEditable` was removed from question names.

duvld:
## Description
Changed the way `XML Value:` and options in choices questions are editable. Introduced input tags and removed `makeEditable`
## Related issues
Fixes #2152


jsapp/xlform/src/view.choices.coffee
```diff
@@ -86,22 +86,22 @@ module.exports = do ->
...
-        $('span', @c).html @model.get("name")
+        $('input', @c).val @model.get("name")
         @model.set('setManually', true)
       else
         @model = new $choices.Option()
         @options.cl.options.add(@model)
-        @p.html("Option #{1+@options.i}").addClass("preliminary")
+        @p.val("Option #{1+@options.i}").addClass("preliminary")
 
-      $viewUtils.makeEditable @, @model, @p, edit_callback: _.bind @saveValue, @
```
magicznyleszek, line 102: Please delete `viewUtils.makeEditable` from `/jsapp/xlform/src/view.utils.coffee` :)

duvld:
I just noticed an oversight, the new input fields for options are not being saved. Submissions probably have to be handled differently due to being input fields now. Looking into it now.

PR #2643, opened 2020-04-29
add lang_header parameter for specify language header values in export

jdugh:
add lang_header parameter. lang_header can be used for specify the language for header values in excel/csv export (you can have _xml header values and _default values for submits datas). If not set, lang arg override lang_header arg value. Need this PR : https://github.com/kobotoolbox/formpack/pull/215

### DIFF
--- a/kpi/models/import_export_task.py
+++ b/kpi/models/import_export_task.py
@@ -348,6 +348,12 @@ class ExportTask(ImportExportTask):
               response values. Specify `_xml` to use question and choice names
               instead of labels. Leave unset, or use `_default` for labels in
               the default language
+    * `lang_header`: optional; the name of the translation to be used just for
+              headers values. Specify `_xml` to use question names in headers
+              instead of labels. Use `_default` for headers labels in the
+              default language. Leave unset for override it and use lang arg
+              value.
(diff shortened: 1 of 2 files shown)


dorey:
Hi @jdugh . Thanks for the PR. Could you add a test that uses this column and shows how the values would change when this parameter is used?

Also, if I understand correctly what this does, it might make more sense to call it `header_lang` or `header_language`.

kpi/models/import_export_task.py
```diff
@@ -462,10 +468,12 @@ def _build_export_options(self, pack):
         group_sep = self.data.get('group_sep', '/')
         translations = pack.available_translations
         lang = self.data.get('lang', None) or next(iter(translations), None)
+        lang_header = self.data.get('lang_header', lang) or next(iter(translations), None)
```
dorey, line 471:
```suggestion
        lang_header = self.data.get('lang_header', lang)
```

because `lang` is the fallback value, and it's set in the line above, we don't need a second fallback value

jnm:
PR was closed accidentally because we removed the underlying base branch. Please carry on the discussion!

PR #2653, opened 2020-05-02
Improve performance of the reports query

mlazowik:
## Checklist

3. [x] Ensure the tests pass
5. [x] If this is a big feature, make sure to prefix the title with `Feature:` and add a thorough description for non-dev folk

## Description

* annotations insted of distinct(): distinct() had to compare big JSON
  fields (you can pass field names to distinct(), but annotations are a
  more canonical solution here IMO)
* defer loading of big JSON fields, as the query is also a base for
  the detail view
## Related issues

https://community.kobotoolbox.org/t/is-assetversion-immutable-solving-performance-issues-with-big-forms/9046

### DIFF
--- a/kobo/apps/reports/views.py
+++ b/kobo/apps/reports/views.py
@@ -1,8 +1,14 @@
 # coding: utf-8
+from django.db.models import Count, Case, When
 from rest_framework import viewsets, mixins
 
+<<<<<<< Updated upstream
 from kpi.constants import PERM_VIEW_SUBMISSIONS, PERM_PARTIAL_SUBMISSIONS
 from kpi.models import Asset
+=======
+
+from kpi.models import AssetVersion, Asset
+>>>>>>> Stashed changes
 from kpi.models.object_permission import get_objects_for_user, get_anonymous_user
 from .serializers import ReportsListSerializer, ReportsDetailSerializer
 
@@ -19,9 +25,18 @@ class ReportsViewSet(mixins.ListModelMixin,
             return ReportsDetailSerializer
 
     def get_queryset(self):
-        
-        # Retrieve all deployed assets first.
-        deployed_assets = Asset.objects.filter(asset_versions__deployed=True).distinct()
+        # Retrieve all deployed assets first. Defer large JSON fields, because this queryset
+        # is a base for the detail view that selects a specific Asset. Without defer()
+        # we might be downloading those fields for Assets that will not be looked at.
+        deployed_assets = Asset.objects.annotate(
+            count_deployed=Count(Case(When(asset_versions__deployed=True, then=1)))
+        ).filter(
+            count_deployed__gt=0
+        ).defer(
+            'content',
+            'report_styles'
+        )
+
         # Then retrieve all assets user is allowed to view
         # (user must have 'view_submissions' on Asset objects)
         required_permissions = [
--- a/kobo/apps/reports/views.py
+++ b/kobo/apps/reports/views.py
@@ -4,7 +4,6 @@ from rest_framework import viewsets, mixins
 
 from kpi.constants import PERM_VIEW_SUBMISSIONS, PERM_PARTIAL_SUBMISSIONS
 from kpi.models import Asset
-
 from kpi.models.object_permission import get_objects_for_user, get_anonymous_user
 from .serializers import ReportsListSerializer, ReportsDetailSerializer
 
--- a/kobo/apps/reports/views.py
+++ b/kobo/apps/reports/views.py
@@ -2,13 +2,9 @@
 from django.db.models import Count, Case, When
 from rest_framework import viewsets, mixins
 
-<<<<<<< Updated upstream
 from kpi.constants import PERM_VIEW_SUBMISSIONS, PERM_PARTIAL_SUBMISSIONS
 from kpi.models import Asset
-=======
 
-from kpi.models import AssetVersion, Asset
->>>>>>> Stashed changes
 from kpi.models.object_permission import get_objects_for_user, get_anonymous_user
 from .serializers import ReportsListSerializer, ReportsDetailSerializer
 


mlazowik:
CI seems to be broken? `E   ModuleNotFoundError: No module named 'mock'` is also present in the last few commits on master.

dorey:
Yes we will be fixing the travis issues shortly.

noliveleger:
FYI: @mlazowik #2661

jnm:
Thanks for drawing our attention to these queries. They definitely could use some help!

Am I right to guess that when you wrote:
> defer loading of big JSON fields, as the query is also a base for the detail view

…you meant "base for the list view"? For the list view, it looks like the only field we need is `uid`(_!_), and deferring the rest should save a good bit of time. The detail view, though, actually needs `report_styles` (as well as `name` and `uid`); we could defer other fields, but I'm not sure it would be worthwhile when loading a single asset instead of a list. You mentioned having forms with 2k questions on the forum; for those, do you see a noticeable difference between `Asset.objects.get(id=<big form>)` and `Asset.objects.defer('content').get(id=<big form>)`? I ask because there's a slight maintenance burden added by `defer()`, `only()`, and friends: if anyone ever needs to use a deferred field in the future, they must remember to stop deferring it or face the penalty of Django executing an entirely new query to retrieve the field.

You identified another choke point where we filter for assets that have deployments:
> annotations insted of distinct(): distinct() had to compare big JSON
fields (you can pass field names to distinct(), but annotations are a
more canonical solution here IMO)

Your annotation is a big improvement over the join + distinct approach, but I think we can do even better now that Django (1.11+) provides https://docs.djangoproject.com/en/2.2/ref/models/expressions/#exists-subqueries. This lets us annotate assets with the existence (true or false) of any deployed version—without actually counting the versions.

For reference, I've included some Postgres `explain analyze` output for the two approaches below. These `COUNT` queries were found by using https://github.com/jazzband/django-debug-toolbar with the `/reports/` list endpoint. The `IN` clauses with large numbers of IDs are addressed by #2671.

#### `Exists()` annotation
```
explain analyze SELECT
    COUNT(*) AS "__count"
FROM
    "kpi_asset"
WHERE ((("kpi_asset"."asset_type" = 'survey'
            AND "kpi_asset"."id" IN (<1882 IDs>))
        OR ("kpi_asset"."asset_type" = 'survey'
            AND "kpi_asset"."id" IN (<5473 IDs>)))
    AND EXISTS (
        SELECT
            U0."id",
            U0."uid",
            U0."asset_id",
            U0."name",
            U0."date_modified",
            U0."_reversion_version_id",
            U0."version_content",
            U0."uid_aliases",
            U0."deployed_content",
            U0."_deployment_data",
            U0."deployed"
        FROM
            "kpi_assetversion" U0
        WHERE (U0. "asset_id" = ("kpi_asset"."id")
            AND U0. "deployed" = TRUE)) = TRUE);

 Aggregate  (cost=161286.31..161286.32 rows=1 width=0) (actual time=53.607..53.607 rows=1 loops=1)
   ->  Bitmap Heap Scan on kpi_asset  (cost=20894.15..161280.78 rows=2215 width=0) (actual time=12.209..53.247 rows=5504 loops=1)
         Recheck Cond: ((id = ANY ('{<1882 IDs>}'::integer[])) OR (id = ANY ('{<5473 IDs>}'::integer[])))
         Filter: (((asset_type)::text = 'survey'::text) AND (alternatives: SubPlan 1 or hashed SubPlan 2))
         Rows Removed by Filter: 1850
         Heap Blocks: exact=5463
         ->  BitmapOr  (cost=20894.15..20894.15 rows=7355 width=0) (actual time=11.390..11.390 rows=0 loops=1)
               ->  Bitmap Index Scan on kpi_asset_pkey  (cost=0.00..6649.97 rows=1882 width=0) (actual time=2.021..2.021 rows=1882 loops=1)
                     Index Cond: (id = ANY ('{<1882 IDs>}'::integer[]))
               ->  Bitmap Index Scan on kpi_asset_pkey  (cost=0.00..14243.07 rows=5473 width=0) (actual time=9.368..9.368 rows=5473 loops=1)
                     Index Cond: (id = ANY ('{<5473 IDs>}'::integer[]))
         SubPlan 1
           ->  Index Scan using kpi_assetversion_51c6d5db on kpi_assetversion u0  (cost=0.43..57.40 rows=9 width=0) (actual time=0.006..0.006 rows=1 loops=5543)
                 Index Cond: (asset_id = kpi_asset.id)
                 Filter: deployed
                 Rows Removed by Filter: 2
         SubPlan 2
           ->  Seq Scan on kpi_assetversion u0_1  (cost=0.00..541625.84 rows=646966 width=4) (never executed)
                 Filter: deployed
 Planning time: 4.252 ms
 Execution time: 53.904 ms
```

#### `Count()` annotation
```
explain analyze SELECT
    COUNT(*)
FROM (
    SELECT
        "kpi_asset"."id" AS Col1,
        COUNT(
            CASE WHEN "kpi_assetversion"."deployed" = TRUE THEN
                1
            ELSE
                NULL
            END) AS "count_deployed"
    FROM
        "kpi_asset"
    LEFT OUTER JOIN "kpi_assetversion" ON ("kpi_asset"."id" = "kpi_assetversion"."asset_id")
GROUP BY
    "kpi_asset"."id"
HAVING ((COUNT(
            CASE WHEN ("kpi_assetversion"."deployed" = TRUE) THEN
                1
            ELSE
                NULL
            END) > 0
        AND "kpi_asset"."id" IN (<1882 IDs>))
    OR (COUNT(
            CASE WHEN ("kpi_assetversion"."deployed" = TRUE) THEN
                1
            ELSE
                NULL
            END) > 0
        AND "kpi_asset"."id" IN (<5473 IDs>)))) subquery;

Aggregate  (cost=536694.58..536694.59 rows=1 width=0) (actual time=193.002..193.002 rows=1 loops=1)
  ->  HashAggregate  (cost=536511.11..536602.84 rows=7339 width=5) (actual time=191.353..192.688 rows=5504 loops=1)
        Group Key: kpi_asset.id
        Filter: (count(CASE WHEN kpi_assetversion.deployed THEN 1 ELSE NULL::integer END) > 0)
        Rows Removed by Filter: 1850
        ->  Nested Loop Left Join  (cost=20897.14..536281.86 rows=45850 width=5) (actual time=12.191..167.738 rows=127381 loops=1)
              ->  Bitmap Heap Scan on kpi_asset  (cost=20896.71..111542.66 rows=7339 width=4) (actual time=12.181..18.776 rows=7354 loops=1)
                    Recheck Cond: ((id = ANY ('{<1882 IDs>}'::integer[])) OR (id = ANY ('{<5473 IDs}'::integer[])))
                    Heap Blocks: exact=5463
                    ->  BitmapOr  (cost=20896.71..20896.71 rows=7355 width=0) (actual time=11.396..11.396 rows=0 loops=1)
                          ->  Bitmap Index Scan on kpi_asset_pkey  (cost=0.00..6649.97 rows=1882 width=0) (actual time=2.008..2.008 rows=1882 loops=1)
                                Index Cond: (id = ANY ('{<1882 IDs>}'::integer[]))
                          ->  Bitmap Index Scan on kpi_asset_pkey  (cost=0.00..14243.07 rows=5473 width=0) (actual time=9.388..9.388 rows=5473 loops=1)
                                Index Cond: (id = ANY ('{<5473 IDs}'::integer[]))
              ->  Index Scan using kpi_assetversion_51c6d5db on kpi_assetversion  (cost=0.43..57.32 rows=55 width=5) (actual time=0.003..0.015 rows=17 loops=7354)
                    Index Cond: (kpi_asset.id = asset_id)
Planning time: 5.326 ms
Execution time: 193.343 ms
```

mlazowik:
Thanks for digging into this!

> …you meant "base for the list view"?

No, I meant the detail view. I should've explained more, see below.

> do you see a noticeable difference between Asset.objects.get(id=<big form>) and Asset.objects.defer('content').get(id=<big form>)? 

As one might expect: no, not really (well, we have since refactored the forms and how the max is around ~500 questions). But this is not what was happening in the detail view, I'm pretty sure that the queryset was evaluated before being narrowed down to a single `Asset`. I see that you're bailing our early now in #2680, so that would not be a problem anymore.

I'll try to find time to look a bit more into this to confirm if it was really evaluated.

> I ask because there's a slight maintenance burden added by defer(), only(), and friends: if anyone ever needs to use a deferred field in the future, they must remember to stop deferring it or face the penalty of Django executing an entirely new query to retrieve the field.

That's a very important downside to what I did, but see above. I _think_ that now that you return the queryset for the detail view early `defer()` is no longer needed.

>  This lets us annotate assets with the existence (true or false) of any deployed version—without actually counting the versions.

This definitely does look better.

Issue #2420, opened 2019-09-23
Add support in the formbuilder for `rank` question type


tinok:
This question type is supported by pyxform, Enketo, and Collect, but not yet in the formbuilder. See http://xlsform.org/en/#rank

The UI implementation is the same as for `select_one` and `select_multiple` questions: 

After choosing entering a question label and choosing 'Rank' question type, users can enter any number of select options in the formbuilder. 

During data collection, users can then change the order of these items as needed.

magicznyleszek:
Note: we currently have a rank question type, but it is not the official one, just an old custom implementation that _<q>when downloaded to XLS or deployed, it ends up being only a group with a special appearance<q>_

nanometrenat:
Hi @magicznyleszek quick comment with user thoughts.

Yes agree that Kobo formbuilder has a question type that says it is a rank type - but as per support doco at https://support.kobotoolbox.org/en/articles/1564770-rating-vs-ranking-question-types it is just a series of select-one dropdowns with validation that you can't select the same option twice.  This is fine for ranking 3 or 4 options, but for questions with 10-12 options to rank it is hard for users to remember which options they have already used.

It would be very useful to have the Enketo/ODK style ranking, where the user orders the available options rather than selecting them separately (https://community.kobotoolbox.org/t/rank-widget/2212/6)

Thanks ever so much!

jnm:
@duvld, now that you're in CoffeeLand, could you try this issue? Before there was a standard XLSForm `rank` type, we (@dorey) implemented a KoBo-specific ranking type (similar to the matrix type) that gets compiled into a group of `select_one` questions as described by @nanometrenat above:
https://github.com/kobotoolbox/kpi/blob/5495a6334b432dfb2f7e5199638af6a34f7fc752/kpi/utils/xlsform_preprocessors/koborank_handler.py#L9-L45

I think we need to support the new `rank` type concurrently with our old `kobo--rank` system so that existing forms keep working as expected. We **do not** want anyone's `kobo--rank` stuff changing unexpectedly to `rank`, because the format of the response data will change significantly.

**I propose that**
if someone… | …then:
-- | --
Adds a new "Ranking" question to a new survey | Use the new XLSForm `rank` type
Adds a new "Ranking" question to an _old_ survey | Use the new XLSForm `rank` type
Edits an old-style `kobo--rank` question in an old survey | Keep using the old-style `kobo--rank` logic
[...]

Issue #553, opened 2016-03-07
Allow making some actions in formbuilder with keyboard


tinok:
## Details

Allow navigating forms using keyboard only as much as possible - to improve accessibility and make usage faster by power users.

Important places:
- Form Builder: user can switch between input fields using the tab key and create a new question without using the mouse.

## Additional details

We need to remove global disabling of `:focus` styles - instead we can try using https://github.com/ten1seven/what-input or just enable them and make prettier than default.

magicznyleszek:
There are multiple different inputs that user would select with keyboard, plus nested groups could require some more visuals - thus "needs-design" label :)

tinok:
@magicznyleszek This is what we had in mind with this: 

![image](https://user-images.githubusercontent.com/673511/79349744-a00c9f80-7f04-11ea-97f0-b0fd19cb53ea.png)

![image](https://user-images.githubusercontent.com/673511/79349974-df3af080-7f04-11ea-87f7-3f48e5f630ea.png)

![image](https://user-images.githubusercontent.com/673511/79350123-0f828f00-7f05-11ea-91f4-0441f8c8c002.png)

duvld:
Note: Adding keyboard actions to choices questions would currently most likely be made in `viewUtils.makeEditable` because of the way pressing "Enter" behaves with `makeEditable` (they have no input fields) 

Since we have an issue (https://github.com/kobotoolbox/kpi/issues/2152) to remove instances of `makeEditable` already, it is probably best to work off of the solution to that issue rather than doing the work twice

magicznyleszek:
I feel that focus styles are good enough and all most needed keyboard navigation is covered. I am closing this, we can make some more advanced changes by creating a new issue :)

PR #2662, opened 2020-05-06
Additional keyboard navigation in formbuilder

### LINKED ISSUE #555: Default options can be replaced with single click
User should be able to replace default option text without having to highlight or delete it (like placeholders)

duvld:
## Description

Main thread: https://www.flowdock.com/app/kobotoolbox/kobo/threads/a571W3FWGd8-BDcwwsK9uPT9EO2

~~PR to https://github.com/kobotoolbox/kpi/tree/2152-drop-makeEditable because these changes depend on using `input` fields instead of `makeEditable` fields. Made seperate PR to differentiate the two branches.~~

⚠️ Note: May have conflicts with eventual PR to add ODKRank type (https://github.com/kobotoolbox/kpi/issues/2420) in this branch: https://github.com/kobotoolbox/kpi/tree/2420-koborank-odkrank-model due to changes in `view.choices.coffee`.

### Changes to be made
Suggested changes: https://github.com/kobotoolbox/kpi/issues/553#issuecomment-614078698

- [x] ENTER and Arrow Key navigation for selecting question
- [x] ENTER navigation for choice questions, pressing ENTER should add new option
- [x] Navigation between labels through tabbing
- [x] ~~Shortcut for adding a new question~~ Navigate to `+` option through tabbing
- [ ] SHIFT-TAB should also show `+` button above last question row
    - Currently not shown due to having `display: none` if not moused over

## One Tab to Rule Them All
Use TAB for navigation across formbuilder.

Current implementation of this user story below:

1. User clicks on [+] to create new question
2. Writes: What is your name? [enter]
3. Presses [right arrow] 3x, [enter] to select text question
4. [+] is highlighted, showing it can be activated with [enter]
5. Presses [enter] > new question box appears
6. Writes: Are you currently displaced? [enter]
7. Presses [enter] to select single select question
8. (first response option is highlighted, meaning user can type to replace placeholder)
9. Writes: Yes [enter]
10. Writes: No [enter]
11. (third response option is highlighted) User doesn't want to add more options, presses [Tab]
12. [+] is highlighted to add another question

553-navigation-story
Added boxes around each tabbable element:
553-tabbing

## Related issues
"ENTER navigation for choice questions, pressing ENTER should add new option" fixes #555 
May have conflicts with eventual PR to add ODKRank type #2420
Part of #553


jsapp/scss/components/_kobo.form-builder.scss
```diff
@@ -343,6 +374,17 @@ $t-form-builder-aside-open: 200ms;
...
     }
+
+    &:focus {
+      outline-color: rgb(77, 144, 254);
+      outline-offset: -2px;
+      outline-style: auto;
+      outline-width: 5px;
+
+      i {
+      outline-color: transparent;
```
magicznyleszek, line 385: indentation

duvld:
> I've found few issues:
> 
> 1. I think that tabbing into `+` to add a question, but only if it is visible, makes it a poor feature. I would  either work more on this here, or create a separete PR to improve it. Maybe make the `+` appear both after and before a question/group if something is focused inside the question/group? Additional complexity happens with nested groups!
> 2. There is a difference between mouse clicking `+` and ENTER activating it:
>    
>    1. MOUSE Screenshot 2020-06-03 at 15 29 59
>    2. ENTER Screenshot 2020-06-03 at 15 29 47
> 3. When adding a question with keyboard, the `+` that appears focused, appears at the end of the form, not just after a question. This can be observed if I have 10 questions, but I add new one at the top, and then `+` appears at the end not just after the one I've added. Similar problem when I keyboard add a question inside a group, I end up with `+` focused outside the group.

For 1. I think I know about this issue and my plan is to approach it with this:
> SHIFT-TAB should also show + button above last question row
> Currently not shown due to having display: none if not moused over

Let me know if this is what you're talking about/what you think :)

For 2. and 3., sometimes old issues are still present for some reason. I think fixed both of these already and they work fine for me on the staging server I set up for this PR: https://kf.tri.kbtdev.org have you tried to recreate them on there as well?

magicznyleszek:
I've found a great css selector that fixes the problems I had with `+` button: https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-within and generally I think this is ready to be merged :ship::it: @duvld do you have any objections to merging this?

duvld:
@magicznyleszek great fix for the `+` button functionality :) functionally I think we're good to go, I did have an issue with the styling on the right side header buttons when they're tabbed:
image
I'm going to try to fix this quickly by removing outlines when focused

PR #2667, opened 2020-05-12
Hide repeat checkbox for matrix

### LINKED ISSUE #2634: Remove repeat checkbox for Question Matrix
## Description

A Matrix question (group) has a "Repeat this group if necessary" checkbox. It is possible to check it, but it doesn't do anything, and Backend doesn't even store it.

Screenshot 2020-04-20 at 17 50 32

We should hide it.

magicznyleszek:
## Description

Took some time to pinpoint the place where it should be disabled :P

## Related issues

Fixes #2634


duvld:
Not specified in the issue (unless I've missed it), so I'm wondering if it was intentional and removed for the same reason for the `Appearance (Advanced)` section is also removed? The code doesn't look like it's supposed to touch that checkbox:

image
(Both checkboxes now missing)

magicznyleszek:
@duvld awesome catch! The problem was that my `return false` was breaking the `for` loop, instead of just moving on. I've pushed a fix for that.

PR #2678, opened 2020-05-15
Fix infinite spinner on submission modal after clicking arrow

### LINKED ISSUE #2669: Single submission screen returns blank when clicking quickly on Next
## Description

When a user quickly clicks on the Next button within the single submission view, a blank modal sometimes appears instead of the next record. Right before the single submission screen comes back blank the main table is reloaded (instead of the single submission). 

## Steps to Reproduce

1. Open first record in table view
2. Quickly click on Next

## Expected behavior 

Second record opens.

## Actual behavior 

Blank modal window without data record.

## Additional details

This doesn't always happen but generally seems to be related to how quickly the user clicks on Next.

magicznyleszek:
## Description

Includes a tiny bit cleanup, sorry!

## Related issues

Fixes #2669


jnm:
@magicznyleszek, we have a problem on `master` with the submission modal—maybe it's related? Imagine that I have four submissions, open the submission modal, and see "1 of 4". If I then click next, I see "2 of 4" :+1:, but then if I click next again, I'm back at "1 of 4". From there, no matter how many times I click next, I remain stuck at "1 of 4" until I close the modal and reopen it.

PR #2718, opened 2020-06-22
Drop IE11 support

### LINKED ISSUE #2716: Handle Internet Explorer 11
Decision: we'll drop support for IE and display a more helpful message.

## Description

The old version of KPI on kobo.humanitarianresponse.info seems to work with IE11:
ocha ie11 Screenshot_2020-06-18_15-10-18

However, the new version `2.020.25` does not load the login screen properly. It allows login but shows a blank page afterwards:
image
image

Exception:
hhi ie11 Screenshot_2020-06-18_15-11-37

## Expected behavior 

Either KPI loads, or the user is redirected to an informative page telling them to use a newer browser. Enketo, for example, has dropped all IE support and shows this message to IE visitors:
image

magicznyleszek:
## Description

Three things here:
- Dropped IE11 from `.browserslistrc` (the file used for determining what browsers to support while building JS files)
- Created a redirect to kf.kobo.local/modern_browsers/ with a message about modern browsers (I copied the message from Enketo)
- Cleaned up some old IE-related things in templates

## Related issues

Fixes #2716


magicznyleszek:
There is not much hardcore changes here, so I allowed myself to merge it 😅

PR #2723, opened 2020-06-26
706 cascade question names

### LINKED ISSUE #706: Cascades change when question name changes
When a user updates the label of an imported cascade question the unique `name` is also updated, but the `choice_filter` value remains as it was originally -- thereby breaking the cascade logic. 

image

duvld:
## Description

Utilize coffeescript model `sluggify` function to match "Choice Filter" value with "Data Column" value before saving form. This syncs up the name changes so that cascade questions don't break.

Future improvements (?):
- Don't let user edit the "Data Column" and "Choice Filter" options or
- Set up front end warnings if user changes something to potentially break this sync (changing "Data Column" value or "Choice Filter" value manually) 

## Related issues

Fixes #706


magicznyleszek:
I found a problem:

1. Import this form [cascade_bug.xls.zip](https://github.com/kobotoolbox/kpi/files/4974291/cascade_bug.xls.zip)
2. Open it in Form Builder
3. Change first question's label to "What is your country?"
4. Click SAVE and reload page
5. Now I see that first question's name is still `country`, but second question has chaned `choice filter:` value to `country=${What_is_your_country}` thus braking it :(

PR #2729, opened 2020-07-16
Handle (undocumented) XLSForm `disabled` column appropriately

jnm:
…when building reports and exports, by upgrading formpack to fix kobotoolbox/formpack#219


jnm:
kpi tests are failing locally
image

jnm:
ok, it should work now that https://github.com/kobotoolbox/kpi/pull/2730 is merged and this is rebased on top of it

PR #2736, opened 2020-07-23
Moved reading CSRF cookie to ajax

### LINKED ISSUE #2717: Impossible to save form after reauthenticating in another tab
I think my change in https://github.com/kobotoolbox/kpi/pull/2588 made this scenario bad:

1. Start editing a form in the form builder;
1. Get logged out somehow (session expires, yours truly messes up sessions on the server, log out in another tab, etc.);
1. Try to save in the form builder;
1. See the error modal with `403: Authentication credentials were not provided.`;
1. Open a new tab in the same browser and log in again;
1. Go back to the tab with the form builder and save again;
1. See the error modal with `Forbidden`;
1. Notice that the server response (not shown in the UI) is `CSRF Failed: CSRF token missing or incorrect.`

Is it possible to detect the CSRF error and retrieve the new CSRF token using an AJAX request?

Related support request: https://community.kobotoolbox.org/t/error-message-while-saving-my-form/10108/6

Another way to reproduce:

1. Start editing a form in the form builder;
1. Open a new tab and visit `/accounts/login/`, which perhaps you've bookmarked;
    * This page always reauthenticates, even if you're already logged in.
1. Return to the tab with the form builder;
1. Try to save and notice that it always fails.

duvld:
## Description

Moved #2728 fix to the ajax request, `x-csrftoken` header now matches `csrftoken` cookie when updated:

image

image

## Related issues

Fixes #2717

### DIFF
--- a/jsapp/js/main.es6
+++ b/jsapp/js/main.es6
@@ -27,14 +27,14 @@ function csrfSafeMethod(method) {
 }
 
 let csrfToken = '';
-try {
-  csrfToken = document.cookie.match(/csrftoken=(\w{64})/)[1];
-} catch (err) {
-  console.error('Cookie not matched');
-}
 
 $.ajaxSetup({
     beforeSend: function(xhr, settings) {
+	try {
+	  csrfToken = document.cookie.match(/csrftoken=(\w{64})/)[1];
+	} catch (err) {
+	  console.error('Cookie not matched');
+	}
         if (!csrfSafeMethod(settings.type) && !this.crossDomain) {
             xhr.setRequestHeader('X-CSRFToken', csrfToken || cookies.get('csrftoken'));
         }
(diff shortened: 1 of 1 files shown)


jsapp/js/main.es6
```diff
@@ -27,14 +27,14 @@ function csrfSafeMethod(method) {
...
-  console.error('Cookie not matched');
-}
 
 $.ajaxSetup({
     beforeSend: function(xhr, settings) {
+	try {
+	  csrfToken = document.cookie.match(/csrftoken=(\w{64})/)[1];
+	} catch (err) {
+	  console.error('Cookie not matched');
+	}
```
jnm, line 37: sorry, this seems great, but could you fix the indentation?

PR #2742, opened 2020-07-30
Send translation settings with saved rows to question library

### LINKED ISSUE #2731: First translation name is removed when adding question to library from formbuilder
## Description

When adding a question to the library containing multiple translations (label languages) the POST request does not include the name of the first translation. 

## Steps to Reproduce

1. Upload this form ([d004.xlsx](https://github.com/kobotoolbox/kpi/files/4936839/d004.xlsx))
2. Open it in the formbuilder
3. Click on the 'add to library' button
4. Check the Form Data that was sent in the network POST request

## Expected behavior 

Question is added to the library containing the same translations (label::Français, label::Sango). 

## Actual behavior 

The question stored to the library only includes `label`and `label::Sango`. See form data:

image

duvld:
## Description

Forms with translations could not transfer their translation settings when saving a row to library. This was due to [`unnullifyTranslations`](https://github.com/kobotoolbox/kpi/blob/fbc6b6b3609a22fc66f44cfe2af65b523cd705cb/jsapp/js/utils.es6#L71) not being called when saving a single row to the question library.

This PR fixes this issue by sending the entire survey to `unnullifyTranslations` then _fishing_ out the row being saved and sending said row and its translation to the question library. This could be potentially heavy if there is a large form - if so adding a single row version of `unnullifyTranslations` may be needed.

## Related issues

Fixes #2731

### DIFF
--- a/jsapp/js/models/surveyScope.es6
+++ b/jsapp/js/models/surveyScope.es6
@@ -2,25 +2,28 @@ import {actions} from '../actions';
 import {
   notify,
   t,
+  unnullifyTranslations,
 } from '../utils';
 
 class SurveyScope {
   constructor ({survey}) {
     this.survey = survey;
   }
-  add_row_to_question_library (row) {
+  add_row_to_question_library (row, assetContent) {
     if (row.constructor.kls === 'Row') {
       var rowJSON = row.toJSON2();
       let content;
       if (rowJSON.type === 'select_one' || rowJSON.type === 'select_multiple') {
         var surv = this.survey.toFlatJSON();
         var choices = surv.choices.filter(s => s.list_name === rowJSON.select_from_list_name);
-        content = JSON.stringify({
-          survey: [
-            row.toJSON2()
-          ],
-          choices: choices || undefined
-        });
+        // content = JSON.stringify({
+        //   survey: [
+        //     row.toJSON2()
+        //   ],
+        //   choices: choices || undefined
+        // });
+        content = unnullifyTranslations(JSON.stringify(surv), assetContent);
+        console.log(content);
       } else {
         content = JSON.stringify({
           survey: [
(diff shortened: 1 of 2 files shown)


jsapp/js/models/surveyScope.es6
```diff
@@ -2,31 +2,51 @@ import {actions} from '../actions';
...
       let content;
+      var surv = this.survey.toFlatJSON();
+      /*
+       * Apply translations "hack" again for saving single questions to library
+       * Since `unnullifyTranslations` requires the whole survey, we need to
+       * fish out the saved row and its translation settings out of the unnullified return
+       */
+      var unnullifiedContent = JSON.parse(unnullifyTranslations(JSON.stringify(surv), assetContent));
+      var settings_obj = unnullifiedContent.settings;
+      var survey_obj = unnullifiedContent.survey;
```
magicznyleszek, line 24: A minor thing - let's try to keep JavaScript variables in `camelCase` :)

jsapp/js/models/surveyScope.es6
```diff
@@ -2,31 +2,51 @@ import {actions} from '../actions';
...
-        });
+        var choices = unnullifiedContent.choices.filter(s => s.list_name === rowJSON.select_from_list_name);
+        for (var i in survey_obj) {
+          if (survey_obj[i].$kuid == row.toJSON2().$kuid) {
+            content = JSON.stringify({
+              survey: [
+                survey_obj[i]
+              ],
+              choices: choices || undefined,
+              settings: settings_obj || undefined
```
magicznyleszek, line 34: Do you really need those `|| undefined`? Would these variables be ever something like `null`, or `""`?

magicznyleszek:
@duvld also feel free to merge this PR when you fix these few minor things :)

PR #2753, opened 2020-08-13
Extract front-end translatable strings

### LINKED ISSUE #2752: Extract all translatable strings from front-end code
Related to https://github.com/kobotoolbox/form-builder-translations/issues/3, we want to automatically extract strings from the UI code without [a manual procedure that uses a desktop application](https://github.com/kobotoolbox/form-builder-translations#kpi-ui-translations). This will cut down on developer busywork and the long delay between application changes and updated translations.

1. `babel-loader` loads ES6 sources and `coffee-loader` loads the CoffeeScript sources (i.e. form builder);
1. We avoid calls to `t()` being turned into things like `_utils__WEBPACK_IMPORTED_MODULE_8__["t"])` by **not** importing `t` but using [`webpack.ProvidePlugin`](https://webpack.js.org/plugins/provide-plugin/) instead;
1. We use https://github.com/grassator/webpack-extract-translation-keys, which hooks into [webpack's JavascriptParser](https://webpack.js.org/api/parser/) to extract all static strings passed to `t()`;
1. We process the resulting file into something useful for upload to Transifex (TBD; let's address as part of https://github.com/kobotoolbox/form-builder-translations/issues/3).

In the future, when all code is ES6 or beyond, we may be better served by https://github.com/ttag-org/ttag.

jnm:
…to the file `jsapp/compiled/extracted-strings.json` when `npm run build` / `webpack --config webpack/prod.config.js` is executed. Going forward:

* `t()` must be the function call used to surround translatable strings, i.e. do not assign `t` to another variable and expect string extraction to continue working;
* call `t()` only with static strings;
* do not import `t()` manually anywhere, e.g. `import {t} from './utils'` must be avoided: `t()` is now provided globally by `webpack.ProvidePlugin`.

Closes #2752


jsapp/js/components/searchcollectionlist.es6
```diff
@@ -162,14 +162,19 @@ class SearchCollectionList extends Reflux.Component {
     if (this.state.searchResultsDisplayed)
       searchResultsBucket = 'searchResultsCategorizedResultsLists';
 
+    const categoryLabels = {
```
magicznyleszek, line 165: I see these are being used in two places, so maybe put'em in `constants.es6`?

magicznyleszek:
@jnm do you want to fix the merge conflict? (and one thing from review :-P)

magicznyleszek (approved):

PR #2559, opened 2020-01-24
Feature: Public collections

magicznyleszek:
~~[don't forget, @jnm, to do #2795!]~~

## Description

This is the big feature from https://community.kobotoolbox.org/t/public-collections-design-improvement/2284/12

Things TODO I know of:
- [x] :bug: long word in asset name breaks header and list view (e.g. `dsadasdasdasdasdasdasdasdasdasdasdasdasjdkhakjshdkjahsdkjahskdjhaksjhdakjshdakjshdkajshdkjahskjdhaksjdhkajshdkajshkdjahskdjahksjhdajshkdahskjdhakjsdhkajshdkajshdkjahskjdhaksjdhkajshdkjahskdjhakjsdhkajshdjkasdhakjshdkjahsdhasdhkahdasdasdasdaasdasd12345`)
- [x] list view could use a "reset ordering and filtering" button somewhere
- [x] data stores should probably fetch completely new data when actions cause changes to assets that  could influence metadata content (I already try to do some magick to avoid fetching the big list as much as possible)
- [x] mobile looks of the list view
- [ ] :bug: using hot reload in dev env causes the list view to not reload properly (infinite spinner)
- [x] :bug: list view needs right padding to account for the scrollbar (display scrollbar always?, detect scrollbar width?)
- [x] use metadata endpoint when possible to avoid fetching whole list of assets with metadata
- [x] add fullscreen toggle button

Future ideas:
- advanced filtering in the search box in header
- modal to choose/confirm collection parent when creating/cloning from within collection
- no value option in filters (metadata)
- uploading a file on collection page should make the upload have collection as parent

## Related issues

Fixes #2293
Fixes #1282 
Fixes #1860 
Fixes #2157 
Fixes #2586 
Fixes #280 
Fixes #2559 
Fixes #2783
Fixes #2813

Fixes partially #1983

Closes #2332
Closes #2381

PR #2774, opened 2020-08-31
Cleanup colors to prepare for design system

### LINKED ISSUE #2790: Introduce pressed-in state for buttons
## Description

Our buttons mostly doesn't change visually when pressed in, so user doesn't get a feedback that the button was pressed successfully. Our (me and @Ig-Rebollo) initial idea is to use a mix of moving button small distance down and change the shadow when pressed. It is possible to see it in action here: https://jsfiddle.net/smutnyleszek/3eLsj8zh/7/.

## Additional note

Hover state should be also changed to use the current color darkened by `10%` (so using https://sass-lang.com/documentation/modules/color#darken `color.scale($some_color, $lightness: -10%);`)

magicznyleszek:
Things done here:
- all misc unnamed colors are now variables
- merged similar colors
- moved colors to separate file
- removed some unused styles and code linked to them

Note: merged this into #2559

Part of #2776 
Fixes #2790


magicznyleszek:
@Ig-Rebollo using `lightness` is a quick way of getting things done ;-) I changed some of them to defined colors. I agree it is not good to overuse them. I renamed colors to better names and merged more of them. Please take a look and I think we're close to merging this to `master` :-)

Ig-Rebollo (commented):
Looking great @magicznyleszek. Just wondering if it would be much work to merge $cool-gray, $cool-silver, $c-ghost and $c-iron into just one mid-range gray, for instance: #CBCED7. 

No text should ever go lighter than $cool-darkgray, so the text currently using $cool-gray or $cool-silver should be switched up to $cool-darkgray. (ultimately we want only 4 options for text-> either $black, $cool-darkgray, or white and $cool-whitegray if the background is dark). 

Borders using these grays though, could be switched to that mid-range gray from above. Backgrounds using $c-iron could go down to use $cool-lightgray.

I'm not sure if this might cause some side effect I overlooked, but hopefully this would also help increasing contrast between text and backgrounds... what do you think?

The accent colors are perfect as you have them :)

Issue #2262, opened 2019-04-26
Add (failing) unit test for library imports


jnm:
(second sim bug)

- [x] add failing test
- [x] fix it

Broken by #2000 / the fix for https://github.com/XLSForm/pyxform/issues/229

Hint: https://github.com/kobotoolbox/kpi/blob/0cde9029f37f3a8bac11f18d6f528801aefd4cca/kpi/models/import_export_task.py#L288

noliveleger:
Just try to take a look. It's indeed broken.
PyXForm skips sheets whose names are not part of `constants.SUPPORTED_SHEET_NAMES`
https://github.com/XLSForm/pyxform/blob/8fb7934886a1e248a39eb802061e618ebceb60a4/pyxform/xls2json_backends.py#L209-L218

Tried to monkey-patch PyXform to add `library` but it still fails a little bit further.

jnm:
pasted from https://www.flowdock.com/app/kobotoolbox/kobo/threads/x-uUMo8M2PfHGoC04PEBcDlmKz4, in the hope that it's helpful to future wanderers:

the import task is being run by Celery, which does not automatically reload when the python code changes and also logs to a separate file.
to get around that, this should work:

    ./run.py -cf exec kpi bash
    sv stop uwsgi
    ps faux # or however you like to list your processes, to make sure that runserver has actually stopped
    # kill runserver if it's still running
    SKIP_CELERY=True ./manage.py runserver_plus 0:8000

now the celery tasks should run inside the `manage.py` process instead of the celery worker processes, and your code changes should take effect right away. `print` statements and the like should appear on the screen along with all the other `runserver` output

jnm:
here's my hack. Anji is working on a better solution\*

    diff --git a/kpi/models/import_export_task.py b/kpi/models/import_export_task.py
    index ead3644a..d2ad55aa 100644
    --- a/kpi/models/import_export_task.py
    +++ b/kpi/models/import_export_task.py
    @@ -665,7 +665,22 @@ class ExportTask(ImportExportTask):
     
     def _b64_xls_to_dict(base64_encoded_upload):
         decoded_str = base64.b64decode(base64_encoded_upload)
    -    survey_dict = xls2json_backends.xls_to_dict(BytesIO(decoded_str))
    +    #survey_dict = xls2json_backends.xls_to_dict(BytesIO(decoded_str))
    +    from collections import OrderedDict
[...]

PR #2798, opened 2020-09-16
Fix bulk library imports

duvld:
## Description

Fixed library upload failure by introducing a "less strict" xls to dict converter.

Refactored sheet converters to a utils function (outside of manager files).

## Related issues

Second checkbox of #2262

### DIFF
--- a/kpi/utils/sheet_converter.py
+++ b/kpi/utils/sheet_converter.py
@@ -8,6 +8,7 @@ from pyxform.errors import PyXFormError
 from pyxform.utils import basestring, unichr, unicode
 from xlrd import XLRDError
 
+
 def convert_xls_to_dict(xls_file_object, strip_empty_rows=True):
     """
 
@@ -51,6 +52,7 @@ def convert_xls_to_dict(xls_file_object, strip_empty_rows=True):
         ss_structure[sheet_name] = sheet_contents
     return ss_structure
 
+
 def convert_xls_to_ss_structure(xls_file_object, strip_empty_rows=True):
     """
     The goal: Convert an XLS file object to a CSV string.
@@ -91,8 +93,10 @@ def convert_xls_to_ss_structure(xls_file_object, strip_empty_rows=True):
         ss_structure[sheet_name] = sheet_contents
     return ss_structure
 
+
 def _iswhitespace(string):
     return isinstance(string, str) and len(string.strip()) == 0
+
 
 def _xls_value_to_unicode(value, value_type):
     """
@@ -124,3 +128,4 @@ def _xls_value_to_unicode(value, value_type):
         # to avoid this issue:
         # https://github.com/modilabs/pyxform/issues/83
         return str(value).replace(chr(160), ' ')
+


Kalyan-Lama:
**Tested the feature of uploading an xlsform to the library and observed the following:**

- [x] xlsform upload possible with the `library` tab
- [x] xlsform upload possible with the `survey` tab
- [ ] When uploading questions from an xlsform the questions seems to be inversely positioned in the question library when compared to those that are positioned in the xlsform. e.g. in an xlsform the questions are ordered as Q1, Q2, Q3 … while in the library they are ordered as … Q3, Q2 and Q1 respectively.

_Question order as seen in an xlsform:_
Image 1
_Question order as seen in the library:_
Image 2
_Reference xlsform for testing:_
[collection_small_questions.xlsx](https://github.com/kobotoolbox/kpi/files/5254607/collection_small_questions.xlsx)

- [ ] When uploading questions (from inside a groups) from an xlsform the questions seems to be inversely positioned in the question library when compared to those that are positioned in the xlsform. e.g. in an xlsform the questions are ordered as Q1, Q2, Q3 … while in the library they are ordered as … Q3, Q2 and Q1 respectively. Additionally, it also considers begin_group and end_group a distinct question.

_Question order as seen in an xlsform:_
Image 3
_Question order as seen in the library:_
Image 4
_Reference xlsform for testing:_
[collection_small_group.xlsx](https://github.com/kobotoolbox/kpi/files/5254660/collection_small_group.xlsx)

- [x] When uploading block questions (with the column header `block`) from an xlsform, the questions are positioned in the exact way as they are in the xlsform i.e. the questions are ordered as Q1, Q2, Q3 in both the xlsform as well as the library.

_Question order as seen in an xlsform:_

_Question order as seen in the library:_

_Reference xlsform for testing:_


_Question order as seen in an xlsform:_

_Question order as seen in the library:_

_Reference xlsform for testing:_

jnm:
Thank you, @Kalyan-Lama, for the extensive testing. I'm glad that many things are working properly. Others, where behavior seems questionable or wrong, likely merit their own issues. This particular pull request was intended to address only the inability to bulk-import questions and blocks into the library by uploading files with a `library` worksheet.

> When uploading questions from an xlsform the questions seems to be inversely positioned in the question library when compared to those that are positioned in the xlsform. e.g. in an xlsform the questions are ordered as Q1, Q2, Q3 … while in the library they are ordered as … Q3, Q2 and Q1 respectively.

This is not very intuitive, but let's address it in a separate issue if needed. Right now, the `library` sheet is a shortcut for uploading multiple library items at once: think of it as a quicker alternative to uploading Q1, Q2, and Q3 as separate XLSForm files (each with a `survey` sheet). The server processes them in this way, sequentially, creating separate library items ("assets") from top to bottom. Once done, the library then displays the newest / most recently modified items first, and in this case, Q3 is the newest item because it was imported last.

> When uploading questions (from inside a groups) from an xlsform the questions seems to be inversely positioned in the question library when compared to those that are positioned in the xlsform. e.g. in an xlsform the questions are ordered as Q1, Q2, Q3 … while in the library they are ordered as … Q3, Q2 and Q1 respectively. Additionally, it also considers begin_group and end_group a distinct question.

It only makes sense to use groups in conjunction with a `block` column. Without a `block` column, each row is treated as a separate item; as described above, think of it as uploading an XLSForm (with a `survey` sheet) that has only a single `begin_group` question. That's not something that would be terribly meaningful. Arguably, we should reject it outright, but that should be a separate issue if it's causing people grief.

> The system also allows users to upload a question (either select one or select multiple) with no choices in the choices tab. Interestingly the system creates a hypothetical choice viz. Option 1 and Option 2 with the XML value as AUTOMATIC.

Ideally, yes, we should have a warning about this. However, "Option 1" and "Option 2" are not added during import: you can verify this by looking at the library item's JSON, e.g. https://kf.beta.kbtdev.org/#/library/asset/aDxC8o5cXyLXJR85YLtNFJ/json. These choices are actually added by the form builder. I tested this and found the same behavior when uploading a regular (non-library) survey. This could be filed as a separate issue, but given more pressing work we have to do, I would defer this until we receive complaints from users.

> Tags present in the library is not seen in the xlsform while downloading them back to the xlsform from library.

That's okay. We allow tags on batch imports to save people time, but we don't yet have a way to do export multiple items at once—and tags would have nowhere to go in an individual XLSForm export.

> Unable to add xlsform with matrix question (designed using begin_kobomatrix end_kobomatrix)

These are a special type of group, and, as above, would only make sense if used in conjunction with a `block` column. Please open a separate issue if the import fails when using a `block` column, as this is something we should fix, although it's perhaps not high-priority.

> Able to upload a matrix question from the xlsform to the library.…But the problem is that it changes the variable name of the matrix question…

Let's tackle this as a separate issue. I would need an example to understand this properly, or to work through it by myself, which I haven't had time to do.

> Project library does not seem to be checking the syntax errors within the forms.

Among the following, do you find syntax checking inconsistent?

* Importing via the `library` sheet
* Importing, still into the library, via the `survey` sheet
* Importing into the (non-library) projects list

I know some syntax problems are not detected until deployment, but if the above aren't consistent, please open a new issue.

Kalyan-Lama:
**My observations while testing the Library (Public Collections):**

- [x] Upload an xlsformwith the `library` tab
- [x] Download as xlsform (block)
- [x] Download as xlsform (question)
- [x] Download as xml (question)
- [x] Deleting question(s)
- [x] Deleting block
- [x] Cloning a block/question
- [x] Managing translation
- [ ] Edit Tags **(Remarks: could not edit tags as the changes did not save)**
- [x] Edit question/blocks through form builder
- [x] Making blocks public
- [x] Subscribing to a public collection
- [x] Able to unsubscribe to a public collection
- [x] Moving questions/blocks to collection
- [x] Moving questions/blocks from one collection to another collection
- [x] Sharing collections with other users (view collection access)
- [x] Sharing collections with other users (edit collection access)
- [ ] Sharing collections with other users (edit collection access) **Remarks:** _Could not delete a shared question. Seeing the following error message:_

image

But `Remove from collection` is however possible. Maybe if this is the case, we need not display the `Delete` as shown in the image below:

image

- [x] Sharing collections with other users (edit collection access) **Remarks:** _Able to add new questions on the block; able to clone a question/block; able to manage translation_

- [ ] **Designing survey project from a library:** When pulling a question from the library, the labels do not show in the form builder UI. The screen is seen something like this:

image

But interestingly, the question label is seen clearly while previewing:

image

But then if you look at the `Form Preview` (shared above) you will also see the choices missing. However, once they are deployed and viewed through `Enketo` everything seems to be normal:

image

- [x] **Search Library>Search (seen in the forum builder UI):** Able to search anything that is within questions (assets) that is seen in the screen i.e. question labels are seen but the choices are not seen/searchable.

- [ ] **Search Library>Search Tags (seen in the forum builder UI):** The search does not show up anything.

- [x] **Search Library>Select Collection Name (seen in the forum builder UI):** Users are able to filter the questions/blocks from the selected collections (if there are multiple collections in the library).

- [x] **Search Library>expand details (seen in the forum builder UI):** Users are able to see the details (questions/tags) when selected.

jnm:
Further discussion about separate issues that may need to be opened as a result of this broad testing is at https://chat.kobotoolbox.org/#narrow/stream/4-KoBo-Dev/topic/Public.20Collection.20feedback/near/5375

jnm:
How was this merged?

1. I changed the base branch for this PR from `master` to `2262-failing-library-upload-test`
1. https://github.com/kobotoolbox/kpi/commit/45d1a5115ef3221ff534e1045f515c98eb523b53 merged this PR's branch into `2332-collection-as-asset-type`
1. I reverted that merge (https://github.com/kobotoolbox/kpi/commit/0d0d9040617daa418273a67db20f136eea70e50b), but that did not remove the commits in this PR
1. I rebased `2262-failing-library-upload-test` on top of `2332-collection-as-asset-type`

In short, all the commits from this PR now exist in the base branch—but they were all reverted—meaning the PR had no ultimate effect.

PR #2805, opened 2020-09-23
[Public collections] Update query parser parameters for front end

noliveleger:
Some changes in the back-end code needed the front-end code to be updated as well.
- `status` is not part of `q` parameter anymore. 
- `parent__uid:null` makes an extra query to DB. To speed-up the response of the back end, `parent:null` is preferable.  

_NB: `parent__uid:<uid>` cannot be avoided because FE doesn't have the parent's pk_

### DIFF
--- a/jsapp/js/dataInterface.es6
+++ b/jsapp/js/dataInterface.es6
@@ -384,6 +384,10 @@ export var dataInterface;
         searchData.metadata = 'on';
       }
 
+      if (params.status) {
+        searchData.status = params.status;
+      }
+
       return $ajax({
         url: `${ROOT_URL}/api/v2/assets/`,
         dataType: 'json',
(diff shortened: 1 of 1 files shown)


jsapp/js/dataInterface.es6
```diff
@@ -421,26 +429,28 @@ export var dataInterface;
...
     searchPublicCollections(params = {}) {
+      params['status'] = 'public-discoverable';
       return this._searchAssetsWithPredefinedQuery(
         params,
-        `${COMMON_QUERIES.get('c')} AND status:public-discoverable`,
+        COMMON_QUERIES.get('c'),
       );
     },
     searchPublicCollectionsMetadata(params = {}) {
+      params['status'] = 'public-discoverable';
```
magicznyleszek, line 450: I would only change `params['status']` to `params.status` in both places :)

magicznyleszek (changes requested):

magicznyleszek (approved):

PR #2806, opened 2020-09-23
Do not add dicts with meta types

### LINKED ISSUE #2801: Meta question types get added as separate assets when uploaded to the library
## Description

When uploading a collection to the Library, each `meta` question (start, end etc.) gets added with it as a separate asset. 

`begin/end_group` tags should also be rejected. We should also tell the user if they want to upload groups of questions to the Library, these tags will only work for library imports if there is a block column preceding it:
image

This can be an issue since meta question types are always added when downloading a form from KPI.

## Steps to Reproduce

1. Download a xlsform from KPI
2. Change the `survey` sheet to `library`
3. Upload the form to the Library

## Expected behavior 

Only non meta questions are added to the Library 

## Actual behavior 

Every question (including start, end, and calculate containing versions) is added to the form as a separate asset

## Additional details

Here's a form that can be the simplest recreation of this issue
[basic-text.xlsx](https://github.com/kobotoolbox/kpi/files/5241604/basic-text.xlsx)

duvld:
## Description

Added a list of strings containing meta question types, and a check before adding to dictionary if that row contains a meta type.

## Related issues

Fixes #2801

### DIFF
--- a/kpi/utils/sheet_converter.py
+++ b/kpi/utils/sheet_converter.py
@@ -8,6 +8,9 @@ from collections import OrderedDict
 from pyxform.errors import PyXFormError
 from pyxform.utils import basestring, unichr, unicode
 
+META_TYPES = ['start', 'end', 'today', 'deviceid', 'username', 'subscriberid',
+              'simserial', 'phonenumber', 'email', 'audit']
+
 def _iswhitespace(string):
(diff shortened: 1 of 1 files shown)


duvld:
Closing as the related issue is closed

PR #2807, opened 2020-09-23
Added check to reject rows that contain version

### LINKED ISSUE #2802: A downloaded form's version is added to library as a separate asset
## Description

When downloading a form from KPI, the form contains a calculate question labelled `__version__`. Uploading a collection to the library containing this creates another asset with this calculate question.  

## Steps to Reproduce

1. Download a form from KPI
2. Form should have a calculate question labelled `__version__`
3. Upload the form to the Library

## Expected behavior 

The version id is not made into a separate asset alongside the questions in a collection

## Actual behavior 

A calculate asset containing the version id shows up in the Library

duvld:
## Description

Added small check for `calculate` question types that contain a form's version
## Related issues

Fixes #2802

### DIFF
--- a/kpi/utils/sheet_converter.py
+++ b/kpi/utils/sheet_converter.py
@@ -11,6 +11,12 @@ from pyxform.utils import basestring, unichr, unicode
 def _iswhitespace(string):
     return isinstance(string, str) and len(string.strip()) == 0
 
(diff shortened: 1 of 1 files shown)


jnm:
Now handled by #4118

PR #2819, opened 2020-10-04
Fix how integer validation handles parentheses

enjeyw:
## Checklist

3. [X] Ensure the tests pass

## Description

Got caught out by this one during the week, so figured it was worth fixing.

Currently the validation criteria
`(. = 123.456)` will be parsed as `{name:'.', operator:'resp_equals', response_value: '123.456'}`
but
`(. = 123)` will be parsed as `{name:'.', operator:'resp_equals', response_value: '123)'}`.

That is, the closing parentheses sneaks its way into `response_value`. Only behaves this way for integers, not for dates, strings or decimal numbers.

This PR fixes this, and adds a bunch of tests onto the `validationLogicParser` (currently just on `equalityCriterionPattern`).

Thought about refactoring out the criterion so they could have their own unit tests, but it seemed overkill...

### DIFF
--- a/jsapp/xlform/src/model.validationLogicParser.coffee
+++ b/jsapp/xlform/src/model.validationLogicParser.coffee
@@ -8,7 +8,7 @@ module.exports = do ->
         (?:                              #   start of a non-matching group
           date\(\'\d{4}-\d{2}-\d{2}\'\)) #     something resembling a date: date('xxxx-xx-xx')
         |                                #   or
-          (?:-?(?:\d+\.\d+|\.\d+|\d+.?)) #     a signed integer or decimal
+          (?:-?(?:\d+\.\d+|\.\d+|\d+))   #     a signed integer or decimal
         |                                #   or
           (?:\'[^']+\')                  #     a string surrounded by single quotes (careful: the quotes are included in the match!)
       )                                  # end of the value group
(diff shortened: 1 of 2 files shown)


jnm:
Hey, thanks for catching this! Looks like a mistake of mine. I'm curious, though: looking at `(?:\d+\.\d+|\.\d+|\d+.?)`, I think my intent (5 years ago :sweat_smile:) was to allow all of:
1. Decimals expressed as `123.456`
2. Decimals expressed as `.456`
3. Integers expressed (albeit strangely) as `123.`
4. Integers expressed as `123`

In other words, my mistake was failing to escape the `.` that handles cases 3 and 4. If this is sensible to you, could you modify your PR so that the fateful line reads:
```coffeescript
(?:-?(?:\d+\.\d+|\.\d+|\d+\.?)) #     a signed integer or decimal
```
…and include a test for case 3 above?

Also, I don't know if you've seen it, but the skip logic parser has exactly the same regex, with exactly the same mistake:
https://github.com/kobotoolbox/kpi/blob/46c8086438ee1ade1deb4a66c73383a7327d261a/jsapp/xlform/src/model.skipLogicParser.coffee#L21

:thinking: Maybe this duplication is what you were referring to when you wrote:
> Thought about refactoring out the criterion so they could have their own unit tests, but it seemed overkill...

I'd be great to fix that as well :pray:

Anyway, whatever you can contribute is most welcome, even if you leave it at what you have already and wait for us to take over. Thanks very much for the well-tested and well-documented contribution.

enjeyw:
sure thing @jnm - will modify this weekend :)

jnm:
@p2edwards, bringing this generous outside contribution up to the point where it's ready to merge would be helpful. We can live with the code duplication for now; the parser fixes and the tests are the most important part.

Issue #2800, opened 2020-09-17
`name` of a library question is used instead of translated labels when that question is added to a form


duvld:
## Description

When editing a form (either from Projects or Library), and dragging a question from the library, the `name` of the question from the xlsform is used rather than it's label if that label is a translation, i.e., `label::English(en)`.

These questions show up properly in the Library, but the bug occurs when trying to add the questions to a form.

## Steps to Reproduce

1. Upload a collection to the library containing labels with translations (question has correct title in the Library)
2. Add that question to a form from the formbuilder

## Expected behavior 

The title of the question is the label

## Actual behavior 

The `name` of the question is used instead of it's label

## Additional details

Here's the simplest form that would produce the bug:
[name-used-not-label.xlsx](https://github.com/kobotoolbox/kpi/files/5241575/name-used-not-label.xlsx)

Here's a more "real" example to test on:
https://github.com/kobotoolbox/kpi/files/5168787/collection_small.xlsx

jnm:
Found #2821 while working on this

PR #2820, opened 2020-10-05
Prevent surveys from appearing in form-builder library sidebar (fix for #1 outlined in #2800)

duvld:
## Description

Fixes issue where saving your form in formbuilder makes the current form show up in library assets sidebar, does so by rejecting `survey` asset types.

## Related issues

Fixes 1. of #2800

### DIFF
--- a/jsapp/js/editorMixins/assetNavigator.es6
+++ b/jsapp/js/editorMixins/assetNavigator.es6
@@ -10,6 +10,8 @@ import {searches} from '../searches';
 import ui from '../ui';
 import mixins from '../mixins';
 
+import { ASSET_TYPES } from '../constants';
+
 import { t } from '../utils';
 
 import {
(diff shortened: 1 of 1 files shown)


jsapp/js/editorMixins/assetNavigator.es6
```diff
@@ -95,13 +95,14 @@ class AssetNavigatorListView extends React.Component {
...
-
       return (
         <bem.LibList m={['done', isSearch ? 'search' : 'default']} ref='liblist'>
           {list.map((item)=> {
             var modifiers = [item.asset_type];
             var summ = item.summary;
-            if (summ.row_count == undefined) {
+            // Library questions can only be of `question` or `block` types
+            // Reject `survey` types to not include current survey on save
+            if (summ.row_count == undefined || item.asset_type == 'survey') {
```
magicznyleszek, line 105: Please use `ASSET_TYPES.survey.id` from `constants.es6`

magicznyleszek: And please add a comment stating that this is a `HACK FIX`, as ideal solution would be to not search for `survey`s in the first place - but I understand that it would require too much work influencing too many places in the code

magicznyleszek: Oh and please use triple equals to check for `asset_type` :)

magicznyleszek (changes requested):

Issue #2821, opened 2020-10-05
`unnullifyTranslations()` assumes `translated` cannot change


jnm:
Consider that we have a form with one named language, `French (fr)`, and only labels—**no hints**:
```json
{
    "schema": "1",
    "survey": [
        {
            "type": "integer",
            "$kuid": "kv4jl55",
            "label": [
                "Au contraire!"
            ],
            "required": false,
            "$autoname": "Au_contraire"
        }
    ],
    "settings": {
        "default_language": "French (fr)"
    },
    "translated": [
        "label"
    ],
    "translations": [
        "French (fr)"
    ]
}
```

If we open this in the form builder, add a hint ("My brother"), and save it:
1. [`unnullifyTranslations()`](https://github.com/kobotoolbox/kpi/blob/b14562ed65f97453ed16e7641fb644aa31dfcacc/jsapp/js/utils.es6#L81-L100) will see that `translated` in the asset content contains only `label`;
1. It will change `label` to `label::French (fr)`, but it will **not** alter `hint`;
1. The front end will `PATCH` the asset with this content:
    ```json
    {
        "settings": [
            {
                "default_language": "French (fr)"
            }
        ],
        "survey": [
            {
                "$autoname": "Au_contraire",
                "$kuid": "kv4jl55",
                "hint": "My brother",
                "label::French (fr)": "Au contraire!",
                "required": "false",
                "type": "integer"
            }
        ]
    }
    ```
1. The back end ([formpack](https://github.com/kobotoolbox/formpack/blob/52f77e3519fd1079ce3451f9a5c7c27002e9f3f0/src/formpack/utils/expand_content.py#L77-L82)) will now treat the form as having two translations: `null` (from the bare `hint`) and `French (fr)`:
    ```json
    {
        "schema": "1",
        "survey": [
            {
                "hint": [
                    null,
                    "My brother"
                ],
                "type": "integer",
                "$kuid": "kv4jl55",
                "label": [
                    "Au contraire!",
                    null
                ],
                "required": false,
                "$autoname": "Au_contraire"
            }
        ],
        "settings": {
            "default_language": "French (fr)"
        },
        "translated": [
            "hint",
            "label"
        ],
        "translations": [
            "French (fr)",
            null
        ]
    }
    ```

What should happen? `unnullifyTranslations()` must recognize that changes made by the form builder can cause `translated` to become invalid. We should probably use a constant list of translatable columns when unnullifying a translation.

PR #2823, opened 2020-10-06
Copy library asset content before parsing it

### LINKED ISSUE #2822: Library items are mutated when added to a survey from the sidebar
OK, we know this seems bad:
https://github.com/kobotoolbox/kpi/blob/d3d39dbc25d64f35e90125b0b58004e63176a649/jsapp/xlform/src/model.inputParser.coffee#L111-L121

and in certain circumstances, it is indeed:
1. Here, we load a library asset from the back end, but we save it in a store—so subsequent loads won't refresh it:
    https://github.com/kobotoolbox/kpi/blob/0deff9668027e73ecc7199132cee908ce25a41b8/jsapp/js/stores.es6#L277-L282
1. [`Survey.loadDict()`](https://github.com/kobotoolbox/kpi/blob/0deff9668027e73ecc7199132cee908ce25a41b8/jsapp/xlform/src/model.survey.coffee#L333) calls `inputParser.parse()` (see above), which mutates `asset.content`: it nullifies the default/first translation and causes subsequent import attempts to fail (see #2800).

My proposal: let's have `addExternalItemAtPosition()` make a copy of `asset.content` before doing anything with it. Easy enough; I'll make a PR. @magicznyleszek, are there tests for this kind of thing?

jnm:
…when importing a library asset into a survey. Fixes #2822 [the GitHub `#` auto-completer is not my friend]

Asking for a review from @duvld, since we worked on this together, but will give the final :+1::-1: to fresh-eyed @magicznyleszek.

## Checklist

1. :question: If you've added code that should be tested, add tests
2. ~~If you've changed APIs, update (or create!) the documentation~~
3. :question: Ensure the tests pass
5. ~~If this is a big feature, make sure to prefix the title with `Feature:` and add a thorough description for non-dev folk~~

### DIFF
--- a/jsapp/js/stores.es6
+++ b/jsapp/js/stores.es6
@@ -16,6 +16,7 @@
 
 import Reflux from 'reflux';
 import {Cookies} from 'react-cookie';
+import clonedeep from 'lodash.clonedeep';
 import dkobo_xlform from '../xlform/src/_xlform.init';
 import {parsed, parseTags} from './assetParserUtils';
 import {actions} from './actions';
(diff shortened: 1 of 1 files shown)


duvld (approved):
:+1: looks good, I cannot produce a null translation from adding questions from the library.
Tried to generate a null translation by adding the following into the form and saving: 
  - blocks with multiple translations, 
  - questions with multiple translations, 
  - questions that have form's default language but not as default,
  - different combinations of blocks and groups with translations

The only way I can generate a null translation in the library is through #2821, or adding translated library questions to a form with no named translations (both are beyond the scope here)

In regards to tests, (in a quick search) there's not many that test [`loadDict()`](https://github.com/kobotoolbox/kpi/blob/0deff9668027e73ecc7199132cee908ce25a41b8/test/xlform/model.tests.coffee#L163-L209) but there's a whole suite that tests `inputParser` wrt translations: https://github.com/kobotoolbox/kpi/blob/master/test/xlform/inputParser.tests.coffee that are run by travis.

magicznyleszek:
@jnm I love the comments you've added <3

duvld:
No need for https://github.com/kobotoolbox/kpi/blob/43b5ef70b9d6e25cdf85f46e73d8c22dd5466765/jsapp/xlform/src/model.inputParser.coffee#L117-L118 anymore now 😉

jnm:
Thank you, @duvld :+1: I wonder if @dorey could comment on how code like `addExternalItemAtPosition()`, which bridges the ES6 React stuff with CoffeeScript / Backbone code, ought to be tested.

PR #2830, opened 2020-10-09
Updating default search behaviour when no explicit term is given by user

### LINKED ISSUE #2814: Search term without field should query `asset.name`
The query parser expands `hello` to `summary:hello`, but we should have it expand to `summary:hello OR name:hello`. Both of these should use `icontains` when translated into `Q` objects.

https://github.com/kobotoolbox/kpi/blob/0deff9668027e73ecc7199132cee908ce25a41b8/kpi/utils/query_parser/query_parser.py#L110-L113

joshuaberetta:
Closes #2814 

The previous default behavior of the search was to compare the search term to values in 'summary__icontains' of the projects list. To refine the search, the user would have to enter search terms and values such as `name__icontains:foo AND summary__icontains:bar`. The default behavior is now more forgiving on the user and does the following for a default search:

user search: `foo`
resulting filter: `summary__icontains:foo OR name__icontains:foo OR uid__icontains:foo OR owner__username__icontains:foo OR tags__name__icontains:foo OR settings__description__icontains:foo`


kpi/utils/query_parser/query_parser.py
```diff
@@ -108,9 +125,23 @@ def notexp(text, a, b, elements):
...
+        if elements[0].text == '':
+            value = _get_value('', elements)
+            # A list of `Q` objects where every value is the same
+            # searched value
+            q_list = [
+                Q(**{field: value}) for field in SEARCH_DEFAULT_FIELD_LOOKUPS
+            ]
+            # combining all the `Q` objects with an `or` operator and
+            # returning the result
+            return reduce(operator.or_, q_list)
```
noliveleger, line 144: 👏

kpi/utils/query_parser/query_parser.py
```diff
@@ -108,9 +125,23 @@ def notexp(text, a, b, elements):
 
     @classmethod
     def term(cls, text, a, b, elements):
-        if elements[0].text == '':
+
+        def _get_value(field, elements):
             # A search term by itself without a specified field
-            field = SEARCH_DEFAULT_FIELD_LOOKUP
+            value = elements[1]
+            # the field value is not used in `process_value()`
```
noliveleger, line 132: Maybe it would be better to wrap field within `. 
> the `field` value ?

kpi/utils/query_parser/query_parser.py
```diff
@@ -108,9 +125,23 @@ def notexp(text, a, b, elements):
 
     @classmethod
     def term(cls, text, a, b, elements):
-        if elements[0].text == '':
+
+        def _get_value(field, elements):
             # A search term by itself without a specified field
-            field = SEARCH_DEFAULT_FIELD_LOOKUP
+            value = elements[1]
```
noliveleger, line 131: Personal preference. I would use a `_` suffix to `value` (and `elements`) to avoid any confusions with `value` (or `elements`) on line 136 & 153. (even it is in own scope and should not shadow `value` from outer scope)

noliveleger:
In term of performance I was afraid it would be slow but it does not seem to be. Except if we try with only 1 or 2 characters (e.g `q=e` or `q=er`). 
I'm wondering whether we ask for 3 characters at least if the field is not provided? what do you think (cc @jnm )

joshuaberetta:
> In term of performance I was afraid it would be slow but it does not seem to be. Except if we try with only 1 or 2 characters (e.g `q=e` or `q=er`).
> I'm wondering whether we ask for 3 characters at least if the field is not provided? what do you think (cc @jnm )

@noliveleger is there also quite a performance hit by filtering on every keystroke?

noliveleger:
> @noliveleger is there also quite a performance hit by filtering on every keystroke - or is that purely frontend?

@joshuaberetta I don't know how it's handled in the frontend, but if there is no delay between keystroke and the ajax calls (to prevent multiple calls at once) , I'm sure It will be an issue when the keyword is only less than 3 characters.

jnm:
we learned from Leszek on the standup that the front end does debounce :)
but i think it's fine to require that search terms contain at least 3 characters

jnm:
Let's go with this approach for now, given its simplicity:
> Totally reject search terms that do not specify fields if they are shorter than 3 characters

I'll open a new issue for the front-end improvements.

kpi/utils/query_parser/query_parser.py
```diff
@@ -114,6 +115,14 @@ def _get_value(_field, _elements):
...
         if elements[0].text == '':
             value = _get_value('', elements)
+
+            # As discussed here: https://github.com/kobotoolbox/kpi/pull/2830
+            # there strain on the server for small search queries without a
+            # specified field. The user will receive an empty list in response
+            # until using 3 or more characters
+            if len(value) < 3:
+                raise BadDefaultSearchException()
+
```
noliveleger, line 128: According to be mentioned before, we could specify the message over here instead.

```python
if len(value) < 3:
    default_detail = _('The search term must be 3 or more characters') 
```

Besides, I think we discussed about using a constant (instead of hard-coding `3`) at top the class. I may be wrong, though.

kpi/filters.py
```diff
@@ -206,7 +207,7 @@ def filter_queryset(self, request, queryset, view):
             q_obj = parse(
                 q, default_field_lookups=ASSET_SEARCH_DEFAULT_FIELD_LOOKUPS
             )
-        except ParseError:
+        except (ParseError, BadDefaultSearchException):
```
noliveleger, line 210: We may want to make to a difference between `ParseError` and `BadDefaultSearchException` to return the correct message to users. 
So I think, we should catch the `BadDefaultSearchException` within the ViewSet.

noliveleger (changes requested):

kpi/views/v2/asset.py
```diff
@@ -537,7 +537,12 @@ def koboform(self, request, *args, **kwargs):
         return Response({'asset': asset, }, template_name='koboform.html')
 
     def list(self, request, *args, **kwargs):
-        queryset = self.filter_queryset(self.get_queryset())
+        # raising an exception if the default search query without a 
+        # specified field is less than a set length of characters - currently 3
+        try:
+            queryset = self.filter_queryset(self.get_queryset())
+        except BadDefaultSearchException as e:
+            raise exceptions.ValidationError(e)
```
noliveleger, line 545: As per discussed, let's catch the exception in `SearchFilter`. 
See previous comment.

kpi/filters.py
```diff
@@ -202,12 +203,17 @@ def filter_queryset(self, request, queryset, view):
             return queryset
 
         try:
-            q_obj = parse(q)
+            q_obj = parse(
+                q, default_field_lookups=ASSET_SEARCH_DEFAULT_FIELD_LOOKUPS
+            )
         except ParseError:
             return queryset.model.objects.none()
```
noliveleger, line 210: As per discussed, I think it's better to handle the exceptions at the same place. 
Let's catch the `SearchQueryTooShortException` in the filter as well.

```python
except ParseError:
            return queryset.model.objects.none()
except SearchQueryTooShortException as e:
            raise e
```

noliveleger (changes requested):

kpi/exceptions.py
```diff
@@ -28,7 +29,7 @@ class KobocatProfileException(Exception):
 
 class BadDefaultSearchException(exceptions.APIException):
     status_code = 400
-    default_detail = 'The search term must be 3 or more characters'
+    default_detail = _('Your query is too short')
```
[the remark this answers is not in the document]
joshuaberetta: I like this approach as you suggested and it makes the code more readable 👍

kpi/tests/test_utils.py
```diff
@@ -210,3 +211,12 @@ def test_query_parser_no_specified_field(self):
...
         )
+
+    def test_query_parser_default_search_too_short(self):
+        # if the search query without a field is less than a specified
+        # length of characters (currently 3), then it should
+        # throw `SearchQueryTooShortException()` from `query_parser.py`
+        query_string = 'fo'
+        with self.assertRaises(SearchQueryTooShortException) as e:
+            parse(query_string, ASSET_SEARCH_DEFAULT_FIELD_LOOKUPS)
+        self.assertTrue('Your query is too short' in str(e.exception))
```
noliveleger, line 222: As per discussed in standup, let's start to use `assert condA == condB` 

Does `str(e)` return the same thing as `str(e.exception)`?

noliveleger (approved):

kpi/tests/test_utils.py
joshuaberetta: @noliveleger that's a good point - I'll check if `str(e)` does the same thing 👍

Issue #2833, opened 2020-10-09
Library & Public Collections UX Review


Ig-Rebollo:
There are several UX issues with the latest implementation of the Public Collections and Library.

Notes from the meeting with Tino, Anji and Leszek on Oct 9, 2020.

- In "My Library" view type column header should be filtering not ordering
- Collection meta on collection page should be all hidden, not half toggle like now - display some data with "Show more details" button. Everything could be just one line by default (unless expanded)
- Collection content is displaying only 10 items (bug ?). Ideally it should display the full list of assets contained in that collection.
- When typing a keyword on the search, can it also look for assets within a collection (in case assets in a collection can only be accessed by clicking in that collection). Searching for a question inside a collection - is it possible from My Library?
- Remove some columns from the collection view (Country, Org)
- Do we want to change all columns to ordering instead of filtering? (Ignacio votes for changing to just sorting temporarily -if we need to get this out soon- until we can do the filtering better. Better not to have a feature than to have it done wrong)
- Move the asset title from header to the content and the search box is always searching global
- **You can't add subscribed collection items in Form Builder (bug!)**
- **In Form Builder in "Add from Library" the no results message shouldn't be draggable (bug!)**
- Should it be possible to add an asset to multiple collections? At the moment, no

The main ideas for library view:
1. Leave current view and order collections first in the table
2. Items not in collection displayed in an "not in collection" custom collection
3. Split "My Library" into two separate items in sidebar: "My Collections" and "My Assets"
  - "My Collections" list the collections you have access too (owned, shared with and subscribed)
  - "My Assets" list the questions, blocks and templatest you have access too (regardless if they are inside collection or not)
4. Leave current view (collections + non-collected items) and add quick filters in sidebar under "My Library":
  - "just collections"
  - "items not in collection"
  - "shared with me"
  - "subscribed"

Ig-Rebollo:
@duvld @tinok @magicznyleszek @jnm 

Leaving here a proposal based on our discussion today. We will need to see:
1. If this is what we all envisioned
2. How much of this is possible/we want to prioritize in the time we have

Starting with the 'My Library', I created this option where you can choose between three categories: overview (shows collections and assets that are not part of any collection), My Collections (only shows collections you created or subscribed to) and My assets (displays all questions and blocks, both those that are part of a collection and not part of any collection)
[...]

Issue #2824, opened 2020-10-06
[Public Collections] Feature polishing feedback


magicznyleszek:
## Description

- [x] disallow making empty collections public (hint: add children check to `assetUtils.isAssetPublicReady` function)
- [x] reset filters button for table is not needed
- [x] move number of children (out from next to name) to a separate column
- [x] the number of children is sometimes wrong `The number in the grey circle (child assets) is wrong, at least sometimes: tinokreutzer's 'Sample Library Blocks' has 60 assets, but in the table's grey circle it says '1'. This is also in the JSON for the search result. Whereas the children count of the asset itself is reliably 60.`
- [x] don't display all columns for My Library `The columns in the Public Collections view are not needed in the main library view since questions and blocks can't have an Organization, Primary Sector, or Country attached to them. This means these rows are almost always empty; only public collections or templates would have this information, but it's not a good use of space to show this in the main library view`
- [x] similarly to above `Sector, Country, and Organization are not needed in the single question block view since there is no way to specify this information`
- [x] Search in library view is not applied to questions and blocks Search terms are only added to the Collections query, not the `asset_type:question OR asset_type:block`
- [x] "Back" button is misaligned on asset page (Chrome?)
- [x] searching for a question inside a collection - is it possible from My Library?
- [x] fix unnecessary search calls (jnm edit: detailed at #2808?)

- [ ] ~~fix filtering: `It works to filter for "Central African Republic" but when I filter by organization (UNDP CAR) or Primary Sector (Economic/Social development) I get 0 results. Expected: I should get the same result each time`~~ moved to https://github.com/kobotoolbox/kpi/issues/2835
- [x] ~~Cloning an asset from a subscribed Collection causes 400/500 errors~~ :bug: moved to https://github.com/kobotoolbox/kpi/issues/2836
- [ ] ~~Collection content is displaying only 10 items~~ :bug: moved to https://github.com/kobotoolbox/kpi/issues/2834
- [x] ~~You can't add subscribed collection items in Form Builder~~ :bug: moved to https://github.com/kobotoolbox/kpi/issues/2832
- [x] ~~Fix search~~ :bug: moved to https://github.com/kobotoolbox/kpi/issues/2814

jnm:
@magicznyleszek, this is ready to close, right?

PR #2838, opened 2020-10-14
Added parenthesis around library default query

duvld:
## Description

A tale of two parenthesis:

The default query to get library assets was: 
`asset_type:question OR asset_type:block OR asset_type:template OR asset_type:collection AND parent:null`
which is equivalent to:
`asset_type:question OR asset_type:block OR asset_type:template OR (asset_type:collection AND parent:null)`
which may be a mistake since `(asset_type:collection AND parent:null)` would be redundant as collections do not have parents.

PR adds parenthesis around the query to make it `(asset_type:question OR asset_type:block OR asset_type:template OR asset_type:collection) AND parent:null`, preventing assets in a collection from showing up in the default query. 

## Related issues

Semi-related to #2833 #2824

### DIFF
--- a/jsapp/js/dataInterface.es6
+++ b/jsapp/js/dataInterface.es6
@@ -429,14 +429,14 @@ export var dataInterface;
       return this._searchAssetsWithPredefinedQuery(
         params,
         // we only want orphans (assets not inside collection)
-        `${COMMON_QUERIES.get('qbtc')} AND parent:null`,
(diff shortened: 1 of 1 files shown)


duvld:
PR is duplicate of an already pushed fix from @magicznyleszek: https://github.com/kobotoolbox/kpi/commit/2cde8d30eb35ba3e656c2bdbb96ec1ee0f2217c4

PR #2840, opened 2020-10-16
Added failing test for getting subscribed collections

duvld:
## Description

Added a failing test that checks if we can grab subscribed collection assets with its `parent__uid`

## Related issues

Part of #2832

Issue #2832, opened 2020-10-09
 Public Collections: formbuilder library sidebar does not display any assets for subscribed collections


duvld:
## Tasks (jnm edit)

- [x] Add a failing unit test
- [x] Apply patch in [comment below](#issuecomment-708168429)
- [x] If results look weird, inspect the queries sent by the front end. For example, after applying this change, the library list will start showing the children of all subscribed collections because `asset_type:question OR asset_type:block OR asset_type:template OR asset_type:collection AND parent:null` is not the same as `(asset_type:question OR asset_type:block OR asset_type:template OR asset_type:collection) AND parent:null` ([see internal discussion](https://chat.kobotoolbox.org/#narrow/stream/4-KoBo-Dev/topic/Public.20Collection.20feedback/near/1989) about parentheses) -- fixed on https://github.com/kobotoolbox/kpi/commit/2cde8d30eb35ba3e656c2bdbb96ec1ee0f2217c4

Once a PR is prepared, let's remember to deploy it on `kf.beta` and check for performance problems.

## Description

Mentioned in #2824, a subscribed collection's assets do not load into "Add from Library" sidebar in the formbuilder.

## Steps to Reproduce
(In public collections https://kf.beta.kbtdev.org)
1. Subscribe to a Public Collection that contains assets
2. Open the formbuilder and open the "Add from Library" sidebar
3. Click on "Select Collection Name"
4. Select subscribed collection

## Expected behavior 

List of the assets from the subscribed collection appears below dropdown

## Actual behavior 

Returns "No search results found" and "0 assets found"

## Additional details

@jnm found the following wrt this issue:
> impossible to see subscribed collection contents in the form builder's library sidebar (because the `parent__uid` query is returning nothing; [read more here](https://chat.kobotoolbox.org/#narrow/stream/4-KoBo-Dev/topic/Public.20Collection.20optimizations/near/861))
>  - but, you may wonder, how does the collection detail view in the library show the children? answer: it gets the list of children directly from the asset detail endpoint

Contents of `read more here`:

>i think it is a Good Idea™ to include only subscribed assets, and not their children, in the asset list results. some of these collection assets can have thousands of children, and there's no way we want all of those being returned all the time
>i.e. i'm happy with the change of behavior from the code in `master`
>
[...]

PR #2841, opened 2020-10-16
Added id's of subscribed collection assets to permission filter

duvld:
## Checklist

1. [x] If you've added code that should be tested, add tests (test is on #2840)
~~2. [ ] If you've changed APIs, update (or create!) the documentation~~
3. [x] Ensure the tests pass
4. [ ] Make sure your code lints and you followed [our coding style](https://github.com/kobotoolbox/kpi/blob/master/CONTRIBUTING.md#use-a-consistent-coding-style) (please check my formatting!)

## Description

When querying for subscribed collections, right now `?=parent__uid:<some uid>` returns nothing. As a temporary fix we union a query for all subscribed collection assets with a query for their parents

## Related issues

Part of #2832

### DIFF
--- a/jsapp/js/searches.es6
+++ b/jsapp/js/searches.es6
@@ -170,6 +170,9 @@ function SearchContext(opts={}) {
         params.string = this.state.searchString;
       }
       if (this.state.parentUid) {
+        // TODO: currently this parent__uid query param returns nothing. See
+        // https://github.com/kobotoolbox/kpi/issues/2832 for the in place
+        // bypass solution
         params.parentUid = `parent__uid:${this.state.parentUid}`;
       }
       if (this.state.parentName) {
--- a/jsapp/js/searches.es6
+++ b/jsapp/js/searches.es6
@@ -170,9 +170,6 @@ function SearchContext(opts={}) {
         params.string = this.state.searchString;
       }
       if (this.state.parentUid) {
-        // TODO: currently this parent__uid query param returns nothing. See
-        // https://github.com/kobotoolbox/kpi/issues/2832 for the in place
-        // bypass solution
         params.parentUid = `parent__uid:${this.state.parentUid}`;
       }
       if (this.state.parentName) {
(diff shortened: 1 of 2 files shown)


jsapp/js/searches.es6
```diff
@@ -170,6 +170,9 @@ function SearchContext(opts={}) {
         params.string = this.state.searchString;
       }
       if (this.state.parentUid) {
+        // TODO: currently this parent__uid query param returns nothing. See
+        // https://github.com/kobotoolbox/kpi/issues/2832 for the in place
+        // bypass solution
```
jnm, line 175: can remove this comment

kpi/filters.py
```diff
@@ -77,8 +77,15 @@ def filter_queryset(self, request, queryset, view):
         # the query to be processed right now. Otherwise, because queryset is
         # a lazy query, Django creates (left) joins on tables when queryset is
         # interpreted and it is way slower than running this extra query.
-        asset_ids = list(owned_and_explicit_shared.union(subscribed)
-                         .values_list('id', flat=True))
+
+        # TODO: Figure out a better way to get subscribed collection's children
+        # As a temporary fix we union a separate query for id's of subscribed
+        # collection parents, bypassing the broken `?q=parent__uid` query
+        asset_ids = list(
```
jnm, line 84: we discussed that this fix is okay long-term, and that `?q=parent__uid` does work (because `SearchFilter` handles it), but a comment here about why it's necessary to check if the user has a subscription to the parent (and not just to the asset itself) would be helpful

kpi/filters.py
```diff
@@ -77,8 +77,15 @@ def filter_queryset(self, request, queryset, view):
...
+            (owned_and_explicit_shared.union(subscribed).union(
+                queryset.filter(parent__in=subscribed).values('pk')
+            )).values_list('id', flat=True)
+        )
         return queryset.filter(pk__in=asset_ids)
```
jnm, line 89: Oli's request to align the `union()`s for better readability

jnm (changes requested):
ready to merge after minor changes

kpi/filters.py
duvld: Changed the comment to:
```
# Since user would be subscribed to a collcetion and not
# the assets themselves, we append children of subscribed
# collections to the queryset in order for `?q=parent__uid`
# queries to return the collection's children
.union(queryset.filter(parent__in=subscribed).values("pk")
```

noliveleger: @duvld, typo :-) `collcetion`

PR #2844, opened 2020-10-21
updating semantics from member_since to date_joined

### LINKED ISSUE #2795: Change `member_since` to `date_joined` in users API
Quick change for simplicity to help future code readers
(change in public collections branches)

joshuaberetta:
Closes #2795 

Minor semantic change for readability and consistency.

### DIFF
--- a/kpi/serializers/v2/user.py
+++ b/kpi/serializers/v2/user.py
@@ -1,4 +1,6 @@
 # coding: utf-8
+import pytz
+
 from django.conf import settings
 from django.contrib.auth.models import User
 from django.contrib.contenttypes.models import ContentType
@@ -34,8 +36,9 @@ class UserSerializer(serializers.HyperlinkedModelSerializer):
                   'public_collections_count',
                   )
 
-    def get_date_joined(self, user):
-        return user.date_joined
+    def get_date_joined(self, obj):
+        return obj.date_joined.astimezone(pytz.UTC).strftime(
+            '%Y-%m-%dT%H:%M:%SZ')
 
     def get_public_collection_subscribers_count(self, user):
         public_collection_ids = self.__get_public_collection_ids(user.pk)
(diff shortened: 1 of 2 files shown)


kpi/serializers/v2/user.py
```diff
@@ -20,7 +20,7 @@ class UserSerializer(serializers.HyperlinkedModelSerializer):
     assets = PaginatedApiField(
         serializer_class=AssetUrlListSerializer
     )
-    member_since = serializers.SerializerMethodField()
+    date_joined = serializers.SerializerMethodField()
```
noliveleger, line 23: I don't think we need to redeclare `date_joined` because it's already part of the model. Just adding it to `fields` should be enough. 
We can override the output though. In that case, we need to use `serializers.SerializerMethodField()`

kpi/serializers/v2/user.py
```diff
@@ -29,12 +29,12 @@ class Meta:
...
                   'username',
                   'assets',
-                  'member_since',
+                  'date_joined',
                   'public_collection_subscribers_count',
                   'public_collections_count',
                   )
 
-    def get_member_since(self, user):
+    def get_date_joined(self, user):
```
noliveleger, line 37: This is already done magically by Django because `date_joined` is already a property of the model `User`. 

As I said above, we can keep it to override the output. In the `CurrentUserSerializer`, `date_joined` is overridden. We could do the same over here for consistency.
What do you think? 

https://github.com/kobotoolbox/kpi/blob/2795-change-member-since-to-date-joined/kpi/serializers/current_user.py#L54-L56

noliveleger (changes requested):

kpi/serializers/v2/user.py
joshuaberetta: @noliveleger I think consistency is nice across the different serializers, so perhaps we either change them all to allow Django to handle them or continue with the current "convention"?

noliveleger (approved):

PR #2848, opened 2020-10-27
Added cycling pagination to collection asset list

### LINKED ISSUE #2834: [Public Collections] Single collection page only shows first page (10) of children with no way of loading more
## Description

So the assumption was that `.children` will contain **all** the children. It was a bad assumption as `.children` have a standardized `count`, `next`, `previous` and `results` (the de facto list of children).

## Expected behavior 

Single collection page uses component from `assetsTable.es6` which has a built-in capability of displaing pagination and a change page callback.

There are two ways of handling this:

1. Easier - Use `AssetsTable`, 
~~provide proper pagination properties and on add a functionality to `assetRoute.es6` that uses the callback to load next or previous page of child assets.~~ 
`AssetsTable` already has pagination properties and expects its logic to be passed on instantiation, `collectionAssetTable` should be the one to pass the logic to it 
~~It would be nice to increase the default limit from `10` to a bigger number.~~ _(I've changed it already to `100`)_
2. ~~Harder - Drop pagination from `AssetsTable` and instead use an infinite scroll solution that will load next pages of results. This change will require to update all the places that uses `AssetsTable`.~~ It was agreed that we don't want infinite scroll :)

duvld:
## Description

Implemented cycling pagination to collection asset table wrapper. Displays up to 100 assets maximum and user can cycle through pages with "Next" and "Previous" buttons. 

## Related issues

Fixes #2834


magicznyleszek:
@duvld unfortunatel this will grow bigger 😢 As I discussed with John on Zulip, currently filtering/ordering in `CollectionAssetsTable` is done on client side using data received from backend - making it only work on a single page of results and thus being unusable for us after introducing pagination of children. 

We decided that it would be best to stop using the collection `.children` (as it is not possible to apply filters to the query) and just make separate query on collection page that will target only assets with parent being this collection.

We would need to drop/change most of the `collectionAssetsTable.es6`. I was thinking about using `myLibraryStore` when entering collection page and applying a search/filter to it - but it would be a problem, because it is possible to display not owned collection (e.g. public collection not subscribed into), so the store doesn't make sense.

magicznyleszek:
@duvld correct me if I'm wrong, but this can be closed in favor of #2882 ?

duvld:
https://github.com/kobotoolbox/kpi/pull/2882 replaces this PR

PR #2851, opened 2020-10-30
Bugfix cloning subscribed collection asset

### LINKED ISSUE #2836: [Public Collections] Cloning asset from a subscribed Public Collection is not possible
## Description

It should be possible to clone an asset from a subscribed Public Collection. This asset should be cloned without parent (normally when you own a collection and clone an asset, the clone has the same parent).

Currently we get `500` error.

clone-fail-pc-beta

magicznyleszek:
## Description

This is a more universal fix - whenever user is cloning an asset that has a parent, the code will make the clone keep the parent only if user has edit permissions on parent. This means that cloning asset from collection shared with me, will add the clone to that collection - and cloning asset from collection subscribed to, will don't add clone to that collection.

## Related issues

Fixes #2836


duvld (commented):
These changes don't clone the asset ~~for me 🙁~~ if I am *not* subscribed, but the option is still there. It no longer has a 400/500 error in the network, but now hangs at "Cloning..." and doesn't finish. There isn't any requests in the network tab either.

Should we allow for cloning without subscribing?

magicznyleszek:
@duvld pushed a fix for the thing you've found :)

duvld:
I can't actually see a non-subscribed asset in order to test the cloning due to #2951 :eyes:

magicznyleszek:
@duvld oh true. You can do this:
1. subscribe to collection
2. open the url to the child asset in new tab
3. unsubscribe from collection
4. refresh the child tab and test out :)

PR #2856, opened 2020-11-02
Export list not appearing fix [will use for v2 API]

### LINKED ISSUE #2855: Export list not appearing
## Description

(on current `master`, but not on production deployments)
Clicking EXPORT creates an export successfully (as seen from 201 response in the developer console), but nothing ever appears in the list of exports:
image

## Steps to Reproduce

1. Create an XLS or CSV export (non-legacy)
2. Observe no exports are listed
3. Go to `/exports/` endpoint
4. Notice that your new export appears
5. Go to `/exports/?q=source:[your asset uid]`
6. Notice that no results appear

## Expected behavior 

The export list endpoint should allow filtering by source asset UID.

## Actual behavior 

Filtering by source UID always returns nothing.

noliveleger:
I decided to use the new query parser.  I cherrypicked @joshuaberetta's work for public collections to use the same technic for exports tasks.

It means the query syntax has changed a little bit. 
Before, we had to use `q=source:<uid>` . 
Now, we have to use either `q=data__source:https://[kpi]/api/v2/assets/<uid>/` or `q=data__source__icontains:<uid>`.

The search may appear more verbose than before, but (IMHO) it's more consistent with what we have in `AssetViewSet`
Users can search with `q:<uid>` . The query will search within `uid`  and `data.source` fields. 

Closes #2855


kpi/constants.py
```diff
@@ -70,3 +70,19 @@
 
 # KC INTERNAL
 PERM_FROM_KC_ONLY = 'from_kc_only'
+
+# Terms that can be used to search and filter return values
+# from a query `q`
+ASSET_SEARCH_DEFAULT_FIELD_LOOKUPS = [
```
noliveleger, line 76: I would get rid of `ASSET_SEARCH_DEFAULT_FIELD_LOOKUPS` and `EXPORT_TASK_SEARCH_DEFAULT_FIELD_LOOKUPS` and I would assign their content to `self.default_field_lookups` in each viewset.  

I kept it like this because of the tests.

jnm:
Fixed in master with #2868. I'll convert this to a draft; we can use it when we create a v2 API for exports

joshuaberetta:
I'm going to close this PR as I've ported the changes to #3011 with the new `/api/v2/exports` endpoint.

noliveleger:
@joshuaberetta You can delete the branch if it's not needed anymore.

PR #2857, opened 2020-11-02
Removing nested children from asset detail endpoint

### LINKED ISSUE #2854: [Public collections] Remove nested children from asset detail endpoint
We added a nested array of `children` to the asset detail endpoint for the public collections overhaul, but we ended up not needing it:
https://chat.kobotoolbox.org/#narrow/stream/4-KoBo-Dev/topic/Public.20Collection.20feedback/near/3841

Instead, the front end will use the asset _list_ endpoint with a query (I believe something like `?q=parent__uid:a1234`).

Please remove the nested children from the detail endpoint. There's no use including them if they're not being used, and removing them should improve response time a little bit.

joshuaberetta:
closes #2854 

Changing the output of the `asset-detail` endpoint from a nested list of children to a count of children, i.e. from:

```
...
"children": [
    ...
]
...
```
to:
```
...
"children": {
    "count": <int>
}
...
```

The tests were also updated to account for this change and to make them pass again by making a minor change to the `KpiTestCase.assert_child_of` method.

_**Note that this change currently breaks the display of child assets when viewing a collection in the frontend.**_

### DIFF
--- a/kpi/serializers/v2/asset.py
+++ b/kpi/serializers/v2/asset.py
@@ -86,6 +86,9 @@ class AssetSerializer(serializers.HyperlinkedModelSerializer):
     # Only add link instead of hooks list to avoid multiple access to DB.
     hooks_link = serializers.SerializerMethodField()
 
+    # children = PaginatedApiField(
+    #     serializer_class="kpi.serializers.v2.asset.AssetListSerializer"
+    # )
 
     children = serializers.SerializerMethodField()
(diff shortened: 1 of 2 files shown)


noliveleger (approved):
I applied the same changes that John asked me to apply in one of my other PRs.

joshuaberetta:
> @joshuaberetta. FYI: I applied some changes to your PR that John asked me to apply in one of my PRs.

@noliveleger - perfect 👍

PR #2862, opened 2020-11-05
Removing exported file on delete

### LINKED ISSUE #2253: Export files are not deleted when trashing an export.
## Steps to Reproduce

1. Export data as XLS (or CSV) 
2. Click Trash icon

## Expected behavior 
File should be deleted

## Actual behavior 
File still remains

## Additional details
It appears when using S3 as default storage, but the issue is maybe present with FileSystem Disk

joshuaberetta:
closes #2253 

Previously, as described in the linked issue, when the trash icon in the UI was pressed to delete the export, the file remained in S3. This was also the case for locally stored exports. When the endpoint `exports/<uid>/` was called with the `DELETE` method, only the postgres record was removed but the artifact remained untouched. In the end a simple method was added to achieve this result: `ExportTaskViewSet.perform_destroy` as well as corresponding tests.

### DIFF
--- a/kpi/views/v1/export_task.py
+++ b/kpi/views/v1/export_task.py
@@ -90,12 +90,18 @@ class ExportTaskViewSet(NoUpdateModelViewSet):
         }, status.HTTP_201_CREATED)
 
     def perform_destroy(self, instance):
-        if 'KPI_DEFAULT_FILE_STORAGE' not in os.environ:
-            print('do something here to delete in S3')
+        if 'KPI_DEFAULT_FILE_STORAGE' in os.environ:
+            try:
+                ExportTask.result.field.storage.delete(
+                    name=str(instance.result)
+                )
+            except:
+                pass
         else:
             ROOT = os.environ.get('KPI_SRC_DIR')
             MEDIA = 'media'
-            file_to_delete = instance.result
-            os.remove(f'{ROOT}/{MEDIA}/{file_to_delete}')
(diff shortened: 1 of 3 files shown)


kpi/views/v1/export_task.py
```diff
@@ -86,3 +86,7 @@ def create(self, request, *args, **kwargs):
                 request=request),
             'status': ExportTask.PROCESSING
         }, status.HTTP_201_CREATED)
+
+    def perform_destroy(self, instance):
+        ExportTask.result.field.storage.delete(name=str(instance.result))
+        return super().perform_destroy(instance)
```
jnm, line 92: Hey @joshuaberetta,
* Would `instance.result.delete(save=False)` work? [[ref](https://docs.djangoproject.com/en/2.2/ref/models/fields/#django.db.models.fields.files.FieldFile.delete)]
* Could this go into a `delete()` method on the `ExportTask` model? [Some deletions may happen through the ORM directly](https://github.com/kobotoolbox/kpi/blob/bf0c8cb6d6804c2bd2d4e78d1f768cac2c121abb/kpi/models/import_export_task.py#L648-L651), not through the HTTP API, and it'd be nice to do the same clean up in those cases.

joshuaberetta: Thanks for the suggestion! I'll look into that 💪

jnm (approved):
Thanks :+1:

PR #2872, opened 2020-11-11
Public collections tweaks

### LINKED ISSUE #2735: New KoBo Toolbox Public Collections UX feedback
**This issue is to compile the feedback on the latest implementation of the Help site.**

So far, elements to address are:

1. The dot indicating you are subscribed is not very clear/visible. I would remove/replace the dot for making the folder icon with the globe and number all blue, and a checkmark immediately to the right of the icon.
image
2. The subscribe and unsubscribe buttons are both blue. I would change the color of 'unsubscribe' to either a grey or red color, to highlight the difference.
image
3. In a specific collection page:
- Back button to go back to the list view
- Indication of being subscribed plus the button to unsubscribe (in grey)
- Leave the 'show full details' out of the table itself (though this is not important and purely about the appearance)
image
4. Back into the collections list page, it feels strange to have pagination if there's also what seems like unlimited scroll. I would suggest either allowing scroll up to a certain point (i.e. 50 items) and then show a button saying 'load more' that displays another 50 and so on; or else not allowing much scrolling at all and implement pagination instead.
image

### LINKED ISSUE #2833: Library & Public Collections UX Review

magicznyleszek:
## Description

All the important functional fixes for Public Collections are included. And almost all of the visual changes from the designs are here.

## Related issues

Fixes #2833 
Fixes #2735


jsapp/js/components/library/assetsTableRow.es6
```diff
@@ -29,8 +24,7 @@ class AssetsTableRow extends React.Component {
     if (
       this.props.asset.asset_type !== ASSET_TYPES.collection.id &&
       this.props.asset.summary &&
-      this.props.asset.summary.row_count &&
-      this.props.asset.summary.row_count >= 2
+      this.props.asset.summary.row_count
```
dorey, line 27: 2 lines could be simplified to: `this.props.asset.summary?.row_count`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

jsapp/js/components/library/assetsTableRow.es6
```diff
@@ -29,8 +24,7 @@ class AssetsTableRow extends React.Component {
...
     ) {
```
jnm, line 28: i don't know what these other guys are talking about, but have you heard about this one weird trick called [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)?

jsapp/js/components/library/assetsTableRow.es6
magicznyleszek: It's first time I see this - thanks! Reminds me of coffee script :)

jsapp/js/components/library/assetsTableRow.es6
magicznyleszek: Alex was first, but thanks nonetheless :)

duvld (commented):
Walked through all the changes with John and he'll follow up with some notes of the quirks we found :smiley:

jsapp/js/components/library/assetsTableRow.es6
jnm: Thanks for indulging us our (bad) joke. I'd never heard of it before today ;)

jnm:
Hey @magicznyleszek, we tested this using https://github.com/kobotoolbox/kpi/tree/public-collections-tweaks-and-2834—to deal with the back end having removed the nested children from the collections endpoint—so take this with a grain of salt :)

1. When viewing someone else's public collection that I have not yet subscribed to, I cannot see the contents; "There are no assets to display" is shown. However, as soon as I click subscribe, I can see the contents as I should.
2. If I view an asset inside a collection (e.g. http://kf.kobo.local:9000/#/library/asset/arpMQwK9amnNNVf5M3Lcnc) directly by URL—or refresh after clicking through the usual way—the breadcrumbs show "Parent Collection" instead of the parent collection's name. That's not a huge deal, but if I then click on "Parent Collection", I see "There are no assets to display". Refreshing again shows me the correct content of the collection. This can be reproduced with private or public collections.
3. In the table that lists collection contents, the three dots menu gets trapped within a scrollable container on Firefox and Chromium-based Brave:
    image
3. We to put this on a back-end person's plate, but I'm writing it here to remember: it seems like `/api/v2/assets/metadata/` never returns anything when queried by the library. It looks like you've built the filter criteria dropdowns to use whatever's on the current page of asset results as a workaround(?)
4. Also back-end stuff: making a 600-asset collection public is slow. Adding a single language to one asset within a 600-asset collection is slow. I'm guessing we need to have reasonable performance for 2000+ assets within a collection :sweat_smile: 
5. Even more back-end stuff: assets with type `question` often have an empty name. The front end correctly compensates for this by using the label of the constituent question as the name; however, when trying to sort by name, this breaks down. The back end attempts to sort by comparing empty strings to each other and returns results that make no sense in the UI.
6. I know the mock ups sometimes use the term "asset" instead of "item", but I think we should avoid showing "asset" in the UI. I just now saw that Tino had mentioned this [before](https://github.com/kobotoolbox/kpi/issues/2833#issuecomment-715666033), so I will read through all of that discussion. I can make the verbiage changes myself after that :)
    * In collection view:
        * Assets count (needed along with items count in footer of table?)
        * "Items" column in table: should be "Questions"? (This appears to show the number of questions within a question/block/template)
7. The markedly tighter border radius on the new make public/make private/subscribe/unsubscribe button came up on the call. What I forgot at the time was that @Ig-Rebollo has been phasing in subtle design changes, so let's assume that this was intentional and not bother him—beyond this ping, sorry!—at this busy time.
    image

I really appreciate everything you've done so far. I know it's been a long haul, but it's looking very nice!

duvld:
>—so take this with a grain of salt :)

Indeed! The first two bugs (listed above and quoted below) are actually from my fix to the removed `.children` endpoint.

>1. When viewing someone else's public collection that I have not yet subscribed to, I cannot see the contents; "There are no assets to display" is shown. However, as soon as I click subscribe, I can see the contents as I should.
>2. If I view an asset inside a collection (e.g. http://kf.kobo.local:9000/#/library/asset/arpMQwK9amnNNVf5M3Lcnc) directly by URL—or refresh after clicking through the usual way—the breadcrumbs show "Parent Collection" instead of the parent collection's name. That's not a huge deal, but if I then click on "Parent Collection", I see "There are no assets to display". Refreshing again shows me the correct content of the collection. This can be reproduced with private or public collections.

The collection table makes this request: https://github.com/kobotoolbox/kpi/blob/accc31b02cfb7742735597407914fe559c8485e8/jsapp/js/dataInterface.es6#L427-L433

and is not returning anything for public collections. In contrast, the `Assets: #` counter in the `AssetInfoBox` does know the proper count because it makes this request:
https://github.com/kobotoolbox/kpi/blob/accc31b02cfb7742735597407914fe559c8485e8/jsapp/js/dataInterface.es6#L308-L315
instead. An important caveat: `.children` is not *removed*, it still contains `count` which is why this works at all.

magicznyleszek:
@jnm 
> When viewing someone else's public collection that I have not yet subscribed to, I cannot see the contents; "There are no assets to display" is shown. However, as soon as I click subscribe, I can see the contents as I should.

To get the collection's asset for the table FE makes this call:

```
http://kf.kobo.local/api/v2/assets/?q=(asset_type%3Aquestion%20OR%20asset_type%3Ablock%20OR%20asset_type%3Atemplate%20OR%20asset_type%3Acollection)%20AND%20parent__uid%3AaTFag99kq3Bt98Zq8n9JoD&limit=100&offset=0&ordering=-date_modified&metadata=on
```

So when I am not subscribed to public collection, this ^ returns empty. My guess is that I don't have permissions to access these assets? When I'm subscribed I get all the assets from it.

So my qustion is - should I have access to someone else's assets that are inside collection and this is some bug? Or it's ok but the message should be different or the table should be hidden until subscribed?

magicznyleszek:
@jnm 
> 2. If I view an asset inside a collection (e.g. http://kf.kobo.local:9000/#/library/asset/arpMQwK9amnNNVf5M3Lcnc) directly by URL—or refresh after clicking through the usual way—the breadcrumbs show "Parent Collection" instead of the parent collection's name. That's not a huge deal, but if I then click on "Parent Collection", I see "There are no assets to display". Refreshing again shows me the correct content of the collection. This can be reproduced with private or public collections.

So I fixed the "There are no assets to display" issue and now going to "Parent Collection" makes it load the content of collection.

As to the "not a huge deal", at the point when you open child asset URL directly, the only thing I know about the parent is the UID. I could be making a call to BE to get the parent to get the name in situation when it is not know, but it would mean that we either show "Parent Collection" that will change into collection name in a half second, or delay displaying breadcrumbs until I make the call. What do you think?

magicznyleszek:
@jnm 
> 3. In the table that lists collection contents, the three dots menu gets trapped within a scrollable container

Fixed :hammer:

magicznyleszek:
@jnm 
> We to put this on a back-end person's plate, but I'm writing it here to remember: it seems like /api/v2/assets/metadata/ never returns anything when queried by the library. It looks like you've built the filter criteria dropdowns to use whatever's on the current page of asset results as a workaround(?)

I actually use the `metadata` from the `/api/v2/assets/` with `metadata=on`. The calls to `/api/v2/assets/metadata/` are done to save calls whenever we handle the assets list change on FE but need the fresh metadata (e.g. a single asset is being updated, we get whole asset in response so we replace it in the list, but need metadata to be refreshed).

magicznyleszek:
@jnm 

> 7. I know the mock ups sometimes use the term "asset" instead of "item", but I think we should avoid showing "asset" in the UI. I just now saw that Tino had mentioned this before, so I will read through all of that discussion. I can make the verbiage changes myself after that :)
>   - In collection view:
>     - Assets count (needed along with items count in footer of table?)
>     - "Items" column in table: should be "Questions"? (This appears to show the number of questions within a question/block/template)

I've changed the label near the "Show full details" to "Children", but the "Item" in the table header is a different story. The column could be either "questions" or "questions/children" depending on the context. Should I change it to "Questions" when displaying a collection content and leave it as "Items" when displaying collections and other assets (and same in the table footer of course)? Or should it be left as "Items" in both tables?

magicznyleszek:
@jnm 
> 8. … (border radius in buttons)

I've looked at the designs again and changed the corners from `2px` to `3px` as it should be and it looks better now :)

jnm:
@magicznyleszek 

> @jnm
> 
> > We to put this on a back-end person's plate, but I'm writing it here to remember: it seems like /api/v2/assets/metadata/ never returns anything when queried by the library. It looks like you've built the filter criteria dropdowns to use whatever's on the current page of asset results as a workaround(?)
> 
> I actually use the `metadata` from the `/api/v2/assets/` with `metadata=on`. The calls to `/api/v2/assets/metadata/` are done to save calls whenever we handle the assets list change on FE but need the fresh metadata (e.g. a single asset is being updated, we get whole asset in response so we replace it in the list, but need metadata to be refreshed).

I don't know that `metadata=on` is doing what we need. Consider that I:
1. go to Public Collections, click dropdown, see English, Français, Malagasy, Urdu (ur)
1. pick Français
1. see now dropdown showing only English, Français, Malagasy
    * because Urdu (ur) never appears together with Français in a collection

I also don't see a way to go back to viewing all languages, i.e. there's no way to remove the language filter.

jnm:
@magicznyleszek, I found an issue with searching:
If I try to search for `owner__username:trevor`, it returns no results, because the query is quoted before being sent: I see `asset_type:collection AND "owner__username:trevor"` in the console, which the backend interprets as one big term called `owner__username:trevor`, not a value `trevor` for the field `owner__username`

jnm:
@magicznyleszek

> I've changed the label near the "Show full details" to "Children"

I think this should be "Items" to match the bottom, left-hand corner of the table. Also, if I started by browsing "Public Collections", I would've seen "Items" in the table header. Once I click into a collection, I might look for the same terminology.

> The "Item" in the table header is a different story. The column could be either "questions" or "questions/children" depending on the context.

OK, I see what you mean now. Could the header be sensitive to the route and show "Items" when the view is "My Library" or "Public Collections", but then change to "Questions" once we're looking at the contents of a collection?

magicznyleszek:
@jnm 
> I don't know that `metadata=on` is doing what we need. Consider that I:
> 
>     1. go to Public Collections, click dropdown, see English, Français, Malagasy, Urdu (ur)
> 
>     2. pick Français
> 
>     3. see now dropdown showing only English, Français, Malagasy
>        
>        * because Urdu (ur) never appears together with Français in a collection

It is intentional that the dropdowns only contain options that are meaningful for the filtered data. So what you describe is not a bug. The question is then is it wrong how it's done now?

> I also don't see a way to go back to viewing all languages, i.e. there's no way to remove the language filter.

Previously there was "Remove filters" button but it was removed. The way to remove filter now is to select it again from the dropdown. It was supposed to be replaced by this advanced searchbox with all filtering and things - so when you select something in filter it would appear in the searchbox as a clearable item.

I made a GH issue for the advanced searchbox to make talking about it easier: #2990 :)

jnm:
> The way to remove filter now is to select it again from the dropdown.

OK, that's good. Thanks. I don't know why I couldn't figure that out :shrug: but it seems like a common approach used by, for example, the GitHub issues list.

jnm:
> It is intentional that the dropdowns only contain options that are meaningful for the filtered data. So what you describe is not a bug. The question is then is it wrong how it's done now?

Yes, I would say that it's wrong. Let's say I have these library items:
* English 1 (languages: English)
* Français 1 (languages: Français)
* Français et Espagnol (languages: Français, Español)

When I go to my library list and click on the Languages drop down, I see:
* English
* Français
* Español

Great! That all makes sense. Let's say that I choose English.

After a while, I'm done working in English and want to switch to Français. I click on the drop down again, but the only option that appears is English. Technically, yes, this is the only language that appears in the current, filtered view of the data, but is it really the only "meaningful" choice? I have to know that I need to:
1. Click English again (to remove that filter)
2. Wait for the results to load
3. Click the drop down again
4. Pick Français

Then, a bit later, I'm done with Français and want to switch to Español. I click the drop down and see options for Français _and_ Español. Hooray, no extra steps?

Maybe the reasons for this are obvious enough when I'm dealing with my own library—and especially if it contains only two items—but if I'm browsing through a larger, unfamiliar list, I might not notice certain languages appearing and vanishing from the drop down, or understand why that would happen.

Are there any other applications you know of that work this way? GitHub's issue list doesn't bother excluding any choices: the filter drop downs always show everything, even choices that would yield zero results due to other filters. The shopping site newegg.com does change the available filter choices after I pick some criteria, but it doesn't take away options in the category that I just chose from:

No filter | After picking 4K Recommended Resolution
-- | --
image | image

jnm:
@magicznyleszek, I see there are a few new commits here that I guess aren't deployed to kf.beta.kbtdev.org yet—and maybe they fix this—but I want to document before I forget:
* Editing a library item by clicking the pencil icon in the list view works fine, but
* Clicking the pencil from the item landing screen navigates to the edit view but does not load the form builder. The form builder does appear after a refresh.
* 
https://user-images.githubusercontent.com/2085013/106780073-a267f380-6615-11eb-8aa3-fadf06442a3e.mp4

magicznyleszek:
@jnm You're right. Filters will be more useful if they will have all options available. I moved this discussion to separate issue: #2996

magicznyleszek:
@jnm I also moved the filters clearing discussion to #2996 ➡️

magicznyleszek:
> @magicznyleszek, I see there are a few new commits here that I guess aren't deployed to kf.beta.kbtdev.org yet—and maybe they fix this—but I want to document before I forget:
> 
>     * Editing a library item by clicking the pencil icon in the list view works fine, but
> 
>     * Clicking the pencil from the item landing screen navigates to the edit view but does not load the form builder. The form builder does appear after a refresh.

This was a tricky one, but I finally nailed it :hammer: :bug:. The problem I think was brought to light by the recent changes that added permission checks. It was here before, but luck made the race condition always win for us. Basically what happened is that coffee code needs `.form-wrap` node to exist (to render the survey editor into it), but it wasn't there and the code never tried again or was prepared for it in any way. It uses a `setTimeout` ugly code, as we need newer React version to do it nicely, but it will work 💯

magicznyleszek:
> @magicznyleszek
> 
> > I've changed the label near the "Show full details" to "Children"
> 
> I think this should be "Items" to match the bottom, left-hand corner of the table. Also, if I started by browsing "Public Collections", I would've seen "Items" in the table header. Once I click into a collection, I might look for the same terminology.
> 
> > The "Item" in the table header is a different story. The column could be either "questions" or "questions/children" depending on the context.
> 
> OK, I see what you mean now. Could the header be sensitive to the route and show "Items" when the view is "My Library" or "Public Collections", but then change to "Questions" once we're looking at the contents of a collection?

Fixed that too :+1:

PR #2877, opened 2020-11-13
Use `name` first to update data column

### LINKED ISSUE #2873: "Data column name" in form builder is not working (regression)
## Description

"Data column name" seems to neither read from nor write to the JSON content of the form.

## Steps to Reproduce

1. Create new form, open with form builder
1. Add question
1. Set "Data column name" to something different than the default for that question
1. Open network console
1. Save the form

## Expected behavior 

The manual question name is saved in the asset content

## Actual behavior 

The manual question name is not sent to the server

## Additional details

If you import an XLSForm and then open it in the form builder, the "Data column name" does not include values from the XLSForm `name` column. Instead, it appears to copy the `label`.

The bug does **not** exist in `2.020.30-hotfix-2450`, but it is present in `master` (currently 8572b97).

duvld:
Issue was caused by a conditional assignment switch made in #2723 and this PR switches it back. See https://chat.kobotoolbox.org/#narrow/stream/4-KoBo-Dev/topic/Form.20builder.20question.20name.20regression/near/5615

## Description

Fix a regression where "Data Column Name" is not saved by the form builder

## Related issues

Fixes #2873

### DIFF
--- a/jsapp/xlform/src/view.rowDetail.coffee
+++ b/jsapp/xlform/src/view.rowDetail.coffee
@@ -269,7 +269,10 @@ module.exports = do ->
         @model.set 'value', value
         @model.deduplicate @model.getSurvey()
       )
-      update_view = () => @$el.find('input').eq(0).val($modelUtils.sluggifyLabel @model._parent.getValue('label') || @model.get("value"))
+      # Update "Data Column Name" with `name` of question first, then use label.
+      # The order was swapped was added to fix kpi#706 but using this order does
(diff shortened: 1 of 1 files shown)


jnm (approved):
Confirmed that this reverts to https://github.com/kobotoolbox/kpi/blame/40704af765b73325f4adf9989db7e86c21b342eb/jsapp/xlform/src/view.rowDetail.coffee#L272

PR #2882, opened 2020-11-14
Copied library route and store functionality to single collection view

### LINKED ISSUE #2834: [Public Collections] Single collection page only shows first page (10) of children with no way of loading more

duvld:
#### Work in progress
Revamps the way we get and display a single collection's assets now that we've removed `.children` from the endpoint. To that end we are trying to mimic `myLibraryRoute` + `myLibraryStore` functionality into `collectionAssetTable` + (new) `singleCollectionStore`.

#### `myLibraryRoute` + `myLibraryStore`
Currently works like this, where the store holds actions that updates the route and in turn updates the table :+1: 
```
myLibraryRoute --(creates/updates)--> assetTable
            \
    (uses for actions)
                 \
      myLibraryStore
```

#### `collectionAssetTable`
Currently works like this, where `collectionAssetTable` holds functionality of a route and store which makes pagination and filtering awkward (see https://github.com/kobotoolbox/kpi/pull/2848#issuecomment-720577387) :-1: 
```
assetRoute --(creates)--> collectionsAssetTable --(creates/updates/"actions")--> assetsTable
```
We want to revamp the flow to like this:
#### `collectionAssetTable` + `singleCollectionStore`
```
assetRoute --(creates)--> collectionAssetTable --(creates/updates)--> assetTable
                                 \
                         (uses for actions)
                                      \
                           singleCollectionStore
```

## Current hurdles:

1. Table not populated when navigating from Library (fixed by https://github.com/kobotoolbox/kpi/pull/2882/commits/21071d1f41b24a44257d5baf15b5d63b0c42bbfe)

2. Filtering on pages that don't have the filtered assets returns nothing. (fixed by https://github.com/kobotoolbox/kpi/pull/2882/commits/173ccc53dfa62460a13c9e90834835462a345c07)
## Related issues

Fixes #2834


jsapp/js/components/library/singleCollectionStore.es6
```diff
@@ -0,0 +1,341 @@
...
+    actions.resources.createResource.completed.listen(this.onAssetCreated);
+    actions.resources.deleteAsset.completed.listen(this.onDeleteAssetCompleted);
+    // These are the changed requests, not all have been updated yet. Trying to
+    // just get assets to show up first
+    actions.library.searchMyCollectionAssets.started.listen(this.onSearchStarted);
+    actions.library.searchMyCollectionAssets.completed.listen(this.onSearchCompleted);
+    actions.library.searchMyCollectionAssets.failed.listen(this.onSearchFailed);
+    // TODO Improve reaction to uploads being finished after task is done:
+    // https://github.com/kobotoolbox/kpi/issues/476
+    actions.resources.createImport.completed.listen(this.fetchDataDebounced);
```
magicznyleszek, line 67: I'm not 100% sure, but I think single collection page doesn't need to react to imports, so we can delete this call (with comments) and delete `fetchDataDebounced` and delete `_` import.

jsapp/js/components/library/singleCollectionStore.es6
```diff
@@ -0,0 +1,341 @@
...
+
+  /**
+   * @return {object} search params shared by all searches
+   */
+  getSearchParams() {
+    const params = {
+      searchPhrase: searchBoxStore.getSearchPhrase(),
+      pageSize: this.PAGE_SIZE,
+      page: this.data.currentPage,
+      uid: getCurrentCollectionUID() // TODO a better way to get UID through listeners?
```
magicznyleszek, line 100: You can remove this TODO comment. Getting uid from url is pretty much ok. We have other mechanism in `mixins.es6` that gets the `uid` from the router parameters, but we should stop using `mixins` functionality in `React` as it's not supported and not recommended.

jsapp/js/components/library/singleCollectionStore.es6
```diff
@@ -0,0 +1,341 @@
...
+      params.filterValue = this.data.filterValue;
+    }
+
+    return params;
+  },
+
+  /**** Same code as `myLibraryStore` after this point ****/
+
+  fetchMetadata() {
+    actions.library.searchMyLibraryMetadata(this.getSearchParams());
```
magicznyleszek, line 115: This needs to be `searchMyCollectionMetadata` (which you need to set up in `actions/library` and `dataInterface`).

Fetching metadata instead of whole assets is useful for server when we know that a small portion of single asset changed.

magicznyleszek: You will also need to listen to `actions.library.searchMyCollectionMetadata.completed.listen(this.onSearchMetadataCompleted);`

magicznyleszek (changes requested):
Take a look at `assetRoute.es6` - there are two callback functions that uses asset's `.children` in their logic. As these are no longer existing on asset data, we need to do some cleanups here :)

jsapp/js/components/library/libraryUtils.es6
```diff
@@ -28,3 +28,27 @@ export function isOnPublicCollectionsRoute() {
     path.split('/')[2] === 'public-collections'
   );
 }
+
+/**
+ * Checks if on `/library/asset/…`
+ */
+export function isOnSingleCollectionRoute() {
```
magicznyleszek, line 35: Just a minor thing - this function name should be rather `isOnLibraryAssetRoute`, as it will work the same regardless if asset type is "collection", "question" or "template" :)

jsapp/js/components/library/libraryUtils.es6
```diff
@@ -28,3 +28,27 @@ export function isOnPublicCollectionsRoute() {
...
+  return (
+    path.split('/')[1] === 'library' &&
+    path.split('/')[2] === 'asset'
+  );
+}
+
+/**
+ * Returns current collection uid
+ */
+export function getCurrentCollectionUID() {
```
magicznyleszek, line 46: Similarly as above: `getCurrentLibraryAssetUID` :)

jsapp/js/components/library/singleCollectionStore.es6
```diff
@@ -0,0 +1,335 @@
...
+import {actions} from 'js/actions';
+import {
+  ORDER_DIRECTIONS,
+  ASSETS_TABLE_COLUMNS
+} from './assetsTable';
+
+// A store that listens for actions on assets from a single collection
+// Extends most functionality from myLibraryStore but overwrites some actions:
+// - searchMyLibraryAssets.* -> searchMyCollectionAssets.*
+// - searchMyCollectionMetadata.completed -> searchMyCollectionMetadata.completed
```
magicznyleszek, line 21: I think you meant:

```
// - searchMyLibraryMetadata.* -> searchMyCollectionMetadata.*
```

jsapp/js/components/library/singleCollectionStore.es6
```diff
@@ -0,0 +1,335 @@
...
+    if (this.abortFetchData) {
+      this.abortFetchData();
+    }
+
+    const params = this.getSearchParams();
+    // Surrounds `filterValue` with double quotes to avoid filters that have
+    // spaces which would split the query in two, thus breaking the filter
+    if (params.filterProperty !== undefined) {
+      params.filterValue = JSON.stringify(params.filterValue); // Adds quotes
+    }
```
magicznyleszek, line 125: Please add the same double quotes improvement to `publicCollectionsStore.es6` and `myLibraryStore.es6`

jsapp/js/dataInterface.es6
```diff
@@ -434,6 +441,13 @@ export var dataInterface;
 
       return this._searchAssetsWithPredefinedQuery(params, query);
     },
+    searchMyCollectionMetadata(params = {}) {
+      return this._searchMetadataWithPredefinedQuery(
+        params,
+        // we only want the currently viewed collection's assets
+        `${COMMON_QUERIES.get('qbtc')} AND parent:${params.uid}`,
```
magicznyleszek, line 448: Shouldn't this also (same as in `searchMyCollectionAssets`) be `parent__uid:` instead of `parent:`?

magicznyleszek (changes requested):
Generally these are all small changes, so you can freely just merge the PR after you make them :ship: :it:

PR #2891, opened 2020-11-20
Asset load bug fix

### LINKED ISSUE #2885: [Public Collections] UI is not refreshed when clicking on projects under `Deployed`, `Drafts` or `Archived`
## Description

When you click on the project link in the sub-menu, the project panel is not refreshed entirely: 
- It seems as though the `asset-detail` endpoint is not being called,
- Additionally, the project title and number of submissions disappears,
- It seems to happen when projects share the same condition of length of submissions (i.e. 0 or more than 0) 

## Steps to Reproduce

1. Create 4 projects and deploy them
    - Project A (2 submissions)
    - Project B  (1 submission)
    - Project C  (0 submissions)
    - Project D  (0 submissions)
2. Two projects with 2 or more submissions
3. Two projects with 0 submissions
4. Expand `deployed` menu. 
5. Click on Project A
6. Click on Project C -> ok
6. Click on Project D ->  bug occurs
6. Click on Project B -> ok
6. Click on Project A -> bug occurs

## Additional details
Also happens when cloning a project and trying to go into it after cloning for the sub-menu.

Present on `kf.beta`, not `kf.master`

magicznyleszek:
## Description

Context: opening asset from sidebar list can lead to either `/summary` (when asset has submissions) or `/landing` (no submissions or undeployed) route. When switching assets that lead to different route (e.g. from `<asset_1_uid>/summary` to `<asset_2_uid>/landing`) the asset loads properly. When switching asset to same route (e.g. from `<asset_1_uid>/summary` to `<asset_2_uid>/summary`) then we had a :bug:. 

The bug came to life when we removed a duplicated endpoin call from `header.es6`, which in some cases unknowingly was serving an accidental purpose to a different component. Instead of reverting the changes, I fixed the bug whilst keeping the calls still not duplicated.

## Related issues

Fixes #2885


jsapp/js/mixins.es6
```diff
@@ -213,6 +213,19 @@ mixins.dmix = {
...
     }
   },
+  // TODO
+  // Fix `componentWillUpdate` and `componentDidMount` asset loading flow.
+  // Ideally we should build a single overaching component that would
+  // handle loading of the asset in all necessary cases in a way that all
+  // interested parties could use without duplication or confusion and with
+  // indication when the loading starts and when ends.
+  componentWillUpdate(newProps) {
+    if (this.props.params.assetid !== newProps.params.assetid) {
```
duvld, line 223: `newProps` will contain just `newForm` as its context when clicking "NEW" (adding a new form), making `newProps.params` return `undefined` and crashing here

duvld (commented):
>Context: opening asset from sidebar list can lead to either /summary (when asset has submissions) or /landing (no submissions or undeployed) route. When switching assets that lead to different route (e.g. from <asset_1_uid>/summary to <asset_2_uid>/landing) the asset loads properly. When switching asset to same route (e.g. from <asset_1_uid>/summary to <asset_2_uid>/summary) then we had a bug.

While the PR does fix this scenario, there are a couple more quirks to how the sidebar affects what is shown. I've made a gif of them below, these involve the headings in each form landing. Not sure if we'd want to fix these on this PR or leave it for another issue



The above gets 'reset' when we switch from `SUMMARY` to `FORM` and vice versa:

jsapp/js/mixins.es6
magicznyleszek: The visual bug you've discovered was actually caused by this crash ^ :)

duvld:
@magicznyleszek there's another rendition of this :bug: that popped out:
- going from any form without data to another form without data creates a messed up state where the title to disappear and the underline for which form tab we are in (Form, Settings, etc.) is gone. For example, if we either:
  - go from Draft to Undeployed/deployed form with no data and vice versa
  - go from Undeployed/deployed form with no data to another Undeployed/deployed form with no data
  we get, in this 'messed up' state, clicking on another form tab goes to the specified tab of the previously open asset

magicznyleszek:
@duvld so by mistake I broke the hackfix 👎 it turned out to be quite easy fix for fix: 828ea5cbdd9ca659635ae66cc9c447bf28667388 (merged it directly to `beta`)

duvld:
Learning about `?.` really helps shorten these `if`'s :smile:

magicznyleszek:
> Learning about `?.` really helps shorten these `if`'s 😄

Yup, I wish I've learned about it earlier :D

PR #2901, opened 2020-11-26
Simplify partial usernames input

magicznyleszek:
## Description

Fix for self-harm calls causing 502s when checking too many `usernames` at once. It simplifies the partial users input. 

Split out from #2900 to be merged quicker.


magicznyleszek:
Closing because we want these changes on `master` - see #2902

PR #2902, opened 2020-11-27
simplify partial usernames input

magicznyleszek:
## Description

Fix for self-harm calls causing 502s when checking too many usernames at once. It simplifies the partial users input.

Split out from #2900 to be merged quicker.

### DIFF
--- a/jsapp/js/components/permissions/userAssetPermsEditor.es6
+++ b/jsapp/js/components/permissions/userAssetPermsEditor.es6
@@ -2,7 +2,6 @@ import React from 'react';
 import reactMixin from 'react-mixin';
 import autoBind from 'react-autobind';
 import Reflux from 'reflux';
-import TagsInput from 'react-tagsinput';
 import Checkbox from 'js/components/checkbox';
 import TextBox from 'js/components/textBox';
 import {stores} from 'js/stores';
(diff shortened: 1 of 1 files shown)


noliveleger:
There is a tiny issue: commas are duplicated in the payload.
For example: 
If I enter `foo, bar,  far, boo` in the `View submissions only from specific users` textbox, it sends 
```json
{"_submitted_by":{"$in":[
    "foo,",
    "bar,",
    "john,",
    "anji"
]
}}
```

noliveleger (changes requested):
~~There is a tiny issue: commas are duplicated in the payload~~

~~For example:~~
~~If I enter `foo, bar,  far, boo` in the `View submissions only from specific users` textbox, it sends~~

> ~~{"_submitted_by":{"$in":[~~
    ~~"foo,",~~
    ~~"bar,",~~
    ~~"john,",~~
    ~~"anji"~~
~~]~~
~~}}~~

noliveleger (approved):

jnm:
@noliveleger, I think your point was valid. Commas are not allowed in usernames:
https://github.com/kobotoolbox/kpi/blob/fc06513ba834b87fd2b0034607bd8307845482dc/kpi/forms.py#L9-L15

PR #2906, opened 2020-12-01
Require unnamed translation to be named before…

jnm:
allowing other translation-related actions. Inspired by a typo I found in #2825 :stuck_out_tongue:

## Description

If an unnamed translation exists, require it to be named before allowing other translation-related actions.

## Related issues

Part of #2800; related to #2825

### DIFF
--- a/jsapp/js/components/modalForms/translationSettings.es6
+++ b/jsapp/js/components/modalForms/translationSettings.es6
@@ -323,7 +323,7 @@ export class TranslationSettings extends React.Component {
                       </bem.FormView__label>
                     }
 
-                    {i !== 0 && !haveUnnamedTranslation &&
+                    {i !== 0 &&
                       <bem.FormView__iconButton
                         data-index={i}
                         onClick={this.changeDefaultLanguage}
@@ -364,7 +364,7 @@ export class TranslationSettings extends React.Component {
                       </bem.FormView__iconButton>
                     }
 
-                    {i !== 0 && !haveUnnamedTranslation &&
+                    {i !== 0 &&
                       <bem.FormView__iconButton
                         data-index={i}
                         onClick={this.deleteLanguage}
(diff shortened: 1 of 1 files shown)


jnm:
@duvld, I labeled this high priority because I'd like to get it out with a release today. Please let me know if you see any problems

duvld (changes requested):
Code looks good and works as expected :) My only concern is the following scenario:

There's no way to set any other language to the default if the user encounters this modal. If the user wants to set the null translation to an existing one (language that is added from a library question that is translated), the modal says "Name/Code must be unique!" and the user has to:
1. create a junk translation
4. set library question language to default
5. delete the junk translation.

This is a pretty specific scenario but it's happened to me a few times so maybe leaving the `set default language` and `delete translation` options in this modal is a viable option. I tested this out and the user still has to name the `unnamed translation` if they decide to use this option -- it just makes the flow described above more straightforward (no need to create junk transltion):

1. Make other language default
2. Name unnamed translation or delete it

jnm:
OK, I understand the scenario. From #2800:
> 3. :yellow_circle: if the library item does not have the language currently being used by the form builder, then the XML / "data column" name gets used as the label instead

In what you describe:
1. the "language currently being used by the form builder" is `null`
1. the library item does not have a `null` translation
1. the library item's XML / "data column" name gets used as the `null` label in the form
1. the library item's language is added to the form as a new, named translation

I'd be ideal if we asked the user which language they'd like to use instead of always taking the XML / "data column", but since we don't have time for that right now, I could take your suggestion:
> maybe leaving the set default language and delete translation options in this modal is a 

I'd intentionally disabled all actions except renaming because I was nervous about misunderstandings or nasty side effects when manipulating unnamed translations, but we're (I hope) sufficiently safeguarded by the form builder refusing to load in situations where nullifying the default translation would overwrite another (unnamed) translation. It halts with:
> Error loading survey:
>
> There is an unnamed translation in your form definition.
> Please give a name to all translations in your form.
> Use "Manage Translations" option from form landing page.

Small nitpick: remember that the "junk translation" is only junk in the most minimal scenario where you:
* don't have any existing form content
* don't add any questions other than the single library item
* don't change the library item's labels in any way after adding it (those changes would be saved in the `null` translation)

PR #2909, opened 2020-12-01
Allow groups to have labels when using `grid-theme`

duvld:
## Description

Fixes problem outlined here: https://github.com/kobotoolbox/kpi/pull/2708#issuecomment-736886276 which is a slight regression that always removes `grid-theme` group labels and isn't necessary. Made the guard stricter.
## Related issues
#2708

### DIFF
--- a/jsapp/js/utils.es6
+++ b/jsapp/js/utils.es6
@@ -115,8 +115,7 @@ export function unnullifyTranslations(surveyDataJSON, assetContent) {
             if (typeof surveyData.settings[0] !== 'undefined'
                 && typeof surveyData.settings[0].style === 'string'
                 && surveyData.settings[0].style.includes('theme-grid')
-                && surveyRow.type === 'begin_group'
-                && surveyRow[translatedProp] === null || surveyRow[translatedProp] === '') {
+                && surveyRow.type === 'begin_group') {
               delete surveyRow[translatedProp];
             }
             surveyRow[`${translatedProp}::${defaultLang}`] = surveyRow[translatedProp];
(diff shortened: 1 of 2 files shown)


jnm (approved):

jsapp/js/utils.es6
```diff
@@ -115,7 +115,8 @@ export function unnullifyTranslations(surveyDataJSON, assetContent) {
             if (typeof surveyData.settings[0] !== 'undefined'
                 && typeof surveyData.settings[0].style === 'string'
                 && surveyData.settings[0].style.includes('theme-grid')
-                && surveyRow.type === 'begin_group') {
+                && surveyRow.type === 'begin_group'
+                && surveyRow[translatedProp] === null || surveyRow[translatedProp] === '') {
```
jnm, line 119: The `&&` operator has higher precedence than `||`, so I don't think this does exactly what you want. If `surveyRow[translatedProp] === ''` is `true`, than the whole expression evaluates to `true`, regardless of everything that comes before the logical or operator. As a simplified example:
```
Welcome to Node.js v12.18.3.
Type ".help" for more information.
> false && false && false || true
true
```
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence#Table.

(All of this is a long-winded way to say "add a few parentheses" :wink:)

jnm (changes requested):

jsapp/js/utils.es6
duvld: Checked just in case my tests were only working because ` || surveyRow[translatedProp] === ''` was taking precedence, added parentheses and it still works :+1:

jnm:
Let me merge this now since it fixes a destructive problem. We can revert it later once kobotoolbox/formpack#227 is merged and KPI is updated to use it.

jnm (approved):

PR #2910, opened 2020-12-02
Improve row-level permission usernames input by switching to commas

magicznyleszek:
## Description

Use commas as the separator between usernames and disallow spaces


jnm (approved):
Tested locally; works well

Issue #2760, opened 2020-08-20
Data cleaning for multiple submissions


jnm:
User flow:
Ideally users should be able to use the table-search as usual, and then select the records they wish to update.
After selecting the records (rows), they would select the question which would lead them to an enketo form.
Here they would enter a new value for the selected question and confirm the update (a list/popup of affected IDs would appear with a prompt to confirm the update to the specific question... just as a safety measure).

See https://docs.google.com/spreadsheets/d/1jBiky5DJ0InsvCO5nPG11a3AtCMQsX_XmHKUnvUFGGc/edit#gid=1219299840

jnm:
* UX work
  * jnm put sshot here of current bulk actions & ask ignacio
  * then how do we choose question <-- actually the important ignacio question
* can FE give snippet of form content directly to Enketo (enketo-core?) and display input control for that question, including choice lists?
* can FE read value entered by user directly from Enketo (whether Enketo field is created directly by FE or a mini-form is created by BE and then loaded in Enketo)?

magicznyleszek:
From my research in regards to FE handling Enketo directly __it is possible__:
- FE will need both https://github.com/enketo/enketo-core and https://github.com/enketo/enketo-transformer
- FE will need to generate (or get from BE) a small XForm with the question<sup>1</sup>
- XForm will pass through `enketo-transformer` to get XML + HTML pair
- XML + HTML pair will be rendered for user with `enketo-core`<sup>2</sup>
- FE will hook up to `submit` button to get the new value (like here https://enketo.github.io/enketo-core/tutorial-00-getting-started.html) and pass it to BE

<sup>1</sup> I don't know how hard will it be to get a small XForm based on question type, especially with more complex types.

<sup>2</sup> It is possible that loading enketo form in our CSS/JS context will break something, so we might need to load it in an iframe and either communicate with iframe's parent to send new value or send it from within iframe.

Ig-Rebollo:
@jnm @magicznyleszek some initial ideas:

I would advocate for not selecting the column (or question) in the table. As that will complicate the navigation significantly there. We can definitely do it if you think it is important for the use cases, but it might mean giving the table an important update in order to make it intuitive. 

My suggestion would be (and I don't know how possible this is) to decide what to edit and what not IN the Enketo view. See mockups below:

First I select the entries I want to edit
[...]

PR #2920, opened 2020-12-10
[UNHCR] Backend support for bulk updating of submissions feature

### LINKED ISSUE #2929: Data cleaning / bulk updating back-end work
Part of #2760. Mostly complete in #2920.

Allow `/api/v2/assets/[asset uid]/data/bulk` to accept `PATCH` that contains:
* multiple unique submission identifiers
* a single, fully-qualified (including group hierarchy) question name
* a new response value

Then write the new value as the response to the question for all identified submissions. Old responses should be overwritten. Insert a new response into the submission if one didn't exist previously. Do this by manipulating the submission XML and re-`POST`ing it to KoBoCAT's `submission` endpoint. Move the `instanceID` into `deprecatedID` and generate a new UUID for `instanceID`, as [this is the way](https://docs.getodk.org/openrosa-metadata/) edits are indicated in OpenRosa.

The requesting user must have permission to change submissions for the project.

joshuaberetta:
A new endpoint has been added to support the bulk updating/editing of submissions that can be accessed though a `PATCH` request to `/api/v2/assets/{uid}/data/bulk/`. This feature will be available through the UI as well as through the API, provided the user has sufficient permissions to edit submissions for an asset.

Each `PATCH` request must contain a payload with an array of `submission_ids` and at least one question field (in its XML format) with an updated value that will be applied to all the submissions corresponding to the supplied `submission_ids`. Some XML fields are protected and those fields will be removed from the payload if present before performing any updates.

Example usage:

`curl -X PATCH https://[kpi]/api/v2/assets/{uid}/data/bulk/`

with a payload following this general format:

```
   {
      "submission_ids": [{integer}],
      "data": {
          <field_to_update_1>: <value_1>,
          <field_to_update_2>: <value_2>,
          <field_to_update_n>: <value_n>
      }
   }
```

In the event that the form has been updated with additional fields and submissions created before the update have been included in the array of `submission_ids`, those new fields will be added to the submissions. If the value to be updated is within a group or nested group, the following must be taken into account:

If the field is within a group called `group_1`, the field name is `question_1` and the new value is `new value`, the payload should contain an item with the following structure:

```json
"group_1/question_1": "new value"
```

Similarly, if there are `N` nested groups, the structure will be:

```json
"group_1/sub_group_1/.../sub_group_n/question_1": "new value"
```

## Description
A shiny new feature that allows for the bulk editing/updating of a project's submissions, to facilitate data cleaning.

## Related issues
- backend support for #2760 
- closes #2929


kpi/deployment_backends/kobocat_backend.py
```diff
@@ -724,6 +725,100 @@ def set_validation_statuses(self, data, user, method):
...
+                )
+                element_to_update_or_new.text = v
+
+            # TODO: might be worth refactoring this as it is also used when
+            # duplicating a submission
+            file_tuple = (_uuid, io.BytesIO(ET.tostring(xml_parsed)))
+            files = {'xml_submission_file': file_tuple}
+            # due to the implementation in kobocat, the `PATCH` is limited to
+            # accepting validation status updates only, so using `POST` is a bit
+            # of a work around
```
jnm, line 796: this is actually part of the OpenRosa spec: https://docs.getodk.org/openrosa-form-submission/
(`deprecatedID` comes from https://docs.getodk.org/openrosa-metadata/)

kpi/deployment_backends/kobocat_backend.py
```diff
@@ -724,6 +725,100 @@ def set_validation_statuses(self, data, user, method):
...
+                else ET.SubElement(xml_parsed.find('meta'), 'deprecatedID')
+            )
+            deprecated_id.text = instance_id.text
+            instance_id.text = uuid_formatted
+
+            # if the form has been updated with new fields and earlier
+            # submissions have been selected as part of the bulk update,
+            # a new element has to be created before a value can be set.
+            # However, with this new power, arbitrary fields can be added
+            # to the XML tree through the api.
```
jnm, line 781: right, it's powerful, but that's what we want

kpi/deployment_backends/kobocat_backend.py
```diff
@@ -863,3 +958,48 @@ def __prepare_as_drf_response_signature(requests_response):
...
+        Validating the request payload for bulk updating of submissions
+        """
+        if not 'submission_ids' in payload:
+            raise KobocatBulkUpdateSubmissionsException(
+                detail=_('`submission_ids` must be included in the payload')
+            )
+
+        if not isinstance(payload['submission_ids'], list):
+            raise KobocatBulkUpdateSubmissionsException(
+                detail=_('`submission_ids` must be a list')
```
jnm, line 973: i haven't looked to see if this is actually the case, but if this is a message intended for a client, and the client sent JSON, then we should call this an "array" to avoid confusion

kpi/exceptions.py
```diff
@@ -62,6 +62,12 @@ class KobocatDuplicateSubmissionException(exceptions.APIException):
     default_code = 'submission_duplication_error'
 
 
+class KobocatBulkUpdateSubmissionsException(exceptions.APIException):
+    status_code = 500
```
jnm, line 66: when `KobocatBulkUpdateSubmissionsException` is raised by `__validate_payload()`—if the messages used there are to be returned to the client—would the client receive a 500?

if a client sends incorrect data to the API, they should receive a 400. please reserve 500 errors for situations where it's us as the developers / sysadmins who are responsible for fixing the problem, i.e. a bug in the code or a server-side system not working properly

kpi/tests/api/v2/test_api_submissions.py
```diff
@@ -527,6 +530,67 @@ def test_duplicate_submission_by_anotheruser_shared_allowed(self):
...
+        super().setUp()
+        self.submission_url = reverse(
+            self._get_endpoint('submission-bulk-update-submissions'),
+            kwargs={
+                'parent_lookup_asset': self.asset.uid,
+            },
+        )
+        self.updated_submission_data = {
+            'submission_ids': ['1', '2'],
+            'q1': '🕺',
```
jnm, line 545: :smile:

joshuaberetta: I have a good laugh every now and then looking through your tests, so I thought I should carry on the tradition 🤡

kpi/exceptions.py
joshuaberetta: Thanks for pointing this out! I've created a new exception class to differentiate client and server-side errors 👍

kpi/deployment_backends/kobocat_backend.py
joshuaberetta: Good point 💯 it was intended for the client, so I've updated it with 'array'

kpi/deployment_backends/kobocat_backend.py
joshuaberetta: 🤘

magicznyleszek:
@joshuaberetta just a question - could this be used to update `_validation_status`?

jnm:
No, because that's not stored in the [submission XML](https://github.com/kobotoolbox/kobocat/blob/d91834ccb174da0fd0305f8e8ae11ed93cce10c0/onadata/apps/logger/models/instance.py#L123), but rather in a separate field on the KoBoCAT `Instance` model:
https://github.com/kobotoolbox/kobocat/blob/d91834ccb174da0fd0305f8e8ae11ed93cce10c0/onadata/apps/logger/models/instance.py#L152

kpi/deployment_backends/kobocat_backend.py
```diff
@@ -886,3 +1005,58 @@ def __prepare_as_drf_response_signature(requests_response):
...
+
+        # Ensuring submission ids are integer values and unique
+        validated_payload['submission_ids'] = list(
+            set(map(int, validated_payload['submission_ids']))
+        )
+
+        # Sanitizing the payload of potentially destructive keys
+        santized_payload = copy.copy(validated_payload)
+        for key in validated_payload:
+            if key == 'submission_ids':
```
jnm, line 1046: aha, so if there were a question named `submission_ids`, we wouldn't be able to bulk update its responses? could the request format be restructured to avoid the possibility of this collision?

kpi/views/v2/data.py
```diff
@@ -217,6 +217,43 @@ class DataViewSet(AssetNestedObjectViewsetMixin, NestedViewSetMixin,
...
+    "group_1/question_1": "new value"
+    </pre>
+
+    Similarly, if there are `N` nested groups, the structure will be:
+
+    <pre class="prettyprint">
+    "group_1/sub_group_1/.../sub_group_n/question_1": "new value"
+    </pre>
+
+
```
jnm, line 256: Thanks for including detailed documentation and a detailed description of the pull request. In the PR description, could you leave your complete text intact but move the `## Description` heading below it, and then write a short (few sentences tops) description beneath that heading that's suitable for inclusion in our release notes ([example](https://community.kobotoolbox.org/t/release-notes-version-2-020-25/9473))? Basically:
```
As much as you want to say, in as much detail as you'd like. This is a great PR
and here are several reasons why! Let me explain in useful detail how you might
take advantage of this PR as a user and understand how it works as a developer.
This text will not be included in the release notes, but the release notes will have
a hyperlink to this PR so that readers can easily access the full details.

## Description
Added great feature to do great stuff.

## Related issues
Closes #xyz, etc. This also appears in the release notes, in the rightmost column.
No changes are needed from what you already have.
```

jnm (changes requested):

kpi/deployment_backends/kobocat_backend.py
joshuaberetta: Because the payload is structured like this (which is the same as `/validation_statuses/`):

```
{
    "submission_ids": [],
    "foo": "bar",
    ...
}
```
I just make sure to pass over `submission_ids` before popping keys off. But you've got a good point about restructuring the payload so this isn't necessary. Something like this might be better?

```
{
    "submission_ids": [],
    "data": {
        "foo": "bar",
        ...
    }
}
```

kpi/deployment_backends/kobocat_backend.py
```diff
@@ -886,3 +1005,58 @@ def __prepare_as_drf_response_signature(requests_response):
...
+        """
+        Preparing the request payload for bulk updating of submissions
+        """
+        payload = json.loads(request_data['payload'][0])
+        validated_payload = cls.__validate_payload(payload)
+
+        # Ensuring submission ids are integer values and unique
+        validated_payload['submission_ids'] = list(
+            set(map(int, validated_payload['submission_ids']))
+        )
```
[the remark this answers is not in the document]
joshuaberetta: Good catch! I'll do that 💪

kpi/deployment_backends/kobocat_backend.py
```diff
@@ -761,6 +770,116 @@ def set_validation_statuses(self, data, user, method):
...
+                        else:
+                            updated_xml_path = '/'.join(accumulated_elements)
+                            ET.SubElement(
+                                xml_parsed.find(updated_xml_path), element
+                            )
+                            accumulated_elements.append(element)
+
+                element_to_update_or_new = (
+                    xml_parsed.find(k)
+                    if xml_parsed.find(k) is not None
```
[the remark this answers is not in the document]
joshuaberetta: I'll look into that, but that's a good change to make in any case 👍

kpi/views/v2/data.py
joshuaberetta: I'll do that 🕺 I read [this article](https://mtlynch.io/code-review-love/) while working on this feature and I was inspired to make this documentation a bit more extensive than I have in the past — I maybe got a bit carried away 🙃

kpi/deployment_backends/kobocat_backend.py
```diff
@@ -761,6 +770,116 @@ def set_validation_statuses(self, data, user, method):
...
+
+        # TODO: This follows a similar reponse message that bulk updating of
+        # `validataion_statuses` will give, however there may be something more
+        # useful to send back to the client
+        if successful_updates > 0:
+            return {
+                'status': status.HTTP_200_OK,
+                'data': {
+                    'detail': _(
+                        '{} submissions of {} attempts have been updated'
```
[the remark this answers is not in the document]
joshuaberetta: It's passed to a `Response` object in the data viewset and expanded following the same pattern I saw implemented there: `return Response(**json_response)` — where it must be JSONified. The other methods in `KobocatDeploymentBackend` typically do something like `return self.__prepare_as_drf_response_signature(...)` and `__prepare_as_drf_response_signature` returns a dict to the viewset hence following that pattern. I'm happy to change that though if we're wanting to move away from that.

kpi/exceptions.py
```diff
@@ -62,6 +62,19 @@ class KobocatDuplicateSubmissionException(exceptions.APIException):
...
+    default_detail = _('An error occurred trying to bulk update the submissions.')
+    default_code = 'bulk_update_submissions_error'
+
+
+class KobocatBulkUpdateSubmissionsClientException(exceptions.APIException):
```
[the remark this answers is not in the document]
joshuaberetta: There is no valid reason 💩 I'll change that!

kpi/views/v2/data.py
jnm: lol: "But I don’t want my reviewer to fall in love with me"
i'll have to read the article :) the documentation you wrote is great, and i do indeed love it. it just won't fit in our silly markdown table that we post to Discourse. i'm very glad that it will remain here in GitHub for future reference, though.
not to dampen the spirits of your new description for the release notes (with "shiny new feature," a phrase that surely makes Tino proud!), but i may edit it to clarify that this is API only—just in case "Backend support" in the title doesn't make it sufficiently clear for non-developer folks.

kpi/deployment_backends/kobocat_backend.py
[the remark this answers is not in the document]
jnm: sorry for being unclear: i was specifically referring to the string `{} submissions of {} attempts have been updated`. if we expect a machine to glean something useful from that, we're not being too nice to our fellow developers by asking them to do string parsing to get the two integers out in order to see whether the operation completed (totally) successfully. 

i did miss earlier the comment that says this was inspired by the way bulk updating validation statuses works (hey, btw, there is a `validataion` typo in the comment :wink:). is [this](https://github.com/kobotoolbox/kobocat/blob/d91834ccb174da0fd0305f8e8ae11ed93cce10c0/onadata/apps/api/viewsets/data_viewset.py#L443-L446) the response you saw? if so, i can see that i'm not evenly applying my string-parsing criticism :)

i would draw one point of distinction here, though: it doesn't appear that bulk updates of validation statuses can fail partially. the `{} submissions have been updated` message is an (admittedly not machine-friendly) informational accompaniment to a HTTP 200 response that always means total success.

i did glance at the ~~HTTP spec~~, er, [someone's interpretation of the HTTP spec](https://stackoverflow.com/questions/8472935/http-status-code-for-a-partial-successful-request#comment114923447_22911940) when i did the previous review and concluded that 200 response was just fine for partial success. trying to think, though, about what the front end would need to do when it receives a response made me wonder if something like `{successes: 7, failures: 3}` might be more helpful than `7 submissions of 10 attempts have been updated`. i did a little searching and can't find a good example of best practices for this. since we're setting a precedent (for ourselves) here, it might be worth a quick chat in Zulip.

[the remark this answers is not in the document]
joshuaberetta: Thanks for the detailed explanation (and catching that sneaky typo) — that makes a lot of sense 💪 The changes have been made according to our discussion on Zulip.

kpi/deployment_backends/kobocat_backend.py
```diff
@@ -1070,3 +1066,48 @@ def __validate_bulk_update_submissions(submissions: list) -> list:
...
+        '''
+
+        # Unfortunately, the response message from OpenRosa is in XML format,
+        # so it needs to be parsed before extracting the text
+        results = [
+            {
+                'uuid': res['uuid'],
+                'status_code': res['response'].status_code,
+                'message': ET.fromstring(res['response']._content)
+                .find('{http://openrosa.org/http/response}message')
```
jnm, line 1092: what if KoBoCAT really misbehaves and doesn't return XML?
```
In [1]: from xml.etree import ElementTree as ET                                                                                      

In [2]: ET.fromstring('502 Bad Gateway').find('{http://openrosa.org/http/response}message')                                          
Traceback (most recent call last):

  File "/usr/lib/python3/dist-packages/IPython/core/interactiveshell.py", line 3331, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)

  File "<ipython-input-2-ce5bd0eb196f>", line 1, in <module>
    ET.fromstring('502 Bad Gateway').find('{http://openrosa.org/http/response}message')

  File "/usr/lib/python3.8/xml/etree/ElementTree.py", line 1320, in XML
    parser.feed(text)

  File "<string>", line unknown
ParseError: syntax error: line 1, column 0
```

kpi/deployment_backends/kobocat_backend.py
```diff
@@ -1070,3 +1066,48 @@ def __validate_bulk_update_submissions(submissions: list) -> list:
...
+            the client
+        '''
+
+        # Unfortunately, the response message from OpenRosa is in XML format,
+        # so it needs to be parsed before extracting the text
+        results = [
+            {
+                'uuid': res['uuid'],
+                'status_code': res['response'].status_code,
+                'message': ET.fromstring(res['response']._content)
```
jnm, line 1091: why `_content` instead of `content`? the leading underscore is a Python convention—not enforced in any way—that indicates a private variable. since `_content` is coming from the Requests library and not our own code, we should probably keep our hands off unless we have good reason to touch it :)

jnm (changes requested):

kpi/deployment_backends/kobocat_backend.py
joshuaberetta: Good point 👍 sorted.

kpi/deployment_backends/kobocat_backend.py
joshuaberetta: I wish I had a good reason for that — thanks for spotting. Sorted 👌

jnm (approved):

PR #2599, opened 2020-03-18
Fix wrong columns being displayed for grouped questions

magicznyleszek:
## Description

I removed columns groups and duplicated grouped questions from table view using a recently added utility function.

## Related issues

Fixes #2249

PR #2939, opened 2020-12-23
2937 table columns not displaying

### LINKED ISSUE #2937: _id column sometimes doesn't show up in table view
## Description

> Users are reporting that _id is not showing on the "Table" view of the interface since last deployment (see attachments). _id appears in the data export, just not on the table view. We checked our accounts and the same is happening - not to all forms, but to some of them

## Steps to Reproduce

1. Create a form with a `note` question type
2. Add some submissions
3. Obserwe all columns that are not other questions (or meta data selected in form settings, e.g. `start`, `simserial`) are not being displayed

## Expected behavior 

`_id` column should always appear and should be available for freezing

## Additional info

This was introduced in #2599

magicznyleszek:
## Description

This bug was caused by `note` type exclusion code (#2599) gone wrong. I've also added all missing meta data in the single submission modal.

## Related issues

Fixes #2937

### DIFF
--- a/jsapp/js/components/submissionDataTable.es6
+++ b/jsapp/js/components/submissionDataTable.es6
@@ -11,6 +11,7 @@ import {
   getSubmissionDisplayData
 } from 'js/submissionUtils';
 import {
+  META_QUESTION_TYPES,
   QUESTION_TYPES,
   SCORE_ROW_TYPE,
   RANK_LEVEL_TYPE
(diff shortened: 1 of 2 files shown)


jsapp/js/components/submissionDataTable.es6
```diff
@@ -301,9 +302,17 @@ class SubmissionDataTable extends React.Component {
...
+        {this.renderMetaResponse(META_QUESTION_TYPES.get('username'), t('username'))}
+        {this.renderMetaResponse(META_QUESTION_TYPES.get('simserial'), t('sim serial'))}
+        {this.renderMetaResponse(META_QUESTION_TYPES.get('subscriberid'), t('subscriber ID'))}
+        {this.renderMetaResponse(META_QUESTION_TYPES.get('deviceid'), t('device ID'))}
+        {this.renderMetaResponse(META_QUESTION_TYPES.get('phonenumber'), t('phone number'))}
+        {this.renderMetaResponse(META_QUESTION_TYPES.get('audit'), t('audit'))}
         {this.renderMetaResponse('__version__', t('__version__'))}
+        {this.renderMetaResponse('_id', t('_id'))}
         {this.renderMetaResponse('meta/instanceID', t('instanceID'))}
         {this.renderMetaResponse('_submitted_by', t('Submitted by'))}
```
jnm, line 317: I lean ever so slightly against changing these labels in case we've (unwittingly?) trained people to expect that they appear in a specific way. These also introduce a discrepancy between the table and the modal, but it's probably a step in the right direction :)

jnm (approved):

PR #2941, opened 2020-12-30
Bulk edit submissions User Interface

magicznyleszek:
## Description

Added UI that allows users to use the new bulk edit submissions endpoint. It can be accessed through Table View after checkbox-selecting multiple submissions/rows.

## Additional details

This PR includes the code from #2920 

~~Things needed:~~
- ~~the support url for bulk editing (link is in the UI, currently it's a `const HELP_ARTICLE_URL = 'https://foo.bar';` ;-)~~

## Related issues

Part of #2760
Related to #2920


magicznyleszek:
Note to self - @joshuaberetta review:

some observations from playing around with the frontend:

- [ ] 1. I noticed that "... of `x results`"  is showing as a link which then selects all the submissions (same behaviour as checking the top-left box in the table, but without being able to uncheck it) — was this intentional?
```html
<a class="bulk-select-link">2 results</a>
```
[results link](/user_uploads/2/88/XALHgmSQZfJtSzwYDjpyncIX/Screen-Shot-2020-12-31-at-3.19.21-PM.png)

- [ ] 2. Things get a bit wonky at some screen sizes
[screen size](https://chat.kobotoolbox.org/user_uploads/2/43/LsCu4HQ-C-sgS6EMzu0wLoEW/Screen-Shot-2020-12-31-at-3.27.31-PM.png)

- [ ] 3. Typo in edit modal:
> "You are currently seeing multiple submissions at once. You can select specific questions to edit or remove reponses in bulk. ..."

"reponses" -> "responses"

- [ ] 4.  I can still click "CONFIRM AND CLOSE" even if no changes are made, resulting in a 400 error and a response of "Payload must contain data to update the submissions" from the backend (the "DISCARD CHANGES" button is disabled until a change is made)
[error](https://chat.kobotoolbox.org/user_uploads/2/94/GKJIRDSw5Gz9SwWTnSyty74C/Screen-Shot-2020-12-31-at-3.49.08-PM.png)

- [ ] 5. If the response is within a group, the frontend doesn't send the correct payload to the backend ie. If the new value is nested in group `foo` and question `bar`, the backend needs to receive something like:
```json
{
	"submission_ids": ["1", "2"],
	"data": {
		"foo/bar": "new value"
	}
}
```
Currently you're sending something like this:
```json
{
	"submission_ids": ["1", "2"],
	"data": {
		"bar": "new value"
	}
}
```
Resulting in some issues. There's a more detail in the [description of PR](https://github.com/kobotoolbox/kpi/pull/2920#issue-535569213) about this.

- [ ] 6. Maybe a question for @**John Milner** : A form with some questions has been deployed and then updated so that some of those original questions are now within a group, i.e. question `bar` is now `foo/bar`, the frontend doesn't allow for changing previous values under question `bar`, therefore ending up with both `bar` and `foo/bar`. How do you suggest handling this?

- [ ] 7. Strangely, sometimes when I sort the table by clicking on the header, the columns rearrange?
[columns.gif](https://chat.kobotoolbox.org/user_uploads/2/e6/fvOTUIixXhaBloLNmtgAXktf/columns.gif)

magicznyleszek:
@jnm I fixed all the FE things Joshua reported, do you want to take a look at it now? :)

jnm:
> 6. Maybe a question for @**John Milner** : A form with some questions has been deployed and then updated so that some of those original questions are now within a group, i.e. question `bar` is now `foo/bar`, the frontend doesn't allow for changing previous values under question `bar`, therefore ending up with both `bar` and `foo/bar`. How do you suggest handling this?

This is something we should improve in the future, but not a departure from the way the front end currently works. When editing a single submission, for example, the responses are loaded into Enketo using the most recently deployed version of the form. Also, the table view and single submission modal do not consult previous versions of the form when determining question types or labels—but the front end does gracefully fall back to using the XML names when it encounters a question that's not present in the current version of the form. Eventually, we should solve this by using the logic in [formpack](https://github.com/kobotoolbox/formpack/) (not by re-writing that logic in the front end).

jnm:
@magicznyleszek, in the interest of getting this merged before an eternity elapses, I'm going to ask @duvld to review it. Please let me know if there's something in here that you'd like me in particular to have a look at. Thanks :)

magicznyleszek:
@jnm I will let you know if something pops up, but I think @duvld will manage :)

duvld (approved):
Posting PM here: I found a small thing: When you open the bulk submission modal for the first time, clicking "SELECT" on one submission closes the modal and refreshes the page instead of populating the response box. This can be React trying to reload the UI after the state change, but it doesn't explain why it would only do it once.

Besides this (which isn't crucial) I'd say the front-end is good to go :)

magicznyleszek:
@duvld I think I managed to fix the issue. The cause was a missing `evt.preventDefault()` in the submit callback. Sometimes weird things happen when you have `<form>` element with submit and buttons inside. Similar issue was in the New Library Item Modal.

I will merge this :)

PR #2949, opened 2021-01-08
[Public Collections] Fix app crashing when language is not English

magicznyleszek:
## Description

Frontend code was wrongly using translated labels as id's to display deployment groups ("Drafts", "Deployed" and "Archived") which was causing the crash when the language was other than English (e.g. endpoint expected `Drafts` but was getting `Brouillons` when set to French).

It also includes a fix for when a graph style override was saved with the UI language other than English. There is no way to set invalid value through UI now and we will ignore existing invalid values and use default style instead.

## Additional details

Also includes a fix for #2950


jnm (approved):

jnm:
Tested locally; works well :+1:

Kalyan-Lama:
For documenting: https://community.kobotoolbox.org/t/problem-accessing-reports-with-graphs-and-charts/14992

PR #2955, opened 2021-01-12
Fix to allow for child assets to be visible in public collections when unsubscribed or non-owner

### LINKED ISSUE #2951: Child assets not visible in list view of unsubscribed public collections
## Description

As described in #2872 in this [comment](https://github.com/kobotoolbox/kpi/pull/2872#issuecomment-751987137):
> When viewing someone else's public collection that I have not yet subscribed to, I cannot see the contents; "There are no assets to display" is shown. However, as soon as I click subscribe, I can see the contents as I should.

## Steps to Reproduce

1. Visit a public collection that has at least one child asset that you aren't subscribed to
2. The list of child assets is empty
3. Visit the url of the child asset(s) directly and they are visible

## Expected behavior 

Visiting an unsubscribed public collection with child assets should list them under the following circumstances (from Zulip conversation [here](https://chat.kobotoolbox.org/#narrow/stream/4-KoBo-Dev/topic/Public.20collections.20tweaks.20backend/near/12011)):

- I have a collection with 5 child assets
- I make the collection public (give anonymous the view permission)
- This permission is inherited by all 5 children
- Anyone should be able to see this collection and its 5 children
- Now, I explicitly remove the anonymous view permission from one of the children (set a "deny" permission on this child)
- Change: anyone should be able to see this collection and 4 out of 5 of its children

## Actual behavior 

List view of child assets is empty.

joshuaberetta:
A new `_get_queryset_for_discoverable_child_assets` method was added to the `KpiObjectPermissionsFilter` to provide this fix. The method pulls the `parent__uid` paramenter from the search query and checks if that parent has a status of `public-discoverable`.

## Description

Allow the contents (child questions, blocks, and templates) of any discoverable, public collection to be visible in the list view to non-owners and those who have not subscribed to the collection.

## Related issues

Closes #2951


jnm:
Minor thing: please write commit messages and descriptions in the imperative mood. "Narrow down" is perfect; "Using" I'd replace with "Use" and "Allowing" I'd replace with "Allow". This is just a request for future stuff; I'm not requesting you do an interactive rebase or something to change existing commits in this PR unless you'd really like to practice that.

joshuaberetta:
@jnm, thanks for the advice 🕺

kpi/filters.py
```diff
@@ -175,6 +186,47 @@ def _get_queryset_for_collection_statuses(self, request, queryset):
...
+
+        self._return_queryset = False
+        PARENT_UID_PARAMETER = 'parent__uid'
+
+        if 'q' not in request.query_params:
+            return queryset
+
+        request_query = request.query_params['q']
+        parent_uid = re.search(
+            f'{PARENT_UID_PARAMETER}:([a-zA-Z0-9]*)', request_query
```
jnm, line 213: Let's not use this as a long-term solution, but for now, I'm going to merge it into `public-collections-tweaks` so that we can do more front-end testing there

kpi/tests/api/v2/test_api_collections.py
```diff
@@ -195,6 +195,16 @@ def test_collection_filtered_list(self):
             response.data['results'][0]['uid'], public_collection.uid
         )
 
+        # Logged in as another user, retrieve all children of
+        # `public_collection`. Should have 1
+        query_string = f'parent__uid:{public_collection.uid}'
+        url = f'{list_url}?q={query_string}'
+        response = self.client.get(url)
+        self.assertTrue(response.data.get('count') == 1)
```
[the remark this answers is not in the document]
joshuaberetta: I agree 👍 I prefer this `assert response.data.get('count') == 1`, but stuck to the style of the immediate neighbours. I'll stick to using `assert` going forward though 👌

kpi/filters.py
joshuaberetta: It did feel very hacky to do it that way... definitely all for a better solution.

Issue #2517, opened 2019-11-25
Include __version__ field in xls/csv exports


tinok:
In general: are there any columns we legitimately want to exclude from exports or the table view? I believe that the table view already has a UI to hide columns if desired. —jnm

## Description

As a user I want to see the version ID for each submission in my XLS and CSV tables so I can attribute the data to different deployed form versions.

## Expected behavior 

__version__ column is included at the beginning of the metadata section (just like in kobocat exports).

## Additional details

Optional, but we could also export the _notes and _tags columns available in KC but not in KPI since users may have stored something there through the KC API.

jnm:
Also should include `_submitted_by`

jnm:
and `_status` :sunglasses:

Kalyan-Lama:
Community link: https://community.kobotoolbox.org/t/submitted-by-field-not-showing-up-in-xls-download-file/15567

jnm:
Closed by #2968

PR #2968, opened 2021-01-22
OSM geoJSON export feature and export field filtering

### LINKED ISSUE #2967: Custom data exports API implementation
To fulfill #2881, we need the back end to support everything to be done in #2963.

I propose we add a new, nested endpoint called `/api/v2/assets/<asset uid>/export-settings/` and a new Django model, related to `Asset`, to store each set of settings. I think the model could contain only a `uid` field, a date field, and a JSON field for everything else. The date should be set initially on `POST` and always updated on `PATCH`, even if the `PATCH` doesn't actually change anything. The list view should show the set of settings with the most recent date first. (The reasons behind this stuff are explained more in #2963).

I can help with any clarification; Olivier and I can both help with any Django questions.

joshuaberetta:
KPI backend changes to incorporate geoJSON export and export field filtering update in formpack. The current `/exports` endpoint can now accept a new export type of `geojson` as well as a new `fields` array property used for selecting which fields to include in the export.

Refer to #2975 for the shape of the new export payload and to [formpack#228](https://github.com/kobotoolbox/formpack/pull/228) for the workings of field filtering and geoJSON export.

## Description
Addition of new geoJSON export feature, new default export fields for all non-legacy exports and field filtering for geoJSON, CSV and XLS (non-legacy) exports.

## Related Issues
blocked by [formpack#228](https://github.com/kobotoolbox/formpack/pull/228)
closes #2967, #2517 
related to #2975


kpi/tests/api/v2/test_api_submissions.py
```diff
@@ -679,145 +679,75 @@ def setUp(self):
...
+                    'properties': {'text': 'Excited'},
+                },
             ],
-            "name": "Two points and one text",
-            "type": "FeatureCollection"
         }
-        self.assertDictEqual(expected_output, json.loads(response.content))
+        assert expected_output == json.loads(response.content)
 
     def test_list_submissions_geojson_other_geo_question(self):
```
jnm, line 718: FYI @joshuaberetta I brought this second test back

jnm (approved):

PR #2974, opened 2021-01-26
Always search for root survey when adding to library

### LINKED ISSUE #2973: Questions from inside a group cannot be added to the library
## Description

Users are not able to add a question from inside a group to the library.

## Steps to Reproduce

1. Upload this xlsform (for ease) which has a group question as a survey project.
[Group Question.xlsx](https://github.com/kobotoolbox/kpi/files/5872197/Group.Question.xlsx)
2. Deploy it.
3. Edit the form.
4. Then try to add question to the library. Nothing happens.
5. Remove the group. Try to add question to the library. Nothing happens.
6. Save the changes. Try to add question to the library. Nothing happens.
7. Refresh page. Try to add question to the library. The question is added to the library.

## Expected behavior 

User should be able to add questions to the library even when the question is within a group.

## Actual behavior 

User's are unable to add questions to the library when the question is within a group.

## Additional details

**Community forum:** https://community.kobotoolbox.org/t/can-not-add-questions-to-library-for-reusing-in-other-projects/15458

duvld:
## Description
More so verbose notes for those interested: 
As a consequence to https://github.com/kobotoolbox/kpi/pull/2742, we need the entire survey passed to `unnullifyTranslations` when saving a formbuilder row to the library. The coffee code nests the survey in 2 `._parent` hierarchies for each group (i.e., if a question is in one group, its survey is in `@model.collection._parent_.parent._initalParams`, if nested in two groups its survey would be in `@model.collection._parent_.parent._parent_.parent._initalParams` etc.)

Currently we are sending a static set of 1 `_parent` (which assumes no nested groups). There's a already a function that iterates through each `_parent` and returns the last one. I've used that function here instead.

## Related issues

Fixes #2973

### DIFF
--- a/jsapp/xlform/src/view.row.coffee
+++ b/jsapp/xlform/src/view.row.coffee
@@ -137,7 +137,7 @@ module.exports = do ->
 
     add_row_to_question_library: (evt) =>
       evt.stopPropagation()
-      @ngScope?.add_row_to_question_library @model, @model.collection._parent._initialParams
+      @ngScope?.add_row_to_question_library @model, @model.getSurvey()._initialParams
 
   class GroupView extends BaseRowView
     className: "survey__row survey__row--group  xlf-row-view xlf-row-view--depr"


magicznyleszek:
This fix is crème de la crème :ok:

jnm:
@duvld, I love the verbosity, but I'm going to move it above `## Description` and add a brief (and hopefully easily-digested) message for the fine folks who read our release notes over at https://community.kobotoolbox.org/c/announcements/7 :smile:

PR #2978, opened 2021-01-27
Don't display non-survey assets in the Projects after they are created

### LINKED ISSUE #2976: New Library assets show up in the Projects List
## Description

This is hard to notice, as it happens only a short while after the Library item is created.

## Steps to Reproduce

1. Go to My Library > NEW
2. Create new collection
3. Wait for the loading and go to Projects List
4. Observe new collection is listed under "Drafts" :bug:

## Additional details

This happens because the `searches` react to asset creation without filtering it's type.

magicznyleszek:
## Description

Immediately after creating new Library Item (collection, template, etc.) it was showing up in the Projects List (/forms) under Drafts. Now it is properly filtered out by type to not include wrong ones when reacting to new item creation.

## Related issues

Fixes #2976


jsapp/js/searches.es6
```diff
@@ -115,16 +118,19 @@ function SearchContext(opts={}) {
...
-      this.rebuildCategorizedList(
-        this.state.defaultQueryResultsList,
-        'defaultQueryCategorizedResultsLists'
-      );
-      this.rebuildCategorizedList(
-        this.state.searchResultsList,
-        'searchResultsCategorizedResultsLists'
-      );
+      // only update things if given asset matches the current context types
+      if (this.state.defaultQueryFilterParams?.assetType.includes(asset.asset_type)) {
```
duvld, line 122: Just a heads up, this may cause a conflict with: https://github.com/kobotoolbox/kpi/blob/17cd0bb21de7c9683f6754f4417f188d76f81d39/jsapp/js/searches.es6#L119-L122 (PR #2984)

duvld (approved):

jsapp/js/searches.es6
duvld: @magicznyleszek, a FYI, didn't notice this until after the fact, but when you delete an asset form the asset list, the callback gives you `asset.assetType` instead of the `asset.asset_type` that it gives you when you create an asset.  I've fixed it with https://github.com/kobotoolbox/kpi/commit/b2022fa7c61c3adde9b86db331d5c986f701f54a.

magicznyleszek: Awesome :pray:

PR #2982, opened 2021-02-01
Fix translation not being used in table header

magicznyleszek:
## Description

Fix translation not being used for `select_one` question types in table header filter dropdown.


jnm (approved):
Tested locally; works great :tada:

Issue #2940, opened 2020-12-29
Always allow non-owners to remove their own permissions


jnm:
1. I own an asset
2. I share it with Stephane
3. Stephane doesn't want it
4. Too bad, Stephane is stuck with it!

We should always allow Stephane to remove the asset from his list, by removing his own access to it.

I should not be able to remove my access to my own asset, however. If I don't want an asset that I own anymore, I can delete it.

Related to #2926
Related to #2648

See also: https://community.kobotoolbox.org/t/refuse-and-remove-previsously-shared-form/4380

PR #2984, opened 2021-02-02
Add dropdown menu item that lets non-owners remove their own permissions

duvld:
## Description

Adds a dropdown menu item on the asset list page that lets a non-owner remove forms that are shared with them from their asset list.

## Additional details

image
Clicking this would trigger a prompt asking them to confirm, similar to the delete draft prompt.

Some front end notes: Asset list is populated by an empty string search. In order for this list to update two things must happen:
1. A listener triggers a refreshing of the search list (within `stores.es6`, `this,trigger()` does this)
  - getting `list.es6` to trigger an empty search will also work
2. The search result must be different, `searches.es6` has a callback that removes a deleted asset from the search stores

## Related issues

Front end of #2940


jsapp/js/searches.es6
```diff
@@ -117,9 +118,15 @@ function SearchContext(opts={}) {
       }
     },
     // remove asset from all search store lists
-    removeAsset(asset) {
-      // only update things if given asset matches the current context types
-      if (this.state.defaultQueryFilterParams?.assetType.includes(asset.assetType)) {
+    removeAsset(asset, isNonOwner) {
```
magicznyleszek, line 121: If I get it right - after adding `actions.permissions.removeAssetPermission.completed` listener, this `asset` function parameter becomes asset or uid? So let's change it to `assetOrUid` as we already do in other places.

Just for future safety please add also this at the beginning of the function:

```js
let asset;
if (typeof assetOrUid === 'object') {
  asset = assetOrUid;
} else {
  asset = stores.selectedAsset.asset || stores.allAssets.byUid[assetOrUid];
}

// non-owner self permission removal only gives an assetUid string, not
// an object; for consistency we make it an object here
if (!asset) {
 asset = {uid: assetOrUid};
}
```

That way we always get `asset` object for the rest of the code.

jsapp/js/searches.es6
```diff
@@ -117,9 +118,15 @@ function SearchContext(opts={}) {
...
+      // only update things if given asset matches the current context types or
+      // a non-owner removed their own permissions
+      if (isNonOwner || this.state.defaultQueryFilterParams?.assetType.includes(asset.assetType)) {
+        // non-owner self permission removal only gives an assetUid string, not
+        // an object; for consistency we make it an object here
+        if (isNonOwner) {
+          asset = {uid: asset}
+        }
```
magicznyleszek, line 129: If you follow the previous comment then you don't need this whole `if` here

jsapp/js/stores.es6
```diff
@@ -343,6 +344,20 @@ stores.allAssets = Reflux.createStore({
...
+    if (this.byUid[assetUid] && isNonOwner) {
+      this.byUid[assetUid].deleted = 'true';
+      this.trigger(this.data);
+      window.setTimeout(()=> {
+        this.data = this.data.filter(function(item){
+          return item.uid !== assetUid;
+        });
+        this.trigger(this.data);
+      }, 500);
+    }
```
magicznyleszek, line 359: Instead of all this code, let's do:

```js
// When non owner self removes all his asset permissions, it's as if the asset was deleted for them
if (isNonOwner) {
  this.onDeleteAssetCompleted({uid: assetUid});
}
```

magicznyleszek (changes requested):

jsapp/js/searches.es6
```diff
@@ -117,9 +118,22 @@ function SearchContext(opts={}) {
       }
     },
     // remove asset from all search store lists
-    removeAsset(asset) {
-      // only update things if given asset matches the current context types
-      if (this.state.defaultQueryFilterParams?.assetType.includes(asset.assetType)) {
+    removeAsset(assetOrUid, isNonOwner) {
+      // only update things if given asset matches the current context types or
+      // a non-owner removed their own permissions
+      if (isNonOwner || this.state.defaultQueryFilterParams?.assetType.includes(asset.assetType)) {
```
magicznyleszek, line 124: @duvld I've found a bug here: `asset.assetType` doesn't exist here. It happend to me when I deleted an asset from My Library.

PR #2997, opened 2021-02-04
Navigate back to the library list when searching

jnm:
This doesn't get all the way there—it always searches "My Library"—and it may be evil.

## Description

Change the search placeholder from "Search Library" to "Search My Library" or "Search Public Library" depending on the context. If viewing an individual library item, navigate back to the library list when pushing enter inside the search box.

### DIFF
--- a/jsapp/js/components/header.es6
+++ b/jsapp/js/components/header.es6
@@ -255,7 +255,7 @@ class MainHeader extends Reflux.Component {
             { this.isLibrary() &&
               <div className='mdl-layout__header-searchers'>
                 <SearchBox
-                  placeholder={t('Search Library')}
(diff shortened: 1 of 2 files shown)


magicznyleszek:
FYI for now the Search Box is hidden for single library assets (that includes collections of course) with 60cfed9c6b9ab0c6ccb4ece4c1a724db97d07bb1.

jnm:
I've flagged this as `donotmerge` for now. We can have a UX discussion about whether it's worthwhile (and make sure that it's implemented correctly) after the initial release of public collections.

magicznyleszek:
This is already fixed, and the PR is very old - closing 📕

PR #3002, opened 2021-02-05
Allow superusers to migrate projects from legacy interface through the API

noliveleger:
## Description

In order to help the support team to migrate users' projects from legacy, an endpoint is added. `/api/v2/users/<username>/migrate`
It calls the management command `sync_kobocat_xforms` as a background task (with Celery)

To launch a task:
`/api/v2/users/<username>/migrate`

To check its state:
`/api/v2/users/<username>/migrate/<task_id>/`


kobo/settings/base.py
```diff
@@ -454,6 +454,7 @@ def __init__(self, *args, **kwargs):
         }
 
 CELERY_BROKER_URL = os.environ.get('KPI_BROKER_URL', 'redis://localhost:6379/1')
+CELERY_RESULT_BACKEND = CELERY_BROKER_URL
```
jnm, line 457: I see we do this already [in KoBoCAT](https://github.com/kobotoolbox/kobocat/blob/d91834ccb174da0fd0305f8e8ae11ed93cce10c0/onadata/settings/kc_environ.py#L51), but does it mean that we store all the results from every task forever? I just checked on HHI production, and there are 4400 keys in the Redis DB, containing stuff like:
```
127.0.0.1:6379[2]> GET celery-task-meta-62bd2cf9-1f53-4bf8-9e47-1a0c9554a1b3
"{\"status\": \"SUCCESS\", \"traceback\": null, \"result\": true, \"task_id\": \"62bd2cf9-1f53-4bf8-9e47-1a0c9554a1b3\", \"children\": []}"
```
I also saw in the [documentation](https://docs.celeryproject.org/en/v4.3.0/userguide/tasks.html?highlight=result#result-backends):
> Backends use resources to store and transmit results. To ensure that resources are released, you must eventually call get() or forget() on EVERY AsyncResult instance returned after calling a task.

kpi/views/v2/asset.py
```diff
@@ -618,6 +620,68 @@ def metadata(self, request):
         metadata = self.get_metadata(queryset)
         return Response(metadata)
 
+    @action(detail=False, methods=['GET'],
+            renderer_classes=[renderers.JSONRenderer],
+            url_path=r'migrate(?:/(?P<task_id>[\d\w\-]+))?')
+    def migrate(self, request, task_id: str = None, **kwargs):
```
jnm, line 626: Just wondering: why in the `AssetViewSet`? It feels crowded in here :blush: 
giphy

kpi/views/v2/asset.py
```diff
@@ -618,6 +620,68 @@ def metadata(self, request):
...
+        1. Call this endpoint with `?username=<username>`
+        2. Fetch url provided to check the state of the Celery task.
+           It can be:
+            - 'PENDING'
+            - 'FAILED'
+            - 'SUCCESS'
+
+        """
+        if not request.user.is_superuser:
+            raise Http404
```
jnm, line 641: Not a big deal, but I think the rationale behind returning a 404 response when denying access is that revealing the existence of an endpoint would be a privacy concern in some cases. I'm not sure this is one of those cases, though.

kpi/views/v2/asset.py
```diff
@@ -618,6 +620,68 @@ def metadata(self, request):
...
+
+        if task_id:
+            from celery.result import AsyncResult
+            res = AsyncResult(task_id)
+            if res:
+                return Response(
+                    {
+                        'status': res.state
+                    },
+                )
```
jnm, line 651: All hail our new god, psf/black? Yikes. Can't say I like this formatting at all.

kpi/views/v2/asset.py
```diff
@@ -618,6 +620,68 @@ def metadata(self, request):
...
+        """
+
+        Really basic endpoint what allows super users to migration projects
+        from KoBoCAT.
+
+        1. Call this endpoint with `?username=<username>`
+        2. Fetch url provided to check the state of the Celery task.
+           It can be:
+            - 'PENDING'
+            - 'FAILED'
```
jnm, line 636: I forgot to restart my Celery workers when testing locally, so my attempt failed with `TypeError: sync_kobocat_xforms() got an unexpected keyword argument 'populate_xform_kpi_asset_uid'`. However, the status never changed from `PENDING` to `FAILED` (or anything else).

I know this is a throw-away, temporary endpoint, but I guess we should be aware that the Celery `res.state` isn't too reliable—unless it's waiting several minutes to retry the task(?)

kobo/settings/base.py
noliveleger: I thought I did :-(  You are totally right. The keys should be dropped / should expire.

kpi/views/v2/asset.py
noliveleger: Because it's related to assets. 
BTW, I really love the GIF. 
Any suggestions to move it somewhere else?

kpi/views/v2/asset.py
noliveleger: Ok, so 403 it will be :-)

kpi/views/v2/asset.py
noliveleger: Oops, no it should have been on one line. I was testing a new shortcut to enable `Black` on those lines. I forgot to revert it.

kobo/settings/base.py
noliveleger: @jnm They are automatically removed after 1 day (by default) by `celery beat`. Obviously, we can lower this value.

https://docs.celeryproject.org/en/latest/userguide/configuration.html#result-expires
> A built-in periodic task will delete the results after this time (celery.backend_cleanup), assuming that celery beat is enabled. The task runs daily at 4am.

kpi/views/v2/asset.py
noliveleger: I moved it under `/api/v2/users/<username>/`

kpi/views/v2/asset.py
noliveleger: I updated the comments

kobo/settings/base.py
jnm: Cool, I didn't know about that. Thanks. I think 1 day is fine

kpi/views/v2/asset.py
jnm: Thanks

jnm (approved):

PR #3007, opened 2021-02-09
WIP moved fullscreen button

duvld:
## Description

An attempt at moving the fullscreen button to the header (now breadcrumbs)


jsapp/scss/components/_kobo.breadcrumbs.scss
```diff
@@ -11,6 +11,16 @@
     background: $kobo-whitegray;
     padding: #{($s-breadcrumbs-height - 40px) * 0.5} 20px;
     height: $s-breadcrumbs-height;
+
+    justify-content: space-between;
+    button {
+      @extend .mdl-button; // Why is having the className insufficent?
```
magicznyleszek, line 17: You have a typo in JSX code: `classname` instead of `className` :)

jsapp/scss/components/_kobo.bem.ui.scss
```diff
@@ -683,4 +683,9 @@ span.svg-icon {
       display: block;
     }
   }
+
+  &.dropzone--fullscreen {
+    @extend .assets-table, .assets-table--fullscreen;
+    position: fixed !important; // Why does the extended position not override?
```
magicznyleszek, line 689: It's because dropzone adds itself an inline style attribute with `position: relative;` and inline styles have stronger specificity than classes

jsapp/js/components/library/myLibraryRoute.es6
```diff
@@ -78,18 +84,34 @@ class MyLibraryRoute extends React.Component {
       );
     }
 
+    // Modifying Dropzone like this feels a bit weird, should we add a wrapper?
```
magicznyleszek, line 87: It sucks, but a wrapper would probably be a bit better solution, as if we update `react-dropzone` npm package then there is a tiny chance this will break.

magicznyleszek (changes requested):
I'm trying to think if this is a good solution, or if we should make it reusable in some way.

If this should stay, we will need a "don't show fullscreen button" property for `assetsTable`… But maybe a better way would be to completely remove "Toggle Fullscreen" from `AssetsTable` and make a fullscreenable wrapper component similar that accepts any content inside and displays a button.

Currently the "Toggle Fullscreen" button appears in few places:
- Project's Data > Table (`.form-view` fullscreen modifier, custom button)
- Project's Data > Reports (`.form-view` fullscreen modifier, custom button)
- Project's Data > Map (`.form-view` fullscreen modifier, custom button)
- My Library (`AssetsTable` built-in footer button)
- Public Collections (`AssetsTable` built-in footer button)
- Single Collection (`AssetsTable` built-in footer button)

We don't support two fullscreen buttons at the same time plus we must remember to un-fullscreen when changing routes.

We might add an action for fullscreening and a store for keeping this information. Each fullscreenable component instance would need to have an id.

magicznyleszek:
Closing this, because it's old and we most probably are going to do some other way to handle this

PR #3017, opened 2021-02-18
added multicheckbox component

duvld:
## Description

Adds component that takes in an array of items to be displayed in a list with a checkbox.

### DIFF
--- a/jsapp/js/components/multiCheckbox.es6
+++ b/jsapp/js/components/multiCheckbox.es6
@@ -29,10 +29,10 @@ class MultiCheckbox extends React.Component {
 
   render() {
     return (
-      <ul className='multi-checkbox multi-checkbox__list'>
+      <ul className='multi-checkbox'>
         {this.props.items.map((item, itemIndex) => {
           return (
-            <li className='multi-checkbox__list-item' key={itemIndex}>
+            <li className='multi-checkbox__item' key={itemIndex}>
               <Checkbox
                 checked={item.checked}
                 disabled={item.disabled}
--- a/jsapp/js/components/multiCheckbox.es6
+++ b/jsapp/js/components/multiCheckbox.es6
@@ -21,33 +21,28 @@ class MultiCheckbox extends React.Component {
     autoBind(this);
   }
 
-  onChange(response) {
+  onChange(itemIndex, isChecked) {
     let updatedList = this.props.items;
-    updatedList[response].checked = !updatedList[response].checked;
+    updatedList[itemIndex].checked = isChecked;
     this.props.onChange(updatedList);
   }
 
   render() {
     return (
-      <div className='project-downloads__column project-downloads__column--right'>
-        <span className='project-downloads__title'>
-          {this.props.label}
-        </span>
-        <ul className='project-downloads__questions-list'>
-          {this.props.items.map((item, n) => {
-            return (
-              <li key={n}>
-                <Checkbox
-                  disabled={item.disabled || false}
-                  checked={item.checked || false}
-                  onChange={this.onChange.bind(this, n)}
-                  label={item.label}
-                />
-              </li>
-            );
-          })}
-        </ul>
-      </div>
+      <ul className='multi-checkbox multi-checkbox__list'>
+        {this.props.items.map((item, itemIndex) => {
+          return (
+            <li className='multi-checkbox__list-item' key={itemIndex}>
+              <Checkbox
+                checked={item.checked}
+                disabled={item.disabled}
+                onChange={this.onChange.bind(this, itemIndex)}
+                label={item.label}
+              />
+            </li>
+          );
+        })}
+      </ul>
     );
   }
 }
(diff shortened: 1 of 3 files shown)


jsapp/js/components/multiCheckbox.es6
```diff
@@ -0,0 +1,55 @@
...
+    this.props.onChange(updatedList);
+  }
+
+  render() {
+    return (
+      <div className='project-downloads__column project-downloads__column--right'>
+        <span className='project-downloads__title'>
+          {this.props.label}
+        </span>
+        <ul className='project-downloads__questions-list'>
```
magicznyleszek, line 36: We need these class names to be something specific, like `multi-checkbox`, `multi-checkbox__list`, `multi-checkbox__list-item`, for sure not reusing the `project-downloads` ones here.

jsapp/js/components/multiCheckbox.es6
```diff
@@ -0,0 +1,55 @@
...
+    updatedList[response].checked = !updatedList[response].checked;
+    this.props.onChange(updatedList);
+  }
+
+  render() {
+    return (
+      <div className='project-downloads__column project-downloads__column--right'>
+        <span className='project-downloads__title'>
+          {this.props.label}
+        </span>
```
magicznyleszek, line 35: If you want the list to have a label, it would need to be optional.

jsapp/js/components/multiCheckbox.es6
```diff
@@ -0,0 +1,55 @@
...
+        <ul className='project-downloads__questions-list'>
+          {this.props.items.map((item, n) => {
+            return (
+              <li key={n}>
+                <Checkbox
+                  disabled={item.disabled || false}
+                  checked={item.checked || false}
```
magicznyleszek, line 42: I think we shouldn't have `|| false` here. For `checked` it is required, so always will be set. As for `disabled` I think that if `item.disabled` will be `undefined`, then React will handle it itself quite well and just not add the `disabled` attribute.

jsapp/js/components/multiCheckbox.es6
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  render() {
+    return (
+      <div className='project-downloads__column project-downloads__column--right'>
+        <span className='project-downloads__title'>
+          {this.props.label}
+        </span>
+        <ul className='project-downloads__questions-list'>
+          {this.props.items.map((item, n) => {
```
magicznyleszek, line 37: Just a nitpick, but instead of `n` let's make it `itemIndex` :)

jsapp/js/components/multiCheckbox.es6
```diff
@@ -0,0 +1,55 @@
...
+ * @prop {MultiCheckboxItem[]} items
+ * @prop {function} onChange - returns whole list whenever any item changes
+ */
+class MultiCheckbox extends React.Component {
+  constructor(props){
+    super(props);
+    autoBind(this);
+  }
+
+  onChange(response) {
```
magicznyleszek, line 24: It's nice you've bound the `onChange` to index, but it means it should be `(itemIndex, isChecked)`.

jsapp/js/components/multiCheckbox.es6
```diff
@@ -0,0 +1,55 @@
...
+    let updatedList = this.props.items;
+    updatedList[response].checked = !updatedList[response].checked;
```
magicznyleszek, line 26: This would be simpler: `updatedList[itemIndex].checked = isChecked;`

jsapp/js/components/multiCheckbox.es6
magicznyleszek: I think you can add new `.scss` file for `multi-checkbox`, don't mind it will be so small :)

magicznyleszek (changes requested):

jsapp/scss/components/_kobo.multi-checkbox.scss
```diff
@@ -0,0 +1,18 @@
...
+  position: absolute;
+  top: 20px;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  width: 100%;
+  overflow: hidden auto;
+  border: 1px solid $kobo-lightgray;
+  border-radius: 2px;
+  margin-top: 6px;
```
magicznyleszek, line 13: I think you need to drop `position`, `top`, `left`, `right`, `bottom` and `margin-top` as these should be set by parent component depending on context (e.g. for Project Downloads we have flexbox layouting, not absolute positioning)

jsapp/scss/components/_kobo.multi-checkbox.scss
```diff
@@ -0,0 +1,18 @@
+.multi-checkbox {
```
magicznyleszek, line 1: You need to add `@import "./components/kobo.multi-checkbox";` somewhere in `jsapp/scss/main.scss`

magicznyleszek (changes requested):

Issue #2995, opened 2021-02-03
[Public Collections] 


magicznyleszek:
## Description

Filters contain only options available in currently filtered data and ~~maybe~~ most probably it's a bad idea.
This was brought up in #2872 (see [this](https://github.com/kobotoolbox/kpi/pull/2872#issuecomment-771590412), [this](https://github.com/kobotoolbox/kpi/pull/2872#issuecomment-772109171), [this](https://github.com/kobotoolbox/kpi/pull/2872#issuecomment-772601426) and [that](https://github.com/kobotoolbox/kpi/pull/2872#issuecomment-772650343) comment).

PR #3026, opened 2021-02-23
Better filter header clearing UX

### LINKED ISSUE #2996: [Public Collections] Assets Table column filter clearing UX improvement
## Description

After you choose filter option from a dropdown (e.g. "Francais (fr)" language) it's not clear how to unselect it (so to be "any language"). _Currently it requires to select the same dropdown option again._ Possible solutions:
  - wait for #2990 and make clearing happen in the search box _(Note: what if we don't have search box visible, but we have assets table with filters - like at single collection route?)_
  - always include "All <name>" (e.g. "All languages") option that will clear the filter
  - add some "x" clear button next to a filter if it has some selection

## Additional details

Connected issues:
- #2835
- #2990

magicznyleszek:
## Description

Adds a small "x" button for clearing filter header in table.

## Related issues

Fixes #2996

### DIFF
--- a/jsapp/js/components/library/assetsTable.es6
+++ b/jsapp/js/components/library/assetsTable.es6
@@ -124,6 +124,11 @@ export default class AssetsTable extends React.Component {
     }
   }
 
+  onClearFilter(evt) {
+    evt.stopPropagation();
+    this.props.onFilterChange(null, null);
+  }
+
   /**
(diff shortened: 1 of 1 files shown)


jnm (approved):
:ok_hand: tested locally and this lets us dodge #2995 for now because clearing the filter entirely (and resetting the drop down to its initial state) is required before switching from one criterion to another

Issue #3023, opened 2021-02-19
Add UI option to remove line breaks when exporting CSV


jnm:
Internal discussion: https://chat.kobotoolbox.org/#narrow/stream/6-Support/topic/Carriage.20Return.20Creates.20additional.20Rows.20on.20CSV.20export/near/17828

Depends on #3022 

This should be in "Advanced options" and only available when CSV is the selected format.

The wording could be "Include line breaks" or "Include line breaks in resp[onses]" depending on what fits.

magicznyleszek:
Note: you can use #3029 as base branch

PR #3029, opened 2021-02-23
Toggle line breaks in csv export

### LINKED ISSUE #3022: Add back-end option to strip line breaks from CSV exports
Could be done here (in kpi) or in formpack. The goal is to replace `\r` and `\n` with spaces in all response values.
Internal discussion: https://chat.kobotoolbox.org/#narrow/stream/6-Support/topic/Carriage.20Return.20Creates.20additional.20Rows.20on.20CSV.20export

- [ ] Add option to export machinery
- [ ] Accept option from API when creating new export task

joshuaberetta:
Allow for stripping of line breaks in CSV exports as discussed [here](https://chat.kobotoolbox.org/#narrow/stream/6-Support/topic/Carriage.20Return.20Creates.20additional.20Rows.20on.20CSV.20export/near/17591) and brought up by the community [here](https://community.kobotoolbox.org/t/text-data-type-and-csv-data-export/13963/14). A new `preserve_breaks` parameter can be set in `/export-settings` and sent to `/exports` that, when `True` will leave the newline and carriage-return characters if they are present and if `False` will remove them before exporting to CSV.

## Description
New CSV export option to control whether or not the newline and carriage-return characters should be preserved.

## Related issues
closes #3022 
back-end support for #3023


noliveleger:
@joshuaberetta, Can you fix the conflicts. I will review it.

PR #3034, opened 2021-02-24
Avoid resetting dates when migrating collections to assets

### LINKED ISSUE #3033: Migrations changing the modification date of all collections
News flash: Django's `bulk_create()` still respects `auto_now` and `auto_now_add`. This causes our collections-to-asset migration to reset the `date_created` and `date_modified` of all collections: https://github.com/kobotoolbox/kpi/blob/01619de2626f197e409feb7283b84dd65fad1a3e/kpi/migrations/0029_copy_collections_to_assets.py#L92

Internal discussion: https://chat.kobotoolbox.org/#narrow/stream/4-KoBo-Dev/topic/Public.20collections.20final.20final.20topic/near/18210

jnm:
## Description

Preserve the creation and modification dates of collections when altering the database structure. If all goes well, this change has no visible effects.

## Related issues

Fixes #3033

### DIFF
--- a/kpi/migrations/0029_copy_collections_to_assets.py
+++ b/kpi/migrations/0029_copy_collections_to_assets.py
@@ -1,11 +1,12 @@
 import sys
 from collections import defaultdict
 
+import django.db.models.deletion
 from django.conf import settings
 from django.db import migrations, models
-import django.db.models.deletion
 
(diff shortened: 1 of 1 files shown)


jnm:
This is a minor change using an existing context manager. Merged :hammer:

PR #3035, opened 2021-02-24
return warning if missing attachment

### LINKED ISSUE #3028: The screen goes white when clicking on the GPS location from DATA>Map
## Description

The screen goes white when clicking on the GPS location from **DATA>Map**.

## Steps to Reproduce

1. Got to **DATA>Map**.
2. Click to one of the GPS locations seen under the **DATA>Map**.
3. Screen goes white.

## Expected behavior 

While clicking to one of the GPS locations from **DATA>Map**, the entries made for that particular submission should pop up.

## Actual behavior 

The screen goes white when clicking on the GPS location from **DATA>Map**.

## Additional details

This is the issue that I see in the console:

image

duvld:
See [internal discussion](https://chat.kobotoolbox.org/#narrow/stream/17-Developer-Community/topic/Data.20Display.20issues/near/18326) for detailed explanation 

## Description

Adds a warning to prevent answered media questions with no attachments from crashing the page

## Related issues
Fixes #3028


jsapp/js/components/submissionDataTable.es6
```diff
@@ -266,14 +266,19 @@ class SubmissionDataTable extends React.Component {
...
+            <img src={attachment.download_small_url}/>
+          </a>
+        );
+      } else {
+        return (<a href={attachment.download_url} target='_blank'>{filename}</a>);
+      }
+    // In the case that an attachment is missing, don't crash the page
     } else {
-      return (<a href={attachment.download_url} target='_blank'>{filename}</a>);
+      return(t('Could not retrieve attachment'));
```
jnm, line 281: maybe i'd be valuable to still see the (un-clickable) filename?

duvld: Good idea :+1:

jsapp/js/components/submissionDataTable.es6
```diff
@@ -266,14 +266,19 @@ class SubmissionDataTable extends React.Component {
...
+            <img src={attachment.download_small_url}/>
+          </a>
+        );
+      } else {
+        return (<a href={attachment.download_url} target='_blank'>{filename}</a>);
+      }
+    // In the case that an attachment is missing, don't crash the page
     } else {
-      return (<a href={attachment.download_url} target='_blank'>{filename}</a>);
+      return(t('Could not retrieve') + ' ' + filename);
```
jnm, line 281: @duvld this is definitely a nitpick when applied to this error-handling code, but it's more important in other contexts: if you include a placeholder within the translatable string, and then replace that placeholder, our translators can do a better job, especially with right-to-left languages. For example:
```javascript
      return(t('Could not retrieve ##filename##').replace('##filename##', filename));
```

The convention used to be a triple underscore as the placeholder, but doing a quick search makes me think we evolved to `##thing##`. @magicznyleszek is that right? Also, is this PR okay to merge?

magicznyleszek: Yes, we are using `##name##` as this makes it clear what we are looking for in the variable plus it is easier to have multiple variables in same string

magicznyleszek: @duvld please make the change John is suggesting and then you can merge this :ship: :it:

PR #3049, opened 2021-03-05
Validate default response value received when creating/updating assets

### LINKED ISSUE #3048: Stored XSS which takes hostile data
## Description of the Vulnerability
Stored XSS which takes hostile data, stores it in a file, a database, or another back-end system, and then at a later stage, displays the data to the user, unfiltered.

Bug Type: Stored XSS
Bug Severity: High
Affected Users: All users

Location of bug: https://forms.ona.io/#/forms/<asset_uid>/edit
Attack string (payloads): `prompt(document.cookie);>#\">#\">#\">`
## Steps to Reproduce
1. Launch a form on Formbuilder
2. Add Question and Select File upload
3. Put your payload in Default Response and Click Save

## Expected behavior 

Creating/ Updating assets with malicious content should fail on save

## Actual behavior 

kpi_vulnerability

WinnyTroy:
## Checklist

3. [x] Ensure the tests pass

## Description

Validate default response value received when creating/updating assets

## Related issues

Fixes #3048

### DIFF
--- a/kpi/models/asset.py
+++ b/kpi/models/asset.py
@@ -383,6 +383,16 @@ class FormpackXLSFormUtils:
             raise ValueError('Duplicate translation: {}'.format(_to))
         _ts[_ts.index(_from)] = _to
 
+    def _contains_invalid_chars(self, content):
+        for row in content['survey']:
+            try:
+                if row['default'] and bool(re.search(
+                        r'[<|>|&]', row['default'])):
+                    raise ValidationError(
+                        'XForm questions settings may contain malicious content')
+            except KeyError:
+                pass
+
 
 class XlsExportable:
     def ordered_xlsform_content(self,
@@ -680,6 +690,7 @@ class Asset(ObjectPermissionMixin,
         self._autoname(self.content)
         self._unlink_list_items(self.content)
         self._remove_empty_expressions(self.content)
+        self._contains_invalid_chars(self.content)
 
         settings = self.content['settings']
         _title = settings.pop('form_title', None)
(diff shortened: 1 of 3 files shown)


kpi/models/asset.py
```diff
@@ -383,6 +383,16 @@ def _rename_translation(self, content, _from, _to):
             raise ValueError('Duplicate translation: {}'.format(_to))
         _ts[_ts.index(_from)] = _to
 
+    def _contains_invalid_chars(self, content):
+        for row in content['survey']:
+            try:
+                if row['default'] and bool(re.search(
```
jnm, line 389: Wouldn't other columns need to be sanitized as well, if we took this approach?

jnm (commented):
Thank you for proposing this contribution, but I do not think we will merge these changes because:
* KoBoToolbox does not appear to be affected by #3048 (please re-open that issue with further information if you disagree)
* Maintaining a list of banned characters for various fields—I think more than `default` would be needed—does not seem workable: there are too many legitimate reasons why form authors might need to use any printable character.
* The responsibility to ensure that user-provided input does not get executed by the browser lies with the front-end code. It's not wrong for the back end to accept any string as a default value, and it's not wrong for the back end to pass a verbatim error message—which may include a snippet of user-provided input—from ODK Validate to the front end. However, the front end must take care when displaying that error message so that the browser does not interpret it as HTML or JavaScript (or CSS, etc.).

In short: if a fix in KPI is needed, it should be made in the ES6 code that displays the error message.

PR #3054, opened 2021-03-08
Fix submission duplication error

### LINKED ISSUE #3053: Fix submission duplication 500 error
## Description

User receives 500 error when attempting to duplicate submissions that do not contain a `start` and `end` element in the XML tree.

## Steps to Reproduce

1. Create a survey with xlsform and exclude `start` and `end` fields
2. Upload to Kobo and deploy
3. Submit data
4. Try and duplicate a submission

## Expected behavior 

The submission is duplicated.

## Actual behavior 

500 Error.

## Additional details

Source of issue:
https://github.com/kobotoolbox/kpi/blob/90fd6c679d11171a792e06ec084c298428a14ae5/kpi/deployment_backends/kobocat_backend.py#L696-L698

joshuaberetta:
## Description

Fix 500 error when attempting to duplicate a submission without `start` or `end` fields.

## Related issues
closes #3053

### DIFF
--- a/kpi/deployment_backends/kobocat_backend.py
+++ b/kpi/deployment_backends/kobocat_backend.py
@@ -694,13 +694,16 @@ class KobocatDeploymentBackend(BaseDeploymentBackend):
         # `start` and `end` are not guaranteed to be included in the XML object
         _uuid, uuid_formatted = self.generate_new_instance_id()
         date_formatted = self.format_openrosa_datetime()
-        update_fields = [
-            ('start', date_formatted),
-            ('end', date_formatted),
-            ('meta/instanceID', uuid_formatted),
-        ]
-        for element_str, update_str in update_fields:
-            self.__update_xml_text(xml_parsed, element_str, update_str)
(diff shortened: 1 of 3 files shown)


jnm (approved):
@joshuaberetta to avoid delays, I made a few changes directly in https://github.com/kobotoolbox/kpi/pull/3054/commits/279278c5fad4b24ebfe1a960868d551db2639d32. I wish we could include a unit test for this, but it seems like we'd have to rewrite a lot of the `MockDeploymentBackend` to use XML.

joshuaberetta:
> @joshuaberetta to avoid delays, I made a few changes directly in https://github.com/kobotoolbox/kpi/pull/3054/commits/279278c5fad4b24ebfe1a960868d551db2639d32. I wish we could include a unit test for this, but it seems like we'd have to rewrite a lot of the `MockDeploymentBackend` to use XML.

@jnm that's perfect — thanks! I know... I started looking at that but realised it would slow down deploying the patch 😅

PR #3060, opened 2021-03-10
Hide non-data rows from bulk edit form

magicznyleszek:
## Description

The `hidden`, `note` and `calculate` questions don't carry submission data, so we hide them from the Bulk Edit Submissions form (and Single Submission Modal).

### DIFF
--- a/jsapp/js/components/modalForms/bulkEditSubmissionsForm.es6
+++ b/jsapp/js/components/modalForms/bulkEditSubmissionsForm.es6
@@ -6,10 +6,7 @@ import {
   getSurveyFlatPaths,
   getFlatQuestionsList,
 } from 'js/assetUtils';
-import {
-  QUESTION_TYPES,
-  FORM_VERSION_NAME,
-} from 'js/constants';
+import {QUESTION_TYPES} from 'js/constants';
 import {bem} from 'js/bem';
 import {actions} from 'js/actions';
 import TextBox from 'js/components/common/textBox';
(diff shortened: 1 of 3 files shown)


jsapp/js/components/modalForms/bulkEditSubmissionsForm.es6
```diff
@@ -173,10 +176,23 @@ class BulkEditSubmissionsForm extends React.Component {
...
    */
   getDisplayData() {
-    const questions = getFlatQuestionsList(this.props.asset.content.survey);
+    let questions = getFlatQuestionsList(this.props.asset.content.survey);
     const flatPaths = getSurveyFlatPaths(this.props.asset.content.survey);
 
-    questions.forEach((question) => {
+    questions = questions.filter((question) => {
+      // we don't want to include special calculate row used to store form version
+      if (question.type === QUESTION_TYPES.calculate.id && question.name === FORM_VERSION_NAME) {
```
duvld, line 184: There is a small edge case that bypasses this check: If a form has multiple versions like [this one](https://github.com/kobotoolbox/kpi/files/6131190/aEYQ8CkLw3QKmynaXUAw8U.xlsx) for example

duvld:
Should we put this on master as well? :D

jsapp/js/components/modalForms/bulkEditSubmissionsForm.es6
magicznyleszek: This is a great comment. It actually makes more sense to just hide all `calculate` rows here (and in the single submission modal)!

magicznyleszek:
> @magicznyleszek Should we put this on master as well? :D

Nah, Bulk Edit is on `beta` only, and the changes I did for the Single Submission Modal are really tiny

PR #3076, opened 2021-03-16
Better routes checking

magicznyleszek:
## Description

Alternative solution to react router's `isActive` for checking wheter a route is your current route. Needed for non-components.

## Related issues

Started because of #3012


jsapp/js/components/library/singleCollectionStore.es6
```diff
@@ -65,7 +65,7 @@ const singleCollectionStore = Reflux.createStore({
    * otherwise wait until route changes to a library (see `onRouteChange`)
    */
   startupStore() {
-    if (!this.isInitialised && isOnLibraryAssetRoute() && !this.data.isFetchingData) {
+    if (!this.isInitialised && isAnyLibraryItemRoute() && !this.data.isFetchingData) {
```
duvld, line 68: I think each `isAnyLibraryItemRoute` in this file should be replaced with `isAnyLibraryRoute`?

duvld (changes requested):

jsapp/js/components/library/singleCollectionStore.es6
magicznyleszek: `isOnLibraryAssetRoute` was looking for `/library/asset/…`, and `isAnyLibraryItemRoute` is looking for `/library/asset/<uid>…`. There is no `/library/asset/` without the `<uid>` part so these two are actually identical. And `isAnyLibraryRoute` is only looking for`/library…` so it would be not strict enough :)

duvld: :+1: gotcha on the logic, I didn't think of that--I should've been more clear ;)

This function always returns `false` because, with an empty parameter, `ROUTES.LIBRARY_ITEM.replace(':uid', uid)` always will return `/library/asset/undefined/` so it's an issue with how we handle empty params in the function instead

magicznyleszek: Oh right! I wasn't passing the `uid` to the function. But actually I don't really care if the uid matches, I only want to know if it's `/library/asset/<anything>`. I updated the `isAnyLibraryItemRoute` function to not expect a parameter

Issue #2767, opened 2020-08-24
Basic case management


jnm:
1) Research and understand relevant X(LS)Form features, e.g. `xml-external` type
2) Add necessary settings and API endpoints to link projects together
3) Add UI for linking projects
4) Modify XForm XML generation code as required

Ig-Rebollo:
@tinok @jnm @dorey here is my first attempt to incorporate this functionality. For the time being I simply focused on what's necessary to make this functionality available. I am also working on a more user friendly way to add the data to the questions once the files are link, though that will come later.

This is the easiest way, or MVP, as we discussed it in our call on Monday. First, a tab in settings:
![MacBook Pro - 1](https://user-images.githubusercontent.com/22472326/92284402-e4a6d080-eecf-11ea-9b1b-0ce0c73f8a4c.jpg)

You can then choose to make a project available for 'referencing' (though I'm sure we can come up with a better word) and who can view this data and reference the project:
![MacBook Pro - 2](https://user-images.githubusercontent.com/22472326/92284504-19b32300-eed0-11ea-8e67-00f8758914a3.jpg)

Then you can go to project B and select from the dropdown:
![MacBook Pro - 3](https://user-images.githubusercontent.com/22472326/92284538-2df72000-eed0-11ea-9e09-f5ea932bf86c.jpg)

And review which documents you have linked so far:
![MacBook Pro - 4](https://user-images.githubusercontent.com/22472326/92284580-436c4a00-eed0-11ea-85b3-0180d28e7cce.jpg)

I'm still working on further designs/options but do let me know what you think!

magicznyleszek:
Looks really promising! And I like the icon.

Maybe instead of "reference" we could use word "link"? We would have "Data linking" in the sidebar menu.

Ig-Rebollo:
I like "data linking", what do you think @tinok 

Also, are there any additional functionalities I missed for this initial option?

tinok:
This does look promising @Ig-Rebollo! My initial thoughts from looking at this over the last days: 

* The main thing I'm missing is the ability to only share some columns (and / or to only include some columns)
* We need a single name for this that's used in all places (the left menu and the main panel). Right now it's: `Record management`, `project as a reference`, and `link to other projects`. It sounds like different features. Let's use "Data linking" for now throughout.
* In the main screen on the right I think we need first an intro paragraph that explains the concept (with a link to a support page) and then two functionalities: 
    * Share data with other projects
    * Import data from another project
[...]

Issue #2983, opened 2021-02-01
The second coming of Prettier


magicznyleszek:
## Description

We need to reintroduce¹ [prettier](https://prettier.io/) in a smart way. The plan:

1. We need to apply the changes in single commit, best to apply to `master` and all open branches
2. This needs to be scripted so we will be able to repeat this whenever we upgrade `prettier`
3. The script needs to be split in two phases, as after running prettier on codebase it requires some manual fixes
4. It's best no other commits are pushed during that, so it should be left to be done on a weekend or at night
5. It's best to test out the script on some other repository first

Some inspiration:
- An article about using prettier on a big codebase in the first time: https://flexport.engineering/upgrading-prettier-on-a-large-codebase-28d56c4de49e

## Additional details

¹We initially introduced prettier back in 2018 with #1778 but soon it was reverted with #1790. The problem with the first attempt was that prettier was supposed to run on pre-commit hook, thus making the changes be applied in a not very controlled way.

bufke:
We should not let perfect be the enemy of good enough and implement prettier. I don't care about how we approach it but with permission I would apply it to all ts/tsx files in one commit and add a CI script that checks it.

PR #3117, opened 2021-03-30
[Feature] Column filtering for basic case management

duvld:
Merge and test #3057 first!

## Description

Extends base UI for exporting/importing data across connected projects with column (question) selection

## Related issues

Column filtering of #2767


jsapp/js/components/modalForms/dataAttachmentColumnsForm.es6
```diff
@@ -0,0 +1,226 @@
...
+      actions.dataShare.attachToParent(this.props.asset.uid, data);
+    }
+  }
+
+  render() {
+    return (
+      <bem.FormModal__form m='data-attachment-columns'>
+        <div className='header'>
+          <span className='modal-description'>
+            {t('You are about to import ##PARENT_NAME##. Select or deselect in the list below to narrow down the number of questions to import.').replace('##PARENT_NAME##', this.props.parent.name)}
```
joshuaberetta, line 172: I'm not sure if you guys try and stick to a text-width of 80, but maybe format this to something like the following (not sure if this works the same as in Python)?

```javascript
t(
  'You are about to import ##PARENT_NAME##. Select or deselect in',
  'the list below to narrow down the number of questions to import.'
).replace('##PARENT_NAME##', this.props.parent.name)
```

or

```javascript
t(
  `You are about to import ${this.props.parent.name}. Select or deselect`,
  'in the list below to narrow down the number of questions to import.'
)
```

jsapp/js/components/modalForms/dataAttachmentColumnsForm.es6
```diff
@@ -0,0 +1,226 @@
...
+          </span>
+
+          <div className='bulk-options'>
+            <span className='bulk-options__description'>
+              {t('Select below the questions you want to import')}
```
joshuaberetta, line 177: Maybe rather one of these?:
1. Select the questions you want to import;
2. Select the questions you want to import from the list below;
3. Select the questions you want to import from below;
4. Select which questions to import;
5. Select which questions to import from below;

jsapp/js/actions/datashare.es6
```diff
@@ -55,13 +78,24 @@ dataShareActions.getSharingEnabledAssets.failed.listen(() => {
...
   alertify.error(response?.responseJSON?.detail || t('Failed to toggle sharing'))
 });
 
+dataShareActions.updateColumnFilters.listen((uid, data) => {
+  dataInterface.patchDataSharing(uid, data)
+    .done(dataShareActions.updateColumnFilters.completed)
+    .fail(dataShareActions.updateColumnFilters.failed);
+});
+dataShareActions.updateColumnFilters.failed.listen((response) => {
+  alertify.error(response.responseJSON);
```
joshuaberetta, line 98: Is `response` and the attribute `responseJSON` guaranteed? Above you have:

```javascript
  alertify.error(response?.responseJSON?.detail || t('Failed to toggle sharing'))
```

jsapp/js/constants.es6
```diff
@@ -70,6 +70,7 @@ export const MODAL_TYPES = {
   ASSET_TAGS: 'asset-tags',
   ENCRYPT_FORM: 'encrypt-form',
   BULK_EDIT_SUBMISSIONS: 'bulk-edit-submissions',
+  DATA_ATTACHMENT_COLUMNS: 'data-attachment-columns',
```
joshuaberetta, line 73: Should there also be an `Object.freeze()` here?

jsapp/js/components/modalForms/dataAttachmentColumnsForm.es6
```diff
@@ -0,0 +1,226 @@
...
+                {t('select all')}
+              </a>
+
+              <span>
+                {t('|')}
+              </span>
+
+              <a onClick={this.onBulkDeselect}>
+                {t('deselect all')}
+              </a>
```
joshuaberetta, line 191: Maybe something like this?

```javascript
{this.state.columnsToDisplay.length > 0 && (
  <a onClick={this.onBulkDeselect}>
    {t('deselect all')}
  </a>
)}
```

joshuaberetta: On second thoughts, maybe this should rather behave more like the data table select-all and deselect-all check box?

Screen Shot 2021-04-07 at 9 42 19 AM

jsapp/js/components/modalForms/connectProjects.es6
```diff
@@ -186,163 +298,293 @@ class ConnectProjects extends React.Component {
...
-              classNamePrefix='kobo-select'
+        <div className={selectClassNames.join(' ')}>
+          <Select
+            placeholder={t('Select a different project to import data from')}
+            options={sharingEnabledAssets}
+            value={this.state.newParent}
+            isLoading={(!this.state.isInitialised || this.state.isLoading)}
+            getOptionLabel={option => option.name}
+            getOptionValue={option => option.url}
+            noOptionsMessage={() => {return t('No projects to connect')}}
```
joshuaberetta, line 400: Maybe rather this?
```javascript
noOptionsMessage={() => t('No projects to connect')}
```

jsapp/js/components/modalForms/connectProjects.es6
```diff
@@ -186,163 +298,293 @@ class ConnectProjects extends React.Component {
...
+
+          {!this.state.isLoading && this.state.attachedParents.length == 0 &&
+            <li className='connect-projects__no-imports'>
+              {t('No data imported')}
+            </li>
+          }
+
+          {!this.state.isLoading && this.state.attachedParents.length > 0 &&
+            this.state.attachedParents.map((item, n) => {
+              return (
```
joshuaberetta, line 505: More personal preference, but maybe this?
```javascript
...
const {filename, parentName, attachmentUrl, etc., etc., etc.} = item
return (
...
)
```

joshuaberetta (changes requested):

jsapp/js/components/modalForms/dataAttachmentColumnsForm.es6
```diff
@@ -0,0 +1,226 @@
...
+      ),
+      actions.dataShare.patchParent.completed.listen(
+        this.onPatchParentCompleted
+      ),
+      actions.dataShare.patchParent.failed.listen(
+        this.onPatchParentFailed
+      ),
+      actions.resources.loadAsset.completed.listen(
+        this.onLoadAssetContentCompleted
+      ),
```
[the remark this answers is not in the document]
duvld: Good catch :) if it failed then the loading would stay forever--doesn't make it clear enough that a request failed. I thought a notification was already being handled in `actions.es6`, but that's not true. Interestingly, nowhere else in the JS code (after a quick search) listens for `actions.resources.loadAsset.failed`, probably because a blank screen is a good enough indicator that something messed up :P

jsapp/js/components/modalForms/dataAttachmentColumnsForm.es6
```diff
@@ -0,0 +1,226 @@
...
+    );
+    this.setModalTitle();
+  }
+
+  componentWillUnmount() {
+    this.unlisteners.forEach((clb) => {clb();});
```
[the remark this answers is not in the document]
duvld: Sure, looks cleaner :+1:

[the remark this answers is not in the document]
duvld: Actually everywhere else does it like this, maybe a question for @magicznyleszek :D

jsapp/js/components/modalForms/dataAttachmentColumnsForm.es6
```diff
@@ -0,0 +1,226 @@
...
+  }
+
+  setModalTitle() {
+    this.props.onSetModalTitle(
+      t('Import data from ##PARENT_NAME##')
+        .replace('##PARENT_NAME##', this.props.parent.name)
+    );
```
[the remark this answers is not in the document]
duvld: It's so that the strings marked for translation (with `t(...)`) make it obvious what is being handled by the code and make translators lives a bit easier, see:
https://github.com/kobotoolbox/kpi/pull/3035#discussion_r589742839

jsapp/js/components/modalForms/dataAttachmentColumnsForm.es6
```diff
@@ -0,0 +1,226 @@
...
+  }
+  onBulkDeselect() {
+    let newList = this.state.columnsToDisplay.map((item) => {
+      return {label: item.label, checked: false}
+    });
+    this.setState({columnsToDisplay: newList})
+  }
+  onLoadAssetContentCompleted(response) {
+    if (
+      response.data_sharing?.fields.length > 0
```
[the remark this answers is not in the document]
duvld: Good catch, to be honest I don't remember anymore--doesn't hurt to be more safe, though. I'll add the existence check here too :+1:

jsapp/js/components/modalForms/dataAttachmentColumnsForm.es6
```diff
@@ -0,0 +1,226 @@
...
+        .replace('##PARENT_NAME##', this.props.parent.name)
+    );
+  }
+
+  onAttachToParentCompleted() {
+    this.props.onModalClose();
+  }
+  onAttachToParentFailed() {
+    this.setState({isLoading: false});
+  }
```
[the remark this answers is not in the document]
duvld: It's definitely better practice to tell the user something in the chance this goes wrong. I'll add error alerts to each endpoint call :+1:

jsapp/js/components/modalForms/dataAttachmentColumnsForm.es6
```diff
@@ -0,0 +1,226 @@
...
+  setModalTitle() {
+    this.props.onSetModalTitle(
+      t('Import data from ##PARENT_NAME##')
+        .replace('##PARENT_NAME##', this.props.parent.name)
+    );
+  }
+
+  onAttachToParentCompleted() {
+    this.props.onModalClose();
+  }
```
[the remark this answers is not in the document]
duvld: I think the modal's state would be trashed by React when it's closed. The loading on the parent modal is handled on the parent through a listener that listens also listens to `attachToParent.completed`, which calls `refreshAttachmentList`, which then get's all attached assets again and in turn sets `isLoading` to `false` :D

jsapp/js/components/modalForms/dataAttachmentColumnsForm.es6
```diff
@@ -0,0 +1,226 @@
...
+
+  onColumnSelected(newList) {
+    this.setState({columnsToDisplay: newList});
+  }
+
+  onSubmit(evt) {
+    evt.preventDefault();
+    this.setState({isLoading: true});
+    this.props.triggerParentLoading();
```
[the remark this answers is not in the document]
duvld: This function triggers loading on the parent modal, which listens to both actions that `onSubmit` can call--listeners are in place on the parent to stop loading if either of them fail

jsapp/js/components/modalForms/dataAttachmentColumnsForm.es6
```diff
@@ -0,0 +1,226 @@
...
+
+    let fields = [];
+    var data = '';
```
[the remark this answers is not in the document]
duvld: Sure :) I like that better. tbh I left `var` in and forgot about it when I copied the structure from another older file

jsapp/js/components/modalForms/dataAttachmentColumnsForm.es6
duvld: I try to stick to 80 and I wanted to do it like the way you're describing, but I'm not entirely sure. We want to set up `prettier.io` integration: https://github.com/kobotoolbox/kpi/issues/2983 and I want to follow their [rationale](https://prettier.io/docs/en/rationale.html#print-width) to prepare for this. In it they say 80 is a "guideline" and their formatter doesn't shorten long strings. But... clearly it's [a point of contention](https://github.com/prettier/prettier/issues/4123) so I'm not sure :sweat_smile: It's something I want to ask @magicznyleszek about though!

jsapp/js/components/modalForms/dataAttachmentColumnsForm.es6
duvld: I copied the text you see from the mock ups, maybe @Ig-Rebollo can comment?

jsapp/js/constants.es6
duvld: cc @magicznyleszek :D

jsapp/js/components/modalForms/dataAttachmentColumnsForm.es6
duvld: My implementation is from the mockups, but Leszek made the data table checkbox a component so I can integrate it in for sure. What do you think @Ig-Rebollo

jsapp/js/components/modalForms/connectProjects.es6
```diff
@@ -25,67 +34,128 @@ class ConnectProjects extends React.Component {
...
+          this.props.asset.content.survey,
+        ),
+      });
+    }
+
     actions.dataShare.getSharingEnabledAssets();
   }
 
+  componentWillUnmount() {
+    this.unlisteners.forEach((clb) => {clb();});
```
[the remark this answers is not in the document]
duvld: will go by the response to that comment :+1:

jsapp/js/components/modalForms/connectProjects.es6
```diff
@@ -186,163 +298,293 @@ class ConnectProjects extends React.Component {
...
+          <h2>{t('Share data with other project forms')}</h2>
+        </bem.FormView__cell>
+
+        <bem.FormView__cell m={['box', 'padding']}>
+          <bem.FormView__form>
+            <span>
+              {t('Enable data sharing to allow other forms to import and use dynamic data from this project. Learn more about dynamic data attachments')}
+              &nbsp;
+              <a href='#'>{t('here')}</a>
+            </span>
```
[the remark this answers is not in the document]
duvld: The mockups had two because I think we want two separate help pages for the two different actions a user can take here, please correct me if I'm wrong @Ig-Rebollo !

jsapp/js/components/modalForms/connectProjects.es6
```diff
@@ -186,163 +298,293 @@ class ConnectProjects extends React.Component {
...
+        </bem.FormView__cell>
 
-          {this.renderSelect()}
+        <bem.FormView__cell m={['box', 'padding']}>
+          <bem.FormView__form>
+            <span>
+              {t('Connect with other project(s) to import dynamic data from them into this project. Learn more about dynamic data attachments')}
+              &nbsp;
+              <a href='#'>{t('here')}</a>
+            </span>
```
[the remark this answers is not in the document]
duvld: same as above

jsapp/scss/main.scss
```diff
@@ -20,6 +20,7 @@
 @import "./components/kobo.text-button";
 @import "./components/kobo.checkbox-and-radio";
 @import "./components/kobo.toggle-switch";
+@import "./components/kobo.multi-checkbox";
 @import "./components/kobo.modal";
 @import "./components/kobo.tooltips";
 @import "./components/kobo.navigation";
```
duvld, line 26: Notes to self from internal discussion:

- Disallow self attachment
- De-load the modal

jsapp/js/components/modalForms/connectProjects.es6
```diff
@@ -94,66 +164,70 @@ class ConnectProjects extends React.Component {
...
-  removeAttachment(newVal) {
+
+  onRemoveAttachment(newVal) {
     this.setState({isLoading: true})
     actions.dataShare.detachParent(newVal);
   }
-  toggleSharingData() {
+
+  onToggleSharingData() {
     var data = JSON.stringify({
```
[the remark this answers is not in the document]
duvld: I think we should keep it consistent with `let` instead https://github.com/kobotoolbox/kpi/pull/3117#discussion_r608100860

jsapp/js/components/modalForms/connectProjects.es6
```diff
@@ -177,6 +251,44 @@ class ConnectProjects extends React.Component {
...
+  onColumnSelected(columnList) {
+    this.setState({isLoading: true});
+
+    let fields = [];
+    columnList.forEach((item) => {
+      if (item.checked) {
+        fields.push(item.label);
+      }
+    })
+    var data = JSON.stringify({
```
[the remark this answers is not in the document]
duvld: I think we should keep it consistent with `let` instead https://github.com/kobotoolbox/kpi/pull/3117#discussion_r608100860

jsapp/js/components/modalForms/connectProjects.es6
```diff
@@ -186,163 +298,293 @@ class ConnectProjects extends React.Component {
   generateAutoname(newParent) {
     if (newParent) {
       let autoname = newParent.name;
-      autoname = autoname.toLowerCase().substring(0, MAX_DISPLAYED_STRING_LENGTH).replace(/(\ |\.)/g, '_');
+      autoname = autoname
+        .toLowerCase()
+        .substring(0, MAX_DISPLAYED_STRING_LENGTH.connect_projects)
+        .replace(/(\ |\.)/g, '_');
```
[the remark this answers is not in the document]
duvld: Nice :D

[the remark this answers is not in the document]
joshuaberetta: @duvld maybe even pull this out into it's own method or into the utils? Seems like something that would be used in other places too.

jsapp/js/components/modalForms/connectProjects.es6
```diff
@@ -94,66 +164,70 @@ class ConnectProjects extends React.Component {
...
+      );
     } else {
-      if (!parentUrl) {
+      if (!this.state.newParent?.url) {
         this.setState({
           fieldsErrors: Object.assign(
-            this.state.fieldsErrors, {emptyParent: t('No project selected')}
+            this.state.fieldsErrors, {parent: t('No project selected')}
           )
         });
```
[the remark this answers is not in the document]
duvld: Nice :) slipped my mind as to how to do this without mutating the object directly. Good catch!

jsapp/js/components/modalForms/connectProjects.es6
duvld: @joshuaberetta sorry if I'm missing something, but wouldn't I still have to iterate through `this.state.attachedParents` with some `item` (current value) to get its `filename, parentName, attachmentUrl, etc., etc., etc.`?

jsapp/js/components/modalForms/connectProjects.es6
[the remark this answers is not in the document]
duvld: Good idea, I was kind of on the fence because it's so simple but we wouldn't want duplicate regex's all over the place :+1:

Ig-Rebollo:
Sorry for the delay @duvld and @joshuaberetta 

Here are some mockups on how to implement these suggestions. 

image

As you can see I included a counter to give users confirmation of **when** the questions have been actually shared (if they click on the 'share' button without selecting any questions, maybe a notification saying 'no questions have been selected' could appear.

The 'loading' could be part of the button itself

image

image

Let me know what you think!

jsapp/js/components/modalForms/dataAttachmentColumnsForm.es6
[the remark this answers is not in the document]
magicznyleszek: Yeah let's do it like that :+1: I have personal issue with that due to too much coffeescripting :P

jsapp/js/components/modalForms/dataAttachmentColumnsForm.es6
magicznyleszek: I think we should leave them long. There is no great way to fit them in 80 characters - splitting it with `+` will break getting strings out from `t(` or make it complex to handle.

jsapp/js/constants.es6
magicznyleszek: Yeah let's add it :) Now I freeze every `const` object I define :)

jsapp/js/components/modalForms/dataAttachmentColumnsForm.es6
magicznyleszek: @duvld I don't know if you got an answer from Ignacio - if no, try asking him on Zulip :)

jsapp/scss/components/_kobo.form-view.scss
```diff
@@ -880,6 +880,68 @@
   }
 }
 
+.form-modal__form.form-modal__form--data-attachment-columns {
+   color: darken($kobo-lightgray, 40%);
```
magicznyleszek, line 884: I know that Ignacio will be adding quite some more grays to our palette. I wonder what should we do with that fact. This particular color is `hsl(230, 12%, 50%)`, which seems like best way would be to rename our grays to `$kobo-gray-50`. Let's add a new color like this and I will make a GH issue for renaming current ones

magicznyleszek: Here it is https://github.com/kobotoolbox/kpi/issues/3235

jsapp/js/utils.es6
```diff
@@ -635,3 +635,18 @@ export function truncateFile(str, length) {
...
+
+/**
+ * Generates a simple lowercase, underscored version of a string. Useful for
+ * quick filename generation
+ *
+ * @param {string} str
+ * @param {number} [startIndex=0]
+ * @param {number} [endIndex=str.length]
+ */
+export function generateAutoname(str, startIndex=0, endIndex=str.length) {
```
magicznyleszek, line 647: Just being curious - does this match how `$autoname`s are generated? Maybe we could leave an url to backend code that makes that?

magicznyleszek (changes requested):

jsapp/js/components/modalForms/dataAttachmentColumnsForm.es6
duvld: There's a few changes Ignacio wanted to make about the look and feel in general, but we decided to do that after an initial release :D

duvld: There's a few changes Ignacio wanted to make about the look and feel in general, but we decided to do that after an initial release :D

jsapp/js/components/modalForms/dataAttachmentColumnsForm.es6
```diff
@@ -0,0 +1,226 @@
...
+            {t('You are about to import ##PARENT_NAME##. Select or deselect in the list below to narrow down the number of questions to import.').replace('##PARENT_NAME##', this.props.parent.name)}
+          </span>
+
+          <div className='bulk-options'>
+            <span className='bulk-options__description'>
+              {t('Select below the questions you want to import')}
+            </span>
+
+            <div className='bulk-options__buttons'>
+              <a onClick={this.onBulkSelect}>
```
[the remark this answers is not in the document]
duvld: Got rid of the `i` "buttons" on the imported list as well, actually it was a net loss in CSS now :D

jsapp/js/utils.es6
duvld: I added a link to how the backend generates autoname for translations as that's the closest to "just a string" I could find in the file

magicznyleszek (approved):

PR #3129, opened 2021-04-13
Fix not redirecting to form list after deleting project

duvld:
PR #2984 left an untouched guard that still looked for `asset`, moved
creation of `asset` from `assetOrUid` before this check
## Description
Fixed bug where if user deleted a project they would not be redirected back to the form list

### DIFF
--- a/jsapp/js/searches.es6
+++ b/jsapp/js/searches.es6
@@ -119,21 +119,24 @@ function SearchContext(opts={}) {
     },
     // remove asset from all search store lists
     removeAsset(assetOrUid, isNonOwner) {
+      let asset;
+      if (typeof assetOrUid === 'object') {
+        asset = assetOrUid;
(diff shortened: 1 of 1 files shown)


jsapp/js/searches.es6
```diff
@@ -119,21 +119,24 @@ function SearchContext(opts={}) {
...
-        }
-
-        // non-owner self permission removal only gives an assetUid string, not
-        // an object; for consistency we make it an object here
-        if (!asset) {
-         asset = {uid: assetOrUid};
-        }
+      if (
+        isNonOwner ||
+        this.state.defaultQueryFilterParams?.assetType.includes(asset.assetType)
```
magicznyleszek, line 138: Do you think we need the fake asset object to also have the `assetType` so this will go through?

PR #3144, opened 2021-04-22
Fix missing help box in formbuilder

### LINKED ISSUE #3138: formbuilder no longer shows the helper box when drag-and-dropping questions
## Description

The formbuilder no longer shows the helper box when drag-and-dropping questions. Before, there was a grey box with a dashed outline to show where the question would land while dragging it. Now it's being removed while dragging with no indication to the user on where it would land when releasing.

## Steps to Reproduce

1. Drag a question in a multi question form in the formbuilder

## Expected behavior 

Outline of question being dragged is wherever the cursor is currently placed.

## Actual behavior 

The moment a question is being dragged its spot disappears from the formbuilder, and moving the cursor up or down doesn't show graphically where the question would be dragged.

## Additional details

Seems to be a recent regression.

duvld:
Small hack fix in comments. We have two different `placeholders` - one for moving questions within the form around and one for dragging questions from the library. It's a bit [hard to notice](https://github.com/kobotoolbox/kpi/blob/7edbc13e5da9982deff701491ee335f89b84f449/jsapp/xlform/src/view.surveyApp.coffee#L466-L527) that the former exists, so the CSS for it was removed in [2774](https://github.com/kobotoolbox/kpi/commit/33cb5c3f75d5c3f3229f89586ba6942bf4dda69b#diff-fd1cc506eb0ff4295748f37cb38f3c37f5f6e473dc67594ddf7a92dda1fd9e69L355-L374).

Since they use different CSS classnames, to avoid giving library specific class names to non-library components I just add the missing CSS to the non-library classname. For the formbuilder revamp we should generalise the placeholders

## Description
The formbuilder used to have a help box displaying where a dragged question will land. Regressed by colour rework PR, fixed here. Also changed colour and thickness

## Related Issues
Fixes #3138

### DIFF
--- a/jsapp/scss/components/_kobo.form-builder.scss
+++ b/jsapp/scss/components/_kobo.form-builder.scss
@@ -467,9 +467,13 @@ $t-form-builder-aside-open: 200ms;
   }
 
   // Semi-hack: lib-list__item already have margins, non-library placeholder
-  // doesn't--for formbuilder revamp all placeholders should be the same
+  // doesn't. lib-list__item should also be transparent--for formbuilder revamp
+  // all placeholders should be the same
   .placeholder {
     margin: 5px 0;
+  }
+  .lib-list__item {
+    background: transparent;
   }
 }
 
--- a/jsapp/scss/components/_kobo.form-builder.scss
+++ b/jsapp/scss/components/_kobo.form-builder.scss
@@ -459,11 +459,17 @@ $t-form-builder-aside-open: 200ms;
 }
 
 .survey-editor {
-  .ui-sortable-placeholder.lib-list__item {
-    border: 1px dashed $kobo-midgray;
+  .ui-sortable-placeholder {
+    border: 2px dashed $kobo-blue;
     visibility: visible !important;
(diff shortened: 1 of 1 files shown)


magicznyleszek (changes requested):
I found two things:

1. When you drag library item into a group - placeholder box has white background (it should be transparent)
2. A bug that is not related but maybe you want to fix it :D If no, I will create a GH issue:
  - when you move a group out of other group the group ends up as empty (it should disappear)

duvld:
> I found two things:
> 
> 1. When you drag library item into a group - placeholder box has white background (it should be transparent)
> 2. A bug that is not related but maybe you want to fix it :D If no, I will create a GH issue:
> 
> * when you move a group out of other group the group ends up as empty (it should disappear)

I think it's best suited for a new issue, as this also happens when you remove a question from a group and that group is now empty. If there are multiple groups, and the only question is in all of them are removed/moved out, should it also remove each parent group? Maybe it's relevant enough to be tackled in #3146 (seems like we can treat last question moving out the same as last question removed)?

PR #3152, opened 2021-04-27
Add non-owner self removal to form landing

### LINKED ISSUE #3151: Add non-owner self removal to form landing
## Description

Alongside or in place of (depending on permissions) to the "Share this project" option in the form landing:

image

And fix surrounding option titles to fit sentence casing styling

duvld:
Currently redirects from form landing after self removal but does not trigger refresh of asset list (the removed asset still shows unless the user refreshes)

## Description

Add the non-owner self removal option introduced in https://github.com/kobotoolbox/kpi/pull/2984 to the actions available in form landing (to compliment where a user would expect the sharing options to be for an owned form):
image
## Related issues
Fixes #3151


jsapp/js/components/formLanding.es6
```diff
@@ -388,6 +395,19 @@ export class FormLanding extends React.Component {
...
+    hashHistory.push(ROUTES.FORMS);
+  }
+  nonOwnerSelfRemoval(evt) {
+    evt.preventDefault();
+    // Listen for permission removal here to avoid manage_asset user removal
+    // from triggering redirect
+    this.listenTo(
+      actions.permissions.removeAssetPermission.completed,
+      this.goToProjectsList
+    );
```
magicznyleszek, line 408: I am wondering if you shouldn't remove the listener afterwards. So there is no bug after some other remove permission call :)

So:

```
nonOwnerSelfRemoval(evt) {
  evt.preventDefault();
  this.cancelNonOwnerSelfRemovalListener = this.listenTo(
    actions.permissions.removeAssetPermission.completed,
    this.nonOwnerSelfRemovalCompleted
  );
}
nonOwnerSelfRemovalCompleted() {
  if (this.cancelNonOwnerSelfRemovalListener) {
    this.cancelNonOwnerSelfRemovalListener()
  }
  this.goToProjectsList();
}
```

jsapp/js/mixins.es6
```diff
@@ -242,7 +242,11 @@ mixins.dmix = {
     } else if (uid) {
       actions.resources.loadAsset({id: uid});
     }
-  }
+  },
+  removeSharing: function(evt) {
+    evt.preventDefault();
+    mixins.clickAssets.click.asset.removeSharing(this.props.params.uid);
+  },
```
magicznyleszek, line 249: I don't think you need to pass `evt` to this function, let's make it only about removing sharing. `evt` is already `preventDefault`-ed :)

magicznyleszek (changes requested):

magicznyleszek:
One more thing I noticed - on the Projects List route the button is in the middle of list, on the Form Landing it's on the bottom

Screenshot 2021-05-03 at 16 23 48

Screenshot 2021-05-03 at 16 27 26

But now I also see that all options are in different order.

I propose: make this self removal option appear as last item and I would make a new issue to cleanup those.

jsapp/js/components/formLanding.es6
duvld: I know it's not meant to be exact, but just to make sure can you please check my changes again :D

Issue #3148, opened 2021-04-23
Review permissions for duplicate submission feature


noliveleger:
## Description

Currently, users need `Edit submissions` to be able to duplicate a submission. When writing the unit tests for row level permissions I wondered whether it would be better to assign `Add submissions` because duplicating a submission is like adding a new submission. 

But @jnm said:
> One wrinkle is that the use case for duplicating submissions assumes the duplicate will usually be edited immediately after creation

That means the owner/ the managers should assign both permissions. It would be make sense, but as @tinok said
> from a UX side I think it's easier to assign one permission instead of two.

If we decide to use both, we would need to document it. Front-end code should be updated too.

related to #2768

tinok:
@noliveleger I would say let's leave it with just `edit submissions` as you have it. Logically someone with that permission is able to do drastic changes to the collected data, so duplicating a row is not something I can imagine someone would want to restrict a collaborator from doing that.

noliveleger:
As per our last discussion: Users will need both permissions:

- `add_submissions` to create the new duplicated submission 
- `change_submissions` to edit the new duplicated submission

PR #3157, opened 2021-04-27
Fix kobocat proxy request user

### LINKED ISSUE #3153: Incorrect user in kobocat proxy request for bulk edit
Currently:
https://github.com/kobotoolbox/kpi/blob/bf4ccc6fbe22d3e01417170ed1286812acdf5874/kpi/deployment_backends/kobocat_backend.py#L713-L715
Should use requesting user, not asset owner in request.

joshuaberetta:
closes #3153

### DIFF
--- a/kpi/deployment_backends/kobocat_backend.py
+++ b/kpi/deployment_backends/kobocat_backend.py
@@ -664,7 +664,7 @@ class KobocatDeploymentBackend(BaseDeploymentBackend):
         return dt.isoformat('T', 'milliseconds')
 
     def duplicate_submission(
-        self, requesting_user_id: int, instance_id: int
+        self, requesting_user: 'auth.User', instance_id: int
     ) -> dict:
         """
         Dupicates a single submission proxied through kobocat. The submission
@@ -673,7 +673,7 @@ class KobocatDeploymentBackend(BaseDeploymentBackend):
         to kobocat.
 
         Args:
-            requesting_user_id (int)
+            requesting_user ('auth.User')
             instance_id (int)
 
         Returns:
@@ -681,7 +681,7 @@ class KobocatDeploymentBackend(BaseDeploymentBackend):
             if successful
         """
         params = self.validate_submission_list_params(
-            requesting_user_id,
+            requesting_user.id,
             format_type=INSTANCE_FORMAT_TYPE_XML,
             instance_ids=[instance_id],
         )
@@ -711,12 +711,12 @@ class KobocatDeploymentBackend(BaseDeploymentBackend):
             method='POST', url=self.submission_url, files=files
         )
         kc_response = self.__kobocat_proxy_request(
-            kc_request, user=self.asset.owner
+            kc_request, user=requesting_user
         )
 
         if kc_response.status_code == status.HTTP_201_CREATED:
             return next(
-                self.get_submissions(requesting_user_id, query={'_uuid': _uuid})
+                self.get_submissions(requesting_user.id, query={'_uuid': _uuid})
             )
         else:
             raise KobocatDuplicateSubmissionException
@@ -775,7 +775,7 @@ class KobocatDeploymentBackend(BaseDeploymentBackend):
         return self.__prepare_as_drf_response_signature(kc_response)
 
     def bulk_update_submissions(
-        self, request_data: dict, requesting_user_id: int
+        self, request_data: dict, requesting_user: 'auth.User'
     ) -> dict:
         """
         Allows for bulk updating of submissions proxied through kobocat. A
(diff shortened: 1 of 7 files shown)


kpi/deployment_backends/kobocat_backend.py
```diff
@@ -865,7 +865,7 @@ def bulk_update_submissions(
                 method='POST', url=self.submission_url, files=files
             )
             kc_response = self.__kobocat_proxy_request(
-                kc_request, user=self.asset.owner
+                kc_request, user=requesting_user
```
noliveleger, line 868: I think I found why you chose to use `self.asset_owner` at the first place. 
`requesting_user` needs `add_submissions` to be allowed to duplicate in KC, otherwise it returns a 403. 
Sending the request with the owner bypasses this validation. 

This brings up this issue: https://github.com/kobotoolbox/kpi/issues/3148

Would you mind to change the tests?

And if you are willing, would you mind to change the front end too? 
i.e. Show the "duplicate" button only to users who have both permissions

Otherwise, push it to one of our front-end friends.

While looking at the front-end code, we should fix this typo. 
Screen Shot 2021-05-07 at 11 14 24

kpi/deployment_backends/kobocat_backend.py
```diff
@@ -664,7 +664,7 @@ def format_openrosa_datetime(dt: Optional[datetime] = None) -> str:
         return dt.isoformat('T', 'milliseconds')
 
     def duplicate_submission(
-        self, requesting_user_id: int, instance_id: int
+        self, requesting_user: 'auth.User', instance_id: int
```
noliveleger, line 667: I cannot choose line 670, but there is a typo 
`Dupicates` instead of `duplicates`

noliveleger (changes requested):

kpi/deployment_backends/kobocat_backend.py
joshuaberetta: - [ ] change back to `self.asset.owner` (also for bulk edit)
- [ ] update tests
- [ ] change permissions to required both `add_submissions` and `change_submissions` otherwise `403`
- [ ] update frontend to only show `DUPLICATE` button when user has both permissions
- [ ] fix typos

kpi/permissions.py
```diff
@@ -260,12 +260,13 @@ class EditSubmissionPermission(SubmissionPermission):
 
 
 class DuplicateSubmissionPermission(SubmissionPermission):
+    # POST permissions are handled in the viewset to account for needing both
+    # `change_submissions` and `add_submissions` permissions
     perms_map = {
         'GET': ['%(app_label)s.view_%(model_name)s'],
-        'POST': ['%(app_label)s.change_%(model_name)s'],
+        'POST': ['%(app_label)s.view_%(model_name)s'],
```
noliveleger, line 267: I think the permissions should:

```python
'POST': [
    '%(app_label)s.change_%(model_name)s',
    '%(app_label)s.add_%(model_name)s',
]
```

kpi/deployment_backends/kobocat_backend.py
```diff
@@ -864,8 +864,9 @@ def bulk_update_submissions(
             kc_request = requests.Request(
                 method='POST', url=self.submission_url, files=files
             )
+            # `asset.owner` permissions needed for kobocat to allow the request
             kc_response = self.__kobocat_proxy_request(
-                kc_request, user=requesting_user
+                kc_request, user=self.asset.owner
```
noliveleger, line 869: Why did you revert it? We still want `requesting_user` to make the request, right?

kpi/views/v2/data.py
```diff
@@ -365,6 +370,18 @@ def duplicate(self, request, pk, *args, **kwargs):
...
+        if not all(
+            perm in self.asset.get_perms(request.user)
+            for perm in [PERM_ADD_SUBMISSIONS, PERM_CHANGE_SUBMISSIONS]
+        ):
+            raise exceptions.PermissionDenied(
+                detail=_(
+                    'User needs both `add_submissions` and '
+                    '`change_submissions` permissions to duplicate a submission.'
+                )
+            )
```
noliveleger, line 384: Not needed, `DuplicateSubmissionPermission` should do the job when the correct permissions are set in `perms_map`

noliveleger (changes requested):

kpi/deployment_backends/kobocat_backend.py
noliveleger: > change back to self.asset.owner (also for bulk edit)
No, we want to keep `requesting_user`

noliveleger:
@joshuaberetta, I needed your fix ASAP do I decided to apply the changes myself to merge it. 
I left the comment as a reference.

kpi/deployment_backends/kobocat_backend.py
joshuaberetta: @noliveleger I needed to revert it for bulk editing. I shouldn't have changed it to begin with.

kpi/deployment_backends/kobocat_backend.py
joshuaberetta: You're right 🤦 I corrected that point with what I actually did:

> change back to self.asset.owner for bulk edit

PR #3174, opened 2021-04-30
Allow users to migrate their XForms from kobocat

### LINKED ISSUE #3150: Self-service sync_kobocat_xforms
Extend #3002 so that people can synchronize their own forms. We used to offer this in a hacky way:

https://github.com/kobotoolbox/kpi/commit/b8351171428bd7a048c6334c68b0064167cf7845#diff-1483cb549cd31083151fa35593cfd9cd1d460bcca38c32a0049ac004d4459f1eL20

We need back-end support for this as well as some minimal UI component to invoke it. I'm unsure whether the UI component will be in KPI or KoBoCAT, but probably KoBoCAT given that we'll be [showing a notice](https://github.com/kobotoolbox/kobocat/issues/700) there?

joshuaberetta:
## Description

In addition to superusers being able to sync any user's forms at the `/api/v2/users/{username}/migrate` endpoint, allow users to migrate their own XForms from kobocat.

## Related issues

closes #3150 
dependency of kobotoolbox/kobocat#706

### DIFF
--- a/kpi/views/v2/user.py
+++ b/kpi/views/v2/user.py
@@ -32,8 +32,10 @@ class UserViewSet(viewsets.GenericViewSet, mixins.RetrieveModelMixin):
             url_path=r'migrate(?:/(?P<task_id>[\d\w\-]+))?')
     def migrate(self, request, task_id: str = None, **kwargs):
         """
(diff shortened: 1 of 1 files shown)


jnm (approved):
I love this.

PR #3177, opened 2021-05-04
Library upload asset type selector

### LINKED ISSUE #3142: Improve library file import (allow choosing asset type)
## Description

Currently all uploaded `xls` files in My Library end up as `block` type. We need to have a nice way to upload a template (a general improvement, but mainly for #2991).

Idea for fix:

1. In the "NEW > Upload" modal add an input to choose uploaded type (template or block)
  - optionally add also an input to set title (prefilled with the xls name)
2. When uploading one file, display the "NEW > Upload" modal with file uploaded
3. When uploading multiple files don't show the modal (leave as is now)

magicznyleszek:
## Description

Whenever user uploads one file (through NEW modal or by dropping on list) in My Library, it is possible to select the desired type, e.g. "template".

## Related issues

Fixes #3142


duvld:
Without backend branch `3005-library-locking-feature-combined` merged in the backend would cause `kobo--locking-profiles` to be appended to an asset's `content`. The fix is merged in here: https://github.com/kobotoolbox/kpi/pull/3177/commits/95c98bec13d6a1cad39b2261be1b57656c1f5824

magicznyleszek:
I'm merging this as it was merged into `beta` already through other branch and there are zero changes ;)

PR #3188, opened 2021-05-07
Fix choice question labels mismatch

### LINKED ISSUE #3183: Report shows same label multiple times
## Description

https://chat.kobotoolbox.org/#narrow/stream/6-Support/topic/Issue.20with.20custom.20report/near/29068

1. Use the particular form: [Label-issues.xlsx](https://github.com/kobotoolbox/kpi/files/6430653/Label-issues.xlsx)
2. Have some data submitted
3. Go to `/forms/<uid>/data/report`
4. Observe duplicated labels:
  Screenshot 2021-05-05 at 23 40 30

It seems that the problem with only labels being wrongly assigned. Please check `reports.es6` creating `reportData[i].data.responseLabels = [];`

magicznyleszek:
## Description

Fix wrong labels being displayed when some choices have same labels as some choices names.

## Additional details

This was happening because the old reports code was expecting responses as `name`s, but instead it was getting `label[0]`s.

This was working 99% of time, because it never found a choice to display and as a fallback was displaying the response it got from backend.

This PR contains some code cleanup changes, as I needed to understand what's happening. The real fix is a one liner in `choice = asset.content.choices.find((item) => {` function near `NOTE` comment - it checks for label instead of name to find a choice. 

This requires #3185 to be merged first.

## Related issues

Fixes #3183


magicznyleszek:
@duvld please do not merge or review it right now

magicznyleszek:
Fixed already

PR #3209, opened 2021-05-12
[Library locking] Add locking to translation modal

### LINKED ISSUE #3205: [Library Locking] All `…label_edit` restrictions should apply to translations modal
Table for editing translations should have some rows disabled.

duvld:
## Description

Adds locking to the translation modal. Specifically, if a question, choice, or group label is locked, it's respective label in every translation is locked in the update translation modal

Known caveats: #3213
## Related issues
Fixes #3205


jsapp/js/components/modalForms/translationTable.es6
```diff
@@ -67,7 +82,18 @@ export class TranslationTable extends React.Component {
...
         minWidth: 130,
-        Cell: (cellInfo) => {return cellInfo.original.original;}
+        Cell: (cellInfo) => {
+          // Disabling has no effect on this cell, but we do it to gray out the
+          // text to indicate that the label is locked
+          // TODO: Figure out what to do for the case of adding a new language
+          //       when there are locked labels. These labels should be unlocked
+          //       for the newly added languages and their translations only.
+          return (
+            <div className={cellInfo.original.isLabelLocked ? 'disabled' : ''}>
```
magicznyleszek, line 92: Maybe make this `rt-td--disabled`, as `disabled` is quite a common name, and could be defined elsewhere (e.g. in some plugin we use and we update)?

magicznyleszek (changes requested):

jsapp/scss/components/_kobo.modal.scss
```diff
@@ -707,6 +711,12 @@ $z-modal-x: 10;
           &:focus {
             background: $kobo-white;
           }
+          // If the sibling spans more lines than its translation, the
+          // .translation background color takes over the rest of the space.
+          // This looks ugly, so we change text color to indicate locking instead
+          &:disabled {
+            color: $kobo-midgray;
+          }
         }
```
magicznyleszek, line 720: I have a proposition:

```patch
diff --git a/jsapp/scss/components/_kobo.modal.scss b/jsapp/scss/components/_kobo.modal.scss
index 90e5039d357ba28185d921b212f1e8ca927f5361..688c04975e920dc2aabce81f6473640564306a5a 100644
--- a/jsapp/scss/components/_kobo.modal.scss
+++ b/jsapp/scss/components/_kobo.modal.scss
@@ -698,16 +698,17 @@ $z-modal-x: 10;
         min-width: 70%;
       }
       .translation {
-        background: $kobo-green-light;
         color: $kobo-darkgray;
-        padding: 2px;
+        padding: 0;
+
         textarea {
+          background: $kobo-green-light;
           display: block;
-          padding: 8px;
+          padding: 10px;
           border: none;
-          background: transparent;
           width: 100%;
           color: $kobo-black;
+          height: 100% !important;
           &:focus {
             background: $kobo-white;
           }
@@ -716,6 +717,7 @@ $z-modal-x: 10;
           // This looks ugly, so we change text color to indicate locking instead
           &:disabled {
             color: $kobo-midgray;
+            background-color: $kobo-whitegray;
           }
         }
       }

```

magicznyleszek: This way we avoid this weird gray on green and take over the visal side. The downside is having a thin white border around green cells, but it's better anyways :)

magicznyleszek (changes requested):

jsapp/scss/components/_kobo.modal.scss
magicznyleszek: Also this comment:
```scss
// If the sibling spans more lines than its translation, the
// .translation background color takes over the rest of the space.
// This looks ugly, so we change text color to indicate locking instead
```
Is no longer needed, as `height: 100% !important;` hack override fixes this

jsapp/js/components/modalForms/translationTable.es6
```diff
@@ -49,14 +66,16 @@ export class TranslationTable extends React.Component {
     // add choice options to translation table
     if (choices && choices.length) {
       choices.forEach((choice) => {
+        let isLabelLocked = this.isLabelLocked(CHOICE_TYPE, choice.name);
```
magicznyleszek, line 69: Instead of introducing this `CHOICE_TYPE` please make two separate functions `isChoiceLabelLocked` and `isRowLabelLocked` :)

jsapp/scss/components/_kobo.modal.scss
```diff
@@ -562,6 +566,18 @@ $z-modal-x: 10;
     &:hover .form-view__cell--translation-name .form-view__icon-button {
       opacity: 1;
       visibility: visible;
+
+      // Hide the on hover button completely if disabled
+      &:disabled {
+        opacity: 0;
+        visibility: hidden;
```
magicznyleszek, line 573: Wouldn't `display: none` be good enough?

magicznyleszek (changes requested):

jsapp/js/components/locking/lockingConstants.es6
```diff
@@ -33,7 +33,7 @@ export const LOCKING_RESTRICTIONS = Object.freeze({
   group_add: {name: 'group_add', label: t('Add group')},
   question_add: {name: 'question_add', label: t('Add question')},
   question_order_edit: {name: 'question_order_edit', label: t('Change question order')},
-  translations_add: {name: 'translations_add', label: t('Manage translations')},
+  language_edit: {name: 'language_edit', label: t('Manage translations')},
```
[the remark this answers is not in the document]
duvld: I like "Change languages" :D

jsapp/js/components/modalForms/translationSettings.es6
```diff
@@ -146,9 +155,9 @@ export class TranslationSettings extends React.Component {
   }
   prepareTranslations(content) {
     let translated = content.translated,
-        translationsLength = content.translations.length,
-        survey = content.survey,
-        choices = content.choices;
+      translationsLength = content.translations.length,
+      survey = content.survey,
+      choices = content.choices;
```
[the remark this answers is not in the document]
duvld: Didn't notice that the vim prettier plugins didn't catch this, i'll double check my linter

jsapp/scss/components/_kobo.modal.scss
duvld: Don't know if you see it on your screen but if I do this the language text that would have the icon next to it moves on hover

jsapp/js/components/modalForms/translationSettings.es6
```diff
@@ -295,51 +333,61 @@ export class TranslationSettings extends React.Component {
...
+                      disabled={
+                        this.state.isUpdatingDefaultLanguage ||
+                        !this.canEditLanguages()
+                      }
                       data-tip={t('Edit language')}
                       className='right-tooltip'
                     >
-                      {this.state.renameLanguageIndex === i &&
+                      {this.state.renameLanguageIndex === i && (
                         <i className='k-icon-close' />
```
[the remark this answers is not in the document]
duvld: made these changes for the relevant files int this PR :+1:

jsapp/js/components/modalForms/translationSettings.es6
```diff
@@ -131,13 +136,17 @@ export class TranslationSettings extends React.Component {
...
+        ),
         labels: {ok: t('Delete'), cancel: t('Cancel')},
         onok: () => {
           this.updateAsset(content);
           dialog.destroy();
         },
-        oncancel: () => {dialog.destroy();},
+        oncancel: () => {
+          dialog.destroy();
+        },
```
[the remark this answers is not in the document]
duvld: Is this a linting option I missed?

jsapp/js/components/modalForms/translationTable.es6
duvld: Okay so choices weren't actually being locked at all :sweat_smile:

This is because (and maybe we should change this) we don't apply the restriction to the choice lists, but rather to the rows that can potentially point to them, i.e., a select question can have the row restriction `choice_label_edit` but the `choice` in the code here doesn't know that.

I remedied it this way: https://github.com/kobotoolbox/kpi/pull/3209/commits/e685f5059372c520111cf64753d1482846375e8c -- I have it set that if a row has `choice_label_edit` locked then the choices in its linked choice list is also locked, even if another row does not have this restriction. Let me know what you think!

jsapp/js/components/modalForms/translationSettings.es6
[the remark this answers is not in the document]
magicznyleszek: Don't think there is linting option that handles that. It's `leszek-linter` :P

jsapp/scss/components/_kobo.modal.scss
magicznyleszek: Oh I see. I would say that it means that the parent styles are not good enough :P Let's do this on `form-view__cell--translation-name`:

```
display: flex;
flex-direction: row;
align-items: center;
```

jsapp/js/components/modalForms/translationTable.es6
magicznyleszek: Seems nice :)

PR #3216, opened 2021-05-13
Fix `AnonymousUser` listed permissions on public asset

### LINKED ISSUE #3215: `AnonymousUser` permissions not listed correctly at `asset-detail` endpoint
## Description

When a project is publicly shared, `AnonymousUser` has `view_asset` permissions. `AnonymousUser` will also get `view_submissions` if submissions are shared publicly. When accessing the `asset-detail` endpoint as `AnonymousUser`, you should be able to see your permissions as well as those of the asset's owner.

## Steps to Reproduce

1. Create a project
2. Share publicly
3. Visit asset detail endpoint

## Expected behavior 

See `AnonymousUser`'s permissions.

## Actual behavior 

Only see asset owner's permissions.

joshuaberetta:
## Related issues

closes #3215

### DIFF
--- a/kpi/utils/object_permission_helper.py
+++ b/kpi/utils/object_permission_helper.py
@@ -32,7 +32,12 @@ class ObjectPermissionHelper:
         # Filtering is done in `get_queryset` instead of FilteredBackend class
         # because it's specific to `ObjectPermission`.
         if not user or user.is_anonymous:
-            queryset = queryset.filter(user_id=affected_object.owner_id)
+            queryset = queryset.filter(
+                user_id__in=[
+                    affected_object.owner_id,
+                    settings.ANONYMOUS_USER_ID,
+                ]
+            )
         elif not affected_object.has_perm(user, PERM_MANAGE_ASSET):
             # Display only users' permissions if they are not allowed to modify
             # others' permissions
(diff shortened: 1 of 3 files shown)


kpi/utils/object_permission_helper.py
```diff
@@ -32,7 +32,12 @@ def get_user_permission_assignments_queryset(cls, affected_object, user):
         # Filtering is done in `get_queryset` instead of FilteredBackend class
         # because it's specific to `ObjectPermission`.
         if not user or user.is_anonymous:
-            queryset = queryset.filter(user_id=affected_object.owner_id)
+            queryset = queryset.filter(
+                user_id__in=[
+                    affected_object.owner_id,
+                    settings.ANONYMOUS_USER_ID,
+                ]
+            )
```
noliveleger, line 40: Great... Because the code is almost the same I would have refactored it a little bit. 

```python3
        # Filtering is done in `get_queryset` instead of FilteredBackend class
        # because it's specific to `ObjectPermission`.
        if not affected_object.has_perm(user, PERM_MANAGE_ASSET):
            filtered_users = [
                affected_object.owner_id,
                settings.ANONYMOUS_USER_ID,
            ]

            # Display only users' permissions if they are not allowed to modify
            # others' permissions
            if not user.is_anonymous:
                filtered_users.append(user.pk)

            queryset = queryset.filter(user_id__in=filtered_users)
        
        return queryset
```
The difference in my code, it does not handle `user is None`. But I looked at the code and `get_user_permission_assignments_queryset` is never called without `user`

noliveleger (approved):

PR #3234, opened 2021-05-24
[Feature] Add support for `background-audio` in the table view

### LINKED ISSUE #3037: Improve how table finds attachment URLs
## Description

Currently https://github.com/kobotoolbox/kpi/pull/2553 is implemented in a such a way that, for each submission that contains media, we traverse every other submission for the media's filename in order to get the attachment URL to display in its table cell.

This should be improved to only search within its own submission to avoid extra work as well as potentially grabbing the wrong attachment URL.

### LINKED ISSUE #3193: Background audio recording functionality
Background recording of a survey is now available in Collect. This issue outlines how this functionality should work in KPI. 

### In the form builder
This feature would record the entire survey, producing a long file with all the questions and answers for a given submission. The background recording is not really a question type, but a form level function. It makes more sense to access this functionality in settings, even if it displays as a question of sorts afterwards in the form builder. See animation below:
backgroundaudio-formbuilder2
backgroundaudio-formbuilder

**Modifications in the form builder**
- [x] Create a section within settings to enable background recording. When activated, display options below (for the time being, the only choice available is audio quality).
- [x] When activated, this function will create a box with the background recording icon and a message indicating that the survey will be recorded.
- [x] The section in settings must include a note highlighting the fact that this feature only works on KoboCollect

### In the Table view
The designs used in this section are those of the new table view UI, which is described in more detail in #3198 and #3199. The addition of the background recording feature to the table view **does not require these updates**, so it could be implemented using the old UI until the updates are possible.
The background audio would be the first of the metadata columns. It would display an audio file icon and the total duration of the recording:
Table view redesign
It would also appear on the single submission view. Ideally, the single submission view could also be revamped as it appears on the mockup. If this is not possible at the moment, the same audio play functionality would be included on top of the data, opposed to Validation status.
Version without any major changes:
singlesub old brec

**Important updates:**
- [x] Background recording should have a dedicated column, and this column should be right after validation status, as a metadata field and not a normal question
- [ ] The column should include file icons that are clickable. On hover they will appear full blue (our buttons blue). On click, they will open a modal (this modal could be styled with new UI or the old one, depending on time available) That would allow users to play the audio.
- [ ] When clicking on the single submission icon, the background recording should available to play, and it should not be treated as a regular question. Either as metadata (in the new UI) or simply on top of the questions/answers table (as in the version with few updates)

For additional details and specs, see [the Figma file here](https://www.figma.com/file/cJhbMFYOsQeHBqVJtV8sXA/Background-recording-feature?node-id=0%3A1)

duvld:
Exists alongside #3276 and contains both formbuilder and table view changes. #3276 exists to review formbuilder code seprately, we should merge that in first

## Description

Adds UI changes outlined in #3193 that implements table view and formbuilder support for `background-audio` recording

## Related issues

Fixes #3193
Fixes #3037


jnm:
@duvld, I'm deleting `2605-pyxform-upgrade`, so please rebase this on the current `beta`. I'm guessing that only commits 2411e38 and f5aefc7 need to be retained.

jsapp/js/components/submissions/table.es6
```diff
@@ -422,10 +422,14 @@ export class DataTable extends React.Component {
         return false;
       }
 
+      // Set ordering of question columns. Meta questions can be prepended or
+      // appended relative to survey questions with an index prefix
+
       // sets location of columns for questions not in current survey version
+      // `y` makes puts this case in front of known meta types
```
magicznyleszek, line 429: This comment probably should be without "makes"

jsapp/js/components/submissions/table.es6
```diff
@@ -461,6 +465,10 @@ export class DataTable extends React.Component {
         case ADDITIONAL_SUBMISSION_PROPS._submitted_by:
             index = 'z92';
             break;
+        // set index for `background-audio` to the front of every column with `_`
+        case 'background-audio':
```
magicznyleszek, line 469: I added `background-audio` in Form Builder, made few submissions, but I don't see this column in the table. Should it appear?

jsapp/js/components/submissions/table.es6
```diff
@@ -461,6 +465,10 @@ export class DataTable extends React.Component {
...
+            index = '_1';
```
magicznyleszek, line 470: Wouldn't `_1` make it appear as first column? Columns with `_` (so `_id`, `_uuid`, etc.) use `z123`, and the other columns use numbers, so underscore would come first

magicznyleszek (changes requested):

jsapp/js/components/submissions/submissionModal.es6
```diff
@@ -349,6 +356,23 @@ class SubmissionModal extends React.Component {
...
+
+                <bem.BackgroundAudioPlayer__audio controls>
+                  <source
+                    src={this.props?.backgroundAudioUrl}
+                    type='audio/x-m4a'
+                  />
+                </bem.BackgroundAudioPlayer__audio>
+              </bem.BackgroundAudioPlayer>
+            </bem.FormModal__group>
+          }
```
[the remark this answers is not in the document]
duvld: The modal is styled with horizontal groups that have flex display, so if there are translations then the translations dropdown would take the place of the background audio player:
image

We could switch the position of the player and the translation, or maybe put the player in the place of the translation if there aren't any?

jsapp/js/components/submissions/table.es6
duvld: Not sure I understand what's wrong :sweat_smile:

jsapp/js/components/submissions/table.es6
duvld: Not sure if you did this but the question will only have submissions if you use ODK/Kobo collect--but I guess the column should be there even if there are no submissions for the question type? This is true for other types (maybe this is what you meant :P)

duvld: Figured out why the column won't show if there's not `background-audio` data, it's because `getSurveyFlatPaths` doesn't include meta be default. Since we're treating `background-audio` like a meta question, it doesn't actually get sent to us with the submission unless it contains data. What do you think we should do? Or should we leave it as is?

jsapp/js/components/submissions/table.es6
duvld: I changed the language in case it was bad wordplay

jsapp/js/components/submissions/submissionModal.es6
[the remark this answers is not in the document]
magicznyleszek: Ah right, the language selector… Let's try to do something not requiring much template logic, and if it's not possible then let's leave it until we build our audio player.

My idea is: let's put all three on the same line with `flex`, give them min width and use `flex-wrap` so it automagically wraps when there is not enough space. Plus each of selectors should also wrap inside and have the label be on top if not enough space (or by default?):

Screenshot 2021-06-18 at 09 17 00

jsapp/js/components/submissions/table.es6
magicznyleszek: Oh right. Let's leave it as is then. It is consistent with other meta questions :+1:

jsapp/js/components/submissions/table.es6
magicznyleszek: I thought it wasn't supposed to be first, sorry :D

jsapp/js/components/submissions/submissionModal.es6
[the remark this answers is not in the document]
duvld: This works for slotting all three components in, but does come with its own caveats, namely the case where there would only be "Validation status":
image

it's no longer right aligned like before making lots of white space. Do you think this is acceptable? or maybe we should just wait for a seprate submission modal rework?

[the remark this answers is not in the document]
magicznyleszek: You could use `:only-child` on the validation status and move it to right with something (perhaps there is some flex property for that, or you could end up using left margin).
If you have both language and validation status, you can use `:first-child` check on language to move it too.
But if this ends up not working in all scenarios, I think you can revert to having background audio in it's own row and we would fix that later :D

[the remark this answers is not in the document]
duvld: I added a wrapper class for the dropdowns (they needed it anyway for the JSX conditionals) and gave it a `only-child` that makes it `width: 100%` and it looks like the best compromise:

- bg audio with both options
image

- bg audio with just validation
image

- no bg audio both options
image

- no bg audio just validation
image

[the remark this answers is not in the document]
magicznyleszek: Looks fine now :+1: :) We will improve upon it when we build the audio component :)

Issue #3266, opened 2021-06-09
Define a wrapper component for react-select


magicznyleszek:
## Description

Currently we use `react-select` (`<Select …/>`) with very similar options for all instances. With new design introduced at #3199 we will use different indicator icon which requires an override custom partial component. It would be best to define our own `KoboSelect` component (in the `jsapp/js/components/common` directory) that will be a wrapper around `react-select` with all the common options included.

## Additional details

Remember about #3248 while working on this.

magicznyleszek:
## Research

Screenshots of most of the usages of `ui.PopoverMenu`:

Projects list more actions menu:
<img width="264" alt="Screenshot 2021-06-10 at 13 27 18" src="https://user-images.githubusercontent.com/2521888/121517428-43f9ba00-c9e7-11eb-82a8-fe0343739b68.png">

Form landing collection menu:
<img width="660" alt="Screenshot 2021-06-10 at 13 28 29" src="https://user-images.githubusercontent.com/2521888/121517517-612e8880-c9e7-11eb-8a39-433aa1043d65.png">

Form landing more actions menu:
<img width="255" alt="Screenshot 2021-06-10 at 13 29 16" src="https://user-images.githubusercontent.com/2521888/121517613-802d1a80-c9e7-11eb-98f0-db26b0ef9947.png">

Account menu:
<img width="303" alt="Screenshot 2021-06-10 at 13 29 44" src="https://user-images.githubusercontent.com/2521888/121517671-8e7b3680-c9e7-11eb-8d7d-548ac48ddd12.png">

Library list more actions menu and library item landing page more actions menu (same wrapper component):
<img width="229" alt="Screenshot 2021-06-10 at 13 30 35" src="https://user-images.githubusercontent.com/2521888/121517883-c2565c00-c9e7-11eb-86ef-14e1b696fe28.png">
<img width="238" alt="Screenshot 2021-06-10 at 13 30 40" src="https://user-images.githubusercontent.com/2521888/121517891-c4b8b600-c9e7-11eb-8ea4-fe7bf53484d1.png">

Public collections table filter:
<img width="238" alt="Screenshot 2021-06-10 at 13 30 40" src="https://user-images.githubusercontent.com/2521888/121518759-c20a9080-c9e8-11eb-87b9-39888eb9af5c.png">

Map view by menu:
[...]

PR #3273, opened 2021-06-10
Don't force form media's `dropzone` look everywhere

duvld:
## Description

Revert global `dropzone` CSS changes, make changes specific to form media's dropzone.

### DIFF
--- a/jsapp/js/components/modalForms/formMedia.es6
+++ b/jsapp/js/components/modalForms/formMedia.es6
@@ -237,7 +237,7 @@ class FormMedia extends React.Component {
             {!this.state.isUploadFilePending && (
               <Dropzone
                 onDrop={this.onFileDrop.bind(this)}
-                className='dropzone-settings dropzone-form-media'
+                className='dropzone-settings kobo-dropzone--form-media'
               >
                 {this.state.fieldsErrors?.base64Encoded && (
                   <bem.FormView__cell m='error'>
@@ -254,7 +254,7 @@ class FormMedia extends React.Component {
             )}
 
             {this.state.isUploadFilePending && (
-              <div className='dropzone-settings dropzone-form-media'>
+              <div className='dropzone-settings kobo-dropzone--form-media'>
                 <LoadingSpinner message={t('Uploading file…')} />
               </div>
             )}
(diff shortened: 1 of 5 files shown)


jsapp/js/components/modalForms/formMedia.es6
```diff
@@ -237,7 +237,7 @@ class FormMedia extends React.Component {
             {!this.state.isUploadFilePending && (
               <Dropzone
                 onDrop={this.onFileDrop.bind(this)}
-                className='dropzone-settings'
+                className='dropzone-settings dropzone-form-media'
```
magicznyleszek, line 240: Should we name it `kobo-dropzone` and `kobo-dropzone--form-media`?

jsapp/scss/components/_kobo.modal.scss
```diff
@@ -770,6 +766,17 @@ $z-modal-x: 10;
   &.dropzone-reject {border-color: $kobo-red;}
 }
 
+// Form media specific dropzone
+.dropzone-form-media {
+  background: $kobo-white;
+  padding: 40px;
+  border: 6px dashed;
+  border-color: $kobo-gray-92;
```
magicznyleszek, line 774: You can go with `border: 6px dashed $kobo-gray-92;` :)

jsapp/js/components/modalForms/formMedia.es6
magicznyleszek: And maybe move all dropzone related styles to separate file in `js/components/common`?

magicznyleszek (changes requested):

jsapp/js/components/modalForms/formMedia.es6
duvld: image

there's a lot of modal specific changes, but I guess the one in `_kobo.modal.scss` is the most generic so I'll move that one :D

duvld: it would be good to make a `dropzone` overhaul and make it a BEM element with `m` tags for each of these modals, but maybe we should do it in a different PR

magicznyleszek: Yeah ultimately it would be nice to have a generic dropzone styles. I'm pondering about the same problem with our dropdowns right now in https://github.com/kobotoolbox/kpi/issues/3266

PR #3276, opened 2021-06-10
[Feature] Adds `background-audio` as a meta question type into formbuilder

### LINKED ISSUE #3275: Intergrate `background-audio` into the formbuilder
## Description

Split from #3193 this issue will be closed after the formbuilder half of the issue is merged into `beta`

duvld:
## Description

Can now add a [`background-audio` meta question](https://docs.getodk.org/form-question-types/#background-audio-recording) to forms via the formbuilder sidebar:

image

## Related issues
Formbuilder side of things of #3193 
Fixes #3275

### DIFF
--- a/jsapp/js/components/metadataEditor.es6
+++ b/jsapp/js/components/metadataEditor.es6
@@ -242,11 +242,11 @@ export default class MetadataEditor extends React.Component {
               <Select
                 className='kobo-select'
                 classNamePrefix='kobo-select'
-                {/*
+                /*
                   defaultValue only displays the default value, it does not
                   append it to the JSON. If there is no quality parameter then
                   it would default to ODK_DEFAULT_AUDIO_QUALITY behind the scenes
-                */}
+                */
                 defaultValue={ODK_DEFAULT_AUDIO_QUALITY}
                 options={AUDIO_QUALITY_OPTIONS}
                 onChange={this.onBackgroundAudioParametersChange}
--- a/jsapp/js/components/metadataEditor.es6
+++ b/jsapp/js/components/metadataEditor.es6
@@ -31,7 +31,7 @@ export default class MetadataEditor extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      metaProperties: []
+      metaProperties: [],
     };
     autoBind(this);
   }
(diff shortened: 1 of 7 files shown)


dorey:
Is `quality=voice-only` the default parameter? When i activated this value it showed up in the sidebar as being "Voice only" but then the JSON structure did not have any parameters reflected.

duvld:
> Is `quality=voice-only` the default parameter? When i activated this value it showed up in the sidebar as being "Voice only" but then the JSON structure did not have any parameters reflected.

`voice-only` is what it will default to if there is no `quality=*` parameter specified, so yeah the default. I didn't want to add a parameter if the user wanted to keep it this quality, and only append the parameters column if they made an explicit decision  

I should probably make a comment explaining this :D

dorey:
image
This text seems to imply that it doesn't work in ODK Collect, which is not the case. I think that in the past we have been OK referring to a generic "Collect" which could mean ODK or KoBoCollect.

jnm:
> > Is `quality=voice-only` the default parameter? When i activated this value it showed up in the sidebar as being "Voice only" but then the JSON structure did not have any parameters reflected.
> 
> `voice-only` is what it will default to if there is no `quality=*` parameter specified, so yeah the default. I didn't want to add a parameter if the user wanted to keep it this quality, and only append the parameters column if they made an explicit decision
> 
> I should probably make a comment explaining this :D

the default could change in the future to something other than "Voice only", so i'd advocate for writing into the JSON when "Voice only" is chosen. maybe add an option to the drop down called "Default" that avoids appending the parameters column?

duvld:
> > > Is `quality=voice-only` the default parameter? When i activated this value it showed up in the sidebar as being "Voice only" but then the JSON structure did not have any parameters reflected.
> > 
> > 
> > `voice-only` is what it will default to if there is no `quality=*` parameter specified, so yeah the default. I didn't want to add a parameter if the user wanted to keep it this quality, and only append the parameters column if they made an explicit decision
> > I should probably make a comment explaining this :D
> 
> the default could change in the future to something other than "Voice only", so i'd advocate for writing into the JSON when "Voice only" is chosen. maybe add an option to the drop down called "Default" that avoids appending the parameters column?

If the user doesn't change anything in the dropdown then no parameters would be appended, it would only append if they explicitly clicked an option. I think this would be better than a "Default" that would imply to me that there is a default that isn't "Voice-only"

dorey (approved):
Tested out locally and `background-audio` saves as expected in the forms.

Issue #2675, opened 2020-05-14
Fetch old form media from KoBoCAT


jnm:
Since KPI is going to be handling form media (#2487, #2537, #2673, #2674) we need a way to get existing form media files from KoBoCAT into KPI. Note that some KoBoCAT `XForm`s simply won't have any corresponding KPI `Asset`, and ~~we should ignore those~~ [we can't ignore them if we're taking away the settings interface in KoBoCAT]. This is kind of like `sync_kobocat_xforms` :neutral_face: although we have no intention of running it periodically.

jnm:
Idea: "deleted" flag on KPI media file object
1. unicorn.png is a media file that was uploaded to KPI
1. person wants to delete unicorn.png; UI or whatever sends HTTP DELETE to KPI back end
1. KPI deletes unicorn.png from storage immediately, and sets "deleted" flag to true on media file object associated with unicorn.png
    * we must hide all media file objects that have the "deleted" flag set to true
1. person redeploys their magical creatures form that previously had unicorn.png in it
1. because unicorn.png media file object is marked as deleted, KPI sends HTTP DELETE to KoBoCAT to remove unicorn.png
1. once KPI receives successful response (or maybe 404 too; see form deletion logic), KPI removes the unicorn.png media file object entirely from the database

this would solve the issue of KoBoCAT media getting unintentionally trashed after we deploy KPI media file handling, but it would not make any of the old KoBoCAT media visible in KPI.

jnm:
i think this is related *enough* to stash here to avoid forgetting about it:

1. the kpi database cannot enforce uniqueness because the filename comes from either the `content` (for an uploaded file) or the `metadata` (for a URL)
1. however, kobocat does enforce uniqueness (of uploaded files and URLs separately?)
1. there are possibly sticky situations where a user would have to go into the kobocat ui and remove a file to solve a conflict(?)
1. can we generally fix this by adding some new parameter to the kobocat endpoint for form media file uploads that would allow overwriting a file with the same name instead of failing with an `IntegrityError` (or whatever it does now)

_Originally posted by @jnm in https://github.com/kobotoolbox/kpi/pull/2537#discussion_r608792158_

joshuaberetta:
> Idea: "deleted" flag on KPI media file object
> 
>     1. unicorn.png is a media file that was uploaded to KPI
> 
>     2. person wants to delete unicorn.png; UI or whatever sends HTTP DELETE to KPI back end
> 
[...]

PR #3278, opened 2021-06-11
Update `sync_kobocat_xform` command

joshuaberetta:
Minor fix to #2675


noliveleger (changes requested):
I cannot comment on the line I would like to, but I would remove the condition `if username` on line 583.
I would expect the media files to be synced when I called `--sync-kobocat-form-media`, 
even if it's all files.  Because it's false by default, if the user adds the option, we should assume they know what they are doing.
With the current behaviour, nothing happens if the username is not provided. 
What do you think? 

Morever, I think we should pass the verbosity to both sub commands `populate_kc_xform_kpi_asset_uid` and `sync-kobocat-form-media`.

joshuaberetta:
Those are good points 👍 I'll make the changes.

noliveleger (approved):

PR #3297, opened 2021-06-16
Optimize `sync_kobocat_form_media`

### LINKED ISSUE #3296: Optimize `sync_kobocat_form_media`
## Description

Optimize `sync_kobocat_form_media` to use less system memory and write to `stdout` while running rather than large JSON output.

joshuaberetta:
## Description

Reduce memory usage while command is executing.

## Related issues

closes #3296

### DIFF
--- a/kpi/management/commands/sync_kobocat_form_media.py
+++ b/kpi/management/commands/sync_kobocat_form_media.py
@@ -122,6 +122,7 @@ class Command(BaseCommand):
             'owner',
             'parent_id',
             'name',
+            '_deployment_data',
         )
         if username is not None:
             assets = assets.filter(owner__username=username)
--- a/kpi/management/commands/sync_kobocat_form_media.py
+++ b/kpi/management/commands/sync_kobocat_form_media.py
@@ -1,7 +1,7 @@
 # coding: utf-8
-import json
 import requests
 import sys
+import time
 from typing import Optional
 
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


jnm:
Ideas for this:
* call `raise_for_status()` on all(?) requests to avoid silent failures
* but log failures of any kind and keep going
* enhancements for large dbs:
  * make easier to resume after failure, maybe by looping over all users (sorted alphabetically by username) and then getting all assets for each user
  * print status info per user e.g.
    1. "starting sync for jnm_demo": having the username logged before we try to do work lets us figure out where a problem lies if the script crashes
    2. "finished sync for jnm_demo: w forms succeeded, x forms failed, y media files already in kpi, z media files added to kpi, in t seconds (n% of users complete)": having a shorter summary per user might make it easier to spot problems. seeing the number of files actually being added would let us catch sync problems (if that number doesn't go down when re-running).

nb: of course a more machine-digestable format (tab-separated, json, etc.) is completely fine and may be better than my quick ideas. the general goal would be to spot obvious problems visually without waiting for the whole job to complete, while maintaining the ability to grep, jq or whatever through the total output for more isolated or subtle problems.

noliveleger:
@jnm, @joshuaberetta . should we merge this?

noliveleger:
I'm merging this PR since it already brings some optimization.
I've open another issue with @jnm's ideas to improve the management command.

PR #3307, opened 2021-06-21
Copy fonts post npm install

magicznyleszek:
## Description

Small internal dependencies workflow improvement for development conveniance.


package.json
```diff
@@ -81,6 +81,7 @@
     "zxcvbn": "^4.4.2"
   },
   "scripts": {
+    "postinstall": "npm run copy-fonts",
```
duvld, line 84: Is this supposed to aid also in things like this: (where it seems like the OS/browser fonts are taking over our custom fonts)?
image
I've been having this issue for a while and I'm not sure why :P the fonts throughout KPI look a bit off/wider than they're supposed to

duvld: Nevermind! The above is a result of different Roboto fonts coming from my OS. On Ubuntu it produces the above but on Arch it's using the correct font.

PR #3309, opened 2021-06-22
Submissions by Country Report

### LINKED ISSUE #3065: Report - Number of interviews conducted per country
## Description
Create a monthly report that includes the number of interviews conducted per country using two metrics:
1. Self-reported location filled in for the KoBo projects (higher priority)
2. Submission country based on IP address to be stored as a counter per country, per month for submission POST requests 

Note: Server logs don’t go back forever

This should also be accessible and downloadable via Django admin 
The country and time frame should be able to be selected when generating the report

## Suggestions
- A counter by country for future interviews
- Update location on authentication and when the single page application is refreshed
- Use an IP look up tool to retrieve the country only

JacquelineMorrissette:
## Checklist

3. [x] Ensure the tests pass

To use the filters on the `superuser_stats/country_report/`, add a `start_date` and `end_date` query
Example:
`https://[KPI]/superuser_stats/country_report/?start_date=YYYY-MM-DD&end_date=YYYY-MM-DD`

## Description

Creates a report for Submission-by-country which can be filtered by a start and end date at the `/superuser_stats/country_report/` URI.

## Related issues

Fixes #3065


kobo/apps/usage_statistics/admin.py
```diff
@@ -114,3 +186,4 @@ def __get_serialized_data(self, request) -> list:
 
 
 admin.site.register(KobocatSubmissionCounter, UserStatisticsAdmin)
+admin.site.register(Asset, SubmissionsByCountry)
```
noliveleger: It would be better to register `ReadOnlyKobocatInstance` and change its verbose name and its app to `usage_statistic` to make it appear under the same section we created earlier for reports. 


FYI: PEP-8, you need a new line at the end the file. (This is can be set automatically with your IDE)

kobo/apps/superuser_stats/tasks.py
```diff
@@ -1,21 +1,81 @@
...
+                xform=xform,
+                date_created__range=(start_date, end_date),
+            ).count()
+
+        row.append(label)
+        row.append(count)
+
+        return row
+
+    CHUNK_SIZE = 1000
```
noliveleger, line 57: Unused?

kobo/apps/superuser_stats/templates/superuser_stats.html
```diff
@@ -0,0 +1,35 @@
...
+          </td>
+        </tr>
+        <tr>
+          <td>
+            <a href="{{ kpi_base_url }}/superuser_stats/user_report/">User Report</a>
+          </td>
+        </tr>
+      </tbody>
+    </table>
+{% endblock %}
```
noliveleger, line 35: I would move this template under `usage_statistics` app.

noliveleger: It seems to have an issue with the CSS of the table. It takes 100% of the width. If the viewport width is too small, the count column is hidden by the filter. 

Screen Shot 2021-06-22 at 10 13 51
Screen Shot 2021-06-22 at 10 14 00

This css rule fixes it. I don't know if you can change the CSS directly or if you need to include a stylesheet that override the behaviour for this module. Last solution, you can add the style within the template.

```css
#changelist table {
    width: calc(100% - 250px);
}
```

noliveleger (changes requested):

kobo/apps/superuser_stats/tasks.py
```diff
@@ -1,21 +1,87 @@
...
+@shared_task
+def generate_country_report(output_filename, start_date, end_date):
+
+    def format_date(d):
+        if hasattr(d, 'strftime'):
+            return d.strftime('%F')
+        else:
+            return d
+
+    def get_row_for_country(code, label):
```
noliveleger, line 32: Would you mind to add function annotations. (Feel free to add them to existing code in this file too). The signature of the inner function is quite simple but it helps to find out which type each variable is pretty quickly.

Moreover, as a side note: I'm used to use different variable names inside inner functions to avoid shadowing  variables in the same scope.   

For example: 
```python
def generate_country_report(output_filename: str, start_date: date, end_date: date):
    ... 
    def get_row_for_country(code_: str, label_: str) -> dict:   # use `code_` instead of `code` and `label_` instead  of `label` 
        row_ = []   # use `row_` instead of `row`
        ...
        return row_
        
        
    with default_storage.open(output_filename, 'wb') as output_file:
        ...
        for code, label in COUNTRIES:

            try:
                row = get_row_for_country(code, label)  # `row`, `code, `label` are not shadowed by `get_row_for_country()` variables. 
            except Exception as e:
                row = ['!FAILED!', 'Country: {}'.format(label), repr(e)]
            writer.writerow(row)     
```            

It's not a big deal, the code works the way it is, but IMHO, it avoids making some mistake and it is more obvious when you debug to find out the value of each variable.

kobo/apps/superuser_stats/admin.py
```diff
@@ -114,3 +196,4 @@ def __get_serialized_data(self, request) -> list:
 
 
 admin.site.register(KobocatSubmissionCounter, UserStatisticsAdmin)
+admin.site.register(ReadOnlyKobocatInstance, SubmissionsByCountry)
```
noliveleger, line 199: We should deactivate the add button. These reports are read-only, aren't they?

Screen Shot 2021-07-01 at 11 06 22

Screen Shot 2021-07-01 at 11 10 28

kobo/apps/superuser_stats/admin.py
```diff
@@ -48,6 +57,79 @@ def queryset(self, request, queryset):
...
+                xform_id_strings = list(assets.values_list(
+                    '_deployment_data__backend_response__id_string', flat=True
+                ).filter(
+                    _deployment_data__backend='kobocat',
+                    asset_type=ASSET_TYPE_SURVEY,
+                ))
+
+                instances_count = ReadOnlyKobocatInstance.objects.filter(
+                    xform__id_string__in=xform_id_strings,
+                ).count()
```
noliveleger, line 123: We may be able to speed-up things a little bit. The number of submissions is already calculated within the `logger_xform` table. 

What about returning the count directly from  `logger_xform` with `SUM()`? I
```python
result = KobocatXForm.objects.filter(xform__id_string__in=xform_id_strings).aggregate(instances_count=Sum('num_of_submissions'))
instances_count = result['instances_count']
```

kobo/apps/superuser_stats/admin.py
```diff
@@ -48,6 +57,79 @@ def queryset(self, request, queryset):
...
+
+                data.append({
+                    'country': name,
+                    'count': instances_count,
+                })
```
noliveleger, line 128: Would it better to initialize `instances_count` to 0 and move these lines out of the `if` statement to always display a line for each country in the HTML table?

noliveleger (changes requested):

PR #3311, opened 2021-06-22
Update supported form media types

### LINKED ISSUE #3310: Update supported asset file media types
## Description

`zip` should be supported for valid form media.

joshuaberetta:
## Related issues

closes #3310

### DIFF
--- a/kpi/models/asset_file.py
+++ b/kpi/models/asset_file.py
@@ -36,7 +36,14 @@ class AssetFile(OpenRosaManifestInterface, models.Model):
     )
 
     ALLOWED_MIME_TYPES = {
-        FORM_MEDIA: ('image', 'audio', 'video', 'text/csv', 'application/xml'),
+        FORM_MEDIA: (
+            'image',
+            'audio',
+            'video',
+            'text/csv',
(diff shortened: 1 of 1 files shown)


kpi/models/asset_file.py
```diff
@@ -36,7 +36,14 @@ class AssetFile(OpenRosaManifestInterface, models.Model):
...
 
     ALLOWED_MIME_TYPES = {
-        FORM_MEDIA: ('image', 'audio', 'video', 'text/csv', 'application/xml'),
+        FORM_MEDIA: (
+            'image',
+            'audio',
+            'video',
+            'text/csv',
+            'application/xml',
+            'application/zip',
```
noliveleger, line 45: What about adding the `application/x-zip-compressed` too?

noliveleger (changes requested):

kpi/models/asset_file.py
joshuaberetta: I saw that the result of `guess_type('something.zip')` in the kpi container results in `application/zip`. It was actually the same for kobocat...

Issue #3199, opened 2021-05-10
Usability improvements to Table View


Ig-Rebollo:
This issue if the second (and optional) half of a series improvements to the table view UI. These improvements are divided in two groups:

1. **Styling** improvements, (issue #3198) which should be done before any other table view related additions; and
2. **Usability** improvements, (this issue) which can be done along with those updates, or later on, depending on timelines and availability.

### Usability improvements
The following updates involve small new functionalities and minor additions to make the experience more user friendly. These are not as urgent as the accessibility improvements, and most are optional, depending on what we have time for. Many can also be postponed, in which case they will be recorded in the Notion idea repository. See list below.
![TableviewNEW](https://user-images.githubusercontent.com/22472326/118162868-428ca000-b3ef-11eb-8d4e-b9c8ab0df917.jpg)

**Making column headers clearer and more useful**
- [ ] Include an arrow with a dropdown menu for easier to understand actions (current sorting is not very intuitive). Note that a sorted column would stay highlighted and include the icon of the type of sorting. The dropdown would also allow you to cancel the sorting. See animation:
![column sorting](https://user-images.githubusercontent.com/22472326/117689363-aa40b200-b187-11eb-8b8b-c8076312a1bc.gif)

**Include more navigation options on top.**
- [ ] While 'filter' might be something more challenging for a later occasion, we could start with a 'hide columns' functionality. This could also be accessible from the column header dropdown described before.
- [ ] Note that in order to do this, 'number of results' and 'rows displayed' should be removed from where they are and included in the table itself.
![hide fields](https://user-images.githubusercontent.com/22472326/117689752-14f1ed80-b188-11eb-9da5-c406298c9cc0.gif)

Additional details and specs can be found [in the following Figma file](https://www.figma.com/file/vbHVtRkdit3dqd5y8JFfXF/Table-view-redesign?node-id=0%3A1).

PR #3317, opened 2021-06-23
Kobo dropdown component

magicznyleszek:
## Description

A new UI component for dropdowns.

## Additional details

`KoboDropdown` will be replacing the outdated `ui.PopoverMenu` component.

## Related issues

Needed for #3199 
Part of #3266

### DIFF
--- a/jsapp/js/components/common/koboDropdown.es6
+++ b/jsapp/js/components/common/koboDropdown.es6
@@ -1,0 +1,53 @@
+import React from 'react';
+import autoBind from 'react-autobind';
+import {bem} from 'js/bem';
+
+/**
+ * A generic dropdown component that accepts any content inside. If you need a
+ * selector dropdown, please look at `react-select`.
+ *
+ * @prop {boolean} opened
+ * @prop {function} onClose
+ *
+ * @prop {boolean} [closeOnMenuClick] `false` by default - closes menu whenever user clicks inside it, useful for simple menu with a list of actions
+ * @prop {boolean} [closeOnBlur] `false` by default - closes menu when it loses focus (e.g. user clicked outside)
+ * @prop {boolean} [closeOnMouseLeave] `false` by default
+ * @prop {boolean} [closeOnEsc] `false` by default
+ * @prop {node} trigger
+ * @prop {function} content the content of dropdown, anything's allowed
+ */
+class KoboDropdown extends React.Component {
+  constructor(props){
+    super(props);
+    autoBind(this);
+  }
+
+  onClose(evt) {
+    this.props.onClose(evt);
+  }
+
+  render() {
+    return (
+      <bem.KoboDropdown>
+        <bem.KoboDropdown__wrapper>
+          <bem.KoboDropdown__input
+            type='checkbox'
+            name={this.props.name}
+            id={this.props.id}
+            onChange={this.onChange}
+            checked={this.props.checked}
+            disabled={this.props.disabled}
+          />
+
+          {this.props.label &&
+            <bem.KoboDropdown__label htmlFor={this.props.id}>
+              {this.props.label}
+            </bem.KoboDropdown__label>
+          }
+        </bem.KoboDropdown__wrapper>
+      </bem.KoboDropdown>
+    );
+  }
+}
+
+export default KoboDropdown;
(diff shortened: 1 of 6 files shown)


jsapp/js/components/common/koboDropdown.es6
```diff
@@ -0,0 +1,202 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import autoBind from 'react-autobind';
+import {bem} from 'js/bem';
+import {KEY_CODES} from 'js/constants';
+import './koboDropdown.scss';
+
+export const KOBO_DROPDOWN_THEMES = {};
+new Set([
```
duvld, line 9: Should we be declaring all local constant sets this way? I've usually just been doing something like `set = [one: {id:'one'}, ...]`

jsapp/js/components/common/koboDropdown.es6
```diff
@@ -0,0 +1,202 @@
...
+  'up-left',
+  'up-center',
+  'up-right',
+  'down-left',
+  'down-center',
+  'down-right',
+]).forEach((codename) => {KOBO_DROPDOWN_PLACEMENTS[codename] = codename;});
+Object.freeze(KOBO_DROPDOWN_PLACEMENTS);
+
+bem.KoboDropdown = bem.create('kobo-dropdown');
```
duvld, line 26: I like this! On top of the other reasons I find that it gives a good flow for the file and it prepares me for the seeing new components :D

jsapp/js/components/common/koboDropdown.es6
```diff
@@ -0,0 +1,202 @@
...
+ * A generic dropdown component that accepts any content inside the menu and
+ * inside the opener.
+ *
+ * NOTE: If you need a select-type dropdown, please use `react-select`.
+ *
+ * You can use some existing content elements:
+ * - bem.KoboDropdown__menuButton - a generic dropdown row button
+ * - bem.KoboDropdown__triggerButton - a generic dropdown row button
+ *
+ * @prop {string} [theme] - `light` by default - one of KOBO_DROPDOWN_THEMES
```
duvld, line 41: Curious, why not use the jsdoc syntax for specifying default values: `param {string} [somebody=John Doe]`?

jsapp/js/components/common/koboDropdown.es6
```diff
@@ -0,0 +1,202 @@
...
+      wrapperMods.push('disabled');
+    }
+
+    return wrapperMods;
+  }
+
+  render() {
+    return (
+      <bem.KoboDropdown m={this.getWrapperModifiers()}>
+        <bem.KoboDropdown__trigger onClick={this.onTriggerClick}>
```
duvld, line 190: I took a sneak peak at your usability improvements branch :P I noticed that the caret:
image
is a little hard to click on (it looks like you have to click slightly above it) Maybe we should add a wrapper class before the `props` content with a minimum height and width?

jsapp/js/components/common/koboDropdown.es6
magicznyleszek: I like this way because you don't risk a typo in the name - in `one: {id: 'one'}` you need "one" and "one" to be identical, and althought it seems pretty simple there is some risk of a bug. With this convoluted way there is none :) Plus the freeze is always nice :D

jsapp/js/components/common/koboDropdown.es6
magicznyleszek: I like it too :)

jsapp/js/components/common/koboDropdown.es6
magicznyleszek: Didn't know about the default value! Changed :)

jsapp/js/components/common/koboDropdown.es6
magicznyleszek: I thought about it and I decided it would be better for the `KoboDropdown` user to be responsible for the trigger clickability area. Otherwise the wrapper with minimum height/width might get in a way. I tried to have the component as generic as possible - that's why the `bem.KoboDropdown__menuButton` is an optional useful BEM not a requirement :)

magicznyleszek: I might add some tweaks to `KoboDropdown` thought while I work on it on that branch and discover something :)

duvld (approved):

PR #3328, opened 2021-07-01
Don't allow illegal matrix column names

### LINKED ISSUE #3326: Form-builder allows invalid matrix question `Data Column Suffix` value
## Description

The `Data Column Suffix` field in the matrix question settings does not replace spaces with underscores as it does with other question names, therefore leaving it open to user error and incomplete data exports, such as in this case: https://community.kobotoolbox.org/t/missing-data-on-export-yet-its-there-in-the-tool-its-a-group-question/20130

The Enketo preview gives this error:
```
[row : 6] Invalid question name [group_hl59k79_header_a space] Names must begin with a letter, colon, or underscore.Subsequent characters can include numbers, dashes, and periods.
```
Strangely the form opens and one can submit without error. When one tries to export with KPI, the data in the matrix is not included because of this now invalid name (fortunately the legacy export includes the data).

## Steps to Reproduce

1. Create a matrix question
2. Enter a value with spaces in the `Data Column Suffix` field
2. Try previewing the form
3. Submit data
4. Try exporting through KPI 

## Expected behavior 

The spaces are replaced to create a valid field value.

## Actual behavior 

Invalid values are allowed through.

magicznyleszek:
## Description

Don't allow matrix column names that contain illegal characters (e.g. spaces). Illegal names caused serious problems for Enketo.

## Related issues

Fixes #3326


jsapp/js/formbuild/containers/KoboMatrix.es6
```diff
@@ -96,17 +96,17 @@ class KoboMatrix extends React.Component {
...
 
     if (type === 'column') {
-      data.get('cols').forEach(function(ch){
+      data.get('cols').forEach(function (ch) {
         names.push(data.getIn([ch, '$autoname']));
       });
     } else {
-      data.get('choices').forEach(function(ch){
-        if (ch.get('list_name') == ln)
+      data.get('choices').forEach(function (ch) {
```
joshuaberetta, line 106: Should `survey`, `choices`, `settings`, `list_name`, `$autoname`, `$autovalue` etc. also be things we can include with the new `createEnum()` function?

jsapp/js/formbuild/containers/KoboMatrix.es6
```diff
@@ -96,17 +96,17 @@ class KoboMatrix extends React.Component {
 
   autoName(val, type, ln = false) {
     var names = [];
-    const kobomatrix_list = this.state.kobomatrix_list;
     var data = this.state.data;
 
     if (type === 'column') {
-      data.get('cols').forEach(function(ch){
+      data.get('cols').forEach(function (ch) {
```
joshuaberetta, line 102: What's your preference for this kind of situation?
```javascript
someArray.forEach((...) => {
  ...
})
```
or 
```javascript
someArray.forEach(function (...) {
  ...
})
```

joshuaberetta (changes requested):
It looks like new behaviour is quite strange — it seems to only be accepting letters and numbers, but I can move the cursor back and then enter an underscore or space between those letters or numbers that will then be converted to an underscore.

jsapp/js/formbuild/containers/KoboMatrix.es6
magicznyleszek: The preference is the arrow function, just for brevity :)

jsapp/js/formbuild/containers/KoboMatrix.es6
magicznyleszek: I would say: "maybe" :D If we'd write more new code that uses some of these names, then we would use an enum, but as this is just a bugfix with some code style cleanup, I would leave it

magicznyleszek:
> It looks like new behaviour is quite strange — it seems to only be accepting letters and numbers, but I can move the cursor back and then enter an underscore or space between those letters or numbers that will then be converted to an underscore.

@joshuaberetta yeah this function we have is a bit weird - the rules for names and the rule for uniqueness causes this weird behaviour. I think I will try to apply the rules only after user ends typing or the input leaves focus

jsapp/js/formbuild/containers/KoboMatrix.es6
```diff
@@ -263,12 +348,12 @@ class KoboMatrix extends React.Component {
...
     const val = this.state.data.getIn([colKuid, 'required']);
-    return val == true || val == 'true' ? true : false;
+    return val === true || val === 'true' ? true : false;
   }
 
   newChoiceOption(e) {
-    var data = this.state.data,
-        listName = null;
+    var data = this.state.data;
+    let listName = null;
```
joshuaberetta, line 356: Is there a reason to have both `let` and `var` here (considering you've now changed that line anyway 💁‍♂️

jsapp/js/formbuild/containers/KoboMatrix.es6
```diff
@@ -176,15 +265,13 @@ class KoboMatrix extends React.Component {
...
+        onok: () => {
           data = data.setIn([colKuid, 'type'], newType);
           data = data.deleteIn([colKuid, 'select_from_list_name']);
           _this.setState({data: data});
           _this.toLocalStorage(data);
         },
-        oncancel: () => {
-          dialog.destroy();
-        }
+        oncancel: dialog.destroy,
```
joshuaberetta, line 274: A naive question, but is there is a reason for this not to be `onOk` and `onCancel`?

jsapp/js/formbuild/containers/KoboMatrix.es6
```diff
@@ -345,14 +428,12 @@ class KoboMatrix extends React.Component {
       title: t('Delete column?'),
       message: t('Are you sure you want to delete this column? This action cannot be undone.'),
       labels: {ok: t('Delete'), cancel: t('Cancel')},
-      onok: (evt, val) => {
-        data = data.update('cols', cols => cols.filterNot(x => x === colKuid));
+      onok: () => {
+        data = data.update('cols', (cols) => cols.filterNot((x) => x === colKuid));
```
joshuaberetta, line 432: Maybe something nicer than `x`? 🙈

joshuaberetta (changes requested):
Works great! 🎉

jsapp/js/formbuild/containers/KoboMatrix.es6
magicznyleszek: Oh, this thing is coming from a plugin -_- can't really do anything with it 🤷

jsapp/js/formbuild/containers/KoboMatrix.es6
magicznyleszek: Right, fixed :)

jsapp/js/formbuild/containers/KoboMatrix.es6
magicznyleszek: Changed it to `col` :)

PR #3337, opened 2021-07-12
Fix Add from Library search form filter and counter

### LINKED ISSUE #3335: Public collections not appearing as filter option in Add from Library
## Description

Subscribed public collections are not listed as a filter option in Form Builder in the aside Add from Library menu. The items from the collections are listed below though.

So in this example below:
- `asda` is my own collection
- there is a public collection `Leszek's collection` that I'm subscribed to (as you can see it doesn't appear as option next to `asda` :bug:)
- `Enter a number` and `Enter a time` both comes from `Leszek's collection`

Screenshot 2021-07-09 at 18 22 08

magicznyleszek:
## Description

Make subscribed public collections appear as a filter option in Form Builder's "Add from Library" search form. Also fix the count in same place.

## Related issues

Fixes #3335


jsapp/js/editorMixins/assetNavigator.es6
```diff
@@ -91,21 +95,24 @@ class AssetNavigatorListView extends React.Component {
         );
     } else {
 
-      window.setTimeout(()=>{
-        this.activateSortable();
-      }, 1);
+      // TODO FIXME
+      // HACK: we activate sortable with timeout so it is rendered :puke:
+      window.setTimeout(() => {this.activateSortable();}, 1);
```
duvld, line 100: I know there's a comment but I'm curious why the timeout is so short and arbitrary :P, does it just need to wait for the list to populate?

jsapp/js/components/list.es6
```diff
@@ -154,55 +154,69 @@ export class ListCollectionFilter extends React.Component {
...
+      actions.library.getCollections.completed.listen(this.onGetCollectionsCompleted)
+    );
     this.queryCollections();
   }
-  queryCollections () {
-    dataInterface.getCollections().then((collections)=>{
-      var availableCollections = collections.results.filter((value) => {
-        return (
-          value.access_types &&
-          !value.access_types.includes(ACCESS_TYPES.public)
```
duvld, line 167: Were we trying to exclude `public` access types before but not now? I'm guessing it's because public collections is now live but just want to be sure

jsapp/js/editorMixins/assetNavigator.es6
```diff
@@ -9,13 +9,13 @@ import {searches} from '../searches';
...
-  ASSET_TYPES
+  ASSET_TYPES,
 } from '../constants';
 import {
   ListSearch,
   ListTagFilter,
   ListCollectionFilter,
-  ListExpandToggle
+  ListExpandToggle,
 } from '../components/list';
```
duvld, line 19: should this be relative?

jsapp/js/editorMixins/assetNavigator.es6
magicznyleszek: It probably might be `0`. This is needed because sortable plugin needs the element to exist (i.e. be already rendered by React). Without timeout this line fails: `var $el = $(ReactDOM.findDOMNode(this.refs.liblist));` as it is `null`. We could do it better for sure, probably by implementing some checks that will see if element is already rendered and if sortable was already activated - then we can safely run the function with each render

jsapp/js/components/list.es6
magicznyleszek: Yeah that's the reason :)

jsapp/js/editorMixins/assetNavigator.es6
magicznyleszek: Fixed the import paths :)

jsapp/js/editorMixins/assetNavigator.es6
duvld: Understood :+1:

PR #3343, opened 2021-07-13
Fix team members UI

magicznyleszek:
## Description

Bring back team members UI from Form Summary view.

## Additional details

This thing:
Screenshot 2021-07-13 at 14 32 40


jsapp/js/components/formSummary.es6
```diff
@@ -283,15 +287,23 @@ class FormSummary extends React.Component {
...
+
   renderTeam() {
-    var team = [];
-    this.state.permissions.forEach(function(p){
-      if (p.user__username && !team.includes(p.user__username) && p.user__username != 'AnonymousUser')
-        team.push(p.user__username);
+    const team = [];
+    this.state.permissions.forEach((perm) => {
+      let username = null;
+      if (perm.user) {
```
duvld, line 295: Was the issue that `user__username` doesn't exist? I don't see anything in `this.state.permissions` that has that attribute at this stage

duvld (approved):

jsapp/js/components/formSummary.es6
magicznyleszek: Yes that's it :)

PR #3362, opened 2021-07-22
Cleanup stylesheets

### LINKED ISSUE #3248: Rearrange Frontend code
## Description

We should group our files by their function, not by their type/extension. TL;DR the simplest example of this idea would be:

bad:
```
/js/components/button.es6
/js/components/table.es6
/css/button.scss
/css/table.scss
/tests/button.tests.js
```

good:
```
/components/button/button.es6
/components/button/button.scss
/components/button/button.tests.js
/components/table.es6
/components/table.scss
```

Also components should define themselves in their `*.es6` file rather than in global `bem.es6` (which should ultimately be empty or almost empty). 

Also instead of importing styles in `main.scss` for the components it is better to import them in the component's `es6` file.

Doable TODO:
1. [x] Understand why we have `kobo.scss` and `main.scss` - possibly integrate `kobo.scss` into `main.scss`
2. [ ] Move `main.scss` to `js` directory (this would become `src` in the future)
3. [x] Move to `jsapp/js/components/common` all stylesheets that are connected to components described in it
4. [ ] Move to `jsapp/js/components/common` all stylesheets that are a css-only common component (buttons?)
5. [x] Check and adapt our `webpack` to be ok with this changes

## Additional information

This will work nicely with #3250

magicznyleszek:
## Description

Move around frontend files in a more logical fashion. Just a file cleanup changes.

## Related issues

Fixes #3248


jsapp/js/components/reports/reports.scss
```diff
@@ -198,27 +200,27 @@
   }
 
   &--vertical {
-    background-image: url("~../img/chart-vertical.png");
```
p2edwards, line 201: Does [the tilde in the url()](https://stackoverflow.com/questions/39535760/what-does-a-tilde-in-a-css-url-do) do anything here? From my reading it doesn't, and can be removed, but I might be missing something.

p2edwards (changes requested):

jsapp/js/components/reports/reports.scss
p2edwards: The tilde could possibly be useful if we use it to set an alias for the images folder. That way the "relative path" doesn't have to change when we change the .scss file (or the way it's being imported.)

Issue #3318, opened 2021-06-24
Improve log in guidance


jnm:
When trying to access the `edit` route anonymously, we tell people to "try reloading the page", and then display an infinite loading spinner without any "LOG IN" button:
![image](https://user-images.githubusercontent.com/2085013/123305554-3d426b00-d4ee-11eb-8b52-7f032408f912.png)
![image](https://user-images.githubusercontent.com/2085013/123305572-43384c00-d4ee-11eb-90df-375b8f82ffeb.png)

For the `landing` route, we still say "try reloading the page" (incorrect advice) and display an infinite spinner, but we *do* show the "LOG IN" button, and clicking that does let the user authenticate and be redirected to the place where they were trying to go originally.
![image](https://user-images.githubusercontent.com/2085013/123305593-4a5f5a00-d4ee-11eb-8208-1efb48a6deb1.png)

Requests:

- [ ] No infinite loading spinners :smiling_imp:
- [ ] Show the "LOG IN" button when needed
- [ ] When appropriate, advise to use the "LOG IN" button instead of reloading the page (I think we do this in some contexts already?)

jnm:
Another one to test: https://kobo.humanitarianresponse.info/#/account-settings shows an infinite loading spinner, with no informational modal, but it does include the "LOG IN" button

jnm:
Another infinite loading spinner:
1. be authenticated
1. go to a form you can't access (e.g. https://kf.kobotoolbox.org/#/forms/agNZGnGtHbpH9hbQPjn6rQ/landing)
1. spinny spin

magicznyleszek:
Note to self: see if previous PR in this area has some useful knowledge: https://github.com/kobotoolbox/kpi/pull/3255

PR #3390, opened 2021-08-02
Cleanup pre-work for login guidance improvement

magicznyleszek:
## Description

Cleanup routes related files as a pre-work for #3318

## Related issues

Part of #3318


jsapp/js/stores.es6
```diff
@@ -216,30 +216,36 @@ stores.session = Reflux.createStore({
...
-      currentAccount: acct,
-    });
+    this.currentAccount = account;
+    this.trigger();
+  },
+
+  onNotLoggedIn(data) {
+    log('login confirmed anonymous', data.message);
+    this.isAuthStateKnown = true;
+    this.trigger();
```
p2edwards, line 244: (this `this.trigger()` was a fix, confirmed on call)

p2edwards (approved):
per video call, thanks for looking at this with me!

PR #3409, opened 2021-08-12
Make authentication required more obvious in permissions UI

### LINKED ISSUE #2620: Make users know authentication is required when using row level permission
## Description

Without "Require authentication…" account option being enabled, the partial permissions sharing options don't work in a way that users would expect.

Let's do this In Sharing Form:

1. Let's check if user has "Require authentication…" disabled, and if so, let's add some warning message:
   > "Anyone can see this blank form and add submissions because you have not set your account [link to settings route] to require authentication"
2. If "Require authentication…` is disabled, then make "Add submissions" always checked and disabled

## Additional details

As per discussed in [this thread](https://www.flowdock.com/app/kobotoolbox/ux-ui/threads/dMEFGTWfLOfl8FFxdDqXOGWGiXd), users may not read [this article](https://support.kobotoolbox.org/en/articles/3345421-row-level-permissions) and skip the important step to check `Require authentication to see forms and submit data` in their account settings.

magicznyleszek:
## Description

When "Authentication required" is not enabled, we display a warning message in "Share project" modal. We also make "Add submissions" always checked (and by the permissions rules "View form" too).

## Additional details

Changes:

Screenshot 2021-08-12 at 08 11 44

## Related issues

Fixes #2620

### DIFF
--- a/jsapp/js/components/permissions/sharingForm.es6
+++ b/jsapp/js/components/permissions/sharingForm.es6
@@ -10,6 +10,7 @@ import {bem} from 'js/bem';
 import {LoadingSpinner} from 'js/ui';
 import {buildUserUrl} from 'utils';
 import {
+  ROUTES,
   ASSET_TYPES,
   ANON_USERNAME,
 } from 'js/constants';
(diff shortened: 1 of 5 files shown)


jnm:
@magicznyleszek I rebased this onto the release branch and solved a few conflicts, but I saved the original branch to https://github.com/kobotoolbox/kpi/tree/2620-auth-required-help__backup in case that makes it easier to get this change into `beta` later.

jnm (approved):

PR #3412, opened 2021-08-12
Fix validation logic mangling bug

### LINKED ISSUE #466: regex with double quotes is stripped on saving form
When using double quotes in manual validation/constraint the formbuilder strips all parts of the code starting with the first double quote. 

To reproduce: 
1. Start form with a text question
1. Add this condition as a manual constraint: `regex ( ., "^[0-9]{10}")`
1. Save the form and reopen (or open in json view)

Expected: The condition is saved as `regex ( ., "^[0-9]{10}")`

Instead: The condition only remains as `regex ( .,`

magicznyleszek:
## Description

Fixes the problem for validation logic with double quotes being mangled on save.

## Related issues

Fixes #466


duvld:
is the branch `466-cleanup-regex-quotes` still being used?

jsapp/xlform/src/mv.validationLogicHelpers.coffee
```diff
@@ -98,11 +98,16 @@ module.exports = do ->
...
-            <input type="text" name="constraint" id="#{@context.helper_factory.current_question.cid}-handcode" class="text" value="#{@criteria}">
+            <input type="text" name="constraint" id="#{@context.helper_factory.current_question.cid}-handcode" class="text" value="">
           </span>
         </div>
       """)
       @textarea = @$handCode.find('#' + @context.helper_factory.current_question.cid + '-handcode')
+      # Textarea starts with empty value - setting it in HTML caused error with
+      # mangled values (due to " character). Set initial value here, so all the
+      # special characters are saved.
+      @textarea.val(@criteria)
```
duvld, line 109: Curious what is the difference between `value=` in the HTML element vs javascript value setting?

jsapp/xlform/src/view.mandatorySetting.coffee
```diff
@@ -21,6 +21,13 @@ module.exports = do ->
     render: ->
       reqVal = @getChangedValue()
       template = $($viewTemplates.$$render("row.mandatorySettingSelector", "required_#{@model.cid}", reqVal))
+
```
duvld, line 24: Slightly off topic, but after searching for the cause of the issue I'm wondering how does the Validation Criteria get rendered differently from Error Message? Both are text fields but Error Message's text with quotes is maintained. I'm looking at:
https://github.com/kobotoolbox/kpi/blob/6e0cfa313b5a57145075f4696bc0c19202a83df3/jsapp/xlform/src/view.rowDetail.coffee#L202-L209

magicznyleszek:
> is the branch `466-cleanup-regex-quotes` still being used?

There's a PR: https://github.com/kobotoolbox/kpi/pull/3411 :)

jsapp/xlform/src/mv.validationLogicHelpers.coffee
magicznyleszek: The problem is not with the `value=` itself, but by the fact that the whole HTML code for this is being set with this long string (`"""`) and there JS is first putting the whole string together, and when it gets to `value="regex"with"quotes"` it recognizes `value="regex"` as attribute with value and then the `with"quotes"` is being ommited as not a valid attribute

magicznyleszek: When using JS to set value, we are operating on JS HTML object, and the value setting functions have a safe way of storing `"` :)

jsapp/xlform/src/view.mandatorySetting.coffee
magicznyleszek: I did not follow up the code to the root, but I think that the difference is that the error message value is being set in a proper JS way, and the validation criteria (before the fix) was trying to put the value in the string. And I don't see that the value is being forwarded with the creation of the error message textbox, so I'm guessing it is being set somewhere else after rendering

duvld (approved):

PR #3314, opened 2021-06-23
[NLP] Allow previewing of submisson media from data table

duvld:
## Description

If submission contains media (specifically audio/background-audio, video or images) then it is now possible to view the media directly from the table view in a pop-up modal.

## Related issues
Fixes #3089

PR #3418, opened 2021-08-16
[NLP] Text preview modal

### LINKED ISSUE #3386: [NLP] Text preview modal
## Description

Similarly to the Table Media Preview, we need a way to open a modal for long text answers. Ultimately it will be possible to either translate those or add coding in the upcoming NLP feature.

duvld:
## Description

Extends #3314 to also have a preview modal for text based questions.

## Related issues
Includes code from #3419 

Fixes #3386


jsapp/js/components/submissions/mediaCell.scss
```diff
@@ -1,6 +1,6 @@
  @import "scss/_colors";
 
- .media-cell {
+.media-cell {
   text-align: center;
```
magicznyleszek, line 4: I think we need to switch this to `left` for the icon + text case. Also the icon button hover state background shape is not perfect, see:

Screenshot 2021-08-18 at 23 08 45

jsapp/js/components/common/modal.scss
```diff
@@ -415,6 +423,11 @@ $z-modal-x: 10;
         margin-right: 12px;
       }
 
+      .table-media-preview__text {
+        text-align: left;
+        padding: 0 32px 32px 32px;
```
magicznyleszek, line 428: We should either force wrap long words, or add a right padding that works, see:

Screenshot 2021-08-18 at 23 17 49

jsapp/js/components/submissions/mediaCell.es6
```diff
@@ -56,25 +95,43 @@ class MediaCell extends React.Component {
...
+
+          <a
+            className='kobo-light-button kobo-light-button--gray'
+            // TODO: point this to submission processing modal
+            href={'#'}
+          >
+            {t('process')}
+
+            <i className='k-icon k-icon-arrow-up'
+              //TODO: Need to add icon for top-right pointing arrow
```
magicznyleszek, line 124: I think we should comment out the button for now, as we are merging to `beta` - or switch base branch to `feature/nlp`

magicznyleszek: BTW. ~~do you need the icon right now?~~ I've just merged the icon you need to `beta` :) Try `k-icon-arrow-up-right`

magicznyleszek (changes requested):

jsapp/js/components/submissions/mediaCell.scss
duvld: I tried the `text-align: left` but I think this extra height is coming from the `line-height` of the text. I got around it by adding a wrapper `div` so that `div` eats the line height instead. Maybe this is a hacky way to do it though :P

jsapp/js/components/submissions/mediaCell.es6
duvld: Great point, I'll comment it out for now but still use the correct icon :D

jsapp/js/components/common/modal.scss
duvld: I was wary of bad breaks but I didn't know that your browser would auto-hyphen words it thinks belong to the dictionary. Even some of the longest words in English are hyphened by the browser:
image

Hopefully not many submissions have 1000+ character words :P If the submission contains a typo such as a long sentence without spaces it would just arbitrarily cut the 'word' though

magicznyleszek (approved):

PR #3421, opened 2021-08-18
allow mousewheel on map

sgigh:
## Description

Allow scrolling on map for best navigation

### DIFF
--- a/jsapp/js/components/map.es6
+++ b/jsapp/js/components/map.es6
@@ -118,7 +118,7 @@ export class FormMap extends React.Component {
 
     var map = L.map('data-map', {
       maxZoom: 17,
-      scrollWheelZoom: false,
+      scrollWheelZoom: true,
       preferCanvas: true
     });
 


jnm:
Thank you, @sgigh.

@Ig-Rebollo, I'm not sure why this was disabled in the first place (but it looks like that was done intentionally). Do you have any position on whether or not turning the scroll wheel should zoom the map?

Ig-Rebollo:
@jnm @sgigh generally speaking, scroll blocking in maps is a standard practice to ensure the scrolling of the map doesn't interfere with the scrolling of the site, particularly for tablets, mobile devices, etc. Not doing so might get users 'stuck' in the map. 

This does not apply so much to our case right now, since the site is mostly focused on desktop and there's nothing else to scroll in the UI as long as it is not opened on a mobile device or similar.

Of course, this opens a whole other discussion about responsiveness in KoBo and how much we can adapt to multiple devices. Following UX best practices, we could enable scrolling in the map for desktop/laptop screens, but we should then block it (or find an alternative approach like the two finger scrolling) for mobile devices and tablets.

magicznyleszek:
This will enable mousewheel for all devices (we want to only enable for desktop). I've created a task for it to not lose this idea: https://www.notion.so/kobotoolbox/Allow-mousewheel-on-map-on-desktop-18b7e515f6548000a55fc19611c27b6e?pvs=4 and am closing it.

PR #3435, opened 2021-08-23
Improve types for question types enum

magicznyleszek:
## Description

Internal types improvement for QESTION_TYPES and ASSET_TYPES.


jsapp/js/constants.ts
```diff
@@ -169,25 +169,45 @@ export const VALIDATION_STATUSES_LIST = [
...
+  collection = 'collection',
+}
+
+interface AssetTypeDefinition {
+  id: AssetTypeName
+  label: string
+}
+
+type AssetTypes = {
+  [P in AssetTypeName]: AssetTypeDefinition
```
duvld, line 189: Is the purpose of `[P in AssetTypeName]` to ensure that each attribute in the `AssetTypeName` enum has a definition?

jsapp/js/constants.ts
```diff
@@ -218,46 +280,66 @@ export const ASSET_FILE_TYPES = {
...
+  MetaQuestionTypeName.end,
+  MetaQuestionTypeName.today,
+  MetaQuestionTypeName.username,
+  MetaQuestionTypeName.simserial,
+  MetaQuestionTypeName.subscriberid,
+  MetaQuestionTypeName.deviceid,
+  MetaQuestionTypeName.phonenumber,
+  MetaQuestionTypeName.audit,
+  MetaQuestionTypeName['background-audio'],
+]) as {[P in MetaQuestionTypeName]: MetaQuestionTypeName};
```
duvld, line 342: Is this setting each attribute in `MetaQuestionTypeName` to the entire type? I'm a bit confused by this and `GroupTypeBeginName`

magicznyleszek: I didn't want to get rid of all those `SOMETHING = createEnum()` right now, but ultimately they should become `enum`s.

`createEnum` function is returning `{[val: string]: string}` which means it's an object with string properties and values. Using `createEnum` as is doesn't allow for the created enum to be more meaningful. The `SOMETHING = createEnum(…) as SomeType` is a way of overriding the returned type. It's not 100% nice TypeScript, but it's good way of going half way.

duvld (approved):

PR #3452, opened 2021-09-02
New bem utility

magicznyleszek:
## Description

A followup TypeScriptization of `bem` utility. It turned out that the previous code was too complex to create types for them and too complex for our usage. I wrote it from scratch with a bit changed usage.

### Files

Previously we had `bem.ts` file with multiple bem components being defined in it, it was also exporting something like an alias for `reactCreateBemElement.es6`. Messy. Now all the components definitions are in `bemComponents.ts`, and you probably should never import the file, as all definitions are stored in `bem` from `bem.ts`. The chaotic neutral file `reactCreateBemElement.es6` is no more, and bem components creation comes from a simplified `bem.ts` now.

### Usage

Previously:

```tsx
import bem from 'js/bem'

bem.FooBar = bem.create('foo-bar', 'section')
bem.FooBar__FumBaz = bem.FooBar.__('fum-baz', 'input')

function render() {
  return <bem.FooBar><bem.FooBar__FumBaz value='test'/></bem.FooBar>
}
```

Now:

```tsx
import bem, {makeBem} from 'js/bem'

bem.FooBar = makeBem(null, 'foo-bar', 'section')
bem.FooBar__FumBaz = makeBem(bem.FooBar, 'fum-baz', 'input')

function render() {
  return <bem.FooBar><bem.FooBar__FumBaz value='test'/></bem.FooBar>
}
```

There is no difference in `JSX` usage. The difference is when creating bem components, we pass parent to `makeBem` funcion and we store (optionally) created elements in `bem` object.

NOTE: we no longer support passing tags with `<` characters (so don't use: `'<input>'`, use `'input'` only)


jsapp/js/bem.ts
```diff
@@ -1,382 +1,103 @@
...
+        if (propEntry[0] !== 'm' && propEntry[0] !== 'className') {
+          newProps[propEntry[0]] = propEntry[1];
+        }
+      });
+
+      return React.createElement(htmlTagName, newProps);
+    }
+  }
+
+  return BemComponent
```
duvld, line 102: Is the class definition being returned here? Or an instance of the class? To me it looks like this function reads:
```
function(...) {
  class Foo {...};
  return Foo;
}
```
and I'm a little confused

duvld: Besides these, this method of bem was :sparkles: clean to read through :+1:

jsapp/js/bem.ts
```diff
@@ -1,382 +1,103 @@
...
+        // Case 1: string
+        classNames.push(`${wholeName}--${this.props.m}`);
+      } else if (Array.isArray(this.props.m)) {
+        // Case 2: array
+        this.props.m.forEach((modifier) => {
+          if (typeof modifier === 'string' && modifier.length >= 1) {
+            classNames.push(`${wholeName}--${modifier}`);
+          }
+        });
+      } else if (typeof this.props.m === 'object') {
```
duvld, line 77: Should we make this a custom type? This reads like we're expecting each entry to have two elements, and if this was something like `{'fb'}` would this result in class names `wholeName__f` and  `wholeName__b`?

duvld:
Besides these, this method of bem was :sparkles: clean to read through :+1:

jsapp/js/bem.ts
magicznyleszek: This returns a class, not an instance. Or speaking differently it returns a component, i.e. the `bem.TextBox` is a React component. So when calling `makeBem()` you are making a new class/component to be used later. And when writing JSX code `<bem.TextBox>…</bem.TextBox>` React takes the provided class and renders it.

magicznyleszek: So it's:

```
function(someParams) {
  class Foo {/* configure it using `someParams` making the returned `Foo` unique */};
  return Foo;
}
```

jsapp/js/bem.ts
magicznyleszek: So the `m` property has these types:

```ts
type BemModifiersObject = {
  [modifierName: string]: boolean
}

interface BemComponentProps extends React.ComponentProps<any> {
  m?: string | string[] | BemModifiersObject
}
```

magicznyleszek: And passing `null` crashes it because it is not a valid value for `m`. Unfortunatly it's not showing up because `editableForm` is not TypeScript yet.

magicznyleszek: If I make this simple code:

```ts
import React from 'react';
import bem, {makeBem} from 'js/bem';

bem.FooBar = makeBem(null, 'foo-bar');

function foo() {
  const isAsideVisible = true;

  return (
    <bem.FooBar m={isAsideVisible ? 'visible' : null} />
  )
}
```

magicznyleszek: I get TypeScript error: `Type 'string | null' is not assignable to type 'string | string[] | BemModifiersObject | undefined'. Type 'null' is not assignable to type 'string | string[] | BemModifiersObject | undefined'.`

magicznyleszek: There is a useful thing in TypeScript - it can understand JSDoc comments and even display them. So I've improved the definition of properties to this:

```ts
interface BemComponentProps extends React.ComponentProps<any> {
  /**
   * Pass a string, or array of strings. For complex modifiers, pass an object:
   * {
   *   'modifier-name': <boolean value whether the modifier should be applied>,
   *   'another-modifier': <boolean>,
   * }
   */
  m?: string | string[] | BemModifiersObject
}
```

And now whenever I hover over `m` in my editor, I see this:

Screenshot 2021-09-14 at 16 51 18

jsapp/js/bem.ts
duvld: Gotcha! well explained :)

duvld (approved):

PR #3439, opened 2021-08-23
Typescriptize assetUtils file

magicznyleszek:
## Description

Adds types to assetUtils file.

## Related issues

Contains the work from these PRs:

- #3429
- #3435

PR #3458, opened 2021-09-10
Typescriptize submissionUtils

magicznyleszek:
## Description

Move `submissionUtils` file to TypeScript.

## Related issues

Contains code from PRs:
- #3439 (and thus also some other PRs)


jsapp/js/components/submissions/submissionUtils.ts
```diff
@@ -12,35 +12,49 @@ import {
...
- */
 class DisplayGroup {
-  constructor(type, label = null, name = null) {
+  public type: DisplayGroupTypeName;
+  /** Localized display label */
+  public label: string | null = null;
+  /** Unique identifier */
+  public name: string | null = null;
+  /** List of groups and responses */
+  public children: Array<DisplayResponse|DisplayGroup> = [];
```
duvld, line 41: Just curious of the syntactic difference between this and `(DisplayResponse | DisplayGroup)[]`?

jsapp/js/components/submissions/submissionUtils.ts
```diff
@@ -338,12 +388,12 @@ function getRowData(name, survey, data) {
...
- * @param {string|null} groupPath - null for root level rows
- * @param {Array<object>} survey
- * @returns {boolean}
  */
-function isRowFromCurrentGroupLevel(rowName, groupPath, survey) {
+function isRowFromCurrentGroupLevel(
+  rowName: string,
+  /** Null for root level rows. */
+  groupPath: string|null,
+  survey: SurveyRow[]) {
```
duvld, line 396: Pretty trivial but is this `)` supposed to be there (instead of on the next line) or a missed linting?

jsapp/js/components/submissions/submissionUtils.ts
```diff
@@ -396,17 +449,46 @@ function getRegularGroupAnswers(data, targetKey) {
...
-    row[CHOICE_LISTS.SELECT] ||
-    row[CHOICE_LISTS.MATRIX] ||
-    row[CHOICE_LISTS.SCORE] ||
-    row[CHOICE_LISTS.RANK]
+export function getMediaAttachment(
+  submission: SubmissionResponse,
+  fileName: string
+): string | SubmissionAttachment {
+  const fileNameNoSpaces = fileName.replace(/ /g, '_');
+  let mediaAttachment: string | SubmissionAttachment = t('Could not find ##fileName##').replace(
```
duvld, line 480: I wonder if it would be better practice to do something like:
```
const attachmentNotFound = t('Could not find ##fileName##').replace(...)
let mediaAttachment: SubmissionAttachment
...
if (mediaAttachment) {...} else return attachmentNotFound
```
to ensure that `mediaAttachment`'s typing? Not a big deal, more curious :D

jsapp/js/dataInterface.d.ts
```diff
@@ -7,6 +7,45 @@ interface FailResponse {
   statusText: string
 }
 
+interface SubmissionAttachment {
```
duvld, line 10: I'm wondering what your process is for finding all the possible attributes for backend responses? Do you scan the code and see what we use or are you checking somewhere in the backend/api?

jsapp/js/dataInterface.d.ts
```diff
@@ -7,6 +7,45 @@ interface FailResponse {
...
+  filename: string
+  instance: number
+  xform: number
+  id: number
+}
+
+interface SubmissionResponse {
+  [questionName: string]: any
+  __version__: string
+  _attachments: SubmissionAttachment[]
```
duvld, line 25: Found a bug, at this line:
https://github.com/kobotoolbox/kpi/blob/80fbec7eab5bc9d16f97e662a56c43be6a64bf67/jsapp/js/components/submissions/table.es6#L944

it has to do with the `row.original` having `_attachments` rather than `row` itself. I think we should change `getMediaAttachment` param to an object having `{_attachment: any}` instead of `SubmissionResponse`, as in this case it would not be that type

I commited a fix here: https://github.com/kobotoolbox/kpi/pull/3458/commits/3e72eaa8d5023b68235c735135d806582066cbc4

jsapp/js/components/submissions/submissionUtils.ts
magicznyleszek: So I found this post from 2017 https://mattferderer.com/difference-between-array-types-in-typescript after a quick look. I didn't find any good article that covers all of the ways, but I think this is the same :)

jsapp/js/components/submissions/submissionUtils.ts
magicznyleszek: Good catch, supposed to be on new line :)

jsapp/js/components/submissions/submissionUtils.ts
magicznyleszek: I tried doing that now, but it turned out I need to do some other stuff around to make it work, so I think it's ok to leave it as is.

jsapp/js/dataInterface.d.ts
magicznyleszek: I'm looking at the endpoint responses and build the `interface`s with that. Then after I write some more TypeScript it turns out some stuff is missing and I add it.

jsapp/js/dataInterface.d.ts
magicznyleszek: Thanks for the fix!

magicznyleszek: > I think we should change `getMediaAttachment` param to an object having `{_attachment: any}` instead of `SubmissionResponse,` as in this case it would not be that type
I don't think I understand - in all two of the `getMediaAttachment` calls we are passing the `row.original`, which is `SubmissionResponse` in both cases

jsapp/js/dataInterface.d.ts
duvld: Makes sense, that's what I had in mind as well :+1:

jsapp/js/dataInterface.d.ts
duvld: > > I think we should change `getMediaAttachment` param to an object having `{_attachment: any}` instead of `SubmissionResponse,` as in this case it would not be that type
> 
> I don't think I understand - in all two of the `getMediaAttachment` calls we are passing the `row.original`, which is `SubmissionResponse` in both cases

You're right. I didn't notice `row.original` was the submission response. I confused it with just `row` :P

PR #3465, opened 2021-09-14
Preserve Paired Data XPath consistency

HaidarZ:
update(paired-data): replace 'data' node by 'item' node to preserve pattern and integrate better with pyxform.

## Description

We’re used to the following syntax for the calculated fields , checkout the query nodes
```
count(instance('survey')/root/item[P_Enumerator])	
```

While just inside Dynamic Data Attachments the rule is changed to use the below
```
count(instance('survey')/root/data[P_Enumerator])	
```

One of the use cases is when you try to integrate data pulled from another form into any widget.

## Related issues

Checkout the discussion thread on the community forum,

https://community.kobotoolbox.org/t/dynamic-data-attachments-calculation-syntax-enhancement/22580

## Note

The following [docs](https://support.kobotoolbox.org/dynamic_data_attachment.html) must be updated if we elevated the PR.
All the `root/data` expressions must be replaced by `root/item`

### DIFF
--- a/kpi/views/v2/paired_data.py
+++ b/kpi/views/v2/paired_data.py
@@ -254,9 +254,11 @@ class PairedDataViewset(AssetNestedObjectViewsetMixin,
         parsed_submissions = []
 
         for submission in submissions:
-            # Use `rename_root_node_to='data'` to rename the root node of each
-            # submission to `data` so that form authors do not have to rewrite
+            # Use `rename_root_node_to='item'` to rename the root node of each
+            # submission to `item` so that form authors do not have to rewrite
             # their `xml-external` formulas any time the asset UID changes,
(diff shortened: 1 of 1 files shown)


jnm:
Pending discussion: https://community.kobotoolbox.org/t/dynamic-data-attachments-calculation-syntax-enhancement/22580/7?u=jnm

jnm:
I believe the use of `data` to be consistent with XForm conventions (see [discussion](https://community.kobotoolbox.org/t/dynamic-data-attachments-calculation-syntax-enhancement/22580/7?u=jnm) also linked above) and therefore decline to merge the change at this time.

I do want to express my gratitude to @HaidarZ for their interest in improving KoBoToolbox. Thank you!

PR #3469, opened 2021-09-15
Make BEM modifiers more flexible

### LINKED ISSUE #3467: Unsaved changes indicator in Form Builder no longer works
## Description

The unsaved changes indicator (`*` in "SAVE" button) is no longer showing up when editing a form in Form Builder. This happens on `beta` branch.

duvld:
## Description

With the merging of #3452 we have a revamped way to use BEM in KPI. There was a missed spot for passing bem modifiers as a array that is a combination of strings and objects, namely here:
https://github.com/kobotoolbox/kpi/blob/ae5fe4518a6eba11957f5ede09ab7e4704009205/jsapp/js/editorMixins/editableForm.es6#L624-L628

PR allows this type though. For future we should maybe figure out all allowed cases/bad cases in the `es6` code

Fixes #3467

### DIFF
--- a/jsapp/js/bem.ts
+++ b/jsapp/js/bem.ts
@@ -1,4 +1,5 @@
-import React from 'react';
+import React from 'react'
+import classnames from 'classnames'
 
 /**
  * USAGE:
@@ -17,7 +18,7 @@ interface bemInstances {
  */
 const bem: bemInstances = {}
 
-export default bem;
+export default bem
 
 type BemModifiersObject = {
   [modifierName: string]: boolean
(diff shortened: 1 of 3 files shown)


dorey:
This functionality was crafted around the npm package `classnames` which builds a list of classnames from an arbitrary list of strings and things.

https://github.com/JedWatson/classnames#readme

```javascript
classNames('foo', 'bar'); // => 'foo bar'
classNames('foo', { bar: true }); // => 'foo bar'
classNames({ 'foo-bar': true }); // => 'foo-bar'
classNames({ 'foo-bar': false }); // => ''
classNames({ foo: true }, { bar: true }); // => 'foo bar'
classNames({ foo: true, bar: true }); // => 'foo bar'
```

magicznyleszek (approved):

PR #3475, opened 2021-09-20
Declare `classnames` module and assume return type

duvld:
## Description

After #3469 was merged, the new module was not declared and it's return type is not specified by the module. See errors below
```
ERROR in /srv/src/kpi/jsapp/js/bem.ts
./jsapp/js/bem.ts 2:23-35
[tsl] ERROR in /srv/src/kpi/jsapp/js/bem.ts(2,24)
      TS7016: Could not find a declaration file for module 'classnames'. '/srv/src/kpi/node_modules/classnames/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/classnames` if it exists or add a new declaration (.d.ts) file containing `declare module 'classnames';`

ERROR in /srv/src/kpi/jsapp/js/bem.ts
./jsapp/js/bem.ts 76:40-48
[tsl] ERROR in /srv/src/kpi/jsapp/js/bem.ts(76,41)
      TS7006: Parameter 'modifier' implicitly has an 'any' type.
ℹ ｢wdm｣: Failed to compile.
```
from their `npm` page: https://www.npmjs.com/package/classnames it looks like it's safe to assume the return will always be `string`.

### DIFF
--- a/jsapp/js/bem.ts
+++ b/jsapp/js/bem.ts
@@ -72,7 +72,7 @@ export function makeBem(
       const wholeName = parent ? `${parent.blockName}__${name}` : name
       classNames.push(wholeName)
 
-      const modifiersList = classnames(this.props.m)
+      const modifiersList: string = classnames(this.props.m)
       modifiersList.split(' ').forEach((modifier) => {
         classNames.push(`${wholeName}--${modifier}`)
       })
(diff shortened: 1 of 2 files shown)


magicznyleszek:
@duvld actually the fix should be this commit: https://github.com/kobotoolbox/kpi/commit/1d5fc37ecec11094709f74d0fc1331d4600e7f43 - could you please try the `beta` branch again making sure you did `npm install`? IIUC the version of `classnames` we had (`2.2.6`) did not have the types in it yet, but the've added it in the one we are using (latest one, i.e. `2.3.1`)

Issue #3097, opened 2021-03-23
[NLP] Usage and Billing Dashboard


magicznyleszek:
## Description

This would be a basic dashboard initially with:
- the usage and billing account section
- a rework of Account views
- submissions per month and GB of storage (needs backend endpoint)

PR #3485, opened 2021-09-27
Typescriptize KoboDropdown and KoboDropdownActions

duvld:
## Description
Typescriptize the `KoboDropdown` component. Needed for #3097


jsapp/js/components/common/koboDropdown.tsx
```diff
@@ -1,27 +1,52 @@
...
+  'up-center' |
+  'up-right' |
+  'down-left' |
+  'down-center' |
+  'down-right'
+
+
+type KoboDropdownProps = {
+  theme: KoboDropdownThemesType,
+  placement: KoboDropdownPlacementsType,
```
magicznyleszek, line 37: You don't need `KoboDropdownThemesType` and `KoboDropdownPlacementsType` - you can use enum here. Enums in TypeScript are both a data to be used in code and a type. See `textBox.tsx` for identical situation :)

jsapp/js/components/common/koboDropdown.tsx
```diff
@@ -1,27 +1,52 @@
...
   KEY_CODES,
 } from 'js/constants';
 import koboDropdownActions from './koboDropdownActions';
 import './koboDropdown.scss';
 
-export const KOBO_DROPDOWN_THEMES = createEnum([
-  'light',
-  'dark',
-]);
+export enum KOBO_DROPDOWN_THEMES {
```
magicznyleszek, line 10: I am wondering how should we name the `enum`s we create in TS. For now the `const`s we had followed the `SOME_NAME` idea, so it's clear that we have constant and enum-ish thing. But with TS we can clearly see what the thing is, as we see the definition whenever we use it. What do you think about just using `SomeName`?

magicznyleszek: I also initially went with naming the enums as singular not plural, but I think we should keep going with plural 😅

jsapp/js/components/common/koboDropdown.tsx
```diff
@@ -1,27 +1,52 @@
...
+
+type KoboDropdownProps = {
+  theme: KoboDropdownThemesType,
+  placement: KoboDropdownPlacementsType,
+  isDisabled: boolean, // disables the dropdowns trigger, thus disallowing opening dropdown
+  hideOnMenuClick: boolean, // hides menu whenever user clicks inside it, useful for simple menu with a list of actions
+  hideOnMenuOutsideClick: boolean, // hides menu when user clicks outside it
+  hideOnEsc: boolean, // hides menu when opened and user uses Escape key
+  triggerContent: React.ReactNode,
+  menuContent: React.ReactNode, // the content of dropdown, anything's allowed
```
magicznyleszek, line 43: I discovered recently that we can use JSDoc for even tiniest part of code for TS to display them. So please change all those to:

```ts
/** the content of dropdown, anything's allowed */
menuContent: React.ReactNode,
```

magicznyleszek (changes requested):

jsapp/js/components/common/koboDropdown.tsx
```diff
@@ -1,27 +1,52 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
```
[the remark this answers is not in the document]
duvld: Should be good now!

jsapp/js/components/common/koboDropdown.tsx
duvld: I was kind of confused by this too. I was thinking of just using `SomeName` for `enums` but it looks like syntax that should be used for types, so it felt a bit odd accessing attributes like `KoboDropdownThemes.light` versus `KOBO_DROPDOWN_THEMES.light`

Although, with what you said in the above comment:
> Enums in TypeScript are both a data to be used in code and a type

makes me feel like we should go with that after all, as it can double as a type and accessed

duvld: Actually if we're following `textBox.tsx` as an example this is already the case :P

magicznyleszek (approved):

PR #3487, opened 2021-09-29
Fix unnecessary table reloads

magicznyleszek:
## Description

Table View was fetching the data unnecessarily when a modal was being opened other than Single Submission Modal.

## Additional details

Current code was not strict enough as to what type of modal changes it was reacting to. It should only react to "submission" type modal - and that is the fix.


p2edwards (approved):
Looks great! 

Tested: 
- Before the change, clicking the "expand arrow" for a text question generated an additional query.
- After this change, clicking the "expand arrow" for a text question did not generate an additional query. ✅
- Submission modal still works: one network query when the modal is activated, and then one network query each time "next" or "prev" is clicked.

PR #3488, opened 2021-09-29
Single processing view improvements

magicznyleszek:
## Description

Some improvements for Single Processing View after first round of testing.

## Related issues

Fixes #3481

PR #3492, opened 2021-09-30
3491 table view improvements

### LINKED ISSUE #3491: Updates to the Table View to fit more data in a normal screen
The size of rows, columns, cells and fonts in the table view seems to be excessive for some users, as they don't see many entries/responses compared to the previous table or even to similar tools (Airtable, Excel...)

**Updates and fixes:**

- [x] Reduce font size to 13px, as it was before
- [x] Reduce height of rows from the current 42px to 36px (I believe this will still provide enough space to breathe as long as the font size is reduced to 13px)
- [x] Use three letter abbreviations for months in dates (`September` becomes `Sep`)
- [x] Reduce default width of `Start` and `End` to 120px
- [x] Change 'Validation Status' to simply 'Validation' since the full name doesn't fit in the default space anyways.
- [x] If possible without cutting the 'not approved' label, also reduce the Validation column to 120px, ideally reducing margins. If the label gets cut by doing this, give it the min width needed for the label to fit.
- [x] Reduce the default column width from 160px to 140px
- [x] Align all column headings to the left (for some reason 'Start' and 'End' and centred in their respective columns)
- [x] Stop using monospaced font in table

cc @tinok

magicznyleszek:
## Description

Stylistic improvements to Table View to make it display more data at once.

## Additional details

Two things of the list are done on #3488 - shorter date format and table header text alignment.

## Related issues

Fixes #3491


duvld (commented):
Everything looks good except the `Not Approved` status seems to be cut off from the new Validation column width:
image

though the kerning of the font looks wrong. maybe it's just a missing font package for me?

magicznyleszek:
@duvld I think that your OS + browser duo just renders Roboto in such ugly way :P I added `5px` more space to accomodate for that fact. Also included a few fixes from the other NLP branch.

PR #3494, opened 2021-10-04
Refactor custom input range

duvld:
## Description

Move the custom input and its styles to an extensible component to avoid duplicating work
## Related issues
Needed for #3097


jsapp/js/components/common/koboRange.tsx
```diff
@@ -0,0 +1,124 @@
...
+  /** uses time display for all required values */
+  isTime?: boolean,
+  /** assumes input is clickable and with move to where it is clicked */
+  onChange?: Function,
+  /** defaults to $kobo-blue */
+  color?: string,
+  /** optional string to append to max */
+  maxUnits?: string,
+  /** optional string to append to value */
+  currentUnits?: string,
```
magicznyleszek, line 35: I think we should have just `units` for both places, as it's highly unlikely they would be different

jsapp/js/components/common/koboRange.tsx
```diff
@@ -0,0 +1,124 @@
...
+  'default' =  'default',
+  'warning' = 'warning',
+  'teal' = 'teal',
+}
+
+type KoboRangeProps = {
+  max: number,
+  value: number,
+  /** uses time display for all required values */
+  isTime?: boolean,
```
magicznyleszek, line 27: When I want `isTime`, what kind of `value` is expected? Seconds, minutes? Would be nice to add it to the comment

jsapp/js/components/common/koboRange.tsx
magicznyleszek: Also if I go with `isTime`, would that mean that the unit is known?

jsapp/js/components/common/koboRange.tsx
```diff
@@ -0,0 +1,124 @@
...
+        </bem.KoboRange__values>
+
+        <bem.KoboRange__progress>
+          <bem.KoboRange__input
+            m={this.props?.color ? this.props.color : KoboRangeColours.default}
+            type='range'
+            max={this.props.max}
+            value={this.state.currentValue}
+            disabled={!this.props.onChange}
+            onChange={this.props?.onChange && this.onChange.bind(this)}
```
magicznyleszek, line 118: For some reason this doesn't trigger on my machine. The visual issue it causes is that the audio player track doesn't update when playing audio.

magicznyleszek (changes requested):

PR #3495, opened 2021-10-06
Icon component

### LINKED ISSUE #3477: Make icon component
## Description

The general idea is to:

1. make the code generating icons also generate a constants file (JSON or TS) with a list of icons available
2. create a component that would be used as:
   ```tsx
   import Icon from 'js/components/common/icon'
   <icon name='arrow-up' size='xl' className={['some', 'optional', 'class', 'names']} />
   ```

Thanks to it being TypeScript we get:
- no longer a chance for us to have nonexistent icon being used
- easier to find icons we have with auto complete
- size parameter

magicznyleszek:
## Description

Introduces a reusable icon component.

## Additional details

Used in one place as example. We should migrate slowly to `<Icon>` from `<i>` while working on other things.

## Related issues

Fixes #3477

### DIFF
--- a/jsapp/js/components/common/icon.tsx
+++ b/jsapp/js/components/common/icon.tsx
@@ -1,0 +1,48 @@
+import React from 'react'
+// Using this type ensures we only have existing icon names
+import {IconName} from '../../../fonts/k-icons'
+import './icon.scss'
+
+/**
+ * Check out `icon.scss` file for exact pixel values.
+ */
+export type IconSize = 'xs' | 's' | 'm' | 'l' | 'xl'
+
+const DefaultSize = 's'
+
+type IconProps = {
+  name: IconName
+  size?: IconSize
+  classNames?: string[]
+}
+
+/**
+ * An icon component.
+ */
+class Icon extends React.Component<IconProps, {}> {
+  constructor(props: IconProps){
+    super(props)
+  }
+
+  render() {
+    let classNames: string[] = []
+    if (
+      Array.isArray(this.props.classNames) &&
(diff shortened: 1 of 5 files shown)


jsapp/js/components/common/icon.tsx
```diff
@@ -0,0 +1,48 @@
...
+ * Check out `icon.scss` file for exact pixel values.
+ */
+export type IconSize = 'xs' | 's' | 'm' | 'l' | 'xl'
+
+const DefaultSize = 's'
+
+type IconProps = {
+  name: IconName
+  size?: IconSize
+  classNames?: string[]
```
duvld, line 16: There's a lot of places in the code where we add `onChange` to icons (`mediaCell.es6` for example). Should we add that to this component or  should we just surround each case of this with a wrapper button?

magicznyleszek: Yeah I think this is a better way. A clickable icon just doesn't feel right when I think about it 🤢 😉

magicznyleszek: But of course if we find any property that would be needed, we will be adding it when the moment comes :)

jsapp/js/components/common/icon.tsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react'
+// Using this type ensures we only have existing icon names
+import {IconName} from '../../../fonts/k-icons'
```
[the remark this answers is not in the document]
magicznyleszek: Fixed. Had to add alias for `jsapp` path in two places.

PR #3498, opened 2021-10-11
Fix dropdown memory leak causing console errors

### LINKED ISSUE #3497: Dropdown component crash when leaving route
## Description

When leaving Table View while a dropdown is opened, there will be an error in console (non leathal).

## Steps to Reproduce

1. Go to form's "Data" tab
2. Open dropdown for a given column name
3. With the dropdown opened click on "Summary" tab
4. `Unable to find node on an unmounted component` :bug:

magicznyleszek:
## Description

Leaving Data route with dropdown opened was causing a console error.

## Additional details

The bug was happening because the function passed to `addEventListener` was not being recognized by `removeEventListener` so the listener was not removed, and the Node stops existing after changing route.

## Related issues

Fixes #3497

### DIFF
--- a/jsapp/js/components/common/koboDropdown.tsx
+++ b/jsapp/js/components/common/koboDropdown.tsx
@@ -69,6 +69,8 @@ export default class KoboDropdown extends React.Component<
     this.state = {isMenuVisible: false};
   }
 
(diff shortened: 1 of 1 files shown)


p2edwards (approved):
Looks good to me.

PR #3505, opened 2021-10-14
Use base image compatible with s390x architecture

jnm:
Sometimes it's easier to show a draft PR than write an issue. This can't be merged because it simply swaps the standard base image out for one that's built for s390x. The work will be to identify a proper multi-architecture solution that can replace both.


jnm:
The organization interested in this work stopped responding long ago.

PR #3560, opened 2021-11-18
Fix missing draft navigation

### LINKED ISSUE #3559: Project navigation disappears when returning from formbuilder
## Description
Project navigation (summary/form/data/settings) is missing in a draft project (not deployed) after I return from editing the form in the formbuilder. The navbar reappears if I go back to the project list and then access my draft project again.
image

Also note the white stripe at the bottom of the screen (indicating that the grey background maybe moved up as the navbar is missing)

## Steps to Reproduce
1. Go to a draft project or create a new project
2. Upload XSLForm **or** create it from scratch with formbuilder
3. In the formbuilder, make some changes and save
4. Close formbuilder
5. You are taken to the 'Form' section but can't see the navigation bar

## Expected behavior 
When closing the formbuilder, I return to the 'Form' section but I am able to see 'Summary', 'Form', 'Data' and 'Settings'. Of course since it is not deployed, 'Data' would be greyed out.

## Actual behavior 
You don't see 'Summary', 'Form', 'Data' or 'Settings'

## Additional details
I could not reproduce this in the OCHA server, **but it did happen as well with deployed projects in the HHI server**

magicznyleszek:
## Description

The draft navigation and header editable title were missing for drafts that were just created.

## Additional details

Also removes the unnecessary `assetid` state property, as we can use the `asset.uid`.

## Related issues

Fixes #3559

### DIFF
--- a/jsapp/js/components/formViewTabs.es6
+++ b/jsapp/js/components/formViewTabs.es6
@@ -51,23 +51,19 @@ class FormViewTabs extends Reflux.Component {
   }
 
   componentDidMount() {
+    // On initial load use the possibly stored asset.
+    this.setState({asset: assetStore.getAsset(this.currentAssetID())})
     this.listenTo(assetStore, this.assetLoad);
   }
 
   assetLoad(data) {
-    var assetid = this.currentAssetID();
-    var asset = data[assetid];
-    this.setState(
-      assign({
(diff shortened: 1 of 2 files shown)


noliveleger (approved):
LGTM

PR #3575, opened 2021-12-01
Fix filtering repeat groups in export

### LINKED ISSUE #3324: Exports with filtered questions break repeat groups
xref: https://community.kobotoolbox.org/t/repeat-group-export-missing/19954

## Description

1. Create form with repeat groups
2. Submit some data to it
3. Export to XLS from KPI UI with filtered questions (all questions can be selected)
4. XLS file omits repeat group sheet

joshuaberetta:
## Description

Fix export filtering for repeat groups.

## Related issues

closes #3324

### DIFF
--- a/kpi/tests/test_mock_data_exports.py
+++ b/kpi/tests/test_mock_data_exports.py
@@ -247,28 +247,25 @@ class MockDataExportsBase(TestCase):
                 'schema': '1',
                 'survey': [
                     {
-                        'name': 'person',
+                        'type': 'begin_group',
+                        'name': 'people',
+                        'label': ['People'],
+                    },
+                    {
                         'type': 'begin_repeat',
-                        '$kuid': 'yl4hr30',
+                        'name': 'person',
                         'label': ['person'],
-                        'required': False,
-                        '$autoname': 'person',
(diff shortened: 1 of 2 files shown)


kpi/tests/test_mock_data_exports.py
```diff
@@ -247,28 +247,25 @@ class MockDataExportsBase(TestCase):
                 'schema': '1',
                 'survey': [
                     {
-                        'name': 'person',
+                        'type': 'begin_group',
+                        'name': 'people',
+                        'label': ['People'],
+                    },
+                    {
```
noliveleger, line 254: What about add another form , e.g. 'Simple nested group' and add a test which uses this form explicitly?

joshuaberetta: @noliveleger, this is the test I added in the previous PR that attempted to fix the filtering issue. So this is just extending that test a little further to cover the previous and new fix so I don't think it needs a new test form. I'm happy to add one if you think that's a better way forward though 👍

noliveleger: That's fine

PR #3581, opened 2021-12-04
NLP transcript and translations editors

### LINKED ISSUE #3092: [NLP] Transcription Editor
## Description

Manual transcription editor
- Should display Languages Selector first (see #3376) to set the transcription language
- It should be possible to change the selected language for transcription
- Transcription editor is just a `<textarea>` and a SAVE button (with some indication of unsaved changes)
- Each question can only have one transcription
- What happens with unsaved changes when we switch to Translations editor?
- It should store the data using #3091 
- Display warning if is possible that multiple users would edit in the same time (check permissions if possible)

## Additional details

Requires #3376 and #3091

### LINKED ISSUE #3094: [NLP] Transcription Translation Editor
## Description

Manual translation editor:
- Should display Languages Selector first (see #3376) to set the translation language
- It should be possible to only have one translation per language
  - we probably want to disable selecting an existing language
  - we need to make sure that a custom typed language is not the same as existing language
- It should be possible to change the selected language for given translation
- Translation editor is just a `<textarea>` and a SAVE button (with some indication of unsaved changes)
- Transcripton can have multiple translations
- It should be possible to add further translations - plus also to change the translation language, delete translation 
- It should be possible to switch between translations
  - What happens with unsaved changes? Block switching translations until saved? Confirmation prompt?
- What happens with unsaved changes when we switch to Transcription editor?
- It should store the data using #3091
- Display warning if is possible that multiple users would edit in the same time (check permissions if possible)

## Additional details

Requires #3376 and #3091

### LINKED ISSUE #3567: [NLP] UX/UI fixes to front end (version in kvar)
### In Table View
- [x] Question: How are the sizes of the modal determined? Could/should we make them more consistent?
- [x] In some cases scroll appears for images. See screenshot for example, if size of modal is not always the same, shouldn't the modal adapt to the size of the pictures? Alternatively, if the modal can't adapt, I would suggest to reduce the size of the image to fit it in the max width/height of the modal.
image
- [x] In audio modal, time takes a while to appear, so bar appears higher than it should for a while (see screenshot). It would be better if the whole audio component waited to load when the time is ready, so the whole thing loads simultaneously. 
image
- [x] Both pictures and the audio component can take a bit to load. We should add a loading icon/animation while they are loading. 
image
- [x] Play icon should be smaller (though grey background on hover should remain the size it is now - so smaller icon in the same hover area) 
image
- ~~Once transcripts or translations are saved, they should be added as new columns to the table view. For MVP, no special treatment at all. Just new columns to the right of the audio they came from.~~ duplicate of #3382

### In Single Processing View
- [x] Some text already appears in transcriptions and translations by default. This makes it difficult to test how it would feel in a real situation. Can we remove the default texts?
- [x] When a transcript/translation I just typed is saved (clicking the 'save' button) the editor should lock again, showing the trash can, pen icon, etc. but instead it simply greys out 'save', and forces me to click on 'back'. This is really strange, it should simply take me back when I click 'save'.
- [x] List of languages seems too small and crammed. Can we increase the font size to 14px?
image
- [x] It would also be nice to increase the space between the columns so they spread nicer over the available area: 
image
- [x] Also a bit of separation between rows would make it nicer: 
image
- [x] Background grey colour is too strong, should be bluish grey from mockups (also in the Figma colours file): #F1F4FC
- [x] Underline indicating the page (transcripts, translations or coding) should always be the same width (50px), centered and smaller than the text. 
image
- [x] Header should be updated to latest version. For this go to issue #3570
- [x] Dropdown for languages should be the blue option from the new dropdown component. Go to issue #3568 for more details
- [x] 'Modified November...' should use the next lighter grey in our system, have less weight, and be right next to the language dropdown. It should also say 'last modified ...'
- [ ] #3617
image
- [ ] Reduce the options on the top left corner of the audio to 'Display' and 'Settings'. Go to #3569 for more info.
- [x] For the time being, let's remove groups from the submission information on the side. We'll have to figure out how to display them later on, but for now I think just showing the questions is better.

### In Downloads
Downloads should be done in the downloads section. Issue for this here (to be created)

### LINKED ISSUE #3570: [NLP] Updated header for Single Processing View
After the user testing sessions, we collected a fair amount of feedback related to the header, which led us to update it entirely. 

This new design for the header should be included in the MVP, as much as possible. 

New designs can be found in [this Figma file](https://www.figma.com/file/ssmaK5p4FYSZUq7O7T1RhY/KoBo-Qualitative-Processing-Prototype?node-id=322%3A9453)

### Changes
- [x] Total height of the header is reduced to 60px
- [x] Question becomes a dropdown, which allows you to select other relevant questions (only questions with some type of media that can be processed - maybe for MVP only audios)
- [x] In this dropdown, each question has an icon to its left. This icon shows the type of media (audio, image, video, etc.). These icons should be visible when opening the dropdown, so it's easier to identify the question you're looking for.
- [x] Submission number is the same functionality but styled differently, to have a similar presence as the question.
- [x] Done button is smaller, uses lowercase, and a 1px border (not 2px)

image

### LINKED ISSUE #3662: [NLP] More NLP related UX review fixes
## Description

- [x] special text ("N/A") for submission modal translation without content
- [x] add simple confirmation modals for deletion of transcript and translation
- [x] for translations there should be a suggestion for languages - use translation languages that are already there for other submissions
- when deleting a translation
  - [x] backend should prune the language if there is no other translation for this language (remove the language from advanced_features) - expect this to happen ( @dorey this is the only TODO item for you here)
  - [x] when deleting a translation or transcript - frontend needs to send unicode "delete character"
  - [x] when we delete a translation, we should display the remaining one or the "begin" stage - confirm it works like that
- [x] allow deleting transcript. Also make sure UI works correctly and the initial flow of requiring transcript before making translations works
- moved to #3676
- [x] fix transcript/translation column name in table view
- [x] we should store languages in `asset.advanced_features.transcript` the same way as with `.translated` and use these languages in suggested languages (same way as for translations)
- [x] display separate columns for different transcript languages
- moved to #3677
- [x] make sure that additional columns are placed next to source questions, now it seems it is broken
- [x] display transcript and translation text as multiline

### LINKED ISSUE #3688: [NLP] LanguageSelector should allow custom languages as optional parameter
For NLP we want to force users to either use our predefined list of languages (these are the only ones that give users a chance of using all NLP features) or contact us about a missing one. We don't want to allow using custom languages. 

But we want to have option in LanguageSelector to use custom language elsewhere in future. 

Let's add a parameter `isCustomAllowed` that will unlock the feature. It might be also good to check if preselected language is valid too.

We should also bring back the "I cannot find my language" button and make a simple article at support website describing the situation and steps to be taken.

### LINKED ISSUE #3738: [NLP] Add "process" button to Single Submission Modal
## Description

Similarly to what we have in Table Media Preview modal, we should display same button here in Single Submission Modal:

image

magicznyleszek:
## Description

This adds a transcript and translations editors to the NLP interface.

## In depth description

Processing is adding additional submission data for question responses. We plan to have processing for audio, video and text question types. We will start with audio only.

Processing is database heavy, so assets need to be have the feature activated first. Activations requires providing a lists of question names and a lists of language codes - this means that asset might be re-activated multiple times, e.g. when a new translation language is added. Backend handles un-activation automagically - e.g. when you delete last translation for given language.

The feature code directory is `jsapp/js/components/processing`. The part we are developing now is related to "single processing" - meaning a processing of single thing, single submission. We are planning to have batch processing too. This influenced the file/component names here. I thought that `processingActions` will have code for both single/batch processing. Now I wonder if it wouldn't be better to make `jsapp/js/components/batchProcessing` and don't merge these two too much.

`singleProcessingRoute.tsx`
- provides the code with `assetUid`, `questionName` and `submissionUuid` through route parameters
- url is `/forms/:uid/data/processing/:questionName/:submissionUuid`
- uses `WorkProtector` component to avoid losing unsaved work
- displays spinners until data (i.e. `singleProcessingStore`) is ready

`processingActions.ts`
- a set of Reflux actions for calling API:
- some of actions are doing chain calls to make sure asset is activated (for given language)

`singleProcessingStore.ts`
- stores all data necessary for rendering the route and all its features
- handles draft transcript and draft translation
- keeps current processing data (from backend)
- handles switching content tabs (Transcript, Translation, and Analysis)
- the only file for now that is using `processingActions` 

`processingBody.ts`
- a BEM component for displaying transcript/translation

`singleProcessingContent.tsx`
- displays three tabs (Transcript, Translation, and Analysis) and their content
- note: Analysis is not part of MVP

`singleProcessingHeader.tsx`
- allows switching to different question or submission to be processed

`singleProcessingPreview.tsx`
- displayed on the aside
- shows the source transcript or translation (with `processingBody`)
- being used while creating a new translation

`singleProcessingSubmissionDetails.tsx`
- displayed on the aside
- shows some more information for currently selected submission

`transcriptTabContent.tsx`
`translationsTabContent.tsx`
- content components for tabs from `singleProcessingContent`
- both have different steps of creating a transcript/translation
- given step of creating/managing is computed from `singleProcessingStore` data

## Related issues

Closes #3094 
Closes #3092
Closes #3570
Closes #3567 
Closes #3662
Closes #3688
Closes #3738


bufke:
Bug - broken audio file results in infinite loading instead of an error message. Upload a broken audio file (such as taking a random file and renaming it .mp3). I'd expect to see a message like "Unable to play audio file" but instead I see loading.

image

bufke:
Not blocking: Given how deeply nested the transcript workflow becomes, I would use react router so that back/forwards works. Maybe for next time. This also probably brings up design questions, like should it use breadcrumbs? The UI is fairly complex so the design time might be well worth it.

bufke:
image

Error on second language transcription when not already added to form

1. Upload mp3 to survey
2. Enter English transcription
3. Enter a second (I used Esperanto, which was not added to the form) 

I'd expect either this to just work or else give me an error about how Esperanto isn't added to the form. Instead I get an error and no indication on how to fix it.

When I reload and try again, it works. Bug me if more help is needed reproducing. I'm not 100% sure on the reproduction steps.

bufke:
Bug: I'm able to get a blank form

1. Go to Translations tab
2. Click new translation
3. Click Translations tab again

I'd expect to see the same thing as in step 1 (a list view of translations). Instead I get a blank page.

image

kobo/apps/subsequences/actions/unknown_action.py
```diff
@@ -0,0 +1,8 @@
+from .base import BaseAction
+
+
+class UnknownAction(BaseAction):
+    ID = 'unknown_action'
+    def load_params(self, params):
+        action_id = params.get('action_id', '?')
+        raise Exception(f'uknown action: {action_id}')
```
bufke, line 8: Spelling

jsapp/js/alertify.ts
```diff
@@ -93,24 +109,25 @@ export function multiConfirm(confirmId, title, message, buttons) {
   const dialog = alertify[confirmId]();
 
   // set up closing modal on ESC key
-  const killMe = (evt) => {
-    if (evt.keyCode === KEY_CODES.ESC) {
+  const killMe = (evt: JQuery.KeyUpEvent) => {
```
bufke, line 112: Is it (or should it be) a goal to remove jquery since we also use react? This line replaces a constant with a JQuery property.

bufke (commented):
I'll leave more comments later, breaking this up a bit.

kobo/apps/subsequences/models.py
```diff
@@ -0,0 +1,25 @@
+from django.db import models
+from kpi.models import Asset
+from django.contrib.postgres.fields import JSONField
+
+
+class SubmissionExtras(models.Model):
+    date_created = models.DateTimeField(auto_now_add=True)
+    date_modified = models.DateTimeField(auto_now=True)
+    uuid = models.CharField(max_length=40, null=True)
```
bufke, line 9: Make this refer to submission_uuid (or whatever the right name is) so that I don't assume it's a primary key for SubmissionExtras itself

magicznyleszek:
> Bug - broken audio file results in infinite loading instead of an error message. Upload a broken audio file (such as taking a random file and renaming it .mp3). I'd expect to see a message like "Unable to play audio file" but instead I see loading.
> 
> image

I've added error handling and displaying of error message when this happens :+1:

magicznyleszek:
> Bug: Error on second language transcription when not already added to form
> 
> image
> 
>     1. Upload mp3 to survey
> 
>     2. Enter English transcription
> 
>     3. Enter a second (I used Esperanto, which was not added to the form)
> 
> 
> I'd expect either this to just work or else give me an error about how Esperanto isn't added to the form. Instead I get an error and no indication on how to fix it.
> 
> When I reload and try again, it works. Bug me if more help is needed reproducing. I'm not 100% sure on the reproduction steps.

I'm not entirely sure what caused the error, but the form languages are in no way connected to the transcript and translation languages - i.e. regardless if form has any language defined (or has just `null`), you can have any transcript and translation language.

magicznyleszek:
> Bug: I'm able to get a blank form
> 
>     1. Go to Translations tab
> 
>     2. Click new translation
> 
>     3. Click Translations tab again
> 
> 
> I'd expect to see the same thing as in step 1 (a list view of translations). Instead I get a blank page.
> 
> image

Fixed :+1:

jsapp/js/alertify.ts
magicznyleszek: Right, I dropped jquery usage from this file - it was being used for ESC key listening :)

kobo/apps/subsequences/models.py
dorey: Good point. I will fix this

jnm (commented):
This is a retrospective look at the README.md

kobo/apps/subsequences/README.md
```diff
@@ -0,0 +1,197 @@
...
+    ID = 'decimal_rounder'
+    def check_submission_status(self, submission):
+        if submission.get('fuel_cost') != None:
+            return ACTION_NEEDED
+        return PASSES
+
+## STEP 4: define the handler
+
+    def run_change(self, submission):
+        _data = submission.get(self._destination_field, {})
```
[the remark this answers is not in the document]
jnm: It's defined by `BaseAction` to be `_supplementalDetails`

kobo/apps/subsequences/README.md
```diff
@@ -0,0 +1,197 @@
...
+
+```
+
+#### Step 5: specify which surveys (`Asset`) should be passed to this handler
+
+Somewhere, either through the API or elsewhere, add relevant details to the asset's `advanced_features` field:
+
+```python
+for asset in Asset.objects.filter(name__contains='fuel'):
+    asset.advanced_features = {
```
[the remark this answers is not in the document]
jnm: Yes! And you'll get rejected if you try to add a new action without also allowing it in the JSON Schema

kobo/apps/subsequences/README.md
```diff
@@ -0,0 +1,197 @@
...
+
+```python
+class DecimalRounder(BaseAction):
+    ID = 'decimal_rounder'
+    # ...
+    # `modify_jsonschema` appended to the class above
+    def modify_jsonschema(self, schema):
+        defs = schema.get('definitions', {})
+        props = schema.get('properties', {})
+        defs['roundedschemadef'] = {
```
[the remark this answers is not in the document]
jnm: Maybe we can go with `#/defs/{self.ID}/roundednumber`

PR #3582, opened 2021-12-04
TableMediaPreview component style tweaks

magicznyleszek:
## Description

This change makes the image or video be always fully visible regardless of screen size, i.e. we no longer display scrollbar, just make the image/video fit (useful for very tall and short images). We also introduce a missing loading spinner for images and audio files.

## Additional details

Moved some styles around. Added an `KoboImage` component for displaying images with loading spinner. Added `mixins.scss` file with some helpful mixins.

## Related issues

Part of #3567


jsapp/js/components/common/audioPlayer.tsx
```diff
@@ -1,7 +1,9 @@
 import React from 'react'
-import autoBind from 'react-autobind'
 import bem, {makeBem} from 'js/bem'
 import KoboRange from 'js/components/common/koboRange'
+import LoadingSpinner from 'js/components/common/loadingSpinner'
+import Icon from 'js/components/common/icon'
+import Button from 'js/components/common/button'
```
duvld, line 6: I edited the TODOs: https://github.com/kobotoolbox/kpi/pull/3582/commits/b7d282699c39e86cae66e7da325248490218b326 the icon is smaller and the padding is bigger. I think it looks okay though, what do you think?

jsapp/js/components/common/koboImage.tsx
```diff
@@ -0,0 +1,62 @@
...
+  componentDidMount() {
+    this.preloadImage()
+  }
+
+  componentWillReceiveProps() {
+    this.preloadImage()
+  }
+
+  preloadImage() {
+    this.setState({isLoading: true})
```
duvld, line 37: same as above

jsapp/scss/mixins.scss
```diff
@@ -0,0 +1,32 @@
+@use "scss/sizes";
+
+@mixin buttonReset() {
```
duvld, line 3: not sure if we need this after new button component?

jsapp/js/components/common/audioPlayer.tsx
magicznyleszek: It's ok :+1: I will revisit all the icons in https://github.com/kobotoolbox/kpi/pull/3635 (WIP) so if it doesn't fit the size Ignacio wanted I will change it then :)

jsapp/js/components/common/koboImage.tsx
magicznyleszek: In audio you're right, but here we can receive a `preloadImage` request via `componentWillReceiveProps`. This could happen if for some reason the image will change after `KoboImage` is rendered. I don't think we have such a case now in code, but it could happen in future, so let's leave it :)

jsapp/scss/mixins.scss
magicznyleszek: Right, I think I copied `mixins.scss` from some other branch that used that mixin. Will delete it here :)

jsapp/js/components/common/audioPlayer.tsx
```diff
@@ -23,52 +25,57 @@ type AudioPlayerState = {
...
+  componentDidMount() {
+    this.prepareAudio()
+  }
+
+  componentWillUnmount() {
+    this.audioInterface.pause()
+  }
+
+  prepareAudio() {
+    this.setState({isLoading: true})
```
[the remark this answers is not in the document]
magicznyleszek: Yes :) Fixed

PR #3586, opened 2021-12-07
Synchronous exports

### LINKED ISSUE #3319: Add more convenient (synchronous) exports
Probably best done by taking a `AssetExportSettings` object and feeding it into a new endpoint that attempts to churn out the whole XLS or CSV export before hitting the two-minute uWSGI / NGINX timeout.

NB: `AssetExportSettings` is incorrectly named. We aren't exporting an `Asset` with those; we're exporting submissions. Could be renamed to `ExportSettings` to follow `ExportTask` convention, or maybe both could get `Data` as a prefix, e.g. `DataExportTask`(?)

joshuaberetta:
## Description

Replace legacy synchronous exports: use formpack and export settings.

## Related issues

closes #3319


kpi/models/import_export_task.py
```diff
@@ -395,7 +396,150 @@ def export_upload_to(self, filename):
     return posixpath.join(self.user.username, 'exports', filename)
 
 
-class ExportTask(ImportExportTask):
+class ExportBase:
```
noliveleger, line 399: 🤔 
Let me suggest something. `ExportBase` is not really a model but provide some functionality to get export object. Moreover the children class do not inherit from it as their base class but their secondary one.
What about call it a mixin (e.g. `ExportObjectMixin`) and move it into the `mixins` folder?
IMO, it would be make more sense to import a mixin in `renderers.py` too (than a base model class).

kpi/renderers.py
```diff
@@ -122,6 +124,54 @@ def render(self, data, accepted_media_type=None, renderer_context=None):
...
+    ) -> Tuple[formpack.reporting.Export, Generator]:
+        view = renderer_context['view']
+        self.user = view.request.user
+        return super().get_export_object(
+            source=view.asset,
+            _async=False,
+        )
+
+
+class SubmissionXLSXRenderer(SubmissionRendererExportBase):
```
noliveleger, line 139: I'm wondering whether `SubmissionRendererExportBase` is really needed?
We save 2 lines of code (4 if we count `SubmissionCSVRenderer`). 

`SubmissionXLSXRenderer` and `SubmissionCSVRenderer` could inherit directly from `renderers.BaseRenderer` and `ExportBase` (aka `ExportObjectMixin`)  and call 

```python
view = renderer_context['view']
self.user = view.request.user
```
just before `self.get_export_object()`

What do you think?

kpi/serializers/v2/asset_export_settings.py
```diff
@@ -38,13 +39,15 @@ class Meta:
...
             'url',
+            'exports_url',
             'name',
             'date_modified',
             'export_settings',
         )
         read_only_fields = (
             'uid',
             'url',
+            'exports_url',
```
noliveleger, line 50: FYI: `SerializerMethodField` is already read-only.  I think that `read_only_fields` is useless in that case.

kpi/renderers.py
```diff
@@ -122,6 +124,54 @@ def render(self, data, accepted_media_type=None, renderer_context=None):
         )
 
 
+class SubmissionRendererExportBase(renderers.BaseRenderer, ExportBase):
+    def get_export_object(
+        self, renderer_context: Dict
+    ) -> Tuple[formpack.reporting.Export, Generator]:
+        view = renderer_context['view']
+        self.user = view.request.user
```
noliveleger, line 132: Do you think it would be useful to add a comment here about why `self.user` needs to be set?

kpi/renderers.py
```diff
@@ -122,6 +124,54 @@ def render(self, data, accepted_media_type=None, renderer_context=None):
...
+    ) -> BytesIO:
+        self.data = data
+        export, submission_stream = self.get_export_object(renderer_context)
+        stream = BytesIO()
+        export.to_xlsx(stream, submission_stream)
+        stream.seek(0)
+        return stream
+
+
+class SubmissionCSVRenderer(SubmissionRendererExportBase):
```
noliveleger, line 157: Obviously, same question here about `SubmissionRendererExportBase`

noliveleger (changes requested):

kpi/models/import_export_task.py
joshuaberetta: That's a great idea ❤️

kpi/renderers.py
joshuaberetta: Good idea 👍 I'll make that change.

kpi/renderers.py
joshuaberetta: Good idea 👍

PR #3600, opened 2021-12-16
Build Docker images with GitLab CI and deploy to beta instance

bufke:
## Description

- Build docker images on commit, tag as both short sha hash and ref (branch/tag). Push to gitlab docker registry.
- On beta (and this branch to test) push to a beta instance via helm chart

CI configuration and builds are public. They depend on a protected gitlab ci variable for pushing in [gitlab settings](https://gitlab.com/kobo_toolbox/kpi/-/settings/repository) for those with access. Information on how permissions work for this aren't related to this specific PR but are important for how it works.

## Related issues

kobotoolbox/kobocat#785

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -8,7 +8,7 @@ build:
   services:
     - docker:dind
   before_script:
-    - echo "$DOCKER_REGISTRY_PASS" | docker login $DOCKER_REGISTRY --username $DOCKER_REGISTRY_USER --password-stdin
+    - docker login -u gitlab-ci-token -p $CI_BUILD_TOKEN registry.gitlab.com
   script:
     - docker pull $CI_REGISTRY_IMAGE:CI_COMMIT_REF_NAME || true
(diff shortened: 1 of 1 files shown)


jnm:
@bufke what's this mean?
> They depend on a protected gitlab ci variable for pushing in [gitlab settings](https://gitlab.com/kobo_toolbox/kpi/-/settings/repository) for those with access.

I see `gitlab-ci-token` and `$CI_*`, but GitLab exposes those by default, right? I was just wondering what we need to pay attention to in settings.

jnm (approved):

bufke:
The GitLab "CI/CD Variable" KUBECONFIG how the CI script has access to make k8s changes. That variable is set to only show in protected branches. This is important as it allows us access in certain branches but not others. Perhaps outside_company branch is authorized for commits, but not allowed to push to the beta k8s app. You are correct that there are additional build in variables for things like authenticating to GitLab's docker registry.

PR #3613, opened 2022-01-10
Fix validation column disappearing

### LINKED ISSUE #3612: Validation column disappears when using "hide fields" functionality
## Description

When hiding any column in "hide fields", the "Validation" column also disappears. The problem doesn't occur when hiding a column through the column header dropdown.

magicznyleszek:
## Description

Validation status column will be always visible and unable to hide.

## Related issues

Fixes #3612


magicznyleszek:
Closed in favour of #3679

PR #3619, opened 2022-01-14
Standardize `country` and `sector` user metadata

jnm:
Previously, the user registration form saved `sector` in the extra user details, while the account settings screen used `primarySector`. In #3611, `primarySector` was changed to `primary_sector`, but the registration form still used `sector`. This changes the account settings screen to use `sector` to be consistent with the registration form and also the project metadata. It also converts lone-string `sector` and `country` values to the label-value object expected by the front end.

## Description

Fixes an issue where sector and country details entered during account registration did not appear in the account settings page.

### DIFF
--- a/kpi/serializers/current_user.py
+++ b/kpi/serializers/current_user.py
@@ -75,18 +75,47 @@ class CurrentUserSerializer(serializers.ModelSerializer):
     def to_representation(self, obj):
         if obj.is_anonymous:
             return {'message': 'user is not logged in'}
+
         rep = super().to_representation(obj)
-        if not rep['extra_details']:
+        if not rep['extra_details'] or not isinstance(
+            rep['extra_details'], dict
+        ):
             rep['extra_details'] = {}
-        elif rep.get('primarySector'):
-            # the front end used to set this with camelCase but has since
-            # been fixed
-            rep['primary_sector'] = rep['primarySector']
-            del rep['primarySector']
+        extra_details = rep['extra_details']
+
+        # the front end used to set `primarySector` but has since been changed
+        # to `sector`, which matches the registration form
+        if extra_details.get('primarySector') and not extra_details.get(
+            'sector'
+        ):
+            extra_details['sector'] = extra_details['primarySector']
+
+        # remove `primarySector` to avoid confusion
+        try:
+            del extra_details['primarySector']
+        except KeyError:
+            pass
+
+        # the registration form records only the value, but the front end
+        # expects an object with both the label and the value.
+        # TODO: store and load the value *only*
+        for field in 'sector', 'country':
+            if isinstance(extra_details.get(field), str):
+                extra_details[field] = {
+                    'label': extra_details[field],
+                    'value': extra_details[field],
+                }
+
         # `require_auth` needs to be read from KC every time
-        if settings.KOBOCAT_URL and settings.KOBOCAT_INTERNAL_URL:
-            rep['extra_details']['require_auth'] = get_kc_profile_data(
-                obj.pk).get('require_auth', False)
+        # except during testing, when KC's database is not available
+        if (
+            settings.KOBOCAT_URL
+            and settings.KOBOCAT_INTERNAL_URL
+            and not settings.TESTING
+        ):
+            extra_details['require_auth'] = get_kc_profile_data(obj.pk).get(
+                'require_auth', False
+            )
 
         return rep
 
--- a/kpi/serializers/current_user.py
+++ b/kpi/serializers/current_user.py
@@ -77,16 +77,18 @@ class CurrentUserSerializer(serializers.ModelSerializer):
             return {'message': 'user is not logged in'}
 
         rep = super().to_representation(obj)
-        if not rep['extra_details'] or not isinstance(
-            rep['extra_details'], dict
+        if (
+            not rep['extra_details']
+            or not isinstance(rep['extra_details'], dict)
         ):
             rep['extra_details'] = {}
         extra_details = rep['extra_details']
 
         # the front end used to set `primarySector` but has since been changed
         # to `sector`, which matches the registration form
-        if extra_details.get('primarySector') and not extra_details.get(
-            'sector'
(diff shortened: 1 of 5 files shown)


kpi/serializers/current_user.py
```diff
@@ -75,18 +75,47 @@ def get_git_rev(self, obj):
     def to_representation(self, obj):
         if obj.is_anonymous:
             return {'message': 'user is not logged in'}
+
         rep = super().to_representation(obj)
-        if not rep['extra_details']:
+        if not rep['extra_details'] or not isinstance(
+            rep['extra_details'], dict
+        ):
```
noliveleger, line 82: We already add a discussion with @joshuaberetta in other PR about this format in `if` statement. 
My eyes always try to find the opening parenthesis next to `if` and make my brain think there is an error. 

What about this instead?:
```python
if (
  not rep['extra_details'] 
  or not isinstance(rep['extra_details'], dict)
):
```

It has also the advantage to put the operator on another line.

There is an internal discussion in Zulip: https://chat.kobotoolbox.org/#narrow/stream/4-KoBo-Dev/topic/Python.20coding.20style/near/5423

kpi/serializers/current_user.py
```diff
@@ -75,18 +75,47 @@ def get_git_rev(self, obj):
...
-            # been fixed
-            rep['primary_sector'] = rep['primarySector']
-            del rep['primarySector']
+        extra_details = rep['extra_details']
+
+        # the front end used to set `primarySector` but has since been changed
+        # to `sector`, which matches the registration form
+        if extra_details.get('primarySector') and not extra_details.get(
+            'sector'
+        ):
```
noliveleger, line 90: I would have the same suggestion for this `if` statement as said above .

Link To be copied

kpi/serializers/current_user.py
```diff
@@ -75,18 +75,47 @@ def get_git_rev(self, obj):
...
         # `require_auth` needs to be read from KC every time
-        if settings.KOBOCAT_URL and settings.KOBOCAT_INTERNAL_URL:
-            rep['extra_details']['require_auth'] = get_kc_profile_data(
-                obj.pk).get('require_auth', False)
+        # except during testing, when KC's database is not available
+        if (
+            settings.KOBOCAT_URL
+            and settings.KOBOCAT_INTERNAL_URL
+            and not settings.TESTING
+        ):
```
noliveleger, line 115: ❤️

kpi/tests/api/v1/test_api_users.py
```diff
@@ -1,7 +1,46 @@
...
+        # `primarySector` should be renamed to `sector`
+        xtradata['primarySector'] = 'camelCase Administration'
+        user.extra_details.save()
+        response = self.client.get(reverse('currentuser-detail'))
+        # lone string should be transformed to object with label and value
+        assert response.data['extra_details']['sector'] == {
+            'label': 'camelCase Administration',
+            'value': 'camelCase Administration',
+        }
+        assert not 'primarySector' in response.data['extra_details']
```
noliveleger, line 27: Shouldn't be `assert 'primarySector' not in response.data['extra_details']` [instead](https://www.python.org/dev/peps/pep-0008/#programming-recommendations)?
:fly: 😗

kpi/tests/api/v1/test_api_users.py
```diff
@@ -1,7 +1,46 @@
...
+
+        # …but only if `sector` doesn't already exist
+        xtradata['sector'] = 'Head Honchoing'
+        user.extra_details.save()
+        response = self.client.get(reverse('currentuser-detail'))
+        assert response.data['extra_details']['sector'] == {
+            'label': 'Head Honchoing',
+            'value': 'Head Honchoing',
+        }
+        assert not 'primarySector' in response.data['extra_details']
```
noliveleger, line 37: `not in`

kpi/tests/api/v1/test_api_users.py
```diff
@@ -1,7 +1,46 @@
...
+    def test_current_user_extra_details_kludges(self):
+        self.client.login(username='someuser', password='someuser')
+        user = User.objects.get(username='someuser')
+        xtradata = user.extra_details.data
+        assert xtradata == {}
+
+        # `primarySector` should be renamed to `sector`
+        xtradata['primarySector'] = 'camelCase Administration'
+        user.extra_details.save()
+        response = self.client.get(reverse('currentuser-detail'))
```
noliveleger, line 21: Would it better to call `reverse(self._get_endpoint('currentuser-detail'))` even if `CurrentUserViewSet` does not exist in `v2` yet? 
When (if?) we port the viewset to a newer version of API, we could copy the tests and they should keep working.
What do you think?

noliveleger (changes requested):

kpi/serializers/current_user.py
jnm: this was formatted by Black :shrug:
i agree that your suggestion is nicer, except that you're indenting by 2 spaces (?)

kpi/tests/api/v1/test_api_users.py
jnm: right

kpi/tests/api/v1/test_api_users.py
jnm: sure

kpi/tests/api/v1/test_api_users.py
jnm: fine

kpi/serializers/current_user.py
jnm: applied the suggested formatting but with 4-space indentation

kpi/serializers/current_user.py
jnm: ok, done

kpi/serializers/current_user.py
noliveleger: Sorry, I did it quick. 4-space it should have been ;-)

PR #3621, opened 2022-01-14
Upgrade react select and hack the unwanted scrollbar bug

magicznyleszek:
## Description

Hackfixes a bug in New Project modal when opening a dropdown caused a scrollbar to appear unwantedly. Also includes an upgrade of `react-select` version.

## Related issues

This is a way around https://github.com/JedWatson/react-select/issues/4108


jsapp/js/components/common/wrappedSelect.tsx
```diff
@@ -1,22 +1,28 @@
...
+class WrappedSelect extends React.Component<WrappedSelectProps> {
   render() {
     return(
       <bem.KoboSelect__wrapper m={{
         'error': Boolean(this.props.error)
       }}>
         <label>
           {this.props.label}
+          <Select {...this.props}/>
           {super.render()}
```
jnm, line 26: maybe this does nothing now?

jnm (approved):

jsapp/js/components/common/wrappedSelect.tsx
magicznyleszek: You're right :)

PR #3637, opened 2022-01-20
Replace instances of "KoBoToolbox" with "KoboToolbox"

### LINKED ISSUE #3636: Use "KoboToolbox" spelling more consistently in the UI
## Description

There are several places where "KoBoToolbox" is used instead of "KoboToolbox".

joshuaberetta:
## Description

Use consistent spelling of "KoboToolbox" and "Kobo" in the UI.

## Related issues

closes #3636


magicznyleszek (changes requested):
Looks good, just one question: should we also replace "KoBo" with "Kobo"? I think there is only a few instaces of it.

PR #3649, opened 2022-01-31
Types for myLibraryStore, tableStore, dataInterface, and mixins

magicznyleszek:
## Description

Types for `tableStore`, with all necessary types for it (`myLibraryStore`, `dataInterface`, and `mixins`.


package.json
```diff
@@ -51,6 +51,7 @@
     "@babel/preset-react": "^7.9.4",
     "@types/jquery": "^3.5.10",
     "@types/lodash": "^4.14.179",
+    "@types/lodash.clonedeep": "^4.5.6",
```
bufke, line 54: Since we don't use lodash (we only use clonedeep, merge, orderby), I suggest we remove @types/lodash and specify only @types/lodash.clonedeep. This has 0 impact on installed modules because the @types lodash/clonedeep has a dependency of lodash. But it better reflects the reality of what we use.

jsapp/js/utils.ts
```diff
@@ -249,7 +249,7 @@ export function escapeHtml(str: string): string {
   return div.innerHTML;
 }
 
-export function renderCheckbox(id: string, label: string, isImportant: boolean) {
+export function renderCheckbox(id: string, label: string, isImportant?: boolean) {
```
bufke, line 252: `function renderCheckbox(id: string, label: string, isImportant = false)` better reflects the intended usage. We seem to only every ignore this param, or set to true. Undefined essentially implies false in how we use it.

jsapp/js/mixins.tsx
```diff
@@ -35,63 +36,172 @@ import {
...
+  SubmissionResponse,
+} from 'js/dataInterface';
 
 const IMPORT_CHECK_INTERVAL = 1000;
 
-var mixins = {};
+interface MixinsObject {
+  contextRouter: {
+    [functionName: string]: Function;
+    context?: any; // Hack to not go into typing the mixins too much
```
bufke, line 54: I would say any usage of `any` is a "hack" we want to remove. Might be better to comment the rare, legitimate usage of any and assume other usages are temporary hacks to remove when possible.

bufke (commented):
Looks mostly good to me. Would be fine to merge after reviewing my comments.

PR #3667, opened 2022-02-08
Fix upload and import modal properties not being applied

magicznyleszek:
## Description

Fix a bug introduced by configurable metadata feature - in new project modal, when uploading or importing a file, the imported name and settings were not applied.


magicznyleszek:
This seems to be fixed elsewhere, but this PR fixes some other tiny thing, so let's merge it

Issue #3563, opened 2021-11-18
[Feature] - 2FA/MFA support 


noliveleger:
Add support for multi-factor authentication to KoBoToolbox application. 

Design: #3552 
Internal discussion: https://chat.kobotoolbox.org/#narrow/stream/4-KoBo-Dev/topic/MFA/near/2045

PR #3674, opened 2022-02-10
[ MFA ] Make the help message editable

noliveleger:
## Description

Super user can modify (and translate) the message to help users who lost their device and the backup codes.

## Related issues

Part of #3563


noliveleger:
@magicznyleszek , the PR also contains some front-end changes. Would you mind to look at them?

kobo/settings/base.py
```diff
@@ -199,7 +199,24 @@
...
+    ),
+    'MFA_I18N_HELP_TEXTS': (
+        json.dumps({
+            'default': (
+                'KoBoToolbox uses Multi-Factor Authentication. '
+                'If you no longer have access to the device registered '
+                'with your account and the backup codes, the KoBo Support Team '
+                'can assist you regaining access to the system.'
+                'To request this support, please contact us at '
+                '[support@kobotoolbox.org](mailto:support@kobotoolbox.org)'
```
jnm, line 212: If it's possible to reference one Constance setting from another, then could you replace this with `SUPPORT_EMAIL`, please? If not, maybe this string could have something like `##support email##` as a placeholder, and then replace that elsewhere(?)

kpi/views/environment.py
```diff
@@ -53,7 +63,12 @@ def get(self, request, *args, **kwargs):
                 processor = None
             value = getattr(constance.config, key)
             if processor:
-                value = processor(value)
+                try:
+                    value = processor(value)
+                except json.JSONDecodeError:
+                    logging.error(f'Environment.py: Could not decode `{key}`')
+                    continue
```
jnm, line 70: This might be smart to keep to avoid bad settings basically crashing the entire server. However, it's also possible to validate the JSON when it's being saved from the admin interface. I don't remember how, exactly, but I did it for the custom metadata stuff :)

jnm (changes requested):

kobo/settings/base.py
noliveleger: I chose the placeholder solution. Could not find a way to make it work as a reference

kpi/views/environment.py
noliveleger: Yep :-) https://github.com/kobotoolbox/kpi/blob/3563-disabled-date-support/kobo/settings/base.py#L249-L254 

Thank you for bushwacking

jnm (approved):

jnm (changes requested):
Unfortunately, I think we will have to change the front-end component of this PR, but we cannot do it until we have a final UX decision: https://chat.kobotoolbox.org/#narrow/stream/7-UX.2FUI/topic/MFA.20help.20text

magicznyleszek (approved):

noliveleger:
FYI @magicznyleszek @duvld will take over the front-end work. Based on latest mockups, the front-end code of this PR must be modified.

magicznyleszek (approved):

jsapp/js/components/account/mfaModals.tsx
```diff
@@ -470,18 +476,18 @@ export default class MFAModals extends React.Component<
...
-                t('Please note that in order to reconfigure two-factor authentication (2FA), the previous set up will need to be deleted. Tokens or recovery codes from the previous set up will not be valid anymore')
+                t('Please note that in order to reconfigure two-factor authentication (2FA), the previous set up will need to be deleted. Tokens or recovery codes from the previous set up will not be valid anymore.')
               }
             </strong>
           </bem.MFAModal__p>
 
           {this.props.modalType === 'reconfigure' &&
             <bem.MFAModal__p>
-              {t("Once your 2FA has been deactivated, you'll be prompted to configure it again. If you cannot complete the process. two-factor authentication will remain disabled for your account. In this case, you can reenable it at any time through the usual process.")}
+              {t("Once your 2FA has been deactivated, you'll be prompted to configure it again. If you cannot complete the process. 2FA will remain disabled for your account. In this case, you can reenable it at any time through the usual process.")}
```
jnm, line 490: note to self: revise

jsapp/js/components/account/securityRoute.tsx
```diff
@@ -139,26 +166,31 @@ export default class SecurityRoute extends React.Component<
...
+                label={this.state.isMfaActive ? t('Enabled') : t('Disabled')}
+                checked={this.state.isMfaActive}
                 onChange={this.onToggleChange.bind(this)}
               />
             </bem.SecurityRow__buttons>
           </bem.SecurityRow__header>
 
           <bem.SecurityRow__description>
-            {t('Two-factor authentication (2FA) is an added layer of security used when logging into the platform. We recommend enabling Two-factor authentication for an additonal layer of protection.')}
+            {t('Two-factor authentication (2FA) is an added layer of security used when logging into the platform. We recommend enabling Two-factor authentication for an additional layer of protection.')}
```
jnm, line 177: note to self: revise (weird capitalization; other messages use 2FA; standardize this)

jsapp/js/components/account/securityRoute.tsx
```diff
@@ -190,6 +222,13 @@ export default class SecurityRoute extends React.Component<
               </bem.MFAOptions__row>
             </bem.MFAOptions>
           }
+
+          {!this.state.isMfaActive && this.state.dateDisabled &&
+            <InlineMessage
+              type='default'
+              message={t('Two-factor authentication was deactivated for your account on the ##date##').replace('##date##', formatDate(this.state.dateDisabled))}
```
jnm, line 229: note to self: revise

jsapp/js/components/permissions/sharingForm.es6
```diff
@@ -118,16 +119,19 @@ class SharingForm extends React.Component {
...
+            <InlineMessage
+              type='warning'
+              icon='alert'
+              message={
+                <span>
+                  {t('Anyone can see this blank form and add submissions to it because you have not set ')}
+                  <a href={`/#${ROUTES.ACCOUNT_SETTINGS}`}>
+                    {t('your account')}
+                  </a>
+                  {t(' to require authentication.')}
```
jnm, line 131: not new to this PR, but unacceptable use of translated strings. the translator needs to see the entire string, not fragments of it before and after the hyperlink. in this case, it doesn't seem like a big deal to add the `<a href=…` using string replacement instead of JSX, but we may need a more standard way to handle this situation where a `## placeholder ##` in a translated string needs to be replaced by rendered JSX. first thought: split the translated string (*after* translation) into the parts before and after `## placeholder ##`

kobo/apps/mfa/templates/mfa_token.html
```diff
@@ -2,23 +2,72 @@
...
+      </section>
+
+      <section class="mfa-token-form__section" data-mfa-token-form-section="help">
+        <h1 class="mfa-token-form__header">
+          {% trans "Verification issues" %}
+        </h1>
+
+        <div class="mfa-token-form__help-text">
+          {{ mfa_help_text | safe }}
+        </div>
```
jnm, line 50: note to self: verify wording and revise as needed

kpi/views/environment.py
```diff
@@ -38,6 +40,18 @@ class EnvironmentView(APIView):
             'OPERATIONAL_PURPOSE_CHOICES',
             lambda text: tuple((line.strip('\r'), line.strip('\r')) for line in text.split('\n')),
         ),
+        (
+            'MFA_I18N_HELP_TEXTS',
+            lambda i18n_texts: {
+                lang: markdown(text)
+                for lang, text in json.loads(
+                    i18n_texts.replace(
+                        '##support email##', constance.config.SUPPORT_EMAIL
```
jnm, line 49: unsure we should substitute `##support email##` here since the superuser has full control over the entire string. administrators may not want to use the same address for general support as for MFA reset issues

jnm (changes requested):

kpi/views/environment.py
jnm: boneheaded comment on my part. if they want a custom email address, they can simply use that and omit `##support email##` from their string.

jsapp/js/components/permissions/sharingForm.es6
jnm: https://chat.kobotoolbox.org/#narrow/stream/4-Kobo-Dev/topic/translated.20sentences.20with.20html.20links

jsapp/js/editorMixins/cascadeMixin.es6
```diff
@@ -49,7 +49,7 @@ export const cascadeMixin = {
         choices: inp
       });
       if (tmpSurvey.choices.length === 0) {
-        throw new Error(t('Paste your formatted table from excel in the box below.'));
+        throw new Error('Paste your formatted table from excel in the box below.');
```
jnm, line 52: @magicznyleszek doesn't this message get shown to the user?

https://github.com/kobotoolbox/kpi/blob/a9a24018d321fad4a83108c51b79fa2d38d263c1/jsapp/js/editorMixins/cascadeMixin.es6#L71-L85

jsapp/js/editorMixins/cascadeMixin.es6
```diff
@@ -60,7 +60,7 @@ export const cascadeMixin = {
       */
       var rowCount = tmpSurvey.rows.length;
       if (rowCount === 0) {
-        throw new Error(t('Paste your formatted table from excel in the box below.'));
+        throw new Error('Paste your formatted table from excel in the box below.');
```
jnm, line 63: I believe this also gets shown to the user (see above)

jnm (changes requested):

jsapp/js/components/permissions/sharingForm.es6
dorey: I think translators can work around an HTML link i.e.

```
t('Anyone can see this blank form and add submissions to it because you have not set
<a href="##accountsettingsurl##">your account</a> to require authentication.')}
```

But this will require using React's `dangerouslySetInnerHTML` syntax. (Which I think is already needed inside `InlineMessage` )

jnm (approved):

Issue #3654, opened 2022-02-02
Edit and view submissions in Enketo without using KoBoCAT


jnm:
We'll build on previous work for Enketo previewing so that KPI provides enough of the OpenRosa API to allow direct integration with Enketo for submission viewing and editing.

@noliveleger, please feel free to correct or expand this description as needed.

## Previous work

What's below was implemented in #2537.

> For previewing, KPI would have to implement minimal OpenRosa form list and form manifest endpoints
> 
> ### Form list (will only ever contain one form in this context)
> ```xml
> <xforms xmlns="http://openrosa.org/xforms/xformsList">
> 	<xform>
> 		<formID>agNZGnGtHbpH9hbQPjn6rQ [asset.uid]</formID>
> 		<name>Media! [asset.name]</name>
> 		<majorMinorVersion/> [seems to be unused]
> 		<version/> [seems to be unused]
> 		<hash>md5:abcdef123567890blah [probably the hash of the asset snapshot xml?]</hash>
> 		<descriptionText>Media! [also asset.name?]</descriptionText>
> 		<downloadUrl>asset snapshot xml url</downloadUrl>
> 		<manifestUrl>https://kc.kobotoolbox.org/jnm_demo/xformsManifest/372177 [new KPI manifest endpoint]</manifestUrl>
> 	</xform>
> </xforms>
> ```
> 
> ### Form manifest [zero, one, or many media]
> ```xml
> <manifest xmlns="http://openrosa.org/xforms/xformsManifest">
> 	<mediaFile>
> 		<filename>nuthatch.png</filename>
> 		<hash>md5:d1f05ced9b4382e0a0b8c6daf32efc28</hash>
> 		<downloadUrl>https://kc.kobotoolbox.org/jnm_demo/xformsMedia/372177/966204.png</downloadUrl>
> 	</mediaFile>
> 	<mediaFile>
> 		…
> 	</mediaFile>
> </manifest>
> ```
> 
> _Originally posted by @jnm in https://github.com/kobotoolbox/kpi/issues/2487#issuecomment-563454871_

PR #3689, opened 2022-02-16
Open edit/preview Enketo Express links with KPI as OpenRosa server

noliveleger:
## Checklist

3. [x] Ensure the tests pass
7. [] Open an issue in the [docs](https://github.com/kobotoolbox/docs/issues/new) if there are UI/UX changes

## Description

Get Enketo Express preview and edit links through KPI directly. Add an OpenRosa server to let Enketo Express submit submissions on edit.

## Related issues

Closes https://github.com/kobotoolbox/kpi/issues/3654
Blocked by kobotoolbox/kobocat#791


kpi/views/v2/asset_snapshot.py
```diff
@@ -126,6 +150,40 @@ def preview(self, request, *args, **kwargs):
...
+        detail=True,
+        trailing_slash='',
+        permission_classes=[EditSubmissionPermission],
+        methods=['HEAD', 'POST'],
+        authentication_classes=[DigestAuthentication],
+    )
+    def submission(self, request, *args, **kwargs):
+        if request.method == 'HEAD':
+            # ToDo figure out what's the valid response
+            return Response({})
```
jnm, line 163: Empty response body is okay. The response code should be 204. Some headers are supposed to be returned as well: `X-OpenRosa-Version`, `X-OpenRosa-Accept-Content-Length`. See https://docs.getodk.org/openrosa-form-submission/#extended-transmission-considerations

kpi/views/v2/data.py
```diff
@@ -376,7 +382,8 @@ def destroy(self, request, pk, *args, **kwargs):
         url_path='(enketo\/)?edit',
     )
     def enketo_edit(self, request, pk, *args, **kwargs):
-        return self._enketo_request(request, pk, action_='edit', *args, **kwargs)
+        submission_id = positive_int(pk)
```
jnm, line 385: ideally we should not 500 if it's the client's fault, such as `http://kf.kobo.local:9000/api/v2/assets/aYgpAJH2fLhTEDjx4ApU3Y/data/haha/enketo/edit/`, which currently raises a `ValueError: invalid literal for int() with base 10: 'haha'`

kpi/views/v2/data.py
```diff
@@ -549,3 +546,89 @@ def _filter_mongo_query(self, request):
...
+        submission_json = deployment.get_submission(
+            submission_id, user, request=request
+        )
+
+        _, submissions_stream = build_formpack(
+            self.asset,
+            submission_stream=[submission_json],
+            use_all_form_versions=True
+        )
+        version_uid = list(submissions_stream)[0][INFERRED_VERSION_ID_KEY]
```
jnm, line 589: probably could move `_infer_version_id()` from where it's currently nested inside `build_formpack()` to some utility file

kpi/views/v2/data.py
```diff
@@ -549,3 +546,89 @@ def _filter_mongo_query(self, request):
...
+
+        try:
+            snapshot = self.asset.versioned_snapshot(version_uid=version_uid)
+        except AssetVersion.DoesNotExist:
+            raise serializers.ValidationError(
+                {'version': t('Version not found')}
+            )
```
jnm, line 596: i don't think this should ever happen since `_infer_version_id()` falls back on the latest deployed version. i suppose there could be a race condition between `_infer_version_id()` and `versioned_snapshot()`, but even then, perhaps we should return a 500 error. it's not something the client can correct.

kpi/views/v2/data.py
```diff
@@ -549,3 +546,89 @@ def _filter_mongo_query(self, request):
...
+        for attachment in attachments:
+            key_ = f'instance_attachments[{attachment.media_file_basename}]'
+            data[key_] = reverse(
+                'attachment-detail',
+                args=(self.asset.uid, submission_id, attachment.pk),
+                request=request,
+            )
+
+        enketo_endpoint = getattr(
+            settings, f'ENKETO_{action_}_INSTANCE_ENDPOINT'.upper()
```
jnm, line 621: i'm not a huge fan of this because it makes it a lot more difficult to answer a question like "where are all the places where `ENKETO_VIEW_INSTANCE_ENDPOINT` is used?"

you already have a conditional `if action_ == 'edit'` at the beginning of the function; i'd request setting the `enketo_endpoint` up there.

jnm (changes requested):

jnm:
scenarios to test:
* manually specified `form_title`
* manually specified `name` (in XLSForm, i.e. the XML root node name)

kpi/views/v2/data.py
jnm: later

kpi/views/v2/data.py
jnm: #3692

PR #3700, opened 2022-02-24
Fix synchronous export failure on insufficient permissions

### LINKED ISSUE #3699: Synchronous exports without sufficient permissions results in `500` response
## Description

Synchronous export attempts for users without sufficient permissions results in a `500` server error response.

## Steps to Reproduce

1. Create form with submissions, data not public
2. Attempt to export as anonymous user
3. 500 error

## Expected behavior 

A `404` is more appropriate.

joshuaberetta:
## Description

Return `404` if user does not have sufficient permissions to create a synchronous export.

## Related issues

closes #3699

### DIFF
--- a/kpi/mixins/export_object.py
+++ b/kpi/mixins/export_object.py
@@ -10,7 +10,7 @@ from formpack.schema.fields import (
     TagsCopyField,
     ValidationStatusCopyField,
 )
-from django.http import Http404
+from django.utils.translation import gettext as t
 
 from kobo.apps.reports.report_data import build_formpack
 from kpi.constants import (
@@ -22,6 +22,11 @@ from kpi.utils.models import resolve_url_to_asset
 
 
 class ExportObjectMixin:
+
+    class InaccessibleData(Exception):
+        def __str__(self):
+            return t('This data does not exist or you do not have access to it')
+
     COPY_FIELDS = (
         IdCopyField,
         '_uuid',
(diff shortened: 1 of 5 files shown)


kpi/mixins/export_object.py
```diff
@@ -70,13 +70,7 @@ def get_export_object(
             PERM_VIEW_SUBMISSIONS not in source_perms
             and PERM_PARTIAL_SUBMISSIONS not in source_perms
         ):
-            # Unsure if DRF exceptions make sense here since we're not
-            # returning a HTTP response
```
jnm, line 74: i think the comment is still valid: `get_export_object()` is called by a celery task as well as your new view.

my proposal would be to:
1. create a new exception class for this kind of permission-denied-during-export error
1. raise the exception here
1. catch the exception in the view and have the view return a 404
1. in the celery task, add a more specific exception handler [ref](https://github.com/kobotoolbox/kpi/blob/c2bb9d3066d6523f270df09283ecf62b4af21709/kpi/models/import_export_task.py#L110-L114) that catches this new exception, marks the job as failed, and sets an appropriate error message

i'll write a draft of this and ask you to review it

jnm (changes requested):

kpi/tests/api/v2/test_api_exports.py
```diff
@@ -356,4 +348,83 @@ def test_synchronous_csv_export(self):
             line.decode() for line in export_content_response.streaming_content
         )
         assert synchronous_export_response.content.decode() == export_content
+        print(synchronous_export_response.content.decode())
```
jnm, line 351: :grin:

jnm (changes requested):
thanks for the new tests :+1: the only thing i found was one stray `print` :smiley:

kpi/mixins/export_object.py
jnm: https://github.com/kobotoolbox/kpi/pull/3713

kpi/tests/api/v2/test_api_exports.py
joshuaberetta: lol 🤦

PR #3707, opened 2022-03-01
Allow Enketo editing to use existing KPI authentication

noliveleger:
Enketo already passes through the `kobonaut` session cookie with its OpenRosa requests, but KPI was denying it with 403 because its POSTs did not include a CSRF token. This takes advantage of enketo/enketo-express#191 to supply Enketo with the CSRF token via cookie and verify that it's sent back properly in the POST data.

## Description

Avoid an extra prompt for username and password when editing submissions.

## Related issues

#3689, kobotoolbox/kobocat#786

### DIFF
--- a/kobo/settings/base.py
+++ b/kobo/settings/base.py
@@ -245,11 +245,6 @@ CONSTANCE_CONFIG = {
         "Options available for the 'operational purpose of data' metadata "
         'field, one per line.'
     ),
-    'ENKETO_AUTH_COOKIE_MAX_AGE': (
-        30 * 24 * 60 * 60,  # one month
-        'In seconds. It should match what Enketo Express sets.\n'
-        '(see https://tinyurl.com/mtc7zr2y)'
-    ),
 }
 CONSTANCE_ADDITIONAL_FIELDS = {
     'metadata_fields_jsonschema': [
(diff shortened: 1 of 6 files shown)


jnm:
@noliveleger, I added my commit on top of yours so that we can reference the `__enketo_meta_username` approach later if needed. (I also rebased onto the release branch.)

I noticed in your original description:
> This class is also deactivated when 2FA is enabled on the user's account.

Do we need to do anything special to make sure this custom `SessionAuthentication` subclass is **allowed** when 2FA is enabled? If 2FA blocks this new approach, I don't think 2FA users will be able to edit their submissions at all, which isn't what we want. Once they've established a KPI session, we can assume that they passed through the "front door" correctly by verifying their second authentication factor.

noliveleger:
> @noliveleger,
> I noticed in your original description:
> 
> > This class is also deactivated when 2FA is enabled on the user's account.
> 

That's ok, the class `EnketoSessionAuthentication` is not extended with the mixin `MFABlockerMixin` which provides the method `validate_mfa_not_active()` that raises an exception in case the current class needed to blocked when MFA is enabled.

noliveleger:
I've tested it locally and it worked great. LGTM

PR #3708, opened 2022-03-01
Enable optional settings for hsts

bufke:
## Description

Enable optional (and off by default) HSTS headers. Controlled via environment variables.

### Potential Alternatives

- As we add more settings, I'd be interested in using django-environ. It would result in less code around determining what is a Boolean.
- Constance is used often, but doesn't feel appropriate for http headers
- Tell users to set this in their CDN or load balancer instead. But doing this in Django is likely easier for many people including our own managed instances.

### Problems

We don't document how to do this anywhere. But it shouldn't be enabled by default either. Adding to kobo-install seems excessive. I'd be in favor of having a place for server settings documentation. Users who look for this kind of think would find it.

### Testing

This isn't easy to test locally as it requires HTTPS. I tested it by pushing it to the beta instance and inspecting the response header. It works.

### DIFF
--- a/kobo/settings/base.py
+++ b/kobo/settings/base.py
@@ -39,11 +39,9 @@ if env.str('PUBLIC_REQUEST_SCHEME', '').lower() == 'https' or SECURE_PROXY_SSL_H
     SESSION_COOKIE_SECURE = True
     CSRF_COOKIE_SECURE = True
 
-if os.environ.get("SECURE_HSTS_INCLUDE_SUBDOMAINS", "False") == "True":
-    SECURE_HSTS_INCLUDE_SUBDOMAINS = True
-if os.environ.get("SECURE_HSTS_PRELOAD", "False") == "True":
-    SECURE_HSTS_PRELOAD = True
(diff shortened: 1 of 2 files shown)


kobo/settings/base.py
```diff
@@ -42,6 +42,12 @@
     SESSION_COOKIE_SECURE = True
     CSRF_COOKIE_SECURE = True
 
+if os.environ.get("SECURE_HSTS_INCLUDE_SUBDOMAINS", "False") == "True":
+    SECURE_HSTS_INCLUDE_SUBDOMAINS = True
+if os.environ.get("SECURE_HSTS_PRELOAD", "False") == "True":
+    SECURE_HSTS_PRELOAD = True
+SECURE_HSTS_SECONDS = os.environ.get("SECURE_HSTS_SECONDS", 0)
+
```
noliveleger, line 50: FYI: We tend to use single quotes for strings, at least for new code :-) 
We keep double quotes for triple double quotes or f-strings containing single quotes.

noliveleger (changes requested):

kobo/settings/base.py
bufke: Updated and rebased on latest beta

PR #3709, opened 2022-03-02
Add django-environ and refactor environment variable setting

bufke:
## Description

- Add django-environ
- Remove dj-database-url (replaced by django-environ)
- Remove RedisHelper (replaced by django-environ cache_url)

I'd prefer to move more to env.list but it only does comma separation, not space. kpi uses space delineation.


noliveleger (approved):
LGTM

PR #3714, opened 2022-03-04
Use query params for synchronous exports

joshuaberetta:
WIP


jnm:
Closed in favor of #3716. Thanks for doing this work so quickly on short notice :heart:

PR #3715, opened 2022-03-07
Update sentry

bufke:
## Description

Update to latest Sentry client SDK.

- Logging should be automatic now, need to confirm.
- Celery integration may just work, need to confirm.
- Name env var RAVEN_DSN to SENTRY_DSN with backwards compat fallback


noliveleger:
@bufke , I also made the upgrade on KC when I upgraded from Python2/Django 1.8. 
I think it would be a good idea to review my work there to match yours and have similar settings for both apps.

bufke:
Good to know. I updated to match your syntax. We still need to discuss log level and traces sample rate.

jnm:
npm failure should be resolved in `beta` now by #3720

bufke:
Rebased!

PR #3717, opened 2022-03-09
Pass REDIS connection as an URL to `django-redis-session`

noliveleger:
@bufke said:

> `SESSION_REDIS` is being set to
> ```
> {'BACKEND': 'django_redis.cache.RedisCache', 'LOCATION': 'redis://:xxxxxx@redis-cache.kobo.private:6380/2'}
> ```
> which is right for built in django things...but not django-redis-sessions. It wants `{'host': 'host', 'port': '6379', 'db': '1', 'password': '', 'prefix': 'session', 'socket_timeout': 1}`

### DIFF
--- a/kobo/settings/base.py
+++ b/kobo/settings/base.py
@@ -786,8 +786,12 @@ MONGO_DB = MONGO_CONNECTION[MONGO_DATABASE['NAME']]
 
 MONGO_DB_MAX_TIME_MS = CELERY_TASK_TIME_LIMIT * 1000
 
-SESSION_ENGINE = "redis_sessions.session"
-SESSION_REDIS = env.cache_url("REDIS_SESSION_URL", default="redis://redis_cache:6380/2")
+SESSION_ENGINE = 'redis_sessions.session'
+# django-redis-session expects a dictionary with `url`
+redis_session_url = env.cache_url(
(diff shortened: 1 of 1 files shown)


kobo/settings/base.py
```diff
@@ -786,8 +786,14 @@ def __init__(self, *args, **kwargs):
 
 MONGO_DB_MAX_TIME_MS = CELERY_TASK_TIME_LIMIT * 1000
 
-SESSION_ENGINE = "redis_sessions.session"
-SESSION_REDIS = env.cache_url("REDIS_SESSION_URL", default="redis://redis_cache:6380/2")
+SESSION_ENGINE = 'redis_sessions.session'
+django_environ_redis_dict = env.cache_url(
+    'REDIS_SESSION_URL', default='redis://redis_cache:6380/2'
+)
+django_environ_redis_dict['url'] = django_environ_redis_dict['LOCATION']
```
bufke, line 793: This retains the LOCATION and ENGINE in the dict which aren't used by SESSION_REDIS. If for some reason there was ever a collision of values, it could introduce expected errors. I'd prefer to make a new dict instead of merging.

bufke (changes requested):

PR #3747, opened 2022-03-30
Debug toolbar enable

bufke:
## Description

Currently one must look up their internal IP address as seen by the docker container to enable django debug toolbar. Which is only ever enabled in dev anyway. Instead, we can control this via a environment variable. In dev, set `DEBUG_TOOLBAR` to "True" to enable it. This does not work in other environments, where debug toolbar isn't enabled at all.


noliveleger (approved):
LGTM. 
Way better than (my :woozy_face:) previous work ;-)

bufke:
I know way back in the day there was no SHOW_TOOLBAR_CALLBACK option and I always hated setting my always changing DHCP ip address or figuring out god knows what some proxy thought I my ip was.

Issue #1349, opened 2017-06-19
Review and merge Gallery work from ADK


pmusaraj:
The gallery in KPI is almost ready, but it requires a review and several UI fixes before it can be merged. 

@pmusaraj to do a review of the UI and frontend code. @dorey will then review the Python code and tests. This work will be done in a new branch under the current repository, and future ADK work can be committed there (this allows us to more easily retain only the file changes relevant to the gallery.)

pmusaraj:
renaming questions in the form builder and then redeploying breaks the gallery in a strange way. Submitted photos prior to renaming the questions are still there, but shown without a grouping and each on one row. Submissions after the renaming are fine. In this screenshot, new submissions are above the red line, old ones below: 

![testing adk gallery kobotoolbox 2017-06-20 13-33-20](https://user-images.githubusercontent.com/368961/27347138-7e79ee7c-55bd-11e7-96be-6419839f9ae1.jpg)

emilykferguson:
Missing Functionality:
- [ ] I'm not able to switch to group the photos by record
- [ ] Other photos for specified record

![image](https://user-images.githubusercontent.com/15673694/27546696-85e92a76-5a61-11e7-9d6c-971267cacc61.png)


UI updates:
- [ ] Question title Should be: Roboto normal, size 16px, color: #59636B 
![image](https://user-images.githubusercontent.com/15673694/27545722-8dd6ead2-5a5e-11e7-8c85-130bb91770a8.png)
- [ ] Number of images Should be: size: 14px, color: #414453 
![image](https://user-images.githubusercontent.com/15673694/27545746-a0a9675c-5a5e-11e7-8a56-b7c0deeeda5a.png)
- [ ] Width between photos should be 8px (it's 12px now)
- [ ] Search bar should say: "Filter by question" and should be a dropdown of the photo related questions
![image](https://user-images.githubusercontent.com/15673694/27545786-bf0d1e3c-5a5e-11e7-9ebc-2df42e8c37e0.png)
- [ ] Photo title font
![image](https://user-images.githubusercontent.com/15673694/27546738-a6cd9024-5a61-11e7-83e5-c4113a596629.png)
- [ ] Color of filter icon: #414453

jnm:
[...]

PR #3749, opened 2022-03-31
Add elemMatch support to mongo helper

bufke:
- [x] Write unit test(s)

## Description

- Add $elemMatch support to mongo helper. Considering what we already allow, this is probably safe. However any modification of a query param to mongo workflow merits a higher level of security review as we could potentially allow injection attacks :rotating_light: 
- Handle mongo operation failure on data api with fallback. 

## Use Cases this solves

- I want to filter submissions to only show results for a specific question inside of a repeating group (which is an array). 
- I maybe want to do this for a future gallery update #1349

### Other thoughts

- What do you think of my handling of two semi related but not 100% related issues in one PR? It felt small enough to group
- Thoughts on my doing some very very mild refactoring while in the area? Technically removing an unused import isn't 100% guaranteed to be safe in case something is importing *. But we only do that in settings.
- I decided to catch the OperationFailure, check if it was some error I know about, and if not raise a warning and give a more generic message. I didn't want to display any messages right from mongo in case there is some way to force mongo to give out unwanted data via exception message.
- I'm only 80% sure we need elemMatch for future gallery update, but it seems good to support it as a standard alone addition to the data api.

## Related issues

Internal bug tracking: KPI-BACKEND-232K

### DIFF
--- a/kpi/tests/api/v2/test_api_submissions.py
+++ b/kpi/tests/api/v2/test_api_submissions.py
@@ -615,11 +615,11 @@ class SubmissionApiTests(BaseSubmissionTestCase):
 
         data = {
             'query': f'{{"{group}":{{"$elemMatch":{{"{group}/{question}":{{"$exists":true}}}}}}}}',
-            'format': 'json'
+            'format': 'json',
         }
         response = self.client.get(self.submission_list_url, data)
         self.assertEqual(response.status_code, status.HTTP_200_OK)
-        self.assertEqual(response.data.get('count'), 1)
+        self.assertEqual(response.data[0].get('_uuid'), submission['_uuid'])
 
     def test_retrieve_submission_as_owner(self):
         """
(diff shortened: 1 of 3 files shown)


kpi/tests/api/v2/test_api_submissions.py
```diff
@@ -599,6 +599,32 @@ def test_list_submissions_asset_publicly_shared_and_shared_with_user_as_anotheru
...
+        ]
+        self.asset.deployment.mock_submissions(self.submissions)
+
+        data = {
+            'query': f'{{"{group}":{{"$elemMatch":{{"{group}/{question}":{{"$exists":true}}}}}}}}',
+            'format': 'json',
+        }
+        response = self.client.get(self.submission_list_url, data)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        if isinstance(response.data, list):
```
bufke, line 622: This is pretty bad but the alternative seems to be writing two 95% identical copies of the test for each api version which I think is worse.

noliveleger (approved):

PR #3759, opened 2022-04-06
Upgrade to python 3.10

noliveleger:
Changes:
- Celery CLI syntax. options must come before app name

Blockers: 
- ~~`pyxform==1.7.0` which requires and freezes`xlrd==1.2.0`. Python 3.9 needs 2.0+~~. Fixed by #3765
- ~~`dicttoxml` . Repo is dead. Has not been updated for 5 years. Have to look at `dict2xml` instead.~~  `dict2xml` works fine.

Removed:
- `anyjson` . Not used anywhere. Have not been updated since 2012. 
- `transifex-client`. Only support 3.9 max. [Support Python 3.10 issue](https://github.com/transifex/transifex-client/issues/332). We use GitHub Actions to compile translations.
- `nose` 

Packages that cannot be upgraded now:

| package  | pinned version | latest_version | |
| ------------- | -------------  | -------------  | ------------- |
| django-trench  | 0.2.3  | 0.3.1  | Needs rewrite in KPI. We should open another issue. |
| psycopg2  | 2.8.4 | 2.9.1  | [UTC issue with Django 2.2](https://stackoverflow.com/questions/68024060/assertionerror-database-connection-isnt-set-to-utc) |
| Werkzeug  | 2.0.3 | 2.1.1  | [issue with Django extension](https://github.com/django-extensions/django-extensions/issues/1715) |
|  PyMongo |  3.12.3 | 4.1.0 | MongoDB version issue: pymongo.errors.ConfigurationError: Server at mongo.kobo.private:27017 reports wire version 5, but this version of PyMongo requires at least 6 (MongoDB 3.6).


noliveleger:
@bufke, I've tagged the PR as "ready for review" but KoBoCAT is not ready yet. I'll leave it to your judgement if it's safe to merge it or we better wait for KoBoCAT to be ready as well.

bufke:
Cool - I expect to have some time to review Monday.

bufke (changes requested):
image :eyes: 

Need to update pytest.yml - it's tempting to have CI test using the built image so that we don't have duplicate versions but I find this can slow it down.

PR #3761, opened 2022-04-07
Add task to flush Enketo preview after some delay

jnm:
## Description

Attempt to solve a persistent Redis memory exhaustion problem caused by the accumulation of many gigabytes of Enketo previews (specifically, cached transformations of their XML). Enketo stores these for 30 days, and that lifetime is not configurable. This change uses the Enketo API to remove each preview from the cache 30 minutes after it was created.

## Related issues

enketo/enketo-express#357


noliveleger (approved):

jnm:
For some reason, this didn't really cut the mustard in production. Oh well. Any future readers: you can tame this problem by configuring a memory limit and eviction policy on your Redis cache server: https://github.com/enketo/enketo-express/issues/357#issuecomment-1333172861.

PR #3780, opened 2022-04-26
Read from `KPI_DATABASE_URL`

### LINKED ISSUE #3748: Read from `KPI_DATABASE_URL`
and fall back to `DATABASE_URL` if it's not present. This should be a few lines' change in a settings file.

See https://github.com/kobotoolbox/kobo-docker/issues/264#issuecomment-1084033452.

JacquelineMorrissette:
## Checklist

3. [x] Ensure the tests pass

## Description
Changed the settings so it would read the database URL from environment variables instead of having it hard coded

## Related issues
closes #3748

### DIFF
--- a/kobo/settings/base.py
+++ b/kobo/settings/base.py
@@ -320,10 +320,12 @@ SKIP_HEAVY_MIGRATIONS = env.bool('SKIP_HEAVY_MIGRATIONS', False)
 
 # Database
 # https://docs.djangoproject.com/en/1.7/ref/settings/#databases
+DATABASES = {}
+if 'KPI_DATABASE_URL' in os.environ:
+    DATABASES['default'] = env.db_url('KPI_DATABASE_URL')
+else:
+    DATABASES['default'] = env.db_url('DATABASE_URL')
 
-DATABASES = {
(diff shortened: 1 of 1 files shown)


kobo/settings/base.py
```diff
@@ -320,10 +320,12 @@ def __init__(self, *args, **kwargs):
 
 # Database
 # https://docs.djangoproject.com/en/1.7/ref/settings/#databases
+DATABASES = {}
+if 'KPI_DATABASE_URL' in env.ENVIRON:
+    DATABASES['default'] = env.db_url(default='KPI_DATABASE_URL')
```
noliveleger, line 325: As I said on KoBoCAT PR, the syntax should be `env.db_url('KPI_DATABASE_URL', default="sqlite:///%s/db.sqlite3" % BASE_DIR)` 
If `DATABASE_URL` is not present, this will fail :-(

noliveleger (changes requested):

jnm:
Thanks for doing this. I only just now realized that we need a change to the testing settings as well:
https://github.com/kobotoolbox/kpi/blob/e50008c694f8e55c8083223b6983a1aa4e1f114d/kobo/settings/testing.py#L8

It's a small thing so I'll just make a PR.

PR #3794, opened 2022-05-03
Upgraded to latest django-storages and inserted azure blob settings

bufke:
## Description

- [x] Upgrade to latest django-storages and port tellable storage workaround
- [x] Add support for Azure blob storage
- [x] Test in Azure
- [x] Test in S3

This PR does not address django-private-storage continuing to use only S3. Therefor this PR only partially support Azure. 

Related to  https://github.com/kobotoolbox/kobocat/pull/816

### DIFF
--- a/kobo/settings/base.py
+++ b/kobo/settings/base.py
@@ -640,10 +640,14 @@ if os.environ.get('AWS_ACCESS_KEY_ID'):
     AWS_SES_REGION_ENDPOINT = os.environ.get('AWS_SES_REGION_ENDPOINT')
 
 
+
 ''' Storage configuration '''
 if 'KPI_DEFAULT_FILE_STORAGE' in os.environ:
-    # To use S3 storage, set this to `storages.backends.s3boto3.S3Boto3Storage`
+    # To use S3 storage, set this to `kobo.apps.storage_backends.s3boto3.S3Boto3Storage`
     DEFAULT_FILE_STORAGE = os.environ.get('KPI_DEFAULT_FILE_STORAGE')
+    if DEFAULT_FILE_STORAGE == 'storages.backends.s3boto3.S3Boto3Storage':
+        # Force usage of custom S3 tellable Storage
+        DEFAULT_FILE_STORAGE = 'kobo.apps.storage_backends.s3boto3.S3Boto3Storage'
     if 'KPI_AWS_STORAGE_BUCKET_NAME' in os.environ:
         AWS_STORAGE_BUCKET_NAME = os.environ.get('KPI_AWS_STORAGE_BUCKET_NAME')
         AWS_DEFAULT_ACL = 'private'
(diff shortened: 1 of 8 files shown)


kpi/deployment_backends/kc_access/shadow_models.py
```diff
@@ -591,7 +591,7 @@ def absolute_mp3_path(self):
             content = self.get_mp3_content()
             kobocat_storage.save(self.mp3_storage_path, ContentFile(content))
 
-        if not isinstance(kobocat_storage, KobocatS3Boto3Storage):
+        if isinstance(kobocat_storage, KobocatFileSystemStorage):
```
bufke, line 594: I reversed this to make anything not local storage, use the S3 like storage. Which maybe works for azure too. I really hate this but it's the only quick way I could think to make it work. It should really not be so S3 specific and just happen to maybe work in other places.

kobo/settings/base.py
```diff
@@ -640,10 +640,14 @@ def __init__(self, *args, **kwargs):
...
 
+
 ''' Storage configuration '''
 if 'KPI_DEFAULT_FILE_STORAGE' in os.environ:
-    # To use S3 storage, set this to `storages.backends.s3boto3.S3Boto3Storage`
+    # To use S3 storage, set this to `kobo.apps.storage_backends.s3boto3.S3Boto3Storage`
     DEFAULT_FILE_STORAGE = os.environ.get('KPI_DEFAULT_FILE_STORAGE')
+    if DEFAULT_FILE_STORAGE == 'storages.backends.s3boto3.S3Boto3Storage':
+        # Force usage of custom S3 tellable Storage
+        DEFAULT_FILE_STORAGE = 'kobo.apps.storage_backends.s3boto3.S3Boto3Storage'
```
noliveleger, line 650: 👏   

I was going to say: we need to open an issue in kobo-install to support the new class. 

- https://github.com/kobotoolbox/kobo-install/blob/5d8cd0c0c9fdd265d0be9cdcdc1ba47fd435585e/templates/kobo-env/envfiles/aws.txt.tpl#L14
- https://github.com/kobotoolbox/kobo-install/blob/5d8cd0c0c9fdd265d0be9cdcdc1ba47fd435585e/templates/kobo-env/envfiles/aws.txt.tpl#L17

kpi/deployment_backends/kc_access/storage.py
```diff
@@ -40,6 +40,12 @@ def __init__(
 
 class KobocatS3Boto3Storage(S3Boto3Storage):
 
-    def __init__(self, acl=None, bucket=None, **settings):
+    def __init__(self, *args, **kwargs):
         bucket = django_settings.KOBOCAT_AWS_STORAGE_BUCKET_NAME
-        super().__init__(acl=None, bucket=bucket, **settings)
+        super().__init__(*args, **kwargs)
```
noliveleger, line 45: 🤔   `bucket` should be enforced to `django_settings.KOBOCAT_AWS_STORAGE_BUCKET_NAME`, right?
Am I missing something or it's not passed to parent anymore? 

BTW, it should `bucket_name` now, `bucket` is deprecated, I have changed it in the Django 3.2 branch.

kpi/deployment_backends/kc_access/storage.py
```diff
@@ -40,6 +40,12 @@ def __init__(
...
+
+
+class KobocatAzureStorage(AzureStorage):
+    def __init__(self, *args, **kwargs):
+        azure_container = django_settings.KOBOCAT_AZURE_CONTAINER
```
noliveleger, line 50: 🤔  Same thoughts as above?  How does KobocatAzureStorage know when to write since `azure_container` is not passed to the parent constructor?

noliveleger (changes requested):

kobo/settings/base.py
bufke: Yeah I'd rather no one know we do this horrible hack :smile: and we may some day be able to remove it. So we don't want under users changing it.

kpi/deployment_backends/kc_access/storage.py
noliveleger: Should not be something like this instead?:
```python
        kwargs['azure_container'] = django_settings. KOBOCAT_AZURE_CONTAINER
        super().__init__(*args, **kwargs)
```

kpi/deployment_backends/kc_access/storage.py
bufke: Updated, I tested by making an alternative bucket name for kc

PR #3810, opened 2022-05-16
Sentry - configure traces rate

bufke:
## Description

Allow editing Sentry traces rate with env var SENTRY_TRACES_SAMPLE_RATE which defaults to 0.05 (%)

## Implementation

I'm also refactoring it just slightly and preferring the env var SENTRY_DSN with backwards compatibility for RAVEN_DSN. Raven no longer exists.

## Related issues

Related to


noliveleger (changes requested):
I think it's time to introduce a linter for the single quote/double quotes.
But for now, let's keep the single quotes ;-)

noliveleger:
@bufke I've just found another place when `RAVEN_DSN` is still present.
https://github.com/kobotoolbox/kpi/blob/sentry-config-traces/docker/init.bash#L33-L36

noliveleger:
> > @bufke I've just found another place when `RAVEN_DSN` is still present. https://github.com/kobotoolbox/kpi/blob/sentry-config-traces/docker/init.bash#L33-L36
> 
> Should we address this in it's own PR?

Let's do this. Chances are pretty low devs use Sentry anyway.

PR #3825, opened 2022-05-25
Remove constance windows style newlines

bufke:
## Description

Fixed bug where constance config would incorrectly parse newlines

## Implementation

`\r\n` is a Windows style new line. It shouldn't be used anywhere and particularly so in unix, such as where a \n is expected. This removes it entirely and adds a legacy remover of `\r` for backwards compatibility or if somehow the windows style line gets in there.

It's not clear to me how this code used to work prior.

### DIFF
--- a/kobo/settings/base.py
+++ b/kobo/settings/base.py
@@ -259,7 +259,7 @@ CONSTANCE_CONFIG = {
         'metadata_fields_jsonschema'
     ),
     'SECTOR_CHOICES': (
-        '\r\n'.join((s[0] for s in SECTOR_CHOICE_DEFAULTS)),
+        '\n'.join((s[0] for s in SECTOR_CHOICE_DEFAULTS)),
         "Options available for the 'sector' metadata field, one per line."
     ),
     'OPERATIONAL_PURPOSE_CHOICES': (
--- a/kpi/forms/registration.py
+++ b/kpi/forms/registration.py
@@ -81,7 +81,7 @@ class RegistrationForm(registration_forms.RegistrationForm):
         # Intentional t() call on dynamic string because the default choices
         # are translated (see static_lists.py)
         self.fields['sector'].choices = (('', ''),) + tuple(
-            (s, t(s)) for s in constance.config.SECTOR_CHOICES.split('\r\n')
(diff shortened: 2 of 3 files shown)


noliveleger (approved):

jnm:
For future reference, this has nothing to do with "Windows" anything; it's part of the standard for `application/x-www-form-urlencoded`. django-constance [changed their behavior](https://github.com/jazzband/django-constance/commit/96bd3e1047d6eac995c9258859b025bb8a6e24fc) by calling the undocumented `django.utils.text.normalize_newlines()` function on values received from the admin form. I think https://github.com/jazzband/django-constance/issues/396 was the wrong action—the person reporting the issue should've changed their default value to match the HTML standard—and this `\n`-only convention likely will introduce a discrepancy with values received from any other multi-line form field from the Django admin or otherwise :shrug: 

From [internal discussion](https://chat.kobotoolbox.org/#narrow/stream/4-Kobo-Dev/topic/constance.20sector.20bug):

> The commit history shows me explicitly adding `\r`: https://github.com/kobotoolbox/kpi/commit/dbe2127fa04637e61868978a3046830582dfd042?diff=unified#diff-fda91693febb37a79e7e40d7252c7851ae75b5e07eded7111299c8a078b9a93eR230
> 
> It surprised me too, but standards dictate that (using the default form encoding) newlines in HTML `textarea`s are represented as CR+LF (`\r\n`)
> https://blogs.warwick.ac.uk/chrismay/entry/how_are_newlines/
> https://code.djangoproject.com/ticket/19251#comment:5
> https://www.w3.org/TR/REC-html40/interact/forms.html#h-17.13.4
> 
> We will need to make sure that the change doesn't break the custom metadata feature #3611. Maybe no further code changes are needed since this PR seems to accidentally work by stripping `\r`s everywhere. At a minimum, the PR description will need to be rewritten so that it's factually correct and doesn't confuse future developers. We can decide if changing the incorrect commit messages is worth the hassle.

A clarifying comment was added to the code in #3831

jnm:
I'm sure absolutely no one cares right now, but in case we ever find ourselves confounded by newline behavior (or have enough free time to propose changes to django-constance, ha ha), a `TextField` given a multi-line value in the Django admin interface (using a Linux browser, fwiw), results in `\r\n`s being saved in the model instance:
image

PR #3835, opened 2022-05-30
Use cache to display number of submissions per project

noliveleger:
## Checklist

3. [x] Ensure the tests pass

## Description

Get the number of all submissions from a pre-calculated field instead of calculating it each time for users who can view all submissions of a project.

## Additional info

It used to be this way before implementing partial permissions. Users with partial permissions still cannot benefit of the cache value because filters for partial permissions are different from one user to another.


noliveleger:
@joshuaberetta. FYI: The tests pass locally. 

Screen Shot 2022-05-30 at 11 46 44

PR #3838, opened 2022-06-01
Support for mongodb replica set

### LINKED ISSUE #3821: Mongodb connection string env var errors when multiple mongo replicas are set
## Description

`env.db_url('MONGO_DB_URL')` db_url is not intended for mongodb. It errors when the string contains commas.

## Steps to Reproduce

Set MONGO_DB_URL to `"mongodb://root:pass@mongodb-0.mongodb-headless:27017,mongodb-1.mongodb-headless:27017,mongodb-2.mongodb-headless:27017"`

## Expected behavior 

Mongo should work with the multiple replicas

## Actual behavior 

`ValueError: Port could not be cast to integer value as '27017,mongodb-1.mongodb-headless:27017,mongodb-2.mongodb-headless:27017'`

## Additional details

Perhaps we can rely on the connect string for the default database, though we need to still default to formhub when unset.

bufke:
## Description

Add support to mongo replica set by setting MONGO_DB_URL environment variable.

## Related issues

Fixes #3821
Related to https://github.com/kobotoolbox/kobocat/pull/830

### DIFF
--- a/kobo/settings/base.py
+++ b/kobo/settings/base.py
@@ -819,9 +819,12 @@ if not (MONGO_DB_URL := env.str('MONGO_DB_URL', False)):
         MONGO_DB_URL = "mongodb://%(HOST)s:%(PORT)s/%(NAME)s" % MONGO_DATABASE
     mongo_db_name = MONGO_DATABASE['NAME']
 else:
-    # Get collection name from the connection string, fallback on 'formhub' if
-    # it is empty or None
-    mongo_db_name = env.db_url('MONGO_DB_URL').get('NAME') or 'formhub'
+    # Attempt to get collection name from the connection string
+    # fallback on KPI_MONGO_NAME or 'formhub' if it is empty or None or unable to parse
+    try:
+        mongo_db_name = env.db_url('MONGO_DB_URL').get('NAME') or 'formhub'
+    except ValueError: # db_url is unable to parse replica set strings
+        mongo_db_name = env.str('KPI_MONGO_NAME', 'formhub')
 
 mongo_client = MongoClient(
     MONGO_DB_URL, connect=False, journal=True, tz_aware=True
--- a/kobo/settings/base.py
+++ b/kobo/settings/base.py
@@ -822,7 +822,7 @@ else:
     # Attempt to get collection name from the connection string
     # fallback on MONGO_DB_NAME or 'formhub' if it is empty or None or unable to parse
     try:
-        mongo_db_name = env.db_url('MONGO_DB_URL').get('NAME', env.str('MONGO_DB_NAME', 'formhub'))
+        mongo_db_name = env.db_url('MONGO_DB_URL').get('NAME') or env.str('MONGO_DB_NAME', 'formhub')
     except ValueError: # db_url is unable to parse replica set strings
         mongo_db_name = env.str('MONGO_DB_NAME', 'formhub')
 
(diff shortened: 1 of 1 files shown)


kobo/settings/base.py
```diff
@@ -819,9 +819,12 @@ def __init__(self, *args, **kwargs):
...
 else:
-    # Get collection name from the connection string, fallback on 'formhub' if
-    # it is empty or None
-    mongo_db_name = env.db_url('MONGO_DB_URL').get('NAME') or 'formhub'
+    # Attempt to get collection name from the connection string
+    # fallback on KPI_MONGO_NAME or 'formhub' if it is empty or None or unable to parse
+    try:
+        mongo_db_name = env.db_url('MONGO_DB_URL').get('NAME') or 'formhub'
+    except ValueError: # db_url is unable to parse replica set strings
+        mongo_db_name = env.str('KPI_MONGO_NAME', 'formhub')
```
noliveleger, line 827: The warning on line 794 would not be accurate anymore 

> `MONGO_DB_URL` is not found. 
>`KPI_MONGO_HOST`, `KPI_MONGO_PORT`, `KPI_MONGO_NAME`, `KPI_MONGO_USER`, `KPI_MONGO_PASS` are deprecated and will not be supported anymore soon.

I would introduce a new variable `MONGO_DB_NAME` instead. Moreover, `MONGO_DB_URL` remove the concept of two different MongoDB connections (one for KPI , one for KoBoCAT). 

Long story short, I would keep `KPI_MONGO_NAME` deprecated and would use `MONGO_DB_NAME` instead.
What about something like:

```diff
-     # Get collection name from the connection string, fallback on 'formhub' if
-     # it is empty or None
-     mongo_db_name = env.db_url('MONGO_DB_URL').get('NAME') or 'formhub'
+    # Attempt to get collection name from the connection string
+    # fallback on KPI_MONGO_NAME or 'formhub' if it is empty or None or unable to parse
+    try:
+        mongo_db_name = env.db_url('MONGO_DB_URL').get('NAME') or env.str('MONGO_DB_NAME', 'formhub')
+    except ValueError: # db_url is unable to parse replica set strings
+        mongo_db_name = env.str('MONGO_DB_NAME', 'formhub')
```

Notice I've also replaced `or 'formhub'` with `or env.str('MONGO_DB_NAME', 'formhub')` on line 825.

noliveleger (changes requested):

kobo/settings/base.py
bufke: Slight code golf `mongo_db_name = env.db_url('MONGO_DB_URL').get('NAME', env.str('MONGO_DB_NAME', 'formhub'))` since the second param is the default.

bufke:
[CI fail](https://gitlab.com/kobo_toolbox/kpi/-/jobs/2583178040) unrelated, but might need a `--force` on npm

kobo/settings/base.py
noliveleger: > So the preferred env vars that we will document are `MONGO_DB_URL` and `MONGO_DB_NAME`, right? Anything else is legacy.

Indeed

noliveleger: I think we should keep the other alternative.

Let's use as an example: 
```python 
MONGO_DB_URL = 'mongodb://user:password@mongo.kobo.tld:27017'  # no collection provided
```

```python 
In [1]: env.db_url('MONGO_DB_URL').get('NAME') or env.str('MONGO_DB_NAME', 'formhub')
In [1]: formhub

In [2]: env.db_url('MONGO_DB_URL').get('NAME', env.str('MONGO_DB_NAME', 'formhub'))
Out[2]: ''
```

Because `NAME` (always) exists in the dict the `db_url('MONGO_DB_URL')` returns. It's an empty string, thus get('NAME', ...) does not fallback on its default.

bufke: Updated and rebased

Issue #3840, opened 2022-06-01
Make Docker build smaller and faster


bufke:
## Description

We do a few things that result in a slow docker build and a very large docker image. Current builds are 812 MB as reported by dockerhub. Other Django apps like GlitchTip and Saleor are 163MB and 223MB for comparison. Some potential room for improvement:

- [x] Make use of docker multi-stage builds and use only python:3.10-slim in the final build.
- [x] Audit and remove unnecessary debian packages
- [ ] Remove any external connection aside from package repositories. For example, current builds attempt to [connect to redis for 40 seconds](https://github.com/kobotoolbox/kpi/issues/3820).
- [x] Make virtualenv read only and don't chown it (very costly)
- [ ] Audit and condense build steps
- [ ] Remove npm/node/node_modules from production image

Reducing build time will improve CI build times for making changes to testing/production environment and improve the developer experience. Reducing image size will allow KoboToolbox to run with less disk space requirements and to pull images faster, which can improve horizontal auto scaling.

bufke:
Most of this was done, can open new issues if needed.

PR #3841, opened 2022-06-01
Reducer docker image size via dockerfile multi-step build

bufke:
## Description

Reduces docker image size, to improve docker pull speed and use less disk space. Removes 860MB of disk space as reported by `docker images`. Note this isn't the same reported image as dockerhub but it should have a huge impact there as well.

## Related issues

Related to #3840

## Testing

This change has a high risk of breaking something.  I removed libproj-dev and gdal-bin which I don't see being used in Python. It's possible they are used elsewhere or I missed how they get used. It's possible they are only relevant to kobocat and were copied here.

This could have an impact on installing python packages at runtime. We will need to identify these use cases and IMO remove them if possible or else add necessary packages to the final image.


noliveleger (approved):
On my env, the image is 1GB less. Nice work ;-)

Issue #3145, opened 2021-04-22
Move away from Reflux


magicznyleszek:
## Description

The situation is that React is alive and growing, but Reflux is dead since Feb 2017 (wow!). There are no plans for future of the project: https://github.com/reflux/refluxjs/issues/550. We should move to some other library, as we will encounter more problems (e.g. currently we have a problem with npm 7+ dependency mismatch).

[Redux](https://redux.js.org/) seems to be a good candidate:

- https://www.fin.com/posts/from-reflux-to-redux/
- https://medium.com/metrosystemsro/migrating-a-react-application-from-reflux-to-redux-281c9a92b607
- https://redux.js.org/recipes/migrating-to-redux/

magicznyleszek:
@dorey did a research into switching to redux, and it seems we shouldn't go this way. @bufke suggested we should try going the react hooks way, and we will attempt that with new gallery 🤞

bufke:
Anyone interested in this topic please review my draft gallery rewrite. I'm proposing a very redux-looking style with only react hooks (not actual redux). Now would be a great time to give feedback or suggest alternatives.

PR #3842, opened 2022-06-01
Typescriptize help bubbles code

magicznyleszek:
## Description

Move help bubble code to typescript and trim half of it that we didn't use anymore.

## Related issues

Preparations for MobX test (#3145)


duvld (approved):
Looks good and tested on my machine

PR #3850, opened 2022-06-06
Upgrade css loader dependency

magicznyleszek:
## Description

Upgrade `css-loader` dependency to make sure it's compatible with Webpack v5.


noliveleger (approved):
LGTM. Could build and run KPI flawlessly

PR #3856, opened 2022-06-09
Downgrade leaflet to fix dependency conflict

magicznyleszek:
## Description

Downgrading `leaflet` a bit to match with the version that `leaflet.markercluster` is using.


magicznyleszek:
Closing this, as it doesn't fix the building error. We need to try different approach

Issue #3867, opened 2022-06-14
One command to generate requirements.txt files


bufke:
## Description

We have `make pip_compile` to regenerate the requirements.txt files. However due to https://github.com/jazzband/pip-tools/issues/1326 and needing backports-zoneinfo this is no longer possible. A workaround is to hand edit the txt files or generate them only from Python 3.8 which isn't provided by this project's Dockerfile.

## Steps to Reproduce

1. Clone repo
2. Build docker image
3. In the container run `make pip_compile`

## Expected behavior 

pip_compile should regenerate requirement.txt files without changing them based on the developer's Python version.

## Actual behavior 

backports.zoneinfo is removed, breaking Python 3.8 support.

PR #3868, opened 2022-06-14
Add pip compile script to work around pip tools python version limitation

bufke:
## Description

Work around pip tools limitation where packages that depend on a certain Python version are not carried over into the requirements txt file when the command is run from a system with certain Python versions. The desire is for the txt file to act as a universal lock file for dependencies. This change appends necessary packages that only should be installed on Python < 3.9.

## Implementation

The script unconditionally adds backports-zoneinfo under the assumption it's only run from Python 3.10+. We support 3.8, but I would expect developers to be using the Dockerfile defined version 3.10.

## Related issues

Fixes #https://github.com/kobotoolbox/kpi/issues/3867


noliveleger:
Awesome! What about KoBoCAT?

Issue #3876, opened 2022-06-17
Submission edits enter authentication loop from recreated asset snapshots


joshuaberetta:
## Description

Due to snapshots [being regenerated](https://github.com/kobotoolbox/kpi/blob/bdc86cf59ecc1f61c55227357dfcf9fea447c9a0/kpi/models/asset.py#L906) on each opened submission edit, only the most recently opened submission will submit successfully to the server. All other open edits in Enketo will enter an authentication loop on submit attempts from [this](https://github.com/kobotoolbox/kpi/blob/bdc86cf59ecc1f61c55227357dfcf9fea447c9a0/kpi/permissions.py#L350-L359) and the edits will be lost.

## Steps to Reproduce

1. Create a project with at least 2 submissions
2. Open a submission (s1) for editing, don't submit
3. Open another submission (s2) for editing from the same project
4. Attempt to submit the first submission (s1)

## Expected behavior 

Both s1 and s2 can be successfully submitted.

## Actual behavior 

Only s2 can be submitted while s1 will enter an authentication loop and prevent submitting.

<img width="556" alt="Screenshot 2022-06-17 at 16 31 57" src="https://user-images.githubusercontent.com/50016008/174319807-cc9f7073-04c9-46bf-90ec-3ca9327137e9.png">

PR #3881, opened 2022-06-21
Set `regenerate=False` as default for snapshots

### LINKED ISSUE #3880: Don't regenerate snapshots for `/xform` view
## Description

The `/xform` view is for debugging xforms in the UI, but if it regenerates a snapshot, then the same problems in #3876 occur. Remove the call with `regenerate=True`.

joshuaberetta:
## Description

Regenerating snapshots has the side effect of sending submission edits into an authentication loop (#3876). Rather default to not regenerating them.

## Related issues

closes #3880

### DIFF
--- a/kpi/models/asset.py
+++ b/kpi/models/asset.py
@@ -778,7 +778,7 @@ class Asset(ObjectPermissionMixin,
 
     @property
     def snapshot(self):
-        return self._snapshot(regenerate=False)
+        return self._snapshot()
 
     @property
     def tag_string(self):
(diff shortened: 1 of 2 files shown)


jnm:
Consult https://chat.kobotoolbox.org/#narrow/stream/4-Kobo-Dev/topic/AssetSnapshot before proceeding

jnm:
Closing as we'll regenerate but explicitly allow multiple snapshots per version: #3996

PR #3883, opened 2022-06-21
Fix KPI issue when the app is not responding anymore even under low load

### LINKED ISSUE #3566: uWSGI hangs sometimes
## Description

uWSGI worker hangs sometimes after respawning. See WID 4 below with RSS 0 and VSZ 0. 

image
## Steps to Reproduce

Unfortunately, couldn't replicate reliably by load testing. 

## Additional details

Self hosted KPI version 2.021.45b 
Sentry not enabled

This seems similar to https://github.com/kobotoolbox/kobocat/issues/377 & https://github.com/DataDog/dd-trace-py/issues/491

Why not set `lazy-apps = true` in uwsgi.ini ?

noliveleger:
## Description

Fix 502/504 errors that KPI starts to return after running few hours/days even if the load is very low.

## Additional info

502/504 can still happens under high load, obviously. This PR addresses a particular case:
uWSGI locks and it is not able to respawn new workers (thus not able to serve any requests anymore). 
It happens because of deadlocks in Python. 
See https://github.com/unbit/uwsgi/issues/1599#issuecomment-336584282 for more details. 

Moreover, to avoid respawning many workers very often, the `cheaper` algorithm has been replace with `busyness` (instead `spare`). It does respond better under high load.

## Related issues

Fixes #3566


kobo/wsgi.py
```diff
@@ -1,16 +1,24 @@
...
+
+# Close DB connections to avoid deadlock when uwsgi forks process.
+# see https://github.com/unbit/uwsgi/issues/1599#issuecomment-336584282
+
+# Close PostgreSQL connections
+for conn in connections.all():
+    conn.close()
+
+# Close MongoDB connections
+settings.MONGO_CONNECTION.close()
```
bufke, line 22: This only exists in settings/testing. It would be good if this code didn't mind if the database doesn't exist either.

noliveleger: Revert the PR to draft until I have tested the changes again.

noliveleger: Ok to be reviewed

PR #3931, opened 2022-07-08
Use mfa_code_length in UI

magicznyleszek:
## Description

Display the same MFA code length in UI as set on backend.


noliveleger:
@magicznyleszek I think we should remove the validation on the length of the token too. Because backup code can be longer than 6 (or whatever value provided by the back end).

magicznyleszek:
@noliveleger good catch. Pushed a commit that fixes that. There was a magic number `6` used in the code

duvld (approved):

PR #3950, opened 2022-07-22
Treat unmarked timezone as UTC, convert _submission_time from UTC to local

### LINKED ISSUE #3941: `_submission_time` needs to be treated as UTC in the UI
## Summary

Please treat all timestamps lacking a time zone offset as UTC. The server always uses Universal Time unless there's an explicit indication otherwise.

## Details

In the response from `https://kf.kobotoolbox.org/api/v2/assets/[asset-id]/data/`, there are a variety of timestamps for each submission. Some of them include time zone offsets:
* `"start": "2022-07-17T09:43:10.266-04:00",`
* `"end": "2022-07-17T13:53:11.481+00:00",`

The front end converts these to local time when displaying them in the table view. So far, so good.

However, `_submission_time` does **not** include an offset:
```json
"_submission_time": "2022-07-17T13:53:11",
```
This is where the trouble starts: the front end interprets this as a local time and displays it in the table view without any conversion. This leads to confusion because `_submission_time` is actually UTC.

Ideally, this would be changed to
```
"_submission_time": "2022-07-17T13:53:11+00:00",
```
…but that would involve rewriting millions of records, or adding an ugly hack to rewrite these timestamps on the fly as they are served by the API.

p2edwards:
## Description

Convert `_submission_time` from UTC to local time in the table view.

## Related issues

Fixes #3941

### DIFF
--- a/jsapp/js/components/submissions/table.es6
+++ b/jsapp/js/components/submissions/table.es6
@@ -725,8 +725,7 @@ export class DataTable extends React.Component {
             }
             if (
               q.type === META_QUESTION_TYPES.start ||
-              q.type === META_QUESTION_TYPES.end ||
-              q.type === ADDITIONAL_SUBMISSION_PROPS._submission_time
+              q.type === META_QUESTION_TYPES.end
             ) {
               return (
                 <span className='trimmed-text'>
@@ -734,6 +733,13 @@ export class DataTable extends React.Component {
                 </span>
               );
             }
+          }
+          if (key === ADDITIONAL_SUBMISSION_PROPS._submission_time) {
+            return (
+              <span className='trimmed-text'>
+                {formatTimeDateShort(row.value)}
+              </span>
+            );
           }
           if (typeof(row.value) === 'object' || row.value === undefined) {
             const repeatGroupAnswers = getRepeatGroupAnswers(row.original, key);
--- a/jsapp/js/utils.ts
+++ b/jsapp/js/utils.ts
@@ -33,7 +33,7 @@ const unzonedTime = new RegExp(/^\d{4}-\d\d-\d\dT\d\d:\d\d:\d\d$/);
  * Given a string like '2022-07-22T00:00:00', appends a 'Z' so it can be interpreted as UTC instead of "local".
  */
 function presumeUtc(timeStr: string): string {
-  // Expecting like	'2022-01-01T00:00:00', or that plus a timezone offset.
+  // Expecting like '2022-01-01T00:00:00', or that plus a timezone offset.
   if (unzonedTime.test(timeStr)) {
     return timeStr + 'Z';
   } else {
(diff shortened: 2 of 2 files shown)


p2edwards:
Oh, oops! This should be a PR on beta, not master.

jsapp/js/utils.ts
```diff
@@ -27,35 +27,49 @@ export function notify(msg: string, atype = 'success') {
...
 }
 
+const unzonedTime = new RegExp(/^\d{4}-\d\d-\d\dT\d\d:\d\d:\d\d$/);
+
+/**
+ * Given a string like '2022-07-22T00:00:00', appends a 'Z' so it can be interpreted as UTC instead of "local".
+ */
+function presumeUtc(timeStr: string): string {
+  // Expecting like	'2022-01-01T00:00:00', or that plus a timezone offset.
+  if (unzonedTime.test(timeStr)) {
```
p2edwards, line 37: I'm not sure if this regexp test is really needed, let me know what you think.
Originally I just wrote `if (timeStr.length === 19)`, but I thought this would make the intent a little more clear.

jsapp/js/components/submissions/table.es6
```diff
@@ -735,6 +734,13 @@ export class DataTable extends React.Component {
               );
             }
           }
+          if (key === ADDITIONAL_SUBMISSION_PROPS._submission_time) {
+            return (
+              <span className='trimmed-text'>
+                {formatTimeDateShort(row.value)}
+              </span>
+            );
+          }
```
p2edwards, line 743: I moved this because `q` wasn't ending up defined for `_submission_time`.

jsapp/js/utils.ts
magicznyleszek: I like that function :) And it's a perfect case for you to write some unit tests! Check out some other tests in `/jsapp/js/utils.tests.es6` - the `formatSeconds` one is pretty concise. Think about some edge cases :) 

To run tests you have to run `npm run test` in terminal. Note that sometimes these tests fail (like 1% chance) due to test suite not finding the headless chrome instace - if that happens, just re-run tests :)

magicznyleszek (changes requested):

jsapp/js/utils.ts
p2edwards: Ah, I like it too, but now I've found the built-in moment function that makes it not needed.

p2edwards: By the way, we don't currently have unit tests for [these formatTime functions](https://github.com/kobotoolbox/kpi/blob/21c5b2241d0c7287728eff223de29862234a7aca/jsapp/js/utils.ts#L30-L60). To me that makes sense since we'd need to mock the local timezone (or even the current date), and since they're just moment wrappers, it doesn't seem worth the complexity. What do you think?

magicznyleszek: It's great you've found it. And agree, no point in testing moment library functions :)

magicznyleszek (approved):

PR #3956, opened 2022-07-28
Resize all image elements to fit in their container by default

### LINKED ISSUE #3945: Images are not scaled in the single submission modal
A large image dominates the single submission modal and actually goes outside the table boundaries:
image
image

_Originally posted by @amks1 in https://github.com/kobotoolbox/kpi/issues/3672#issuecomment-1189652765:_
> It's not necessary to return the resized image, there can be a CSS style with:
> ```
> max-width: 100%;
> object-fit: contain;
> ```
> for the `` tag which would solve the interface problem.
> 
> image

p2edwards:
## Description

Prevent large images from stretching across table cell boundaries in the single submission modal.

## Related issues

Fixes #3945

## Developers

Introduces new global css on `img` tags.

```scss
img {
  max-width: 100%;
  object-fit: contain;
}
```

We can override this as needed on individual components, but it's a good default. ([ref](https://github.com/kobotoolbox/kpi/issues/3945#issuecomment-1191152571))

## Screenshots
<details>
  <summary>Before</summary>
  Large image breaks the layout
</details>
<details>
  <summary>After</summary>
  Large image is contained in the table
</details>

### DIFF
--- a/jsapp/scss/stylesheets/partials/_base.scss
+++ b/jsapp/scss/stylesheets/partials/_base.scss
@@ -65,6 +65,11 @@ video {
   vertical-align: middle;
 }
 
+img {
+  max-width: 100%;
+  object-fit: contain;
+}
+
 fieldset {
   border: 0;
   margin: 0;


duvld (commented):
looks good :)

PR #3959, opened 2022-07-28
Hide "Security" tab entirely if it is useless

jnm:
## Description

In account settings, hide "Security" entirely when it is not useful instead of leaving it visible but disabled.

### DIFF
--- a/jsapp/js/components/account/accountSidebar.tsx
+++ b/jsapp/js/components/account/accountSidebar.tsx
@@ -78,16 +78,19 @@ export default class AccountSidebar extends React.Component<
           </bem.FormSidebar__label>
           }
 
-          <bem.FormSidebar__label
-            m={{selected: this.isSecuritySelected()}}
-            href={'#' + ROUTES.SECURITY}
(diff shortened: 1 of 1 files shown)


jsapp/js/components/account/accountSidebar.tsx
```diff
@@ -78,16 +78,19 @@ export default class AccountSidebar extends React.Component<
...
-            <bem.FormSidebar__labelText>
-              {t('Security')}
-            </bem.FormSidebar__labelText>
-          </bem.FormSidebar__label>
+          { /* hide "Security" entirely if nothing there is available */
+            envStore.isReady && envStore.data.mfa_enabled &&
+            <bem.FormSidebar__label
+              m={{selected: this.isSecuritySelected()}}
+              href={'#' + ROUTES.SECURITY}
+              disabled={ !(envStore.isReady && envStore.data.mfa_enabled) }
```
jnm, line 86: left this in case we revert later

jnm:
This has been included in release 2.022.24a

PR #3960, opened 2022-07-28
added calculate and barcode question types to enable filtering in table view

amks1:
## Checklist

3. [ ] Ensure the tests pass

## Description

`calculate`  and `barcode` type questions are not 'filterable' in the table view by default. There is no reason why this can't be done, because the responses to these questions are always of type `TEXT`. This code addition has been tested in a production environment for several months and it works very well.

I cannot think of a situation where it is undesirable to have this option, but if there is some such case, the PR can be rejected.

### DIFF
--- a/jsapp/js/components/submissions/table.es6
+++ b/jsapp/js/components/submissions/table.es6
@@ -770,6 +770,8 @@ export class DataTable extends React.Component {
       QUESTION_TYPES.date.id,
       QUESTION_TYPES.time.id,
       QUESTION_TYPES.datetime.id,
+      QUESTION_TYPES.barcode.id,
+      QUESTION_TYPES.calculate.id,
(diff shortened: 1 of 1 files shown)


jnm:
@magicznyleszek if you decide to merge this, please edit the title and description to make them suitable for the table of changes that we generate for our release notes. The title should be succinct and use the imperative: something like "Allow filtering for barcode and calculate questions in the table view" would be okay. There may not be a need for a `## Description` at all, unless it adds something that a regular user reading the release notes would find helpful. The original text can be moved outside `## Description`, or that heading could simply be deleted.

magicznyleszek:
Closing this in favor of #3973. I tested the idea, it works great. Just decided to additionally move things around :wink:

Thanks :pray:

PR #3978, opened 2022-08-19
Avoid crash when kobocat database does not exist

jnm:
## Description

In a fresh-install situation, such as when `kobo-docker/.vols` has been removed, the kobocat database may not yet exist when kpi configuration checks run. That led to a `django.db.utils.OperationalError: FATAL:  database "kobocat" does not exist` error caused by `check_for_kpi_data_in_kobocat_database()`. This change allows the configuration checks to pass without a kobocat database. Both kobocat and kpi use the `wait_for_postgres.bash` [script](https://github.com/kobotoolbox/kobo-docker/blob/aee0ecfd5005cf2c666c35afac1ca3075c328506/scripts/wait_for_postgres.bash) to create their databases, but each application only creates its own database (for example, kpi does not create a database for kobocat).

### DIFF
--- a/kpi/utils/two_database_configuration_checker.py
+++ b/kpi/utils/two_database_configuration_checker.py
@@ -2,6 +2,7 @@
 from django.conf import settings
 from django.core.checks import Error
 from django.db import connections
+from django.db.utils import OperationalError
 from django.utils.translation import gettext as t
 
 from kpi.management.commands.is_database_empty import (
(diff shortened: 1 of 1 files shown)


jnm:
Explanation for why it worked before: `./manage.py wait_for_database` did not invoke the configuration checker under Django 2.
```
root@kpi:/srv/src/kpi# git diff kpi/utils/two_database_configuration_checker.py
diff --git a/kpi/utils/two_database_configuration_checker.py b/kpi/utils/two_database_configuration_checker.py
index c9b451b..7d810f1 100644
--- a/kpi/utils/two_database_configuration_checker.py
+++ b/kpi/utils/two_database_configuration_checker.py
@@ -141,6 +141,7 @@ class TwoDatabaseConfigurationChecker:
             self.check_for_kpi_data_in_kobocat_database,
             self.check_for_migration_from_shared_database,
         ]
+        print('does this even happen?')
         for check in checks:
             if not check():
                 # Each check depends on the one before; it makes no sense to
root@kpi:/srv/src/kpi# ./manage.py wait_for_database && echo great
great
root@kpi:/srv/src/kpi# django-admin version
2.2.27
```

noliveleger:
Test it from a brand new environment and worked perfectly

noliveleger (approved):

PR #3979, opened 2022-08-22
Add Stripe data sync to track public kf.kobotoolbox.org usage limit exemptions

bufke:
## Description

Add Stripe model syncing for tracking kf.kobotoolbox.org usage exemptions.

## Technical decisions

- stripe django app is added conditionally. I usually dislike this, but in this case I think it's appropriate. Self host users probably don't want this large app installed.
- Organizing code by features organization and stripe. Stripe depends on organization. Maybe some day organization can be useful by itself.
- Some porting of code from [GlitchTip stripe module](https://gitlab.com/glitchtip/glitchtip-backend/-/tree/master/djstripe_ext) but not 1-1.


kobo/apps/stripe/serializers.py
```diff
@@ -10,11 +10,13 @@ class Meta:
...
 
 class PlanSerializer(serializers.ModelSerializer):
+
     product = BaseProductSerializer()
 
     class Meta:
         model = Plan
-        fields = ('id', 'nickname', 'amount', 'metadata', 'product')
+        exclude = ('djstripe_id',)
+        # fields = ('id', 'nickname', 'amount', 'metadata', 'product')
```
bufke, line 19: No commented out code please

kobo/apps/stripe/views.py
```diff
@@ -32,6 +32,11 @@ class SubscriptionViewSet(
     mixins.UpdateModelMixin,
     viewsets.GenericViewSet,
 ):
+
+    renderer_classes = (
```
bufke, line 36: Why don't we want the defaults?

kobo/apps/stripe/views.py
```diff
@@ -11,3 +24,63 @@ class ProductViewSet(viewsets.ReadOnlyModelViewSet):
...
+    mixins.UpdateModelMixin,
+    viewsets.GenericViewSet,
+):
+    model = Subscription
+    serializer_class = SubscriptionSerializer
+    lookup_field = 'id'
+    permission_classes = (IsAuthenticated,)
+
+    def get_queryset(self):
+        user_orgs = self.request.user.organizations_organization.values_list(
```
bufke, line 46: Coercing to a python list is probably resulting in an extra query. When you add tests please make use of `self.assertNumQueries` to ensure we aren't calling extra queries. Be extra careful of lists calling one query per row. In the future subscription history may get complicated and it's a good habit to just always be in.

kobo/apps/stripe/views.py
```diff
@@ -11,3 +24,63 @@ class ProductViewSet(viewsets.ReadOnlyModelViewSet):
...
+            'id',
+            flat=True,
+        ).all()
+        queryset = Subscription.objects.filter(
+            customer__subscriber_id__in=user_orgs,
+        )
+        return queryset
+
+    def list(self, request, *args, **kwargs):
```
bufke, line 55: We do we need these custom list, retrieve functions? Is there a way we could use the default implementation instead?

PR #3984, opened 2022-08-30
Create endpoint for total and daily submission counts

JacquelineMorrissette:
## Checklist

3. [x] Ensure the tests pass

## Description
Create an API endpoint for daily and total counts of submissions


kpi/tests/utils/mock.py
```diff
@@ -50,7 +50,7 @@ class MockAttachment(MP3ConverterMixin):
     Mock object to simulate ReadOnlyKobocatAttachment.
     Relationship with ReadOnlyKobocatInstance is ignored but could be implemented
     """
-    def __init__(self, pk: int, filename: str, mimetype: str = None, **kwargs):
+    def  __init__(self, pk: int, filename: str, mimetype: str = None, **kwargs):
```
noliveleger, line 53: Please remove the extra space ;-)

kpi/serializers/v2/asset_counts.py
```diff
@@ -0,0 +1,23 @@
...
+        fields = ('daily_submission_counts', 'total_submission_count')
+
+    def get_daily_submission_counts(self, asset):
+        today = timezone.now().date()
+        if 'days' in self.context:
+            start_date = today - relativedelta(days=int(self.context['days']))
+        else:
+            start_date = today - relativedelta(days=31)
+        daily_counts = asset.deployment.get_daily_counts(timeframe=(start_date, today))
+        return daily_counts
```
noliveleger, line 20: We should validate that `days` is an integer and a setting instead of hard-coding 31. 

Something like

```python 
    days = self.context['days']   # it is always set is the viewset , see comment below
    try:
        days = int(days)
    except ValueError:
        raise serializers.ValidationError(t('not an integer... blah blah blah'))
    
    start_date = today - timedelta(days=days)
```

kpi/serializers/v2/asset_counts.py
```diff
@@ -0,0 +1,23 @@
+from dateutil.relativedelta import relativedelta
```
noliveleger, line 1: Why using `relativedelta`? What's the advantage against `datetime.timedelta`?

kpi/tests/api/v2/test_api_assets.py
```diff
@@ -495,6 +496,54 @@ def test_asset_version_id_and_content_hash(self):
         self.assertEqual(response.data['version__content_hash'],
                          self.asset.latest_version.content_hash)
 
+    def test_count_endpoint(self):
```
noliveleger, line 499: This test is with the owner if I'm not wrong. Would you mind to add tests with AnonymousUser and regular user? 
For the regular user, the test could start without 'view_submissions' then retry with `view_submissions` to ensure it pass with proper permission.

kpi/views/v2/asset.py
```diff
@@ -343,6 +366,21 @@ def content(self, request, uid):
...
+        methods=['GET'],
+        renderer_classes=[renderers.JSONRenderer],
+    )
+    def counts(self, request, *args, **kwargs):
+        asset = self.get_object()
+        if not asset.has_deployment:
+            raise Http404
+        context = self.get_serializer_context()
+        if 'days' in request.query_params:
+            context['days'] = request.query_params['days']
```
noliveleger, line 380: You can simplify this with `context['days'] = request.query_params.get('days', settings.MY_SETTING)`

I cannot find it but I was sure you already create a setting for a similar number of days? Maybe in another PR?

kpi/views/v2/asset.py
```diff
@@ -343,6 +366,21 @@ def content(self, request, uid):
             'data': asset.to_ss_structure(),
         })
 
+    @action(
+        detail=True,
+        methods=['GET'],
+        renderer_classes=[renderers.JSONRenderer],
```
noliveleger, line 372: We may need to restrict the access with specific permission classes...

kpi/deployment_backends/kobocat_backend.py
```diff
@@ -682,6 +684,16 @@ def get_attachment_objects_from_dict(self, submission: dict) -> QuerySet:
...
 
+    def get_daily_counts(self, timeframe: tuple[date, date]) -> dict:
+        daily_counts = ReadOnlyKobocatDailyXFormSubmissionCounter.objects.values(
+            'date', 'counter'
+        ).filter(
+            xform_id=self.xform_id,
+            date__range=timeframe,
+        )
+        data = {str(count['date']): count['counter'] for count in daily_counts}
+        return data
```
noliveleger, line 695: Hmmm.. We will need to handle the users with partial permissions. 
Let's discuss about it.

noliveleger (changes requested):

noliveleger (changes requested):

kpi/deployment_backends/kobocat_backend.py
noliveleger: As per our discussion, users with partial submissions cannot user daily counters. 
We are going to use `MongoDB.aggregate()` to calculate the correct value on the fly. Not as fast as daily submission counters but after few tests on production servers, it is pretty fast thanks to indexes.

kpi/serializers/v2/asset_counts.py
```diff
@@ -14,12 +15,9 @@ class Meta:
...
 
     def get_daily_submission_counts(self, asset):
-        today = timezone.now().date()
+        today = timezone.now().date() + datetime.timedelta(days=1)
         request = self.context['request']
-        if 'days' in self.context:
-            start_date = today - relativedelta(days=int(self.context['days']))
-        else:
-            start_date = today - relativedelta(days=31)
+        start_date = today - datetime.timedelta(days=int(self.context['days']))
```
noliveleger, line 20: We should validate `self.context['days']` is an integer and raise a `ValidationError` if it's not the case.

kpi/tests/api/v2/test_api_asset_counts.py
```diff
@@ -0,0 +1,261 @@
...
+            },
+            {
+                '__version__': self.asset.latest_deployed_version.uid,
+                'q1': 'a1',
+                '_submitted_by': '',
+                '_submission_time': '2022-09-12T17:31:33',
+            },
+        ]
+
+        self.asset.deployment.mock_submissions(submissions)
```
noliveleger, line 52: I think these lines could be in `setUp()` since we are using always the same content in this class.

kpi/tests/api/v2/test_api_asset_counts.py
```diff
@@ -0,0 +1,261 @@
...
+        count_url = reverse(
+            self._get_endpoint('asset-counts-list'),
+            kwargs={'parent_lookup_asset': self.asset_uid}
+        )
+        # Test anonymous user with public access to data
+        anonymous_user = User.objects.get(username='AnonymousUser')
```
noliveleger, line 156: You can import the utility instead: `from kpi.utils.object_permission import get_anonymous_user`

kpi/tests/api/v2/test_api_asset_counts.py
```diff
@@ -0,0 +1,261 @@
...
+        self.asset.assign_perm(anonymous_user, PERM_VIEW_ASSET)
```
noliveleger, line 157: `assign_perm` is smart enough to assign implied permissions. `PERM_VIEW_ASSET` is implied when assigning `PERM_VIEW_SUBMISSIONS`.

kpi/tests/api/v2/test_api_asset_counts.py
```diff
@@ -0,0 +1,261 @@
...
+        self.asset.assign_perm(anonymous_user, PERM_VIEW_SUBMISSIONS)
```
noliveleger, line 158: `PERM_VIEW_ASSET` only ;-)

kpi/tests/api/v2/test_api_asset_counts.py
```diff
@@ -0,0 +1,261 @@
...
+        ]
+
+        self.asset.deployment.mock_submissions(submissions)
+        count_url = reverse(
+            self._get_endpoint('asset-counts-list'),
+            kwargs={'parent_lookup_asset': self.asset_uid}
+        )
+        # Test another user without perms
+        self.client.login(username='anotheruser', password='anotheruser')
+        response = self.client.get(count_url)
```
noliveleger, line 208: Assert the `anotheruser` does **NOT** have `PERM_VIEW_ASSET`

kpi/tests/api/v2/test_api_asset_counts.py
```diff
@@ -0,0 +1,261 @@
...
+
+        self.asset.deployment.mock_submissions(submissions)
+        count_url = reverse(
+            self._get_endpoint('asset-counts-list'),
+            kwargs={'parent_lookup_asset': self.asset_uid}
+        )
+        response = self.client.get(count_url)
+        # Test another user with perms
+        anotheruser = User.objects.get(username='anotheruser')
+        self.asset.assign_perm(anotheruser, PERM_VIEW_SUBMISSIONS)
```
noliveleger, line 257: 👀 `PERM_VIEW_ASSET`

kpi/views/v2/asset_counts.py
```diff
@@ -0,0 +1,68 @@
...
+    </pre>
+
+    > Example
+    >
+    >       curl -X GET https://[kpi]/api/v2/assets/aSAvYreNzVEkrWg5Gdcvg/counts/?days=7
+
+    ### CURRENT ENDPOINT
+    """
+    parent_model = Asset
+    permission_classes = [ViewSubmissionPermission]
```
noliveleger, line 58: The permission class should be restricting on `PERM_VIEW_ASSET` only ;-)

noliveleger (changes requested):

kpi/serializers/v2/asset_counts.py
JacquelineMorrissette: The number of days gets turned into a string when passed as a URL parameter, so it needs to be converted regardless, would it be better to convert it to an integer in the viewset?

noliveleger: @JacquelineMorrissette , I meant we should validate the value of `self.context['days']` is digit, not characters before cast to `int()`. If it's not then raise the ValidationError. 
So no, do not convert it in the viewset. Let the serializer do the job is the validation passes.

kpi/tests/api/v2/test_api_asset_counts.py
noliveleger: You got it right. Sorry for the confusion

kpi/views/v2/asset_counts.py
noliveleger: Once again, my mistake. That's the correct permission class.

kpi/serializers/v2/asset_counts.py
```diff
@@ -0,0 +1,45 @@
...
+
+class AssetCountsSerializer(serializers.Serializer):
+    daily_submission_counts = serializers.SerializerMethodField()
+    total_submission_count = serializers.SerializerMethodField()
+
+    class Meta:
+        fields = ('daily_submission_counts', 'total_submission_count')
+
+    def get_daily_submission_counts(self, asset):
+        today = timezone.now().date() + datetime.timedelta(days=1)
```
noliveleger, line 18: `get_daily_counts` already handles it for us. No need to add the extra timedelta.

noliveleger (approved):

Issue #3988, opened 2022-09-08
Separate Celery queues for REST Services (Hook), imports, and exports


jnm:
A slow REST Service cripples the service by clogging the queue, which is shared with critical functions like importing and exporting.

Step 1 will be separating the queues; following that, we'll have to devise a way to prevent one misbehaving REST Service from blocking the others.

PR #3989, opened 2022-09-08
Move background tasks for REST Services to their own queue

noliveleger:
## Checklist

3. [x] Ensure the tests pass

## Description

Use a different queue to send submissions to REST Services external servers to avoid blocking imports and exports running in the background.

## Related issues

Part of #3988

### DIFF
--- a/kobo/apps/hook/tests/test_email.py
+++ b/kobo/apps/hook/tests/test_email.py
@@ -34,7 +34,7 @@ class EmailTestCase(HookTestCase):
     def test_notifications(self):
         self._create_periodic_task()
         first_log_response = self._send_and_fail()
-        failures_reports.delay()
+        failures_reports.apply_async(queue='kpi_low_priority_queue')
         self.assertEqual(len(mail.outbox), 1)
 
         expected_record = {
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,6 +44,7 @@ RUN mkdir -p "${NGINX_STATIC_DIR}" && \
     mkdir -p ${CELERY_PID_DIR} && \
     mkdir -p ${SERVICES_DIR}/uwsgi && \
     mkdir -p ${SERVICES_DIR}/celery && \
+    mkdir -p ${SERVICES_DIR}/celery_low_priority && \
     mkdir -p ${SERVICES_DIR}/celery_beat && \
     mkdir -p "${INIT_PATH}"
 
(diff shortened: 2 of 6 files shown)


docker/run_celery.bash
```diff
@@ -11,7 +11,8 @@ exec celery -A kobo worker --loglevel=info \
     --hostname=kpi_main_worker@%h \
     --logfile=${KPI_LOGS_DIR}/celery.log \
     --pidfile=/tmp/celery.pid \
-    --exclude-queues=sync_kobocat_xforms_queue \
+    --queues=kpi_queue \
+    --exclude-queues=kpi_low_priority_queue \
```
jnm, line 15: I have to admit: I don't fully understand these options or why we went with `--exclude-queues` in the first place, if indeed we only wanted to run tasks from one queue. `celery worker --help` says this, which doesn't really answer my questions:
```
Queue Options:
  --purge, --discard
  -Q, --queues COMMA SEPARATED LIST
  -X, --exclude-queues COMMA SEPARATED LIST
  -I, --include COMMA SEPARATED LIST
```
Maybe this helps?
> A worker instance can consume from any number of queues. By default it will consume from all queues defined in the [task_queues](https://docs.celeryq.dev/en/stable/userguide/configuration.html#std-setting-task_queues) setting (that if not specified falls back to the default queue named celery).
>
> You can specify what queues to consume from at start-up, by giving a comma separated list of queues to the [-Q](https://docs.celeryq.dev/en/stable/reference/cli.html#cmdoption-celery-worker-Q) option: …

I'm just wondering if it would be clearer to say which queues we were consuming instead of which ones we weren't, especially if we add more than two queues in the future.

kobo/apps/hook/tests/test_email.py
```diff
@@ -34,7 +34,7 @@ def _create_periodic_task(self):
     def test_notifications(self):
         self._create_periodic_task()
         first_log_response = self._send_and_fail()
-        failures_reports.delay()
+        failures_reports.apply_async(queue='kpi_low_priority_queue')
```
jnm, line 37: It's too bad that we have to specify the queue when calling the task instead of in the task definition itself, but I don't see any way to do the latter. I found https://stackoverflow.com/questions/64606915/specify-queue-for-a-task-class-in-celery, but it looks like what they're doing there is completely undocumented.

docker/run_celery.bash
noliveleger: I have to admit I was not sure myself. I kept `--exclude-queues` because of current syntax. But it would be more obvious to get rid of it.

kobo/apps/hook/tests/test_email.py
noliveleger: I did not want to change the code a lot for this urgent matter but we could replace the `shared_task` decorator with the `celery.app` decorator and use the `queue` argument, e.g:

```
@shared_task
def failures_reports():
    ...
    
    failures_reports.apply_async(queue='kpi_low_priority_queue')
```

with 

```
from kobo.celery import celery_app

@celery_app.task(queue='kpi_low_priority_queue')
def failures_reports():
    ...
    failures_reports.delay()
```

or use maybe some [manual routing](https://docs.celeryq.dev/en/stable/userguide/routing.html#manual-routing)?

jnm: Is `@celery_app.task(queue='kpi_low_priority_queue')` documented? It's probably staring me in the face :flushed:

docker/run_celery.bash
jnm: OK, let's stick with the tried-and-true for this quick change.

jnm:
Just a note: the django-constance autoscale settings affect both queues in my testing. It doesn't really matter because these autoscale settings are removed in #3863.

Issue #3987, opened 2022-09-07
Brackets in choice value breaks drop-down filter in table


jeverling:
## Description

Hi all, when you have a XLSFrom with choices, where the values contains brackets (and presumably other special characters), the drop-down filter in the data table doesn't work.

For example in the _choices_ sheet:

| list_name | name                  | label                 |
|-----------|-----------------------|-----------------------|
| products  | Groundnuts (In Shell) | Groundnuts (In Shell) |
| products  | Groundnuts (Shelled)  | Groundnuts (Shelled)  |

If you now try to filter using the table view in kobotoolbox, it doesn't show any of the matching choices because the filtering uses a regex instead of the exact value, e.g.: `query={"product":{"$regex":"Groundnuts (In Shell)","$options":"i"}}`

![image](https://user-images.githubusercontent.com/121946/188883346-424e3ef1-57c3-47fb-bc68-39bdef4f8bd7.png)

This only happens when a form is imported from a XLSForm, the formbuilder generates safe _XML values_, e.g. `groundnuts__in_shell`.

Of course it's much better to use safe values for the choices, but if the drop-down didn't use a regex for filtering and instead just the exact value, it should also work for strings containing spaces and other characters.

Is there a specific reason to use a regex in this place? Would you consider replacing this with an exact query instead of the regex?

jeverling:
Hi Leszek, thanks for your answer!
Weird, I tried uploading the following form to https://kf.kobotoolbox.org/, and was able to upload and deploy it: 
[form_broken_choice_values.xlsx](https://github.com/kobotoolbox/kpi/files/9560478/form_broken_choice_values.xlsx)

Anyway, that choice values with spaces will not be allowed is a good solution.

PR #3999, opened 2022-09-16
Better filtering for select_multiple columns in Data Table

### LINKED ISSUE #3897: Data table not filtering correctly.
## Description

There is a bug on the data table showing submitted form information athttps://kobo.humanitarianresponse.info
On a column for a multiple-choice question, when one is selected starting with the letter A, the filter does not work.

### LINKED ISSUE #3922: Add select filter in Data Table for select_multiple columns
## Description

Currently we have a header filter for `select_one`, we should add similar filter for `select_multiple`.

magicznyleszek:
## Description

Changes the text input filtering into a dropdown with all available choices. Additionally fixes the issue with filtering by not filtering by labels properly.

## Related issues

Fixes #3897
Fixes #3922

### DIFF
--- a/jsapp/js/components/submissions/table.es6
+++ b/jsapp/js/components/submissions/table.es6
@@ -770,9 +770,10 @@ export class DataTable extends React.Component {
     let frozenColumn = tableStore.getFrozenColumn();
 
     columnsToRender.forEach(function (col) {
-      // TODO: see if this can work for select_multiple too
-      // See: https://github.com/kobotoolbox/kpi/issues/3922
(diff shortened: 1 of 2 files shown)


jnm:
@magicznyleszek, does this really fix #3987 a.k.a.
> Additionally fixes the issue with filtering by not filtering by labels properly.

:question: 

This attracted my attention while writing the release notes. I don't see how the changes in the diff would correct the name vs. label problem.

magicznyleszek:
@jnm it fixes it, because these small changes causes the multiple filtering no longer being a text input (causing that bug), but a dropdown (displays Labels, but filters via names)

PR #4003, opened 2022-09-18
Make `envStore` use MobX

duvld:
Moved to #4030.

### DIFF
--- a/jsapp/js/envStore.ts
+++ b/jsapp/js/envStore.ts
@@ -1,6 +1,6 @@
-import Reflux from 'reflux';
 import {actions} from 'js/actions';
 import type {EnvironmentResponse} from 'js/dataInterface';
+import {makeAutoObservable} from 'mobx';
 
 export interface EnvStoreDataItem {
   value: string;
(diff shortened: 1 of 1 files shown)


jsapp/js/envStore.ts
```diff
@@ -115,20 +112,19 @@ class EnvStore extends Reflux.Store {
     }
 
     this.isReady = true;
-    this.trigger(this.data);
   }
 
-  getLanguages() {
+  get languages() {
```
p2edwards, line 117: Interesting rename. I didn't see any usage for `getLanguages(` or `languages(`, internally or externally -- other class functions here are just using `this.data.all_languages` directly.

magicznyleszek: I think that store was built with all different useful methods that some didn't find any use. I think it would make sense to update all `this.data.all_languages` usages to `this.languages`, or to just remove this method :P Especially that `feature/nlp` will remove `all_languages` from here :)

jsapp/js/envStore.ts
```diff
@@ -14,38 +14,38 @@ export interface EnvStoreFieldItem {
...
-
-  getProjectMetadataField(fieldName: string): EnvStoreFieldItem | boolean {
+  public all_languages: EnvStoreDataItem[] = [];
+  public interface_languages: EnvStoreDataItem[] = [];
+  public submission_placeholder = '';
+  public mfa_localized_help_text: {[name: string]: string} = {};
+  public mfa_enabled = false;
+  public mfa_code_length = 6;
+  public stripe_public_key: string | null = null;
+  public stripe_pricing_table_id: string | null = null;
```
p2edwards, line 38: What's the reason for marking these all as `public`?

jsapp/js/envStore.ts
duvld: Removing this makes the most sense to me 👍

duvld: Made this change over in #4030

PR #4024, opened 2022-09-19
Replace test (video) file with one having content

jnm:
…and reduce size from 10M+ to <600K (although this space will not be reclaimed within Git).

Credit to @ninapaley for said content


jnm:
> While smaller and more fun, I suggest clipping it to just a couple seconds containing voice audio. This would save on testing cpu cycles. We may end up processing this a few times in ffmpeg in unit tests. If you change it, do rewrite history.

The video content is intentional, but I truncated the clip to 16 seconds, down from the original 59s.

PR #4025, opened 2022-09-20
Fix "different root names" Enketo error when editing a submission

### LINKED ISSUE #3996: Enketo "Error trying to parse XML record. Different root nodes" when editing
1. `AssetSnapshot` should be many-to-one with `AssetVersion` instead of one-to-one
1. We should start cleaning up `AssetSnapshots` frequently so their quantity doesn't get (more) out of control (kpi#2434)
1. We should not delete anything when [making](https://github.com/kobotoolbox/kpi/blob/14aeab5d6015a3cb0ab9ea236f27f58ba98af5a5/kpi/models/asset.py#L945) a new snapshot, even if the snapshot uses the same version
  * Use case: at times the XML root node name of a submission can differ from the XML for a particular `AssetVersion`. We [deal with that](https://github.com/kobotoolbox/kpi/blob/14aeab5d6015a3cb0ab9ea236f27f58ba98af5a5/kpi/views/v2/data.py#L628-L640) by changing the XML retrieved from the `AssetVersion` to explicitly match the root node name of the submission. Previously, it was fine to store that manipulated XML in an `AssetSnapshot`. However, with kpi#3877 (fix for the editing problem), we now (almost) never regenerate a snapshot unless the `AssetVersion` changes.
1. We should return to ~~re~~generating a new snapshot for each edit, with the appropriately adjusted root node name.

Internal discussion: https://chat.kobotoolbox.org/#narrow/stream/4-Kobo-Dev/topic/AssetSnapshot

noliveleger:
## Checklist

3. [x] Ensure the tests pass

## Description

Always regenerate (and update the root of the `<instance>` node) of the XML of the form when editing a submission to match the submission root name.

## Related issues

Closes #3996

### DIFF
--- a/kpi/models/asset.py
+++ b/kpi/models/asset.py
@@ -906,10 +906,6 @@ class Asset(ObjectPermissionMixin,
             self.create_version()
 
     @property
-    def snapshot(self):
-        return self._snapshot(regenerate=False)
-
-    @property
     def tag_string(self):
         try:
             tag_list = self.prefetched_tags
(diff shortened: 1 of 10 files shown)


jnm (approved):
:champagne:

Kalyan-Lama:
Community: https://community.kobotoolbox.org/t/error-trying-to-parse-xml-record-different-root-nodes/33586

PR #4026, opened 2022-09-22
Show more text in Data Table column headers

### LINKED ISSUE #4000: Make Data Table 🧮 columns headers 💆‍♂️ text wrap ↩️
Make Data Table columns headers text wrap if too long (which is almost always) with a limit to 3 lines.

magicznyleszek:
## Description

Display up to 3 lines of text for column headers, trimmed if longer. Users can still see the whole text when hovering.

## Additional details

This change is open to future improvements - mainly the number of lines being displayed can be configurable easily.

## Related issues

Fixes #4000


jsapp/js/components/submissions/table.scss
```diff
@@ -91,6 +91,7 @@ $z-table-frozen: 10;
         font-size: 20px;
         color: colors.$kobo-gray-55;
         vertical-align: top;
+        display: inline-block;
```
duvld, line 94: Fixes Safari bug :D

PR #3976, opened 2022-08-17
Superuser user list endpoint and export

joshuaberetta:
🚧

PR #4038, opened 2022-10-04
Remove newlines in user details export

joshuaberetta:
## Description

Remove newline characters in user details CSV export that cause issues in Excel, Data Studio, etc.

## Related issues

fix for #3976


kobo/apps/superuser_stats/tasks.py
```diff
@@ -212,6 +212,13 @@ def flatten_metadata_inplace(metadata):
             if isinstance(v, dict) and 'value' in v:
                 metadata[k] = v['value']
 
+    def get_row_value(row, col):
```
noliveleger, line 215: What about using annotations typing even if it's a small inner function?

noliveleger (changes requested):

kobo/apps/superuser_stats/tasks.py
joshuaberetta: Good idea 👍

noliveleger (approved):

PR #4033, opened 2022-09-28
Use latest deployed version of the form for submission edits

joshuaberetta:
## Description

Fix previous change in kpi#3734 that used latest version of the form for submission edits, not latest **deployed** version and add a `version_uid` attribute to the submission API edit and view response for easier debugging and transparency.

PR #4047, opened 2022-10-07
Update submission version with edit to match form version

### LINKED ISSUE #4041: Editing a submission should update __version__ to match form version used for edit
## Description

Editing a submission with a different form version to the one it was originally submitted with may result in missing data in the export when `fields_from_all_versions=True`. Formpack [uses the `__version__` attribute for each submission](https://github.com/kobotoolbox/formpack/blob/482172c7f84974b73383ca7f0b5012810dad5799/src/formpack/reporting/export.py#L124-L138) to know how to format the data for exporting. The `__version__`'s value should be updated to match the current version of the form used to edit the submission.

## Steps to Reproduce

1. Have a form (`v1`) with a media question (`image_1`)
1. Submit data (`s1`)
1. Add another media question (`image_2`) to the form (`v2`)
1. Edit `s1` (using `v2`) to add an image for `image_2` question
1. Export data with with `fields_from_all_versions=True`
1. Only `image_1` question has data
1. Export data with `fields_from_all_versions=False`
1. `image_1` and `image_2` have data

## Additional details

Discussion: https://chat.kobotoolbox.org/#narrow/stream/4-Kobo-Dev/topic/Submission.20edits.20and.20attachments/near/157775

joshuaberetta:
## Description

Update the submission `__version__` value to use the latest deployed version of the form (#4033) for both single and bulk edit operations. This ensures that, if submissions are edited with a form version different to their original, the added data will be included in the export when `fields_from_all_versions` is set to `True`.

## Related issues

closes #4041

### DIFF
--- a/kpi/deployment_backends/kobocat_backend.py
+++ b/kpi/deployment_backends/kobocat_backend.py
@@ -131,6 +131,8 @@ class KobocatDeploymentBackend(BaseDeploymentBackend):
         Returns:
             dict: formatted dict to be passed to a Response object
         """
+        version_uid = self.asset.latest_deployed_version.uid
+
         submission_ids = self.validate_access_with_partial_perms(
             user=user,
             perm=PERM_CHANGE_SUBMISSIONS,
@@ -182,6 +184,10 @@ class KobocatDeploymentBackend(BaseDeploymentBackend):
             )
             deprecated_id_or_new.text = instance_id.text
             instance_id.text = uuid_formatted
+
+            # Update the `__version__` value in the submission to match the
+            # version of the form used for editing
+            xml_parsed.find('__version__').text = version_uid
 
             # If the form has been updated with new fields and earlier
             # submissions have been selected as part of the bulk update,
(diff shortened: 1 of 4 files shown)


jnm:
Left some comments in https://github.com/kobotoolbox/kpi/issues/4041#issuecomment-1421675422.

I'm concerned especially about the bulk editing part of this. I really don't think someone who chooses to bulk edit responses to  Question A, shared by both versions in a situation like this:

* Version 1: Question A | Question B | Question C
* Version 2: Question A | Question D | Question E

…would then expect that all submissions collected with Version 1 of the form would now be treated as if they were from Version 2, thus preventing responses to Question B and Question C from being included in exports.

It seems like a fair amount of work, but the right thing to do would be to restrict bulk editing only to questions that appear in all versions used by the selected submissions. If submissions from Version 1 and Version 2 in example above were selected for bulk editing, we should only allow responses to Question A to be altered.

If someone needs to change the effective version of the form used by a submission, that should be a deliberate action.

PR #4072, opened 2022-10-11
Refactor ApiTokenDisplay component

magicznyleszek:
## Additional details

Testing out functional components syntax that we want to adapt for all future code.


jsapp/js/components/apiTokenDisplay.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  useEffect(() => {
+    if (isVisible && token === null) {
+      const fetchToken = async () => {
+        setIsFetching(true);
+        try {
+          const result = await dataInterface.apiToken();
+          setToken(result.token);
+        } catch (error) {
+          setToken(null);
```
bufke, line 33: This is unnecessary. token must be null already to get here.

Also, when the user leaves this page and goes back, it would be null again until clicking show. Then refetch. Right? It should be possible to sometimes update the token when reloading the page.

jsapp/js/components/apiTokenDisplay.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const [isVisible, setIsVisible] = useState(false);
+
+  const toggleTokenVisibility = () => {
+    setIsVisible(!isVisible);
+  };
+
+  useEffect(() => {
+    if (isVisible && token === null) {
+      const fetchToken = async () => {
+        setIsFetching(true);
```
bufke, line 28: Over-engineering alert :exclamation: probably don't make this change

I believe a more direct translation of setState to hooks would be to use one useState with an object. [Example](https://stackoverflow.com/a/53575023). This allows you to set multiple state objects in command. Just like setState({foo: 1, bar: 2}) would change both foo and bar in one command.

When the user clicks toggle, I would prefer to see one synchronous action result in one state transition. As the state gets more complex, useReducer could be considered (but not in this simple component). 

Action > Synchronous state change > Async Side effect > Synchronous state change.

But all of this philosophy is to deal with more complex state transitions. It helps the developer understand that one action causes X state changes all in one place (rather than searching multiple places for it).

Why I don't think this change is truly necessary to make, is that it will alter your effect usage. Running that effect on every state change, IMO, will be more like class based components. But what you're doing here isn't really wrong. It's justifiable to create multiple states and trigger the effect function only with one state change. Sure you could refactor and throw in a branching statement to only do things on the state changes you care about - but meh.

bufke (changes requested):

jsapp/js/components/apiTokenDisplay.tsx
magicznyleszek: Remove the unnecessary nullification 🦅 👁️

jsapp/js/components/apiTokenDisplay.tsx
```diff
@@ -0,0 +1,62 @@
...
+        try {
+          const result = await dataInterface.apiToken();
+          setToken(result.token);
+        } finally {
```
bufke, line 32: I don't think this works, you need a catch even if the catch doesn't do anything.

magicznyleszek: It does work, try in console: `try {throw 'foo'} finally {console.log('bar')} `

bufke: It's valid syntax, but it's not ignoring the error. See how the uncaught error is present with finally but not with catch.

image

[Further reading](https://stackoverflow.com/questions/5764107/is-try-without-catch-possible-in-javascript)

jsapp/js/components/apiTokenDisplay.tsx
magicznyleszek: I will not do this then :)

PR #4084, opened 2022-10-17
Fix Account Plan route styles

magicznyleszek:
## Description

Fixes a small visual issue that caused `account/plan` route specific styles to be applied randomly in other places of the app.


duvld (approved):
Thanks for fixing the color of the range index!

PR #4088, opened 2022-10-20
single submission view of transcripts and translations

dorey:
builds a sorted object of keys to specify which rows should be displayed


jnm:
The tip of this branch reappears as 6f45214818b2049612d40ad3df1eebc3bf35b3fe and was merged as part of #4107

Issue #3977, opened 2022-08-18
Refactor modal dialog system


p2edwards:
## Description

Refactor modal dialog system, dropping alertify.dialog.

We probably want:

- New styles from #3643
- Better behavior for nested modals
  - e.g. `'Esc'` should close one popup at a time
- Use typescript, mobx, for new or rewritten components

Internal discussion: https://chat.kobotoolbox.org/#narrow/stream/7-UX.2FUI

## Related Issues

Implement #3643 - Component design for modals
Also part of #3837 - replace/drop Alertify

Follow-up to PR https://github.com/kobotoolbox/kpi/pull/3972#discussion_r949377341 
> `// TODO: port from alertify.dialog to new modal system`

magicznyleszek:
Some notes:
- it would be useful to check if we can refactor `jsapp/js/components/common/modal.es6` to be a useful "base" component, instead of deprecating it and writing a component from scratch
- looking at specs, some configuration would be needed:
  - size: `standard` and `large`
  - title with:
    - optional left side icon
    - optional "x" close button
    - name (mostly a `string`, but should work with HTML content)
  - closing by outside click is optional (enabled by default)
  - closing by esc key is optional (enabled by default)
- other requirements:
  - we need a way to programatically open (and close) a modal (i.e. call a function)
  - we need a way to know if a modal is open
  - the above suggest we need a uid for modals (`utils.ts` have `generateUid` function that would be good enough, opening a modal should assing uid to id and return it)
  - some min/max sizes (see [Figma file](https://www.figma.com/file/Q04O151tZ2r4ObqL860Nqu/KoboToolbox-Design-System-2.0?node-id=874%3A539))
  - header will always be there (TBD?)
  - content below header can take whole available area (e.g. a video modal)
  - footer is optional, and should probably allow any content, but with suggestion to use predefined simple bem components:
    - footer wrapper
    - footer left, center, and right containers (all optional) - so we can put buttons inside of these

magicznyleszek:
Another note: besides those custom modals we have, we also need to create a prompt modal that will be using the base conponent(?)

Issue #3769, opened 2022-04-14
Use designbook instead of manual design-system we have


magicznyleszek:
## Description

We should start using [storybook.js](https://storybook.js.org) and migrate all `jsapp/js/designSystem`

bufke:
Should we deploy the beta branch to say storybook.kbtdev.org? Adding multiple branches to multiple sites would get more complicated.

Issue #2809, opened 2020-09-25
Improve keyboard navigation in Form Manager UI


magicznyleszek:
## Description

We need to remove global disabling of `:focus` styles - instead we can try using https://github.com/ten1seven/what-input or just enable them and make prettier than default.

This is needed because of general accessibility and to enable users to fill in "Share Form" or other components with just a keyboard.

## Current status

There are focus styles for most text inputs (a `1px` blue underline). There are none for buttons, radios, checkboxes and links.

## Additional details

Work connected to #553

I don't recall there is a name for "the part of KPI that isn't Form Builder or Login", so I will be using "Form Manager".

magicznyleszek:
AFAIK what this still needs to be closed is:
- test tabbing throught whole UI
- fix ugly styling (invisible size, maring or padding mishaps that are ok without focus outline, but look weird otherwise)
- add `tabindex="0"` to all things that should be tabbable but aren't (e.g. `<KoboDropdown>`s trigger and options)

PR #4092, opened 2022-10-21
Refactor modal

magicznyleszek:
## Description

Introducing new modal component with demo available at `/design-system`.

## Related issues

Part of #3977


jsapp/js/components/modals/koboModal.tsx
```diff
@@ -0,0 +1,48 @@
...
+  onRequestClose: () => void;
+  size?: KoboModalSize;
+  children: React.ReactNode;
+  /**
+   * Whether it should close when user hits Esc or clicks on overlay.
+   * NOTE: disabling Esc key may introduce an accessibility issue.
+   */
+  isDismissableByDefaultMeans?: boolean;
+  /** Renders a `data-testid` attribute in the DOM. */
+  testId?: string;
```
bufke, line 23: Was there any conversation about this? It seems like a new concept. Why use testId over id?

bufke (approved):
It works as I'd expect a react modal to work and it looks like something in Kobo. 

If you wanted to take it a step further, a big thing with modals is [accessibility](https://reactcommunity.org/react-modal/accessibility/) and how it grabs focus. I think you'd get that for "free" with react modal. You could put some tab indexes in the design system demo and see if it works. We don't often consider this, but we should start in newer components.

Prettier changed a couple things for me in kobo modal and kobo modal content, did you run it before committing?

Good to merge, pending a review of my comments.

bufke:
Out of scope, but kind of bad. This change actually increases our bundle size due to `import DesignSystemApp from 'js/designSystem/designSystemApp';` in main.es6. Every time our design system app gets larger, it makes the entire JS app larger.

Could be solved either by [storybook](https://github.com/kobotoolbox/kpi/issues/3769) or lazy loading the design system, which involves adding it to a route.

jsapp/js/components/modals/koboModal.tsx
p2edwards: Is it `data-cy` with a new name?

magicznyleszek: There is a built-in `testId` property in `react-modal` and I thought it would be better to use it instead of not having any cypress selector. But I did find a way to use `data-cy` with it :)

magicznyleszek:
@bufke I did enable the built-in accessibility now. We have this old issue #2809 for re-enabling visible focus styles, it would be a great start in this area 🤞  As for the bundle size - it is bad, and the `design-system` shouldn't be a part of the main bundle for sure, but I think we just need to find some time for implementing that storybook in the future to address it properly :)

PR #4098, opened 2022-10-31
Do not override files when using Azure Storage

noliveleger:
## Checklist

3. [x] Ensure the tests pass

## Description

Make Django Storage behave the same with AWS S3 and Azure Blob Storage.

## Additional info

`AZURE_OVERWRITE_FILES` can be set with an environment variable (with the same name). It is `False` by default.

## Related issues

Related to kobotoolbox/kobocat#853


kobo/settings/base.py
```diff
@@ -714,7 +714,7 @@ def __init__(self, *args, **kwargs):
         AZURE_ACCOUNT_KEY = env.str('AZURE_ACCOUNT_KEY')
         AZURE_CONTAINER = env.str('AZURE_CONTAINER')
         AZURE_URL_EXPIRATION_SECS = env.int('AZURE_URL_EXPIRATION_SECS', None)
-        AZURE_OVERWRITE_FILES = True
+        AZURE_OVERWRITE_FILES = env.bool('AZURE_OVERWRITE_FILES', False)
```
bufke, line 717: What's a use case where an admin would choose to set to True - and thus break their own kobo instance? If there is no use case, then delete this line. The [default](https://django-storages.readthedocs.io/en/latest/backends/azure.html) is False.

kpi/models/import_export_task.py
```diff
@@ -689,15 +685,79 @@ def _run_task(self, messages):
...
 
+    def get_absolute_filename(self, filename: str) -> str:
+        """
+        Get absolute filename related to storage root.
+        """
+
+        storage_class = self.result.storage
+        filename = self.result.field.generate_filename(self, filename)
+
+        # We used to call `self.result.save(filename, ContentFile(b''))` before
```
bufke, line 710: My take on this style of comment to consider and do with what you may. Concise is easier to read. You and I care deeply about the history and why changes are made. But is it really relevant what we used to do? I would reword to be more terse about needing this for Azure. If it's relevant to keep the old example, keep it but in the style of "This seems like a good idea but won't work". If not needed, that's even better and you can ditch it. Do keep the django bug reference if it's still relevant here.

kpi/models/import_export_task.py
```diff
@@ -689,15 +685,79 @@ def _run_task(self, messages):
...
+        # ```
+        # But it does not work with AzureStorage.
+        # AzureStorage raises an error if we try to write into an existing file
+        # when setting `AZURE_OVERWRITES_FILES` is False.
+        # Unfortunately, `self.result.save()` does few things that we need to
+        # reimplement here:
+        # - Create parent folders (if they do not exist) for local storage
+        # - Get a unique filename if filename already exists on storage
+
+        # Copied from `FileSystemStorage._save()`
```
bufke, line 728: :cry: guess this is what you meant by "Not super proud". If you had more time to give, this would be where to spend it. But we need a quick solution obviously.

bufke (changes requested):
This generally is fine to merge after responding to my comments. Feel free to merge after unless you want to discuss anything further.

kpi/models/import_export_task.py
noliveleger: Indeed :-(

kobo/settings/base.py
noliveleger: Gone!

PR #4102, opened 2022-11-02
React Router 6 + Modular account pages

### LINKED ISSUE #4103: Refactor stores.session to use MobX (drop Reflux)
## Description
See code at js/stores.es6 and all the places that uses it.

## Additional details
Followup to https://github.com/kobotoolbox/kpi/issues/3145.

bufke:
- Refactor accounts sidebar to be functional component and implements the plan link condition the way it was originally intended, via the subscriptionStore.
- Split up routes and route constants so that *mostly* only the account directory knows about it. There are some exceptions. The router imports accounts without lazy loads. Sometimes ACCOUNT_ROUTES is imported elsewhere to link to.
- Move account related components into account directory
- Mostly replaces session store reflux with mobx version. The js version of accountSettingsRoute still uses the reflux version. This fixes #4103. This does result in loading account data twice but it's only temporary and isn't a very expensive api call.
- accountSettingsRouteNew.tsx will replace accountSettingsRoute
- Make Account interface closer to reality, or at least the reality that isn't true but we assume is true.

>  {/* There shouldn't be a nested <a> tag here, NavLink already generates one */}

Should be fixed by someone better at styling.


jsapp/js/components/permissions/sharingForm.es6
```diff
@@ -18,7 +19,8 @@ import {
   ASSET_TYPES,
   ANON_USERNAME,
 } from 'js/constants';
-import {ROUTES} from 'js/router/routerConstants';
+import './sharingForm.scss';
+import {ACCOUNT_ROUTES} from 'js/account/routes';
 import './sharingForm.scss';
```
p2edwards, line 24:
```suggestion
import {ACCOUNT_ROUTES} from 'js/account/routes';
import './sharingForm.scss';
```
duplicate import

jsapp/js/editorMixins/editableForm.es6
```diff
@@ -68,7 +67,8 @@ const RECORDING_SUPPORT_URL = 'recording-interviews.html';
 
 export default assign({
   componentDidMount() {
-    this.props.router.setRouteLeaveHook(this.props.route, this.routerWillLeave);
+    // TODO router6 might be a Prompt?
+    // this.props.router.setRouteLeaveHook(this.props.route, this.routerWillLeave);
```
p2edwards, line 71: Is this still TODO?

p2edwards (approved):

PR #4108, opened 2022-11-08
Fix NLP merge related typescript errors

bufke:
This fixes some type definition issues from merging NLP into beta.


jsapp/js/components/library/singleCollectionStore.ts
```diff
@@ -24,6 +24,7 @@ import type {AssetTypeName} from 'js/constants';
 
 interface SingleCollectionStoreData {
   isFetchingData: boolean;
+  isTranscriptInProgress: boolean;
```
magicznyleszek, line 27: This shouldn't even be in this file. I see @jnm added this line and the other one with `isTranscriptInProgress`. Both of these should be removed.

jsapp/js/actions.d.ts
```diff
@@ -149,13 +149,7 @@ export namespace actions {
...
       setMessageReadTime: GenericDefinition;
     };
-    const library: {
-      moveToCollection: GenericDefinition;
-      subscribeToCollection: GenericDefinition;
-      unsubscribeFromCollection: GenericDefinition;
-      searchMyLibraryMetadata: GenericDefinition;
-      searchMyLibraryAssets: GenericDefinition;
-    };
+    const library: any;
```
magicznyleszek, line 152: This is fine 👌

magicznyleszek (changes requested):

jsapp/js/components/library/singleCollectionStore.ts
jnm: Right, that's junk that I added and should've removed after changing it to `isPollingForTranscript`

PR #4122, opened 2022-11-11
4116 remove columns from xlsform exports

dorey:
Removes columns starting with `$` from XLSForm exports


dorey:
Accidentally opened to `main` :anger:

PR #4118, opened 2022-11-10
Always set the `__version__` calculate field with the latest version identifier

noliveleger:
## Checklist

1. [x] If you've added code that should be tested, add tests
2. [ ] If you've changed APIs, update (or create!) the documentation
3. [x] Ensure the tests pass
4. [x] Make sure that your code lints and that you've followed [our coding style](https://github.com/kobotoolbox/kpi/blob/master/CONTRIBUTING.md)
5. [x] Write a description of your work suitable for publishing on [our forum](https://community.kobotoolbox.org/tag/release-notes)
6. [x] Mention any related issues in this repository (as #ISSUE) and in other repositories (as kobotoolbox/other#ISSUE)
7. [ ] Open an issue in the [docs](https://github.com/kobotoolbox/docs/issues/new) if there are UI/UX changes

## Description

Do not expose `__version__` field anywhere except in submission data and omit its alternatives such as `_version_`, `_version_00x_` etc...

## Related issues

Closes #1465

PR #4123, opened 2022-11-11
remove fields starting with $ from XLSForm exports

dorey:
Removes fields starting with `$` from XLSForm exports


jnm:
We can rely on the unit test in #4118

PR #4160, opened 2022-11-24
Fix 404 when date range parameters are omitted

### LINKED ISSUE #4157: Admin users reports always return 404 when one of the range parameters is not specified.
## Steps to Reproduce

1. Go to super user statistic report 
2. Specify a start date but no end date (e.g: `?start_date=2022-11-01`)
3. Click on csv link

## Expected behavior 

CSV is generated and can be downloaded after a while

## Actual behavior 

CSV is not generated and returns a 404

noliveleger:
## Checklist

3. [ ] Ensure the tests pass

## Description

Reports are generated even if do not contain data (thus do not return **404 Not Found** endlessly)

## Related issues

Closes #4157
Blocked by #4159

### DIFF
--- a/kobo/apps/superuser_stats/views.py
+++ b/kobo/apps/superuser_stats/views.py
@@ -36,6 +36,7 @@ def country_report(request):
     per country as reported by the user
     """
     today = timezone.now().date()
+    first_of_month = today.replace(day=1)
     base_filename = 'country-report_{}_{}_{}.csv'.format(
         re.sub('[^a-zA-Z0-9]', '-', request.META['HTTP_HOST']),
         today,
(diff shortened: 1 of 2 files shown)


kobo/apps/superuser_stats/views.py
```diff
@@ -152,8 +156,8 @@ def domain_report(request):
         f'<code style="background: lightgray">start_date</code> parameter to <code style="background: lightgray">YYYY-MM-DD</code> and/or the '
         f'<code style="background: lightgray">end_date</code> parameter to <code style="background: lightgray">YYYY-MM-DD</code>.<br><br>'
         f'<b>Example:</b><br>'
-        f'<a href="{url}?start_date={today}&end_date={tomorrow}">'
-        f'  {url}?start_date={today}&end_date={tomorrow}'
+        f'<a href="{url}?start_date={today}&end_date={today}">'
+        f'  {url}?start_date={today}&end_date={today}'
```
JacquelineMorrissette, line 160: I used the today and tomorrow's dates in the examples to make sure it was clear which date should be the earlier one. Should we keep it that way? Otherwise it looks good to merge

noliveleger: Right. As you've noticed I removed `tomorrow` everywhere. We were using `tomorrow` to be sure to include `today` inside the date range. There were an issue with that when filtering SQL queries with date objects and datetime fields. 


```python
today = '2022-11-25'
tomorrow = '2022-11-26'

# Working with datetime
Model.objects.filter(date_created__range=(today, today))  # return None
Model.objects.filter(date_created__range=(today, tomorrow))  # return every object created today

# Working with dates
Model.objects.filter(date_created__date__range=(today, today))  # return every object created today
Model.objects.filter(date_created__date__range=(today, tomorrow))  # return every object created today and tomorrow
```

It was a little bit confusing, especially when you only want to narrow down with the end date.
For example: User wants everything until '2022-11-10', they had to use `?end_date=2022-11-11` 

If you want, we can change the default value of `start_date` and update the example accordingly.  Also, write down explicitly the `end_date` is inclusive.

JacquelineMorrissette (approved):

PR #4163, opened 2022-11-28
Add django-allauth for Azure authentication (SSO) and email verification

bufke:
## Description

- Migrate to django-allauth accounts framework
- Require email verification when changing an email
- Require POST for logout.
- Make email field of Current User ViewSet (/me/) read-only
- Provide new /me/emails/ ViewSet to add/list/delete unconfirmed emails, in order to change the user's primary email.
- Add Microsoft Graph (Azure) OAuth provider
- Change user interface for editing user settings and security settings. Move email and change password to security page.

## Implementation Details

django-allauth was deemed the best option for it's support of many oauth clients. Drawbacks include a lack of true multi-tenant configuration and lack of proper OIDC support. We could add both features in theory, ourselves.

- Port django registration templates/logic to allauth
- Make logout require a POST (best practice) frontend was modified to support this
- Add Microsoft Azure Active Directory authentication via django-allauth package but disabled by default

Renamed branch https://github.com/kobotoolbox/kpi/pull/4034

[Docs change](https://github.com/kobotoolbox/docs/issues/314)


jsapp/js/account/security/apiToken/apiTokenSection.component.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import ApiTokenDisplay from 'js/components/apiTokenDisplay';
```
bufke, line 2: Let's move this as a sibling component, it isn't reused

jsapp/js/account/security/apiToken/apiTokenSection.module.scss
```diff
@@ -0,0 +1,55 @@
...
+    input {
+      font-family: mdl.$font_mono;
+    }
+  }
+}
+
+.optionsSection {
+  text-align: right;
+  // FIXME: Why does text-align do nothing?
+  display: ruby;
```
magicznyleszek, line 49: This is the first time I see `ruby` 🧑‍🎓 So why is it here? I've read (right now) that this is being used mostly for annotations in east asian documents

magicznyleszek: Perhaps it would make more sense if we'd use something more generic?

magicznyleszek: But if not, try `ruby-align` instead of `text-align`

jsapp/js/account/security/apiToken/apiTokenSection.module.scss
```diff
@@ -0,0 +1,55 @@
...
+  flex: 3;
+}
+
+.optionsSection > * {
+  margin-left: 12px;
```
magicznyleszek, line 54: `sizes.$x12`

jsapp/js/account/security/email/emailSection.component.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const [session] = useState(() => sessionStore);
+
+  const [email, setEmail] = useState<EmailState>({
+    isLoading: true,
+    emails: [],
+    newEmail: '',
+  });
+
+  useEffect(() => {
+    if (email.isLoading) {
```
magicznyleszek, line 25: Maybe naming it `!email.isInitialised` would make more sense here?

jsapp/js/account/security/email/emailSection.module.scss
```diff
@@ -0,0 +1,60 @@
...
+}
+
+.optionsSection {
+  flex: 3;
+  text-align: right;
+
+  button {
+    margin-top: sizes.$x12;
+    // FIXME: Why does text-align do nothing?
+    right: -78%;
```
magicznyleszek, line 43: Maybe because of flex? You could change `.optionsSection` to `display: flex` and align items with flex properties

jsapp/js/account/security/email/emailSection.module.scss
```diff
@@ -0,0 +1,60 @@
...
+
+  p.blurb {
+    flex: 14;
+    margin: 0;
+  }
+
+  > i {
+    flex: 1;
+    color: colors.$kobo-amber;
+  }
```
magicznyleszek, line 59: Wouldn't just `flex: 1` on `p.blurb` do the work (without `flex: anything` on `> i`)?

jsapp/js/account/security/mfa/mfaSection.component.tsx
```diff
@@ -163,21 +164,23 @@ export default class SecurityRoute extends React.Component<{}, SecurityState> {
...
-          {t(
-            'Two-factor authenication (2FA) verifies your identity using an authenticator application in addition to your usual password. ' +
-              'We recommend enabling two-factor authenication for an additional layer of protection.'
-          )}
-        </bem.SecurityRow__description>
+          <bem.SecurityRow__description>
+            {t(
+              'Two-factor authentication (2FA) verifies your identity using an authenticator application in addition to your usual password. ' +
+                'We recommend enabling two-factor authentication for an additional layer of protection.'
+            )}
```
magicznyleszek, line 171: I wonder if this `+` would work with how the strings are extracted for transifex… Maybe it would be better to do this:
```ts
{t(
  'Two-factor authentication (2FA) verifies your identity using an authenticator application in addition to your usual password.'
)}
{t(
  'We recommend enabling two-factor authentication for an additional layer of protection.'
)}            
```

jsapp/js/account/security/sso/ssoSection.component.tsx
```diff
@@ -0,0 +1,75 @@
...
+      : [];
+
+  const disconnectSocialAccount = () => {
+    if (socialAccounts.length) {
+      const socialAccount = socialAccounts[0];
+      deleteSocialAccount(socialAccount.provider, socialAccount.uid).then(
+        () => {
+          sessionStore.refreshAccount();
+        }
+      );
```
magicznyleszek, line 22: Would `.then(sessionStore.refreshAccount.bind(sessionStore))` work?

jsapp/js/account/security/sso/ssoSection.component.tsx
```diff
@@ -0,0 +1,75 @@
...
+      <div className={styles.titleSection}>
+        <h2 className={styles.title}>{t('Single-Sign On')}</h2>
+      </div>
+      {socialAccounts.length === 0 ? (
+          <div className={styles.bodySection}>
+            <div className={styles.securityDescription}>
+              {t(
+                "Connect your KoboToolbox account with your organization's identity provider for single-sign on (SSO). Afterwards, you will only " +
+                  'be able to sign in via SSO unless you disable this setting here. This will also update your email address in case your current ' +
+                  'address is different.'
```
magicznyleszek, line 37: Same potential risk with the `+` strings. Maybe we should ask John about this transifex extracting…

jsapp/js/account/security/sso/ssoSection.module.scss
```diff
@@ -0,0 +1,41 @@
...
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  flex-wrap: wrap;
+}
+
+.optionsSection {
+  text-align: right;
+  // FIXME: Why does text-align do nothing?
+  display: ruby;
```
magicznyleszek, line 39: Same as above: either not use `display: ruby` (because possibly not proper usage) or use `ruby-align`

jsapp/scss/stylesheets/partials/_registration.scss
```diff
@@ -1,5 +1,8 @@
 @use "scss/_variables";
 
+// A rare light mid-blue
+$kobo-light-mid-blue: #c0e3ff;
```
magicznyleszek, line 4: Is this Ignacio approved? :D Because maybe it should go to `colors` (in `kobo-common` too)?

jsapp/scss/stylesheets/partials/_registration.scss
```diff
@@ -11,7 +14,7 @@
   width: 100%;
   min-height: 100%;
   position: relative;
-  color: $kobo-white;
+  color: $kobo-gray-96;
   padding: 20px;
```
magicznyleszek, line 18: With so much changes here, I would try using `sizes.scss` and fix whole file :)

magicznyleszek (changes requested):

jsapp/js/account/security/email/emailSection.component.tsx
bufke: We should standardize the language. I take "isLoading" to mean it's waiting for some async fetch of data. isInitialised I would interpret as the first load has completed.  Or perhaps more generically it just means the initial value was set. isLoading may flip back and forth, while isInitialised would load flip from false to true exactly once. isInitialised is helpful for things like a "app shell" UI that shouldn't show again later.

But this effect's design needs improved. It's being used as if it was what I describe as isInitialised. Omitting the second param of useEffect causes it to run every time (unnecessary). It works because isLoading starts true. This is untrue. The API data is not loading before the component even mounted. I think the intention here was to run once on mount. Therefore it should have a second param of `[]` and no additional if statements.

Let's keep the language conversation going. I would personally submit two booleans:

- isLoading
- isFirstLoadComplete

To me this is unambiguous.

jsapp/js/account/security/apiToken/apiTokenSection.module.scss
p2edwards: Fixed this along with some other alignment issues in [`a39eeae46`](https://github.com/kobotoolbox/kpi/pull/4163/commits/a39eeae465a468ce1dfdb59b9e0b5919bc53745c)

After:
After

<details><summary>Before:</summary>
Before

</details>

jsapp/js/account/security/email/emailSection.module.scss
p2edwards: display: flex on the parent does all we need to do.

jsapp/js/account/security/email/emailSection.component.tsx
p2edwards: In [`4fe338a05`](https://github.com/kobotoolbox/kpi/pull/4163/commits/4fe338a050f0fb173c46952a55c02889d0000c8f), I removed isLoading from the state and passed `[]` to useEffect.

jsapp/scss/stylesheets/partials/_registration.scss
magicznyleszek: Actually I have another request. I see looking at `kobo-common` project that we've decided to update the `kobo-mid-blue` color value to `$kobo-mid-blue: hsl(206deg 100% 88%); // #c0e3ff`. So could you please:
1. remove `$kobo-light-mid-blue` that you've defined here
2. update `$kobo-mid-blue` in `/jsapp/scss/_colors.scss` to have 88%
3. use `$kobo-mid-blue` here instead :heart:

magicznyleszek (changes requested):

magicznyleszek (approved):

Issue #4080, opened 2022-10-14
Projects metadata download button


magicznyleszek:
A simple button atop table that downloads projects metadata for current view.

Requires #4079

Issue #4058, opened 2022-10-10
Place ProjectsActionButtons on top of table, add checkboxes, update functionality


magicznyleszek:
- ~~`AssetRoute` is using that - and it's ok, pass "single row selected" and just check if everything works correctly~~
- add checkbox for each row in `ProjectsTable`
- no bulk operations for MVP
- all available buttons should reflect the number and type of selected rows plus the permissions
- also move the subscribe/unsubscribe button for collections
- move buttons around, the ones ouside dropdown are:
  - edit
  - share
  - download metadata
  - delete
  - archive
  - duplicate

Requires #4053

Issue #4056, opened 2022-10-10
Update visual side (CSS styles) of ProjectsTable


magicznyleszek:
- consider updating usage of colors to `kobo-common` in place of local `_colors.scss`

Requires #4053

Issue #4055, opened 2022-10-10
Add infinitescroll to ProjectsTable and remove pagination


magicznyleszek:
See if this is doable before all the stores are done. Should be fine - just provide proper properties and callbacks, we can hook it up later.

Requires #4053

PR #4167, opened 2022-11-30
Cleanup ProjectsTable component

### LINKED ISSUE #4054: Remove filtering code from ProjectsTable
- there are two dropdowns for language and sector that needs to be removed
- this task simply removes the functionality

Requires #4053

magicznyleszek:
## Description

Preparing code for further work. Trying to make PRs manageable in size :heart:

## Notes for reviewer

- Modifies `ProjectsTable` in significant way, but mostly rearranging the code, removing and simplifying parts. It is mainly a preparation for final styles and using back-end.
- Splits out `ProjectsTableHeader` into separate file
- Changing `ProjectsTableRow` into functional component
- Hiding non-selected columns is done purely on front-end
- Adds infinite scroll, but it's not testable yet, as there is no endpoint hooked up
- Removes filtering code from the `ProjectsTable` as it will be handled by `ProjectsFilter` component
- There is "Export all data" button that is not hooked up yet
- The custom view is now at `#/projects/1`, not `#/projects/custom/1`

Built atop `duplicate-projects-table` branch (PR #4151). Please change base branch to `feature/regional_projects` after that PR is merged.

## Related issues

Fixes #4054
Part of #4055 
Part of #4056 
Part of #4058 
Part of #4080


jsapp/js/projects/projectsTable/projectsTableRow.module.scss
```diff
@@ -0,0 +1,143 @@
...
+      min-width: sizes.$x14;
+      width: sizes.$x14;
+      margin-left: sizes.$x3;
+      display: inline-block;
+
+      &.k-icon-filter-arrows {
+        font-size: sizes.$x10;
+        margin-left: 0 !important;
+      }
+    }
```
p2edwards, line 119: Not sure that these get applied yet - I didn't see them in the browser console.
Do they need to invoked in projectTableHeader.tsx?

p2edwards (approved):

jsapp/js/projects/projectsTable/projectsTableRow.module.scss
magicznyleszek: Right, so `.k-icon` was being used in the header row (header is also using `ProjectsTableRow`). With CSS Modules it is no longer possible to reference some other selector that is not defined in the same file. But nontheless, these styles are not needed, and most of this file will get rewritten, so I removed them :)

p2edwards (approved):

PR #4201, opened 2023-01-05
Add Storybook

### LINKED ISSUE #3769: Use designbook instead of manual design-system we have

bufke:
Resolves #3769
Depends on #4163

http://storybook.kbtdev.org

image

# Benefits

- Reduce bundle size (as it's a separate bundle)
- Reduce amount of code for example components
- Community tooling such as a11y


jsapp/js/components/common/icon.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+  component: Icon,
+  argTypes: {},
+} as ComponentMeta<typeof Icon>;
+
+const Template: ComponentStory<typeof Icon> = (args) => <Icon {...args} />;
+
+export const Primary = Template.bind({});
+Primary.args = {};
+
+export const AllIcons = () => (
```
magicznyleszek, line 17: I like this list. I think we can delete `/jsapp/k-icons-html-template.hbs` (and the mention in `/scripts/generate_icons.js`)

magicznyleszek (changes requested):

.eslintrc.js
```diff
@@ -1,12 +1,13 @@
...
+      cy: 'readonly',
+      Cypress: 'readonly'
     },
-  ],
-};
+    rules: {
+      semi: ['warn', 'never'] // Cypress style
+    }
+  }]
+};
```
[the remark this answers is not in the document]
bufke: This was edited by a storybook migration script.

jsapp/js/components/common/icon.stories.tsx
bufke: Ok, removed it and the generate icons script still seems to run.

jsapp/js/components/common/button.tsx
```diff
@@ -63,7 +62,7 @@ interface ButtonProps {
   isFullWidth?: boolean;
   /** Additional class names. */
   classNames?: string[];
-  onClick: Function;
+  onClick: (event: any) => void;
```
bufke, line 65: Any because making it more precise breaks usage. Note that (e: any): void is more precise than Function, which is (any) => any

magicznyleszek:
@bufke can we delete the branch?

PR #4207, opened 2023-01-10
Show blue focus outline for keyboard navigation

bufke:
## Description

Allow keyboard navigation via enabling focus state

## Related issues

Fixes or mitigates https://github.com/kobotoolbox/kpi/issues/2809

# Commentary

This greatly improves the situation for keyboard usage and thus accessibility. Recommend we take much better over perfect. We can add other improvements later.


jsapp/scss/stylesheets/partials/_base.scss
```diff
@@ -1,3 +1,4 @@
+@use 'scss/_colors';
```
magicznyleszek, line 1: Please use `@use '~kobo-common/src/styles/colors';` instead

magicznyleszek (changes requested):

jsapp/scss/stylesheets/partials/_base.scss
duvld: Merging beta into the branch fixed it :)

PR #4225, opened 2023-01-20
Fix assets submission count missing with 'view_submissions' of project views

### LINKED ISSUE #4194: Revisit and refactor permission checks for `ProjectView` django app
The permission checks in #4091 are ok but it could certainly simplified and refactored. 
Let's keep an issue open to revisit when the time permits.

noliveleger:
## Checklist

3. [x] Ensure the tests pass

## Description

Returns the correct number instead of 0

## Related issue

Close #4194

## Additional info

`asset.has_perm(user, permission)` returns `True` if `user` belongs to a project view that has `permission` selected for `asset`. Previous behaviour was `False`.
 
`change_metadata` permission has been rename `change_metadata_asset` to be consistent with other permissions.


kpi/permissions.py
```diff
@@ -121,6 +120,38 @@ def has_object_permission(self, request, view, obj):
...
+        if (
+            method == 'PATCH'
+            and user_has_project_view_asset_perm(
+                obj, user, PERM_CHANGE_METADATA_ASSET
+            )
+        ) or (
+            method == 'GET'
+            and user_has_project_view_asset_perm(obj, user, PERM_VIEW_ASSET)
+        ):
+            return True
```
jnm, line 150: We discussed how this gets [properly limited](https://github.com/kobotoolbox/kpi/blob/dc615657b24f10287ea7ab186a1c3710d07185d2/kpi/serializers/v2/asset.py#L180-L192) in the serializer, but we should add a test for that

jnm (changes requested):
@noliveleger good to go after adding the test we discussed!

kpi/permissions.py
noliveleger: Done in https://github.com/kobotoolbox/kpi/pull/4225/commits/ff12597fa445e07d2acf0415c48292c927d96802

kpi/tests/api/v2/test_api_assets.py
```diff
@@ -529,12 +529,58 @@ def test_project_views_for_anotheruser_can_change_metadata(self):
...
+        assert asset_detail_response.data['settings'] == settings
+        # `summary` and `content` should have not
+        assert asset_data['summary'] != summary
+        assert asset_data['content'] != content
+        assert self._sorted_dict(asset_data['summary']) == self._sorted_dict(
+            asset_data['summary']
+        )
+        assert self._sorted_dict(asset_data['content']) == self._sorted_dict(
+            asset_data['content']
+        )
```
jnm, line 582: @noliveleger these are like `assert x == x` are they not? :monocle_face: Reformatted:
```python
assert self._sorted_dict(asset_data['summary'])
    == self._sorted_dict(asset_data['summary'])

assert self._sorted_dict(asset_data['content'])
    == self._sorted_dict(asset_data['content'])
```

noliveleger: Oops :-P  Indeed. Not what I wanted to do. 
That's how you make tests pass no matter what :-P

jnm (approved):

PR #4227, opened 2023-01-23
Allow 2FA with Single Sign-On

noliveleger:
## Checklist

3. [x] Ensure the tests pass

## Description

Display 2FA form after logging with SSO

## Additional info

Restore the redirection to `/admin/` when admin users try to login to `/admin/` directly.


kobo/apps/accounts/adapter.py
```diff
@@ -1,10 +1,52 @@
...
+                'form': mfa_token_form,
+            }
+
+            return TemplateResponse(
+                request=request,
+                template='mfa_token.html',
+                context=context,
+            )
+
+        return super().pre_login(request, user, **kwargs)
```
jnm, line 48: In any situation where you have a pattern like:
```python
if condition:
    # many lines of code
    return a

return b
```
Please, please rewrite this as:
```python
if not condition:
    return b  # i'll call this "bailing out early" for discussion's sake

# many lines of code
return a
```

This particular example, where `# many lines of code` is only 21 lines, is not too hard to read. Imagine, though, that there's enough code that it doesn't fit on a single screen, or that there are multiple nested conditionals or loops: it starts getting hard to keep track of the context while reading.

I hate to say this, because it's crudely reductionist and risks encouraging dogmatic behavior, but try to minimize amount of nesting/indentation, and where it's necessary (as of course it often is), try to minimize the number of lines inside the indented block. This applies even to logic that doesn't match the "bail out early" kind of pattern described above; an example  from earlier today of deeply-nested and large conditionals that are challenging is [this part of kobo-install](https://github.com/kobotoolbox/kobo-install/blob/e6274462282733cf92d588b3f8366259707a5758/helpers/config.py#L1578-L1731). In particular, the `            else:` (indented 12 spaces) on line 1709 makes me scroll up 124 lines (multiple screens in my editor and on GitHub) to find its matching `if`. There may very well _not_ be a way to reduce indentation and make the code easier to read simply by reordering the logic, but in that case, the logic should be separated into different functions.

kobo/apps/accounts/adapter.py
```diff
@@ -1,10 +1,52 @@
...
+                initial={'ephemeral_token': ephemeral_token_cache}
+            )
+            next_url = kwargs.get(
+                'redirect_url', resolve_url(settings.LOGIN_REDIRECT_URL)
+            )
+            current_site = get_current_site(request)
+            context = {
+                REDIRECT_FIELD_NAME: next_url,
+                'site': current_site,
+                'site_name': current_site.name,
```
jnm, line 37: Can we avoid introducing the Django sites framework? It's in kobocat and it's annoying; it seems like we've gotten this far in kpi without using it. I see that `django.contrib.sites` has _not_ been added to `INSTALLED_APPS`, meaning we can't actually use the `Site` model; is this invocation of the sites framework a fancy way to call `request.get_host()` for `site_name`[^1], or does the template do more stuff with the `site` object?

[^1]:
    ```
    In [1]: from django.contrib.sites.shortcuts import get_current_site

    In [2]: class C: pass

    In [3]: req = C()

    In [4]: get_current_site(req).name
    ---------------------------------------------------------------------------
    <snip>
    AttributeError: 'C' object has no attribute 'get_host'

    In [5]: req.get_host = lambda: 'foo.com'

    In [6]: get_current_site(req).name
    Out[6]: 'foo.com'
    ```

jnm (changes requested):

kobo/apps/accounts/adapter.py
```diff
@@ -1,10 +1,52 @@
...
 
 class AccountAdapter(DefaultAccountAdapter):
+
+    def pre_login(self, request, user, **kwargs):
+
+        auth_method = (
+            get_mfa_model()
+            .objects.filter(is_active=True, user=user)
+            .first()
+        )
```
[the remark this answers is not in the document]
noliveleger: Totally!

kobo/apps/accounts/adapter.py
noliveleger: Because it's inheriting for `LoginView`, I thought it was required because of the [context passed to the template ](https://github.com/django/django/blob/main/django/contrib/auth/views.py#L110-L121).

But it seems to work without it.

kobo/apps/accounts/adapter.py
```diff
@@ -17,35 +16,29 @@ class AccountAdapter(DefaultAccountAdapter):
...
+            REDIRECT_FIELD_NAME: next_url,
+            'view': MfaTokenView,
+            'form': mfa_token_form,
+        }
+
+        return TemplateResponse(
+            request=request,
+            template='mfa_token.html',
+            context=context,
         )
```
jnm, line 41: :smiling_face_with_three_hearts:

kobo/apps/accounts/adapter.py
jnm: OK. I guess I don't totally understand the relationship with `LoginView`, but since neither `mfa_token.html` nor `account/base.html` use `site` or `site_name`, we should be fine

jnm (approved):
approved based on code review, but @bufke is going test the various workflows

bufke (changes requested):
We discussed some security considerations in internal chat.

PR #4242, opened 2023-02-01
make activation email configurable

JacquelineMorrissette:
## Checklist

3. [x] Ensure the tests pass

## Description

Make the activation email customize-able from the admin interface

## Notes for reviewer:
* Created a model to store custom template options
* Configured the templates to be customized
* Created a Template Tag to import custom content to the templates
* Created a Template Tag to convert placeholders to designated content (As discussed with Olivier, this is best way to do it cleanly)


kobo/apps/accounts/models.py
```diff
@@ -1,8 +1,15 @@
 from allauth.account.signals import email_confirmed
+from django.contrib import admin
 from django.db import models
 from django.dispatch import receiver
 
 
+class EmailContent(models.Model):
+    email_name = models.CharField(max_length=120)
```
bufke, line 8: email_name + section_name must be unique_together, right? There can't be two versions of the same content block on the same email. email_name is the template name, yes? It should have choices that allow for only valid choices. Perhaps section should be the same, but I think that's too tedious and would be fine leaving it free.

kobo/apps/accounts/templatetags/customize_email_content.py
```diff
@@ -0,0 +1,25 @@
...
+        'section_name', 'content'
+    ).filter(email_name=email_name)
+    email_content_dict = {}
+    for email_content in email_contents:
+        email_content_dict[email_content['section_name']] = email_content['content']
+    return email_content_dict
+
+
+@register.simple_tag
+def get_variables(section_content, activate_url, user):
```
bufke, line 21: The name sounds generic, but is really activation link specific. Let's rename it unless there is a strong feeling it should be made more general purpose.

bufke (changes requested):
I like the general plan here.

Can we some docstrings to explain what it all does? At least on the model class. Perhaps with help_text as well.

Needs a unit test.

kobo/apps/accounts/templatetags/customize_email_content.py
JacquelineMorrissette: It's intended to be generic because it covers the activation link, the username, and anything else we may be asked to create placeholders for. If we do one tag for each one, these templates will get extremely messing quick. I can make it slightly less generic as `replace_placeholders` or something

kobo/apps/accounts/models.py
JacquelineMorrissette: They will be added in the next commits, I forewent them for now in favour of getting the prototype done as quickly as possible as asked

kobo/apps/accounts/tests/test_email_content.py
```diff
@@ -0,0 +1,118 @@
...
+            email_name='email_confirmation_signup_message',
+            section_name='section_one',
+            content=''
+        )
+        email_subject = EmailContent.objects.create(
+            email_name='email_confirmation_signup_message',
+            section_name='subject',
+            content=''
+        )
+        # Using the randomint to make sure that usernames and emails
```
bufke, line 65: I would fix the issue directly instead of working around it. 

- If you'd like it to not really cache then use `@override_settings(CACHES={'default': {'BACKEND': 'django.core.cache.backends.dummy.DummyCache'}})` 
- If you'd like isolated but functional cache - use LocMemCache. I'm not sure if this will clear between test cases but you could either clear it manually in setUp or set a unique cache location value per test.

That avoids the need for randomness. If you did want to use randomness regardless, that's fine but you should seed it. Importing and using kpi.tests.kpi_test_case.BaseTestCase will automatically set a seed. This creates pseudo random data. It's "random" but will process the same "random" results every time.

bufke (changes requested):
While I don't think every test needs assertNumQueries, we designed this system explicitly to fetch all content blocks in one query per email. Therefore, I would test that idea with a single assertNumQueries in a case where we expect 2+ blocks to be used. We should find it only adds 1 additional query.

kobo/apps/accounts/tests/test_email_content.py
```diff
@@ -0,0 +1,131 @@
...
+        }
+        default = "Thanks for signing up with KoboToolbox!"
+        request = self.client.post(self.signup_url, data)
+        user = get_user_model().objects.get(email=email)
+        assert request.status_code == status.HTTP_302_FOUND
+        self.client.login(username=user.username, password=user.password)
+        self.client.get(reverse('account_email_verification_sent'))
+        assert len(mail.outbox) == 1
+        assert mail.outbox[0].subject == 'Activate your KoboToolbox Account'
+        self.assertNumQueries(1)
```
bufke, line 96: This doesn't do anything. It needs to be used as a context manager, see django docs. Generally when testing, please ensure the test works as intended by making it fail. In this case, no number will ever fail.

bufke (changes requested):

bufke (changes requested):
Let's discuss

PR #4246, opened 2023-02-02
Allow manual column resizing in Regional Projects Table

### LINKED ISSUE #4175: Columns should be resizable
- Each column should have mininum width
- A small area between columns should be draggable and would cause a resize
- We should store the widths in `/me` endpoint

Code example: https://stackoverflow.com/a/62437093/2311247

p2edwards:
## Description

Columns can be resized in the Regional Projects Table by click-and-drag in the column headers.

## Details

- [x] Add "title" attribute to column headers, in case column name is cut off.
- [x] Add resize handles that appear when hovering the title area
- [x] Drag resize handles to resize columns
resize-handles

## Related issues

Fixes #4175


jsapp/js/projects/projectsTable/colResize.tsx
```diff
@@ -0,0 +1,271 @@
...
+  const draggingColumnRef: React.MutableRefObject<string | null> = useRef(null);
+  /** The pageX when we started interacting with the resize handler */
+  const dragStartXRef: React.MutableRefObject<number | null> = useRef(null);
+  /** The width of the header cell when we initially touched its resize handler */
+  const dragStartWidthRef: React.MutableRefObject<number | null> = useRef(null);
+  /** The pageX at the previous mousedown/mousemove.
+      Used for de-duping mousemove */
+  const dragPrevXRef = useRef(0);
+  /** The column width we used in the previous mousedown/mousemove.
+      Used for de-duping mouse renders. */
```
p2edwards, line 139: typo: I meant to say mousemove renders.

jsapp/js/projects/projectsTable/colResize.tsx
```diff
@@ -0,0 +1,271 @@
...
+      //  - Only relevant if drag is already happening
+      //  - Update column width state for current column
+      if (
+        e.type === 'mousemove' &&
+        // Event logic
+        isDraggingRef.current &&
+        // Type safety
+        draggingColumnRef.current !== null &&
+        dragStartXRef.current !== null &&
+        dragStartWidthRef.current !== null &&
```
p2edwards, line 160: These null checks / nullability could be removed and the code shortened, if we trust the correctness of the mouseup and mousedown handlers. 

When I was writing these initially I wanted to make sure the mousemove would never accidentally use a symbolic 0 or -1 due to a missed initialization. I thought I would clear these back to 'null' on mouseup, but actually letting the old values linger doesn't cause any practical problems, so I didn't bother.

By the time I added dragPrevXRef and dragPrevWidthRef, I fully trusted the isDraggingRef to guard mousemove and I trusted the mousedown handler to initialize the rest of the variables properly, so I simplified their types by initializing to 0. 

So I'll probably follow suit and remove the nullability and these null checks.

jsapp/js/projects/projectsTable/projectsTableRow.module.scss
```diff
@@ -42,7 +42,7 @@ $projects-table-hover-bg: colors.$kobo-gray-98;
   &[data-field='ownerOrganization'],
   &[data-field='status'],
   &[data-field='submissions'] {
-    width: 12.5%;
+    width: 13%;
```
p2edwards, line 45: This removed a spurious ellipsis for default screen width - which was really being caused I think, by invisible space being truncated. The real solve for this would probably involve being careful about whitespace at the component level, and may involve cooperation with [our formatter](https://prettier.io/docs/en/options.html#html-whitespace-sensitivity).

jsapp/js/projects/projectsTable/projectsTableHeader.module.scss
```diff
@@ -55,3 +61,92 @@
...
+$cell-line-width: sizes.$x1;
+$cell-line-vertical-offset: sizes.$x6;
+// Theme colors
+$cell-line-color: colors.$kobo-gray-85;
+$resizer-line-color: colors.$kobo-blue;
+$resizer-area-color: colors.$kobo-light-blue;
+
+// Resizer "pad" - interactive area
+.resizer {
+  cursor: ew-resize; // col-resize <-|->   ew-resize <->
```
p2edwards, line 82: I asked Ignacio which cursor he preferred, and he liked the `ew-resize` version for hover.

When I added the full-body cursor during drag, I switched it to `col-resize`, and I'm pleased with the way the two work in concert.

jsapp/js/projects/projectsTable/colResize.tsx
```diff
@@ -0,0 +1,271 @@
...
+function columnWidthReducer(
+  columnWidths: ColumnWidths,
+  action: ColumnWidthReducerAction
+) {
+  switch (action.type) {
+    case 'resize':
+      return {
+        ...columnWidths,
+        [action.column]: action.width,
+      };
```
p2edwards, line 67: This reducer shrank while I was writing it. It now has only 1 action, dispatched in 1 place. I kept it around, but its sole invocation in the event handler:

```tsx
          dispatch({
            type: 'resize',
            column: draggingColumnRef.current,
            width: newWidth,
          });
```

 could be replaced with something like this:

 ```tsx
          setColumnWidths((columnWidths) => ({
            ...columnWidths,
            [draggingColumnRef.current]: newWidth,
          }))
```

I'm somewhat inclined to keep it around for reference purposes, but I could go either way.

File: jsapp/js/projects/projectsTable/colResize.tsx [hunk 6]
```diff
@@ -0,0 +1,271 @@
...
+  const dragPrevWidthRef = useRef(0);
+
+  /**
+   * The singular DOM event handler for column resizing.
+   * One consolidated function. It switches on event.type, delegates when
+   * necessary, and exits early.
+   */
+  const handlerRef = useRef((e: Event) => {
+    // TypeScript guard
+    if (e instanceof MouseEvent) {
```
magicznyleszek, line 149: Could we do this instead of wrapping whole function in `if` statement?

```ts
if (e instanceof MouseEvent === false) {
  return;
}
```

This way the guard is here on top, and the rest of the code doesn't need to be indentend so much :)

jsapp/js/projects/projectsTable/colResize.tsx
```diff
@@ -0,0 +1,271 @@
...
+      }
+    }
+  });
+  const eventTypes = ['mousedown', 'mouseup', 'mousemove', 'contextmenu'];
+
+  // Mount/Unmount effect: Attach/remove handlers on the body
+  useEffect(() => {
+    // Mount! Add handlers on body
+    eventTypes.forEach((eventType) => {
+      document.body.addEventListener(eventType, handlerRef.current);
```
magicznyleszek, line 235: When `ColumnResizer` is being rendered (i.e. user is displaying `#/projects/abc123` route) then every mouse move will trigger the callback. The callback will not do anything unless mouse is pressed of course, but perhaps we could change this. What if we only add handlers after `mousedown` (or even better: a `mousedown` on the resizer element), and remove them after `mouseup`/`contextmenu`? This way the callback funcion wouldn't be called at all unless user is actually resizing.

magicznyleszek: I might be too cautious with this thoug. Perhaps todays browsers handle these do-nothing functions with no performance loss. But still, this feels untidy :)

jsapp/js/projects/projectsTable/colResize.tsx
```diff
@@ -0,0 +1,271 @@
...
+    // Unmount! Remove handlers on body
+    return () => {
+      eventTypes.forEach((eventType) => {
+        document.body.removeEventListener(eventType, handlerRef.current);
+      });
+    };
+  }, []);
+
+  // Log confirms minimal re-rendering.
+  // console.log('Rendered resizer!');
```
magicznyleszek, line 247: Let's remove this comment

jsapp/js/projects/projectsTable/colResize.tsx
```diff
@@ -0,0 +1,271 @@
...
+
+  // Render a dynamic style tag to update all custom widths in a single step,
+  // without modifying any attributes managed elsewhere by React.
+  return (
+    <>
+      <style>
```
magicznyleszek, line 253: I like this, it's a clever solution to render a style tag :)

jsapp/js/projects/projectsTable/colResize.tsx
```diff
@@ -0,0 +1,271 @@
...
+          (column) =>
+            `
+            [data-field="${column}"] {
+              width: ${columnWidths[column]}px !important;
+              max-width: ${columnWidths[column]}px !important;
+            }
+            `
+        )}
+      </style>
+      <style>
```
magicznyleszek, line 264: Do we need separate style tags?

jsapp/js/projects/projectsTable/projectsTableHeader.module.scss
```diff
@@ -55,3 +61,92 @@
...
+}
+.resizer:hover::after {
+  opacity: 1;
+  transition: opacity 0.2s;
+}
+
+// Cell width indicators on column hover
+.columnRoot,
+.columnRoot:hover {
+  &::before,
```
magicznyleszek, line 124: Let's hide the `::before` line on the first visible column. Something like `.columnRoot:first-child::before {display: none !important;}` should be fine.

magicznyleszek (changes requested):

jsapp/js/projects/projectsTable/colResize.tsx
p2edwards: Oh, yep, if it works!

jsapp/js/projects/projectsTable/colResize.tsx
p2edwards: It does seem like the right thing to do :) I combined them all to make it easier to read/write/test but I think it's turned out simple enough.

jsapp/js/projects/projectsTable/colResize.tsx
p2edwards: Probably not. IIRC I went with it on random speculation that React might have more fun diffing if the one thing is in its own style tag, but I have no reason to believe that. XD

jsapp/js/projects/projectsTable/projectsTableHeader.module.scss
p2edwards: Oh interesting! I included it for consistency

p2edwards: ✅

jsapp/js/projects/projectsTable/colResize.tsx
```diff
@@ -0,0 +1,271 @@
...
+ * Currently a minimum of 108px.
+ * Can be hardcoded per-field.
+ */
+function minColWidth(fieldname: string) {
+  const defaultMinimumWidth = 108;
+  const minimums: ColumnWidths = {
+    countries: 116,
+    dateModified: 144,
+    dateDeployed: 144,
+  };
```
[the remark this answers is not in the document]
p2edwards: Decided:

- Keep the min-widths in the ColumnResizing file for now (instead of mixing them in with other field constants)
   - Easier to see, e.g., the only 3 field exceptions that are defined so far
   - Make it easier to define more following the pattern without jumping/scrolling in multiple files
- [ ] Use type ProjectFieldName instead of generic 'string'

(via internal discussion)

jsapp/js/projects/projectsTable/colResize.tsx
```diff
@@ -0,0 +1,271 @@
...
+ * handles.
+ */
+export default function ColumnResizer() {
+  // State used in rendering new styles
+  const [columnWidths, dispatch] = useReducer(
+    columnWidthReducer,
+    {} as ColumnWidths
+  );
+  // Polish. This could perhaps be merged into a single state with isDragging
+  const [shouldRenderBodyCursor, setShouldRenderBodyCursor] = useState(false);
```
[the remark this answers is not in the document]
p2edwards: Actually, we can't combine them, both are needed for reasons.

- **isDraggingRef**: Event handlers can't read up-to-date values of React state. So functionally, the event handlers need isDraggingRef for their logic, and it can't be state.
- **shouldRenderBodyCursor**, aka **isDragging** (state): We do want to trigger some re-rendering when starting/stopping a drag event, so we need state variable here.

jsapp/js/projects/projectsTable/colResize.tsx
p2edwards: I moved just the move handler out of this function and toggle it separately. I'm keeping the other handlers combined to simplify the code.

PR #4251, opened 2023-02-07
Allow filtering by country in projects list

### LINKED ISSUE #4176: Improve ProjectsFilter to use countries list
- Use countries list from `/environment` for the `countries` field when creating a filter
- Should be either a dropdown or an autocomplete field (try `<datalist>` tag https://stackoverflow.com/a/33985301/2311247?)
- it is needed, because the filter requires a code (e.g. `DEU` for "Germany") to work, and there is no place we display codes

duvld:
Also changed `KoboSelect` to use `value` instead of `id` for consistency (with older code and backend). 

## Description
Users can now select country by full name in projects filter list.

## Related issues

Fixes #4176

### DIFF
--- a/jsapp/js/projects/projectViews/projectsFilterEditor.tsx
+++ b/jsapp/js/projects/projectViews/projectsFilterEditor.tsx
@@ -154,6 +154,7 @@ export default function ProjectsFilterEditor(props: ProjectsFilterEditorProps) {
               type='outline'
               size='m'
               isClearable
+              isSearchable
               placeholder={t('Country')}
               selectedOption={props.filter.value || ''}
               options={COUNTRIES}
--- a/jsapp/js/projects/projectViews/projectsFilterEditor.tsx
+++ b/jsapp/js/projects/projectViews/projectsFilterEditor.tsx
@@ -68,7 +68,7 @@ export default function ProjectsFilterEditor(props: ProjectsFilterEditorProps) {
         (filterDefinition) => filterDefinition.availableConditions.length >= 1
       )
       .map((filterDefinition) => {
-        return {label: filterDefinition.label, id: filterDefinition.name};
+        return {label: filterDefinition.label, value: filterDefinition.name};
       });
 
   const getConditionSelectorOptions = () => {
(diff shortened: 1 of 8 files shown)


jsapp/js/projects/projectViews/projectsFilterEditor.tsx
```diff
@@ -75,11 +79,14 @@ export default function ProjectsFilterEditor(props: ProjectsFilterEditorProps) {
     return fieldDefinition.availableConditions.map(
       (condition: FilterConditionName) => {
         const conditionDefinition = FILTER_CONDITIONS[condition];
-        return {label: conditionDefinition.label, id: conditionDefinition.name};
+        return {label: conditionDefinition.label, value: conditionDefinition.name};
       }
     );
   };
 
+  const isCountryFilterSelected = () =>
```
magicznyleszek, line 87: Turn this into simple const value instead of function. This is a functional component - each render runs whole function, so there is no need for a helper funcion so simple as this :)

jsapp/js/projects/projectViews/projectsFilterEditor.tsx
```diff
@@ -123,19 +130,40 @@ export default function ProjectsFilterEditor(props: ProjectsFilterEditorProps) {
       {/* Filter value */}
       <div className={styles.column}>
         {!props.hideLabels && (
-          <span className={styles.label}>{t('Value')}</span>
+          <span className={styles.label}>
+            {isCountryFilterSelected() ? t('Select country') : t('Value')}
+          </span>
```
magicznyleszek, line 135: Let's leave it as it was before. If you add multiple filters you will see what I mean :) These labels are working as a "column headers" kinda. I decided to do it this way, so it's possible to use a filter editor with labels as singular instance.

magicznyleszek (changes requested):

jnm:
The PR title and description need to be updated before merging; see https://chat.kobotoolbox.org/#narrow/stream/4-Kobo-Dev/topic/PR.20description.20gripe.20thread/near/201360

duvld:
Github doesn't seem to like me editing this PR. I edited it twice already but it didn't save. I'll try it again after making these changes

jnm:
Ah, that's annoying! :flushed: Let me know if it continues to fail

magicznyleszek (approved):

PR #4258, opened 2023-02-09
Feature/sso signup set default email

david-code:
## Checklist

    - no change to APIs
3. [x] Ensure the tests pass

## Description

Override django-allauth library defaults to correctly populate user email on OpenID Connect signup

## Explanation

I ran into a problem setting up Kobo SSO signin with an OpenID Connect provider. Although the [OpenID userinfo endpoint](https://openid.net/specs/openid-connect-core-1_0.html#UserInfo) returns the user's email, it is not pre-populated in the user signup form. We would like to have it so the user does not need to re-enter their email at this step, so this PR adds the functionality to populate the email when provided.

The problem is actually not with the code in `kpi` itself, but with the OpenIDProvider in the `django-allauth` library. I fixed it by customizing the Adapter class. In the long term I'd like to submit a patch to the library to fix it this problem. However I'm not sure when this will be possible, and as I already found a pretty easy fix, I though I'd submit it here.

Please don't hesitate to point out any problems you see or if you would like to tackle this another way. I'm not super familiar with the kobo code base yet and I'm very open to suggestions!

## Related issues

(Didn't create an issue for this, but I can if it's desired.)

### DIFF
--- a/kobo/apps/accounts/adapter.py
+++ b/kobo/apps/accounts/adapter.py
@@ -22,9 +22,12 @@ class AccountAdapter(DefaultAccountAdapter):
         return user
 
 
+# The following class is a work around for https://github.com/pennersr/django-allauth/issues/3257
+# Also see https://github.com/pennersr/django-allauth/issues/3241
+# PR fix: https://github.com/pennersr/django-allauth/pull/3242
 class SocialAccountAdapter(DefaultSocialAccountAdapter):
     def populate_user(self, request, sociallogin, data):
         user = super().populate_user(request, sociallogin, data)
-        if sociallogin.email_addresses:
+        if not user.email and sociallogin.email_addresses:
             user.email = sociallogin.email_addresses[0].email
         return user
--- a/kobo/apps/accounts/adapter.py
+++ b/kobo/apps/accounts/adapter.py
@@ -1,5 +1,6 @@
 from allauth.account.adapter import DefaultAccountAdapter
 from allauth.account.forms import SignupForm
+from allauth.socialaccount.adapter import DefaultSocialAccountAdapter
 from constance import config
 from django.db import transaction
 
@@ -18,4 +19,12 @@ class AccountAdapter(DefaultAccountAdapter):
             user.extra_details.data.update(extra_data)
             if commit:
                 user.extra_details.save()
+        return user
+
+
+class SocialAccountAdapter(DefaultSocialAccountAdapter):
+    def populate_user(self, request, sociallogin, data):
+        user = super().populate_user(request, sociallogin, data)
+        if sociallogin.email_addresses:
+            user.email = sociallogin.email_addresses[0].email
         return user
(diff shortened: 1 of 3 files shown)


kobo/apps/accounts/adapter.py
```diff
@@ -19,3 +20,11 @@ def save_user(self, request, user, form, commit=True):
             if commit:
                 user.extra_details.save()
         return user
+
+
+class SocialAccountAdapter(DefaultSocialAccountAdapter):
+    def populate_user(self, request, sociallogin, data):
+        user = super().populate_user(request, sociallogin, data)
+        if sociallogin.email_addresses:
```
bufke, line 28: I think this code deserves a comment about how it's an upstream bug. Can you open an upstream bug report and link in the comment. That will make it easier to remove/refactor later.

This only happens with OIDC, correct? Could we check if user.email exists here as well. There is no need to set the email again. It probably doesn't matter unless there is some flow where it would be different or the new populated user has multiple emails (impossible at this time, I think).

kobo/apps/accounts/tests/test_social_account.py
```diff
@@ -29,3 +34,61 @@ def test_delete(self):
         res = self.client.delete(url)
         self.assertEqual(res.status_code, 204)
         self.assertFalse(self.user.socialaccount_set.exists())
+
+
+class ProviderTestCase(APITestCase):
+    def setUp(self):
+        # modify settings to add test social account provider
+        with contextlib.ExitStack() as exit_stack:
```
bufke, line 42: Can override_settings be used instead?

bufke (changes requested):
Looks good otherwise, thanks for contributing. I can re-review once changes are made.

bufke:
One more request, please edit the `## Description` to be a concise explanation suited for our changelog. Examples [here](https://community.kobotoolbox.org/tag/release-notes). It's fine and good to keep more information in the pull request, just put it under another section. Thanks.

kobo/apps/accounts/tests/test_social_account.py
david-code: Thanks, didn't notice this can be used as a class decorator.

kobo/apps/accounts/adapter.py
david-code: @bufke It opened an issue, but it looks like there is another issue that already addresses this: https://github.com/pennersr/django-allauth/issues/3241. It already has an associated PR that will fix this issue. I'll update the comment in the code here to point to that issue

bufke (approved):
It would be preferable to update dango-allauth over this PR. If there is no allauth release by the time other SSO work is finished, then this PR is fine to merge. We should remove it later when allauth is fixed.

david-code:
@bufke Looks like django-allauth made a new [release](https://pypi.org/project/django-allauth/#history). This should render this PR unecessary.

bufke:
https://github.com/kobotoolbox/kpi/issues/4399 Will replace this

PR #4268, opened 2023-02-10
Fix anonymous access regression

### LINKED ISSUE #4266: Anonymous project access not working in 2.023.04
## Description

Limited UI access to public projects was added by #3255 but seems to have stopped working.

## Steps to Reproduce

1. In project sharing settings, check "Anyone can view this form"
2. Copy "Shareable link"
3. Attempt to access link in new browser session
4. Get redirected to login page

## Expected behavior 

Prior to 2.023.04, "Shareable link" redirects to that link plus `/landing`, aka the "FORM" tab of the project. An error notification saying "Authentication credentials were not provided" does appear, which is a bug (now tracked at #4269), but the project is still viewable.

## Actual behavior 

Navigating to the "Shareable link" while not authenticated redirects to the login page.

## Additional details

Unclear if this is related to the switch to django-allauth or some front-end change

jnm:
All routes, instead of only the root route, were being redirected to login whenever a session was not established. Fixes #4266.

### DIFF
--- a/jsapp/js/router/allRoutes.es6
+++ b/jsapp/js/router/allRoutes.es6
@@ -127,7 +127,7 @@ const AllRoutes = class AllRoutes extends React.Component {
         <Tracking />
         <Routes>
           <Route path={ROUTES.ROOT} element={<App />}>
-            <Route path='' element={<Navigate to={ROUTES.FORMS} replace />} />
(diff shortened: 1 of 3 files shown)


jnm:
The spurious pop-up "Authentication credentials were not provided" errors are now tracked by #4269

dorey:
Confirmed this fixes the redirect bug

PR #4278, opened 2023-02-16
Add API endpoint for Stripe Products

bufke:
## Description

Add API endpoint for Stripe Products. Used to increase limits on public servers such as kf.kobotoolbox.org

## Implementation

This is entirely public. It's just a list of known products. It doesn't even require user login. It filters out inactive plans and products.

### DIFF
--- a/kobo/apps/stripe/views.py
+++ b/kobo/apps/stripe/views.py
@@ -35,5 +35,5 @@ class ProductViewSet(viewsets.ReadOnlyModelViewSet):
         plan__active=True,
     ).prefetch_related(
         Prefetch("plan_set", queryset=Plan.objects.filter(active=True))
-    )
+    ).distinct()
     serializer_class = ProductSerializer
--- a/kobo/apps/stripe/views.py
+++ b/kobo/apps/stripe/views.py
@@ -1,12 +1,13 @@
 from django.conf import settings
-from djstripe.models import Product, Subscription
+from django.db.models import Prefetch
+from djstripe.models import Plan, Product, Subscription
 from rest_framework import mixins, renderers, viewsets
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
 
-from kobo.apps.stripe.serializers import (
-    SubscriptionSerializer
-)
+from kobo.apps.stripe.serializers import SubscriptionSerializer
+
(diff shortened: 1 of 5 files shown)


kobo/apps/stripe/views.py
```diff
@@ -25,3 +26,14 @@ def get_queryset(self):
...
+
+class ProductViewSet(viewsets.ReadOnlyModelViewSet):
+    queryset = Product.objects.filter(
+        active=True,
+        livemode=settings.STRIPE_LIVE_MODE,
+        plan__active=True,
+    ).prefetch_related(
+        Prefetch("plan_set", queryset=Plan.objects.filter(active=True))
+    )
+    serializer_class = ProductSerializer
```
JacquelineMorrissette, line 39: As is, the detail view doesn't work. Is a detailed view of this endpoint needed? If not, it might be better to use the generic viewset with a ListModelMixin like the subscription view

bufke: Oh I see it needs a distinct. In the list view as well, there are duplicates when there are multiple plans. I will fix this.

kobo/apps/stripe/views.py
```diff
@@ -1,12 +1,13 @@
...
 from rest_framework import mixins, renderers, viewsets
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
 
-from kobo.apps.stripe.serializers import (
-    SubscriptionSerializer
-)
+from kobo.apps.stripe.serializers import SubscriptionSerializer
+
+from .serializers import ProductSerializer
```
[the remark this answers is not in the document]
bufke: This is what isort is doing. We could adjust the rules though I personally think it's a poor use of our time.

kobo/apps/stripe/views.py
JacquelineMorrissette: I think it needs to have `lookup_field` set to `id` instead of the `pk` too, because if I try to use the `id` listed in the list view, I get `NOT FOUND`

bufke: It works fine for me 
image

Perhaps you're confusing the numeric primary ID with the stripe ID? I could just remove the detail view for now to avoid confusion.

JacquelineMorrissette (approved):

PR #4287, opened 2023-02-21
Add private social apps

### LINKED ISSUE #4263: Allow setting SSO providers to "private"
## Description

The goal of this feature will be to let SSO provider applications be set as private applications. This means the following:

- The application name will not show up as an option in the SSO signup page. If the application is the only SSO application, the option to use SSO will not be displayed. Instead the application itself can directly link to the callback URL to initiate SSO signin.
- The application needs also to be removed from API endpoints returning information on SSO providers. These include `/me` and `/environment`, but I will also check for others.

I'll add a comment below outlining my proposed solution once I've had a chance to examine the code more closely.

david-code:
## Checklist

3. [x] Ensure the tests pass

## Description

Restrict visibility of private social application login.

## Explanation

I implemented the changes discussed in #4263 . From my searches, this should remove all of the appearances of "private" social apps.

I have a test that I haven't commited yet - I need to set up a test Redis instance. Regardless, the rest should be ready for review.

## Related issues

Fixes #4263


david-code:
@bufke The django admin looks like this currently:

image

Any idea how to make it so the two "ACCOUNTS" sections are merged?

bufke:
Yeah that's not ideal. It's because of the allauth.account django app and our internal kobo.apps.accounts django app.

Have any good name suggestions? I don't think they should be combined as they are not the same django app and one is upstream. But essentially kobo.apps.accounts is "Things we extend related to allauth.account". We could choose to set either a verbose name and be done, or consider renaming the python module. I've used "foo_ext" before for extension but I also hate that :laughing: Better than two accounts. "Account Extensions"

If you have no better name, and I hope you do, then maybe set the verbose name of kobo.apps.accounts to Account Extension. The plural version should be Account Extensions I suppose. The python namespace would be a bother to change and I don't mind it as much when seeing the full path kobo.apps.accounts.

david-code:
> If you have no better name, and I hope you do, then maybe set the verbose name of kobo.apps.accounts to Account Extension. The plural version should be Account Extensions I suppose. The python namespace would be a bother to change and I don't mind it as much when seeing the full path kobo.apps.accounts.

I'll think about a bit about the name and get back to you!

Another question @bufke : you'll see that I am basically filtering the same private `SocialApp`s by removing any applications that have linked `SocialApplicationCustomData` objects aka private apps. I'd like to make a function `filter_private_social_apps` that would centralize this knowledge. However I'm not sure were to put this function in the app/module hierarchy, as it will be used between apps. Do you have any suggestions?

david-code:
@bufke I haven't had a great idea for the Django admin name. I was thinking one of the following (instead of accounts)
- Account Management (accurate, but not clear what the difference with Accounts is)
- Logins
- Authentication
I think Logins would be OK, it's at least accurate.

kobo/apps/accounts/views.py
```diff
@@ -52,4 +52,15 @@ class SocialAccountViewSet(
     permission_classes = (IsAuthenticated,)
 
     def get_queryset(self):
-        return super().get_queryset().filter(user=self.request.user)
+        private_apps = [
```
bufke, line 55: I don't think we need to exclude them here.

The environment api should return only public apps. If I'm a regular user, I don't know that Terraso has a private social app. 

But I login with a private social app, it's fine for me to see this. I could imagine a future where it lists all my social app connections and lets me disconnect any of them.

kpi/serializers/current_user.py
```diff
@@ -81,6 +80,19 @@ def get_git_rev(self, obj):
...
             return False
 
+    def get_social_accounts(self, user):
+        private_apps = [
+            app["provider"]
+            for app in SocialApp.objects.filter(
+                custom_data__isnull=False
+            ).values("provider")
+        ]
+        qs = SocialAccount.objects.filter(user=user).exclude(
```
bufke, line 90: I think this would run multiple queries when it could just run one.

bufke (changes requested):

bufke:
Credit to @jnm - he suggests set Django admin to show "Account Extras" for kobo.apps.accounts

and set verbose_name_plural if needed to ensure it doesn't have a double S or something weird.

kobo/apps/accounts/views.py
david-code: @bufke Yeah, that makes sense. I'll remove this.

kpi/serializers/current_user.py
david-code: Yes, you're right. I'll figure out how to simplify this to one query.

david-code:
@bufke Fixed up the comments you left. The tests that were failing now pass on my local machine.

kpi/serializers/current_user.py
bufke: Actually I don't think we need any filter here either. As I mentioned prior, it's not a problem to know that you yourself are a member of a private SSO app. The only place we need to restrict is the Login/Register page, which comes from /environment/

bufke (changes requested):
I tested this out a bit locally and it seems fine aside from my last comment. Once that is changed, we can merge.

kpi/serializers/current_user.py
david-code: Haha right of course! Will fix this up too.

kobo/apps/accounts/views.py
```diff
@@ -1,5 +1,5 @@
 from allauth.account.models import EmailAddress
-from allauth.socialaccount.models import SocialAccount
+from allauth.socialaccount.models import SocialAccount, SocialApp
```
bufke, line 2: I don't think this import is intended.

bufke (approved):
LGTM

david-code:
> We need to rebase/merge. @david-code if you have time could you do this? Or I can at some point.

I will do it, the conflict seems trivial.

bufke:
@david-code it looks like there are now two 0002 migrations. The CI test errors out

```
 E           django.core.management.base.CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0002_add_emailcontent_model, 0002_socialappcustomdata in accounts).
E           To fix them run 'python manage.py makemigrations --merge'
```

david-code:
> @david-code it looks like there are now two 0002 migrations. The CI test errors out
> 
> ```
>  E           django.core.management.base.CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0002_add_emailcontent_model, 0002_socialappcustomdata in accounts).
> E           To fix them run 'python manage.py makemigrations --merge'
> ```

@bufke should be good now!

PR #4289, opened 2023-02-21
Fix error of management command `populate_asset_date_deployed`  not populating all deployed assets.

noliveleger:
## Description

Use the latest **deployed** version of `Asset` objects (even if it is not the latest version) to populate the `date_deployed` field.


kpi/models/asset.py
```diff
@@ -110,16 +110,9 @@ def create(self, *args, children_to_create=None, tag_string=None, **kwargs):
...
-        https://docs.djangoproject.com/en/2.2/ref/models/expressions/#django.db.models.Exists
+        Filter for deployed assets (i.e. assets without a null value for `date_deployed`)
         """
-        deployed_versions = AssetVersion.objects.filter(
-            asset=OuterRef('pk'), deployed=True
-        )
-        return self.annotate(deployed=Exists(deployed_versions)).filter(
-            deployed=True
-        )
+        return self.exclude(date_deployed__isnull=True)
```
jnm, line 115: This is cool but seems like it would break the app if the management command hasn't completed yet. If that's true, we should include this optimization in a subsequent release

noliveleger: kpi#4328 will handle it in the next release ;-)

PR #4290, opened 2023-02-21
Fix move to collection feature

### LINKED ISSUE #4288: Move to collection feature doesn't show all collections
## Description

In Library, all questions, blocks and templates have an option in More Actions drodpown (the "…" one) to move them into a collection you own. It seems there is a bug that causes only some collections to appear as available.

## Steps to Reproduce

1. Have 1 public and 1 private (non-public) collection
2. Have a template
3. Open "…" 
4. 🐛  there is no "Move to" section

## Expected behavior 

"Move to" section in more actions dropdown should be available containing both of your collections.

magicznyleszek:
## Description

All user owned collection now will appear in more actions menu under "Move to" section.

## Code review notes

This was introduced during router6 upgrade. Route changes and session changes turned out to be important for `ownedCollectionsStore` :) I've added few comments to be more clear. I also found a good use for [`when`](https://mobx.js.org/reactions.html#when) function and I've updated few other components that should've been using it instead of wrongly using `observer()`.

## Related issues

Fixes #4288

### DIFF
--- a/jsapp/js/components/library/ownedCollectionsStore.ts
+++ b/jsapp/js/components/library/ownedCollectionsStore.ts
@@ -1,6 +1,6 @@
 import findIndex from 'lodash.findindex';
 import Reflux from 'reflux';
-import { observer } from 'mobx-react';
+import {when} from 'mobx';
 import sessionStore from 'js/stores/session';
 import {actions} from 'js/actions';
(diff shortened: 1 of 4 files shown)


bufke:
Seems to work. The docker build failure is a timeout and seems unrelated.

PR #4296, opened 2023-02-24
Add usage page to track current user stats

duvld:
## Description

Allow user to check the following stats on the new route `/account/usage`:
- Monthly submissions
- Total storage (in GB)
- Monthly transcription minutes
- Monthly translation characters

## Related issues

Part of billing


webpack/dev.server.js
```diff
@@ -6,7 +6,7 @@ const CircularDependencyPlugin = require('circular-dependency-plugin');
 var isPublicDomainDefined = process.env.KOBOFORM_PUBLIC_SUBDOMAIN &&
   process.env.PUBLIC_DOMAIN_NAME;
 var publicDomain = isPublicDomainDefined ? process.env.KOBOFORM_PUBLIC_SUBDOMAIN
-  + '.' + process.env.PUBLIC_DOMAIN_NAME : 'localhost';
+  + '.' + process.env.PUBLIC_DOMAIN_NAME : '192.168.1.43';
```
bufke, line 9: Did you intend to commit this?

jsapp/js/account/usage.api.ts
```diff
@@ -0,0 +1,26 @@
+import {fetchGet} from 'jsapp/js/api';
+
+interface AssetUsage {
+  asset: string;
+  asset__name: string;
+  submission_count_current_month: number;
+  submission_count_all_time: number;
+  nlp_usage_current_month: any,
```
bufke, line 8: Use unknown instead of any. Then if we end up using this later, TS will error and we can substitute in the right value. With any, it won't error and we don't think to go back and update it.

jsapp/js/account/usage.component.tsx
```diff
@@ -0,0 +1,115 @@
...
+    <div className={styles.root}>
+      <h2>{t('Your account total use')}</h2>
+
+      {usage.showBanner && (
+        <div className={styles.header}>
+          <Icon name='alert' size='s' />
+          <div className={styles.article}>
+            <p>
+              {t(
+                'Please note these figures are only updated once per day. Numbers may not reflect immediately recent changes in account usage. For any questions concerning usage, please read the '
```
bufke, line 55: I'm pretty sure that @JacquelineMorrissette told me this isn't true. I think I just made this up in my head one day.

bufke (changes requested):
Looks good otherwise. I'm not qualified to review CSS, maybe see if someone else can. Approved to merge pending the requested changes

jsapp/js/account/usage.component.tsx
JacquelineMorrissette: Can confirm, all our counters are updated in real time :+1:

webpack/dev.server.js
duvld: Oops!

bufke (approved):
JS looks good to me. Once someone reviews the style, let's merge.

srartese (approved):
The styles look great and it resizes well with smaller screen sizes. I think keeping the width there is fine since you don't want the box to get massive with such little text.

PR #4297, opened 2023-02-27
Fix bottom links overlapping each other in login/registration form on mobile screens

magicznyleszek:


p2edwards (approved):

p2edwards:
"Create an account" and "Forgot password?" now render in a stack on narrow screens instead of running into each other.

Screen Shot 2023-02-27 at 10 08 20 AM

PR #4300, opened 2023-02-27
Fix 'Organization Website' not being saved in Account Settings.

p2edwards:


jsapp/js/account/accountSettingsRoute.tsx
```diff
@@ -281,7 +283,7 @@ const AccountSettings = observer(() => {
                 value={form.fields.organizationWebsite}
                 onChange={onAnyFieldChange.bind(
                   onAnyFieldChange,
-                  'organization_website'
+                  'organizationWebsite'
```
p2edwards, line 286: This is one kind of error that TypeScript is not equipped to detect.

PR #4301, opened 2023-02-27
Temporarily disable saving settings on instances with required user metadata fields

magicznyleszek:


jnm:
I think we can close this if #4302 does what is needed :pray:

magicznyleszek:
Superseded by #4302

PR #4309, opened 2023-03-03
Add Organization Owner's email to Organization model

JacquelineMorrissette:
## Checklist

3. [x] Ensure the tests pass

## Description
Add the Organization Owner's email to the Organization model to allow it to meet `dj-stripe`'s requirements

## Other Notes
* All customer's are now syncing to the Customer Stripe Model
* This needs to be tested thoroughly on the billing staging server with the actual webhook, I tested it to the best of my ability locally, but I cannot test with the webhook
* This PR also add's the missing `DJSTRIPE_SUBSCRIBER_MODEL_REQUEST_CALLBACK` and corresponding method to the settings file. Inspiration taken from David's work on GlitchTip

### DIFF
--- a/kobo/apps/organizations/models.py
+++ b/kobo/apps/organizations/models.py
@@ -15,12 +15,10 @@ class Organization(AbstractOrganization):
     @property
     def email(self):
         """
-        This exists to make dj-stripe happy
+        As organization is our customer model for Stripe, Stripe requires that
+        it has an email address attribute
         """
-        return self.owner.organization_user.user.emailaddress_set.get(
-            primary=True,
-            verified=True
-        )
+        return self.owner.organization_user.user.email
 
 
 class OrganizationUser(AbstractOrganizationUser):
--- a/kobo/apps/organizations/models.py
+++ b/kobo/apps/organizations/models.py
@@ -12,6 +12,16 @@ from organizations.abstract import (AbstractOrganization,
 class Organization(AbstractOrganization):
     uid = KpiUidField(uid_prefix='org')
 
+    @property
+    def email(self):
+        """
+        This exists to make dj-stripe happy
+        """
+        return self.owner.organization_user.user.emailaddress_set.get(
+            primary=True,
+            verified=True
+        )
+
 
 class OrganizationUser(AbstractOrganizationUser):
     pass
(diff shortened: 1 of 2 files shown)


kobo/apps/organizations/models.py
```diff
@@ -12,6 +12,16 @@
 class Organization(AbstractOrganization):
     uid = KpiUidField(uid_prefix='org')
 
+    @property
+    def email(self):
+        """
+        This exists to make dj-stripe happy
+        """
+        return self.owner.organization_user.user.emailaddress_set.get(
```
bufke, line 20: User.email is the canonical email address to read. I know it's not normalized and not clear this should be used. Can you think of a good place to document this? I'm not sure because there is no code we wrote that defines this. It's the upstream Django auth.User and allauth.account.EmailAddress. If you can't think of a better place, maybe a code comment in settings/base.py near the other ACCOUNT_ settings.

kobo/apps/organizations/models.py
```diff
@@ -12,6 +12,16 @@
 class Organization(AbstractOrganization):
     uid = KpiUidField(uid_prefix='org')
 
+    @property
+    def email(self):
+        """
+        This exists to make dj-stripe happy
```
bufke, line 18: While I appreciate a snarky comment when reading the code - the PR detail is the better place for this. Let's explain why this is needed. As I understand, djstripe needs this because Stripe requires an email on a Customer. Since our "customer" is an Organization - we need a means to get one and only one email for the Organization customer. You may want to review the original bug's stacktrace to ensure my understanding is correct.

bufke (changes requested):

kobo/apps/organizations/models.py
JacquelineMorrissette: I understand why it's needed, I habitually wrote dj-stripe

kobo/settings/base.py
```diff
@@ -575,7 +575,20 @@ def __init__(self, *args, **kwargs):
...
+
+def organization_request_callback(request):
+    """
+    `DJSTRIPE_SUBSCRIBER_MODEL` needs a custom callback function because we aren't
+    using the default model `auth.user`
+    Gets an organization instance from the user passed through `request`
+    """
+    user = request.user
+    if user:
+        return user.organizations_organization.get()
```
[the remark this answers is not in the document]
JacquelineMorrissette: It's no longer needed and works without it

bufke (approved):

PR #4311, opened 2023-03-06
Harden CSP and remove all inline JavaScript

jnm:
Also apply `content-disposition: attachment` header when serving form media file content; serve `ConfigurationFile` content (i.e. logos and login background images) directly instead of via redirects; modify CSP to allow access to Humanitarian OpenStreetMap layers

## Description
Improve security by explicitly naming locations that are allowed to load JavaScript. Those running their own instances of Kobo will still need to set `ENABLE_CSP=True` to take advantage of this content security policy (CSP).

Additionally, direct browsers to download form media files instead of rendering them when opened directly, as opposed to being embedded in a form.

## Related issues
Related to #3710


hub/models.py
```diff
@@ -45,13 +52,33 @@ def __str__(self):
...
+            return HttpResponseNotModified()
+
+        if request.method == 'HEAD':
+            response = HttpResponse()
+        else:
+            response = FileResponse(content.open())
+
+        response['Content-Type'] = mimetype
+        response['Content-Length'] = size
+        response["Last-Modified"] = http_date(mtime)
```
noliveleger, line 80: Any reason why using double-quotes here?

jnm: whoops, copy & paste :flushed: fixed!

noliveleger:
@jnm FYI , LGTM, I would have merged it ;-)

PR #4317, opened 2023-03-07
Stripe Checkout link endpoints

LMNTL:
## Checklist

3. [x] Ensure the tests pass

## Description

Adds two KPI endpoints, one that generates a Stripe checkout session and returns a payment link, and one that creates a Customer Portal link.

### DIFF
--- a/kobo/apps/stripe/tests/test_link_creation_api.py
+++ b/kobo/apps/stripe/tests/test_link_creation_api.py
@@ -1,66 +1,61 @@
-import stripe
-
-from django.contrib.auth.models import User
-from django.urls import reverse
-
-from djstripe.models import Customer, Price, Subscription
-from model_bakery import baker
-from rest_framework import status
-from unittest.mock import patch
-
-from kobo.apps.organizations.models import Organization
-from kpi.tests.kpi_test_case import BaseTestCase
-
-
-class TestCheckoutLinkAPITestCase(BaseTestCase):
-
-    fixtures = ['test_data']
-
-    def setUp(self):
-        self.someuser = User.objects.get(username='someuser')
-        self.client.force_login(self.someuser)
-
-    def _insert_data(self):
-        organization = baker.make(Organization)
-        organization.add_user(self.someuser, is_admin=True)
-        self.customer = baker.make(Customer, subscriber=organization)
-        price = baker.make(Price, active=True, product__active=True)
-        self.subscription = baker.make(
-            Subscription,
-            customer=self.customer,
-            items=[{'price': price}],
-            livemode=False,
-        )
-        self.url = reverse(
-            'checkoutlinks',
-            kwargs={
-                'organization_uid': organization.uid,
-                'price_id': price.id
-            }
-        )
-
-    @patch("stripe.Customer.modify")
-    @patch("djstripe.models.Customer.get_or_create")
-    @patch("stripe.checkout.Session.create")
-    def test_creates_organization(
-        self, stripe_customer_modify_mock, customer_get_or_create_mock, stripe_checkout_create_mock,
-    ):
-        self._insert_data()
-        stripe_customer_modify_mock.return_value = None
-        customer_get_or_create_mock.return_value = Customer
-        stripe_checkout_create_mock.return_value = {'url': 'https://https://checkout.stripe.com/c/pay/cs_test_a1NbsdWp'}
-        response = self.client.post(self.url)
-        assert response.status_code == status.HTTP_200_OK
-        assert response.data['url'].startswith('https://checkout.stripe.com')
-
-    def test_get_endpoint(self):
-        self._insert_data()
-        response_get_list = self.client.get(self.url_list)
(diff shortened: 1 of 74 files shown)


kobo/apps/stripe/tests/test_link_creation_api.py
```diff
@@ -0,0 +1,66 @@
...
+    @patch("stripe.Customer.modify")
+    @patch("djstripe.models.Customer.get_or_create")
+    @patch("stripe.checkout.Session.create")
+    def test_creates_organization(
+        self, stripe_customer_modify_mock, customer_get_or_create_mock, stripe_checkout_create_mock,
+    ):
+        self._insert_data()
+        stripe_customer_modify_mock.return_value = None
+        customer_get_or_create_mock.return_value = Customer
+        stripe_checkout_create_mock.return_value = {'url': 'https://https://checkout.stripe.com/c/pay/cs_test_a1NbsdWp'}
```
bufke, line 51: Typo?

kobo/apps/stripe/views.py
```diff
@@ -1,20 +1,136 @@
...
+class CheckoutLinkView(
+    APIView
+):
+    permission_classes = (IsAuthenticated,)
+    serializer_class = CheckoutLinkSerializer
+
+    @staticmethod
+    def generate_payment_link(price_id, user, organization_uid):
+        try:
+            organization = Organization.objects.get(uid=organization_uid)
```
bufke, line 38: Use a get_or_create

bufke: We discussed making org uid Nullable - meaning the backend should create a new org

kobo/apps/stripe/tests/test_link_creation_api.py
LMNTL: Good catch!

kobo/apps/stripe/views.py
```diff
@@ -1,12 +1,137 @@
...
+    permission_classes = (IsAuthenticated,)
+    serializer_class = CheckoutLinkSerializer
+
+    @staticmethod
+    def generate_payment_link(price_id, user, organization_uid):
+        if organization_uid:
+            # Look up the specific organization provided
+            try:
+                organization_user = OrganizationUser.objects.get(user=user)
+                organization = Organization.objects.get(
```
bufke, line 42: Combine these into just one query. Something like `Organization.objects.get(organization_users__user=user, uid=organization_uid)`

kobo/apps/stripe/views.py
```diff
@@ -1,12 +1,137 @@
...
+                    uid=organization_uid
+                )
+            # If we can't find the organization or organization user, bail
+            except ObjectDoesNotExist:
+                return Response(
+                    {'status': r'Organization does not exist'}, status=status.HTTP_403_FORBIDDEN
+                )
+        else:
+            # find the organization the user belongs to, otherwise make a new one
+            organization_user, _ = OrganizationUser.objects.get_or_create(user=user)
```
bufke, line 52: Use `from organizations.utils import create_organization` instead. Used in organizations/views.py (not sure inefficient it is, but you won't be refreshing queries at least like it does there.)

kobo/apps/stripe/views.py
```diff
@@ -1,12 +1,137 @@
...
+            organization, _ = Organization.objects.get_or_create(users=user)
+        # Make sure the organization owner is the same as the logged-in user
+        if organization.owner.id != organization_user.id:
```
bufke, line 55: Add this to an above query instead of making this check here. I'm pretty sure the organization.owner relation is causing yet another query. 1 query by itself isn't worth worrying about but there's a lot here. I think the code would be more readable if it was more at the database level too. 

When org uid is set, make it something like `Organization.objects.get(owner_user=user, uid=organization_uid)`
When it isn't set, the create_organization will already guarantee ownership.

kobo/apps/stripe/views.py
```diff
@@ -1,12 +1,137 @@
...
+            },
+            mode="subscription",
+            payment_method_types=["card"],
+            success_url=f'{settings.KOBOFORM_URL}/#/plans?checkout_complete=true',
+        )
+        return session
+
+    def post(self, request):
+        serializer = CheckoutLinkSerializer(data=request.query_params)
+        if not serializer.is_valid():
```
bufke, line 89: Passing raise_exception=True to is_valid would be slightly more concise.

kobo/apps/stripe/views.py
```diff
@@ -1,12 +1,137 @@
...
+            return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+        serializer_data = serializer.validated_data
+        price_id = serializer_data['price_id']
+        try:
+            organization_uid = serializer.validated_data['organization_uid']
```
bufke, line 94: Nit-pick, but I would write this as `serializer.validated_data.get('organization_uid'). 1 line vs 4 and just as readable.

bufke (changes requested):

PR #4329, opened 2023-03-08
Stripe endpoint tests

LMNTL:
## Checklist

3. [X] Ensure the tests pass

## Description

Adds tests for the Stripe checkout link and customer portal API endpoints. Also fixes some issues with the endpoints discovered while getting tests passing.

### DIFF
--- a/kobo/apps/stripe/views.py
+++ b/kobo/apps/stripe/views.py
@@ -1,7 +1,6 @@
 import stripe
 
 from django.conf import settings
-from django.core.exceptions import ObjectDoesNotExist'
 from django.db.models import Prefetch
 
 from djstripe.models import Customer, Price, Product, Subscription, SubscriptionItem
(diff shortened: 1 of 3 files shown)


kobo/apps/stripe/views.py
```diff
@@ -82,22 +74,18 @@ def start_checkout_session(customer_id, price_id, organization_uid):
...
         organization_uid = serializer.validated_data.get('organization_uid')
-        try:
-            Price.objects.get(
-                active=True,
-                product__active=True,
-                id=price_id
-            )
-        except ObjectDoesNotExist:
-            return Response({'status': 'Price not found or inactive'}, status=status.HTTP_404_NOT_FOUND)
+        Price.objects.get(
```
bufke, line 84: Is the idea here that we actively want it to error when no price exists? I don't think I have an opinion between a GET with 404 vs POST. I'm not sure I like this uncaught DoesNotExist possibility. An end user could not understand our API and send random price ids that don't exist. That should generate a 4xx level error. I could see some justification to log a warning to help get our attention but I don't see it as an error if the end user of the API can generate it.

It could be solved by adjusting the serializer to actually reference the Price by id instead of just a CharField. Unsure if this would make the code easier to read or not.

Not a blocker but let's change or discuss this.

bufke (approved):

PR #4331, opened 2023-03-09
Document how to run Python Tests

bufke:
## Description

Document how to run backend tests with pytest.

## Details

I decided to add this to the KPI README as it seems like basic run instructions. An alternative would be the kobotoolbox.github.io project, but that felt more like best practices and expanding docs. Other basic instructions are already on this readme, such as how to modify dependencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,13 @@ If you do not want to upgrade at this time, please use the [`shared-database-obs
 
 Python dependencies are managed with `pip-compile` and `pip-sync` from the [`pip-tools`](https://github.com/jazzband/pip-tools/) package. The dependencies are listed in [`dependencies/pip/`](./dependencies/pip/), with core requirements in [`dependencies/pip/requirements.in`](./dependencies/pip/requirements.in). You may use `pip` directly with one of the compiled `dependencies/pip/*.txt` files, but consider using instead the `pip-sync`. **Do not** add new dependencies directly to the *compiled* `dependencies/pip/*.txt` files; instead, update the relevant the *source* `dependencies/pip/*.in` file(s), and execute `./pip-compile.sh` after any changes. You can pass arguments to `pip-compile.sh` with e.g. `./pip-compile.sh --upgrade-package=xlwt`.
 
+## Python Tests
+
+Pytest is used for backend tests. It's easiest to run tests inside the docker container.
+
+1. Ensure Python dev dependencies are installed. `pip install -r dependencies/pip/dev_requirements.txt`
+2. Run pytest. An example to run tests at a specific path and reuse the database: `pytest --reuse-db kobo/apps/foo`. --reuse-db speeds up the test, but should be omitted to remake the database fresh.
+
 ## Downloading and compiling the translations
 
 * Pull the submodule into the `locale` directory with `git submodule update`.


README.md
```diff
@@ -16,6 +16,13 @@ If you do not want to upgrade at this time, please use the [`shared-database-obs
 
 Python dependencies are managed with `pip-compile` and `pip-sync` from the [`pip-tools`](https://github.com/jazzband/pip-tools/) package. The dependencies are listed in [`dependencies/pip/`](./dependencies/pip/), with core requirements in [`dependencies/pip/requirements.in`](./dependencies/pip/requirements.in). You may use `pip` directly with one of the compiled `dependencies/pip/*.txt` files, but consider using instead the `pip-sync`. **Do not** add new dependencies directly to the *compiled* `dependencies/pip/*.txt` files; instead, update the relevant the *source* `dependencies/pip/*.in` file(s), and execute `./pip-compile.sh` after any changes. You can pass arguments to `pip-compile.sh` with e.g. `./pip-compile.sh --upgrade-package=xlwt`.
 
+## Python Tests
+
+Pytest is used for backend tests. It's easiest to run tests inside the docker container.
```
jnm, line 21: What do you really need to run the tests? I think this is the most valuable information we can add to the README in order to reduce wasted time and frustration.

I've just confirmed that you must have:
1. A PostgreSQL server running and configured via `KPI_DATABASE_URL`, e.g.
    ```
    export KPI_DATABASE_URL=postgis://postgres:insecure-postgres-kobo-dev@10.6.6.1:60666/koboform
    ```
1. A Redis server running and configured as both the session and cache back end, e.g.
    ```
    export REDIS_SESSION_URL=redis://:insecure-redis-kobo-dev@10.6.6.1:60667/1
    export CACHE_URL=redis://:insecure-redis-kobo-dev@10.6.6.1:60667/2
    ```
1. At least 5 UI languages enabled :upside_down_face: one of which must be Arabic (but perhaps not for long, since we are finally adding a default for this?):
    ```
    export DJANGO_LANGUAGE_CODES="en fr es ar zh-hans hi ku"
    ```

We should state these specific requirements, since they are useful for everyone (not just Docker folks) and also because "inside the docker container" doesn't really mean anything unless the developer has a priori knowledge of what Docker setup they're supposed to be using.

For a Docker-based setup, we should recommend that people use the developer mode of kobo-install (and link to it) if they don't want to set up and configure database servers manually. In the future, we may have a more officially-sanctioned replacement for kobo-install as a development environment, but we're not there yet. People are currently using kobo-install, discussing it on the forum, receiving occasional support there from the core team, etc.

README.md
```diff
@@ -16,6 +16,13 @@ If you do not want to upgrade at this time, please use the [`shared-database-obs
...
+
+1. Ensure Python dev dependencies are installed. `pip install -r dependencies/pip/dev_requirements.txt`
+2. Run pytest. An example to run tests at a specific path and reuse the database: `pytest --reuse-db kobo/apps/foo`. --reuse-db speeds up the test, but should be omitted to remake the database fresh.
```
jnm, line 24: I'd prefer to let people read about `pytest` on their own and decide if `--reuse-db` is appropriate. Personally, that's an option that I use on occasion, but not as the norm. If you're confident that running it as the norm is fine, then leave the tip, but please use backticks around code consistently.

jnm (changes requested):

bufke:
@jnm responding to all your comments - I think the docs should balance between a user who already knows about kobo-install, pytest, django testing on one hand. On the other - a user who doesn't know this and is perhaps using Docker without even knowing it.

What do you think about a middle ground of avoiding getting into KPI_DATABASE_URL and is either using kobo-install or kobo-compose or their own Docker Compose based solution? If not using Docker, it's quite a bit harder to run these tests. I'm happy to also discuss Docker based dev vs not, but not in this issue.

PR #4367, opened 2023-03-17
Improve XML parsing with defusedxml

JacquelineMorrissette:
## Checklist

3. [x] Ensure the tests pass

## Description
Switch to defusedxml for xml parsing

## Other notes
The defusedxml library doesn't replace everything in the native XML library so it's only been changes where necessary

## Related issues
kobotoolbox/kobocat#869

### DIFF
--- a/kpi/deployment_backends/kobocat_backend.py
+++ b/kpi/deployment_backends/kobocat_backend.py
@@ -11,16 +11,18 @@ from collections import defaultdict
 from datetime import date, datetime
 from typing import Generator, Optional, Union
 from urllib.parse import urlparse
-from xml.etree import ElementTree as ET
+from xml.etree.ElementTree import ElementTree, ParseError
 try:
     from zoneinfo import ZoneInfo
 except ImportError:
     from backports.zoneinfo import ZoneInfo
 
 import requests
+from defusedxml.ElementTree import fromstring, tostring
+from defusedxml.lxml import tostring
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
-from lxml import etree
+from lxml.etree import SubElement
 from django.core.files import File
 from django.db.models import Sum
 from django.db.models.query import QuerySet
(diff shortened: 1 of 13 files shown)


kpi/deployment_backends/kobocat_backend.py
```diff
@@ -11,16 +11,18 @@
...
-from xml.etree import ElementTree as ET
+from xml.etree.ElementTree import ElementTree, ParseError
 try:
     from zoneinfo import ZoneInfo
 except ImportError:
     from backports.zoneinfo import ZoneInfo
 
 import requests
+from defusedxml.ElementTree import fromstring, tostring
+from defusedxml.lxml import tostring
```
jnm, line 22: Hi @JacquelineMorrissette, could you double-check this PR with a few things in mind?
1. `fromstring` and `tostring` are awfully generic-sounding things. I think the usual convention of `… import ElementTree as ET`, an then using stuff like `ET.fromstring` is good to avoid ambiguity.
    1. Watch out for shadowing one import with another, as in this case where the second `tostring` import completely replaces the first.
1. Why are we still using `xml.etree.ElementTree` and `lxml` versus `defusedxml.ElementTree` and `defusedxml.lxml`? Are there certain things we need that the defusedxml alternatives don't provide?

Edit: whoops, I just saw this in defusedxml's readme:
> defusedxml.lxml
> DEPRECATED The module is deprecated and will be removed in a future release.

Given that, please don't use `defusedxml.lxml` anywhere. Try to see how our uses of `lxml` could be replaced with `ElementTree`, and don't hesitate to ask questions :)

jnm (changes requested):

kpi/deployment_backends/kobocat_backend.py
JacquelineMorrissette: 1. Will do, I'll give it another pass
2. The reason I kept some uses of `xml.etree.ElementTree` (and others) is because the defusedxml library isn't a complete replacement for most features of these and only replaces specific functionalities such as `tostring` and `fromstring`. For example: defusexml doesn't have an equivalent to `from xml.etree.ElementTree import ElementTree`
The documentation here lists what it replaces, in case it helps: https://pypi.org/project/defusedxml/#defusedxml-elementtree

noliveleger:
Closed by #5031

PR #4371, opened 2023-03-20
Endpoint to list purchased one-time add-ons

LMNTL:
## Checklist

3. [ ] Ensure the tests pass

## Description

Adds a new internal endpoint at `api/v2/stripe/addons` that lists any one-time add-ons purchased by the logged-in user's owned organization. Needed for the "Your Active Add-ons" section of the Plans page.

## Sample API response:
image
```
{
    "count": 1,
    "next": null,
    "previous": null,
    "results": [
        {
            "metadata": {
                "price_id": "price_1MmO5sAR39rDI89sfdKUjter",
                "organization_uid": "orgPDBt2kqPpuG9a5bYsCd4S"
            },
            "created": "2023-03-20T14:06:13Z",
            "payment_intent": "succeeded"
        }
    ]
}
```

## Note
This PR will only list add-ons purchased through a Stripe checkout session initiated from the Plans page (or using the `api/v2/checkout-link` endpoint.) This is because it stores the price metadata on the checkout Session.

### DIFF
--- a/kobo/apps/stripe/serializers.py
+++ b/kobo/apps/stripe/serializers.py
@@ -1,12 +1,23 @@
 from django.core.exceptions import SuspiciousOperation, ValidationError
 from djstripe.models import (
-    Customer,
+    Session,
     Price,
     Product,
     Subscription,
     SubscriptionItem,
 )
 from rest_framework import serializers
+
+
+class OneTimeAddOnSerializer(serializers.ModelSerializer):
+    payment_intent = serializers.SlugRelatedField(
+        slug_field='status',
+        read_only=True,
+        many=False,
+    )
+    class Meta:
+        model = Session
+        fields = ('metadata', 'created', 'payment_intent',)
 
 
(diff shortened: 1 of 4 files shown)


bufke (changes requested):
This looks good, it just needs a unit test. You have my approval to merge once this is addressed.

Session seems odd to me but appears to be the correct Stripe way. I guess this would also return subscription checkouts, but that might be fine? What do you think? We could filter them out if that's easier. The UI is always showing a list of purchased addons or a list of subscriptions. It's never listing them both, mixed.

LMNTL:
@bufke Adding the unit tests now.

As written, this endpoint will only show non-subscription add-on purchases (via the `mode='payment'` argument). The frontend will already have a list of subscription-based add-on purchases, because it calls to `api/v2/stripe/subscriptions`, which doesn't care whether a subscription is an add-on or a main plan (Community, Professional). So my thought was to just have a second endpoint and combine the results (subscription + one-time add-ons) together on the frontend.

PR #4385, opened 2023-03-24
Feature/add subscription tiers

srartese:
## Description
This is a feature - Added the plans page for subscriptions with stripe. Users are able to see current plan and upgrade through stripes checkout to a different subscription plan.

Also removes a too-narrow assertion in [test_link_creation_api.py](https://github.com/kobotoolbox/kpi/pull/4385/commits/7e4b4357168b924212d228bc3e06beba4a9fb6b6#diff-7d81d02e1346c12275169d92c8f79dd9dd58e05b95a3fc3a31aba1492374e4e4).


jsapp/js/account/plan.component.tsx
```diff
@@ -0,0 +1,339 @@
...
+          prices: filteredPrices.length ? filteredPrices[0] : null,
+        };
+      });
+      return filterAmount.filter((product: any) => product.prices);
+    }
+  };
+
+  const isSubscribedProduct = (product: any) => {
+    return product.name === state.subscribedProduct?.name;
+  };
```
LMNTL, line 96: It looks like I didn't write this function correctly, which is why the "Your Plan" tab isn't showing up on the frontend.

```suggestion
  const isSubscribedProduct = (product: Product) => {
    if( product.prices.unit_amount === 0 && !state.subscribedProduct?.length ) {
      return true;
    }
    return product.name === state.subscribedProduct?.name;  };
  };
```

jsapp/js/account/plan.component.tsx
```diff
@@ -0,0 +1,339 @@
...
+          case 'addons':
+            const currentItemIcon = `feature_addons_check_${currentResult}`;
+            if (featureItem.includes('feature_addons_') && !featureItem.includes('feature_addons_check') && price.name === plan) {
+              const keyName = `feature_addons_${currentResult}`;
+              let iconBool = false;
+              const itemName: string = price.metadata[keyName];
+              if (price.metadata[currentItemIcon]) {
+                iconBool = JSON.parse(price.metadata[currentItemIcon]);
+              }
+              listItems.push({icon: iconBool, item: itemName});
```
bufke, line 179: There's a lot of repeated logic and short lived variables like iconBool. I think this could be refactored. But I'd have to play around with it to give more specific advice.

LMNTL (changes requested):
This is looking good! I have some changes to request and some general comments.

The frontend will need to be refactored a bit to account for an organization having multiple subscriptions, since some add-ons are recurring. But I think we only need to cross that bridge once the add-ons section has been implemented, so it's not blocking this PR.

jsapp/js/account/planRoute.tsx
```diff
@@ -8,7 +8,7 @@ import type {ServiceUsage} from './dataUsageStore';
 import {handleApiFail} from 'js/utils';
 import {ROOT_URL} from 'js/constants';
 import type {PaginatedResponse} from 'js/dataInterface';
-import './planRoute.scss';
+import {setState} from "reflux";
```
bufke, line 11: I don't see this being used anywhere

jsapp/js/account/plan.component.tsx
```diff
@@ -0,0 +1,339 @@
...
+                    ))
+                  }
+                </ul>
+                {!isSubscribedProduct(price) &&
+                  <button
+                    className={[
+                      styles.resetButton,
+                      styles.upgradeBtn,
+                    ]
+                      .join(' ')}
```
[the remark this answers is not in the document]
srartese: I saw recently some of the Devs talking about how to do this - its a great recommendation so we should bring it to that conversation. I will tag you in it!

jsapp/js/account/plan.component.tsx
```diff
@@ -0,0 +1,339 @@
...
+                              {listItem.item}</li>
+                            :
+                            <li key={listItem.item}><div className={styles.iconContainer}><Icon name='close' size='m' /></div>{listItem.item}</li>
+                          )
+                        ))
+                      }
+                    </ul>
+                  </div>
+                }
+              </div>
```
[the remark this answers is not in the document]
srartese: I am going to leave it for now because I do think there is a good way to refactor and reuse the code so its not this long

jsapp/js/account/plan.component.tsx
```diff
@@ -0,0 +1,339 @@
...
+    setButtonDisabled(buttonsDisabled);
+    $.ajax({
+      dataType: 'json',
+      method: 'POST',
+      url: `${ROOT_URL}/api/v2/stripe/customer-portal?organization_uid=${state.organization?.uid}`,
+    })
+      .done(function (res) {
+        window.location.assign(res.url);
+      })
+      .fail(handleApiFail)
```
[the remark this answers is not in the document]
srartese: For now I set up an alert ( its not the prettiest) but I do think it would be nice to have it in a different type of popup.

jsapp/js/account/stripe.api.ts
```diff
@@ -0,0 +1,76 @@
...
+export interface BaseProduct {
+  id: string;
+  name: string;
+  description: string;
+  type: string;
+  metadata: {[key: string]: string};
+}
+
+export interface BasePrice {
+    id: string;
```
bufke, line 13: Don't forget prettier :smiley:

kobo/apps/stripe/tests/test_link_creation_api.py
```diff
@@ -55,7 +55,6 @@ def test_rejects_invalid_query_params(
         url = self._get_url({'price_id': 'test', 'organization_uid': 'test'})
         response = self.client.post(url)
         assert response.status_code == status.HTTP_400_BAD_REQUEST
-        assert response.data['price_id'].count('Invalid price ID') == 1
```
bufke, line 58: It would be good to add a comment in the Pull Request as to why this was done. But no code comment is needed.

jsapp/js/account/plan.component.tsx
```diff
@@ -0,0 +1,440 @@
...
+        filterToggle: !state.filterToggle,
+      };
+    default:
+      return state;
+  }
+}
+
+export default function Plan() {
+  const [state, dispatch] = useReducer(planReducer, initialState);
+  const [expandComparison, setExpandComparison] = useState(false);
```
bufke, line 68: I'm curious why you choose to call the top variable "state" which sounds like it might contain the entire component's state but then added more useState's. I'm not necessarily against this if you feel that these boolean states are very unrelated.

jsapp/js/account/plan.component.tsx
```diff
@@ -0,0 +1,339 @@
...
+        window.location.assign(res.url);
+      })
+      .fail(handleApiFail)
+      .always(() =>
+        setButtonDisabled(!buttonsDisabled)
+      );
+  };
+
+  // Get feature items and matching icon boolean
+  const getListItem = (listType: string, plan: string) => {
```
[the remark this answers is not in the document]
bufke: We discussed previously that I have a slight lean towards using useMemo more often for consistency and why not. But it's true that it probably won't ever matter how many "extra" times this JS is executed. I'll make the call to say that this is not a merge blocker but happy to keep the conversation going.

jsapp/js/account/plan.component.tsx
```diff
@@ -0,0 +1,440 @@
...
+  const checkMetaFeatures = () => {
+    if (state.products.length > 0) {
+      filterPrices().map((price) =>
+        Object.keys(price.metadata).map((featureItem: string) => {
+          if (
+            featureItem.includes(`feature_support_`) ||
+            featureItem.includes(`feature_advanced_`) ||
+            featureItem.includes(`feature_addon_`)
+          ) {
+            setShowExpand(true);
```
bufke, line 188: This is a computed value, there is no need to use state. The function should return a boolean instead.

jsapp/js/account/plan.component.tsx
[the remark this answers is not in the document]
bufke: Worth discussing and not a blocker.

jsapp/js/account/plan.component.tsx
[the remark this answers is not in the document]
bufke: Let's not forget to do this - I do find the file as a whole a bit hard to read with all the nesting and branching jsx logic. Agreed that it's not a blocker.

bufke (commented):
This is getting there - I tried to carefully note what needs changed now vs what might be worth discussing/doing later and is not a merge blocker.

kobo/apps/stripe/tests/test_link_creation_api.py
srartese: @LMNTL I am guessing this was one of your changes?

LMNTL: Yes! I refactored `CheckoutLinkView` and its serializer, which changed the error that would get produced and broke that test. Then I realized the test case was too narrow (really we just care that we got the right status) so I removed that assertion.

jsapp/js/account/plan.component.tsx
```diff
@@ -0,0 +1,442 @@
...
+            {!isSubscribedProduct(price) && (
+              <button
+                className={[styles.resetButton, styles.upgradeBtn].join(' ')}
+                onClick={() => upgradePlan(price.prices.id)}
+                aria-label={`upgrade to ${price.name}`}
+                disabled={buttonsDisabled}
+                aria-disabled={buttonsDisabled}
+              >
+                {t('Upgrade')}
+              </button>
```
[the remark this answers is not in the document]
magicznyleszek: I think we should move the bottom margin of button to the line element to avoid unnecessary bottom space when details are collapsed

jsapp/js/account/plan.component.tsx
```diff
@@ -0,0 +1,442 @@
...
+                  disabled={buttonsDisabled}
+                  aria-disabled={buttonsDisabled}
+                  aria-label={`manage your ${price.name} subscription`}
+                >
+                  {t('Manage')}
+                </button>
+              )}
+            {expandComparison && (
+              <div>
+                <div className={styles.line} />
```
magicznyleszek, line 290: Nitpicking: we could use the ancient `<hr>` here instead of `<div>` ❤️ Then you could target it with some `.planContainer :global hr`

jsapp/js/account/plan.component.tsx
```diff
@@ -0,0 +1,442 @@
...
+              'We offer add-on options to increase your limits or the capacity of certain features for a period of time. Scroll down to learn more and purchase add-ons.'
+            )}
+          </p>
+          <p className={styles.enterpriseDetails}>
+            {t(
+              'If your organization has larger or more specific needs, contact our team to learn about our enterprise options.'
+            )}
+          </p>
+          <div className={styles.enterpriseLink}>
+            <a href='https://www.kobotoolbox.org/contact/' target='_blanks'>
```
magicznyleszek, line 423: @jnm do you think this contact url should come from Back-end?

jsapp/js/account/plan.component.tsx
```diff
@@ -0,0 +1,442 @@
...
+
+      {showExpand && (
+        <div
+          className={styles.expandBtn}
+          role='button'
+          onClick={() => setExpandComparison(!expandComparison)}
+        >
+          {' '}
+          {expandComparison ? t('Collapse') : t('Display full comparison')}
+        </div>
```
magicznyleszek, line 438: This could also use the `<Button>` component with `isFullWidth` prop.

jsapp/js/account/plan.module.scss
```diff
@@ -0,0 +1,232 @@
...
+}
+
+.intervalToggle label {
+  padding: sizes.$x8 sizes.$x12;
+  color: colors.$kobo-dark-blue;
+}
+
+.intervalToggle input[type="radio"]:checked+label {
+  background-color: colors.$kobo-blue;
+  animation-name: blink;
```
magicznyleszek, line 43: I don't see this animation being defined anywhere

jsapp/js/account/plan.module.scss
```diff
@@ -0,0 +1,232 @@
...
+  line-height: 24.5px;
+}
+
+.enterpriseLink a {
+  color: colors.$kobo-dark-blue;
+  text-decoration: underline;
+  font-weight: 500;
+}
+
+.listTitle {
```
magicznyleszek, line 149: This text should be aligned perfectly with the list items text

jsapp/js/account/plan.component.tsx
```diff
@@ -0,0 +1,442 @@
...
+                >
+                  {t('Manage')}
+                </button>
+              )}
+            {expandComparison && (
+              <div>
+                <div className={styles.line} />
+                {getListItem('support', price.name).length > 0 && (
+                  <ul>
+                    <li className={styles.listTitle}>
```
magicznyleszek, line 293: It would be more semantic to not make the list title one of `<li>`s. See https://stackoverflow.com/questions/1141639/how-to-semantically-provide-a-caption-title-or-label-for-a-list-in-html

jsapp/js/account/plan.module.scss
```diff
@@ -0,0 +1,232 @@
...
+  color: colors.$kobo-dark-blue;
+  text-align: center;
+  font-size: sizes.$x18;
+  padding-bottom: sizes.$x16;
+  font-weight: 600;
+}
+
+.line {
+  border-bottom: 1px solid colors.$kobo-gray-85;
+  border-radius: 5px;
```
magicznyleszek, line 97: The `border-radius` should be removed

jsapp/js/account/plan.module.scss
```diff
@@ -0,0 +1,232 @@
...
+  width: 100%;
+  margin-bottom: 20px;
```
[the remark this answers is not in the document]
magicznyleszek: Also as noted in other comment - it would be simpler if only the line would have top and bottom margins, and both button and the list below wouldn't

jsapp/js/account/plan.module.scss
magicznyleszek: Also the top margin should be removed and the `<ul>` element should handle margins. I think a clean solution would be to have top margins only on `list + list`. This way the line element would be a nice divider with margins, and then the lists will have gaps between them. Make sure to handle this margin cleanup after you change the list title HTML to be more semantic

jsapp/js/account/plan.component.tsx
```diff
@@ -0,0 +1,442 @@
...
+            id='switch_right'
+            name='switchToggle'
+            value='month'
+            onChange={() => dispatch({type: 'month'})}
+            checked={state.filterToggle}
+          />
+          <label htmlFor='switch_right'> {t('Monthly')}</label>
+        </form>
+        <div
+          className={styles.currentPlan}
```
magicznyleszek, line 222: I find it a bit confusing that it's not the part of the `.planContainer` that is subscribed. Would it be hard to change it now?

jsapp/js/account/plan.component.tsx
magicznyleszek, line 1: Are we not doing the "Compare plans" button? I don't know what decisions were made, just verifying :)

Screenshot 2023-03-29 at 16 28 13

jsapp/js/account/plan.component.tsx
```diff
@@ -0,0 +1,442 @@
+import React, {useEffect, useReducer, useState} from 'react';
+import styles from './plan.module.scss';
+import {
```
magicznyleszek, line 3: When importing types please use `import type` and split this into two imports

jsapp/js/account/plan.component.tsx
```diff
@@ -0,0 +1,442 @@
...
+  isLoading: boolean;
+  subscribedProduct: BaseSubscription;
+  intervalFilter: string;
+  filterToggle: boolean;
+  products: Product[];
+  organization: null | Organization;
+}
+
+// An interface for our action
+interface dataUpdates {
```
magicznyleszek, line 27: We follow PascalCase convention for interface names :)

jsapp/js/account/plan.component.tsx
```diff
@@ -0,0 +1,442 @@
...
+  };
+
+  const checkMetaFeatures = () => {
+    if (state.products.length > 0) {
+      filterPrices().map((price) =>
+        Object.keys(price.metadata).map((featureItem: string) => {
+          if (
+            featureItem.includes(`feature_support_`) ||
+            featureItem.includes(`feature_advanced_`) ||
+            featureItem.includes(`feature_addon_`)
```
magicznyleszek, line 186: Single quote please

jsapp/js/account/plan.component.tsx
```diff
@@ -0,0 +1,442 @@
...
+          ) {
+            setShowExpand(true);
+          } else {
+            setShowExpand(false);
+          }
+        })
+      );
+    }
+  };
+  if (!state.products.length) return null;
```
magicznyleszek, line 196: We expect single line `if` to still include curly braces

jsapp/js/account/plan.component.tsx
```diff
@@ -0,0 +1,442 @@
...
+            {expandComparison && (
+              <div>
+                <div className={styles.line} />
+                {getListItem('support', price.name).length > 0 && (
+                  <ul>
+                    <li className={styles.listTitle}>
+                      {price.metadata.feature_support_title}
+                    </li>
+                    {getListItem('support', price.name).map((listItem) =>
+                      listItem.icon ? (
```
magicznyleszek, line 297: This triggers the linter [no-confusing-arrow](https://eslint.org/docs/latest/rules/no-confusing-arrow) rule. I see this whole function contents are repeated 3 times for different `getListItem`s. Let's create a `renderListItem` function for this to make it DRY.

jsapp/js/account/stripe.api.ts
```diff
@@ -0,0 +1,82 @@
...
+
+export interface Portal {
+  url: string;
+}
+
+const PRODUCTS_URL = '/api/v2/stripe/products/';
+const SUBSCRIPTION_URL = '/api/v2/stripe/subscriptions/';
+const ORGANIZATION_URL = '/api/v2/organizations/';
+const CHECKOUT_URL = '/api/v2/stripe/checkout-link';
+const PORTAL_URL = '/api/v2/stripe/customer-portal';
```
magicznyleszek, line 56: I was thinking about this for some time last week. Do you think it would make sense to have a single file somewhere at the top directory with a list of all API endpoints we use? It would be nice to use these instead of hardcoding them

srartese: Yeah I think that would be really helpful because then it reduces having to update a URL in multiple locations/files if the URL changes. I do think it would make sense to separate endpoints based on use, for example, this would live in a Stripe Endpoint file and we have a directory for all endpoint files

jsapp/js/account/plan.component.tsx
```diff
@@ -0,0 +1,442 @@
...
+                      <div className={styles.iconContainer}>
+                        <Icon
+                          name='check'
+                          size='m'
+                          classNames={
+                            price.name === 'Professional plan'
+                              ? [styles.tealCheck]
+                              : [styles.stormCheck]
+                          }
+                        />
```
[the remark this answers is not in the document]
srartese: Yeah I think it would be nice to pass in color as a parameter and if one isn't provided we stick with the default. I see it being useful in other scenarios as well. I will add as a to do on my list for improvements but not blockers

jsapp/js/account/plan.component.tsx
srartese: I was talking to Igancio about the option of this also filling data to the form when it redirects as well. Specifically adding a topic  and having it fill in 'Enterprise' or whatever it may be

jsapp/js/account/plan.component.tsx
srartese: As far as I am aware not for the MVP

jsapp/js/account/plan.component.tsx
srartese: Why split it into two imports?

jsapp/js/account/plan.module.scss
```diff
@@ -0,0 +1,232 @@
...
+  column-gap: sizes.$x20;
+}
+
+.intervalToggle {
+  grid-column: 1 / 5;
+  grid-row: 1/2;
+  justify-self: center;
+  background-color: colors.$kobo-cloud;
+  padding: sizes.$x10 sizes.$x4;
+  border-radius: sizes.$x4;
```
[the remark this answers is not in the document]
srartese: > Figma file has `border-radius` at `6px` and padding at `3px`

So this is true for the blue button - if you looks close the background pops up a little behind the blue and is the cloud. That has a different radius than the active one.

jsapp/js/account/plan.module.scss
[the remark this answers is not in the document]
srartese: If there is something that does have a variable size should I add it? What are our rules for adding sizes?

jsapp/js/account/plan.component.tsx
srartese: Haven't thought of an `<hr>` in a while! I love it

jsapp/js/account/plan.component.tsx
srartese: So in a few examples for using useReducer it used "state" which I thought made sense. I added the other useStates instead of filtering it through the reducer because I felt that the reducer was being used for more data driven items while these were visual items

jsapp/js/account/plan.component.tsx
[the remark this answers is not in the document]
srartese: I am open to moving it to a "useMemo" - my only thought against was to follow best practice for only using when it can improve performance since it could do the opposite if we standardize adding it. Honestly I find it hard to tell with some of these smaller functions when it is useful and when its not

jsapp/js/account/plan.component.tsx
srartese: I'm not sure I follow because then it would get rendered multiple times.

jsapp/js/account/plan.module.scss
srartese: > 

It should be already, are you seeing something off?

srartese: > This text should be aligned perfectly with the list items text

It should be already, are you seeing something off?

jsapp/js/account/plan.module.scss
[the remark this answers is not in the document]
magicznyleszek: The rule for missing size variable is something like this:
- if the size seems weird and unusual (e.g. `23px`, `17px` etc.) and we have something almost identical we should use the existing one (e.g. `$x18` instead of `17px`)
- if the size seems something common (e.g. `32px`, repetitions of 10 or 5, etc.) we should probably add it, unless there is something similar already that works well in given situation (e.g. Figma uses `25px`, but we have already `$x24` and the difference in the visible UI is minimal - we should go with existing 24)
- in other situation just add the size and use it :)

jsapp/js/account/plan.component.tsx
magicznyleszek: I thought about it being put inside that `.map` loop, just with a conditional rendering:

```ts
{filterPrices().map((price: Price, i: number) => (
  {isSubscribedProduct(price) &&
    <div
      className={styles.currentPlan}
…
```

magicznyleszek: This way you will get it only once and it would be easier to style it being positioned above proper price box. You'd propably need to add one more wrapper inside that loop

jsapp/js/account/stripe.api.ts
magicznyleszek: Separate files are nice, but on the other hand the number of API endpoints we have is not so big and it might fit nicely in single file :)

magicznyleszek: Perhaps next to `api.ts` as `api.endpoints.ts`?

jsapp/js/account/plan.component.tsx
```diff
@@ -0,0 +1,403 @@
...
+            expandBool = true;
+            break;
+          }
+        }
+      });
+    }
+    return expandBool;
+  };
+
+  const returnListItem = (type: string, name: string): ReactNode => {
```
magicznyleszek, line 206: Could we do something even less confusing here? Like a function that renders a list of features with a title:

```ts
const renderFeaturesList(
  title?: string, 
  items: Array<{
    icon: 'positive' | 'positive_pro' | 'negative', 
    label: string
  }>
) {
  …
}
```

This way we could simplify this even further :)

jsapp/js/account/plan.module.scss
```diff
@@ -0,0 +1,224 @@
...
+}
+
+.btnSpacePlaceholder { 
+  margin: sizes.$x20 0;
+  height: sizes.$x32;
+}
+
+.planContainer :global hr {
+  border-bottom: 1px solid colors.$kobo-gray-85;
+  border-top: 1px solid colors.$kobo-gray-85;
```
magicznyleszek, line 132: Screenshot 2023-04-03 at 12 06 59

This line should be `1px` height, now it is rendered as `2px` :)

jsapp/js/account/plan.component.tsx
```diff
@@ -0,0 +1,403 @@
...
+                  size='m'
+                  label={t('Manage')}
+                  onClick={managePlan}
+                  aria-label={`manage your ${price.name} subscription`}
+                  aria-disabled={buttonsDisabled}
+                  isDisabled={buttonsDisabled}
+                />
+              )}
+            {price.name === 'Community plan' && (
+              <div className={styles.btnSpacePlaceholder} />
```
magicznyleszek, line 335: Why do we need this? Now it is being rendered unnecessarily here:

Screenshot 2023-04-03 at 12 01 50

jsapp/js/account/plan.component.tsx
magicznyleszek: As an argument for moving this - because it is being rendered separately, it's not synced properly for me now:
Screenshot 2023-04-03 at 13 20 32

magicznyleszek (changes requested):

jsapp/js/account/plan.component.tsx
```diff
@@ -0,0 +1,403 @@
...
+            listItems.push({icon: iconBool, item: itemName});
+          }
+        }
+      })
+    );
+    return listItems;
+  };
+
+  const checkMetaFeatures = () => {
+    let expandBool = false;
```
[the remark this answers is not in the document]
srartese: Absolutely and I will add notes to our notes from the front-end meeting

jsapp/js/account/plan.component.tsx
```diff
@@ -0,0 +1,403 @@
...
+            gridRow: 0,
+            gridColumn: 1 + filterPrices().findIndex(isSubscribedProduct),
+            display:
+              filterPrices().findIndex(isSubscribedProduct) >= 0 ? '' : 'none',
+          }}
+        >
+          {t('your plan')}
+        </div>
+        {filterPrices().map((price: Price, i: number) => (
+          <div className={styles.planContainer} key={i}>
```
LMNTL, line 277: React gets kind of iffy when using an index as a key - see [here](https://react.dev/learn/rendering-lists), under 'Pitfall'. This should get replaced with `key={price.id}` or some other unique value.

jsapp/js/account/plan.component.tsx
srartese: So the purpose was the create space when there is no button - I am a little confused why a button is showing there

jsapp/js/account/plan.module.scss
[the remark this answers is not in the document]
srartese: Yeah the more we work with sizes and spacing I am thinking we should give it a try especially since on my screen I am having trouble recreating this view. I will switch it over to flexbox

jsapp/js/account/stripe.api.ts
srartese: Thats fair, I wasn't sure how many we were talking about in total.  I do like `api.endpoints.ts`

jsapp/js/account/plan.module.scss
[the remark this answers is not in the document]
srartese: Sounds good!

jsapp/js/account/plan.component.tsx
srartese: I agree now and moved it

LMNTL (approved):

jsapp/js/account/plan.component.tsx
```diff
@@ -0,0 +1,435 @@
...
+import {
+  getOrganization,
+  getProducts,
+  getSubscription,
+  postCheckout,
+  postCustomerPortal,
+} from './stripe.api';
+import Icon from '../components/common/icon';
+import Button from 'js/components/common/button';
+import {render} from 'react-dom';
```
magicznyleszek, line 19: You're not using it in this file :)

jsapp/js/account/plan.module.scss
```diff
@@ -0,0 +1,232 @@
...
+  padding-bottom: sizes.$x16;
+  font-weight: 600;
+  line-height: sizes.$x20;
+}
+
+.btnPlacement {
+  margin: 0 auto;
+}
+
+.planContainer > Button {
```
magicznyleszek, line 122: `.planContainer :global .k-button` would be more precise :)

jsapp/js/account/plan.module.scss
```diff
@@ -0,0 +1,232 @@
...
+.priceTitle {
+  color: colors.$kobo-dark-blue;
+  text-align: center;
+  font-size: sizes.$x18;
+  padding-bottom: sizes.$x16;
+  font-weight: 600;
+  line-height: sizes.$x20;
+}
+
+.btnPlacement {
```
magicznyleszek, line 118: Unused code

jsapp/js/account/plan.module.scss
```diff
@@ -0,0 +1,232 @@
...
+  appearance: none;
+}
+
+.intervalToggle label {
+  color: colors.$kobo-dark-blue;
+  padding: 0 sizes.$x12;
+  font-size: sizes.$x18;
+  line-height: sizes.$x32;
+  display: inline-block;
+  width: sizes.$x100;
```
magicznyleszek, line 50: Make this `min-width`, so with non-English it will adapt

magicznyleszek (changes requested):
Please make sure to run your linters :) Let me (or Phil!) know if you have any issues with setting up your editor.

Let's make this review round the final polishing :)

magicznyleszek (changes requested):
Please make sure to run your linters :) Let me (or Phil!) know if you have any issues with setting up your editor.

Let's make this review round the final polishing :)

srartese:
> Please make sure to run your linters :) Let me (or Phil!) know if you have any issues with setting up your editor.
> 
> Let's make this review round the final polishing :)

Linting should be cleaned up now, the only one I wasn't sure about it the any warning. I am still learning about the typescript lingo. It is targeting this `prodData?: any;` which can be used to send data or be empty/undefined. I'm not sure how to write it without causing more erros

jsapp/js/account/plan.component.tsx
```diff
@@ -0,0 +1,435 @@
...
+            onChange={() => dispatch({type: 'month'})}
+            checked={state.filterToggle}
+          />
+          <label htmlFor='switch_right'> {t('Monthly')}</label>
+        </form>
+
+        <div className={styles.allPlans}>
+          {filterPrices().map((price: Price) => (
+            <div className={styles.stripePlans} key={price.id}>
+              {isSubscribedProduct(price) ? (
```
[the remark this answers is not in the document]
srartese: Yeah if there is nothing subscribed it should default to the community plan, this might be fixed in the updates coming from Jessica's branch. Do you know we could get it in this state to test?

jsapp/js/account/plan.component.tsx
```diff
@@ -0,0 +1,435 @@
...
+                <div
+                  className={styles.currentPlan}
+                  style={{
+                    display:
+                      filterPrices().findIndex(isSubscribedProduct) >= 0
+                        ? ''
+                        : 'none',
+                  }}
+                >
+                  {t('your plan')}
```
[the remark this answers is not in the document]
srartese: I think this also get covered by Jessica's updates in the PR I just approved. Should get merged into this branch tomorrow 🎉

jsapp/js/components/common/icon.tsx
```diff
@@ -9,18 +9,20 @@ import './icon.scss';
 export type IconSize = 'l' | 'm' | 's' | 'xl' | 'xs' | 'xxs';
 
 const DefaultSize = 's';
+const DefaultColor = 'inherit';
 
 interface IconProps {
   name: IconName;
   size?: IconSize;
   classNames?: string[];
+  color?: string;
```
magicznyleszek, line 18: Let's make this not any `string`, but a list of limited colors available (from our colors list). Similarly to `export type ButtonColor = 'blue' | 'light-blue' | 'red' | 'storm';`, but instead of defining all that make sense, please start with the ones you use :)

jsapp/js/components/common/icon.tsx
```diff
@@ -36,12 +38,13 @@ class Icon extends React.Component<IconProps, {}> {
...
+    const color = this.props.color || DefaultColor;
+    classNames.push(`k-icon--color-${color}`);
+
     classNames.push('k-icon');
     classNames.push(`k-icon-${this.props.name}`);
 
-    return (
-      <i className={classNames.join(' ')}/>
-    );
+    return <i className={classNames.join(' ')} style={{color: `${color}`}} />;
```
magicznyleszek, line 47: Continuing the above comment: instead of inline style, let's create and use class names :) We don't want any color, we want restricted list of colors

jsapp/js/api.endpoints.ts
```diff
@@ -2,6 +2,8 @@ export const endpoints = {
   PRODUCTS_URL: '/api/v2/stripe/products/',
   SUBSCRIPTION_URL: '/api/v2/stripe/subscriptions/',
   ORGANIZATION_URL: '/api/v2/organizations/',
-  CHECKOUT_URL: '/api/v2/stripe/checkout-link', /** Expected parameters: price_id and organization_uid **/
-  PORTAL_URL: '/api/v2/stripe/customer-portal', /** Expected parameter: organization_uid  **/
+  CHECKOUT_URL:
+    '/api/v2/stripe/checkout-link' /** Expected parameters: price_id and organization_uid **/,
+  PORTAL_URL:
+    '/api/v2/stripe/customer-portal' /** Expected parameter: organization_uid  **/,
```
magicznyleszek, line 8: Move these comments to the line above, instead of endline. This way it will still work fine with editors and it would look cleaner :)

jsapp/js/components/common/icon.tsx
```diff
@@ -36,12 +38,13 @@ class Icon extends React.Component<IconProps, {}> {
     const size = this.props.size || DefaultSize;
     classNames.push(`k-icon--size-${size}`);
 
+    const color = this.props.color || DefaultColor;
+    classNames.push(`k-icon--color-${color}`);
```
magicznyleszek, line 42: Now I'm thinking it might be more elegant to add `k-icon--color-${color}` if `props.color` was provided, and add nothing if not provided. This way we don't really need the `inherit` default :)

jsapp/js/components/common/icon.stories.tsx
```diff
@@ -18,7 +18,7 @@ export const AllIcons = () => (
   <div style={{display: 'flex', flexWrap: 'wrap'}}>
     {(Object.keys(IconNames) as Array<keyof typeof IconNames>).map((icon) => (
       <div style={{flex: '1 1 200px'}}>
-        <Icon name={icon} size='l' />
+        <Icon name={icon} size='l' color='inherit' />
```
magicznyleszek, line 21: I'm not sure how to do a dropdown of available colors with option to select nothing in storybook, but we should do it here :)

magicznyleszek (changes requested):

jsapp/js/components/common/icon.tsx
srartese: I was curious if this was the way we wanted to go about it because there wasn't a good way to just assign variables. I will update

PR #4230, opened 2023-01-25
Toolchain: Better out-of-the-box experience in VS Code

p2edwards:
The goal of this PR is to enable a good out-of-the-box experience in at least one text editor, supporting all of the tools we currently use:

- [x] EditorConfig 
- [x] Black (python formatter)
- [x] ESLint
- [x] Prettier (ts/js formatter)
- [x] Stylelint
- [x] Coffeelint
- [x] TypeScript language server not crashing :)

This branch contains work for **VS Code**. The `kpi.code-workspace` will suggest all needed extensions when opened. Everything ought to work from a fresh install without additional config needed.

## Description

Add a `kpi.code-workspace` file that developers can use to configure VS Code with recommended extensions, including code linters and autoformatters.

## 📄 Usage

Launch the `kpi.code-workspace` with VS Code. Install the Workspace Recommended extensions when prompted. 

The settings in `kpi.code-workspace` should make the extensions work without any additional configuration in VS Code. To run all the linters and formatters, you'll need to install npm packages.

- ✨ Works when installed
    - **EditorConfig**
    - **Black**
    - **Coffeelint**
- 📦 Relies on local `npm install`
    - **ESLint** 
    - **Prettier** 
    - **Stylelint**
- 🐍 Relies on virtualenv in `.venv` for best results
    - **Python (Microsoft)**

To turn on Format and Auto Fix on Save, use the Formatting toggle ✔, bottom right. 

To view output from ESLint, Stylelint, Coffeelint, and the Python and TypeScript/JavaScript error checkers, see the "Problems" tab.

PR #4387, opened 2023-03-24
Toolchain: Warn about undefined CSS Module classes in TypeScript

p2edwards:
## Description

Show a warning in your code editor if you use an undefined CSS module class in a TypeScript file.

## Details

To help catch issues like [the wrong class name being used](https://github.com/kobotoolbox/kpi/pull/4360/commits/5851469a15ee047dbfe8cdad31b303ac55217ba5).

One caveat: It's a TypeScript Language Server plugin, so it'll work in editors but not with`tsc`, `webpack`, or anything we run in CI.

### Notes for reviewer

There used to be an ESLint plugin that did this, but it's currently unmaintained, and it didn't work when I tried it.

There's some talk in the `typescript-plugin-css-modules` repo of [adding an ESLint plugin](https://github.com/typescript-eslint/typescript-eslint/issues/3645), so if that happens we could perhaps use it.

To test this PR, `npm install`, then open the [`kpi.code-workspace`](https://github.com/kobotoolbox/kpi/pull/4230) in VS Code. See what happens when you modify something like `className={styles.root}` in a TypeScript file.

Screen Shot 2023-03-24 at 1 19 11 PM

Further config might be needed to get this functionality in other editors (Sublime, coc.nvim).

---

Changes in this PR:

- `node_modules` - Install [`typescript-plugin-css-modules`](https://github.com/mrmckeb/typescript-plugin-css-modules) - the TypeScript server plugin
- `tsconfig.json` - Set `typescript-plugin-css-modules` as a plugin
- `kpi.code-workspace`: Configure VS Code to use the TypeScript version in node_modules, which will use the plugin in our `tsconfig.json`.
- While we're here, updated our node_modules TypeScript to 5.0.2. This wasn't strictly necessary, but since TypeScript doesn't follow semver, 5.0.0 wasn't a major breaking change, and I figured it's a good time to update.


p2edwards:
According to Leszek, this works in Sublime if you have these plugins:

- `LSP`
- `LSP-typescript`
- `LSP-css`

p2edwards:
@magicznyleszek Thanks for merging!

PR #4388, opened 2023-03-27
Fix comment in .browserslistrc

jnm:
It still included IE11, contradicting the actual content of the file :blush:

## Description

It contained an outdated URL


LMNTL (approved):
Tested the link, looks good!

Issue #4313, opened 2023-03-06
Small UI tweaks to Projects Table


magicznyleszek:
#### Done in #4408 - Project Table UI/UX Adjustments

 - [x] Adjust Project Name hover link styles 
 - [x] Make all column contents left-aligned
 - [x] Click anywhere in the Project Name cell to navigate to that project. Don't navigate when clicking other cells.
      - Clicking on non-project name cells could also toggle the selection for that row, acting like the checkbox. This would making quick actions easier to access. We could try it.
        - Another idea: Clicking on the row "solo's" the clicked row, to make it easier to access the quick actions without "un-clicking" the previous selected checkbox
      - Keep gray row background on hover as is - helps with legibility on wide tables and makes the UI feel more like it's "listening" to user input
- [x] Reposition dropdown triangle to the left side of the dropdown label (field name), keeping the "sorting indictor" on the right, and giving the sorting indicator a different color.
    - [x] Color the "sorting indicator" teal or alt-blue.
    - [x] Align the label (field name) text so its left edge ideally lines up with the cell text in the same column
- [x] Make whole header cell clickable to trigger dropdown
    
#### Done in #4246 - Allow manual column resizing

- [x] Resizing tweaks
    - [x] Ideally keep resizing indicators visible during drag even when the cursor leaves the area
    - [x] Darken resizer indicators (did this during the call)
    - [x] Remove left indicator for leftmost column as suggested by Leszek's PR review
      - [x] Need to re-establish this again since the checkbox selection merge changes what selector should be used

#### Done elsewhere

- [x] Adjust tooltip styles to avoid cropping -- [internal discussion](https://chat.kobotoolbox.org/#narrow/stream/4-Kobo-Dev/topic/My.20Projects.20redo/near/221758) -- fixed [here](https://github.com/kobotoolbox/kpi/commit/ec542a167e26d58fe6f9ecae9e4c4d93edc7ffbb)

#### Not done yet

- [ ] related Searchbox behavior tweaks
 
Reference discussions (internal):

- [table](https://chat.kobotoolbox.org/#narrow/stream/7-UX.2FUI.2FProduct/topic/Frontend.20.2F.20design.20weekly.20checkins/near/207998)
- [searchbox](https://chat.kobotoolbox.org/#narrow/stream/7-UX.2FUI.2FProduct/topic/Frontend.20.2F.20design.20weekly.20checkins/near/211103)

p2edwards:
[...]

PR #4408, opened 2023-04-10
Project Table UI/UX adjustments

p2edwards:
## Description

Improve usability of Projects Table. Make multiple selection easier, and make dropdown behavior more intuitive.

## Details

Usability improvements from the list in #4313

- Make 'Project name' look more like a link during hover
- Clicking anywhere in the row, except 'Project name' cell, now toggles the checkbox.
- Rearrange dropdown indicators and recolor sorting indicators
- Make whole header cell clickable to trigger dropdown
- Left-align every column for consistency

Technical:

- Some refactoring around checkbox styles
- Slight performance improvement on column resizing

## Related issues

Part of #4313


jsapp/js/projects/projectsTable/projectsTableRow.module.scss
```diff
@@ -119,19 +108,82 @@ $projects-table-hover-bg: colors.$kobo-gray-98;
...
+      &::before {
+        left: 0 - sizes.$x32;
+        width: sizes.$x32;
+      }
+      // uncrop element
+      overflow: visible;
+      contain: inline-size style;
+
+      :global input.checkbox__input { // general appearance tweaks
+        &::after {border-radius: 1px;} // better-looking inner corners
```
magicznyleszek, line 141: This actually makes it look worse (at least on Firefox and Chrome on retina screen), see scaled up screenshot:
Screenshot 2023-04-19 at 22 54 17

And withouth `border-radius` the issue is still present, but it's a tiny bit smaller:
Screenshot 2023-04-19 at 23 01 57

Is it different on your machine?

jsapp/js/projects/projectsTable/projectsTableRow.module.scss
```diff
@@ -119,19 +108,82 @@ $projects-table-hover-bg: colors.$kobo-gray-98;
...
+        &:focus-visible { // add keyboard focus styles
+          outline: 3px solid colors.$kobo-alt-blue;
```
magicznyleszek, line 143: Could we add this in `/jsapp/js/components/common/checkbox-and-radio.scss` instead? I think it would be beneficial to all checkboxes :)

jsapp/js/projects/projectsTable/projectsTableRow.module.scss
```diff
@@ -119,19 +108,82 @@ $projects-table-hover-bg: colors.$kobo-gray-98;
...
     }
+
+    // Various hover rules to tell the user what will happen:
+    // - click project cell  ->  navigate to project
+    // - click other cell    ->  toggle the checkbox for project
+
+    // Hovering any cell  ->  Outline the checkbox
+    &[data-field='checkbox'] {
+      :global .checkbox {
+        border-radius: sizes.$x2;
```
magicznyleszek, line 163: I think instead of adding this alien `border-radius` to `.checkbox` it would be better to target `.checkbox__input` :)

jsapp/js/projects/projectsTable/projectsTableRow.module.scss
```diff
@@ -119,19 +108,82 @@ $projects-table-hover-bg: colors.$kobo-gray-98;
...
+        outline: sizes.$x1 solid colors.$kobo-alt-blue;
+      }
+
+      &:hover {
+        // Do an even bigger outline if you're actually near the checkbox
+        :global .checkbox {
```
magicznyleszek, line 169: `.checkbox__input` here too :)

magicznyleszek (changes requested):

jsapp/js/projects/projectsTable/projectsTableRow.module.scss
p2edwards: Yep, it was different for me, I could take screenshots. It might depend on the zoom level, too.

magicznyleszek: I took screenshots at 100% zoom and then resized them (quickly in Preview.app so not pixel perfect).

But I think that the issue might be actually in how the Checkbox component was built (by me :P). Perhaps there is a way to set the rounded border in such way that it overflow:hidden on the blue background…

If the difference on your non-retina is bigger than on my retina, let's keep this 1px :)

p2edwards: Resolved by https://github.com/kobotoolbox/kpi/pull/4442

jsapp/js/projects/projectsTable/projectsTableRow.module.scss
p2edwards: Sure. After your radio/checkbox update, that's `checkbox.scss` and `radio.scss`.

jsapp/js/projects/projectsTable/projectsTableHeader.module.scss
```diff
@@ -152,6 +168,7 @@ $resizer-width: sizes.$x16 + $line-width;
   opacity: 1;
 }
 // Always hide leftmost width indicator
-.columnRoot:first-child::before {
+// nth-child(2) because of checkbox column
+.columnRoot:nth-child(2)::before {
   display: none;
 }
```
magicznyleszek, line 174: The first column header cell (empty one, above checkboxes) has pointer cursor visible when mouseover-ing. It's not clickable, so we should use the default cursor.

magicznyleszek (changes requested):

PR #4409, opened 2023-04-11
Exclude fields from My Projects route

### LINKED ISSUE #4378: Hide some fields from My Projects route
> I think these three fields should be removed from both the filter and the customized fields modals because we don't expose that data for regular projects, only for the special "project views"
> - Owner full name
> - Owner email
> - Owner organization

magicznyleszek:
## Description

Hide "Owner full name", "Owner email", and "Owner organization" columns for My Projects route. These pieces of information are not available for endpoint that is handling this route (in contrast to a custom Project View endpoint).

## Related issues

Fixes #4378


LMNTL (approved):
Works excellently in my testing.

PR #4417, opened 2023-04-17
Always recreate fonts when syncing npm dependencies in dev mode

noliveleger:
## Description

Fix error when `/jsapp/fonts` folder is missing while building the front-end assets in dev mode.


noliveleger:
@p2edwards this PR is little bit different from what you suggested.
It displays missing dependencies but nothing else.

p2edwards (approved):

Issue #4346, opened 2023-03-14
Side settings: Render side elements based on settings


magicznyleszek:
Ideally all elements should be separate components. This will ease out with rendering them based on settings. Make sure to have some default settings too

Issue #4345, opened 2023-03-14
Side settings: Modal with list of selectable elements


magicznyleszek:
Modal
- List of toggle switches
  - we display all transcripts, translations and some built in things
  - not reorderable yet, just a list with fixed order
- Reset and Apply buttons
- Keep the selection stable while in this page.
  - Remember most recent configuration as long as Single Processing Route is open

PR #4423, opened 2023-04-20
Add a display settings modal to the single processing view

duvld:
Used a `Map<LanguageCode | StaticDisplays, boolean>` to store the display and it's status. Decided on a `Map` because it would be useful to know:
1. Map as a whole would list what is to be displayed in the modal
2. The boolean would show if the switch is on or not

This is because certain tabs show certain displays (except for Analysis). E.g., I did not show transcripts and translations for the Transcript and Translations tabs respectfully because it is already visible in the content section. Navigating through the tabs populate and clear the `Map` depending on what needs to be shown. Clicking the available toggles in each view changes the sidebar based on the `boolean` value.

Downsides could be legibility and moving the displays around in the future.

## Description

Add a modal to allow users to reorganize the displays (audio file, submission data, translations, transcriptions) in the single processing view. User can now:
- turn on or off displays
- display multiple translations at once

## Related issues
#4345 #4346


jsapp/js/components/processing/singleProcessingDisplaySettings.tsx
```diff
@@ -0,0 +1,98 @@
...
+import {AsyncLanguageDisplayLabel} from '../languages/languagesUtils';
+
+interface SingleProcessingDisplaySettingsState {
+  isModalOpen: boolean;
+}
+
+export default function SingleProcessingDisplaySettings() {
+  const [store] = useState(() => singleProcessingStore);
+
+  const [modal, setModal] = useState<SingleProcessingDisplaySettingsState>({
```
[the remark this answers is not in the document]
magicznyleszek: Then a bit below you can remove `toggleModal` altogether and update the functions to simpler `() => setIsModalOpen(true)` or `() => setIsModalOpen(false)` (like in comments below)

jsapp/js/components/processing/singleProcessingPreview.module.scss
```diff
@@ -3,7 +3,7 @@
 @use './singleProcessingVariables';
 
 .root {
-  margin-top: singleProcessingVariables.$s-single-processing-tabs;
+  margin-top: sizes.$x32;
```
magicznyleszek, line 6: I see that this needs some work. Let's try to remember to somehow use that variable here

jsapp/js/components/processing/singleProcessingSidebar.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  return (
+    <div className={style.root}>
+      <SingleProcessingDisplaySettings />
+
+      {Array.from(translations).map((translation) => {
+        if (
+          displays.get(translation.languageCode)
+        ) {
+          return <SingleProcessingTranslation singleTransx={translation} />;
```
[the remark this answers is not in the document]
magicznyleszek: But if this component is for displaying translation or transcript, rename it to something like `<TransxDisplay>`

jsapp/js/components/processing/singleProcessingStore.ts
```diff
@@ -111,6 +116,12 @@ class SingleProcessingStore extends Reflux.Store {
   private isSubmissionLoaded = false;
   private isProcessingDataLoaded = false;
 
+  private displays = new Map<LanguageCode | StaticDisplays, boolean>([
```
magicznyleszek, line 119: Please create a `/**` comment to describe the contents of this, especially what the boolean stands for

jsapp/js/components/processing/singleProcessingStore.ts
```diff
@@ -936,6 +983,63 @@ class SingleProcessingStore extends Reflux.Store {
...
+      this.populuateTranslationDisplays();
+      // If transcript display is on (coming from translations) then leave it on.
+      this.displays.set(
+        StaticDisplays.Transcript,
+        this.displays.get(StaticDisplays.Transcript) || false
+      );
+    }
+  }
+
+  setStaticDisplay(display: StaticDisplays) {
```
magicznyleszek, line 1011: I think it would be cleaner if instead of display name, we would also give a boolean `isEnabled` here, so `setStaticDisplay(display: StaticDisplays, isEnabled: boolean) {`. The `<ToggleSwitch>`es that uses this function would use their boolean callback value :)

jsapp/js/components/processing/singleProcessingStore.ts
```diff
@@ -936,6 +983,63 @@ class SingleProcessingStore extends Reflux.Store {
...
+          !this.displays.get(StaticDisplays.Data)
+        );
+        break;
+      case StaticDisplays.Transcript:
+        this.displays.set(
+          StaticDisplays.Transcript,
+          !this.displays.get(StaticDisplays.Transcript)
+        );
+        break;
+    }
```
magicznyleszek, line 1031: Do we need the switch case statement here? Would TS allow simple `this.displays.set(display, isEnabled)`?

jsapp/js/components/processing/singleProcessingStore.ts
```diff
@@ -936,6 +983,63 @@ class SingleProcessingStore extends Reflux.Store {
...
+
+    this.trigger(this.displays);
+  }
+
+  setTranslationDisplay(display: LanguageCode) {
```
magicznyleszek, line 1036: Could we try making this a single function for both static and translation displays? It might require some TS wizardry, but it should be doable :)

jsapp/js/components/processing/singleProcessingSubmissionData.tsx
```diff
@@ -0,0 +1,58 @@
...
+  // If submission data is not ready yet, just don't render the list.
+  if (!submissionData) {
+    return null;
+  }
+
+  // If there is a source, we don't want to display these submission details,
+  // as we want the most space possible for the source text.
+  if (singleProcessingStore.getSourceData() !== undefined) {
+    return null;
+  }
```
magicznyleszek, line 33: With the DisplaySettings, do we still need this? User can set what they display themselves now

jsapp/js/components/processing/singleProcessingSubmissionDetails.tsx
```diff
@@ -154,6 +154,6 @@ export default class SingleProcessingSubmissionDetails extends React.Component<S
   }
 
   render() {
-    return [this.renderMedia(), this.renderDataList()];
+    return [this.renderMedia()];
```
magicznyleszek, line 157: ~~I think this file needs some cleanup :) AFAIU `renderDataList` can be removed? and this will only render media?~~ Do we delete this file?

jsapp/js/components/processing/singleProcessingTranslation.tsx
```diff
@@ -0,0 +1,63 @@
...
+import type {Transx} from 'js/components/processing/singleProcessingStore';
+import styles from './singleProcessingPreview.module.scss';
+import bodyStyles from './processingBody.module.scss';
+import {AsyncLanguageDisplayLabel} from 'js/components/languages/languagesUtils';
+
+interface SingleProcessingTranslationProps {
+  singleTransx: Transx;
+}
+
+export default function SingleProcessingTranslation(props: SingleProcessingTranslationProps) {
```
magicznyleszek, line 13: Following the comments above: this should be `TransxDisplay`?

magicznyleszek (changes requested):

jsapp/js/components/processing/singleProcessingDisplaySettings.tsx
```diff
@@ -0,0 +1,98 @@
...
+                      checked={entry[1]}
+                    />
+                  </li>
+                );
+              } else {
+                return null;
+              }
+            })}
+          </ul>
+        </KoboModalContent>
```
magicznyleszek, line 94: One thing I've noticed - Figma mockups have "Reset" and "Apply selection" buttons. So toggle switches in modal act like a draft that applies to the store only after "Apply selection" is clicked. Plus also we need to have some default displays selection for the "Reset" button

magicznyleszek (changes requested):

jsapp/js/components/processing/processingBody.module.scss
```diff
@@ -107,6 +107,8 @@
   width: 100%;
   overflow-y: auto;
   flex: 1;
+  // TODO: have a 'full' option that turns off other displays
+  max-height: sizes.$x200;
```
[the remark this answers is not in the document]
duvld: Didn't realize this affected the left :P I think we should still have some max height on the Transx displays on the sidebar though. If there are 2 or more bigger Transx displays it would push any other display off.

I think we would need:
1. A max height for the Transx displays
2. A limit to the amount displayed at once (that fits with the max height)
3. A way to "full screen" the display

Or we could overflow the entire sidebar for MVP?

jsapp/js/components/processing/sidebar/sidebarDisplaySettings.module.scss
```diff
@@ -0,0 +1,46 @@
...
+  }
+}
+
+.options {
+  height: 100%;
+  padding-bottom: sizes.$x24;
+  overflow-y: scroll;
+}
+
+footer {
```
magicznyleszek, line 35: If a CSS module file has tag selector, it is being applied globally to everything (to all `<footer>`s). CSS modules can only do localized styling with class selectors.

jsapp/js/components/processing/sidebar/sidebarDisplaySettings.module.scss
```diff
@@ -0,0 +1,46 @@
...
+  padding-top: sizes.$x12;
+  padding-bottom: sizes.$x12;
+  justify-content: center !important;
+  // Some basic overflow styling just in case the list is very long
+  position: fixed;
+  bottom: 0;
+  right: 0;
+  left: 0;
+  background: colors.$kobo-white;
+  border-radius: sizes.$x6;
```
magicznyleszek, line 45: All of these styles are actually not needed. The `kobo-modal__footer` element is always at the bottom, and the `kobo-modal__content` never overflows it.

jsapp/js/components/processing/sidebar/sidebarDisplaySettings.module.scss
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  span::after {
+    content: ')';
+  }
+}
+
+.options {
+  height: 100%;
+  padding-bottom: sizes.$x24;
```
magicznyleszek, line 31: This is not needed. `kobo-modal__content` has bottom padding already.

jsapp/js/components/processing/sidebar/sidebarDisplaySettings.module.scss
```diff
@@ -0,0 +1,46 @@
...
+  padding-top: sizes.$x18;
+  padding-bottom: sizes.$x10;
+}
+
+.description {
+  padding-bottom: sizes.$x30;
+}
+
+.display {
+  padding-bottom: sizes.$x40;
```
magicznyleszek, line 16: Last display shouldn't get this padding, you can just add `:not(:last-child)` to the selector

jsapp/js/components/processing/sidebar/sidebarDisplaySettings.module.scss
```diff
@@ -0,0 +1,46 @@
...
+    content: '(';
+  }
+
+  span::after {
+    content: ')';
+  }
+}
+
+.options {
+  height: 100%;
```
magicznyleszek, line 30: This `height` doesn't do anything here

jsapp/js/components/processing/sidebar/sidebarDisplaySettings.module.scss
```diff
@@ -0,0 +1,46 @@
...
+  span::before {
+    content: '(';
+  }
+
+  span::after {
+    content: ')';
+  }
+}
+
+.options {
```
magicznyleszek, line 29: Unfortunately it needs `margin: 0 !important` to override the styles from `kobo-modal` - it has some styles for `> ul` and `> p` to work with text content (turns out it was a bad idea :P)

jsapp/js/components/processing/sidebar/sidebarDisplaySettings.module.scss
```diff
@@ -0,0 +1,46 @@
...
+
+.root {
+  display: flex;
+  justify-content: right;
+  padding-top: sizes.$x18;
+  padding-bottom: sizes.$x10;
+}
+
+.description {
+  padding-bottom: sizes.$x30;
```
magicznyleszek, line 12: Looking at the Figma file, it looks like `.description` should have something more like `40px` and `.display` should have `35px`. Let's use `sizes.$x36` that we already have for both :)

jsapp/js/components/processing/sidebar/sidebarDisplaySettings.module.scss
magicznyleszek: I think we should remove this whole `footer {…}` from here. Please in `koboModal.scss` file add `.kobo-modal__footer--centered` that simply has `justify-content: center;`. And then in tsx file use `<KoboModalFooter isCentered>`. And finally edit `koboModalFooter.tsx` and add the `isCentered` prop and `<bem.KoboModal__footer m={props.isCentered ? 'isCentered' : ''}>`.

jsapp/js/components/processing/sidebar/sidebarDisplaySettings.module.scss
```diff
@@ -0,0 +1,46 @@
...
+}
+
+.display {
+  padding-bottom: sizes.$x40;
+}
+
+.hasTranslation {
```
magicznyleszek, line 19: Maybe `.wrapWithParens` or something similar would make more sense?

jsapp/js/components/processing/sidebar/sidebarDisplaySettings.tsx
```diff
@@ -0,0 +1,130 @@
...
+      return <strong>{t('Submission data')}</strong>;
+    } else {
+      return <strong>{t('Original file (Audio)')}</strong>;
+    }
+  }
+
+  return (
+    <div className={styles.root}>
+      <Button
+        classNames={[styles.displaySettings]}
```
magicznyleszek, line 43: AFAIK you don't have any styles for `.displaySettings`

jsapp/js/components/processing/sidebar/sidebarDisplaySettings.tsx
```diff
@@ -0,0 +1,130 @@
...
+        color='storm'
+        startIcon='settings'
+        onClick={() => setIsModalOpen(true)}
+      />
+      <KoboModal
+        isOpen={isModalOpen}
+        onRequestClose={() => setIsModalOpen(false)}
+        size='medium'
+      >
+        <KoboModalHeader iconColor='storm'>
```
magicznyleszek, line 56: `iconColor` is unnecessary, as you don't have any icon here

jsapp/js/components/processing/sidebar/sidebarSubmissionDetails.module.scss
```diff
@@ -1,6 +1,6 @@
 @use '~kobo-common/src/styles/colors';
 @use 'scss/sizes';
-@use './singleProcessingVariables';
+@use '../singleProcessingVariables';
 
 .mediaWrapper {
```
magicznyleszek, line 5: Can we split this file into `sidebarSubmissionData.module.scss` and `sidebarSubmissionMedia.module.scss`? AFAIK they don't share any of the styles

jsapp/js/components/processing/sidebar/sidebar.tsx
```diff
@@ -0,0 +1,51 @@
+import React, {useState} from 'react';
+import singleProcessingStore, {
+  StaticDisplays,
+} from 'js/components/processing/singleProcessingStore';
+import TransxDisplay from 'js/components/processing/transxDisplay';
+import SidebarDisplaySettings from 'js/components/processing/sidebar/sidebarDisplaySettings';
+import type {AssetResponse} from 'jsapp/js/dataInterface';
+import SingleProcessingSubmissionData from 'js/components/processing/sidebar/sidebarSubmissionData';
+import SingleProcessingSubmissionMedia from 'js/components/processing/sidebar/sidebarSubmissionMedia';
```
magicznyleszek, line 9: Both of these imports have different internal names, let's use them to avoid confusion. I.e. there is no `SingleProcessingSubmissionData` to be found, but there is a `SidebarSubmissionData`

jsapp/js/components/processing/transxDisplay.tsx
```diff
@@ -0,0 +1,59 @@
...
+import {AsyncLanguageDisplayLabel} from 'js/components/languages/languagesUtils';
+
+import styles from './transxDisplay.module.scss';
+
+interface TransxDisplayProps {
+  transx: Transx;
+}
+
+export default function TransxDisplay(props: TransxDisplayProps) {
+  function renderLanguageAndDate() {
```
magicznyleszek, line 14: I've learned recently about [useCallback hook](https://react.dev/reference/react/useCallback) that is useful for functions defined inside a functional component. I don't think is crucial here, but wanted to make sure you know about it. It basically wraps `renderLanguageAndDate`, so it's not re-defined with every render (as it currenly happens)

jsapp/js/components/processing/transxDisplay.module.scss
```diff
@@ -3,7 +3,9 @@
 @use './singleProcessingVariables';
 
 .root {
-  margin-top: singleProcessingVariables.$s-single-processing-tabs;
+  // TODO: Use the $s-single-processing-tabs variable here somehow.
+  // Right now 60px is too big if there are more than 1 translations.
+  margin-top: sizes.$x32;
```
magicznyleszek, line 8: After you introduce `gap` in `.displays` (in `sidebar.module.scss`), this TODO would be resolved, as margin will no longer be needed here

jsapp/js/components/processing/sidebar/sidebarSubmissionData.tsx
```diff
@@ -0,0 +1,52 @@
+import React, {useState} from 'react';
```
magicznyleszek, line 1: This file needs prettier to be run over it :) Make sure you re-run prettier over all the files you touch just to be sure

magicznyleszek (changes requested):

jsapp/js/components/processing/sidebar/sidebar.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+import styles from './sidebar.module.scss';
+
+interface SidebarProps {
+  asset: AssetResponse;
+}
+
+export default function Sidebar(props: SidebarProps) {
```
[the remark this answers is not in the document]
duvld: Todo still

jsapp/js/components/processing/sidebar/sidebar.module.scss
```diff
@@ -0,0 +1,16 @@
...
+
+.root {
+  height: 100%;
+}
+
+.displays {
+  padding-bottom: singleProcessingVariables.$s-single-processing-tabs;
+  display: flex;
+  flex-direction: column;
+  height: 100%;
```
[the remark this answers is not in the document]
duvld: Adding `overflow-y: auto;` fixes the described issue, but `max-height: 100%` produces this:
image

Not sure why, but keeping `height: 100%` seems to work as expected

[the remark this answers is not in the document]
duvld: I'm guessing CSS has some cracks somewhere that interprupt `max-height: 100%` down to the text boxes

jsapp/js/components/processing/sidebar/sidebar.module.scss
```diff
@@ -0,0 +1,16 @@
+@use '../singleProcessingVariables';
+
+.root {
+  height: 100%;
+}
+
+.displays {
+  padding-bottom: singleProcessingVariables.$s-single-processing-tabs;
+  display: flex;
```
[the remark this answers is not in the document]
duvld: Todo for later

jsapp/js/components/processing/sidebar/sidebar.module.scss
```diff
@@ -0,0 +1,16 @@
...
+  flex-direction: column;
+  height: 100%;
+}
+
+.displays > :first-child {
+  margin-top: 0 !important;
+}
```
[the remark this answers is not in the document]
duvld: I think I used to hack because depending on if it shows translations or not, the first child of `.displays` would change, and depending on which display is first, the `margin-top` would be needed again. This seemed to be less complicated solution

jsapp/js/components/processing/sidebar/sidebarDisplaySettings.tsx
```diff
@@ -0,0 +1,130 @@
...
+              setIsModalOpen(false);
+            }}
+          />
+          <Button
+            label={<strong>{t('Apply selection')}</strong>}
+            type='full'
+            color='light-blue'
+            size='m'
+            onClick={() => {
+              store.applyDisplay();
```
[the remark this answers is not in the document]
duvld: todo for later

jsapp/js/components/processing/sidebar/sidebarDisplaySettings.module.scss
duvld: Note for potential cleanup: this needed `justify-content: center !important`

jsapp/js/components/processing/sidebar/sidebarSubmissionDetails.module.scss
duvld: todo for later

jsapp/js/components/processing/transxDisplay.tsx
duvld: Immediately asked if using this without dependencies makes sense and found this:
https://stackoverflow.com/questions/55026139/whats-the-difference-between-usecallback-with-an-empty-array-as-inputs-and-u

makes sense in hindsight but just wanted to note this

jsapp/js/components/processing/transxDisplay.module.scss
duvld: Todo for later

magicznyleszek (approved):

PR #4424, opened 2023-04-24
Billing page multisite compatibility / Change subscription route

LMNTL:
## Checklist

3. [X] Ensure the tests pass

## Description

Changes the ID field of Organizations to work with dj-stripe webhooks on multiple sites.
Adds an API route to change existing Stripe subscriptions to a different price.

## Notes
This PR keeps the `uid` field on organizations intact, except for changing its prefix. A future PR could condense that so we use the `id` field everywhere, since we only need one UUID to pass around, but that's not necessary at this time.


bufke:
> A future PR could condense that so we use the id field everywhere

I think we should just go ahead and do that - there's no reason to have two UUID's. If this breaks existing data, it's better to deal with it sooner while we have few.

kobo/apps/stripe/views.py
```diff
@@ -46,6 +44,96 @@ def get_queryset(self):
         ).prefetch_related('payment_intent')
 
 
+# Change an existing subscription to a new price
```
bufke, line 47: I think we could add more documentation, either in the class docstring or on some static methods (or both). This class does a few things but it's not clear exactly what at a glance.

kobo/apps/organizations/migrations/0003_alter_organization_uid.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 3.2.15 on 2023-04-20 15:14
+
+from django.db import migrations
+import kpi.fields.kpi_uid
+
+
+class Migration(migrations.Migration):
```
bufke, line 7: These migrations could be combined.

bufke (changes requested):

hub/migrations/0008_auto_20230308_0137.py
```diff
@@ -1,23 +0,0 @@
-# Generated by Django 3.2.15 on 2023-03-08 01:37
```
bufke, line 1: Why was this file removed?

bufke (approved):
LGTM, see my one comment but it's minor. I ran it locally and created an organization to test.

Issue #4422, opened 2023-04-19
'your plan' shows on both monthly and annual plans after upgrading


srartese:
## Description

The "your plan" message should only show when you're viewing the relevant monthly/annual toggle

## Steps to Reproduce

1. Go to https://kf.billing.kbtdev.org/#/account/plan or checkout feature/billing and navigate to #/account/plan.
2. Make sure the account being used is with no plan (community plan)
3. Click on the 'Upgrade' button for a plan
4. Checkout with payment (test card here https://stripe.com/docs/testing)
5. Navigate back to the Plans page
6. Toggle between month and year 

## Expected behavior 
When page redirects the banner should only show on the monthly/annual plan you upgraded to when you toggle between the two filters.

## Actual behavior 
When page redirects the banner is showing on both the monthly and annual plan for the upgraded name (IE Professional) when toggling between the two filters.

## Additional details

The banner is being applied based on the name of the plan and not the interval. If they are on the community plan, show annual by default.

PR #4437, opened 2023-05-09
Bugfix/4422 your plan show only on annual or monthly

srartese:
## Checklist

3. [x] Ensure the tests pass

## Description

When page redirects the banner now only shows on the monthly/annual plan you upgraded to when you toggle between the two filters. It will load the filter to show based on what you are subscribed to. If there is no plan subscribed then show annual community plan by default. When subscribed to the pro plan show on the opposite filter a 'change plan' button so the user has a call to action when their plan is not visible. 

## Related issues
#4422


bufke (approved):
Approved once CI issues are resolved.

bufke (changes requested):
We need to make it more obvious what the current plan is.

jsapp/js/account/plan.component.tsx
```diff
@@ -92,21 +92,26 @@ export default function Plan() {
...
+      state.subscribedProduct !== null &&
+      state.subscribedProduct.length > 0
+    ) {
+      const subscribedFilter =
+        state.subscribedProduct[0].items[0].price.human_readable_price.split(
+          '/'
+        )[1];
+      dispatch({type: subscribedFilter});
+    }
+  }, [state.subscribedProduct]);
```
bufke, line 113: This could be a computed function, right?

bufke (changes requested):

jsapp/js/account/plan.component.tsx
LMNTL: Instead of using `price.human_readable_price` to figure out whether a subscription is monthly or yearly, you should probably use [`price.recurring.interval`](https://stripe.com/docs/api/prices/object#price_object-recurring-interval_count). Otherwise, this function would need to be rewritten if we changed how we display prices (or if `dj-stripe` changes how they compute `human_readable_price`.)

srartese: Synced with Jessica on it, after some research dj-stripe does not offer us such an easy bridge currently but we are in the works of a new approach.

LMNTL (changes requested):

jsapp/js/account/plan.component.tsx
```diff
@@ -92,21 +92,21 @@ export default function Plan() {
   const [shouldRevalidate, setShouldRevalidate] = useState(false);
   const [searchParams] = useSearchParams();
   const didMount = useRef(false);
-  const hasActiveSubscription = useMemo(() => {
```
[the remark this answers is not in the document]
srartese: Ahh I forgot to add this back in, thanks doing that!

jsapp/js/account/plan.component.tsx
```diff
@@ -212,17 +226,32 @@ export default function Plan() {
 
   const isSubscribedProduct = useCallback(
     (product: Price) => {
-      if (hasActiveSubscription !== undefined) {
-        if (!product.prices.unit_amount && !hasActiveSubscription) {
+      if (
+        !product.prices.unit_amount &&
+        state.intervalFilter === 'year' &&
```
LMNTL, line 231: Getting better; now the 'your plan' banner shows up for the Community plan when it's annual, but not when showing monthly plans. I'm guessing it's related to this line.

LMNTL (changes requested):

jsapp/js/account/plan.component.tsx
srartese: Yup! I just found this while testing and pushed a change to fix it

LMNTL: Unfortunately, it's still only showing for the annual plan. I'd suggest reverting the last commit and just removing line 231 (`state.intervalFilter === 'year' &&`).

LMNTL (changes requested):

LMNTL (approved):

PR #4443, opened 2023-05-18
Improve long-running trash-bin tasks duration

noliveleger:
## Description

Make deletion tasks run shortly and increase the limit of time they can run


jnm (approved):
I'm worried about the long execution time crowding out other jobs in the same queue. If the task is able to "nibble away" at the projects each time it runs, and its progress isn't rolled back when it times out, then I'd favor a lower timeout and larger retry count

PR #4449, opened 2023-05-24
Add a `status` filter in Django Admin to filter objects in trash

noliveleger:
## Description

It narrows down the list of objects matching the status filter.


jnm (approved):
Tested locally; looks good :+1:

PR #4454, opened 2023-05-29
4394 nlp additional columns for grouped question appear after the entire group

### LINKED ISSUE #4394: [NLP] Additional columns for grouped question appear after the entire group
Given the survey structure:
1. Begin group
2. Audio question
3. Text question
4. End group

…the data table view should have columns in this order once a transcript is added:
1. Audio question
2. Transcript of audio question
3. Text question

However, it currently uses this order instead:
1. Audio question
2. Text question
3. Transcript of audio question

`@srartese`, I think this is an issue in the front-end code, but I haven't investigated extensively.

srartese:
## Checklist

3. [x] Ensure the tests pass

## Description
When viewing the data table with submissions if the survey has groups the transcription will appear in the wrong order and after other questions when it should be inserted directly after the audio question it is related to. 

## Related issues

Fixes #4394

### DIFF
--- a/jsapp/js/components/submissions/table.es6
+++ b/jsapp/js/components/submissions/table.es6
@@ -8,7 +8,9 @@ import bem from 'js/bem';
 import LoadingSpinner from 'js/components/common/loadingSpinner';
 import {stores} from 'js/stores';
 import ReactTable from 'react-table';
-import ValidationStatusDropdown, { SHOW_ALL_OPTION } from 'js/components/submissions/validationStatusDropdown';
+import ValidationStatusDropdown, {
+  SHOW_ALL_OPTION,
+} from 'js/components/submissions/validationStatusDropdown';
 import {DebounceInput} from 'react-debounce-input';
 import {
   PERMISSIONS_CODENAMES,
@@ -63,7 +65,7 @@ import {
   userCan,
   userCanPartially,
   isSubmissionWritable,
-  } from 'js/components/permissions/utils';
+} from 'js/components/permissions/utils';
 
 const DEFAULT_PAGE_SIZE = 30;
 
@@ -71,7 +73,7 @@ const DEFAULT_PAGE_SIZE = 30;
  * @prop {object} asset
  */
 export class DataTable extends React.Component {
-  constructor(props){
+  constructor(props) {
     super(props);
     this.state = {
       isInitialized: false, // for having asset with content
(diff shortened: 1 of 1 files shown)


dorey:
Does [this](https://github.com/kobotoolbox/kpi/pull/4454/files#diff-b2e814da86e17e70a841f42ca7576dc647ca91115799e0442cd1a1e97a51f200R736) check all the groups or just one group? I.e. if there are nested groups would the same bug show up?

dorey:
Ok, I re-ran the changes made by `prettier` and compared. I'd probably opt for rewriting the git history so that these changes are in their own commit

```
diff --git b/jsapp/js/components/submissions/table.es6 a/jsapp/js/components/submissions/table.es6
index d5228618f..bcc8bfb4d 100644
--- b/jsapp/js/components/submissions/table.es6
+++ a/jsapp/js/components/submissions/table.es6
@@ -730,11 +730,23 @@ export class DataTable extends React.Component {
           // Detect supplemental details column and put it after its source column.
           if (q === undefined && key.startsWith(SUPPLEMENTAL_DETAILS_PROP)) {
             const supplementalColumnSource = key.split('/')[1];
-            const sourceColumn = columnsToRender.find(
-              (columnToRender) =>
-                columnToRender.id === flatPaths[supplementalColumnSource]
-            );
-
+            // Add extra step if grouped
+            const sourceCleaned = supplementalColumnSource
+              .replace(/-/g, '/')
+              .split('/')[1];
+            let sourceColumn;
+
+            if (survey[0].type === 'begin_group') {
+              sourceColumn = columnsToRender.find(
+                (columnToRender) =>
+                  columnToRender.id === flatPaths[sourceCleaned]
+              );
+            } else {
+              sourceColumn = columnsToRender.find(
+                (columnToRender) =>
+                  columnToRender.id === flatPaths[supplementalColumnSource]
+              );
+            }
             if (sourceColumn) {
               // This way if we have a source column with index `2`, we will set
               // the supplemental details column to `2__supplementalDetails/…`
@@ -754,7 +766,6 @@ export class DataTable extends React.Component {
       if (q && q.type) {
         columnIcon = renderQuestionTypeIcon(q.type);
       }
-
       columnsToRender.push({
         Header: () => {
           const columnName = getColumnLabel(
@@ -763,7 +774,6 @@ export class DataTable extends React.Component {
             this.state.showGroupName,
             this.state.translationIndex
           );
-
           const columnHXLTags = getColumnHXLTags(
             this.props.asset.content.survey,
             key
```

dorey:
splitting commit into `formatting` change and code change

PR #4456, opened 2023-06-01
Use the correct labels and ordering for grouped transcript and translation fields

### LINKED ISSUE #4395: [NLP] Grouped transcript and translation columns do not use the question label in the table view
If you make a form with a grouped audio question, like this:

type | name | label
-- | -- | --
audio | I_m_outside_the_group | I'm outside the group
begin_group | group_po7lr63 | I'm a group
audio | I_m_inside_the_group | I'm inside the group!
end_group |   |  

…and then you add a transcript of the grouped audio question, instead of seeing "transcript (en} | I'm inside the group!", i.e. "transcript ([language])" plus the _label_ of the question in the UI, you'll instead see "transcript (en) | group_po7lr63-I_m_inside_the_group", which uses the **name** of the question instead of the label. The label should be shown unless you've switched the UI to show XML names and values, which is not the default.

After glancing at the `-` to `/` substitution in #4454, I wonder if that change wouldn't also fix this issue.

image

The column header for the transcript column uses the group and question names and not the labels.

srartese:
## Checklist

3. [x] Ensure the tests pass

## Description

When viewing the data table with submissions if the survey has groups the column header for the transcript column uses the group and question names and not the labels.

## Related issues

Fixes #4395
Part of #4454

### DIFF
--- a/jsapp/js/components/submissions/table.es6
+++ b/jsapp/js/components/submissions/table.es6
@@ -8,7 +8,9 @@ import bem from 'js/bem';
 import LoadingSpinner from 'js/components/common/loadingSpinner';
 import {stores} from 'js/stores';
 import ReactTable from 'react-table';
-import ValidationStatusDropdown, { SHOW_ALL_OPTION } from 'js/components/submissions/validationStatusDropdown';
+import ValidationStatusDropdown, {
+  SHOW_ALL_OPTION,
+} from 'js/components/submissions/validationStatusDropdown';
 import {DebounceInput} from 'react-debounce-input';
 import {
   PERMISSIONS_CODENAMES,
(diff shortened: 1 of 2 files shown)


jnm:
@srartese thanks for working on this! A few minor pointers about PRs:
1. You can get GitHub to recognize that a PR fixes or closes an issue by using keywords like `fixes #thing`; see https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests
2. Please write PR titles in the imperative, with an emphasis on what the PR does, and don't worry about including issue numbers there. This one could be something like "Use the correct labels and ordering for grouped transcript and translation fields".
3. The description could talk about how the fix affects the data table and maybe (hopefully! :crossed_fingers:) the single submission modal (the thing the eyeball icon opens) as well, which was also broken in the same way at least before this PR: 
image

dorey:
Starting a new branch with `723d44b` out of the git history. Will reference this PR
https://github.com/kobotoolbox/kpi/pull/4454#issuecomment-1583015477

PR #4457, opened 2023-06-02
Drop underscore

### LINKED ISSUE #3652: Replace underscore usage with lodash and drop it
## Description

We use both `lodash` and `underscore` and it's not really logical, as both are very much alike. `lodash` twice as heavy, but is much faster, so let's keep it.

magicznyleszek:
## Description

Inner cleanup. Removes unnecessary `underscore` dependency in favor of granular `lodash` packages.

## Related issues

Fixes #3652


srartese:
The code changes look good and runs on my end without errors. Looks like some files could use some formatting and linting fixes though.

magicznyleszek:
@srartese I will leave the linting/formatting fixes to be done on some other occasion, as this PR would grow to big :)

PR #4482, opened 2023-06-08
Password Validation Configuration

### LINKED ISSUE #4462: Build a way to configure custom requirements in Constance
We need:
- Toggle on/off and configure each and every validator

JacquelineMorrissette:
## Checklist

3. [x] Ensure the tests pass

## Description

Add password validation configuration to Constance  and add variables to the `/environment` endpoint

## Related issues
Closes #4462


kobo/settings/base.py
```diff
@@ -349,6 +349,38 @@
...
+    ),
+    # Build-in password validator
+    'USER_ATTRIBUTE_SIMILARITY_VALIDATOR': (
+        True,
+        "Custom user attribute similarity validation is enabled",
+    ),
+    # Built-in password validator
+    'COMMON_PASSWORD_VALIDATION': (
+        '',
+        'Add to list of common passwords for password validation',
```
magicznyleszek, line 371: Is it easy to extend the built-in Django list with custom words? It is not required for the task (i.e. we are good with the default list), but it would be nice to have if it's not very time consuming

kobo/settings/base.py
```diff
@@ -349,6 +349,38 @@
...
+    ),
+    # Custom password Validator
+    'CUSTOM_CHARACTER_CLASSES_ENABLED': (
+        '^(?=.*[a-z])(?=.*[A-Z])(?=.*\d)(?=.*[@$!%*?&])[A-Za-z\d@$!%*?&]',
```
magicznyleszek, line 375: Are we able to fulfill the requirements with single line regex? My regex knowledge is not strong, but I think it wouldn't be possible to have multiple classes **and** "3 out of 4" feature. 

Maybe we could have a list of classes instead? Something like (needs better names probably):

```
character_classes_validation_minumum_true: 3,
character_classes_validation: [
  '<first regex>',
  '<second regex>',
  '<third regex>',
  '<fourth regex>'
]
```

kpi/views/environment.py
```diff
@@ -83,6 +83,12 @@ class EnvironmentView(APIView):
                 ).exists()
             )
         ),
+        'PASSWORD_VALIDATION_CUSTOM_RULES_ENABLED',
+        'MINIMUM_LENGTH_VALIDATION',
+        'USER_ATTRIBUTE_SIMILARITY_VALIDATOR',
+        'COMMON_PASSWORD_VALIDATION',
+        'CUSTOM_CHARACTER_CLASSES_ENABLED',
+        'MOST_RECENT_PASSWORD_VALIDATION',
```
magicznyleszek, line 91: i'm not sure what does this file do, but if I'm right and it's the definition for `/environment` endpoint, then I can say we don't need all these here exposed in API.

Front end will need to know nothing about custom rules, only the disabling of zxcvbn (other task).

kobo/settings/base.py
magicznyleszek: I'm thinking that it would be ideal if all these singular regexes could be described in admin UI. So if superuser adds `<some regex>` they could describe it, e.g.

```
…classes: [
  ['Uppercase letters', '<regex>']
]
```

This way it would be easier to remember what each regexes is supposed to do. Plus it would give us opportunity to give some default values as examples - either as actual default or in documentation

kobo/settings/base.py
```diff
@@ -349,6 +349,38 @@
...
     ),
+    # Enable the use of validators with custom rules
+    'PASSWORD_VALIDATION_CUSTOM_RULES_ENABLED': (
+        True,
+        'Enable custom password validation rules. '
+        'The subsequent rules will not work unless this field is Enabled',
+    ),
+    # Built-in password validator
+    'MINIMUM_LENGTH_VALIDATION': (
+        10,
```
magicznyleszek, line 360: How do we handle in django disabling such rule? Do we simply require superuser to type `0` to disable it?

kobo/settings/base.py
JacquelineMorrissette: I don't think it will be hard to do at all since we're doing very similar work: https://stackoverflow.com/a/58367983

kpi/views/environment.py
JacquelineMorrissette: Sounds good, I will remove it then

kobo/settings/base.py
JacquelineMorrissette: Sure, I like this idea

kobo/settings/base.py
```diff
@@ -349,6 +349,38 @@
         'having the system empty it automatically.',
         'positive_int_minus_one',
     ),
+    # Enable the use of validators with custom rules
+    'PASSWORD_VALIDATION_CUSTOM_RULES_ENABLED': (
```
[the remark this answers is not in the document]
JacquelineMorrissette: No, I don't think we need to have a global toggle just for custom rules. I'll change it so it's just for disabling zxcvbn

PR #4487, opened 2023-06-13
Improve labeling/ordering of transcripts/translations for grouped audio questions

### LINKED ISSUE #4394: [NLP] Additional columns for grouped question appear after the entire group

### LINKED ISSUE #4395: [NLP] Grouped transcript and translation columns do not use the question label in the table view

srartese:
## Description

Place transcripts and translations immediately after their related audio questions, and use field labels instead of XML values when 'Labels' is selected in the Table display option.

These fixes for grouped audio apply to the table view column headers and the single submission model rows.

## Related issues

Fixes #4394
Fixes #4395

## Reviewer notes

Diffs from commits ignored in `.git-blame-ignore-revs` commits still show up in the 'Files changed' view for reviewers. So it's still better to isolate formatting commits to their own PR.


p2edwards (approved):
I added a few commits with a fix and some code simplifications, and I've updated the PR description for the release notes.

p2edwards:
Rebased recent changes on current release branch.

Issue #4491, opened 2023-06-13
Add support for User Metadata custom translation labels


JacquelineMorrissette:
Add the support for custom translations can be used on labels for the User Metadata fields

Issue #4490, opened 2023-06-13
Use configured user metadata fields in the user's profile settings


JacquelineMorrissette:
Use configured user metadata fields in the user's profile settings
The fields need to respect the custom labels and the required-ness of the field configuration

JacquelineMorrissette:
Done with #4591

PR #3831, opened 2022-05-26
Explain django-constance newline handling

bufke:
This change better explains the reasoning for new line handling, necessary with django-constance 2.7+

## Related issues

Related to https://github.com/kobotoolbox/kpi/pull/3825

PR #4497, opened 2023-06-14
Add labels to User and Project metadata

### LINKED ISSUE #4488: Allow labels to be configured in Django Admin interface
Labels need to be configurable in the admin interface.
Full name needs to be marked as required

* Constance needs to allow the `label` field in the JSON schema for 'USER_METADATA_FIELDS` 
* JSON validation needs to be updated
* Constance documentation needs to be updated

### LINKED ISSUE #4492: Allow Project Metadata Fields to be configured in Django Admin interface
Labels for Project Metadata need to be configurable in the admin interface.
Allow description to be marked as required

* Constance needs to allow the label field in the JSON schema for 'PROJECT_METADATA_FIELDS`
* JSON validation needs to be updated
* Constance documentation needs to be updated

JacquelineMorrissette:
## Checklist

3. [x] Ensure the tests pass

## Description
Add custom label support and custom translation support for User and Project Metadata fields

## Related issues
closes #4488 
closes #4492
part of #4490 
part of #4491


kobo/settings/base.py
```diff
@@ -274,36 +274,48 @@
...
+            {'name': 'organization', 'required': False,},
+            {'name': 'organization_website', 'required': False,},
+            {'name': 'sector', 'required': False,},
+            {'name': 'gender', 'required': False,},
+            {'name': 'bio', 'required': False,},
+            {'name': 'city', 'required': False,},
+            {'name': 'country', 'required': False,},
+            {'name': 'twitter', 'required': False,},
+            {'name': 'linkedin', 'required': False,},
+            {'name': 'instagram', 'required': False,},
```
jnm, line 287: Which tool is adding these trailing commas, assuming it's a tool and not you doing it manually? :stuck_out_tongue: I can't reproduce it with python-black.

Let's keep unnecessary changes to a minimum. These, in particular, don't seem very helpful. The commas at the very end of the lines are nice because they allow us to add, remove, and reorder items at the end without worrying about commas and without adding noise to the diff because an adjacent line had to have a trailing comma changed.  However, the trailing comma within the dictionary doesn't help with this.

kobo/settings/base.py
```diff
@@ -274,36 +274,48 @@
...
         ]),
         # The available fields are hard-coded in the front end
         'Display (and optionally require) these metadata fields for users. '
-        "Possible fields are 'organization', 'organization_website', "
+        "Possible fields are 'full_name', 'organization', 'organization_website', "
         "'sector', 'gender', 'bio', 'city', 'country', 'twitter', 'linkedin', "
-        "and 'instagram'",
+        "and 'instagram'.\n\r"
```
jnm, line 293: I've never seen `\n\r` before. The reverse of that, `\r\n` is common to indicate a single line break (new line + carriage return) and is part of many [standards](https://github.com/kobotoolbox/kpi/pull/3825#issuecomment-1155501572), but at some point django-constance [changed](https://github.com/kobotoolbox/kpi/pull/3831) to use only `\n` for line breaks in **values**. However, that doesn't necessarily mean anything for these field *descriptions*.

I was curious so I did some tests: it looks like your `\n\r` is being interpreted as 2 consecutive newlines and translated to `<br><br>` in the Django Admin HTML. I tested and the same thing can be achieved with `\n\n`, which should be easier for a human reader to understand as two consecutive newlines.

kobo/settings/base.py
```diff
@@ -274,36 +274,48 @@
...
+            #     {'default': 'Operational Purpose'},
+            # },
+            # {'name': 'collects_pii', 'required': False, 'label': {'default': 'Collects PII'},},
         ]),
         # The available fields are hard-coded in the front end
         'Display (and optionally require) these metadata fields for projects. '
-        "Possible fields are 'sector', 'country', 'operational_purpose', and "
-        "'collects_pii'.",
+        "Possible fields are 'sector', 'country', 'operational_purpose', "
+        "'collects_pii', and 'description'\n\r"
```
jnm, line 317: See previous comment about line breaks

kpi/fields/jsonschema_form_field.py
```diff
@@ -66,11 +80,20 @@ def __init__(self, *args, **kwargs):
...
+                    'name': {'type': 'string',},
+                    'required': {'type': 'boolean',},
+                    'label': {
+                        'type': 'object',
+                        'uniqueItems': True,
+                        'properties': {
+                            'default': {'type': 'string',},
+                        },
+                        'required': ['default',],
+                        'additionalProperties': True,
```
jnm, line 95: LGTM except for the trailing commas within single-line dictionaries and lists; see previous comment

kobo/settings/base.py
```diff
@@ -274,36 +274,48 @@
...
-            # {'name': 'collects_pii', 'required': False},
+            {'name': 'sector', 'required': False, 'label': {
+                'default': 'Sector',
+                'some-other-language': (
+                    'This is an example and will never display because some-other-language '
+                    'is not a valid language code.'
+                ),
+            },},
+            {'name': 'country', 'required': False, 'label': {'default': 'Country'},},
+            # {'name': 'description', 'required': False, 'label': {'default': 'Description'},},
```
jnm, line 308: Oops, I don't think "description" should be disabled by default

kobo/settings/base.py
```diff
@@ -274,36 +274,48 @@
...
         'metadata_fields_jsonschema'
     ),
     'PROJECT_METADATA_FIELDS': (
         json.dumps([
-            {'name': 'sector', 'required': False},
-            {'name': 'country', 'required': False},
-            # {'name': 'operational_purpose', 'required': False},
-            # {'name': 'collects_pii', 'required': False},
+            {'name': 'sector', 'required': False, 'label': {
+                'default': 'Sector',
```
jnm, line 301: I see the "Make label field optional" [commit](https://github.com/kobotoolbox/kpi/pull/4497/commits/bff0901589c91ef8cb511a8ea248158e6dddea44), but `label` is still included in these project metadata defaults even if it was removed from the user metadata defaults. ~~We should be consistent.~~ Er, it's not just consistency: if we override `label` by default, then the translated defaults in Transifex won't get used.

Still, we do need a way to show the `some-other-language` example.

Looking more, we have a prior problem with `MFA_LOCALIZED_HELP_TEXT`: the default (hard-coded) value we provide is not marked as a translatable string, and even if it were, there's no way for it to be localized [in the API response](https://github.com/kobotoolbox/kpi/blob/bff0901589c91ef8cb511a8ea248158e6dddea44/kpi/views/environment.py#L63-L71) unless the superuser translates it. (cc @noliveleger)

kpi/tests/api/test_api_environment.py
```diff
@@ -103,6 +103,24 @@ def test_template_context_processor(self):
...
+                if 'label' in item.keys():
+                    assert 'default' in item['label'].keys()
+                
+        self.client.logout()
+        response = self.client.get(self.url, format='json')
+        assert response.status_code == status.HTTP_200_OK
+        user_metadata = response.data['user_metadata_fields']
+        project_metadata = response.data['project_metadata_fields']
+        check_metadata_for_label(user_metadata)
+        check_metadata_for_label(project_metadata)
```
jnm, line 121: I'm not sure I understand the purpose of this test. It looks like it only operates on our hard-coded default values. Could you explain under what circumstances this test would fail?

jnm (changes requested):

kobo/settings/base.py
```diff
@@ -274,48 +274,42 @@
...
+            {'name': 'sector', 'required': False,},
+            {'name': 'country', 'required': False,},
+            {'name': 'description', 'required': False},
+            # {'name': 'operational_purpose', 'required': False},
+            # {'name': 'collects_pii', 'required': False},
         ]),
         # The available fields are hard-coded in the front end
         'Display (and optionally require) these metadata fields for projects. '
         "Possible fields are 'sector', 'country', 'operational_purpose', "
         "'collects_pii', and 'description'\n\r"
```
jnm, line 309: This `\n\r` is still at large :detective:

kobo/settings/base.py
jnm: This `\n\r` is still at large :detective:

kobo/settings/base.py
jnm: OK, I see you put an example of using another language into the help text. This seems good :+1:

kpi/tests/api/test_api_environment.py
jnm: No test is probably better than an ineffective test that lends a false sense of security, but we probably should have unit tests for this feature

kpi/tests/api/test_api_environment.py
```diff
@@ -101,7 +101,7 @@ def test_template_context_processor(self):
         context = RequestContext(HttpRequest())  # NB: empty request
         template = Template('{{ config.TERMS_OF_SERVICE_URL }}')
         result = template.render(context)
-        self.assertEqual(result, constance.config.TERMS_OF_SERVICE_URL)
+        self.assertEqual(result, constance.config.TERMS_OF_SERVICE_URL)
```
jnm, line 104: Stray space :)

kpi/tests/api/test_api_environment.py
jnm: We're just passing through JSON from Constance to the `/environment` endpoint in this PR, and that doesn't really need a test. I may want to manipulate the JSON so that it only returns the label for the currently-active language, but I can do that in another PR. That work would need unit tests.

PR #4503, opened 2023-06-19
Upgrade to Storybook 7

bufke:
A (mostly) automated upgrade to storybook 7. I had to --force the package installs because reflux (and maybe others) creates impossible peer dependencies.


.storybook/main.js
```diff
@@ -76,16 +64,18 @@ module.exports = {
...
-
     if (process.env.MEASURE) {
       const smp = new SpeedMeasurePlugin();
       return smp.wrap(config);
     }
     return config;
   },
+  docs: {
+    autodocs: true
+  }
```
LMNTL, line 78: I like the autodocs! They might need some work to get fully functional, though.  `KoboPrompt` and `KoboModal` don't close and obscure the docs. `KoboDropdown` and `KoboSelect` also clip outside of the doc view's bounds when expanded, but the docs are still accessible and the story displays fine on the Primary tab so that's probably not worth spending much energy on.

LMNTL (changes requested):

.storybook/main.js
bufke: A migration added this. I could go either way on it. Could disable it or just ignore the defects for now.

LMNTL: I went back and forth about it, and I think the immediate benefit for frontend devs from having code examples outweighs having some wonky Storybook bits, so I'm approving. If you don't mind, could you create an issue/task to update the docs pages?

LMNTL (approved):

PR #4505, opened 2023-06-20
Upgrade React Router to 6.14.2 and use upstream usePrompt

bufke:
- Upgrade React Router to ~~6.13~~ 6.14.2
- Use React Router Data API compatible router
- Replace custom usePrompt with upstream react router usePrompt
- Remove history router kludge

# Reviewer Notes

- es6 files were renamed to be importable by typescript (es6 extension is not a recognized standard)
- The usage of the injected router is scary, it avoids an import loop. The injected router should not be used on new code (this it's in legacy.tsx).
- setTimeout is used on reflux states. This is necessary because the states get initialized before the router is injected. A better solution would be to remove reflux.
- allRoutes.es6 got split into a new file routes.tsx - some typescript is better than none.
- Oh no - the new typescript routes found a "impossible" route that was never used before! Did it used to do something?
- legacy.tsx getCurrentRoute now uses router.state. This is necessary to get the correct order of routing for isFooRouteActive functions to work. Without this, they return the prior route.
- unstable_usePrompt doesn't inspire confidence, but is surely better than our prior kludge solution mentioned on a github issue.
- history and router are not API compatible. :warning: :bug: we need to review this further.
- There is a new lazy function from react router for lazy loading. I wonder if we should use that instead of the react one? Maybe it would be less code or could avoid the need for Suspend.


jsapp/js/components/formEditors.js
magicznyleszek, line 1: Do you think this would be a good moment to rename all `.es6` files?

jsapp/js/components/header/searchBoxStore.ts
```diff
@@ -30,7 +30,7 @@ class SearchBoxStore extends Reflux.Store {
   };
 
   init() {
-    history.listen(this.onRouteChange.bind(this));
+    setTimeout(() => router.subscribe(this.onRouteChange.bind(this)));
```
magicznyleszek, line 33: If we end up with these ugly `setTimeout`s (which are fine hacks IMHO), it would be good to leave a comment somewhere. Adding same comment to each `setTimeout` would ba an overkill, but perhaps a comment in `legacy.tsx` would suffice.

jsapp/js/components/processing/processingUtils.ts
```diff
@@ -54,5 +54,5 @@ export function openProcessing(
   const route = ROUTES.FORM_PROCESSING.replace(':uid', assetUid)
     .replace(':qpath', qpath)
     .replace(':submissionEditId', submissionEditId);
-  history.push(route);
+  router.naviate(route);
```
magicznyleszek, line 57: typo

jsapp/js/router/legacy.tsx
```diff
@@ -84,3 +83,8 @@ export function routerGetAssetId() {
   }
   return null;
 }
+
+export let router: any = null;
```
magicznyleszek, line 87: Just a note: let's make sure we will change this `any` to something useful before merging

jsapp/js/router/router.tsx
```diff
@@ -0,0 +1,361 @@
...
+              />
+            }
+          />
+          {/**
+           * TODO change this HACKFIX to a better solution
+           *
+           * Used to force refresh form sub routes. It's some kine of a weird
+           * way of introducing a loading screen during sub route refresh.
+           * See: https://github.com/kobotoolbox/kpi/issues/3925
+           **/}
```
magicznyleszek, line 334: As mentioned on Zulip, we can drop this 🎉 

> I've looked at that `/forms/<uid>/reset` route. It is being used only in `triggerRefresh()` function from `formViewTabs.es6`. The idea was that when user clicks a menu item, e.g. "Table" from "DATA", the route would reload itself. This is interesting, but since I broke it (of course it was me) over 2 years ago (see https://github.com/kobotoolbox/kpi/commit/f14005c9bdeabd9246e0267e637e60772c57a4fb#diff-cd8026c65e0451326d9e9eb4de92a70018810bf708c2e70affcdab16773a93b1L251) it stopped working at all

jsapp/js/components/formEditors.js
bufke: I don't have a strong opinion. It would be nice to not have them. It takes longer to convert them to ts. We should do this a separate PR.

jsapp/js/components/header/searchBoxStore.ts
bufke: Good idea, I'll comment the router variable.

jsapp/js/router/legacy.tsx
bufke: `import type { Router } from "@remix-run/router";`

p2edwards (commented):
Since we got to talk today, I'll try changing some of the unhandled `router?.` to `router!.` to surface any issues with the router initialization.

I also want to subject it to more manual QA before merging.

It will break some work-in-progress branches like https://github.com/kobotoolbox/kpi/pull/4535.

jsapp/js/router/router.tsx
p2edwards: FORM_KOBOCAT should be removed while we're here, it's superceded by the feature at `/forms/:uid/settings/media`

jsapp/js/components/formViewTabs.es6
```diff
@@ -64,6 +64,7 @@ class FormViewTabs extends Reflux.Component {
 
   triggerRefresh(evt) {
     if ($(evt.target).hasClass('active')) {
+      // ROUTES.FORM_RESET
```
p2edwards, line 67: I added this comment for searchability, since removing this section was missed [here](https://github.com/kobotoolbox/kpi/pull/4505#discussion_r1238709429)

jsapp/js/components/map.es6
```diff
@@ -27,46 +27,51 @@ import {
   QUERY_LIMIT_DEFAULT,
 } from '../constants';
```
p2edwards, line 28: Sorry in advance for `map.es6`. Best to review these commits separately.

- 🤖 Prettier and ESLint autofixes [here](https://github.com/kobotoolbox/kpi/pull/4505/commits/be3a9b6041e451b02c671220be50b5ca33cae1a8)
- ✋ Manual ESLint fixes [here](https://github.com/kobotoolbox/kpi/pull/4505/commits/16f242bd05aafa5b6ae8bb5d31670ccb0e568aa1)
	- Manually hoist variables to make scope clearer and use 'let'
	- Carefully replace 8 loose comparisons with strict ones (`===`, `!==`)
		- [https://dorey.github.io/JavaScript-Equality-Table/](https://dorey.github.io/JavaScript-Equality-Table/) was useful here
- ✋ Fix map screen crash [here](https://github.com/kobotoolbox/kpi/pull/4505/commits/bcdbaf0a42c2576446e7f48c873e68e981a209ef)

.git-blame-ignore-revs
```diff
@@ -0,0 +1,4 @@
+# .git-blame-ignore-revs
+
+# Applied automated eslint autofixes and prettier to map.es6
+be3a9b6041e451b02c671220be50b5ca33cae1a8
```
p2edwards, line 4: This entry prevents [be3a9b6] from dominating the [online git-blame view for map.es6](https://github.com/kobotoolbox/kpi/blame/968da59c86307ad2afe929c41bf7b8c101418caa/jsapp/js/components/map.es6).

Reference:
- https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view
- https://www.michaelheap.com/git-ignore-rev/

Internal discussion:

- https://chat.kobotoolbox.org/#narrow/stream/4-Kobo-Dev/topic/Prettier.2FESLint.20Autofixes/near/238951

jsapp/js/router/router.tsx
```diff
@@ -0,0 +1,345 @@
...
+            path={ROUTES.FORM_EDIT}
+            element={
+              <PermProtectedRoute
+                protectedComponent={FormPage}
+                requiredPermissions={[PERMISSIONS_CODENAMES.view_asset]}
+              />
+            }
+          />
+        </Route>
+        <Route path='*' element={FormNotFound} />
```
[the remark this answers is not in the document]
p2edwards: **Edit:** This wasn't quite right yet, `element=` needed to be a JSX not a function name, and then since FormNotFound is lazy-loaded, wrapped in suspense. Fixed [here](https://github.com/kobotoolbox/kpi/pull/4505/commits/968da59c86307ad2afe929c41bf7b8c101418caa).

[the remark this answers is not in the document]
bufke: Yeah TS picked up a few problems. Fix seems fine.

jsapp/js/components/formViewTabs.es6
bufke: Ok. I guess this is why django has that reverse function.

PR #4506, opened 2023-06-21
Updating usage endpoints for organizations and subscriptions

LMNTL:
## Checklist

3. [X] Ensure the tests pass

## Description

Updated api/v2/service_usage to account for organizations and subscriptions.

#### Included changes to `service_usage`
- Refactored from 7N queries per asset to 7-9 queries flat.
- Endpoint can now optionally be queried by organization ID, returning usage data for each user in an organization.
 - When using the `organization_id` parameter, endpoint now tracks usage based on the organization's subscription date (if they have a subscription)
 - Endpoint also returns the start date for the `current_month` and `current_year` totals
#### General changes:
- `service_usage` and `asset_usage` endpoints now return yearly counts for submissions and NLP usage.
- Submissions and NLP usage are now tracked by day instead of month. A separate PR for KoboCat will address retaining daily submission counters for 366 days (one leap year) instead of 31 days. [Link](https://github.com/kobotoolbox/kobocat/tree/feature/usage-limits-api)
- Small changes to the usage page to keep it even with the updated API.


kpi/serializers/v2/service_usage.py
```diff
@@ -62,64 +80,180 @@ def get_storage_bytes(self, asset):
...
+        # Get the organization ID passed in from the query parameters
+        organization_id = self.context['request'].query_params.get('organization_id')
+        if organization_id:
+            try:
+                organization = Organization.objects.filter(
+                    owner__organization_user__user=user,
+                    id=organization_id,
+                )
+                # If the user is in an organization, get all org users so we can query their total org usage
+                organization_users = OrganizationUser.objects.filter(organization=organization).select_related('user')
```
bufke, line 172: I think you could reduce the code here by something like (pseudo code)

User.objects.filter(orgnaization__id=organization_id)

Often it's best to base the FROM part of the query on the table you actually want to return a list of, when possible at least. Then join that table to where you need to apply a filter.

The 404 may not be right here. I kind of don't care and this is pedantic, but a filter should simply filter the existing data. If that filter returns nothing. But if it's easier to raise either a 404 or 400 - I'm fine with it.

kpi/serializers/v2/service_usage.py
```diff
@@ -62,64 +80,180 @@ def get_storage_bytes(self, asset):
...
+                users = [org_user.user.id for org_user in list(organization_users)]
+            except ObjectDoesNotExist:
+                # Couldn't find organization or organization users, raise an error
+                raise Http404
+            # Get the organization's subscription, if they have one
+            subscription = Subscription.objects.filter(
+                status__in=['active', 'past_due', 'trialing'],
```
bufke, line 179: We should probably make this list a constant. Even though it's used just once, it's more readable to understand why these three strings are special. Just the variable name itself would be good documentation.

bufke (changes requested):

jsapp/js/account/plan.component.tsx
```diff
@@ -27,7 +27,8 @@ import Button from 'js/components/common/button';
 import classnames from 'classnames';
 import LoadingSpinner from 'js/components/common/loadingSpinner';
 import {notify} from 'js/utils';
-import {BaseProduct} from "js/account/subscriptionStore";
+import {BaseProduct} from 'js/account/subscriptionStore';
```
bufke, line 30: I know you're just running prettier, but this import isn't being used. We should remove it.

kpi/serializers/v2/service_usage.py
```diff
@@ -1,22 +1,37 @@
+from django.db.models import Sum, Q
+from django.db.models.functions import Coalesce
+from django.utils import timezone
+from djstripe.models import Subscription
```
bufke, line 4: This import appears to generate a RuntimeError at startup when stripe isn't configured. We decided that stripe should be optional. I suspect you could do a condition check and run organizations.something to fetch check if there is a subscription without needing to import.

PR #4509, opened 2023-06-22
Organizations API

bufke:
## Description

Allow creation/updating/deleting of organizations via API.

## Notes

I left in the create on list hack for now - though it really only runs on list view now. It won't do anything silly on other api views.


kobo/apps/organizations/tests/test_organizations_api.py
```diff
@@ -46,3 +49,16 @@ def test_api_returns_org_data(self):
...
+        data = {'name': 'edit'}
+        res = self.client.patch(self.url_detail, data)
+        self.assertContains(res, data['name'])
+
+        user = baker.make(User)
+        self.client.force_login(user)
+        org_user = self.organization.add_user(user=user)
+        with self.assertNumQueries(3):
+            res = self.client.patch(self.url_detail, data)
+        self.assertEqual(res.status_code, 403)
```
LMNTL, line 64: Small question: why do you use `assertNumQueries` for a request that will 403? I don't think it's wrong, but I'm curious. I'd expect enforcing a certain number of queries for a successful request, but not for one that (hopefully) isn't accessing resources.

LMNTL (commented):
This is looking good! I just have a question about a test case.

kobo/apps/organizations/tests/test_organizations_api.py
bufke: The real reason is that I use that to debug, getting a printout of sql statements. I probably just left it here by mistake. I agree that it should be moved to a success path instead.

bufke: OK rebased and better tests (including new list view test). Thanks for spotting this.

LMNTL (approved):

PR #4516, opened 2023-06-27
Removes legacy interface button from drawer

### LINKED ISSUE #4515: Remove legacy projects link from UI
## Description

For some time already we don't really need it. We gave users some time to stop using it. Now it is the time to remove that button:

image

Please check if we use that icon anywhere else and remove it from the repo :)

duvld:
## Description

Removes the link to the legacy interface on the drawer.

## Related issues
Fixes #4515


jsapp/js/dataInterface.ts
```diff
@@ -627,7 +627,7 @@ export interface AccountResponse {
   email: string;
   server_time: string;
   date_joined: string;
-  projects_url: string;
+  projects_url: string; // TODO: no longer used, kept here for response typing
```
magicznyleszek, line 630: I feel like keeping this as TODO would make us think we would like to change this, and most probably we wouldn't - unless Back end stops sending it in the response.

Let's change it to a descriptive comment (the JSDoc one in line above it) stating that this is a link to a legacy view containing a list of projects.

jsapp/js/components/drawer.es6
```diff
@@ -180,16 +180,6 @@ const Drawer = observer(class Drawer extends Reflux.Component {
...
             <HelpBubble/>
           }
-          { sessionStore.isLoggedIn &&
-            sessionStore.currentAccount.projects_url &&
-            <a href={sessionStore.currentAccount.projects_url}
-              className='k-drawer__link'
-              target='_blank'
-              data-tip={t('Projects (legacy)')}
-            >
-              <i className='k-icon k-icon-globe' />
```
magicznyleszek, line 190: Please delete that `globe.svg` from repo - just make sure to leave the `globe-alt.svg` 😉

magicznyleszek (changes requested):

PR #4517, opened 2023-06-28
Remove build frontend code in pytest CI step

bufke:
Do not build frontend assets when running CI pytest step. Upgrade django-webpack-loader so that FakeWebpackLoader can be used in tests.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -37,12 +37,6 @@ jobs:
           - 6379:6379
     steps:
     - uses: actions/checkout@v2
-    - uses: actions/setup-node@v2
-      with:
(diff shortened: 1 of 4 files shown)


.github/workflows/pytest.yml
```diff
@@ -63,9 +63,6 @@ jobs:
       run: npm install
     - name: Update translations
       run: git submodule init && git submodule update --remote && python manage.py compilemessages
-    # Intercom tests in Python fail if the front-end code has not been built
-    - name: Build front-end code
-      run: SKIP_TS_CHECK=true npm run build
```
p2edwards, line 68: Seems like we can skip the node and `npm install` steps too now.

PR #4518, opened 2023-06-28
Set date and time correctly when project is deployed

### LINKED ISSUE #4504: Bug: deployed date is wrong
The deployed date is sometimes off. It seems that it happens only for some older projects. It doesn't happen with new projects.

Reproduction:

Case A:

1. Create some new project
2. Deploy that project
3. Go to `#/projects/home`
4. Notice that "Date deployed" is correct ✅ 

Case B

1. Have some existing old project on https://kf.tri.kbtdev.org/
2. Deploy that project
3. Go to `#/projects/home`
4. Notice that "Date deployed" is wrong :bug:

---

Real life example:
1. I have an old draft project:
  ```
  {
    "url": "https://kf.tri.kbtdev.org/api/v2/assets/aDv6L57hqjGkeH4fc3zoBW/",
    "date_created": "2023-03-23T00:11:08.724213Z",
    "date_modified": "2023-06-15T14:03:51.605925Z",
    "date_deployed": null,
    …
  ```
2. I deploy it
3. My now deployed project has new dates:
  ```
  { 
    "url": "https://kf.tri.kbtdev.org/api/v2/assets/aDv6L57hqjGkeH4fc3zoBW/",
    "date_created": "2023-03-23T00:11:08.724213Z",
    "date_modified": "2023-06-19T21:39:17.659034Z",
    "date_deployed": "2023-06-15T14:03:51.664186Z",
    …
  ```
4. The `date_modified` and `date_deployed` should be almost identical, but the difference is around 4 days :bug:

noliveleger:
## Description

Update modified date of asset version when it is saved and update asset deployed date with it.

## Related issues

Closes #4504

### DIFF
--- a/kpi/deployment_backends/mixin.py
+++ b/kpi/deployment_backends/mixin.py
@@ -75,6 +75,7 @@ class DeployableMixin:
         latest_version.deployed = True
         latest_version.save()
         self.date_deployed = latest_version.date_modified
+
         if save:
             self.save(
                 update_fields=['date_deployed'],
--- a/kpi/deployment_backends/mixin.py
+++ b/kpi/deployment_backends/mixin.py
@@ -75,7 +75,6 @@ class DeployableMixin:
         latest_version.deployed = True
         latest_version.save()
         self.date_deployed = latest_version.date_modified
-
         if save:
             self.save(
                 update_fields=['date_deployed'],
--- a/kpi/models/asset_version.py
+++ b/kpi/models/asset_version.py
@@ -75,19 +75,6 @@ class AssetVersion(models.Model):
             self.date_modified.strftime('%Y-%m-%d %H:%M'),
             ' (deployed)' if self.deployed else '')
 
-    def save(
-        self,
-        force_insert=False,
-        force_update=False,
-        using=None,
-        update_fields=None,
-    ):
-        if self.pk:
-            self.date_modified = timezone.now()
-
-        super().save(
-            force_insert=force_insert,
-            force_update=force_update,
-            using=using,
-            update_fields=update_fields,
-        )
+    def save(self, *args, **kwargs):
+        self.date_modified = timezone.now()
+        super().save(*args, **kwargs)
--- a/kpi/models/asset_version.py
+++ b/kpi/models/asset_version.py
@@ -74,3 +74,20 @@ class AssetVersion(models.Model):
             self.asset.uid, self.uid,
             self.date_modified.strftime('%Y-%m-%d %H:%M'),
             ' (deployed)' if self.deployed else '')
+
+    def save(
+        self,
+        force_insert=False,
+        force_update=False,
+        using=None,
+        update_fields=None,
+    ):
+        if self.pk:
+            self.date_modified = timezone.now()
+
+        super().save(
+            force_insert=force_insert,
+            force_update=force_update,
+            using=using,
+            update_fields=update_fields,
+        )
--- a/kpi/tests/test_asset_versions.py
+++ b/kpi/tests/test_asset_versions.py
@@ -1,8 +1,14 @@
 # coding: utf-8
 import json
 from copy import deepcopy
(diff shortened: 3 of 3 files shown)


kpi/models/asset_version.py
```diff
@@ -74,3 +74,20 @@ def __str__(self):
             self.asset.uid, self.uid,
             self.date_modified.strftime('%Y-%m-%d %H:%M'),
             ' (deployed)' if self.deployed else '')
+
+    def save(
+        self,
+        force_insert=False,
```
bufke, line 80: This would be better as *args and **kwargs. That way if django ever changed it, this code wouldn't break.

kpi/deployment_backends/mixin.py
```diff
@@ -75,6 +75,7 @@ def _mark_latest_version_as_deployed(self, save: bool = False):
         latest_version.deployed = True
         latest_version.save()
         self.date_deployed = latest_version.date_modified
+
```
bufke, line 78: Probably a mistake to commit this newline.

kpi/models/asset_version.py
```diff
@@ -74,3 +74,20 @@ def __str__(self):
...
+        force_update=False,
+        using=None,
+        update_fields=None,
+    ):
+        if self.pk:
```
bufke, line 85: Why wouldn't it be set on creation? And if so, it may be cleaner to use `auto_now=True` in the model - that's a pretty typical way to build modified dates.

bufke (changes requested):

kpi/models/asset_version.py
noliveleger: Because on the creation it will be pick the default, so no need to override it again. But I guess it does not hurt anyway.
AFAIK, we try to avoid `auto_now` (and `auto_now_add`) as much as possible because it alters the field even if we don't want to, e.g. running a management command that updates the model without changing dates. 
We can discuss about this with @jnm if needed.

kpi/models/asset_version.py
noliveleger: Sure! 
But many places in the code (because of me 😇 ) are using the current signature and not *args, **kwargs. Should we create an issue to change them or should we wait until it break - if it breaks one day?

kpi/deployment_backends/mixin.py
noliveleger: Indeed.

kpi/models/asset_version.py
bufke: For new/modified code - I would advise using *kwargs, unless upstream documentation says otherwise. If I was using a variable such as force_insert, I would also declare that explicitly instead of fetching from kwargs.

IMO It's not a valuable time investment going back and changing this. Yes - if it breaks, we can change it.

kpi/models/asset_version.py
bufke: If we avoid auto_now - I'd probably just drop the if statement - I think it's more readable to see that it should always be set. I can't imagine a performance difference here. 

auto_now

> The field is only automatically updated when calling [Model.save()](https://docs.djangoproject.com/en/4.2/ref/models/instances/#django.db.models.Model.save). The field isn’t updated when making updates to other fields in other ways such as [QuerySet.update()](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#django.db.models.query.QuerySet.update), though you can specify a custom value for the field in an update like that.

auto_now is one of those Python only implementations that will would cause a "useless" migration. I certainly don't want to make a slow migration that does absolutely nothing. Do you know if it is indeed slow, maybe postgres would be smart enough to not rewrite tons of disk. Could discuss more if desired. If we could avoid slow migrations, then I think auto_now is better. But not worth a painful migration. Do we have any existing policy for small things like help_text? My personal and maybe controversial policy is to not make "useless" migrations. Let django raise a warning about it.

PR #4521, opened 2023-07-04
Clarify where (re)deployment writes to database

jnm:
This stuff helps me understand how `Asset._deployment_data` gets written to the database during the deployment process.

Merging into `feature/my-projects` because `kpi/serializers/v2/deployment.py` does not exist in the `beta` branch.

### DIFF
--- a/kpi/deployment_backends/mixin.py
+++ b/kpi/deployment_backends/mixin.py
@@ -9,11 +9,11 @@ from .base_backend import BaseDeploymentBackend
 
 class DeployableMixin:
 
-    def async_media_files(self, force=True):
+    def sync_media_files_async(self, always=True):
         """
         Synchronize form media files with deployment backend asynchronously
         """
-        if force or self.asset_files.filter(
+        if always or self.asset_files.filter(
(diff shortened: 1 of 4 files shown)


kpi/deployment_backends/mixin.py
```diff
@@ -62,14 +62,14 @@ def has_deployment(self):
     def set_deployment(self, deployment: BaseDeploymentBackend):
         setattr(self, '__deployment_backend', deployment)
 
-    def _mark_latest_version_as_deployed(self, save: bool = False):
+    def _mark_latest_version_as_deployed(self, save: bool = True):
```
jnm, line 65: I could only find this used in one other place: `sync_kobocat_xforms`. It already had `save=True`, so I left it alone.

noliveleger (approved):
LGTM. 

I misread your comment (omit `v2`): 

> Merging into feature/my-projects because kpi/serializers/v2/deployment.py does not exist in the beta branch.


Right, `DeploymentSerializer` exists only in `v1` in beta.

PR #4522, opened 2023-07-04
Improve redeployment logic

jnm:
Redeploy unconditionally, even when `version` in the `_deployment_data` has not changed. See internal discussion at https://chat.kobotoolbox.org/#narrow/stream/4-Kobo-Dev/topic/Deployment.20date.20feedback/near/245527.

Always update `date_deployed` on the asset even if the version has not been modified. Likewise, do not update `date_modified` on the version when deploying if that same version is already deployed.

## Description

Always update "Date deployed" when "REDEPLOY" is clicked, even if the form version is the same (i.e. the content has not been changed). Other things about the project may have changed, e.g. form media files.

### DIFF
--- a/kpi/serializers/v2/deployment.py
+++ b/kpi/serializers/v2/deployment.py
@@ -47,16 +47,29 @@ class DeploymentSerializer(serializers.Serializer):
         asset = self.context['asset']
         deployment = asset.deployment
 
-        if 'backend' in validated_data and \
-                validated_data['backend'] != deployment.backend:
+        if (
+            validated_data.pop('backend', deployment.backend)
+            != deployment.backend
+        ):
             raise serializers.ValidationError(
-                {'backend': 'This field cannot be modified after the initial '
-                            'deployment.'})
+                {
+                    'backend': (
+                        'This field cannot be modified after the initial '
+                        'deployment.'
+                    )
+                }
+            )
+
+        if validated_data.keys() == set(('active',)):
+            # This request is only changing the active (aka archived) state,
+            # not actually doing a deployment
+            deployment.set_active(validated_data['active'])
+            return deployment
 
         self._raise_unless_current_version(asset, validated_data)
         asset.deploy(
             backend=deployment.backend,
-            active=validated_data.get('active', deployment.active)
+            active=validated_data.get('active', deployment.active),
         )
 
         return deployment
--- a/kpi/serializers/v2/deployment.py
+++ b/kpi/serializers/v2/deployment.py
@@ -40,9 +40,9 @@ class DeploymentSerializer(serializers.Serializer):
 
     def update(self, instance, validated_data):
         """
-        If a `version_id` is provided and differs from the current
-        deployment's `version_id`, the asset will be redeployed. Otherwise,
-        only the `active` field will be updated
+        Redeploy unconditionally. If people want to deploy again the same
+        version that's already deployed, let them. They may have a good reason
+        for doing so, e.g. updating form media files.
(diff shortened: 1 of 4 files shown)


kpi/tests/api/v2/test_api_assets.py
```diff
@@ -1747,6 +1749,66 @@ def test_asset_redeployment(self):
...
+            0
+        ]
+        assert asset_response.data['version_id'] == deployed_version['uid']
+        # why does this fail?
+        '''
+        assert(
+            asset_response.data['deployed_version_id']
+            == deployed_version['uid']
+        )
+        '''
```
jnm, line 1776: @noliveleger, `asset_response.data['deployed_version_id']` is still `None` here :flushed:
image

jnm: OK, it's because `MockDeploymentBackend.connect()` does not set `version` in the `_deployment_data`:
https://github.com/kobotoolbox/kpi/blob/f7e99fde4e49dc3ec6afcdd3954229ea83b3b9ca/kpi/deployment_backends/mock_backend.py#L146-L156

Compare to `KobocatDeploymentBackend`:
https://github.com/kobotoolbox/kpi/blob/f7e99fde4e49dc3ec6afcdd3954229ea83b3b9ca/kpi/deployment_backends/kobocat_backend.py#L322-L328

kpi/serializers/v2/deployment.py
```diff
@@ -53,21 +53,10 @@ def update(self, instance, validated_data):
...
-            # Set the `active` flag without touching the rest of the deployment
-            deployment.set_active(active)
-            # If we (re)activate the asset, let's synchronize its media files
-            if active:
-                asset.sync_media_files_async(always=False)
+        self._raise_unless_current_version(asset, validated_data)
+        asset.deploy(
+            backend=deployment.backend,
+            active=validated_data.get('active', deployment.active)
+        )
```
noliveleger, line 60: Changing this behaviour updates `date_deployed` even when we archive a project. Is it really what we want? 
IMO, `date_deployed` should not change on archiving?

kpi/tests/api/v2/test_api_assets.py
```diff
@@ -1747,6 +1749,84 @@ def test_asset_redeployment(self):
...
+        assert (
+            deployed_version['date_modified']
+            == asset_response.data['deployed_versions']['results'][0][
+                'date_modified'
+            ]
+        ), (
+            'Redeploying the same version, unmodified, should not change the'
+            ' modification date of that version'
+        )
+
```
noliveleger, line 1829: If you agree with my comments above, I would also add a test to validate the `date_modified` stays the same on archiving.

kpi/serializers/v2/deployment.py
jnm: Good point, although I think changing `date_deployed` when (un)archiving is a lesser evil compared to *not* changing `date_deployed` when someone adds or removes a form media file and hits "REDEPLOY", which is the behavior without this PR.

I'll see about updating `date_deployed` only when the form version changes or when there are form media files to synchronize.

jnm: Simpler approach: when the PATCH request contains only `active` (and optionally `backend`), do not update `date_deployed` :ok_hand:

kpi/tests/api/v2/test_api_assets.py
jnm: Done!

PR #4525, opened 2023-07-04
Revert of "Toolchain: Warn about undefined CSS Module classes in TypeScript"

magicznyleszek:
## Description

Undo #4387 as it was causing errors.

## Related issues

Reverts #4387


magicznyleszek:
Revert wasn't needed, it was a mistake :)

PR #4529, opened 2023-07-06
Make project row name a normal link

### LINKED ISSUE #4524: Bug: project row needs to be a link
> Hi @**Leszek**, I don't think we can release the new view with the project links changed from `<a>` to `<span>`. That seems like a pretty serious usability regression to me. The browser does not recognize them as links and they cannot be control/command clicked or right clicked to be opened in new tabs, etc.

magicznyleszek:
## Description

This improves the UX of Custom Projects and My Projects views, by allowing users to interact with a regular HTML link when clicking on project name.

## Related issues

Fixes #4524


p2edwards (approved):
Looks good. I added a focus ring for keyboard support.

PR #4535, opened 2023-07-10
Typescriptize MainHeader component

magicznyleszek:
## Description

Internal code improvements.

## Code Review Notes

I had to simplify it a bit, but didn't want to go full refactor on it. I need this to be TS for some other work. Since it's not a small change, I'm not entirely sure I want it merged into `feature/my-projects`, but still want it reviewed. The plan would be to wait for the feature to be merged (soon 🤞), and then change base of this PR to the other feature branch I'm working on.


jsapp/js/assetUtils.ts
```diff
@@ -269,40 +270,40 @@ export function isAssetPublic(permissions?: Permission[]) {
  * For getting the icon class name for given asset type. Returned string always
  * contains two class names: base `k-icon` and respective CSS class name.
```
p2edwards, line 271: Please update this comment to match new behavior

jsapp/js/bemComponents.ts
```diff
@@ -154,12 +154,6 @@ bem.FormView__map = makeBem(bem.FormView, 'map');
 bem.FormView__mapButton = makeBem(bem.FormView, 'map-button');
 bem.FormView__mapList = makeBem(bem.FormView, 'map-list');
 
-bem.MainHeader = makeBem(null, 'main-header', 'header');
-bem.MainHeader__icon = makeBem(bem.MainHeader, 'icon', 'i');
-bem.MainHeader__title = makeBem(bem.MainHeader, 'title');
-bem.MainHeader__counter = makeBem(bem.MainHeader, 'counter');
-
-
```
p2edwards, line 162: Love seeing things disappear from this file

jsapp/js/components/header/mainHeader.component.tsx
```diff
@@ -99,7 +71,10 @@ const MainHeader = class MainHeader extends Reflux.Component {
   }
 
   renderGitRevInfo() {
-    if (sessionStore.currentAccount?.git_rev) {
+    if (
+      'git_rev' in sessionStore.currentAccount &&
+      sessionStore.currentAccount?.git_rev
```
p2edwards, line 76: Confused by this change. At least, if 'git_rev' is `in sessionStore.currentAccount`, then we shouldn't need `?.` here, right?

jsapp/js/assetUtils.ts
magicznyleszek: Updated the comment :)

jsapp/js/components/header/mainHeader.component.tsx
magicznyleszek: Right, I've changed it to check if both of the values we use for rendering are strings.

PR #4538, opened 2023-07-12
Improve fetch api handler to reject non-ok responses

magicznyleszek:
## Description

Make Update Password Form use more detailed error messages.

### DIFF
--- a/jsapp/js/api.ts
+++ b/jsapp/js/api.ts
@@ -25,7 +25,20 @@ const fetchData = async <T>(path: string, method = 'GET', data?: Json) => {
     headers,
     body: JSON.stringify(data),
   });
+
   const contentType = response.headers.get('content-type');
+
+  if (!response.ok) {
+    if (contentType && contentType.indexOf('application/json') !== -1) {
+      try {
+        return Promise.reject(await response.json());
+      } catch {
+        return Promise.reject(response);
+      }
+    }
+    return Promise.reject(response);
+  }
+
   if (contentType && contentType.indexOf('application/json') !== -1) {
     return (await response.json()) as Promise<T>;
   }
--- a/jsapp/js/api.ts
+++ b/jsapp/js/api.ts
@@ -36,15 +36,15 @@ const fetchData = async <T>(path: string, method = 'GET', data?: Json) => {
     };
 
     if (contentType && contentType.indexOf('application/json') !== -1) {
+      failResponse.responseText = await response.text();
       try {
-        failResponse.responseJSON = await response.json();
-        try {
-          failResponse.responseText = await response.text();
-        } finally {}
+        failResponse.responseJSON = JSON.parse(failResponse.responseText);
       } catch {
-        return Promise.reject(failResponse);
+        // If the response text is not a proper JSON, we simply don't add it to
+        // the rejection object.
       }
     }
+
     return Promise.reject(failResponse);
   }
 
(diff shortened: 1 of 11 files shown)


jsapp/js/api.ts
```diff
@@ -25,7 +25,42 @@ const fetchData = async <T>(path: string, method = 'GET', data?: Json) => {
...
+
+  // For server issues, we simply reject with no useful data. We expect the UI
+  // to not react with any notification or some other indicator.
+  if (
+    response.status === 401 ||
+    response.status === 403 ||
+    response.status === 404 ||
+    response.status >= 500
+  ) {
+    notify(t('Server error'));
```
p2edwards, line 39:
```suggestion
    notify(t('Server error'), 'error');
```

jsapp/js/api.ts
```diff
@@ -25,7 +33,54 @@ const fetchData = async <T>(path: string, method = 'GET', data?: Json) => {
...
+    ) {
+      let errorMessage = t('An error occurred');
+      errorMessage += ' ';
+      errorMessage += response.status;
+      errorMessage += ' ';
+      errorMessage += response.statusText;
+
+      // We only want to display the server issues notification once, even with
+      // multiple calls failing
+      notifyServerErrorThrottled(errorMessage);
```
p2edwards, line 63: I want to hold off on this for now—I thought about this some more and I don't throttling these is a good idea.

Two days ago, when we had a [double submit issue](https://github.com/kobotoolbox/kpi/pull/4538/commits/b7e6e3ee473bd27b5a27b8b72f406c289aae6d2d), the first indication was that there were 2 identical toast notifications.

Seeing lots of errors when one thing is broken isn't fun, but discarding errors without showing them at all is usually worse — particularly if the errors aren't duplicates and some may be more informative than others.

Related: Once toast notifications disappear, you can't review them at all. At least for our own debugging purposes, I want to start logging them to the browser console.

jsapp/js/components/languages/languagesListStore.ts
```diff
@@ -56,7 +56,7 @@ export default class LanguagesListStore {
 
   private onAnyFail(response: FailResponse) {
     this.isLoading = false;
-    notify(response.responseText, 'error');
+    notify(response.responseText || t('Unknown error'), 'error');
```
p2edwards, line 59: If we have the status code for these, I think I want to add them to the message like you did with the generic handler, because it could help in situations where someone takes a screenshot and sends it to the support team.

p2edwards: Could we use utils.ts `handleApiFail(FailResponse)` here?

p2edwards (approved):
I made a few changes:

- Tried to standardize around 'An error occurred' + status codes, rather than 'Unknown error'
- Started logging notify() calls to browser console, so when these things disappear we can still read them
- Turned off error notification throttling for now — I feel like in the near future this would do more harm than good.

---

Not super specific to this PR, but it came up during testing and was a fairly easy fix:

- Refined behavior of 'Query too short' message
   - these are batched by id so you no longer get a whole stack of these as you type a query
   - and it goes away when a successful query result comes back.
   - Also changed the icon from the 'task failed successfully' green checkmark to no icon.
   - If the assets fail to list for some other reason, it's a yellow warning box.

---

We're not catching every kind of error consistently, and I think the code organization could use some improvement. But let's merge this because it fixes the password update problem.

PR #4540, opened 2023-07-14
Add Custom Password Guidance text

### LINKED ISSUE #4468: Custom guidance text configuration
We will display it in UI near all the password forms. This allows superusers to instruct users about the password requirements. It should be a simple single line string.

It might be a good idea to be up front in the admin, with some instruction for that guidance text config, that this guidance text must match the rules that are configured. If guidance text doesn't match the validators

JacquelineMorrissette:
## Checklist

3. [x] Ensure the tests pass

## Description

Add custom guidance text to be set by superusers in Constance settings for password complexity

## Related issues

closes #4468


kobo/settings/base.py
```diff
@@ -438,7 +464,7 @@
         {'widget': 'django.forms.Textarea'},
     ],
     'mfa_help_text_fields_jsonschema': [
-        'kpi.fields.jsonschema_form_field.MfaHelpTextField',
+        'kpi.fields.jsonschema_form_field.CustomTextWithTranslations',
         {'widget': 'django.forms.Textarea'},
     ],
```
noliveleger, line 469: Good idea to rename `MfaHelpTextField'`. 👏 ❤️ 

You can get rid of this block and use on the one you added for `custom_password_guidance_text_field`. 
But please rename `custom_password_guidance_text_field` to `custom_text_with_translations`. Then use it for both MFA and Custom password guidance.

kobo/settings/base.py
```diff
@@ -422,9 +422,35 @@
         'Enable most recent password validation which will prevent the user from '
         'reusing the most recent password.',
     ),
+    'ENABLE_CUSTOM_PASSWORD_GUIDANCE_TEXT': (
+        False,
+        'Enable custom password guidance text to help users create their passwords.',
+    ),
+    'CUSTOM_PASSWORD_GUIDANCE_TEXT': (
+        json.dumps({
+            'default': '',
```
noliveleger, line 431: Because its appearance is controlled by `ENABLE_CUSTOM_PASSWORD_GUIDANCE_TEXT`, let's have a real default based on default rules.
While #4560 is not merged, we should follow the same technic used in #4549 and make the default string translatable, i.e:
- Add the translatable string to `static_list.py`
- Copy the same string in `base.py` - without `t()`
- Create the unit tests to ensure strings match (see `test_mfa_help_text_default_translation` in #4549).
- Use the constant from `static_list.py` in environment endpoint and its unit tests

noliveleger (changes requested):

PR #4541, opened 2023-07-14
Use `python3` for `npm copy-fonts` command

jnm:
Some environments do not automatically link `python` to `python3`, but all environments with Python 3 installed _should_ have `python3` present--hopefully!

@magicznyleszek can you check please if this causes any problems with your environment and merge if all is okay?


magicznyleszek:
WOMM :)

PR #4548, opened 2023-07-22
Resolve conflicts and api changes from react-router-6-13 branch

p2edwards:
# Description

Integrate changes from the react-router 6.14 upgrade into the `feature/my-projects` branch.

## Details

Since `feature/my-projects` is long-lived and contains its own modifications to routing, the merge with beta after #4505 has to be executed carefully.

## Notes for reviewer

I'd like @magicznyleszek to review this merge-conflict-resolution as the author of `feature/my-projects`.

The "Files changed" view will bury the most important manual merge decisions amongst large diffs that come directly from #4505. To try to create a better view, and highlight some of the important changes, I created https://github.com/kobotoolbox/kpi/compare/TMP-naive-merge-for-comparison...merge-my-projects-react-router-6-14. (It's not perfect but it's better than the default view.)

I'm reasonably confident in the quality of this merge, but it's possible there's something subtle that's missed. Let's:

- go over the changes and spot-check here & now …
- … and keep our eyes open after merging & do another QA pass before `feature/my-projects` lands in `beta`.

[link to internal chat thread](https://chat.kobotoolbox.org/#narrow/stream/4-Kobo-Dev/topic/My.20Projects.20redo/near/249946)

## Related issues

Caused by #4505 — Upgrade to React Router 6.14 and use upstream usePrompt
Needed for #4535 — Typescriptize MainHeader component


jsapp/js/components/processing/singleProcessingStore.ts
```diff
@@ -171,7 +169,7 @@ class SingleProcessingStore extends Reflux.Store {
   init() {
     this.resetProcessingData();
 
-    history.listen(this.onRouteChange.bind(this));
+    setTimeout(() => router!.subscribe(this.onRouteChange.bind(this)));
```
magicznyleszek, line 172: :eyes:

jsapp/js/mixins.tsx
```diff
@@ -126,367 +119,419 @@ interface MixinsObject {
...
-  _forEachDroppedFile(params: CreateImportRequest = {}) {
-    const totalFiles = params.totalFiles || 1;
+    _forEachDroppedFile(params: CreateImportRequest = {}) {
+      const totalFiles = params.totalFiles || 1;
 
-    const router = this.props.router;
-    const isProjectReplaceInForm = (
-      this.props.context === PROJECT_SETTINGS_CONTEXTS.REPLACE
-      && routerIsActive('forms')
-      && router.params.uid !== undefined
```
magicznyleszek, line 337: I wonder if instead of creating this one off const, we should just move the logic to the place that is using it.

I am thinking about this because I've noticed that the uid check is no more here, and I see it wasn't needed, but also the whole const seems to be not needed :P

PR #4549, opened 2023-07-22
Allow default MFA help text to be translated along with the rest of the application's translatable strings

jnm:


kobo/settings/base.py
```diff
@@ -235,6 +235,9 @@
     'MFA_LOCALIZED_HELP_TEXT': (
         json.dumps({
             'default': (
+                # It's terrible, but if you change this, you must also change
+                # `MFA_DEFAULT_HELP_TEXT` in `static_lists.py` so that
+                # translators receive the new string
```
jnm, line 240: Perhaps a more sane approach would be to have superusers leave the default value empty when they do not want to override it. However, doing that would take away our ability to use the default string to show an instructional example. Copying and pasting the string seemed like the lesser evil :upside_down_face:

kpi/tests/api/test_api_environment.py
```diff
@@ -33,41 +35,44 @@ def setUp(self):
...
             'mfa_enabled': constance.config.MFA_ENABLED,
-            'mfa_localized_help_text': lambda i18n_texts: {
-                lang: markdown(text)
-                for lang, text in json.loads(
-                    constance.config.MFA_LOCALIZED_HELP_TEXT.replace(
-                        '##support email##',
-                        constance.config.SUPPORT_EMAIL
-                    )
-                ).items()
-            },
```
jnm, line 67: Oops, it was impossible for this to fail. Always make sure your tests can fail! I'll take some blame here, too, because making values automatically get `assertEqual()`ed while requiring callables to do their own assertions leaves some clarity to be desired.

kpi/tests/api/test_api_environment.py
```diff
@@ -33,41 +35,44 @@ def setUp(self):
             'community_url': constance.config.COMMUNITY_URL,
             'frontend_min_retry_time': constance.config.FRONTEND_MIN_RETRY_TIME,
             'frontend_max_retry_time': constance.config.FRONTEND_MAX_RETRY_TIME,
-            'project_metadata_fields': lambda x: \
```
jnm, line 36: Sorry about the quantity of formatting changes. The smallest unit of code that I could invoke python-black on was the dictionary.

kpi/tests/api/test_api_environment.py
```diff
@@ -156,7 +172,7 @@ def test_mfa_per_user_availability_while_globally_enabled_as_anonymous(self):
     def test_social_apps(self):
         # GET mutates state, call it first to test num queries later
         self.client.get(self.url, format='json')
-        queries = 19
+        queries = 20
```
jnm, line 175: This query is being executed twice for some reason:
```sql
SELECT "constance_config"."id", "constance_config"."key", "constance_config"."value" FROM "constance_config" WHERE "constance_config"."key" = 'MFA_LOCALIZED_HELP_TEXT' LIMIT 21
```
I had to give up caring about this for now. I have another branch that I'm about to PR anyway that reduces (or at least was reducing) the number of queries from 19 to 18, so… :shrug:

jnm: https://github.com/kobotoolbox/kpi/pull/4550/files#r1271244110

kpi/tests/api/test_api_environment.py
noliveleger: Who am I to complain about receiving few lines of new formatting changes in a review? 😇

PR #4550, opened 2023-07-22
Clean up the environment (API endpoint)

jnm:
Everyone should do it!

The environment endpoint was [originally designed](https://github.com/kobotoolbox/kpi/commit/5d9d1a76660c9cc96ef0edc817db54db8e2a4b91#diff-30cf9a36a5430764f922f8c83031d70189659a16196c33571358ccf1a309aefc) to copy settings from django-constance into a JSON response with a minimum of extra processing. It had since grown to be unwieldy with the inclusion of many different kinds of data from different sources. This is an attempt to clean that up and make it more manageable.

### DIFF
--- a/kpi/tests/api/test_api_environment.py
+++ b/kpi/tests/api/test_api_environment.py
@@ -172,7 +172,7 @@ class EnvironmentTests(BaseTestCase):
     def test_social_apps(self):
         # GET mutates state, call it first to test num queries later
         self.client.get(self.url, format='json')
-        queries = 20
+        queries = 18
         with self.assertNumQueries(queries):
             response = self.client.get(self.url, format='json')
         self.assertEqual(response.status_code, status.HTTP_200_OK)
--- a/kpi/views/environment.py
+++ b/kpi/views/environment.py
@@ -34,7 +34,7 @@ class EnvironmentView(APIView):
     GET-only view for certain server-provided configuration data
     """
 
-    CONFIGS_TO_EXPOSE = [
+    SIMPLE_CONFIGS = [
         'TERMS_OF_SERVICE_URL',
         'PRIVACY_POLICY_URL',
         'SOURCE_CODE_URL',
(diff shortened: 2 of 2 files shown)


kpi/tests/api/test_api_environment.py
```diff
@@ -172,7 +172,7 @@ def test_mfa_per_user_availability_while_globally_enabled_as_anonymous(self):
     def test_social_apps(self):
         # GET mutates state, call it first to test num queries later
         self.client.get(self.url, format='json')
-        queries = 20
+        queries = 18
```
jnm, line 175: image

kpi/views/environment.py
```diff
@@ -43,38 +43,82 @@ class EnvironmentView(APIView):
...
+            key.lower(): getattr(constance.config, key)
+            for key in cls.SIMPLE_CONFIGS
+        }
+
+    JSON_CONFIGS = [
+        'FREE_TIER_DISPLAY',
+        'FREE_TIER_THRESHOLDS',
+        'PROJECT_METADATA_FIELDS',
+        'USER_METADATA_FIELDS',
+    ]
```
noliveleger, line 60: I would have grouped "static" variables together at the top of the class and grouped the methods below that. 
Just my own preference.

noliveleger (approved):

kpi/views/environment.py
jnm: Thanks, but no thanks! These fields are grouped intentionally next to the method that operates on them. @noliveleger

kpi/tests/api/test_api_environment.py
```diff
@@ -172,7 +172,7 @@ def test_mfa_per_user_availability_while_globally_enabled_as_anonymous(self):
...
         with self.assertNumQueries(queries):
             response = self.client.get(self.url, format='json')
```
jnm, line 177: Can one review a merged PR? :thinking:

jnm (commented):
Example review of an already-merged PR

PR #4552, opened 2023-07-24
Remove unused variable report__base

bufke:
report__base was used in FORM_REPORT_OLD which was already removed. The variable is set but never read. This change removes it.

### DIFF
--- a/jsapp/js/components/formSubScreens.js
+++ b/jsapp/js/components/formSubScreens.js
@@ -55,13 +55,11 @@ export class FormSubScreens extends React.Component {
     }
 
     var iframeUrl = '';
-    var report__base = '';
     var deployment__identifier = '';
 
     if (this.state.uid != undefined) {
       if (this.state.deployment__identifier != undefined) {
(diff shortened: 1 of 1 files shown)


jnm (approved):

jsapp/js/components/formSubScreens.js
```diff
@@ -55,13 +55,11 @@ export class FormSubScreens extends React.Component {
...
     var iframeUrl = '';
-    var report__base = '';
     var deployment__identifier = '';
 
     if (this.state.uid != undefined) {
       if (this.state.deployment__identifier != undefined) {
         deployment__identifier = this.state.deployment__identifier;
-        report__base = deployment__identifier.replace('/forms/', '/reports/');
       }
       switch (this.props.router.location.pathname) {
```
p2edwards, line 64: I would have suggested removing `var deployment__identifier = ''` and the whole `if (…deployment__identifier…) { … }` block, as it appears to be unused.

PR #4553, opened 2023-07-24
Upgrade to node 18 and latest npm version

bufke:
- Upgrade to Node 18
- Set explicit legacy peer deps
- Use >= instead of specific version
- Use npm engine-strict instead of other methods like .node_version

## Developer impact

Running `npm i` with a semi up to date node version is all that is needed.


.gitlab-ci.yml
```diff
@@ -21,7 +21,7 @@ build:
 deploy-beta:
   stage: deploy
   image:
-    name: alpine/helm:3.12.0
+    name: alpine/helm:3.12.2
```
bufke, line 24: yeah yeah it's unrelated but so small

PR #4354, opened 2023-03-14
(tmp) github actions testing

p2edwards:
This is a multipurpose draft/temp PR for testing changes for github-actions.

The main purpose is to iterate on issues like #4337, without cluttering other branches or PR's with lots of small commits that are mainly just for triggering slightly different test runs. Any results found here will probably be ported into other PRs rather than merging this one.

PR #4554, opened 2023-07-25
(tmp) node 18 github actions testing

p2edwards:
The chrome tests `npm test` in the `node-18` branch are failing.

To iteratively troubleshoot without cluttering the commit history too much, I've spun this out to a draft PR for testing. Similar to #4354 (tmp) github actions testing.

---

Cache key safety, keep useful tooling

- Add node version and .npmrc to the cache key in CI. When these are modified in any branch, we definitely shouldn't reuse the same cache packages, as it could cause CI to fail in unrelated runs.
- Restore .node-version and .nvmrc. They are useful tooling for devs who work on multiple node projects.
  - Alternatively, we could add them to .gitignore. Yet, I'd prefer to we keep a centralized config that is known to work and matches what we run in CI. It's not hard to maintain these, and we could upgrade them on a regular schedule.
- Remove --legacy-peer-deps from npm install command in CI, as it should be already applied by .npmrc.
  - We're not asking devs to add that flag manually from now on, nor is it in the Dockerfile. It adds confusion to put it again here.

---

- [ ] Figure out what's going on with mocha-chrome
- [ ] Audit the update-translation-strings action, too. Wouldn't want that to break later.


.github/workflows/npm-test.yml
```diff
@@ -23,10 +26,10 @@ jobs:
           cache-name: cache-node-modules
         with:
           path: node_modules
-          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json', 'patches/**/*.patch') }}
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-node-v${{ env.NODE_VERSION }}-${{ hashFiles('**/package-lock.json', 'patches/**/*.patch', '.npmrc') }}
       - name: Install JavaScript dependencies (npm install)
         if: steps.cache-nodemodules.outputs.cache-hit != 'true'
-        run: npm install --legacy-peer-deps
+        run: npm install
```
bufke, line 32: :+1:

.node-version
```diff
@@ -0,0 +1 @@
+v18.17.0
```
bufke, line 1: My issue with these are that we set the version in many places. Which means soon we'll have many places that disagree. The patch version needs updated constantly, so it's laborious to maintain. What behavior are you hoping to get out of these for development? Can it be done with less places to update and ideally set to something more like ^18.

PR #4559, opened 2023-07-28
Hide meta/rootUuid from table view

jnm:
## Description
This field is not ready for general use yet, but curious people have [discovered](https://community.kobotoolbox.org/t/new-column-in-table-view-meta-rootuuid/43454) it.

When (if) it comes back, it should be positioned at the end of the table.


srartese (approved):
Tested locally and the filter is not showing in table view fields

PR #4560, opened 2023-07-28
Support Django Constance settings with translatable strings 

noliveleger:
## Description

Add a wrapper class to be used python objects such as `dict` or `list` with translatable strings

### DIFF
--- a/kobo/apps/constance_backends/utils.py
+++ b/kobo/apps/constance_backends/utils.py
@@ -1,0 +1,23 @@
+from __future__ import annotations
+
+import json
+
+from typing import Union
+from kpi.utils.json import LazyJSONSerializable
+
+
+def to_python_object(
+    object_or_str: Union['LazyJSONSerializable', str]
+) -> Union[list, dict]:
+    """
+    Returns a python object from `object_or_str`.
+    """
+    # If a Constance setting is a `LazyJSONSerializable` object,
+    # `constance.config.SETTING` does not return the same type of object depending
+    # on whether the setting was saved in its back end or not. From the back end,
+    # it returns the string representation. Otherwise, it returns the
+    # `LazyJSONSerializable` object itself.
+    if isinstance(object_or_str, LazyJSONSerializable):
+        return object_or_str.object
+
+    return json.loads(object_or_str)
--- a/kobo/apps/constance_backends/utils.py
+++ b/kobo/apps/constance_backends/utils.py
@@ -12,11 +12,8 @@ def to_python_object(
     """
     Returns a python object from `object_or_str`.
     """
-    # If a Constance setting is a `LazyJSONSerializable` object,
-    # `constance.config.SETTING` does not return the same type of object depending
-    # on whether the setting was saved in its back end or not. From the back end,
-    # it returns the string representation. Otherwise, it returns the
-    # `LazyJSONSerializable` object itself.
+    # Make constance.config.SETTING return a consistent value when SETTING is a
+    # LazyJSONSerializable object
     if isinstance(object_or_str, LazyJSONSerializable):
         return object_or_str.object
 
--- a/kobo/settings/base.py
+++ b/kobo/settings/base.py
@@ -333,7 +333,7 @@ CONSTANCE_CONFIG = {
         'minutes of transcription, '
         'number of translation characters',
         # Use custom field for schema validation
-        'free_tier_threshold_jsonschema'
+        'free_tier_threshold_jsonschema',
     ),
     'FREE_TIER_DISPLAY': (
         LazyJSONSerializable({
--- a/kobo/settings/base.py
+++ b/kobo/settings/base.py
@@ -1,5 +1,4 @@
 # coding: utf-8
-import json
 import logging
 import os
 import re
@@ -16,6 +15,7 @@ from django.urls import reverse_lazy
 from django.utils.translation import get_language_info, gettext_lazy as t
 from pymongo import MongoClient
 
+from kpi.utils.json import LazyJSONSerializable
 from ..static_lists import EXTRA_LANG_INFO, SECTOR_CHOICE_DEFAULTS
 
 env = environ.Env()
@@ -233,11 +233,8 @@ CONSTANCE_CONFIG = {
         'Enable two-factor authentication'
     ),
     'MFA_LOCALIZED_HELP_TEXT': (
-        json.dumps({
-            'default': (
-                # It's terrible, but if you change this, you must also change
-                # `MFA_DEFAULT_HELP_TEXT` in `static_lists.py` so that
-                # translators receive the new string
+        LazyJSONSerializable({
+            'default': t(
                 'If you cannot access your authenticator app, please enter one '
                 'of your backup codes instead. If you cannot access those '
                 'either, then you will need to request assistance by '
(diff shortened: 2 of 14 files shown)


kpi/tests/api/test_api_environment.py
```diff
@@ -109,14 +111,19 @@ def test_template_context_processor(self):
...
+
         with mock.patch(
             'hub.utils.i18n.t', return_value=MOCK_TRANSLATION_STRING
         ) as mock_t:
             response = self.client.get(self.url, format='json')
             assert response.json()['mfa_localized_help_text'] == markdown(
                 MOCK_TRANSLATION_STRING
             )
-            assert mock_t.call_args.args[0] == MFA_DEFAULT_HELP_TEXT
+            assert mock_t.call_args.args[0] == default_translation._proxy____args[0]
```
jnm, line 126: looks scary! can we coerce `default_translation` to a `str` to achieve the same?

kpi/utils/json.py
```diff
@@ -0,0 +1,42 @@
...
+    Wrapper for python objects (such as lists or dicts) which contain lazy gettext
+    objects and could not be interpreted when the application loads.
+
+    It can be JSON serialized.
+    """
+    def __init__(self, o):
+        self.object = o
+
+    def __repr__(self):
+        return f'<LazyJSONSerializable object={type(self.object)}>'
```
jnm, line 28: just curious about the decision to hide the content here. i see the object returned by `gettext_lazy()` has this:
```
        def __repr__(self):
            return repr(self.__cast())
```
i don't know what `__cast()` does, but in effect you do get the value back out when calling `repr()`:
```
In [7]: thingy = gettext_lazy('stringy')

In [8]: type(thingy)
Out[8]: django.utils.functional.lazy.<locals>.__proxy__

In [9]: repr(thingy)
Out[9]: "'stringy'"
```

static/admin/constance/css/kobo-constance.css
```diff
@@ -0,0 +1,5 @@
+#content-main.constance #changelist-form #results tr td:nth-child(2) p {
+  display: block;
+  unicode-bidi: embed;
+  white-space: break-spaces;
+}
```
jnm, line 5: note to self: do a visual before/after comparison

jnm (changes requested):

static/admin/constance/css/kobo-constance.css
jnm: looks good :+1:

kobo/apps/constance_backends/utils.py
```diff
@@ -0,0 +1,23 @@
...
+    object_or_str: Union['LazyJSONSerializable', str]
+) -> Union[list, dict]:
+    """
+    Returns a python object from `object_or_str`.
+    """
+    # If a Constance setting is a `LazyJSONSerializable` object,
+    # `constance.config.SETTING` does not return the same type of object depending
+    # on whether the setting was saved in its back end or not. From the back end,
+    # it returns the string representation. Otherwise, it returns the
+    # `LazyJSONSerializable` object itself.
```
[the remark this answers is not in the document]
noliveleger: Short answer: yes.


Long answer with more details to what I tried to explain in my comments. 

Actually, it is saved in [PostgreSQL then cached in Redis](https://github.com/kobotoolbox/kpi/blob/1c8a08fae6a2cff20019d08c56197d478bf1b31e/kobo/settings/base.py#L516-L518) after the first access of the variable. 
 
So when we call `constance.config.MY_ENV`, `constance`:
- gets [the default value](https://github.com/jazzband/django-constance/blob/master/constance/base.py#L16)
- try to get a [value from the back end](https://github.com/jazzband/django-constance/blob/master/constance/base.py#L19-L20), but if it does not exist, falls back on the default value (i.e. `LazyJSONSerializable` object).

After the first call of `constance.config.MY_ENV`, `constance` saves a serialized version  of `MY_ENV` in the back end and cached it Django CACHE (if enabled obviously). So in production, we are always getting a `str` even if its default has not been overridden.  In unit tests, PostgreSQL is always empty because tables are wiped after each test and Django CACHE is disabled.

kpi/tests/api/test_api_environment.py
noliveleger: I wanted to be sure the behaviour of this test some the same as before. If we coerce `default_translation` to as `str`, it does not give the exact result. 

```python 
In [1]: from django.utils.translation import activate, gettext_lazy as t

In [2]: activate('fr')

In [3]: s = t('Create an account')

In [4]: str(s)
Out[4]: 'Ouvrir un compte'

In [5]: s._proxy____args[0]
Out[6]: 'Create an account'
```

kobo/settings/base.py
```diff
@@ -335,15 +333,13 @@
...
-        json.dumps(
-            {
-                'name': None,
-                'feature_list': [],
-            }
-        ),
+        LazyJSONSerializable({
+            'name': None,
+            'feature_list': [],
+        }),
```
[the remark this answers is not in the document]
noliveleger: 🤔 Hmm. You're totally right. Still need to add a one line of code to handle the first access to the variable though [link to be provided].

kpi/migrations/0051_set_free_tier_thresholds_to_default.py
```diff
@@ -12,7 +14,7 @@ def reset_free_tier_thresholds(apps, schema_editor):
         'transcription_minutes': None,
         'translation_chars': None,
     }
-    setattr(config, 'FREE_TIER_THRESHOLDS', json.dumps(thresholds))
+    setattr(config, 'FREE_TIER_THRESHOLDS', LazyJSONSerializable(thresholds))
```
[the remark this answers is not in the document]
noliveleger: Gone with last commit.

kobo/settings/base.py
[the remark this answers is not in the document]
noliveleger: You are right. I've just committed some changes to remove `LazyJSONSerializable` in `base.py` where it was not needed. 
Then I realized it was because of the formatting in Django Admin. If we want to use `dict` or `list` directly, their string representation does not keep the formatting that `LazyJSONSerializable` provides (without overriding Constance templates) :-(

kpi/utils/json.py
noliveleger: During my tests, I got an `AppRegistryNotReady` error but I cannot reproduce it anymore. I'll return `__str__()` like it was at the very beginning.

kobo/settings/base.py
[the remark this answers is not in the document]
jnm: No problem :+1: Let's definitely keep `LazyJSONSerializable` if it improves formatting

kpi/tests/api/test_api_environment.py
jnm: But if you don't `activate('fr')`, then `str(default_translation)` would work? I changed it locally and the test passes.

However, the whole point might be moot. Do we even need to examine what was passed as the argument to `t()` anymore now that you've gotten away from my split-brain thing of duplicating the string in `static_lists.py`?

noliveleger: > str(default_translation) would work?

Indeed 

> Do we even need to examine what was passed as the argument to t()

Well, we may not need to test anymore.

kobo/apps/constance_backends/utils.py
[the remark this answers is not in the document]
jnm: Thanks, very educational! I don't think we're going to explain all this in the comment (unless you think we should :thinking:) and current comment as-is leaves me wanting more detail to understand what it means.

Maybe we should write a simpler comment such as:
> Make `constance.config.SETTING` return a consistent value when `SETTING` is a `LazyJSONSerializable` object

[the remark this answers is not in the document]
noliveleger: No, straight to the point ;-) Love it

jnm (approved):

PR #4561, opened 2023-07-28
Usage limits 2fa fixes

LMNTL:
## Checklist

3. [X] Ensure the tests pass

## Description

Small changes to how MFA configuration is sent to the browser.

## Notes

The frontend environment variable `mfa_allowed` has been split into two values: `mfa_allowed`, which tracks whether MFA is enabled on the server level, and `mfa_available_to_user`, which tracks whether the user's MFA per-user availability.

### DIFF
--- a/kpi/views/environment.py
+++ b/kpi/views/environment.py
@@ -13,7 +13,6 @@ from allauth.socialaccount.models import SocialApp
 from kobo.static_lists import COUNTRIES
 from kobo.apps.hook.constants import SUBMISSION_PLACEHOLDER
 from kobo.apps.accounts.mfa.models import MfaAvailableToUser
-from kobo.apps.accounts.utils import user_has_paid_subscription
 from kpi.utils.object_permission import get_database_user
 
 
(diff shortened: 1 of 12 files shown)


kpi/views/environment.py
```diff
@@ -3,16 +3,16 @@
...
 from markdown import markdown
 from rest_framework.response import Response
 from rest_framework.views import APIView
-from allauth.socialaccount.models import SocialApp
 
-from kobo.static_lists import COUNTRIES
-from kobo.apps.hook.constants import SUBMISSION_PLACEHOLDER
 from kobo.apps.accounts.mfa.models import MfaAvailableToUser
+from kobo.apps.hook.constants import SUBMISSION_PLACEHOLDER
+from kobo.static_lists import COUNTRIES
```
noliveleger, line 15: ❤️

kpi/views/environment.py
```diff
@@ -141,4 +129,8 @@ def get(self, request, *args, **kwargs):
             settings.STRIPE_PUBLIC_KEY if settings.STRIPE_ENABLED else None
         )
         data['social_apps'] = social_apps
+        data['mfa_available_to_user'] = MfaAvailableToUser.objects.filter(
```
noliveleger, line 132: This name `mfa_available_to_user` gave me hard time. Maybe because of my memory of a golden-fish, but I always have to look at the code why it returns `false` if the the per-user availability list is empty. 
At the first look, it's not obvious that it tells if user is part of the list or not. 

Do you think we could find something more obvious. (I don't have a good idea).

noliveleger (changes requested):

kpi/views/environment.py
LMNTL: Good point! I didn't think to update that after changing what was getting returned from `/environment`. I went with `mfa_per_user_availability`, like the admin setting, let me know if that's clearer.

PR #4562, opened 2023-07-31
Remove raven package, add sentry npm package.

bufke:
Update Sentry JS SDK to latest (from Raven). There should be no end user impact and is not a breaking change because it still falls back to RAVEN_JS_DSN. This should fix CSP reporting. I tested:

- An exception in the react app
- An exception that was blocked by the sdk itself in main.es6. I don't know why this happens but it seems to be default and expected behavior. Possibly [related](https://stackoverflow.com/a/76579210/443457)
- Using the legacy window.Raven variable via both `window.Raven` and `Raven`
- Added a script tag to code.jquery.com and ensured CSP was sent successfully

I chose to rename most raven variables to sentry, with the exception of the coffeescript files.


srartese (commented):
Code looks good although I am not the best for Python review. I tested the site itself and all is working as normal. I'd love to chat on how you triggered the exceptions and where that data is being stored.

srartese (approved):

p2edwards (approved):

p2edwards (approved):

PR #4563, opened 2023-08-01
Add password guidance text to django templates

JacquelineMorrissette:
## Checklist

1. [ ] If you've added code that should be tested, add tests
2. [x] If you've changed APIs, update (or create!) the documentation
3. [x] Ensure the tests pass
4. [x] Make sure that your code lints and that you've followed [our coding style](https://github.com/kobotoolbox/kpi/blob/master/CONTRIBUTING.md)
5. [x] Write a description of your work suitable for publishing on [our forum](https://community.kobotoolbox.org/tag/release-notes)
6. [ ] Mention any related issues in this repository (as #ISSUE) and in other repositories (as kobotoolbox/other#ISSUE)
7. [ ] Open an issue in the [docs](https://github.com/kobotoolbox/docs/issues/new) if there are UI/UX changes

## Description

Add the `CUSTOM_PASSWORD_GUIDANCE_TEXT` set in Constance to the Password Reset page and Sign Up Page

PR #4574, opened 2023-08-08
Display password guidance text in Update Password Form

### LINKED ISSUE #4469: Display guidance text
Display guidance text in UI in all password forms. 

Requires #4468
~~Needs design from @Ig-Rebollo~~

magicznyleszek:
## Notes

This is the React app part of adding guidance text. There is also place for guidance text in the Django HTML templates - this will be done in other PR (see #4563).

To test it, go to `/admin/constance/config/` and turn on proper flag.

## Related issues

Fixes #4469


jsapp/js/account/security/password/updatePasswordForm.component.tsx
```diff
@@ -93,6 +95,15 @@ export default function UpdatePasswordForm() {
 
   return (
     <form className={styles.root} onSubmit={submitPasswordForm}>
+      {envStore.data.enable_custom_password_guidance_text && (
```
LMNTL, line 98: If the `envStore` isn't ready the first time this renders, the guidance text doesn't show up until the user starts typing (triggering a state change). Could be fixed by checking if `envStore.isReady` is true.

LMNTL (changes requested):

jsapp/js/account/security/password/updatePasswordForm.component.tsx
magicznyleszek: I pushed a fix that uses `useEffect` :)

magicznyleszek: @LMNTL that is educational, thanks :) Pushed a final fix 😉

PR #4582, opened 2023-08-11
Update CSP settings for GA4

### LINKED ISSUE #4567: Update CSP to allow googletagmanager.com
CSP needs to be updated to work with #4322. Please assign David as the reviewer.

https://chat.kobotoolbox.org/#narrow/stream/4-Kobo-Dev/topic/CORB.20blocking.20CSP/near/251151
```
kf.kobotoolbox.org/:1 Refused to load the script 'https://www.googletagmanager.com/gtag/js?id=UA-26003863-7' because it violates the following Content Security Policy directive: "script-src 'self' https://kc.kobotoolbox.org https://ee.kobotoolbox.org *.google-analytics.com *.google-analytics.com https://cdn.ravenjs.com https://sentry.kbtdev.org https://sentry.kbtdev.org https://js.stripe.com https://js.stripe.com". Note that 'script-src-elem' was not explicitly set, so 'script-src' is used as a fallback.
```

magicznyleszek:
## Related issues

Fixes #4567

### DIFF
--- a/kobo/settings/base.py
+++ b/kobo/settings/base.py
@@ -726,10 +726,10 @@ CSP_IMG_SRC = CSP_DEFAULT_SRC + [
 CSP_FRAME_SRC = CSP_DEFAULT_SRC
 
 if GOOGLE_ANALYTICS_TOKEN:
-    google_domain = '*.googletagmanager.com'
-    CSP_SCRIPT_SRC.append(google_domain)
-    CSP_CONNECT_SRC.append(google_domain)
-    CSP_IMG_SRC.append(google_domain)
+    # Taken from https://developers.google.com/tag-platform/tag-manager/csp#google_analytics_4_google_analytics
+    CSP_SCRIPT_SRC.append('https://*.googletagmanager.com')
+    CSP_CONNECT_SRC.extend(['https://*.google-analytics.com', 'https://*.analytics.google.com', 'https://*.googletagmanager.com'])
+    CSP_IMG_SRC.extend(['https://*.google-analytics.com', 'https://*.googletagmanager.com'])
 if RAVEN_JS_DSN_URL and RAVEN_JS_DSN_URL.scheme:
     raven_js_url = RAVEN_JS_DSN_URL.scheme + '://' + RAVEN_JS_DSN_URL.hostname
     CSP_SCRIPT_SRC.append('https://cdn.ravenjs.com')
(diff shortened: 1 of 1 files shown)


bufke:
@magicznyleszek would you like me to test this or just review the change. I see that when we upgraded to GA4, we switched our deployment method to GTM. If you describe your testing process to me, I'd be happy to consider if it was sufficient. If I should test it, then we should discuss internally how I can go about doing so and what access I may need. Or I suppose I could sign up for my own account and test with, as that workflow seems intended.

Are you confident that the google-analytics.com domain no longer needs allowed since switching to GA4 and GTM? The [documentation](https://developers.google.com/tag-platform/tag-manager/csp) is saying otherwise. See section Google Analytics 4 (Google Analytics)

bufke (changes requested):
@magicznyleszek would you like me to test this or just review the change. I see that when we upgraded to GA4, we switched our deployment method to GTM. If you describe your testing process to me, I'd be happy to consider if it was sufficient. If I should test it, then we should discuss internally how I can go about doing so and what access I may need. Or I suppose I could sign up for my own account and test with, as that workflow seems intended.

Are you confident that the google-analytics.com domain no longer needs allowed since switching to GA4 and GTM? The [documentation](https://developers.google.com/tag-platform/tag-manager/csp) is saying otherwise. See section Google Analytics 4 (Google Analytics)

magicznyleszek:
@bufke I thought that since we are using GA4, and we are serving googletagmanager via script src, we only need this one thing:
Screenshot 2023-08-11 at 18 30 42

bufke:
Isn't that saying you need to allow more right under the circle? But let us not speculate. Did you test this out or, if it's urgent and you don't have time, you can ask me to.

kobo/settings/base.py
```diff
@@ -726,10 +726,10 @@ def __init__(self, *args, **kwargs):
 CSP_FRAME_SRC = CSP_DEFAULT_SRC
 
 if GOOGLE_ANALYTICS_TOKEN:
-    google_domain = '*.google-analytics.com'
-    CSP_SCRIPT_SRC.append(google_domain)
-    CSP_CONNECT_SRC.append(google_domain)
-    CSP_IMG_SRC.append(google_domain)
+    # Taken from https://developers.google.com/tag-platform/tag-manager/csp#google_analytics_4_google_analytics
```
bufke, line 729: Good idea

bufke (approved):
Approved pending testing

magicznyleszek:
Note for future: testing this locally went like this:
1. set GA code in `./run.py --setup`
2. enabled CSP locally
3. turned off all possible blockers
4. confirmed that updated `Content-Security-Policy` appears in headers
5. confirmed that google script is being loaded
6. confirmed that google analytics page was in fact gathering data

PR #4604, opened 2023-08-30
Add 'Unsaved changes' state

### LINKED ISSUE #4584: Block navigation if there is unsaved data in Analysis tab
Update the status text too:
- "Saved" should always mean that the latest backend response exactly matches the current frontend state.
- "Saving…" should always mean there's a pending frontend autosave awaiting a backend 200, which hasn't failed or returned yet.
- We can say "Changes made" if we're still in the debounce phase and haven't started a "Saving…" call yet.
- While we're in "Saving…" or "Changes made" state, we should block navigation with an unsaved changes prompt, including for up/down arrows to switch submissions, reroute through URL bar, or closing the tab.

duvld:
## Description
Unsaved changes done. WIP on navigation block

## Related issues
Fixes #4584


duvld:
Closed in favor of #4616 (rebased)

PR #4605, opened 2023-08-30
Return a 401 HTTP code when user is not authenticated

noliveleger:
## Description

Make the response more obvious to users when they are disconnected (and they have to reconnect) 

## Notes

By default, Django returns 403 even if the Python code explicitly returns a 401 with session authentication. 
See [this comment](https://github.com/kobotoolbox/kpi/pull/4605/commits/e76f70d483f1aaebce4c2bf0ffe2dbd7a396659d#diff-58d8fa84360c411219ab22a6b24e9523b9fe21387a01257e917142b7ebe0b1baR110-R120) for more details. 

401 Unauthorized
The HyperText Transfer Protocol (HTTP) 401 Unauthorized response status code indicates that the client request has not been completed because it lacks valid authentication credentials for the requested resource.


kpi/authentication.py
```diff
@@ -106,6 +106,24 @@ def prepare_response_with_csrf_cookie(request, response):
...
+    does never return 401's, because they cannot fill the WWW-Authenticate header
+    with a valid value in the 401 response. As a result, we cannot distinguish
+    calls that are not unauthorized (401 unauthorized) and calls for which the
+    user does not have permission (403 forbidden).
+    See https://github.com/encode/django-rest-framework/issues/5968#issuecomment-39935282
+
+    We do set authenticate_header function in SessionAuthentication, so that a
+    value for the WWW-Authenticate header can be retrieved and the response code
+    is automatically set to 401 in case of unauthenticated requests.
+    """
```
jnm, line 121: This needs a rewrite, but otherwise the PR LGTM

jnm (changes requested):

PR #4608, opened 2023-08-31
Update mass user email script

LMNTL:
## Checklist

3. [ ] Ensure the tests pass

## Description

Updates to the mass user emails script.

## Notes

Added `USER_FILTER` and `USER_EXCLUDE` so the body of the script no longer needs to be edited for non-complex queries.

These changes have been battle-tested by sending a large amount of emails from kpi public servers.

### DIFF
--- a/scripts/mass_user_emails.py
+++ b/scripts/mass_user_emails.py
@@ -46,8 +46,7 @@ ACTIVE_SINCE = timezone.now() - relativedelta(months=6)
 #     'organizations_organization__djstripe_customers__subscriptions': None,
 #     'id': 1,
 # }
-USER_FILTERS = {
-}
+USER_FILTERS = {}
 
 
 # An array of dicts containing additional queries on the User.
@@ -62,11 +61,7 @@ USER_FILTERS = {
 #         'id': 5
 #     }
 # ]
-USER_EXCLUDE = [
-    {
-        'organizations_organization__djstripe_customers__subscriptions': None,
-    }
-]
+USER_EXCLUDE = [{}]
 
 start_time = time.time()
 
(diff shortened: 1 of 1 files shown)


scripts/mass_user_emails.py
```diff
@@ -129,17 +168,23 @@ def run(*args):
 
     for user in eligible_users.iterator(chunk_size=500):
         for attempts in range(MAX_SEND_ATTEMPTS):
-            response = send_email(
-                ses, user.email, configuration=configuration_set
-            )
-            status = response['ResponseMetadata']['HTTPStatusCode']
+            try:
+                response = send_email(
+                    ses, user.email, configuration=configuration_set
```
jnm, line 173: `user.email_cleaned` is available here thanks to your annotation, and it might be nice to use it instead of an arbitrarily-chosen `user+arbitrary@domain`. However, to do that, the regex you send to PostgreSQL needs to be adjusted (see next comment)

scripts/mass_user_emails.py
```diff
@@ -212,12 +258,23 @@ def get_eligible_users(user_detail_email_key, force=False):
...
+    # regex finds 'name+filter@example.com' type addresses
+    return eligible_users.annotate(email_cleaned=Func(
+            Lower(F('email')),
+            Value(r'\+[^*]*@'),
+            Value(''),
+            Value('g'),
+            function='REGEXP_REPLACE',
+            output_field=CharField(),
+    )).distinct(
+        'email_cleaned'
```
jnm, line 277: I think this doesn't quite do what's expected. My reading of the regex replacement is:
1. Find any sequence of a literal `+` ,followed by zero or more characters that are not `*`, followed by a `@`
1. …and replace all that with nothing
1. …and repeat this until no matches remain.

The result is, if I have:
| `email=` | …then I get `email_cleaned=` |
| - | - |
| john@john.john | john@john.john |
| john+clone2@john.john | johnjohn.john |
| john+clone99@john.john | johnjohn.john |

After the `distinct()` call, the emails remaining as recipients are:
* john@john.john and
* either john+clone99@john.john or john+clone2@john.john (these are returned indeterminately by PostgreSQL).

I'd suggest cleaning up the regex and then taking advantage of your annotation to send mail to `user.email_cleaned`; that way, messages always go to `john@john.john` instead of `john+cloneblah@john.john`.

jnm (changes requested):

scripts/mass_user_emails.py
LMNTL: Thank you for noticing that! I've updated both points and tested the regex (in context and from the shell).

jnm (approved):

PR #4616, opened 2023-09-06
Adds 'unsaved changes' state and navigation block when editing QA questions

### LINKED ISSUE #4584: Block navigation if there is unsaved data in Analysis tab

duvld:
## Description
Basic navigation block and replaces 'Saved' with 'Unsaved changes' if there are changes not yet uploaded.

WIP for tags input.

## Related issues

Fixes #4584

### DIFF
--- a/jsapp/js/components/processing/analysis/analysisQuestions.reducer.ts
+++ b/jsapp/js/components/processing/analysis/analysisQuestions.reducer.ts
@@ -6,6 +6,12 @@ import {applyUpdateResponseToInternalQuestions} from './utils';
 export interface AnalysisQuestionsState {
   /** Whether any async action is being done right now. */
   isPending: boolean;
+  /**
+   * True if user has made a change and the change has not been requested OR
+   * backend differs from frontend. This is determined by a 200 response or the
+   * pending state.
+   */
+  changesDetected: boolean;
   questions: AnalysisQuestionInternal[];
   /**
    * A list of uids of questions with definitions being edited. I.e. whenever
@@ -32,6 +38,7 @@ type AnalysisQuestionReducerType = (
 
 export const initialState: AnalysisQuestionsState = {
   isPending: false,
+  changesDetected: false,
   questions: [],
   questionsBeingEdited: [],
 };
@@ -79,6 +86,7 @@ export const analysisQuestionsReducer: AnalysisQuestionReducerType = (
         questions: [newQuestion, ...state.questions],
         // We immediately open this question for editing
         questionsBeingEdited: [...state.questionsBeingEdited, newUuid],
+        changesDetected: true,
       };
     }
     case 'deleteQuestion': {
@@ -97,12 +105,17 @@ export const analysisQuestionsReducer: AnalysisQuestionReducerType = (
       return {
         ...state,
         isPending: false,
+        changesDetected: false,
         questions: action.payload.questions,
       };
     }
     case 'startEditingQuestion': {
       return {
         ...state,
+        // We can put this on the text input of titles and select
+        // questions for slight UX improvement but this is simpler
+        // and understandable.
+        changesDetected: true,
         questionsBeingEdited: [
           ...state.questionsBeingEdited,
           action.payload.uuid,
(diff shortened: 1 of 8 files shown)


jsapp/js/components/processing/analysis/analysisQuestions.reducer.ts
```diff
@@ -6,6 +6,12 @@ import {applyUpdateResponseToInternalQuestions} from './utils';
 export interface AnalysisQuestionsState {
   /** Whether any async action is being done right now. */
   isPending: boolean;
+  /**
+   * True if user has made a change and the change has not been requested OR
+   * backend differs from frontend. This is determined by a 200 response or the
+   * pending state.
+   */
```
magicznyleszek, line 13: I think this would be more precise:

> It's `true` if user has made a change in the UI and the Back end has not been updated yet. Every completed call to API will change it back to `false`.

jsapp/js/components/processing/analysis/analysisQuestions.reducer.ts
```diff
@@ -97,12 +105,17 @@ export const analysisQuestionsReducer: AnalysisQuestionReducerType = (
...
+        hasUnsavedWork: false,
         questions: action.payload.questions,
       };
     }
     case 'startEditingQuestion': {
       return {
         ...state,
+        // We can put this on the text input of titles and select
+        // questions for slight UX improvement but this is simpler
+        // and understandable.
```
magicznyleszek, line 117: I don't understand this comment 😅

jsapp/js/components/processing/analysis/editors/analysisQuestionEditor.component.tsx
```diff
@@ -18,6 +18,7 @@ import singleProcessingStore from 'js/components/processing/singleProcessingStor
 import clonedeep from 'lodash.clonedeep';
 import {handleApiFail} from 'js/utils';
 import type {FailResponse} from 'js/dataInterface';
+import {func} from 'prop-types';
```
magicznyleszek, line 21: Unused import?

jsapp/js/components/processing/analysis/editors/analysisQuestionEditor.component.tsx
```diff
@@ -153,39 +154,46 @@ export default function AnalysisQuestionEditor(
...
 
+  function onSubmit(evt: React.FormEvent<HTMLFormElement>) {
+    evt.preventDefault();
+  }
+
   return (
     <>
       <header className={styles.header}>
-        <div className={commonStyles.headerIcon}>
+        <form className={styles.form} onSubmit={onSubmit}>
```
magicznyleszek, line 164: Can we call it `headerForm`?

jsapp/js/components/processing/analysis/utils.ts
```diff
@@ -287,3 +287,13 @@ export async function updateResponseAndReducer(
     dispatch({type: 'updateResponseFailed'});
   }
 }
+
+export function hasUnsavedWork(
```
magicznyleszek, line 291: You can delete this function, and use one liners instead

magicznyleszek (changes requested):

jsapp/js/components/processing/analysis/analysisQuestions.reducer.ts
magicznyleszek: So perhas:

> Instead of checking changes on every input in the edited question, we assume that, as soon as user starts to edit a question, there are unsaved changes. This is not ideal UX, but it's much simpler and still logical.

magicznyleszek (approved):

PR #4624, opened 2023-09-07
Fix wrong language (i.e. English) being picked instead of detected language in language selector

### LINKED ISSUE #2313: Changing language seems broken
## Description

Switching language in the Header Account menu doesn't seem to work. It just reloads page and everything is still in english. Possibly related to #2302

noliveleger:
## Description

Always save detected language to Django language cookie when it does not exist already to avoid falling back on English

## Notes

Overload Django `LocaleMiddleware` to use the detected language by the middleware to set the cookie language when it is not present. It lets the front end load the correct value and set it when rendering the language dropdown menu. 
Moreover, it will update `extra_details.last_ui_language` with correct value to get more accurate statistics.

## Related issues

Fixes #2313

### DIFF
--- a/hub/middleware.py
+++ b/hub/middleware.py
@@ -1,5 +1,30 @@
-# coding: utf-8
+from django.conf import settings
+from django.middleware.locale import LocaleMiddleware as DjangoLocaleMiddleware
 from django.utils.deprecation import MiddlewareMixin
+
+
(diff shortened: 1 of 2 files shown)


JacquelineMorrissette (approved):
This is working on my end. :+1:

PR #4629, opened 2023-09-08
Hide qual_note questions from Data Table and Single Submission Modal

### LINKED ISSUE #4619: Don't display `qual_note` in the Data Table

magicznyleszek:
## Checklist

3. [x] Ensure the tests pass

## Description

The `qual_note` Analysis Question type doesn't contain any useful data, so we don't display it in the Project Data UI.

## Notes

We have two places that are responsible for data displaying (one for Data Table and other for Single Submission Modal). I added an exclusion in both of these places.

## Related issues

Fixes #4619


jsapp/js/components/submissions/submissionUtils.ts
```diff
@@ -108,16 +110,23 @@ export class DisplayResponse {
 
 /**
  * Returns a sorted object of transcript/translation keys
+ *
+ * Note: we omit returning `qual_note` questions.
  */
-export function sortAnalysisFormJsonKeys(additionalFields: Array<{source: string, dtpath: string}>) {
+function sortAnalysisFormJsonKeys(additionalFields: AnalysisFormJsonField[]) {
```
duvld, line 116: Was this made before AnalysisFormJsonField was created/extended? There's a lot more mandatory fields than the parameter it replaced

jsapp/js/components/processing/analysis/constants.ts
```diff
@@ -16,6 +16,10 @@ export type AnalysisQuestionType =
   | 'qual_tags'
   | 'qual_text';
 
+// We use it on other parts of code that is not related to analysis and is not
```
duvld, line 19: I'm not sure what this comment means, I feel like it should say something like "We don't need to show qualitative note questions outside of analysis"? Or maybe I'm misunderstanding

jsapp/js/assetUtils.ts
```diff
@@ -520,21 +521,26 @@ export function injectSupplementalRowsIntoListOfRows(
   // on Back end, to build a list of columns grouped by source question
   const additionalFields = asset.analysis_form_json?.additional_fields || [];
   const extraColsBySource: Record<string, AnalysisFormJsonField[]> = {};
-  additionalFields.forEach((add_field: AnalysisFormJsonField) => {
-    const sourceName: string = add_field.source;
+  additionalFields.forEach((field: AnalysisFormJsonField) => {
+    // We don't want to display notes in the UI, so we omit it here:
```
duvld, line 525: Maybe a bit more verbose of a comment?

jsapp/js/components/submissions/submissionUtils.ts
magicznyleszek: I was following the code that is using that function. It's actually being used only in one place, in `getSubmissionDisplayData`, which is passing `asset.analysis_form_json?.additional_fields` into it, and this prop is typed as `additional_fields: AnalysisFormJsonField[];`

jsapp/js/components/processing/analysis/constants.ts
magicznyleszek: I wanted to justify having a `const` here, while that `AnalysisQuestionType` should be more than enough. I rewrote the comment, but maybe the best way is to not have a comment at all here?

My thinking right now is that before TypeScript we had these long elaborate enums to ensure every piece of code is using the same strings. After moving some code to TS it become clear, that having a type is safe enough, and keeping const/enums for list of names of things is overdoing things.

jsapp/js/assetUtils.ts
magicznyleszek: Rewrote it here and in `submissionUtils.ts` too

PR #4635, opened 2023-09-13
Account for null XForm instances in service_usage endpoint

LMNTL:
## Checklist

3. [X] Ensure the tests pass

## Description

Fix `/api/v2/service_usage/` calculating NLP and submission counts incorrectly.

## Notes

Deleting a form with submissions currently causes the submission counter to decrement. The endpoint previously did handle this correctly, but it was refactored as part of `feature/usage-limits`, which introduced this regression. See [kobocat#895](kpi/deployment_backends/kc_access/shadow_models.py) for additional background.

### Depends on kobotoolbox/kobocat#895

See [issue](https://github.com/orgs/kobotoolbox/projects/60/views/1?pane=issue&itemId=38370697) on Release 2.023.37 QA board.

### DIFF
--- a/kpi/serializers/v2/service_usage.py
+++ b/kpi/serializers/v2/service_usage.py
@@ -141,7 +141,7 @@ class ServiceUsageSerializer(serializers.Serializer):
 
         # Only use fields we need to improve SQL query speed
         user_assets = (
-            Asset.objects.only(
+            Asset.all_objects.only(
                 'pk',
                 'uid',
                 '_deployment_data',
--- a/kpi/serializers/v2/service_usage.py
+++ b/kpi/serializers/v2/service_usage.py
@@ -155,10 +155,10 @@ class ServiceUsageSerializer(serializers.Serializer):
                 asset_type=ASSET_TYPE_SURVEY,
                 # Make sure we're only getting assets that are deployed
                 _deployment_data__has_key='backend',
-            ).values('uid')
+            )
         )
 
-        asset_list = [asset['uid'] for asset in user_assets]
+        asset_list = [asset.uid for asset in user_assets]
 
         xforms = KobocatXForm.objects.only('bytes_sum', 'id', 'kpi_asset_uid').filter(
             kpi_asset_uid__in=asset_list,
--- a/kpi/serializers/v2/service_usage.py
+++ b/kpi/serializers/v2/service_usage.py
@@ -149,16 +149,19 @@ class ServiceUsageSerializer(serializers.Serializer):
                 'name',
             )
             .select_related('owner')
+            .exclude(_deployment_data=None)
             .filter(
                 owner__in=self._users,
                 asset_type=ASSET_TYPE_SURVEY,
                 # Make sure we're only getting assets that are deployed
                 _deployment_data__has_key='backend',
-            )
+            ).values('uid')
         )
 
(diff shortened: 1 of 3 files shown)


kpi/serializers/v2/service_usage.py
```diff
@@ -141,14 +141,15 @@ def _get_per_asset_usage(self, user):
...
                 'uid',
                 '_deployment_data',
                 'owner_id',
                 'name',
             )
             .select_related('owner')
+            .exclude(_deployment_data=None)
             .filter(
                 owner__in=self._users,
                 asset_type=ASSET_TYPE_SURVEY,
```
noliveleger, line 155: `user_assets` seems to be used no where else than to be used to init `asset_list` 
I would probably get rid of `user_assets` and assign directly `asset_list` with the results of queryset with `values_list()`

Moreover, the query could simplified with the new `date_deployed` field. Only assets of type `'survey'` are deployed. We just need to check whether the value of the field is not null to get all deployed projects.

So what about this? 

```python 
        asset_list = list(
            Asset.all_objects.values_list('uid', flat=True).filter(
                owner__in=self._users,
                date_deployed__isnull=False,
            )
        )
        
        xforms = KobocatXForm.objects.only('bytes_sum', 'id', 'kpi_asset_uid').filter(
            kpi_asset_uid__in=asset_list,
        )
```

kpi/serializers/v2/service_usage.py
```diff
@@ -180,7 +183,9 @@ def _get_per_asset_usage(self, user):
                 'date', 'xform', 'counter'
             )
             .filter(
-                xform__in=xforms,
+                # get submission counters for all deployed assets OR the null xform counter
+                Q(user__in=[kpi_user.id for kpi_user in self._users], xform=None) |
```
noliveleger, line 187: Same logic for `self._users` as `asset_list` above. It seems to be used only to create this list.  
What about renaming `self._users` to `self._user_ids`  and get the ids directly when fetching all users from the organization.

```python
    def _get_per_asset_usage(self, user):
         self._user_ids = [user.pk]

        # Get the organization ID passed in from the query parameters
        organization_id = self.context['request'].query_params.get(
            'organization_id'
        )
        self._get_organization_details(organization_id)
        <snippet>

    def _get_organization_details(self, organization_id=None):
         <snippet>
        # If the user is in an organization, get all org users so we can query their total org usage
        self._user_ids = User.objects.values_list('pk', flat=True).filter(
            organizations_organization__id=organization_id
        )
```

BTW, since `organization_id` is only used to be passed to `_get_organization_details()`, I would move all its assignment logic inside `_get_organization_details()`.  `_get_per_asset_usage()` is super long and deserves to be splitted ;-)

kpi/serializers/v2/service_usage.py
```diff
@@ -265,12 +270,12 @@ def _get_organization_details(self, organization_id=None):
         if not organization_id:
             return
 
-        organization = Organization.objects.filter(
+        self._organization = Organization.objects.filter(
```
noliveleger, line 273: I think we can keep `organization` as it only lives within the scope of `_get_organization_details()`. 
Any reasons to make it a protected attribute of the class?

noliveleger (changes requested):

PR #4645, opened 2023-09-19
Typescriptize permissions files

magicznyleszek:
## Checklist

3. [x] Ensure the tests pass

## Description

Internal code improvements.

## Notes

The changes in this PR include:
- Removed few Reflux actions that are no longer used.
- Typescriptized all files from `jsapp/js/components/permissions/`.
- Deduplicated and improved some types.
- Created `js/users/utils` for user-related helper functions.
- Replaced `stores.userExists` with `userExistence.store.ts`.
- Fixed some styles in the Share modal.
- Replaced `PermValidator` with a helper function.
- Removed safety check from `buildUserUrl` as we no longer pass url to it.
- Some of the Front-end code was not in sync with data from endpoints, I had to fix that.

The main gist is not to refactor this completely or improve in great detail. The goal is to migrate code into TS with some cleanups and no bugs introduced :) It's not meant to be perfect, just a medium sized step into the right direction 🦶➡️☑️  

## Related issues

Includes code from #4651, so that PR should ideally be reviewed first :)


p2edwards (approved):

p2edwards (changes requested):
Wait, there's a new problem that happens if you remove someone from the sharing section.

magicznyleszek:
@p2edwards I've fixed merge conflicts ;)

jsapp/js/app.js
```diff
@@ -86,7 +86,6 @@ class App extends React.Component {
         <React.Fragment>
           <Tracking />
           <ToasterConfig />
-          <PermValidator />
```
p2edwards, line 89: @magicznyleszek Remove this too? 
```js
import PermValidator from 'js/components/permissions/permValidator';
```

magicznyleszek:
@p2edwards I've tested it more and found three things:
- added cache invalidation for `UserExistence` store, so that when someone checks for user that doesn't exist, and then at some point in near future that user is being created, check will return true value (instead of memoized one)
- removed unnecessary fallback for getting uid
- fixed a bug - `SharingForm` was adding listeners that were never removed, so when user visited multiple projects, in each opening sharing settings, the code was making multiple calls (each for visited project), and randomly (possibly based on API call response time) displaying permissions from one of them

jsapp/js/users/userExistence.store.ts
```diff
@@ -0,0 +1,67 @@
...
+  /** Resolves with either memoized value or API call response. */
+  async checkUsername(username: string): Promise<boolean> {
+    // If few minutes passed since the last check, we want to wipe the memoized
+    // values, since there is a chance a user might be created in the meantime
+    // and we don't want to risk storing untrue `false` that will never be fixed
+    // (unless user reloads the browser).
+    const now = new Date();
+    const diffMs = (now.valueOf() - this.lastCheckDate.valueOf());
+    const diffMins = Math.round(((diffMs % 86400000) % 3600000) / 60000);
+    if (diffMins >= 3) {
```
p2edwards, line 36: What's all this conversion about?

```ts
const diffMins = Math.round(((diffMs % 86400000) % 3600000) / 60000);
if (diffMins >= 3) {
```

I have some problems with this code.

- It's hard to read.
- Rounding minutes before comparing `(diffMins >= 3)` makes 2½ minutes the limit instead of 3.
- The modulo `((…) % 3600000)` creates a brief window every hour where cache invalidation doesn't happen. 
- `% 86400000` is redundant with `% 3600000` (they divide evenly).

Suggestion:

- Compare the epoch time in milliseconds.

```ts
const limitMs = 3 * 60 * 1000; // 3 minutes
if (diffMs >= limitMs) {  /* invalidate cache */ }
```

```suggestion
    const limitMs = 3 * 60 * 1000; // 3 minutes
    if (diffMs >= limitMs) {
```

magicznyleszek: A lesson to not trust trusted stack overflow responses 😉

p2edwards (approved):

p2edwards (approved):

PR #4660, opened 2023-09-26
Show storage add-ons in usage limits

LMNTL:
## Checklist

3. [X] Ensure the tests pass

## Description

Previously purchase storage add-ons are now used in usage limit calculations. Corrects some issues that were causing incorrect limits to be displayed.

## Notes

We previously determined whether to show the `FREE_TIER_THRESHOLDS` based on the logged-in user's join date. But for multi-user organizations, that meant each user in the organization could potentially see a different set of limits for their organization. This PR changes that so custom tiers are only shown if the _organization owner_ joined before `FREE_TIER_CUTOFF_DATE`. If the user isn't in an organization, it still defaults to checking the user's join date.

### DIFF
--- a/jsapp/js/account/stripe.api.ts
+++ b/jsapp/js/account/stripe.api.ts
@@ -120,6 +120,38 @@ export async function getSubscriptionInterval() {
   return 'month';
 }
 
+const defaultLimits: AccountLimit = Object.freeze({
+  submission_limit: 'unlimited',
+  nlp_seconds_limit: 'unlimited',
+  nlp_character_limit: 'unlimited',
+  storage_bytes_limit: 'unlimited',
+});
+
+function getLimitsForMetadata(
+  metadata: {[key: string]: string},
+  limitsToCompare: false | AccountLimit = false
+) {
+  const limits: Partial<AccountLimit> = {};
+  for (const [key, value] of Object.entries(metadata)) {
+    // if we need to compare limits, make sure we're not overwriting a higher limit from somewhere else
+    if (limitsToCompare) {
+      if (
+        !limitsToCompare?.[key as keyof AccountLimit] ||
+        (value !== 'unlimited' &&
+          value <= limitsToCompare[key as keyof AccountLimit])
+      ) {
+        continue;
+      }
+    }
+    // only use metadata needed for limit calculations
+    if (key in defaultLimits) {
+      limits[key as keyof AccountLimit] =
+        value === 'unlimited' ? value : parseInt(value);
+    }
+  }
+  return limits;
+}
+
 export async function getAccountLimits() {
   await when(() => subscriptionStore.isInitialised);
   const subscriptions = [...subscriptionStore.subscriptionResponse];
@@ -129,8 +161,11 @@ export async function getAccountLimits() {
   let metadata;
   let hasFreeTier = false;
   if (activeSubscriptions.length) {
-    // get metadata from the user's subscription
-    metadata = activeSubscriptions[0].items[0].price.product.metadata;
+    // get metadata from the user's subscription (prioritize price metadata over product metadata)
+    metadata = {
+      ...activeSubscriptions[0].items[0].price.product.metadata,
+      ...activeSubscriptions[0].items[0].price.metadata,
+    };
   } else {
     // the user has no subscription, so get limits from the free monthly price
     hasFreeTier = true;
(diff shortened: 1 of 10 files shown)


jsapp/js/account/stripe.api.ts
```diff
@@ -120,17 +119,52 @@ export async function getSubscriptionInterval() {
   return 'month';
 }
 
+const defaultLimits: AccountLimit = Object.freeze({
```
bufke, line 122: It looks like we use ALL_CAPS for constants elsewhere in code.

jsapp/js/account/stripe.api.ts
```diff
@@ -120,17 +119,52 @@ export async function getSubscriptionInterval() {
...
+  metadata: {[key: string]: string},
+  limitsToCompare: false | AccountLimit = false
+) {
+  const limits: Partial<AccountLimit> = {};
+  for (const [key, value] of Object.entries(metadata)) {
+    // if we need to compare limits, make sure we're not overwriting a higher limit from somewhere else
+    if (limitsToCompare) {
+      if (
+        !limitsToCompare?.[key as keyof AccountLimit] ||
+        (value !== 'unlimited' &&
```
bufke, line 139: I don't like that I can add a typo here and the code compiles. value ends up as a string and not a number | unlimited. I'm tempted to say that unlimited should be represented as a null, enum, variable, or improve the types somehow. I'll leave this to you to act upon or not. I think I prefer null because "Your limit is null" seems like a correct thing to say, from a weird programmer context at least. But changing to null may be a too large of a change to consider.

If you feel that value could in reality contain weird stuff like "unlimitted" then a simple solve would just be to make something like `const UNLIMITED_TOKEN = "unlimited";` to prevent the programmer from making a mistake.

jsapp/js/account/stripe.api.ts
```diff
@@ -120,17 +119,52 @@ export async function getSubscriptionInterval() {
...
+          value <= limitsToCompare[key as keyof AccountLimit])
+      ) {
+        continue;
+      }
+    }
+    // only use metadata needed for limit calculations
+    if (key in defaultLimits) {
+      limits[key as keyof AccountLimit] =
+        value === 'unlimited' ? value : parseInt(value);
```
bufke, line 148: Wait but `parseInt("unlimitted")` can return NaN. How is TS ok with this?

jsapp/js/account/stripe.api.ts
```diff
@@ -120,17 +119,52 @@ export async function getSubscriptionInterval() {
...
+  nlp_seconds_limit: 'unlimited',
+  nlp_character_limit: 'unlimited',
+  storage_bytes_limit: 'unlimited',
+});
+
+function getLimitsForMetadata(
+  metadata: {[key: string]: string},
+  limitsToCompare: false | AccountLimit = false
+) {
+  const limits: Partial<AccountLimit> = {};
```
bufke, line 133: Do we really need this partial here? Is there a way limits can be created without init'ed to {}? Consider most of my comments as optional.

bufke (changes requested):
I'm not a big fan of unit tests for JS, but I feel that getLimitsForMetadata and others have gotten too complex and merits it. Doesn't need to be now and happy to discuss. Tightening up the types (or simplifying) would mitigate my concern. It feels like the business logic has just gotten a bit too complex from where it started. It's not longer obvious at a glance as to what it's doing.

kpi/views/environment.py
```diff
@@ -174,17 +175,26 @@ def process_other_configs(request):
...
-            }
-            data['free_tier_display'] = {
-                'name': None,
-                'feature_list': [],
-            }
+        if request.user.id:
+            # default to checking the user's join date
+            date_joined = request.user.date_joined.date()
+            # if the user is in an organization, use the organization owner's join date instead
+            if request.user.organizations_organization:
```
bufke, line 182: Could be simplified as one ORM query. I'm pretty sure you're doing a query on this very line. All you really want to fetch here is the date_joined (or null) of the org owner. Not 100% sure, but if you think about it from the org owner's perspective you might be able to simplify. Fetch that one row, filter away others, get the date. At a glance that sounds like one query and less logical statements.

jsapp/js/account/stripe.api.ts
LMNTL: We need the limits to represent a subset of `AccountLimit` here because some of the products (like one-time add-ons) include only some of the `AccountLimit` keys.

image

kobo/apps/stripe/constants.py
```diff
@@ -4,3 +4,15 @@
     'past_due',
     'trialing',
 ]
+
+FREE_TIER_NO_THRESHOLDS = {
```
bufke, line 8: Possibly out of scope. If there was ever a good case for type hints, this is it. Lacking type hints, a DRF serializer on the Environments endpoint would help ensure correctness. But the environments endpoint doesn't have one. IMO this should be considered in a refactor. You could ponder if the added complexity here merits adding a serializer. While not as powerful as real type hints, the serializer would provide a runtime guarantee that data is returned in the expected format. Currently it's a untyped dict that is returned, which cannot provide proof that the API consistently returns the expected format the Frontend needs to accept.

bufke (changes requested):
We discussed that the Environment API be better off not being user specific but that might be out of scope.

PR #4664, opened 2023-09-26
Fix mini audio play button requiring two clicks to play

p2edwards:
## Checklist

1. If you've added code that should be tested, add tests
2. If you've changed APIs, update (or create!) the documentation
3. [x] Ensure the tests pass
7. Open an issue in the [docs](https://github.com/kobotoolbox/docs/issues/new) if there are UI/UX changes

## Description

Fix a problem with the audio question play button in the data table

## Notes

While testing #4487 in my browser (Chrome on KDE / X), I noticed a glitch with the MiniAudioPlayer component (present on local and production).

The glitch was you'd have to click the play button twice to get it to play.

- In production it would log an error to the console.
- In dev the first click would also show a full page blocking splash screen, which was pretty disruptive for testing.

blast

> **AbortError**
> The [play() request was interrupted](https://developer.chrome.com/blog/play-request-was-interrupted/) because the media was **removed from the document**.

The word "removed" here was a little misleading — it was never even added.

By attaching the `<audio>` element to the DOM and accessing its methods with a ref, instead of holding it as an unattached element, this problem went away. As a bonus, we could then simplify some of the event handling code (using React's SyntheticEvents system)

When I told Sara about this she couldn't reproduce the original glitch in her browser. So it might not have a broad impact, but at least it's a nice clean up.

## Related issues

Follow-up on #4487

### DIFF
--- a/jsapp/js/components/common/miniAudioPlayer.scss
+++ b/jsapp/js/components/common/miniAudioPlayer.scss
@@ -10,6 +10,10 @@
   flex-wrap: nowrap;
   color: colors.$kobo-blue;
   height: button.$button-height-s;
+
+  > audio {
+    display: none;
+  }
 }
 
 .mini-audio-player.mini-audio-player--is-loading {
--- a/jsapp/js/components/common/miniAudioPlayer.tsx
+++ b/jsapp/js/components/common/miniAudioPlayer.tsx
@@ -65,23 +65,7 @@ class MiniAudioPlayer extends React.Component<
   }
 
   componentDidMount() {
-    // NOTE: 'metadata' causes an immediate download of part of the file (to get
-    // the metadata), usually requires around 30-50KB, but it may vary. Some
-    // browser may simply download whole file.
-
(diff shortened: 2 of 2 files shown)


jsapp/js/components/common/miniAudioPlayer.scss
```diff
@@ -10,6 +10,10 @@
   flex-wrap: nowrap;
   color: colors.$kobo-blue;
   height: button.$button-height-s;
+
+  > audio {
```
srartese, line 14: Since I can't see the double click error what this is targeting? Nothing breaks or looks different, just curious

srartese (approved):

jsapp/js/components/common/miniAudioPlayer.scss
p2edwards: We've been rendering our own play button for the audio player, and we still are. This rule prevents a browser from displaying a built-in widget, now that we're actually attaching an `<audio>` tag to the DOM.

PR #4666, opened 2023-09-26
Cache service usage endpoint

LMNTL:
## Checklist

3. [X] Ensure the tests pass

## Description

Caches the `api/v2/service_usage` endpoint to prevent slowdown on sites with a large number of users/projects. Adds a 'last updated' widget to the Usage page. Allows promotion codes to be used when checking out for a Plan.

## Notes

* Added a new setting/env var, `ENDPOINT_CACHE_DURATION`.
* Refactored `service_usage` endpoint to use a detail view (`/api/v2/organizations/{ORGANIZATION_ID}/service_usage/`) instead of POSTing the organization ID.
* The badge on the usage page is based on the cached age of the response from the service usage endpoint (user or organization). To reset it, log out and back in.

### DIFF
--- a/jsapp/js/account/useUsage.hook.ts
+++ b/jsapp/js/account/useUsage.hook.ts
@@ -1,8 +1,9 @@
 import {createContext, useEffect, useState} from 'react';
 import type {RecurringInterval} from 'js/account/stripe.api';
 import {getSubscriptionInterval} from 'js/account/stripe.api';
-import {truncateNumber} from 'js/utils';
+import {formatRelativeTime, truncateNumber} from 'js/utils';
 import {getUsageForOrganization} from 'js/account/usage.api';
+import moment from 'moment';
 
 export interface UsageState {
   storage: number;
@@ -13,6 +14,7 @@ export interface UsageState {
   currentYearStart: string;
   trackingPeriod: RecurringInterval;
   isPeriodLoaded: boolean;
+  lastUpdated?: String | null;
   isLoaded: boolean;
 }
 
@@ -26,6 +28,7 @@ export function useUsage() {
     currentYearStart: '',
     trackingPeriod: 'month',
     isPeriodLoaded: false,
+    lastUpdated: '',
     isLoaded: false,
   });
 
@@ -46,10 +49,16 @@ export function useUsage() {
     if (!usage.isPeriodLoaded) {
       return;
     }
-
     getUsageForOrganization().then((data) => {
       if (!data) {
         return;
+      }
+      let lastUpdated: UsageState['lastUpdated'] = null;
+      if ('headers' in data && data.headers instanceof Headers) {
+        const lastUpdateDate = data.headers.get('date');
+        if (lastUpdateDate) {
+          lastUpdated = formatRelativeTime(lastUpdateDate);
+        }
       }
       setUsage((prevState) => {
         return {
@@ -70,6 +79,7 @@ export function useUsage() {
             ],
           currentMonthStart: data.current_month_start,
           currentYearStart: data.current_year_start,
+          lastUpdated: lastUpdated,
           isLoaded: true,
         };
       });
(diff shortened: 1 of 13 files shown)


kpi/views/v2/service_usage.py
```diff
@@ -1,16 +1,22 @@
...
-from rest_framework.decorators import action
 from rest_framework.response import Response
 
 from kpi.permissions import IsAuthenticated
 from kpi.serializers.v2.service_usage import ServiceUsageSerializer
 from kpi.utils.object_permission import get_database_user
 
 
+@method_decorator(cache_page(60 * 30), name='retrieve')
+@method_decorator(cache_page(60 * 30), name='list')
```
noliveleger, line 17: Let's use a setting for this (which could even be configurable with env variable, then fallback on your default). We could have a shorter timeout in dev settings than prod. 

Maybe 30 minutes is a little bit too long, what about reverting it to 15?

kpi/views/v2/service_usage.py
```diff
@@ -83,6 +90,9 @@ def list(self, request, *args, **kwargs):
         )
         return Response(data=serializer.data)
 
-    @action(methods=['post'], detail=False)
-    def post(self, request, *args, **kwargs):
-        return self.list(request, *args, **kwargs)
+    def retrieve(self, request, pk=None):
```
noliveleger, line 93: We'd rather keep DRF signature, i.e. `def retrieve(self, request, pk=None, *args, **kwargs)`. 
(for example, `format` does not work without those parameters)

🤔 Anyway, this endpoint does not seem RESTFul compliant to me. `pk` is not a unique identifier of an `service_usage` model. 

I'd rather see something like:

- `/api/v2/organizations/<pk>/service_usage/`
- `/api/v2/service_usage/`

It may involve more front-end (and back-end) changes but I think it's worth it ;-)

noliveleger (changes requested):

jsapp/js/account/useUsage.hook.ts
```diff
@@ -46,11 +49,17 @@ export function useUsage() {
     if (!usage.isPeriodLoaded) {
       return;
     }
-
     getUsageForOrganization().then((data) => {
       if (!data) {
         return;
       }
+      let lastUpdated: UsageState['lastUpdated'] = null;
+      if ('headers' in data && data.headers instanceof Headers) {
```
noliveleger, line 57: Could be it simplified with `if (data?.headers instanceof Headers)`?

noliveleger, line 57: Could be it simplified with `if (data?.headers instanceof Headers)`?

kobo/settings/base.py
```diff
@@ -1363,6 +1363,9 @@ def dj_stripe_request_callback_method():
     'default': env.cache(default='redis://redis_cache:6380/3'),
 }
 
+# How long to retain cached responses for kpi endpoints
+ENDPOINT_CACHE_DURATION = env.str('ENDPOINT_CACHE_DURATION', 60 * 15)  # 15 minutes
```
noliveleger, line 1367: I think we should better use `env.int()` instead of `env.str()` here? Even the default value is an integer?

noliveleger (changes requested):

kpi/views/v2/service_usage.py
LMNTL: I've extracted this out to `ENDPOINT_CACHE_DURATION`. This setting is used by `/service_usage/` and `/stripe/products/`, but that can be refactored if we later need to set those cache settings independently.

The only thing I'm not sure about is using a shorter timeout in dev settings; I'm worried that could make it easier to miss unwanted results of caching while debugging.

jsapp/js/account/useUsage.hook.ts
LMNTL: Unfortunately, TS complains here without `in` acting as a type guard:

image

image

PR #4676, opened 2023-10-04
Added logic to support all audio formats in the player

### LINKED ISSUE #4597: Data Tab view using the wrong endpoint for audio attachments
## Description

While collecting background data with Collect, an error occurs while trying to play AMR files.

## Steps to Reproduce

1. Create a new form
2. Enable background audio 
3. Set the audio quality to `Voice-Only`
4. Save and deploy the form
5. Download the form on Collect
6. Complete a submission with the device collecting background audio
7. Send the submission
8. Refresh web UI
9. Go to the form Data tab
10. Try to play audio and receive an error

## Expected behavior 

The Audio should play without issue

## Actual behavior 

There is an error when attempting to play the `.amr` file

## Additional details

The `/attachments` endpoint loads the `.amr` file without issue, and the file is playable on download. 
This is probably the culprit: 
```Quote
AMR file is not one of the supported audio files that html5 audio tag supports. Please see the list below for supported audio formats:

http://en.wikipedia.org/wiki/HTML5_Audio#Supported_audio_coding_formats

I would recommend you convert your amr file to mp3 or ogg and then upload it and play it on your server.
``` 
https://stackoverflow.com/a/27915473

We should be using the audio-conversion feature of the `/attachments/` endpoint here to avoid issues like this
`/api/v2/assets/{asset_uid}/data/{data_id}/attachments/?xpath=Path_to_audio_question&format=mp3`

srartese:
## Checklist

3. [x] Ensure the tests pass

## Description
Added a check for the type of audio file and if the player doesn't support it then it is sent to the endpoint to format it to an mp3.  This will allow types like `.AMR` and `.AAC` to be uploaded via Enketo or specifically the background audio from collect to be played back in the audio player. 

## Related issues

Fixes #4597


jsapp/js/components/submissions/submissionUtils.ts
```diff
@@ -531,11 +532,35 @@ export function getMediaAttachment(
...
+        && !attachment.mimetype.includes('mpeg')
+        && !attachment.mimetype.includes('/wav')
+        && !attachment.mimetype.includes('ogg')
+      )
+      {
+        const lastItem = attachment.filename.split("/").pop();
+        const questionPath = Object.keys(submission).find(key => submission[key] === lastItem);
+
+        const newAudioURL = 
+        `${ROOT_URL}/api/v2/assets/${submission._xform_id_string}/data/${attachment.instance}/attachments/?xpath=${questionPath}&format=mp3`;
```
magicznyleszek, line 549: When `questionPath` is `undefined`, this will crash the app.

Steps to break it:

1. Use a simple form with audio question 
[simple-audio-form.xlsx](https://github.com/kobotoolbox/kpi/files/12874349/simple-audio-form.xlsx)
2. Add submission using `flac` file (e.g. https://github.com/ietf-wg-cellar/flac-test-files/blob/main/subset/01%20-%20blocksize%204096.flac)
3. Go to Data Table and click this tiny play button. You will see `400` error on this:
   `GET
http://kf.kobo.local/api/v2/assets/aJpiarEu6gxjLrg7wuXpDq/data/34/attachments/?xpath=undefined&format=mp3` and another error: `Uncaught (in promise) DOMException: The media resource indicated by the src attribute or assigned media provider object was not suitable.`

magicznyleszek: I'm not sure, but perhaps `Object.keys` above should be `Object.values`?

magicznyleszek (changes requested):

jsapp/js/components/submissions/submissionUtils.ts
srartese: So this problem is because when you upload the file it gets renamed by removing the '_' in the file. Unfortunately it's not as simple as just replacing the `_` with spaces because it only does it for the original file name but numeric values are getting added to the end of the file and that is our new file name.

For example: 
We upload: `01 - blocksize 4096.flac`
The attachment file name is `01_-_blocksize_4096-12_47_9.flac`
The question value is: `01 - blocksize 4096-12_47_9.flac`

We were comparing the file name to the question value in order to get the question path name but clearly that doesn't work when the name of the file is changing between the file name and the actual value of the question file. 
I'm working on the solution now and making sure theres not another, much cleaner way of doing this.

PR #4684, opened 2023-10-13
[TASK-91] Add organization type signup

bufke:
## Description

Add new user field, organization type. This field can be enabled in Django admin constance settings.

## Notes

Add organization type field to registration and profile page. The field is a dropdown with fixed options.


bufke:
I cannot set myself as reviewer, but I need to review the latest commit.

bufke:
@p2edwards when I test this, I don't see organization type on the account settings page. It appears on registration.

p2edwards:
@bufke Account settings page will be another task / PR since it crosses into the single page app territory.
Test would be nice, though

PR #4686, opened 2023-10-17
Fix 500 error when loading projects list with PostgreSQL 13 and older

noliveleger:
## Description

Make the default projects sorting (i.e. `deployed`, `draft`, `archive`) work with older versions of PostgreSQL.

## Notes 
`OrderRandom` ORM utility has been renamed to a more explicit name: `OrderCustom` 

Supersedes #4685

### DIFF
--- a/kpi/utils/django_orm_helper.py
+++ b/kpi/utils/django_orm_helper.py
@@ -1,5 +1,7 @@
-# coding: utf-8
+from __future__ import annotations
+
 import json
+from typing import Optional, Literal
 
 from django.db.models import Lookup, Field
 from django.db.models.expressions import Func, Value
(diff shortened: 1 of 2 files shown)


kpi/utils/django_orm_helper.py
```diff
@@ -38,13 +38,19 @@ def __init__(self, expression: str, keyname: str, increment: int, **extra):
...
 
-class OrderRandom(Func):
+class OrderCustom(Func):
 
     function = 'array_position'
     template = '%(function)s(ARRAY%(order_list)s, %(expressions)s)'
     arity = 1
 
-    def __init__(self, expression: str, order_list: list, **extra):
+    def __init__(self, expression: str, order_list: list, array_type: str, **extra):
```
bufke, line 47: array_type: str is a required argument. It's allowed to be `Literal['varchar']` to get the desired varchar behavior else it will use the implicit cast. Thus to use implicit cast, it must be called with `array_type='unicorns'` to function as None isn't accepted.

How about making the type `Optional[literal['varchar']] = None`

If we're going to place the type hint, we should as well make the type useful. Then it can be omitted to use implicit cast and mypy would error if set to unicorns, which is silly.

bufke (changes requested):

bufke (approved):

PR #4689, opened 2023-10-18
Fix replaceSupportEmail function

### LINKED ISSUE #3910: Improve replaceSupportEmail utility function
## Description

At `utils.ts` file we have `replaceSupportEmail` function that should start using `support_email` from `environment/` API.

magicznyleszek:
## Checklist

3. [x] Ensure the tests pass

## Description

Makes the constanse setting `SUPPORT_EMAIL` being displayed in error notifications.

## Notes

Changes here:
- Had to move some things around to avoid circular dependency errors:
  - `handleApiFail` from `js/utils` to `js/api`
  - some text related functions were moved from `js/utils` to `js/textUtils`, because `replaceSupportEmail` requires `envStore`, and importing it into `js/utils` is causing dependency errors. I created new file for general text related utility functions and moved ones that will not require refactor.
- `envStore` is not using `actions` anymore, just making direct call with `api.ts`
- `replaceSupportEmail` is using email from environment endpoint
- removed `getEnvironment` Reflux action :)

## Related issues

Fixes #3910


jsapp/js/envStore.ts
```diff
@@ -122,9 +157,17 @@ class EnvStore {
...
-    actions.auth.getEnvironment.completed.listen(this.onGetEnvCompleted.bind(this));
-    actions.auth.getEnvironment();
+  async fetchData() {
+    try {
+      const response = await fetchGet<EnvironmentResponse>(ENV_ENDPOINT);
+      this.onGetEnvCompleted(response);
+    } catch (err) {
+      const errorObj = err as FailResponse;
+      handleApiFail(errorObj);
+    }
```
LMNTL, line 170: This `catch` block is kind of redundant - because `fetchGet()` (without the `notifyAboutError` option set to `false`) already calls `handleApiFail()`. So if I make the `/environment` endpoint error, I see two errors:

image

LMNTL (changes requested):

jsapp/js/envStore.ts
magicznyleszek: Right! Pushed a fix :)

LMNTL (approved):

PR #4691, opened 2023-10-20
Avoid query in the API call for myProjectsRoute

magicznyleszek:
## Checklist

3. [x] Ensure the tests pass

## Description

Stop including non-surveys in My Projects route search results.

## Notes

The problem is that for `myProjectsRoute` we pass `${ROOT_URL}/api/v2/assets/?q=${COMMON_QUERIES.s}` as `baseUrl`, and then we also add another `q=` to the final API call.

Things changed
- Added a helper function that plucks existing `q=` from `baseUrl`
- Removing `?…` from `baseUrl` and reapplying every query afterwards

I'm not sure if this is the best solution, as there were two possibilities:
- Option 1: what I did :) Which is mangling the `baseUrl` inside `customViewStore` (name might be confusing, it is handling both `customViewRoute` and `myProjectsRoute`)
- Option 2: instead of passing whole long url with `q=`, we would add another parameter to `customView.setUp`, so that `baseUrl` would not contain `q=` and `additionalQueries` would be passed.

### DIFF
--- a/jsapp/js/projects/projectViews/utils.tests.ts
+++ b/jsapp/js/projects/projectViews/utils.tests.ts
@@ -1,8 +1,43 @@
 import chai from 'chai';
 import type {ProjectsFilterDefinition} from './constants';
-import {removeIncorrectFilters, buildQueriesFromFilters} from './utils';
+import {
+  removeIncorrectFilters,
+  buildQueriesFromFilters,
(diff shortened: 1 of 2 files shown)


magicznyleszek:
Closing in favor of #4694 so it would go to next release ❤️

PR #4693, opened 2023-10-23
Create better error page

### LINKED ISSUE #3653: 500 error shows "Access denied" message
:monocle_face: Not quite:
image
**[end of screenshot]**

My preferences below—@Ig-Rebollo, what do you think?
1. Show the above message *only* for 404 errors.
2. For a 403, show the above message *without* the "Either…or this page simply doesn't exist" language. It exists and the client simply does not have access.
3. For a 401, redirect to the login page, or, if that's problematic, show the same message as (2).
4. For any 5xx, display some variation on "please try again, and contact support if you keep seeing this message." The header should be some kind of apology or "something went wrong", but definitely not "access denied".

Handy reference: https://httpstatusdogs.com/

See also #3318

Canonical discussion now at https://www.notion.so/kobotoolbox/Create-better-error-pages-9ae143db568c4b8ba6766d0750bd0faa

RuthShryock:
## Description

Created more detailed error pages so that users have a better understanding of what went wrong and how to get support.

## Notes

Updated the accessDenied.tsx file and added conditional statements to change the error message displayed based on the type of error received. 

## Related issues

Fixes #3653


jsapp/js/router/accessDenied.tsx
```diff
@@ -17,14 +17,48 @@ export interface AccessDeniedProps {
...
   let messageText;
+  let headerText;
+  let bodyText;
+  let errorNumber;
+  const loggedIn = t(
+    `Please try logging in using the header button or [contact the support team] if you think it's an error.`
+  );
+  const loggedOut = t(
+    `Please [contact the support team] if you think it's an error.`
+  );
```
LMNTL, line 28: The text of these two messages should be switched. When I go to the error page while logged in, it tells me to log in:

image

And it doesn't tell me to log in if I'm logged out:

image

LMNTL (changes requested):
Looking good overall! I have a couple changes to request.

jsapp/js/router/accessDenied.tsx
```diff
@@ -17,15 +17,51 @@ export interface AccessDeniedProps {
...
+  // Conditionally rendering error message based on number
+  switch (errorNumber){
+    case 403 || 401:
+      headerText = t(`Access Denied`);
+      bodyText = t(`You don't have access to this page.`);
+      if (sessionStore.isLoggedIn) {
+        messageText = loggedIn;
+      } else {
+        messageText = loggedOut;
+      }
```
LMNTL, line 47: It looks like this block also gets repeated two times. How would you feel about pulling this out into a function that returns either `loggedIn` or `loggedOut`?

jsapp/js/router/accessDenied.tsx
```diff
@@ -17,15 +17,51 @@ export interface AccessDeniedProps {
...
+      break;
+    case 404:
+      headerText = t(`Access Denied`);
```
LMNTL, line 50: Looking better! It looks like you set `headerText = t('Access Denied');` twice in this block; what if you initialized it at the start to `let headerText = t('Access Denied');` and removed those two lines?

LMNTL (changes requested):

LMNTL (approved):

PR #4696, opened 2023-10-24
Update ChangePlanView

LMNTL:
## Checklist

3. [X] Ensure the tests pass

## Description

Update Stripe backend to allow purchasing recurring storage add-ons.

## Notes

This PR also changes how the billing page works. Originally, it used the `ChangePlanView` endpoint to modify a subscription from our backend when upgrading/downgrading. Now, if a customer is upgrading, they're sent to Stripe to change plans, using the `CustomerPortalView` endpoint to generate a checkout link that works for any two prices (even if they're not included in the default Stripe portal configuration.)

A future PR (branch: real-stripe-tests) will include testing for endpoints using real (testmode) Stripe data, since the current tests are limited by the amount of Stripe functions that are being mocked.

## Related issues

Sub-PR of [kpi#4692](https://github.com/kobotoolbox/kpi/pull/4692/commits)

### DIFF
--- a/kobo/apps/stripe/views.py
+++ b/kobo/apps/stripe/views.py
@@ -149,7 +149,7 @@ class ChangePlanView(APIView):
         )
         return Response({'status': 'scheduled'})
 
-    def post(self, request):
+    def get(self, request):
         serializer = ChangePlanSerializer(data=request.query_params)
         serializer.is_valid(raise_exception=True)
         price = serializer.validated_data.get('price_id')
--- a/kobo/apps/stripe/views.py
+++ b/kobo/apps/stripe/views.py
@@ -160,7 +160,7 @@ class ChangePlanView(APIView):
         )
         return Response({'status': 'scheduled'})
 
-    def get(self, request):
+    def post(self, request):
         serializer = ChangePlanSerializer(data=request.query_params)
         serializer.is_valid(raise_exception=True)
         price = serializer.validated_data.get('price_id')
--- a/kobo/apps/stripe/views.py
+++ b/kobo/apps/stripe/views.py
@@ -54,12 +54,12 @@ class ChangePlanView(APIView):
     If the user is downgrading to a lower price, it will schedule the change at the end of the current billing period.
 
     <pre class="prettyprint">
-    <b>POST</b> /api/v2/stripe/change-plan/?subscription_id=<code>{subscription_id}</code>&price_id=<code>{price_id}</code>
+    <b>GET</b> /api/v2/stripe/change-plan/?subscription_id=<code>{subscription_id}</code>&price_id=<code>{price_id}</code>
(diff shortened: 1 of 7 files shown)


kobo/apps/stripe/views.py
```diff
@@ -308,42 +307,51 @@ def generate_portal_link(user, organization_id, price):
...
+                """
+                Recurring add-ons aren't included in the default billing configuration.
+                This lets us hide them as an 'upgrade' option for paid plan users.
+                Here, we try getting the portal configuration that lets us switch to the provided price.
+                """
+                current_config = next(
+                    (config for config in all_configs if (
+                            config['active'] and
+                            config['livemode'] == settings.STRIPE_LIVE_MODE and
+                            config['metadata'].get('portal_price', None) == price.id
```
noliveleger, line 320: Is `config['metadata'].get('portal_price', None)` always an integer? Ex: `'2' != 2`

kobo/apps/stripe/views.py
```diff
@@ -357,7 +365,13 @@ def generate_portal_link(user, organization_id, price):
                             },
                         ],
                         'subscription': customer['subscriptions__id'],
-                    }
+                    },
+                    'after_completion': {
+                        'type': 'redirect',
+                        'redirect': {
+                            'return_url': f'{settings.KOBOFORM_URL}/#/account/plan?checkout={price.id}',
```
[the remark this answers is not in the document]
LMNTL: Good eye! This one's slightly different from the other case - the `return_url` gets sent to Stripe by `stripe.billing_portal.Session.create` [here](https://github.com/kobotoolbox/kpi/pull/4696/files/cfc5a991d0dafa4047fb29995d97487476d6ca60..7eb5ff56db6e3321c7f17e3802687c650a4bae40#diff-9542dd6326eeb038c20ad61ef1c8c7df85d6adda4e0c258e2c430bc4d61e2ce2R378-R383). It tells Stripe [where to send the user after they finish their transaction](https://stripe.com/docs/api/customer_portal/sessions/create#create_portal_session-flow_data-after_completion-redirect-return_url). So this value never gets directly sent to the front-end.

kobo/apps/stripe/views.py
LMNTL: Stripe metadata values/keys are [always strings](https://stripe.com/docs/api/metadata), and `price.id` should be validated by the `CustomerPortalSerializer`. Let me know if I'm wrong, as is often the case 😅 The expected value of `price.id` would be `price_[some alphanumeric string]`, although someone messing around in the database could put in some other (string) value.

noliveleger (approved):

noliveleger:
CI is still broken for Python 3.10 because of coverage 422 error.

PR #4697, opened 2023-10-24
Make Button component more closely follow ARIA conventions

LMNTL:
## Checklist

3. [X] Ensure the tests pass

## Description

Make buttons in kpi click in response to the space bar or or enter key.

## Notes

The Button element was focusable, but not able to be activated by users navigating with the keyboard. Buttons now activate their `onClick` handler in response to `keyUp` events.

Also, the entire `<button/>` was disabled when `isDisabled` was set - this has been changed to using [`aria-disabled`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-disabled) to keep the element focusable, retaining the same styling.

## Related issues

Sub-PR of [kpi#4692](https://github.com/kobotoolbox/kpi/pull/4692/)

### DIFF
--- a/jsapp/js/components/common/button.tsx
+++ b/jsapp/js/components/common/button.tsx
@@ -158,7 +158,7 @@ const Button = (props: ButtonProps) => {
       className={classNames.join(' ')}
       type={props.isSubmit ? 'submit' : 'button'}
       aria-disabled={props.isDisabled}
-      onClick={props.onClick}
+      onClick={handleClick}
       onKeyUp={onKeyUp}
       {...additionalButtonAttributes}
     >
--- a/jsapp/js/components/common/button.tsx
+++ b/jsapp/js/components/common/button.tsx
@@ -43,7 +43,7 @@ ButtonToIconAloneMap.set('s', 'm');
 ButtonToIconAloneMap.set('m', 'l');
 ButtonToIconAloneMap.set('l', 'l');
 
-interface ButtonProps {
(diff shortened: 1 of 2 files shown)


jsapp/js/components/common/button.tsx
```diff
@@ -141,12 +141,25 @@ const Button = (props: ButtonProps) => {
...
+      handleClick(event);
+    }
+  };
+
   return (
     <button
       className={classNames.join(' ')}
       type={props.isSubmit ? 'submit' : 'button'}
-      disabled={props.isDisabled}
+      aria-disabled={props.isDisabled}
```
magicznyleszek, line 160: This change allows me to tab-navigate into a disabled button and activate it with spacebar/enter. You can test it at `#/forms/<uid>/settings/media`. My guess would be that `onClick={props.onClick}` has to be changed into `onClick={handleClick}` to fix that.

magicznyleszek: Just a thought: it would be good to announce on Zulip that this is our new direction as for disabled UI elements - they should be tab-navigable with visible focus, just `0.5` opacity and not activable. We could reference https://ux.stackexchange.com/questions/103239/should-disabled-elements-be-focusable-for-accessibility-purposes

magicznyleszek (changes requested):

magicznyleszek (approved):

PR #4705, opened 2023-10-27
Drop deprecated Moment.js package in favor of Day.js

LMNTL:
## Checklist

3. [X] Ensure the tests pass

## Notes

Moment.js has been in maintenance mode for a while, and its last update was in 2022. We've needed to replace it for a bit now - [Day.js](https://github.com/iamkun/dayjs) is a drop-in replacement that's actively maintained. It gives us better tree-shaking options (import plugins individually instead of using everything in `'moment'`) so we can reduce bundle size a bit, and it relies on immutable datatypes, unlike Moment.js.

This PR replaces the few places where we were using Moment.js with Day.js. In my testing, all functionality remains the same.

### DIFF
--- a/jsapp/js/main.es6
+++ b/jsapp/js/main.es6
@@ -5,6 +5,8 @@
 
 require('jquery-ui/ui/widgets/sortable');
 import dayjs from 'dayjs';
+// import specific locales for dayjs
+import 'js/locales';
 import AllRoutes from 'js/router/allRoutes';
 import RegistrationPasswordApp from './registrationPasswordApp';
 import {AppContainer} from 'react-hot-loader';
--- a/jsapp/js/main.es6
+++ b/jsapp/js/main.es6
@@ -4,7 +4,7 @@
  */
 
 require('jquery-ui/ui/widgets/sortable');
-import moment from 'moment';
+import dayjs from 'dayjs';
 import AllRoutes from 'js/router/allRoutes';
 import RegistrationPasswordApp from './registrationPasswordApp';
 import {AppContainer} from 'react-hot-loader';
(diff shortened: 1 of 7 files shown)


jsapp/js/main.es6
```diff
@@ -15,8 +15,8 @@ import {csrfSafeMethod, currentLang} from 'utils';
 require('../scss/main.scss');
 import Modal from 'react-modal';
 
-// Tell moment library what is the app language
-moment.locale(currentLang());
+// Tell dayjs library which language the app is using
+dayjs.locale(currentLang());
```
p2edwards, line 19: 1. Do we need to do anything differently to make sure the locales actually load? 
    - > By default, Day.js comes with English locale **only**. If you need other locales, you can load them on demand. 
      > — [DayJS i18n — Changing Locale](https://day.js.org/docs/en/i18n/changing-locale#:~:text=By%20default%2C%20Day.js%20comes%20with%20English%20locale%20only.)
2. We call `dayjs.locale(currentLang())` here and in utils.ts. Is there something going on with imports (tree shaking or lazy loading) that dictates we do this more than once? 
    - (submissionsCountGraph implicitly depends on the global locale setting because it uses localized formatting functions.)

LMNTL: 1. Good catch! I've added 'locales.ts' to cover importing those files in one place. I'm not 100% sure that it's the ideal solution - we can't import them dynamically based on the backend response, so it means another place that means updating if we add another language to the kpi interface. That could be changed to importing all locales, with the downside that we'd be loading some locales unnecessarily.

2. Nope, I just wanted to be excessively careful, but you're right that it's unnecessary. I've left the call in `main.es6`.

LMNTL:
I'm dropping this PR back into the draft status because Day.js degrades the quality of date translations.

On this PR:
image

On beta (with Moment):
image

I'll take a look at Luxon as an alternative.

PR #4707, opened 2023-10-27
Add override for error toast message to fetchData options

LMNTL:
## Checklist

3. [X] Ensure the tests pass

## Description

Improved how API errors are displayed and logged in the frontend.

## Notes

The goal is to improve `fetchData` error handling by allowing an optional custom error message to be provided for the toast notification. We pass the error text along with the HTTP status/code to Raven (as long as there's not a more detailed HTML-parsed error available.) This should provide a unified way to respond to simple fetch errors.

## Related Issues

Sub-PR of [kpi#4692](https://github.com/kobotoolbox/kpi/pull/4692/)


jsapp/js/api.ts
```diff
@@ -33,20 +34,24 @@ export function handleApiFail(response: FailResponse) {
     message = htmlDoc.getElementsByClassName('errormsg')?.[0]?.innerHTML;
   }
 
-  if (!message) {
-    message = t('An error occurred');
+  const htmlMessage = message;
```
jnm, line 37: I get a little bit confused at first with the multiple different things that `message` can represent and how `htmlMessage` weaves in between them. Could we have some separately-named variables like, for example, a `responseMessage` that would be either the plain-text or extracted-from-HTML `errormsg` from the server? Then, `message` perhaps could be only what's being built up to display to the user.

jsapp/js/api.ts
```diff
@@ -33,20 +34,24 @@ export function handleApiFail(response: FailResponse) {
...
+      message += '\n\n' + t('Your connection is offline');
     }
   }
 
+  // show the prettiest version of the error to the user
   notify.error(message);
 
-  window.Raven?.captureMessage(message);
+  // prefer sending the HTML error to Raven, since it should contain more detailed information
+  window.Raven?.captureMessage(htmlMessage || message);
```
jnm, line 54: Kind of the same gripe about variable naming above, and more of a comment about the human-readable names and comments vs. the logic itself: the "HTML error" here "works for Werkzeug Debugger only" according to the comment on lines 29-30. That only runs locally, and so we would not be interested in sending those messages to Raven → Sentry/GlitchTip, but rather showing them immediately to the front-end developer.

If I'm reading the logic right, though, `htmlMessage` actually also would include an unparsed, plain-text version of the server's response, and so that would perhaps have more helpful information, even in production, to send via Raven :ok_hand: 

(And it does seem that `notify.error(message)`  would show the unpretty parsed-from-HTML message from Werkzeug / `manage.py runserver` in the notification for the developer's benefit, following the seeming intent of the original code :+1: It'd only be hidden if a `toastMessage` were specified, right? Again, I think the logic's fine—it's just an unclear naming thing for me.)

Thanks :)

jnm (changes requested):

jsapp/js/api.ts
LMNTL: Good point! I ended up renaming `message` to `responseMessage`, since it gets initialized to `response.responseText`, and then moving things around a bit to fit. Let me know if that doesn't work.

jnm (approved):
Appreciated!

PR #4709, opened 2023-10-30
[TASK-114] Split out AccountFieldsEditor from AccountSettingsRoute

magicznyleszek:
## Checklist

3. [x] Ensure the tests pass

## Description

Internal code improvement in preparation for upcoming feture.

## Notes

I will be needing `AccountFieldsEditor` for the TOS feature. 

Changes here:
- New component `AccountFieldsEditor` that displays a selected list of user metadata fields and allows editing their values (locally)
- `AccountSettingsRoute` is using `AccountFieldsEditor` now
- Added error option to `KoboSelect`
- Updated `kobo-common` version to latest one (needed it for linter rules for enums)

Needed for #4711


RuthShryock (approved):
Everything looks good to me!

jsapp/js/components/common/koboSelect.tsx
```diff
@@ -70,6 +71,7 @@ interface KoboSelectProps {
   onChange: (newSelectedOption: string | null) => void;
   'data-cy'?: string;
   placeholder?: string;
+  error?: string;
```
LMNTL, line 74: This should be `error?: string | undefined;` - a bit counter-intuitive, but you explicitly pass it values that are possibly `undefined` (`props.errors?.country`, `props.errors?.gender`, etc.) instead of just omitting the prop. Should fix some TS errors.

jsapp/js/account/accountFieldsEditor.component.tsx
```diff
@@ -0,0 +1,281 @@
...
+
+          {/* Instagram */}
+          {isFieldToBeDisplayed('instagram') && (
+            <div className={styles.row}>
+              <TextBox
+                startIcon='logo-instagram'
+                placeholder={getLabel('instagram')}
+                value={props.values.instagram}
+                onChange={onAnyFieldChange.bind(onAnyFieldChange, 'instagram')}
+                errors={props.errors?.instagram}
```
LMNTL, line 273: The `errors` prop of all the `TextBox` components has a type mismatch - passing `undefined` when the prop is expecting `false`. One option:

```code
errors={props.errors?.instagram || false}
```

jsapp/js/account/accountFieldsEditor.component.tsx
```diff
@@ -0,0 +1,281 @@
...
+  }
+
+  /**
+   * Field will be displayed if it's enabled on Back end and it's not omitted
+   * in `displayedFields`.
+   */
+  function isFieldToBeDisplayed(name: UserFieldName) {
+    return (
+      name in metadata &&
+      (!props.displayedFields || props.displayedFields.includes(name))
```
LMNTL, line 87: This boolean chain could be condensed a bit using `props.displayedFields?.includes(name)`. But that's a small thing.

jsapp/js/account/account.constants.ts
```diff
@@ -0,0 +1,29 @@
...
+  | 'organization_website'
+  | 'sector'
+  | 'gender'
+  | 'bio'
+  | 'city'
+  | 'country'
+  | 'require_auth'
+  | 'twitter'
+  | 'linkedin'
+  | 'instagram';
```
LMNTL, line 13: This could get condensed to one line using `keyof` and `AccountFieldsValues`. You might then need to change UserFieldNames to a [Record type](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type) instead of a generic.

jsapp/js/account/accountFieldsEditor.module.scss
```diff
@@ -0,0 +1,9 @@
+@use 'scss/sizes';
+
+.row {
+  width: 100%;
```
LMNTL, line 4: On the old account profile page, the Country and City lines were both half-width:

image

Not sure if that's an intentional change or not.

jsapp/js/components/common/koboSelect.scss
```diff
@@ -100,6 +100,15 @@ $k-select-menu-padding: sizes.$x6;
...
 
+.k-select__error {
+  font-size: sizes.$x12;
+  line-height: 1.6;
+  font-weight: 400;
+  font-style: normal;
+  color: colors.$kobo-red;
+  margin-top: sizes.$x6;
+}
+
```
LMNTL, line 111: Could this message get positioned somehow so that the layout doesn't jump when it appears/disappears? Also, it would be nice if the error state had the same red border as the `TextBox`.

LMNTL (changes requested):

package.json
```diff
@@ -19,7 +19,7 @@
     "jquery": "^3.5.1",
     "jquery-ui": "1.12.1",
     "jszip": "^3.4.0",
-    "kobo-common": "https://github.com/kobotoolbox/kobo-common.git#semver:1.28.0",
+    "kobo-common": "https://github.com/kobotoolbox/kobo-common.git#semver:1.33.0",
```
p2edwards, line 22: [xref](https://github.com/kobotoolbox/kobo-common/commits/main): allow snake cases in enums

jsapp/js/account/accountFieldsEditor.component.tsx
magicznyleszek: Do you see this being raised as type mismatch in command line or in editor? I don't have any errors when running webpack. My Sublime Text is also handling this without problems:
Screenshot 2023-11-02 at 17 12 42

jsapp/js/components/common/koboSelect.tsx
magicznyleszek: Do you happen to have actual TS errors with this? I'm asking because I don't have errors, and I wonder if my setup is not catching the errors :)

jsapp/js/account/accountFieldsEditor.component.tsx
magicznyleszek: If there are no `props.displayedFields`, I need this to return `true`. If I use `props.displayedFields?.includes(name)` AFAIU this would work differently. I've added helpful comments :)

jsapp/js/account/account.constants.ts
magicznyleszek: I like that `Record<>`, thanks for showing it to me :) I've updated the code to use it and moved some types around

jsapp/js/account/accountFieldsEditor.module.scss
magicznyleszek: Right, forgot to update it. I will make this follow the figma designs as close as possible (given the fact that not all fields displayed in Figma exist in this PR)

jsapp/js/components/common/koboSelect.scss
magicznyleszek: The layout jumping is actually consistent with how the TextBox error is being displayed. I think this will require us having similar errors in all the common components, and then unfiying the functionality.

magicznyleszek: I added the styles for the select trigger (for all the types: outline, blue, and gray)

LMNTL: I agree that it's consistent with how the `TextBox` works, but I wonder if keeping it consistent is worth it in this case - especially since the refactor of the of the `KoboSelect` fields makes them look more distinct from the `TextBox`es. You could keep the same layout/position and avoid the content jumping by adding a very short animation to a height property/transform of the error message.

Still, I take your point - this shouldn't hold up this task👍I'll mark it as resolved.

jsapp/js/account/accountFieldsEditor.module.scss
magicznyleszek: So given these facts:
1. There is no Figma design for Account Settings
2. There is Figma design for Registration screen, which has a few more and few less fields than Account Settings
3. Some of the fields in Figma design for Registration screen don't exist yet in the code

I've changed the order of the fields and put some in the same row. I've also added `label` prop to `KoboSelect` and fixed some styles in it (could use some more tweaking and fixing, but I didn't want to go too far with it). And I've changed that `TextBox` L size to match the `Button` (again). It should fix most of the UI problems with the `TextBox` (that I've introduced with that big size)

LMNTL (changes requested):
Looking good! I have one last small change to request: the select elements on the old account profile could accept tab focus, but ones in this PR can't. It looks like adding `tabindex={0}` to `bem.KoboDropdown` (or some other appropriate element in the component tree) would fix that.

LMNTL (approved):

PR #4712, opened 2023-10-31
Create new TOS api endpoint

RuthShryock:
## Checklist

3. [X] Ensure the tests pass

## Description

API addition for TOS feature. 

## Notes

Added a new url for `me/tos/` and added an endpoint for POST requests which saves the current time.


kpi/tests/api/test_api_tos.py
```diff
@@ -0,0 +1,28 @@
...
+        self.user = User.objects.create_user(
+            username='someuser',
+            password='someuser',
+            is_active=True,
+        )
+        self.client.login(username='someuser', password='someuser')
+
+    def test_post(self):
+        # Prepare and send the request with empty payload
+        time = now().strftime('%Y-%m-%dT%H:%M:%SZ')
```
RuthShryock, line 21: Might want to consider declaring this string: `'%Y-%m-%dT%H:%M:%SZ` as a constant as it is used at least 10 times in the repository.

kpi/tests/api/test_api_tos.py
```diff
@@ -0,0 +1,28 @@
...
+
+class TOSTestCase(BaseTestCase):
+    def setUp(self) -> None:
+        self.url = reverse(self._get_endpoint('tos'))
+        self.user = User.objects.create_user(
+            username='someuser',
+            password='someuser',
+            is_active=True,
+        )
+        self.client.login(username='someuser', password='someuser')
```
bufke, line 17: You can shorten this

```python
from model_bakery import baker
...
self.user = baker.make('auth.User', username='spongebob', email='me@sponge.bob')
self.client.force_login(self.user)
```

kobo/apps/accounts/tests/test_api_tos.py
```diff
@@ -0,0 +1,26 @@
...
+    def setUp(self) -> None:
+        self.url = reverse(self._get_endpoint('tos'))
+        self.user = baker.make(
+            'auth.User', username='spongebob', email='me@sponge.bob'
+        )
+        self.client.force_login(self.user)
+
+    def test_post(self):
+        # Prepare and send the request with empty payload
+        time = now().strftime('%Y-%m-%dT%H:%M:%SZ')
```
p2edwards, line 19: This looks like an 8601 string. I wonder if there's something like an `.isoformat()` function in Python that's equivalent to this. 

- JS: [`.toDateString()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString)
- wikipedia: [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601)

p2edwards: (was talking to David and he says y'all considered this already)

- date formatting code like this is present elsewhere
    - it might make sense to use a constant or another function
    - but being that this api is short-term/internal, and the PR is pretty small, it makes sense to be consistent with it rather than departing from the norm or changing a bunch of unrelated code for uniformity

kpi/views/tos.py
```diff
@@ -0,0 +1,17 @@
+# coding: utf-8
```
bufke, line 1: We discussed moving this to kobo/apps as well

bufke (approved):

RuthShryock:
Tested 3.10 locally which passes all the tests. I think the build fail is related to this [issue](https://chat.kobotoolbox.org/#narrow/stream/4-Kobo-Dev/topic/GitHub.20CI.20Python.20broken/near/294990), but merging it for now.

PR #4715, opened 2023-11-03
Upgrade django-allauth to 0.57.0

LMNTL:
## Checklist

3. [X] Ensure the tests pass

## Notes

This PR bumps the version of `django-allauth` from 0.54.0 to 0.57.0 ([changelog](https://docs.allauth.org/en/latest/release-notes/recent.html#id4)). The major benefit is that SSO providers are able to be configured directly from the django admin, instead of through environment variables.

Since the `provider_id` field is required by the `{% provider_login_url %}` field, and the model itself can't be changed from our side, this PR also makes the `provider_id` field mandatory in the allauth admin.


bufke:
1. Create two open id providers. In my test one was Azure and the other was junk.
2. Go to http://localhost/accounts/login/

MultipleObjectsReturned at /accounts/login/

kobo/apps/django_allauth/admin.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class RequireProviderIdSocialAppForm(SocialAppForm):
+    def __init__(self, *args, **kwargs):
+        super(SocialAppForm, self).__init__(*args, **kwargs)
+        # require the provider_id in the admin, since we can't make it required on allauth's model
+        self.fields['provider_id'].required = True
+
+    def clean_provider_id(self):
+        reserved_keywords = ['kobo']
```
bufke, line 15: IMO if we are blocking in django admin, it would be fine to remove the "Please do not use `kobo` as the `id` value" check. But I'll defer to you.

bufke (changes requested):

kobo/apps/django_allauth/admin.py
LMNTL: Good point, especially since the code to parse OIDC sub-providers has been removed - it's now guarding against an impossible-to-achieve state (at least for someone not trying to break it.)

LMNTL: Actually, nevermind - we still need `check_socialaccount_providers`, because we also need to check the `id` of any social accounts, and that gets set on an env var.

PR #4725, opened 2023-11-11
Disable editing for media type questions with uploads

RuthShryock:
## Checklist

3. [X] Ensure the tests pass

## Description

This feature restricts editing for questions supporting uploads within the bulk-edit modal.

## Notes

Added the `disableEdit` function which determines if the Edit text button is disabled (and if so, the color changed as well) based on the `question.type`.


jsapp/js/components/modalForms/bulkEditSubmissionsForm.es6
```diff
@@ -212,6 +212,12 @@ class BulkEditSubmissionsForm extends React.Component {
     }
   }
 
+  disableEdit(questionType) {
```
LMNTL, line 215: Since this returns a boolean, the name of the function should get changed to match our [coding conventions](https://kobotoolbox.github.io/articles/kpi-frontend-development) - something like `isEditDisabled`.

jsapp/js/components/modalForms/bulkEditSubmissionsForm.es6
```diff
@@ -253,8 +259,9 @@ class BulkEditSubmissionsForm extends React.Component {
 
         <bem.SimpleTable__cell>
           <bem.KoboTextButton
-            m='blue'
+            m={this.disableEdit(question.type) ? 'grey' : 'blue'}
             onClick={this.selectQuestion.bind(this, question)}
+            disabled={this.disableEdit(question.type)}
```
LMNTL, line 264: It's preferable to avoid using the `disabled` attribute on buttons, since it hides information from screen readers and removes the button from the tab order. (The first part of [this article](https://a11y-101.com/development/aria-disabled) outlines a bit of the dilemma.) You can check out [this PR](https://github.com/kobotoolbox/kpi/pull/4697) for an example with the `<Button>` component.

I think you'd only need to take two steps to make that change:

- Use `aria-disabled` instead of `disabled`
- In the `onClick` handler function (`selectQuestion`), don't do anything if `this.disableEdit` is `true`

See [this internal thread](https://chat.kobotoolbox.org/#narrow/stream/4-Kobo-Dev/topic/Front-end.20coding.20style/near/291031) for additional details.

LMNTL (approved):

PR #4726, opened 2023-11-13
Optional TOS feature with /environment flag

RuthShryock:
## Checklist

3. [X] Ensure the tests pass

## Notes

Added `terms_of_service__sitewidemessage__exists` flag to the `/environment` endpoint so that the frontend can display the TOS feature depending on the server.

### DIFF
--- a/kpi/tests/api/test_api_environment.py
+++ b/kpi/tests/api/test_api_environment.py
@@ -117,7 +117,7 @@ class EnvironmentTests(BaseTestCase):
             'custom_password_localized_help_text': markdown(
                 I18nUtils.get_custom_password_help_text()
             ),
-            'terms_of_service__sitewidemessage__exists' : SitewideMessage.objects.filter(slug='terms_of_service').exists(),
+            'terms_of_service__sitewidemessage__exists': False,
         }
 
     def _check_response_dict(self, response_dict):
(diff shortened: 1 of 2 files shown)


kpi/tests/api/test_api_environment.py
```diff
@@ -116,6 +117,7 @@ def setUp(self):
             'custom_password_localized_help_text': markdown(
                 I18nUtils.get_custom_password_help_text()
             ),
+            'terms_of_service__sitewidemessage__exists' : SitewideMessage.objects.filter(slug='terms_of_service').exists(),
```
p2edwards, line 120: This appears to match the implementation logic — and so it doesn't exactly test the logic itself (or catch a regression)

Could we set up a couple of simple/exact test cases to cover each possibility? 

- (1) set a sitewide message called terms_of_service to a static value, assert env's terms_of_service__sitewidemessage__exists reads true
- (2) remove or don't set a sitewide message called terms_of_service, assert env's terms_of_service__sitewidemessage__exists reads false

kpi/tests/api/test_api_environment.py
```diff
@@ -305,3 +307,30 @@ def test_social_apps(self):
...
+        # Delete SitewideMessage object and check that it doesn't exsist
+        SitewideMessage.objects.filter(slug='terms_of_service').delete()
+        exists = SitewideMessage.objects.filter(
+            slug='terms_of_service'
+        ).exists()
+        self.assertFalse(exists)
+        self.assertEqual(
+            self.dict_checks['terms_of_service__sitewidemessage__exists'],
+            exists,
+        )
```
p2edwards, line 336: quickfix spelling

```suggestion
    def test_tos_sitewide_message_exists(self):
        # Create SitewideMessage object and check that it exists
        SitewideMessage.objects.create(
            slug='terms_of_service',
            body='tos agreement',
        )
        exists = SitewideMessage.objects.filter(
            slug='terms_of_service'
        ).exists()
        self.assertTrue(exists)
        self.assertNotEqual(
            self.dict_checks['terms_of_service__sitewidemessage__exists'],
            exists,
        )

    def test_tos_sitewide_message_does_not_exist(self):
        # Delete SitewideMessage object and check that it doesn't exist
        SitewideMessage.objects.filter(slug='terms_of_service').delete()
        exists = SitewideMessage.objects.filter(
            slug='terms_of_service'
        ).exists()
        self.assertFalse(exists)
        self.assertEqual(
            self.dict_checks['terms_of_service__sitewidemessage__exists'],
            exists,
        )
```

p2edwards (approved):

PR #4728, opened 2023-11-14
Feature: Per project anonymous access

duvld:
## Description

Moves account level option for anonymous form submission to a per-project level.

Adds a checkbox to the form landing page, under the submission type to allow anonymous submissions.

Removes:
- `require_auth` from the `/me` endpoint response
- "Require authentication when..." checkbox from account settings page
- Frontend logic that required `require_auth`

**Frontend ready for review**


LMNTL (changes requested):
Looking good so far 👍 I want to test the front-end locally first before I approve, but I'm having trouble with the `redis` config (which I'm talking with @noliveleger about.) In the meantime, I'm leaving some comments I had after looking through the code.

jsapp/js/components/formLanding.js
```diff
@@ -10,124 +10,173 @@ import sessionStore from 'js/stores/session';
...
+      anonymousPermissions: publicPerms,
+      anonymousSubmissions: anonCanAdd ? true : false,
+    });
+  }
+  updateAssetAnonymousSubmissions() {
+    const permission = this.state.anonymousPermissions.filter(
+      (perm) =>
+        perm.permission ===
+        permConfig.getPermissionByCodename(PERMISSIONS_CODENAMES.add_submissions).url
+    )[0];
```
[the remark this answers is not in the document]
duvld: This was a direct copy from `publicSharePermissions.component.tsx`, but since I'm adding more logic here with a state (instead of getting it from props) I think changing it to `.find()` is a better solution 👍

LMNTL (approved):

jsapp/js/components/permissions/publicShareSettings.component.tsx
```diff
@@ -64,19 +69,34 @@ class PublicShareSettings extends React.Component<PublicShareSettingsProps> {
...
           />
+          <a
+            href={envStore.data.support_url + HELP_ARTICLE_ANON_SUBMISSIONS_URL}
+            target='_blank'
+            data-tip={t(
+              'Allow anyone to see this form and add submissions. Click the icon to learn more.'
+            )}
+          >
+            <Icon size='s' name='help' color='storm' />
+          </a>
```
LMNTL, line 88: It looks like this could get pulled out into its own component - the only difference (that I can tell) between this and [this bit](https://github.com/kobotoolbox/kpi/pull/4728/files/80383614fe58675f5372b19b163087127c1971a8..f6db90642ae3efd0d10d612ac52cc0ac33a84f48#diff-419411dcab5d5a1abd3e658254891eab203590e61a388f4a173c4f090bd20501R428-R445) appear to be the `checked` and `onChange` props.

jsapp/js/components/permissions/sharingForm.scss
```diff
@@ -138,6 +139,25 @@ $s-gray-row-spacing: 10px;
...
+  padding-top: sizes.$x8;
+  border-bottom: 1px solid colors.$kobo-gray-96;
+
+  .toggle-switch {
+    padding-right: sizes.$x8;
+  }
+
+  .checkbox__label {
+    max-width: 100% !important;
+  }
```
LMNTL, line 158: Not sure if this needs to be removed or re-factored, but it might've been missed in the checkbox -> toggle refactor.

LMNTL (changes requested):

LMNTL (approved):

PR #4731, opened 2023-11-17
Rename partial permissions variables

magicznyleszek:
## Checklist

3. [x] Ensure the tests pass

## Description

Internal cleanup

## Notes

Renamed everything related to partial permissions (for a selected list of users) to make room for more partial permissions (for particular responses). So basically we use `…PartialByUsers` instead of `…Partial`, because we will be adding `…PartialByResponses` :)

Needed for `feature/extend-partial-perms`


jsapp/js/components/permissions/permConstants.ts
```diff
@@ -81,20 +81,20 @@ type CheckboxNameRegular =
...
+  | 'submissionsValidatePartialByUsers'
+  | 'submissionsDeletePartialByUsers';
+/** All checkboxes names combined. */
+export type CheckboxNameAll = CheckboxNameRegular | CheckboxNamePartialByUsers;
+/** Name of lists of usernames for a partial permissions checkboxes. */
+export type PartialByUsersListName =
+  | 'submissionsViewPartialByUsersList'
+  | 'submissionsEditPartialByUsersList'
+  | 'submissionsDeletePartialByUsersList'
+  | 'submissionsValidatePartialByUsersList';
```
LMNTL, line 97: If you turned `CheckboxNamePartialByUsers` and `PartialByUsersListName` into `enum`s, it looks like you [could delete](https://github.com/kobotoolbox/kpi/pull/4731/files#diff-16bcdd306655aaf1598bac0a5f0932c51c2f9bad9572e7fc721aa749b170f960R199-R230) `getPartialByUsersListName()` and `getPartialByUsersCheckboxName()`. (A frozen `const` object would also work, but at that point it's just an `enum` with more steps.)

LMNTL (changes requested):

jsapp/js/components/permissions/permConstants.ts
magicznyleszek: Nice idea, I will try doing it on `feature-extend-partial-perms--add-partial-by-responses` as I think I have more of these functions there already

PR #4732, opened 2023-11-17
Organization_type skip logic in signup form and back-end validation

RuthShryock:
## Checklist

3. [X] Ensure the tests pass

## Notes

Added skip logic and tests based on if the superuser has enabled and/or required the `organization_type` field.

In forms.py:
- If the organization type is enabled, initialize the signup form with the `organization_type` required and other organization fields to not be required
- When the user submits the form, we check if the `organization_type` is none and if it is not, we require the `organization` and `organization_website` fields. 

In current_user.py:
- When validating the extra_details, if the `organization_type` field is enabled and the value is none, change the required values of `organization` and `organization_website` to `false`.

### DIFF
--- a/kobo/apps/accounts/forms.py
+++ b/kobo/apps/accounts/forms.py
@@ -131,14 +131,27 @@ class KoboSignupMixin(forms.Form):
 
             field = self.fields[field_name]
             # Part of 'skip logic' for organization fields
+            #     The 'Organization Type' dropdown hides 'Organization' and
+            # 'Organization Website' inputs if the user has selected
+            # 'I am not associated with an organization'. In that case the
+            # back end accepts omitted or blank values for organization and
+            # organization_website, even if they're 'required'.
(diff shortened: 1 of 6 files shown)


kobo/apps/accounts/forms.py
```diff
@@ -129,7 +130,16 @@ def __init__(self, *args, **kwargs):
                 continue
 
             field = self.fields[field_name]
-            field.required = desired_field.get('required', False)
+            # Part of 'skip logic' for organization fields
+            if (
+                desired_metadata_fields.get('organization_type')
+                and field_name in ['organization', 'organization_website']
+                and desired_field.get('required', False)
```
p2edwards, line 137: double check that 'False' is right here; looks weird during call — make sure this is covered by tests

p2edwards, line 137: None is falsey so remove False (noisy)

p2edwards (approved):

PR #4740, opened 2023-11-28
Add tooltip Component for disabled buttons

### LINKED ISSUE #4611: Improve Button disabled styles
## Description

We need the tooltip for disabled button to work.

## Ideas

We can modify the `Button` component to stop using `pointer-events` (and `opacity: 0.5`) and thus allowing tooltips to work on disabled button. We would need to add some `:not([disabled])` for some selectors (like hover styles or cursor) - to be checked if doable.

Make tooltips readable. In our design system, disabled buttons are styled with `opacity: 0.5`. All DOM descendents including tooltips can never be more than 50% opaque after that, so tooltips would need a wrapper anyway or do css hacks case-per-case (possible here because of single icon buttons, but not feasible in the general case.) I don't see a good way forward with this, other than to normalize tooltip wrappers, or redesign disabled buttons (and any other input fields where this would apply).

We might alternatively make the `button.tsx` file wrap the button if `tooltip` was provided.

## The override

For now let's use `<span data-tip={t('Blah blah')}>` wrapper for button :)

## Additional details

See https://github.com/kobotoolbox/kpi/pull/4580#discussion_r1305595918

RuthShryock:
## Checklist

3. [X] Ensure the tests pass

## Notes

Added a `tooltip` component which has been added as an attribute to the `Button` component and wrote a tooltip.stories.tsx for documentation. Updated all the projectQuickAction and projectBulkAction icons to use the new tooltip component. 

## Related issues

Fixes #4611


jsapp/js/components/common/tooltip.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+
+interface TooltipProps {
+  text?: string;
+  ariaLabel?: string;
+  className?: string;
```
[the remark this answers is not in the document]
magicznyleszek: And please add the option to change it in the stories :)

magicznyleszek (changes requested):

jsapp/js/components/common/tooltip.tsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+
+export interface TooltipProps {
+  text: string;
+  ariaLabel: string;
+  alignment?: 'right' | 'left' | '';
```
magicznyleszek, line 6: Let's have `export type TooltipAlignment = 'right' | 'left' | 'center';`. This way we can reference it elsewhere (e.g. in `button.tsx`) without the need to duplicate ourselves.

magicznyleszek: Also `'center'` is better than empty string with comments "it means centered" :)

jsapp/js/components/common/tooltip.tsx
```diff
@@ -0,0 +1,36 @@
...
+ */
+const Tooltip: React.FC<TooltipProps> = ({
+  text,
+  ariaLabel,
+  alignment,
+  children,
+}) => (
+  <span
+    data-tip={text}
+    className={`${alignment}-tooltip`}
```
magicznyleszek, line 30: To make `'center'` the de facto default, you can have `${alignment || 'center'}-tooltip`. It's ok if `center-tooltip` doesn't have any styles

jsapp/js/components/common/tooltip.tsx
```diff
@@ -0,0 +1,36 @@
...
+ * component, that provides styling but not other functionaltiy such as
+ * allowing tooltips to work on disabled buttons.
+ *
+ * @param {string} text - Content of the tooltip.
+ * @param {string} ariaLabel - Accessible label for screen readers.
+ * @param {'right' | 'left' | ''} alignment - Position of the tooltip (centered
+ * if undefined or set to an empty string).
+ * @param {React.ReactNode} children - Button that triggers the tooltip when
+ * hovered.
+ */
```
magicznyleszek, line 21: This is great, just insted of all those `@param`s, we can describe each of parameters inside `TooltipsProps` interface - adding `/** comment */` above each of them. See `textBox.tsx` for example code. Sorry for not being clearer.

That double asterisk comment is being understood by text editors and being displayed, for example in Sublime Text:
Screenshot 2023-12-11 at 13 29 19

magicznyleszek (changes requested):

magicznyleszek (approved):

jsapp/js/components/common/button.tsx
```diff
@@ -154,27 +154,39 @@ const Button = (props: ButtonProps) => {
...
-
-      {/* Ensures only one icon is being displayed.*/}
-      {!props.startIcon && props.endIcon && (
-        <Icon name={props.endIcon} size={iconSize} />
-      )}
-
-      {props.isPending && (
-        <Icon name='spinner' size={iconSize} classNames={['k-spin']} />
+    <>
+      {props.tooltip !== undefined && (
```
magicznyleszek, line 158: I've noticed that if you don't add tooltip, this renders nothing. We need the button to be rendered with or without a tooltip :)

magicznyleszek: I thought that somethign is wrong with Storybook, because it wasn't rendering anything for Button, so I've added a commit that improves `button.stories.tsx`. Please take a look at what it does when testing the tooltip:
Screenshot 2023-12-13 at 15 19 35

magicznyleszek (changes requested):

magicznyleszek (approved):

PR #4742, opened 2023-11-29
Enable range selection of checkboxes

### LINKED ISSUE #3995: Enable selecting range of checkboxes
## Feature request: Enable quick selection of checkbox ranges

I think especially with the batch edit features, it would be really handy to be able to select a range of checkboxes.
Often this can be done by activating a checkbox, then keeping _Shift_ pressed while selecting another - checkboxes between the selected two will be selected as well.

This functionality does not work in the _kpi_ data table, but I think it would be a helpful addition.

Right now, you can only select all visible checkboxes or have to select all desired checkboxes individually.

image

RuthShryock:
## Checklist

3. [X] Ensure the tests pass

## Notes

Modified the `bulkUpdateChange` function to handle range selection by updating the state of selected checkboxes between the initial and final checkbox selection while the `Shift` key is pressed. 

## Related issues

Fixes #3995

### DIFF
--- a/jsapp/js/components/submissions/table.es6
+++ b/jsapp/js/components/submissions/table.es6
@@ -1228,22 +1228,34 @@ export class DataTable extends React.Component {
    * @param {boolean} isChecked
    */
   bulkUpdateChange(sid, isChecked) {
-    const {selectedRows, lastChecked} = this.state;
+    const {selectedRows, lastChecked, shiftSelection} = this.state;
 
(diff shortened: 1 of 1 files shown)


LMNTL (changes requested):
This works pretty well! The only change I would request is fairly small: it might be better if selecting the range only happened if you held the shift key down while selecting the start *and* end checkboxes (instead of just the end checkbox.)

PR #4746, opened 2023-12-04
onUpdateStatus: submission_ids when filter active

### LINKED ISSUE #4745: Batch update of all submissions doesn't work when filter is in place
## Description

Hi, when I try to batch update the status of all submissions, this works. The payload is the following:

`{"confirm":true,"validation_status.uid":null}`

However, when a filter is active to select a subset of submissions, the batch update doesn't work. The payload contains `"query"` then instead of `"confirm":true`:

`{"query":{"_submission_time":"2022"},"validation_status.uid":"validation_status_approved"}`

Selecting individual submissions works, but selecting and batch updating all submissions does not. 

## Steps to Reproduce

1. Select all submissions and apply a batch status update -> works
2. Filter the submissions by a field that has a filter, then select all and apply a batch status update -> doesn't work

## Expected behavior 

The batch update for all matching submissions should also work when a filter is applied.

## Actual behavior 

The batch update for all submissions fails, when a filter is active.

## Additional details

See below video for a demonstration of the issue.

[](https://github.com/kobotoolbox/kpi/assets/121946/65639079-9d1f-49c6-a86e-e6668ba875e8)

jeverling:
Fixes #4745

### DIFF
--- a/jsapp/js/components/submissions/tableBulkOptions.es6
+++ b/jsapp/js/components/submissions/tableBulkOptions.es6
@@ -58,17 +58,9 @@ class TableBulkOptions extends React.Component {
     // setting empty value requires deleting the statuses with different API call
     const apiFn = newStatus === null ? actions.submissions.bulkDeleteStatus : actions.submissions.bulkPatchStatus;
 
-    if (this.props.selectedAllPages) {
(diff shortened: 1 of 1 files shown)


magicznyleszek:
Fixed by #4833

PR #4748, opened 2023-12-05
Fix /organizations/{ORG ID}/service_usage/ and /stripe/products/ endpoint cache misses

LMNTL:
## Checklist

3. [X] Ensure the tests pass

## Description

Always cache the `/organizations/{ORG ID}/service_usage/` and `/stripe/products/` endpoints, even if a cookie has changed.

## Notes

Fixes an error that was causing the `/organizations/{ORG ID}/service_usage/` and `/stripe/products/` endpoints to never serve cached data on the NA/EU servers. The problem was identified as a constantly-updating Google Analytics cookie - django caching takes the `Vary` headers into account when deciding to serve from cache ([Django docs link](https://docs.djangoproject.com/en/3.1/topics/cache/#using-vary-headers)), and different middleware will add `Vary` headers (including `Vary: cookie`).

In addition, the `_now` property of the `ServiceUsageSerializer` and `AssetUsageSerializer` was being initialized incorrectly; the value was only generated at the moment Python loaded the model. This has been changed to use the `__init__` method.

## Testing

Make sure you have ENDPOINT_CACHE_DURATION in your environment variables (measured in seconds.) It should be a long enough duration for you to test successive requests.

To reproduce the issue locally for testing, add the following line to `/jsapp/js/account/usage.api.ts`:
```diff
const ORGANIZATION_USAGE_URL =
   '/api/v2/organizations/##ORGANIZATION_ID##/service_usage/';

 export async function getUsage(organization_id: string | null = null) {
+  document.cookie = Date.now().toString();
   if (organization_id) {
     return fetchGet<UsageResponse>(
       ORGANIZATION_USAGE_URL.replace('##ORGANIZATION_ID##', organization_id),
```

You can then go to Account Settings -> Usage in kpi to see the API request in action. Make sure the `date` header on each response stays the same between successive requests. Then make a submission and confirm that the count stays the same.

Log in as another user - confirm that you see the second user's count on the Usage page, instead of the first user's.


kpi/views/v2/service_usage.py
```diff
@@ -14,15 +10,14 @@
 from kpi.utils.object_permission import get_database_user
 
 
-# @method_decorator(cache_page(settings.ENDPOINT_CACHE_DURATION), name='list')
-# @method_decorator(vary_on_cookie, name='list')
```
noliveleger, line 18: ❤️

PR #4749, opened 2023-12-06
Require "edit submissions", not "add submissions", for working with transcripts and translations

jnm:
## Checklist

3. [x] Ensure the tests pass

## Description

Doing any kind of work with transcripts or translations—whether adding a new one or editing an existing one—now always requires the "edit submissions" permission.

### DIFF
--- a/kobo/apps/subsequences/api_view.py
+++ b/kobo/apps/subsequences/api_view.py
@@ -1,4 +1,5 @@
 import json
+from copy import deepcopy
 
 from jsonschema import validate
 from jsonschema.exceptions import ValidationError as SchemaValidationError
@@ -48,8 +49,17 @@ def _check_asr_mt_access_if_applicable(user, posted_data):
                     raise PermissionDenied('ASR/MT features are not available')
 
 
+class AdvancedSubmissionPermission(SubmissionPermission):
+    """
+    Regular `SubmissionPermission` maps POST to `add_submissions`, but
+    `change_submissions` should be required here
+    """
+    perms_map = deepcopy(SubmissionPermission.perms_map)
+    perms_map['POST'] = ['%(app_label)s.change_%(model_name)s']
+
+
 class AdvancedSubmissionView(APIView):
-    permission_classes = [SubmissionPermission]
+    permission_classes = [AdvancedSubmissionPermission]
     queryset = Asset.objects.all()
     asset = None
 
--- a/kobo/apps/subsequences/tests/test_submission_extras_api_post.py
+++ b/kobo/apps/subsequences/tests/test_submission_extras_api_post.py
@@ -14,6 +14,13 @@ from kobo.apps.languages.models.language import Language, LanguageRegion
 from kobo.apps.languages.models.transcription import (
     TranscriptionService, TranscriptionServiceLanguageM2M)
 from kpi.utils.fuzzy_int import FuzzyInt
+from kpi.constants import (
+    PERM_ADD_SUBMISSIONS,
+    PERM_CHANGE_ASSET,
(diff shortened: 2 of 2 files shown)


kobo/apps/subsequences/api_view.py
```diff
@@ -48,8 +49,17 @@ def _check_asr_mt_access_if_applicable(user, posted_data):
...
 
 
+class AdvancedSubmissionPermission(SubmissionPermission):
+    """
+    Regular `SubmissionPermission` maps POST to `add_submissions`, but
+    `change_submissions` should be required here
+    """
+    perms_map = deepcopy(SubmissionPermission.perms_map)
+    perms_map['POST'] = ['%(app_label)s.change_%(model_name)s']
+
```
noliveleger, line 59: I know if only few lines but I would have preferred to keep the permission class in its own file (to be consistent with the rest of the app).

kobo/apps/subsequences/tests/test_submission_extras_api_post.py
```diff
@@ -88,6 +95,115 @@ def test_translation_revisions_stored_properly(self):
...
+        resp = self.client.post(schema['url'], modified, format='json')
+        assert resp.status_code == 404
+
+        self.asset.assign_perm(other_user, PERM_CHANGE_ASSET)
+        resp = self.client.post(schema['url'], modified, format='json')
+        assert resp.status_code == 404
+
+        self.asset.assign_perm(other_user, PERM_VIEW_SUBMISSIONS)
+        resp = self.client.post(schema['url'], modified, format='json')
+        assert resp.status_code == 403
```
noliveleger, line 172: Not related to this PR but I wondering why `PERM_VIEW_ASSET`, `PERM_CHANGE_ASSET` return a 404 while `PERM_VIEW_SUBMISSIONS` returns a 403. I was assuming that as soon as user had `PERM_VIEW_ASSET`, users would receive a 403 when they don't have required permissions on nested objects.

noliveleger (commented):
PR is ok and LGTM. 
Just have comments / questions before merging.  Feel free to merge it if my comments are not relevant.

kobo/apps/subsequences/tests/test_submission_extras_api_post.py
jnm: Good question; I'd assume it's just the usual DRF behavior of:
* Did you send a GET for a specific object that you are not allowed to see? We're not telling you whether or not it exists, so 404
* Did you try to take a disallowed action (POST, PATCH) on an object that you *are* allowed to see? → 403

I didn't expect nested objects to behave differently, but we can look into it more if you think something doesn't make sense

kobo/apps/subsequences/api_view.py
jnm: Yeah, I thought about it. I didn't want to make a tiny file in the `subsequences` app, and I wasn't sure about putting it in the same file as `SubmissionPermission`, where the name "advanced" might be confusing to read without the context of `subsequences`. For best readability, I would just leave it here next to the only thing that uses it, but if more extensive `subsequences` permission work is needed in the future, or the permissions class gets used in other places, then I'd be in favor of moving to a separate file.

kobo/apps/subsequences/tests/test_submission_extras_api_post.py
noliveleger: Right! that's what I'm expecting too. 

I missed [an important line](https://github.com/kobotoolbox/kpi/blob/require-change-not-add-for-transcripts-translations/kpi/permissions.py#L343)! 

I was thinking the model was `Asset`.  Everything is good.

noliveleger (approved):

PR #4751, opened 2023-12-06
[TASK-383] [TASK-330] Preserve response when updating Analysis Question definition

magicznyleszek:
## Checklist

3. [x] Ensure the tests pass

## Notes

When we get the updated question object from API endpoint, it doesn't have the response in it (as it comes from other endpoint). Thus we need to update everything in the stored question, but the response.

This is built atop the #4750 PR.

### DIFF
--- a/jsapp/js/components/processing/analysis/editors/analysisQuestionEditor.component.tsx
+++ b/jsapp/js/components/processing/analysis/editors/analysisQuestionEditor.component.tsx
@@ -134,13 +134,22 @@ export default function AnalysisQuestionEditor(
           updatedQuestions
         );
 
-        // Step 4: update reducer's state with new list after the call finishes
-        analysisQuestions?.dispatch({
-          type: 'updateQuestionCompleted',
-          payload: {
-            questions: getQuestionsFromSchema(response?.advanced_features),
-          },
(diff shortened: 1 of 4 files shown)


jsapp/js/components/processing/analysis/editors/analysisQuestionEditor.component.tsx
```diff
@@ -134,13 +134,22 @@ export default function AnalysisQuestionEditor(
...
+        if (currentNewQuestion) {
+          // Step 4: update reducer's state with new list after the call finishes
+          analysisQuestions?.dispatch({
+            type: 'updateQuestionCompleted',
+            payload: {question: currentNewQuestion},
+          });
+        } else {
+          // This should never happen :) I.e. the list of questions from
+          // `response` will include the question, it's just the `.find`
+          // that has a possibility to return `undefined` :shrug:
```
LMNTL, line 151: I know this should in theory be an unreachable code path, but would it make sense to just `throw` an error here so it bubbles up to the `catch` block?

LMNTL (changes requested):
Just one small change requested, otherwise good to merge 👍

PR #4757, opened 2023-12-08
Partial submissions bug fixes

bufke:
- Fix bug in MongoHelper.get_permission_filters_query where a list of filters would not be transformed correctly into an pymongo query
- Remove duplicate MongoHelper.is_attribute_invalid
- partial permissions support simple OR filters and support implied partial permissions
- Includes https://github.com/kobotoolbox/kpi/pull/4718

# DO NOT MERGE

- [ ] Sharing user interface matches this new partial permissions syntax
- [ ] Unit test for and/or logic on submissions.

### DIFF
--- a/kpi/tests/api/v2/test_api_asset_permission_assignment.py
+++ b/kpi/tests/api/v2/test_api_asset_permission_assignment.py
@@ -813,13 +813,13 @@ class ApiBulkAssetPermissionTestCase(BaseApiAssetPermissionTestCase):
                     {
                         'url': PERM_VALIDATE_SUBMISSIONS,
                         'filters': [
-                            [{'my_question': 'my_response1'}, {'my_question': 'my_response2'}]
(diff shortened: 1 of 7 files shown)


kpi/tests/api/v2/test_api_asset_permission_assignment.py
```diff
@@ -659,3 +657,241 @@ def test_no_assignments_saved_on_error(self):
...
+        )
+
+        bulk_endpoint = reverse(
+            self._get_endpoint('asset-permission-assignment-bulk-assignments'),
+            kwargs={'parent_lookup_asset': self.asset.uid}
+        )
+        response = self.client.post(
+            bulk_endpoint, assignments, format='json'
+        )
+        # It would be better to 400 here
```
noliveleger, line 896: Agreed!

PR #4761, opened 2023-12-11
[TASK-335] [TASK-337] [TASK-338] Disable non permitted UI

magicznyleszek:
## Checklist

3. [x] Ensure the tests pass

## Description

The UI for editing Analysis Questions will be disabled for users with insufficient permissions. Also fixes issue with duplicated error notifications.

## Notes

Changes here:
- Avoid displaying duplicated error notifications for editing questions definitions and for editing responses
- Disable analysis question deleting, editing and reordering if user doesn't have `manage_asset` permission
- Disable editing analysis question responses if user doesn't have `change_submissions` permission

### DIFF
--- a/jsapp/js/components/processing/analysis/list/analysisQuestionRow.component.tsx
+++ b/jsapp/js/components/processing/analysis/list/analysisQuestionRow.component.tsx
@@ -66,7 +66,7 @@ export default function AnalysisQuestionRow(props: AnalysisQuestionRowProps) {
   // Responding to analysis question requires `edit_submissions` permission.
   const hasEditSubmissionsPermissions = (() => {
     const asset = assetStore.getAsset(singleProcessingStore.currentAssetUid);
-    return userCan('edit_submissions', asset);
+    return userCan('change_submission', asset);
   })();
 
   const isDragDisabled =
--- a/jsapp/js/components/processing/analysis/list/analysisQuestionRow.component.tsx
+++ b/jsapp/js/components/processing/analysis/list/analysisQuestionRow.component.tsx
@@ -24,6 +24,8 @@ import classnames from 'classnames';
 import singleProcessingStore from 'js/components/processing/singleProcessingStore';
 import {handleApiFail} from 'js/api';
 import type {FailResponse} from 'js/dataInterface';
+import assetStore from 'js/assetStore';
+import {userCan} from 'js/components/permissions/utils';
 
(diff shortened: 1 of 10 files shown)


jsapp/js/components/processing/analysis/list/analysisQuestionRow.component.tsx
```diff
@@ -55,7 +57,20 @@ export default function AnalysisQuestionRow(props: AnalysisQuestionRowProps) {
     return null;
   }
 
-  const isDragDisabled = analysisQuestions.state.isPending;
+  // Reordering analysis questions requires `manage_asset` permission.
+  const hasManagePermissions = (() => {
+    const asset = assetStore.getAsset(singleProcessingStore.currentAssetUid);
+    return userCan('manage_asset', asset);
+  })();
```
LMNTL, line 64: It looks like this IIFE gets duplicated in jsapp/js/components/processing/analysis/responseForms/commonHeader.component.tsx - might be good to import it instead. Or pull it out into a custom hook.

LMNTL, line 64: Not sure if this is a bug, an upstream issue, or just me doing something I shouldn't: if you remove the `manage_asset` permission from a user, the questions still show up as being editable in the UI.

some user in admin

what some user sees in the UI

But then they get a 403 when submitting a change:
verboten

It looks like this is because `userCan()` assumes the asset owner has full permissions:
```ts
  // If you own the asset, you can do everything with it
  if (asset.owner__username === currentUsername) {
    return true;
  }
```

LMNTL (changes requested):

jsapp/js/components/processing/analysis/list/analysisQuestionRow.component.tsx
jnm: Thanks, @LMNTL: this is good to flag, but in this case, it's the third option :wink: 
We actually [grant all model-level permissions](https://github.com/kobotoolbox/kpi/blob/92b84b5c24f6c4ee78508e1eb672e88df89c2cbc/kpi/signals.py#L22-L37) to all users when they are first created and then manage permissions at the object level. We discourage people from mucking around in the Django admin permission settings and emphasize that changes there are unsupported. We also broadly assume that the owner has full control of the objects they own.

PR #4763, opened 2023-12-12
Add endpoint to test kpi health

RuthShryock:
## Checklist

3. [X] Ensure the tests pass

## Description

Endpoint which makes no connections to databases or other external services and is helpful for HTTP health checks.

### DIFF
--- a/kobo/apps/service_health/test_service_health_minimal.py
+++ b/kobo/apps/service_health/test_service_health_minimal.py
@@ -7,14 +7,14 @@ from django.urls import reverse
 class ServiceHealthMinimalTestCase(TestCase):
     url = reverse('service-health-minimal')
 
-    @responses.activate
     def test_service_health_minimalhealth(self):
-        responses.add(responses.GET, settings.ENKETO_INTERNAL_URL, status=200)
-        responses.add(
-            responses.GET,
-            settings.KOBOCAT_INTERNAL_URL + 'service_health/minimal/',
-            status=200,
-        )
         res = self.client.get(self.url)
-        self.assertContains(res, "ok")
+        self.assertEqual(res.status_code, 200)
+
+        # Check that the response content is "ok"
+        expected_content = 'ok'
+        actual_content = res.content.decode('utf-8')
+        self.assertEqual(actual_content, expected_content)
+
+        # Ensure that no database queries were executed
         self.assertNumQueries(0)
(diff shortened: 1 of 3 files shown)


kobo/apps/service_health/test_service_health_minimal.py
```diff
@@ -0,0 +1,18 @@
...
+    url = reverse('service-health-minimal')
+
+    def test_service_health_minimal(self):
+        res = self.client.get(self.url)
+        self.assertEqual(res.status_code, 200)
+
+        # Check that the response content is "ok"
+        expected_content = 'ok'
+        actual_content = res.content.decode('utf-8')
+        self.assertEqual(actual_content, expected_content)
```
bufke, line 15: You can replace all of this with `self.assertContains(res, 'ok')`  [Docs](https://docs.djangoproject.com/en/5.0/topics/testing/tools/#django.test.SimpleTestCase.assertContains)

kobo/apps/service_health/test_service_health_minimal.py
```diff
@@ -0,0 +1,18 @@
...
+
+        # Ensure that no database queries were executed
+        self.assertNumQueries(0)
```
bufke, line 18: Look up the docs here, it needs to be used as a context manager. If you want to test that it does something, cause the test to fail by putting in some random SQL query.

bufke (changes requested):

kobo/apps/service_health/urls.py
```diff
@@ -1,7 +1,9 @@
 from django.urls import path
 
 from .views import service_health
+from .views import service_health_minimal
 
 urlpatterns = [
     path('service_health/', service_health, name="service-health"),
+    path('service_health/minimal', service_health_minimal, name="service-health-minimal"),
```
bufke, line 8: Add a `/` at the end. That's the Django convention.

kobo/apps/service_health/test_service_health_minimal.py
```diff
@@ -0,0 +1,13 @@
+from django.test import TestCase
+from django.urls import reverse
+
+
+class ServiceHealthMinimalTestCase(TestCase):
+    url = reverse('service-health-minimal')
+
+    def test_service_health_minimal(self):
+        with self.assertNumQueries(0):
```
bufke, line 9: I would add a docstring here explaining what this api is and why it shouldn't have any queries. You can omit the comment "Check that the response content contains" because the documentation for assertContains is available in the django project. So the aim is to explain why our code is doing and why. But we don't need to explain how django works, even if that might be helpful to someone very new. The risk here is documenting too much would actually make the documentation less readable - eventually.

bufke (approved):
Approved to merge pending addressing my comments.

PR #4796, opened 2024-01-05
Url validation and required input improvements

jamesrkiger:
This PR adds URL validation for the organization website field in the ToS and account settings forms. It also includes and finishes @p2edwards work on displaying red asterisks for required fields instead of adding (required) in the label. I have added `aria-required` to the relevant inputs for accessibility.


p2edwards (approved):
Awesome. Thanks!

PR #4813, opened 2024-01-19
Add handling for plan changes and cancellations

jamesrkiger:
## Checklist

3. [x] Ensure the tests pass

## Description

This PR adds handling for scheduled changes to a user's subscription: cancellation, product changes (currently these are always downgrades, but the logic here doesn't assume this), and price changes (currently these are always downgrades from annual to monthly. The logic **does** assume this. We would need to add some more logical parsing to handle other scenarios), It also subsumes the "your plan" section into the "Your usage" section, rather than keeping a separate heading.

## Notes

Adds a fetch on the products endpoint to the "your plan" section of the usage page.

### DIFF
--- a/jsapp/js/account/stripe.types.ts
+++ b/jsapp/js/account/stripe.types.ts
@@ -205,6 +205,14 @@ export enum ChangePlanStatus {
   'error' = 'error',
 }
 
+export enum SubscriptionChangeType {
+  CANCELLATION,
+  RENEWAL,
+  PRODUCT_CHANGE,
+  PRICE_CHANGE,
+  NO_CHANGE
+}
+
 export type ChangePlan =
   | {
       status: ChangePlanStatus.success | ChangePlanStatus.pending;
--- a/jsapp/js/account/stripe.types.ts
+++ b/jsapp/js/account/stripe.types.ts
@@ -72,7 +72,7 @@ interface SubscriptionSchedule {
 }
 
 export interface SubscriptionInfo {
-  plan: PlanInfo;
+  plan: number;
   djstripe_created: string;
   djstripe_updated: string;
   id: string;
@@ -83,7 +83,7 @@ export interface SubscriptionInfo {
   application_fee_percent: any;
   billing_cycle_anchor: string;
   billing_thresholds: any;
-  cancel_at: any;
+  cancel_at: string | null;
   cancel_at_period_end: boolean;
   canceled_at: any;
   collection_method: string;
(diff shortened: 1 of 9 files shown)


jsapp/js/account/stripe.types.ts
```diff
@@ -72,7 +72,7 @@ interface SubscriptionSchedule {
 }
 
 export interface SubscriptionInfo {
-  plan: PlanInfo;
+  plan: number;
```
jamesrkiger, line 75: FYI, I noticed while working on this that plan here returns as the PK of the plan, rather than a nested Plan field. I figured the interface was probably just outdated so changed it accordingly.

LMNTL: Good call! [Plans are deprecated in favor of prices](https://stripe.com/docs/api/plans), so we avoid using them.

jsapp/js/account/usage/yourPlan.component.tsx
```diff
@@ -1,25 +1,47 @@
...
 import LimitNotifications from 'js/components/usageLimits/limitNotifications.component';
 
+type subscriptiontype =
+  | 'cancellation'
+  | 'renewal'
+  | 'productChange'
+  | 'priceChange'
+  | '';
+
+const badgeColorKeys: {[key: string]: BadgeColor} = {
```
LMNTL, line 21: This constant should probably be in ALL_UPPERCASE to fit our coding conventions.

jsapp/js/account/usage/yourPlan.component.tsx
```diff
@@ -1,25 +1,47 @@
...
+import {PlanNames, BaseProduct, Product} from 'js/account/stripe.types';
+import {getProducts} from '../stripe.api';
 import LimitNotifications from 'js/components/usageLimits/limitNotifications.component';
 
+type subscriptiontype =
+  | 'cancellation'
+  | 'renewal'
+  | 'productChange'
+  | 'priceChange'
+  | '';
```
LMNTL, line 19: There are a lot of string comparisons to these values; the code would be less vulnerable to accidental typos (e.g. `cancelation`) if you turned this into an `enum` and then used it to test/assign values.

LMNTL (changes requested):
Looking good! I have some comments, mostly because I'm reviewing a lot of code. 👍

jsapp/js/utils.ts
```diff
@@ -123,6 +123,16 @@ export function formatDate(
...
 
+/**
+ * Takes a Unix timestamp. Returns a UTC string
+ */
+export function unixToUtc(
+  time: number,
+): string {
+  let myMoment = moment.unix(time);
+  return myMoment.utc().toISOString();
+}
```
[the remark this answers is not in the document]
jamesrkiger: Good call. I went with 'convert' instead of 'format' to distinguish it from functions that return human readable strings.

LMNTL (approved):

PR #4814, opened 2024-01-19
Add dialog to announce new features in the UI

duvld:
## Description

Adds a new dialog component that can be placed on new features which may not stand out at first glance.

## Notes

Dialog is a wrapper for whatever component needs to be featured. CSS places the dialog relative to the left-most part of the child. On close, the dialog adds a variable to localstorage and will not show again unless changed.

## Related issues

Internal discussion: 
https://chat.kobotoolbox.org/#narrow/stream/7-Kobo-UX.2FUI.2FProduct/topic/Project-level.20control.20of.20anonymous.20submission

Part of: #4728


jsapp/js/stores.es6
```diff
@@ -142,6 +142,7 @@ stores.pageState = Reflux.createStore({
     if (this._onHideModal) {
       this._onHideModal();
     }
+    console.log('stores porper', this.state);
```
LMNTL, line 145: This should get removed 👍

jsapp/js/components/newFeatureDialog.module.scss
```diff
@@ -0,0 +1,49 @@
...
+  pointer-events: none;
+  border-width: 6px;
+  margin-left: calc(6px * -1);
+  bottom: 100%;
+  border-bottom-color: colors.$kobo-dark-blue;
+}
+
+.header {
+  color: colors.$kobo-white !important;
+  font-weight: 600 !important;
```
LMNTL, line 37: Do these two lines need the `!important`? The styling appears to be the same without it (although that could be down to differences in browser/platform.)

Generally, we want to avoid using `!important` - if you need it, your CSS selector could probably stand to be more specific (`div.header` etc.)

jsapp/js/components/newFeatureDialog.module.scss
```diff
@@ -0,0 +1,49 @@
...
+
+.dialog::before {
+  content: " ";
+  left: 5%;
+  border: solid transparent;
+  height: 0;
+  width: 0;
+  position: absolute;
+  pointer-events: none;
+  border-width: 6px;
```
LMNTL, line 29: The pixel dimensions in this file should probably use the values from `scss/sizes`.

LMNTL (changes requested):

jsapp/js/components/newFeatureDialog.component.tsx
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  return (
+    <div className={cx(styles.root, {className: className})}>
+      <div className={styles.wrapper}>{children}</div>
+      {showDialog && !disabled && (
+        <div className={styles.dialog}>
+          <div className={styles.header}>
+            {t('New feature')}
+            <Button
```
LMNTL, line 51: One thing I missed: when this button has focus, it gets a white background: 
image

Maybe a different button type would be better? Or some CSS to make that look right.

LMNTL (approved):

jnm (approved):
Functionality looks good :+1: Thanks!

jnm (changes requested):
Oops, actually, there's an issue. The notification dismissal needs to be stored per user. I'll see if I can make a quick change myself. (If Patrick logs in, dismisses the notification, logs out, and later Kate logs in, Kate should still see the message.)

jnm (approved):

PR #4815, opened 2024-01-19
Fix question counter to reflect accurate number of questions throughout the app 

### LINKED ISSUE #3951: Form Landing view questions counter is wrong
## Description

That small text with question counter is counting all rows - groups and questions. It shouldn't count groups and any other non-answerable rows (like a `note`).

In this example below I have two groups, first with 2 questions, and second with 1 question. It shows 5, but it should show 3:

Screenshot 2022-07-22 at 14 15 40

RuthShryock:
## Checklist

3. [X] Ensure the tests pass

## Notes

Updated the `row_count` variable to increment based on the `type` of question.
NOTE: this updates the value of `row_count` throughout the application so if we do not want this change, I can create a new variable in the asset summary to just change the question count in the form landing view page. 

## Related issues

Fixes #3951

### DIFF
--- a/kpi/utils/asset_content_analyzer.py
+++ b/kpi/utils/asset_content_analyzer.py
@@ -81,6 +81,7 @@ class AssetContentAnalyzer:
 
         names_by_quality = {'good': [], 'bad': [], 'ok': []};
         index = 0
+        exclude_types = ['note', 'start', 'end', 'begin_group', 'end_group']
         for row in self.survey:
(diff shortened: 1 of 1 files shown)


magicznyleszek (changes requested):
This works for deployed projects, ~~but drafts still display wrong number - both in Summary and Form tabs~~ and after rebuliding for drafts too

PR #4816, opened 2024-01-19
Mark org speed test as 'performance', exclude performance tests from running without flag

LMNTL:
## Checklist

3. [X] Ensure the tests pass

## Description

Prevents running a test that should only be run manually from running automatically.


kobo/apps/stripe/tests/test_organization_usage.py
```diff
@@ -138,7 +138,7 @@ def test_doesnt_include_org_users_with_invalid_plan(self):
             self.expected_file_size() * self.expected_submissions_single
         )
 
-    @pytest.mark.xfail
+    @pytest.mark.performance
```
jnm, line 141: Now that the test is excluded by default, would it be worth cranking up the number of model instances to a number that you would want to use when you run performance tests? (Or, if not, adding a comment describing how someone might use the test to test performance meaningfully?)

pytest.ini
```diff
@@ -2,3 +2,9 @@
 testpaths = kobo kpi hub
 env =
     DJANGO_SETTINGS_MODULE=kobo.settings.testing
+addopts =
+    -m "not performance"
+markers =
+    [
+        performance: marks test for performance that may unexpectedly fail on slower platforms (not executed by default - select with '-m "performance"')
+    ]
```
jnm, line 10: What about moving the entire contents of this file into `pyproject.toml` instead of adding these settings twice? [and then deleting `pytest.ini`]

jnm (changes requested):

jnm (approved):

PR #4822, opened 2024-01-25
Add custom Submission plans for Teams plan

LMNTL:
## Checklist

3. [x] Ensure the tests pass

## Description

Allows users to purchase multiple quantities of a plan in Stripe. Improve the UI for users who have a scheduled plan change.

## Notes

Sub-PRs that have already been reviewed:

* #4819 - Use subscription quantity in usage limit calculations
* #4813 - Add badges for plan downgrades and cancellations
* #4842 - Clean up Stripe product/price interfaces
* #4838 - Separate the Add-ons page into its own top-level page in Account Settings.
* #4857 - Send users to either the Add-ons or Plans page after Stripe checkout/portal, depending on their current subscription
* #4845 - Adds badges reflecting current add-on subscription and/or upcoming changes to add-on subscription (i.e. downgrades and cancellations).

## Media

Teams with custom submissions:
https://www.loom.com/share/a85fd5bc63844862a3f332bdbe0a9ef2

Add-ons and badges:
https://www.loom.com/share/2cab6691bdb54f109557e493f7e9315b?sid=6aca3890-9f1a-4867-8d4c-3dab284e1d84


jsapp/js/account/useProducts.hook.ts
```diff
@@ -0,0 +1,38 @@
+import {createContext, useEffect, useState} from 'react';
+import type {Product} from 'js/account/stripe.types';
+import {getProducts} from 'js/account/stripe.api';
+import useWhenStripeIsEnabled from 'js/hooks/useWhenStripeIsEnabled.hook';
+
+export interface ProductsState {
+  products: Product[];
+  isLoaded: boolean;
```
bufke, line 8: Do we use this enough to abstract "isLoading" into a base interface? Once we add refresh logic, we'll also likely want a isInitialLoad or whatever chosen phrase.

jsapp/js/utils.ts
```diff
@@ -123,6 +123,14 @@ export function formatDate(
   return myMoment.format(format);
 }
 
+/**
+ * Takes a Unix timestamp. Returns a UTC string
+ */
+export function convertUnixTimestampToUtc(time: number): string {
+  let myMoment = moment.unix(time);
```
bufke, line 130: Should we use moment on new code, it's [devs recommend against it](https://momentjs.com/docs/#/-project-status/)

jsapp/js/rootContextProvider.component.tsx
```diff
@@ -0,0 +1,11 @@
+import React, {ReactNode} from 'react';
+import {BillingContextProvider} from 'js/account/billingContextProvider.component';
+
+/* This is a context provider that wraps the root element (in ./app.js)
+ * Use this if you need to share context between, for example, the sidebar and the main page
```
bufke, line 5: "the sidebar and the main page" is a little vague and if we add a signed out state, this won't be true. Maybe just emphasize using this as little as possible and only when it's truly needed as global state. 

Imagine I make a new Foo Feature with a foo page and a foo nav. I may need a foo context. But I don't need to add it to RootContext.

jsapp/js/account/plans/planContainer.component.tsx
```diff
@@ -151,67 +203,140 @@ export const PlanContainer = ({
...
+                          parseInt(
+                            product.price.metadata?.nlp_seconds_limit || '0'
+                          ))) /
+                      60
+                    ).toLocaleString()
+                  )
+                  .replace(
+                    '##plan_interval##',
+                    product.price.recurring!.interval
+                  )}
```
bufke, line 298: All these functions deep in the jsx gets hard to read. Could it just be a function?

jsapp/js/account/plans/useDisplayPrice.hook.tsx
```diff
@@ -0,0 +1,26 @@
+import {useMemo} from 'react';
+import {Price} from 'js/account/stripe.types';
+import {getAdjustedQuantityForPrice} from 'js/account/stripe.utils';
+
+export const useDisplayPrice = (
+  price?: Price | null,
+  submissionQuantity = 1
+) => {
+  return useMemo(() => {
```
bufke, line 9: I guess the purpose of making this a hook and not just a simple "selector" function is to embed the useMemo? Just checking my understanding. Does it offer an advantage over a function without it? The memo depends on the passed in submissionQuantity, price. I'm not asking whether useMemo is merited generally or not, but is it actually executing the way I think, where it runs again if the variable passed in changes, in the scope of where the function is called from.

jsapp/js/account/stripe.api.ts
```diff
@@ -50,9 +54,13 @@ export async function getOrganization() {
...
  */
-export async function postCheckout(priceId: string, organizationId: string) {
+export async function postCheckout(
+  priceId: string,
+  organizationId: string,
+  quantity = 1
+) {
   return fetchPost<Checkout>(
-    `${endpoints.CHECKOUT_URL}?price_id=${priceId}&organization_id=${organizationId}`,
+    `${endpoints.CHECKOUT_URL}?price_id=${priceId}&organization_id=${organizationId}&quantity=${quantity}`,
```
bufke, line 63: Part of me wants this to be a key-value mapping to represent the query params. It doesn't bother me when it's only 1 or 2. Feel free to ignore.

```
interface checkoutParams {
     priceId: string,
     etc
     
`${endpoints.CHECKOUT_URL}?new URLSearchParams(params)`
```

jsapp/js/account/stripe.utils.ts
```diff
@@ -127,3 +115,103 @@ export const getSubscriptionsForProductId = (
...
+    adjustedQuantity = Math.ceil(adjustedQuantity);
+  }
+  if (transform?.round === 'down') {
+    adjustedQuantity = Math.floor(adjustedQuantity);
+  }
+  return adjustedQuantity;
+};
+
+/**
+ * Tests whether a new price/quantity would cost less than the user's current subscription.
```
bufke, line 196: :laughing: that definition of what a downgrade is - if it costs more it **must** be better.

bufke (commented):
I reviewed all of the TypeScript at a high level. We discussed how some day we may want to remove things from RootContext. We may want to add organization update logic sooner.

jsapp/js/account/plans/useDisplayPrice.hook.tsx
LMNTL: Correct - instead of the hook being run on every render, it only updates once when the price and/or submission quantity change. It's a small optimization, but if I add a `console.log` into the body of the `useMemo()`, then test without the `useMemo()`, it does log fewer times.

jsapp/js/account/usage/usage.api.ts
```diff
@@ -73,3 +72,40 @@ export async function getUsage(organization_id: string | null = null) {
     errorMessageDisplay: t('There was an error fetching usage data.'),
   });
 }
+
+export async function getAssetUsage(url?: string) {
```
bufke, line 76: Could be `(url = ASSET_USAGE_URL)`

jsapp/js/account/usage/usage.api.ts
```diff
@@ -73,3 +72,40 @@ export async function getUsage(organization_id: string | null = null) {
...
+    includeHeaders: true,
+    errorMessageDisplay: t('There was an error fetching asset usage data.'),
+  });
+}
+
+export async function getAssetUsageForOrganization(
+  pageNumber: number,
+  order?: ProjectsTableOrder
+) {
+  let organizations;
```
bufke, line 88: This is an implicit any

jsapp/js/account/usage/usage.api.ts
```diff
@@ -73,3 +72,40 @@ export async function getUsage(organization_id: string | null = null) {
...
+  try {
+    organizations = await getOrganization();
+  } catch (error) {
+    return await getAssetUsage(ASSET_USAGE_URL);
```
bufke, line 92: Maybe explain what's going on here. Either inline or for the whole function.

jsapp/js/account/usage/usage.api.ts
```diff
@@ -73,3 +72,40 @@ export async function getUsage(organization_id: string | null = null) {
...
+  }
+
+  let apiUrl = ORGANIZATION_ASSET_USAGE_URL.replace(
+    '##ORGANIZATION_ID##',
```
bufke, line 96: I'd probably define the urls as functions rather than use a custom string token to be replaced. But this was done outside of this PR, so out of scope. I'm not sure what is most popular for react devs to do.

jsapp/js/account/usage/usage.api.ts
```diff
@@ -73,3 +72,40 @@ export async function getUsage(organization_id: string | null = null) {
...
+  ).replace('##PAGE_NUM##', pageNumber.toString());
+
+  if (
+    order?.fieldName &&
+    order.direction &&
+    (order.direction === 'ascending' || order.direction === 'descending')
+  ) {
+    const orderingPrefix = order.direction === 'ascending' ? '' : '-';
+    const fieldDefinition = PROJECT_FIELDS[order.fieldName as ProjectFieldName];
+    apiUrl += `&ordering=${orderingPrefix}${fieldDefinition.apiOrderingName}`;
```
bufke, line 107: We're inconsistent on how to generate query param strings. I see some usage of URLSearchParams. I think a strategy that puts them in a map and then converts to a string at the end is cleaner.

jsapp/js/hooks/useApiFetcher.hook.ts
```diff
@@ -3,44 +3,70 @@ import {useCallback, useEffect, useState} from 'react';
 export interface ApiFetcherOptions {
   /* If set, the ApiFetcher will execute the fetch() function once per the
    * provided number of seconds. Will only fetch when the window is visible.
-   * *Don't* use this for fetches that mutate data on the server (POST, DELETE, PUT).
+   * *Don't* use this for fetches that mutate data on the server (DELETE, PUT, some POST requests).
```
bufke, line 6: Is there a concise way to explain "some". I don't mean to suggest every comment become a tomb, but if there is a short word or two to replace "some" then it would be better.

bufke (changes requested):

Issue #4393, opened 2023-03-30
[NLP] Cannot delete the first translation added to a project


jnm:
The back end does not always respect the special delete character `⌫`. The problem is consistently reproducible on the first translation of the first submission of a newly created project.

How to reproduce:
1. Create a new project;
2. Add an audio question to the form;
3. Deploy the project;
4. Use Enketo to make one submission, uploading any short audio file;
5. Go to the table view;
6. In the column for the audio question, click the "Open" button;
7. Click "begin";
8. Choose any language;
9. Click "manual";
10. Type anything;
11. Click "Save";
12. Click the "TRANSLATIONS" tab;
13. Click "begin";
14. Choose any language;
15. Click "manual";
16. Type anything;
17. Click "Save";
18. Click the trash can icon;
19. Click "DELETE";
20. See the expected payload sent to the back end, e.g.
    ```json
    {
      "submission": "e21bdfc5-2e24-435a-8179-aea95d1b7fb9",
      "i_m_listening": {
        "translation": {
          "fr": {
            "value": "⌫",
            "languageCode": "fr"
          }
        }
      }
    }
    ```
21. See that the response still includes the translation you attempted to delete:
    ```json
    {
      "i_m_listening": {
        "transcript": {
          "value": "transcript",
          "revisions": [],
          "dateCreated": "2023-03-30 13:47:12",
          "dateModified": "2023-03-30 13:47:12",
          "languageCode": "en"
        },
        "translation": {
          "fr": {
            "value": "traduction",
            "revisions": [],
            "dateCreated": "2023-03-30T13:47:21Z",
            "dateModified": "2023-03-30T13:47:21Z",
            "languageCode": "fr"
          }
        }
      }
    }
    ```

Internal discussion: https://chat.kobotoolbox.org/#narrow/stream/4-Kobo-Dev/topic/2.2E022.2E44.20Release/near/181085

jnm:
Fixed by #4825

PR #4825, opened 2024-01-30
Fix NLP translation editing

jnm:
Internal discussion: https://chat.kobotoolbox.org/#narrow/stream/4-Kobo-Dev/topic/can't.20edit.20translations


noliveleger (approved):

jnm:
Bonus: this fixes #4393 according to my testing

PR #4760, opened 2023-12-11
Improve appearance of Signup and User Profile fields

p2edwards:
Various WIP changes to `feature/tos` (#4699)

**Links**:
- [Internal thread for testing & feedback](https://chat.kobotoolbox.org/#narrow/stream/4-Kobo-Dev/topic/New.20required.20fields.20.28TOS.20update.29/near/308436)
- [Internal Notion card](https://www.notion.so/kobotoolbox/Require-extra-information-when-creating-accounts-epic-924597918852497da871e2d4de2181b3?p=7b3e54c54a8a48eb807241cf58e3e2f9&pm=s) (list of planned or considered changes)

PR #4827, opened 2024-01-30
Keyboard-accessible Select component for TOS/Accounts

p2edwards:
## Related issues

Part of #4760

## Notes

Keyboard-accessible Select component for TOS/Accounts

- Supports browser conventional keyboard interactions
- Renders option menu to support designs not supported by `<select>`
- Uses 'current' coding conventions (hooks, scss modules)


jsapp/js/components/special/koboSelect3.tsx
```diff
@@ -0,0 +1,391 @@
...
+      <button
+        className={cx(
+          styles.trigger,
+          props.value || styles.placeholding,
+          props.error && styles.hasError,
+        )}
+        type='button'
+        // ARIA. There's room for improvement here.
+        // https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/combobox_role
+        aria-role='combobox'
```
LMNTL, line 291: It looks this this should be `role` instead of `aria-role` - it fixes this error in the console:

image

(Which is delightfully inconsistent on React's part.)

LMNTL (changes requested):
It's pretty close, most of my comments are on code style rather than functionality. You might want to review the comments, too, in case some of them are WIP.

LMNTL (approved):
Thank you kindly!

PR #4833, opened 2024-02-05
Fix bulk update validation status for filtered data and improve some Data Table filters

### LINKED ISSUE #4745: Batch update of all submissions doesn't work when filter is in place

magicznyleszek:
## Checklist

3. [x] Ensure the tests pass

## Description

Bulk updating validation status is now correctly targeting submissions that were selected in Data Table with filtering included. Enabled filtering for more columns in Data Table: `range`, `rank`, `score`, `select_multiple_from_file` and `select_one_from_file`. Changed the bahaviour of filtering for `decimal`, `integer`, `range`, `rank`, `score`, `select_one`, and `select_one_from_file` questions - now the filter value will match exactly the response.

## Notes

Changes:
- Introduced `buildFilterQuery` utility function that is producing query for both Data Table and TableBulkOptions, ensuring consistency and making future adjustments easier
- Changed how filtering works for some columns - now it matches the selected option exactly (previously "yes" was matching both choices: "yes" and "yessica", and respectively "1" was matching all "1", "21", and "8979111888"). Columns affected are:
  - `decimal`
  - `integer`
  - `range`
  - `rank`
  - `score`
  - `select_one`
  - `select_one_from_file`
- Enabled filtering for more columns:
  - `range`
  - `rank`
  - `score`
  - `select_multiple_from_file`
  - `select_one_from_file`
- Introduced some constants and cleaned up few lines of code

## Related issues

Fixes #4745 
Surpasses #4746


jsapp/js/components/submissions/tableConstants.ts
```diff
@@ -83,10 +105,31 @@ export const TEXT_FILTER_QUESTION_TYPES = [
...
 ];
+
+/**
+ * These are question types that will be filtered by the exact filter value
+ * (i.e. filter value is exactly the response). Any question type not on this
+ * list will be filtered by responses that include the value (i.e. filter value
+ * is part of the response).
+ */
+export const FILTER_EXACT_TYPES: AnyRowTypeName[] = [
+  QuestionTypeName.decimal,
```
duvld, line 128: I don't think I understand this exact value change. I'd expect this to mean that, using decimal as an example, if my submissions were `2.2` and `2.3` then the filter would not show either of these as neither are exactly equal to `2`. But I don't see it behaving like that in the data table:

image

jsapp/js/components/submissions/tableConstants.ts
```diff
@@ -59,21 +61,41 @@ CELLS_WIDTH_OVERRIDES[VALIDATION_STATUS_ID_PROP] = 125;
...
-  QUESTION_TYPES.datetime.id,
-  QUESTION_TYPES.barcode.id,
-  QUESTION_TYPES.calculate.id,
+  QuestionTypeName.barcode,
+  QuestionTypeName.calculate,
+  QuestionTypeName.date,
+  QuestionTypeName.datetime,
+  QuestionTypeName.decimal,
+  QuestionTypeName.integer,
+  QuestionTypeName.range,
```
duvld, line 90: It looks like `range`, `rank`, and `score` got reversed somehow. Here they are on beta:
image
`range` cannot be filtered but `rank` and `score` can

And on the branch:
image
`range` can be filtered but `rank` and `score` cannot

jsapp/js/components/submissions/tableConstants.ts
duvld: Actually it looks like all the mentioned exact types still follow the previous scheme of "1" matching "12", "1.1" etc.

jsapp/js/components/submissions/tableUtils.ts
```diff
@@ -272,3 +278,73 @@ export function getAllDataColumns(
...
+      }
+      // Apart from few exceptions in above cases, we tend to treat all columns
+      // (`filter.id`s) with the same algorithm
+      default: {
+        // We assume `filter.id` is the question name (Data Table column name)
+        const foundRow = survey.find((row) => getRowName(row) === filter.id);
+
+        // Some question types needs the data to be filtered by exact values
+        // (e.g. "yes" shouldn't mach "yessica" or "yes and no")
+        if (foundRow && FILTER_EXACT_TYPES.includes(foundRow.type)) {
```
duvld, line 332: The PR does fix the bug reported in #4745 but I'm kind of curious about this change here. Should we really be changing the default behaviour to match exact values (without any option)? I'd imagine this would be fairly different from what a user would be used to.

jsapp/js/components/submissions/tableConstants.ts
magicznyleszek: I've used this [ALL TYPES FORM.xlsx](https://github.com/kobotoolbox/kpi/files/14904929/ALL.TYPES.FORM.xlsx) for testing.

I've tested it now with every question type and it works as expected. All `FILTER_EXACT_TYPES` are matching exact strings, and all others are partial matching.

jsapp/js/components/submissions/tableUtils.ts
magicznyleszek: Tino gave this change green light, see https://chat.kobotoolbox.org/#narrow/stream/4-Kobo-Dev/topic/Bulk.20update.20validation.20status.20bug/near/341412

There is also a spreadsheet that lists all question types and what behaviour we want filtering wise: https://docs.google.com/spreadsheets/u/1/d/1tVOeg7Sdu64qh22ku0KsGjs9t-Fl3g-DBVQa0htiRPQ/edit#gid=0

jsapp/js/components/submissions/tableConstants.ts
magicznyleszek: Could you check if your code goes into proper place (near `FILTER_EXACT_TYPES.includes…`) in `buildFilterQuery` function?

duvld: Looks good now!

PR #4834, opened 2024-02-06
Optimization: Avoid fetching all anonymous user permissions every time

noliveleger:
## Description

Speed up API response for asset related endpoints

## Notes

Instead of fetching the all permissions each time, only fetch permissions for the subset of asset ids that the current page of the response is rendering.

### DIFF
--- a/kpi/mixins/object_permission.py
+++ b/kpi/mixins/object_permission.py
@@ -780,7 +780,7 @@ class ObjectPermissionMixin:
 
     @staticmethod
     @cache_for_request
-    def __get_all_user_permissions(user_id: int, asset_ids: list = []) -> dict:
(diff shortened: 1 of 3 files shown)


kpi/mixins/object_permission.py
```diff
@@ -783,11 +780,13 @@ def __get_all_object_permissions(object_id):
 
     @staticmethod
     @cache_for_request
-    def __get_all_user_permissions(user_id):
+    def __get_all_user_permissions(user_id: int, asset_ids: list = []) -> dict:
```
jnm, line 783: Let's use an immutable default just in case some future developer modifies `asset_ids` inside the function

jnm (changes requested):

jnm (approved):

PR #4838, opened 2024-02-08
Separate add ons page

jamesrkiger:
## Checklist

3. [x] Ensure the tests pass

## Description

Adds a separate add-ons page under accounts. Introduces "ProductsContext" to fetch a list of products without duplicating API calls.

## Notes

Currently I am using the Plan component for both the add-ons and plans pages. This was the quickest solution given the shared code for both of these views, but now that I'm done with everything I think they should probably be split up. Open to feedback on this during review.


LMNTL (changes requested):

jsapp/js/account/stripe.api.ts
```diff
@@ -217,8 +217,7 @@ const getStripeMetadataAndFreeTierStatus = async () => {
     // the user has no subscription, so get limits from the free monthly product
     hasFreeTier = true;
     try {
-      const products = await getProducts();
-      const freeProduct = products.results.filter((product) =>
+      const freeProduct = products.filter((product) =>
```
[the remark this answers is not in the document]
jamesrkiger: I have removed it as you suggested, though I had to add an await statement to avoid a race condition with products loading.

jsapp/js/account/plans/plansRoute.component.tsx
```diff
@@ -0,0 +1,17 @@
...
+export default function plansRoute() {
+  const usage = useUsage();
+  const products = useProducts();
+
+  return (
+    <UsageContext.Provider value={usage}>
+      <ProductsContext.Provider value={products}>
+        <Plan />
+      </ProductsContext.Provider>
+    </UsageContext.Provider>
```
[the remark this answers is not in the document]
jamesrkiger: I have gone ahead and moved the context providers to a new `BillingOutlet`. It was a little awkward because the routes aren't nested under their own url path, so they are all index routes now. Let me know what you think.

jsapp/js/account/useProducts.hook.ts
```diff
@@ -0,0 +1,37 @@
...
+      setProducts(() => {
+        return {
+          products: products.results,
+          isLoaded: true,
+        };
+      });
+    });
+  }, []);
+
+  return products;
```
LMNTL, line 32: It looks like there's a bug when Stripe isn't enabled. You can replicate by commenting out the `STRIPE_TEST_SECRET_KEY` and `STRIPE_LIVE_MODE` environment variables. This fetch happens even when Stripe is disabled, and 404s. Instead of wrapping this fetch in `useEffect`, you may want to take a look at `useWhenStripeIsEnabled` (which is basically just `useEffect` that only runs once we've confirmed Stripe is active.)

LMNTL (changes requested):

jsapp/js/account/useProducts.hook.ts
jamesrkiger: @LMNTL Thanks for catching this. I have added the useWhenStripeIsEnabled wrapper as you suggested. 

I also updated the isFullyLoaded memo on the usage page to check for either products being loaded or stripe not being enabled so we don't get an infinite loading spinner there when Stripe is disabled.

jsapp/js/account/usage/usage.component.tsx
```diff
@@ -45,7 +45,10 @@ export default function Usage() {
   const location = useLocation();
 
   const isFullyLoaded = useMemo(
-    () => usage.isLoaded && productsContext.isLoaded && limits.isLoaded,
+    () =>
```
LMNTL, line 48: This dependency array needs to include `limits.stripeEnabled`, since it's referenced in the memo function. (I keep footgunning myself with this.) Once that's done, feel free to merge!

LMNTL (approved):

PR #4841, opened 2024-02-11
Fix "Could not find file.ext" error with repeated groups

noliveleger:
## Description

Fix attachments which do not load when adding media questions to repeated groups.

## Notes

I've tested it with repeated groups, nested groups and repeated nested groups. ~~I used only one level for nested group but it should world for edge cases too.~~ Tried it with 2 levels of nesting. 👍 

:warning: Because **XPaths are detected from the latest deployed version only**, if the nesting of groups changes from one version to another, the images may no longer be displayed.


noliveleger:
Closed in favor of #4843

PR #4842, opened 2024-02-12
Stripe product and price interface cleanup

jamesrkiger:
## Checklist

3. [x] Ensure the tests pass

## Description

This PR aims to improve the accuracy and readability of our frontend interfaces related to Stripe products and prices. Previously we had two extensions of the BaseProduct interace: Product and Price, with the only difference between them being that a Product's `prices` field contained an array of price objects, while a Price's `prices` field contained only a single price (obtained by filtering on the frontend). I have renamed the latter `FilteredPriceProduct`. Not the most elegant name, but I think it gets the point across (and I'm open to other suggestions). The bulk of the changes in this PR are related to this. But I also removed some duplicate interfaces in the types file and adjusted some interfaces to match our API.

### DIFF
--- a/jsapp/js/account/plans/addOnList.component.tsx
+++ b/jsapp/js/account/plans/addOnList.component.tsx
@@ -2,7 +2,7 @@ import React, {useCallback, useEffect, useState} from 'react';
 import useWhen from 'js/hooks/useWhen.hook';
 import subscriptionStore from 'js/account/subscriptionStore';
 import type {
-  BasePrice,
+  Price,
   Organization,
   Product,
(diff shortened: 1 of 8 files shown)


LMNTL (changes requested):
This is a well-needed fix, thank you 🙏🏽 My only comment is the one you solicited: `FilteredPriceProduct` is a little ambiguous as to whether it's a `Price` or `Product` object. (It's *definitely* less ambiguous than the existing code 😅) Would you be cool with something like `SinglePricedProduct`? I think that has room for improvement as well, so feel free to modify it.

PR #4844, opened 2024-02-12
Fix error when trying to edit collection, block or question permissions

noliveleger:
## Description

Remove `add_submissions` permission from anonymous user on non survey type assets

## Notes

Added by mistake with 0055 migration in this release


kpi/migrations/0056_fix_asset_permissions.py
```diff
@@ -0,0 +1,42 @@
...
+def remove_add_submissions_for_anonymous_user(apps, schema_editor):
+
+    Asset = apps.get_model('kpi', 'Asset')  # noqa
+    ObjectPermission = apps.get_model('kpi', 'ObjectPermission')  # noqa
+    Permission = apps.get_model('auth', 'Permission')  # noqa
+    permission_id = Permission.objects.get(codename=PERM_ADD_SUBMISSIONS).pk
+
+    ObjectPermission.objects.filter(
+        deny=False,
+        inherited=False,
```
jnm, line 21: Should we have this restriction in the filter? It shouldn't ever be valid to have `add_submissions` on non-`survey` assets

jnm (approved):

PR #4850, opened 2024-02-16
Fix broken admin reports

noliveleger:
## Description

Fix error introduced in `2.024.04` where reports crash before being fully generated.

## Notes 

`get_storage_class()` [is deprecated since Django 4.2](https://docs.djangoproject.com/en/4.2/ref/files/storage/#django.core.files.storage.get_storage_class)


bufke (approved):
I suggest that in the future we add both a unit test and linting tool that would have found this and failed CI

PR #4852, opened 2024-02-20
Remove mislabeled 'display transcript' setting in NLP when there's no transcript

LMNTL:
## Checklist

3. [X] Ensure the tests pass

## Description

Removed a duplicate 'Original File (audio)' setting that was present when viewing a submission in NLP without a transcript.

## Notes

Original UI:
image

Updated UI:
image

[internal xref](https://chat.kobotoolbox.org/#narrow/stream/4-Kobo-Dev/topic/Qualitative.20Analysis.20WIP/near/343660)


jsapp/js/components/processing/sidebar/sidebarDisplaySettings.tsx
```diff
@@ -32,14 +32,17 @@ export default function SidebarDisplaySettings() {
   const availableDisplays = store.getAvailableDisplays(activeTab);
```
magicznyleszek, line 32: I like that we are making sure that the `getStaticDisplayText` returns data that makes sense. But I think that `getAvailableDisplays` should too. So if there is no transcript, it shouldn't propose it (i.e. shouldn't `outcome.push(StaticDisplays.Transcript);`)

magicznyleszek: If we fix `getAvailableDisplays`, then I think we don't need all that:

```
                const staticDisplayText = getStaticDisplayText(staticDisplay);
                if (staticDisplayText === null) {
                  return null;
                }
```

magicznyleszek (changes requested):

magicznyleszek (approved):

PR #4855, opened 2024-02-23
Polyfill useId hook and improve button accessibility

LMNTL:
## Checklist

3. [X] Ensure the tests pass

## Description

Improves labels on buttons to make them more accessible.

## Notes

Adds a basic polyfill for the [useId](https://react.dev/reference/react/useId) hook from later versions of React. (Once we do upgrade to a React version that supports `useId`, this function can get removed - the API is the same.)

By default, a button takes its contents as its accessible label. The `startIcon` of the the Kobo common button, for example, renders as text that would prepend `props.label`, confusing screen readers. `aria-labelledby` with an actual `<label>` is preferable to using `aria-label`, but requires a static ID. `useId` provides that ID, giving an example of this pattern.


p2edwards (approved):
Looks good to me, thanks for making this!

PR #4743, opened 2023-11-29
[Feature] Transfer project ownership

noliveleger:
## Description

Allow users to transfer their project to another user

## Notes

The back end support several projects at once per invite but the UI only support one project per invite.
Once a project is being transferred, it sends emails to previous owner and new owner. 
it also adds in-app message to all other users who have access to that the project.

In this PR, `deployment__identifier` (i.e. link to KoboCAT form representation) to is removed from asset back-end response when project is deployed. It's not 100% related to transfer but the link was based on owner's username. So it should have been updated too. Since the endpoint has been removed in a previous version, the link returns a 404.

PR #4863, opened 2024-02-28
Use API `v2` urls in unit tests `languages` & `audit_log` apps 

noliveleger:
## Description

`v1` is deprecated and new code should use `v2` endpoints.

## Notes

Apps urls are now included in API `v2` Router instead of being added in `url_patterns`. 
It lets Django pass the namespace in the viewname (e.g. `reverse(<namespace>:<viewname>)`)

## Related issues

Part of #4743

### DIFF
--- a/kpi/urls/router_api_v2.py
+++ b/kpi/urls/router_api_v2.py
@@ -2,9 +2,11 @@
 from django.urls import path
 from rest_framework_extensions.routers import ExtendedDefaultRouter
 
+from kobo.apps.audit_log.urls import router as audit_log_router
 from kobo.apps.hook.views.v2.hook import HookViewSet
 from kobo.apps.hook.views.v2.hook_log import HookLogViewSet
 from kobo.apps.hook.views.v2.hook_signal import HookSignalViewSet
+from kobo.apps.languages.urls import router as language_router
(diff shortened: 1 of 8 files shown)


kpi/urls/router_api_v2.py
```diff
@@ -150,6 +152,13 @@ def get_urls(self, *args, **kwargs):
                        ServiceUsageViewSet, basename='service-usage')
 router_api_v2.register(r'users', UserViewSet)
 
+
+# Merge django apps routers with API v2 router
+# All routes are under `/api/v2/` within the same namespace.
+router_api_v2.registry.extend(language_router.registry)
```
bufke, line 158: If we are adding this, can we then remove `path('', include(router.urls)),` in audit_log.urls and languages.urls? It seems like the URL is being added twice.

And also remove 

```
    re_path(r'^api/v2/', include('kobo.apps.languages.urls')),
    re_path(r'^api/v2/', include('kobo.apps.audit_log.urls')),
```

bufke (changes requested):
Otherwise lgtm. If all you need to do is take my suggestion and delete a couple lines, then please go ahead and merge once CI passes.

PR #4873, opened 2024-03-12
[TASK-542] Display automated transcription estimate

magicznyleszek:
## Checklist

3. [x] Ensure the tests pass

## Description

After user starts automated transcription, we now display an estimate and spinner.

## Notes

Changes here:
- Removed `bem.Loading` and cleaned up/refactored `LoadingSpinner` component
- Added `LoadingSpinner` to storybook
- Added `CenteredMessage` component (mainly as replacement for weird functionality in `LoadingSpinner` that I removed)
- Introduce `getAttachmentForProcessing` utility function to use it in multiple places
  - I've created a new file for it, because adding it to any other file caused circular dependency errors
- In the `StepConfigAuto` for transcript we now display a spinner and time estimate
  - I've added two utility functions to make it easier: `secondsToTranscriptionEstimate` and `getAudioDuration`


jsapp/js/components/processing/transcript/transcript.utils.ts
```diff
@@ -0,0 +1,59 @@
...
+  const estimateSeconds = Math.round(sourceSeconds * 0.5 + 10);
+  if (estimateSeconds < 45) {
+    return t('less than a minute');
+  } else if (estimateSeconds >= 45 && estimateSeconds < 75) {
+    return t('about 1 minute');
+  } else if (estimateSeconds >= 75 && estimateSeconds < 150) {
+    return t('about 2 minutes');
+  } else {
+    return t('about 3 minutes');
+  }
```
LMNTL, line 58: I'm a little confused about the math here - the jumps between the different minutes seem inconsistent, and `estimateSeconds` doesn't seem to represent actual seconds. Is this based solely on the formula, or did you do some real-life testing to get these values?

This should probably also handle values greater than a hard-coded limit of 3 minutes - saying 'more than 3 minutes' would be okay if the `estimateSeconds` value grows inconsistently as the actual time taken goes up.

jsapp/js/components/common/bigSpinner.component.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import styles from './bigSpinner.module.scss';
+
+/** Displays an animated blue spinner - for white backgrounds only. */
+export default function BigSpinner() {
+  return (
+    <span className={styles.bigSpinner}/>
+  );
+}
```
LMNTL, line 9: Does this need to be a separate component from `<LoadingSpinner/>`? The only real difference I see is the style/size of the spinner element/icon, and it's missing a label (visible or just ARIA-accessible), which `<LoadingSpinner/>` has. It might be worth considering adding this as a size option for `<LoadingSpinner/>` - or you could just give it some sort of accessible label.

LMNTL (changes requested):

jsapp/js/components/common/bigSpinner.component.tsx
magicznyleszek: Didn't think about incorporating it into LoadingSpinner. Good idea. I ended up improving it in greater deal than just adding big version to it :)

jsapp/js/components/common/bigSpinner.component.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import styles from './bigSpinner.module.scss';
+
+/** Displays an animated blue spinner - for white backgrounds only. */
```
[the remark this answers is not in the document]
magicznyleszek: Good idea :)

LMNTL (approved):

PR #4874, opened 2024-03-13
Django 4.2

noliveleger:
## Notes

All pip dependencies have been upgraded to the latest version (supported by Python 3.8)

Changes: 
- Update `django-constance` to 3.1
  - Removed `'constance.backends.database'` from `INSTALLED_APPS` 
- CI matrix includes Python 3.8, 3.10 and 3.12 ([More details](https://django-constance.readthedocs.io/en/latest/changes.html#v3-0-0-2023-07-27))
- CI PostgreSQL is bumped from 11 to 14 ([Django dropped support for PostgreSQL 11](https://docs.djangoproject.com/en/4.2/releases/4.2/#dropped-support-for-postgresql-11))
- Use Psycopg 3.x
- Use PyMongo 4.x
- Use new [`STORAGES`](https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-STORAGES) instead of `DEFAULT_FILE_STORAGE`
- Replace `unique_together` with Indexes ([More details](https://docs.djangoproject.com/en/4.2/releases/4.2/#index-together-option-is-deprecated-in-favor-of-indexes))

## Related PR
kobotoolbox/kobocat#922


.github/workflows/pytest.yml
```diff
@@ -21,10 +21,10 @@ jobs:
       ENKETO_REDIS_MAIN_URL: redis://localhost:6379/0
     strategy:
       matrix:
-        python-version: ['3.8', '3.10']
+        python-version: ['3.8', '3.10', '3.12']
     services:
       postgres:
-        image: postgis/postgis:11-3.2
+        image: postgis/postgis:14-3.2
```
bufke, line 27: Should we bump this to 14-3.4 which seems to be later?

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.10 as build-python
+FROM python:3.8 as build-python
```
bufke, line 1: Why downgrade this? I recommend against it. It could be considered a major breaking change for anyone customizing our app. If the concern is testing against minimal supported versions, let CI handle that.

kobo/settings/base.py
```diff
@@ -1091,14 +1093,15 @@ def dj_stripe_request_callback_method():
     if region := env.str('AWS_S3_REGION_NAME', False):
         AWS_S3_REGION_NAME = region
 
-
 ''' Storage configuration '''
+default_file_storage = 'django.core.files.storage.FileSystemStorage'
```
bufke, line 1097: You could refer to the django default via `django.conf.global_settings` instead of redefining the setting.

kobo/settings/base.py
```diff
@@ -1130,6 +1133,16 @@ def dj_stripe_request_callback_method():
     KOBOCAT_MEDIA_PATH = os.environ.get('KOBOCAT_MEDIA_PATH', '/srv/src/kobocat/media')
 
 
+STORAGES = {
+    'default': {
+        'BACKEND': default_file_storage,
+    },
+    'staticfiles': {
+        'BACKEND': 'django.contrib.staticfiles.storage.StaticFilesStorage',
```
bufke, line 1141: What happens if we omit staticfiles entirely? If that doesn't work, refer to from django.conf import global_settings

bufke (commented):
LGTM so far, I'll review again once it's marked as ready.

PR #4876, opened 2024-03-15
Fix 'Background Audio Files do not play in the submission preview'

### LINKED ISSUE #4859: Background Audio Files do not play in the submission preview
## Description
Background audio files do not play in the submission preview. This is true for both AMR and M4A files. 
It looks like there is no audio attached to the audio player. On inspection of the element, there is no `src` for the audio player to play. 

image
xref: https://chat.kobotoolbox.org/#narrow/stream/6-Kobo-Support/topic/Background.20Audio.20not.20playing.20with.20Editing.20Mode!

p2edwards:
## Related issues

Fix #4859

### DIFF
--- a/jsapp/js/components/submissions/table.es6
+++ b/jsapp/js/components/submissions/table.es6
@@ -1150,7 +1150,8 @@ export class DataTable extends React.Component {
       ) {
         const backgroundAudioUrl = getMediaAttachment(
           row.original,
-          row.original[backgroundAudioName]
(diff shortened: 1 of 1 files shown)


p2edwards:
Tested this on a staging server; 
- form created with KoboToolbox's Form Builder
- background audio collected with KoboCollect

<table>
  <tr>
    <td>Before
    <td>After
  <tr>
    <td>
    <td>
</table>

PR #4879, opened 2024-03-20
One-time addons usage details

jamesrkiger:
## Checklist

3. [ ] Ensure the tests pass

## Description

This PR adds an add-on details modal to the containers on the usage page, featuring a breakdown of usage across subscriptions and one-time add-ons.

## Notes

1. There are two figures in the usage breakdown that are currently filled in with `0` pending development of a backend system for tracking one-time addon usage across time. I have marked them with TODOs. 

2. While we do not currently support one-time addons for storage, in the usage containers and child components I have opted for generalization to avoid having to build in more checks for displaying/not-displaying these modals. On the one hand, this could be slightly confusing to someone unfamiliar with our payment system looking at the code. On the other hand, I think it makes both present and future development simpler.

3. I have tried to avoid making type/code changes that impact other aspects of our frontend payment system. So, for example, I opted for some bloat in the usage component limit state rather than making deeper changes to how we handle limit data, which now needs to be broken down into recurring limits (coming from recurring subscriptions) and remaining limits (from one-time add-ons in addition to recurring limits). Since we are likely going to be moving some of these calculations to the backend, I figured it would be best to save bigger changes for later when we have a more solid idea of what the data coming from the backend looks like. 

4. I made a small addition to the Kobo modal code to allow for white headers. Unlike the other color header classes, the new white header style class removes bottom padding (since the white background color itself creates empty space between the header and the modal contents).  It's a bit simplistic, but got the job done for the present purpose. As always, I'm open to suggestions for improvement here.


jsapp/js/account/usage/usage.component.tsx
```diff
@@ -121,6 +138,56 @@ export default function Usage() {
...
+          : [];
+      default:
+        return [];
+    }
+  }
+
+  // Find out if any usage type has one-time addons so we can
+  // adjust the formatting of the usage containers to accommodate
+  // a detail link.
+  const useAddonsLayout = useMemo(() => {
```
LMNTL, line 175: Because we're using a lot of react hooks, `useAddonsLayout` should get renamed to avoid confusion. [Hooks always start with `use`](https://react.dev/learn/reusing-logic-with-custom-hooks#hook-names-always-start-with-use), and this is a boolean (although it's generated by a hook.)

LMNTL (changes requested):
Nice work! I have some small change requests, but this is looking good.

LMNTL (approved):

PR #4881, opened 2024-03-21
fix button styles in certain modals

LMNTL:
## Checklist

3. [X] Ensure the tests pass

## Description

Fixes incorrect button styles in form modals.

## Notes

Any label inside of a form modal was getting extra styling applied due to changing the `<Button/>` label from a `<span>` to a `<label>`. I chose to fix it in the most low-impact way I could see available, to avoid any other styling mishaps.

## Related issues

Related to #4855

### DIFF
--- a/jsapp/scss/components/_kobo.bem.ui.scss
+++ b/jsapp/scss/components/_kobo.bem.ui.scss
@@ -461,13 +461,6 @@ iframe {
     &:not(.inline) {
       display: block;
     }
-
-    &.next-to-checkbox {
-      display: inline-block !important;
-      margin-left: 10px;
-      max-width: 90%;
-      vertical-align: top;
-    }
   }
 
   input[type='text'],
--- a/jsapp/scss/components/_kobo.bem.ui.scss
+++ b/jsapp/scss/components/_kobo.bem.ui.scss
@@ -87,7 +87,7 @@
 
 .text-small {
   font-size: 0.85em;
-  opacity: .75;
+  opacity: 0.75;
 }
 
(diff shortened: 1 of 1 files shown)


jsapp/scss/components/_kobo.bem.ui.scss
```diff
@@ -446,7 +447,7 @@ iframe {
     }
   }
 
-  label {
+  label:not(.k-button__label) {
```
magicznyleszek, line 450: Could you add a small comment that we are (kinda) hacking it here for the reason of unwanted overrides? It would be easier in future to find this place and add another exception :) After that feel free to merge it

LMNTL: Good point - it's better if this needs to be touched as little as possible in the future. I refactored a bit to make this less of a hack - let me know if you think it still merits a comment. 👍

jsapp/scss/components/_kobo.bem.ui.scss
```diff
@@ -456,17 +457,17 @@ iframe {
...
+    &:not(.inline) {
+      display: block;
+    }
 
-  label.next-to-checkbox {
-    display: inline-block !important;
-    margin-left: 10px;
-    max-width: 90%;
-    vertical-align: top;
+    &.next-to-checkbox {
```
magicznyleszek, line 465: We can delete this, it's not being used

magicznyleszek (changes requested):

magicznyleszek (approved):

PR #4882, opened 2024-03-25
Add new endpoint to get project usage for users based on their organization

RuthShryock:
## Checklist

3. [X] Ensure the tests pass

Created a paginated endpoint and updated the project breakdown component to call the endpoint and implement the pagination in full.

### DIFF
--- a/jsapp/js/account/usage/usage.api.ts
+++ b/jsapp/js/account/usage/usage.api.ts
@@ -83,16 +83,11 @@ export async function getUsageForOrganization() {
 }
 
 export async function getAssetUsage(url?: string) {
-  try {
-    const apiUrl = url || ASSET_USAGE_URL;
-    return fetchGet<AssetUsage>(apiUrl, {
-      includeHeaders: true,
-      errorMessageDisplay: t('There was an error fetching asset usage data.'),
-    });
-  } catch (error) {
-    console.error('Error fetching data:', error);
-    throw error;
-  }
+  const apiUrl = url || ASSET_USAGE_URL;
+  return fetchGet<AssetUsage>(apiUrl, {
+    includeHeaders: true,
+    errorMessageDisplay: t('There was an error fetching asset usage data.'),
+  });
 }
 
 export async function getAssetUsageForOrganization(url?: string) {
--- a/jsapp/js/account/usage/usage.api.ts
+++ b/jsapp/js/account/usage/usage.api.ts
@@ -52,7 +52,7 @@ const ORGANIZATION_USAGE_URL =
 
 const ASSET_USAGE_URL = '/api/v2/asset_usage/';
 const ORGANIZATION_ASSET_USAGE_URL =
-  '/api/v2/organizations/##ORGANIZATION_ID##/asset_usage/';
+  '/api/v2/organizations/##ORGANIZATION_ID##/asset_usage/?page=##PAGE_NUM##';
 
 export async function getUsage(organization_id: string | null = null) {
   if (organization_id) {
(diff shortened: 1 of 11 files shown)


jsapp/js/account/usage/usage.api.ts
```diff
@@ -63,3 +81,32 @@ export async function getUsageForOrganization() {
...
+export async function getAssetUsage(url?: string) {
+  try {
+    const apiUrl = url || ASSET_USAGE_URL;
+    return fetchGet<AssetUsage>(apiUrl, {
+      includeHeaders: true,
+      errorMessageDisplay: t('There was an error fetching asset usage data.'),
+    });
+  } catch (error) {
+    console.error('Error fetching data:', error);
+    throw error;
```
LMNTL, line 94: Hmm, `fetchGet()` should take care of displaying an error message via a toast, then send the error to our reporting service, which would make this redundant. It _does_ `console.error` (see `jsapp/js/utils.js`), but as you've pointed out through code, it'll only show the `errorMessageDisplay` value, not the error itself. (I think that's my 'oops' originally - I added the `errorMessageDisplay` option and the handling for it.)

So what if, instead of this `catch` block (and the similar one in `jsapp/js/account/usage/usageProjectBreakdown.tsx`), you pass the (optional) real error message to `notify()` in the `ToastOptions`?  Then use it in `handleApiFail()` on line 67.

jsapp/js/account/usage/usageProjectBreakdown.tsx
```diff
@@ -1,158 +1,100 @@
...
-
-  if (projectData.results.length === 0 || projectData.results.some((result) => !result.assets)) {
+  if (projectData.results.length === 0) {
     return <LoadingSpinner/>;
   }
 
+const calculateRange = (): string => {
+  const totalProjects = parseInt(projectData.count);
+  const startRange = (currentPage - 1) * 8 + 1;
+  const endRange = Math.min(currentPage * 8, totalProjects);
```
LMNTL, line 63: The page size of `8` should get turned into a constant, here and anywhere else it shows up in the front-end 👍 It'll make our lives easier if, say, someone asks us to show 50 results per page somewhere down the line 😅

jsapp/js/account/usage/usageProjectBreakdown.tsx
```diff
@@ -1,158 +1,100 @@
...
+  return `${startRange}-${endRange} of ${totalProjects}`;
+};
+
+const removePrefix = (url: string): string => url.replace('http://kf.kobo.local', '');
```
LMNTL, line 67: You should be able to get the the value to replace here using `ROOT_URL`, I think.

LMNTL (changes requested):
Thank you for this! I do have changes to ask for, but I want to emphasize that I think this PR is a solid chunk of work, and feels pretty clean. 👍

jsapp/js/account/usage/usageProjectBreakdown.tsx
```diff
@@ -69,90 +34,63 @@ const ProjectBreakdown = () => {
         };
       });
 
-      setProjectData((prevData) => {return {...prevData, results: updatedResults};});
+      if (isComponentMounted) {
```
LMNTL, line 37: What does `isComponentMounted` do here? If I comment all references to it out, everything seems to function the same, but I could easily be missing something.

jsapp/js/account/usage/usageProjectBreakdown.tsx
```diff
@@ -69,90 +34,63 @@ const ProjectBreakdown = () => {
...
-          )
-          .fail(onAnyFail);
-      })
-      .fail(onAnyFail);
+    fetchData();
 
     return () => {
       setIsComponentMounted(false);
     };
+  }, [isComponentMounted]);
```
LMNTL, line 50: When I open the tab, the links to the projects work:
image

But after I switch pages, it looks like the asset UID is missing:
image

Looks like that's happening because this `useEffect()` doesn't run when the page changes.

Here's what I'd suggest: what if `getAssetUsageForOrganization()` took a page number (`currentPage`), instead of a URL, and you made this `useEffect()` depend on `currentPage`? Then you don't have to do any fetching in `handleClick()` - just increment/decrement the page number. You'll need to do something similar when adding sorting, so this should be good practice 👍 

This would also allow us to, in the future, take a page number in the URL (e.g. `/#/account/usage?p=2`) and jump

LMNTL (changes requested):

jsapp/js/account/usage/usageProjectBreakdown.tsx
RuthShryock: It was previously used when I had the double fetch call before I had this endpoint. So yes thank you! It should be removed.

kobo/apps/stripe/tests/test_organization_usage.py
```diff
@@ -13,9 +13,12 @@
 from kobo.apps.organizations.models import Organization, OrganizationUser
 from kobo.apps.trackers.submission_utils import create_mock_assets, add_mock_submissions
 from kpi.tests.api.v2.test_api_service_usage import ServiceUsageAPIBase
+from kpi.tests.api.v2.test_api_asset_usage import AssetUsageAPITestCase
+from kpi.models.asset import Asset
```
LMNTL, line 17: One last tiny thing, after which please merge this: these two new imports should get sorted alphabetically. 👍

LMNTL (approved):

PR #4883, opened 2024-03-25
Update webpack-dev-middleware

p2edwards:
## Description

Update webpack-dev-middleware to a newer version.

## Notes

The  [path traversal vulnerability](https://github.com/advisories/GHSA-wr3j-pwj9-hqq6) patched by this update does not seem to affect us directly:

- We don't use the `writeToDisk` option mentioned in the CVE
- We don't expose webpack-dev-server to public traffic nor use it in production

That said, it's good to update anyway. Someone self-hosting or developing might be running things a little differently.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -161,7 +161,7 @@
         "webpack": "^5.76.1",
         "webpack-bundle-tracker": "^1.5.0",
         "webpack-cli": "^4.9.2",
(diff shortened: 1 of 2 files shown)


p2edwards:
Closed by #4992 (upgrades webpack-dev-middleware to ^6.1.3, which includes the patch from 6.1.2)

PR #4885, opened 2024-03-25
Avoid storing KC domain in asset deployment data

bufke:
## Notes

Remove kobocat domain from kpi asset deployment data. This allows the domain of the kobocat server to change without breaking any linking.

The solution avoids saving the domain on any new data. It also normalizes (strips the domain) existing data.

## Alternatives

- Do nothing, this entire chunk of code will go away some day
- There is no need to store the url '/api/v1/forms/3' at all. It would be better to use only the existing form ID to construct the full URL. This would lower complexity and disk usage. However it would be a larger change for code that will be removed anyway
- Write a data migration to delete existing domain data. I felt this is unnecessary as the entire data set will be dropped later.

## Depends on

https://github.com/kobotoolbox/kpi/pull/4874


noliveleger (changes requested):

noliveleger:
LGTM. Let's merge it as soon as Django 4.2 is ready.

noliveleger:
> There is no need to store the url '/api/v1/forms/3' at all. It would be better to use only the existing form ID to construct the full URL. This would lower complexity and disk usage. However it would be a larger change for code that will be removed anyway

Agreed.  Could be part of the mid-term Kobocat as django-app migration.

PR #4886, opened 2024-03-27
File cleanup and grouped maintenance task

jamesrkiger:
## Checklist

3. [x] Ensure the tests pass

## Description

This PR creates a new perform_maintenance task that will include maintenance tasks that need to be run daily. Instead of scheduling each task separately, we can run each of them one at a time as part of `perform_maintenance`.  The only task included at the moment is the existing `remove_unused_markdown_files`. I have added a unit test for this task and made some slight optimizations to the task itself. The updated task uses `_raw_delete` to remove orphaned markdown files. Using the class delete method results in an unnecessary db call to delete upload_reference files that we already know don't exist.

## Notes

I tried using the new (as of Django 4.2) `InMemoryStorage` class in the interest of optimizing testing and removing the need to store media files in testing, but it broke a number of other tests and not in a way that seemed easily resolved.

### DIFF
--- a/kobo/settings/base.py
+++ b/kobo/settings/base.py
@@ -966,19 +966,19 @@ CELERY_BEAT_SCHEDULE = {
     'send-hooks-failures-reports': {
         'task': 'kobo.apps.hook.tasks.failures_reports',
         'schedule': crontab(hour=0, minute=0),
-        'options': {'queue': 'kpi_low_priority_queue'}
+        'options': {'queue': 'kpi_low_priority_queue'},
     },
     # Schedule every 30 minutes
     'trash-bin-garbage-collector': {
         'task': 'kobo.apps.trash_bin.tasks.garbage_collector',
         'schedule': crontab(minute=30),
-        'options': {'queue': 'kpi_low_priority_queue'}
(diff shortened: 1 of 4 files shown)


bufke:
Target beta instead of master

kobo/settings/base.py
```diff
@@ -966,19 +966,19 @@ def dj_stripe_request_callback_method():
...
         'schedule': crontab(minute=30),
-        'options': {'queue': 'kpi_low_priority_queue'}
+        'options': {'queue': 'kpi_low_priority_queue'},
     },
-    # Schedule every monday at 00:30
-    'markdown-images-garbage-collector': {
-        'task': 'kobo.apps.markdownx_upload.tasks.remove_unused_markdown_files',
-        'schedule': crontab(hour=0, minute=30, day_of_week=0),
-        'options': {'queue': 'kpi_low_priority_queue'}
+    # Run maintenance every day at 20:00 UTC
```
bufke, line 977: I check our Global and EU server usage and 20:00 UTC is perfect

bufke (approved):
I tested the queries in a production data set too and it lgtm

PR #4889, opened 2024-03-28
[TASK-623] Allow users to select which form questions to display in the single processing side bar

duvld:
## Description

Adds a checkbox that allows choosing which questions to display in the single processing sidebar. Checkbox only appears if "Submission Data" is toggled on.

## Notes

Because of the way `SubmissionDataList` hides questions, the displayed checkbox and the information sent to the store had to be inverted i.e., we send a list of 'hidden' questions rather the list of checked boxes.

## Related issues

NLP addons: https://www.notion.so/kobotoolbox/NLP-Qualitative-analysis-improvements-11b843e113dd478ca33010eefa2814b1?p=e1a31c80b74f4d569bf2505aa68fb9e2&pm=c


jsapp/js/components/processing/sidebar/sidebarDisplaySettings.tsx
```diff
@@ -132,6 +222,20 @@ export default function SidebarDisplaySettings() {
...
           </ul>
+
+          {props.asset?.survey &&
+            selectedDisplays.includes(StaticDisplays.Data) && (
+              <div className={styles.questionList}>
+                <strong>{t('Select the submission data to display.')}</strong>
+                <div className={styles.checkbox}>
+                  <MultiCheckbox
+                    type='bare'
+                    items={getCheckboxes()}
```
magicznyleszek, line 233: There is a bug. Steps are:
1. Have a project with multiple questions and have multiple submissions submitted
2. In Display Settings change displayed questions (e.g. leave only one displayed)
3. See that in sidebar the questions are hidden
4. Change submission (header arrow up or down)
5. See that in sidebar the questions are hidden
6. In Display Settings you will notice that all questions are selected :bug:

magicznyleszek: I think though that this is a semi good thing. Ideally we would like to keep all the display settings between the submissions, so the selected questions too.

So for some reason the selected questions are being reset

magicznyleszek (dismissed):

jsapp/js/components/processing/sidebar/sidebarDisplaySettings.tsx
magicznyleszek: I also see that these changes from our Zulip chat are not here yet:

- Put ~~it~~ *questions list* underneath "Submission data" - because if I understand this correctly, if I toggle off "Submission Data", all these checkboxes should be disabled? We could indent it from the left so it's clear it's a "sub section"
- Overflow the list of checkboxes; for max height I think we should go for 4.5, 5.5 or 6.5 (or more? not sure how much space we have there) of the height of single checkbox row (so If checkbox row is 24px, and we go with 5.5, max height would be 132px)
- After putting checkboxes underneath "Submission data" I think we don't need header inside the frame, but we could put some explanation above it (in regular text), like so (we already have "Select questions to be exported" in Data Downloads, so I think we can use something similar):
```
[x] SubmissionData
    Select questions to be displayed
    <checkboxes…>
```

Are you sure you've pushed all commits? :)

jsapp/js/components/processing/sidebar/sidebarDisplaySettings.tsx
```diff
@@ -158,6 +266,7 @@ export default function SidebarDisplaySettings() {
             color='light-blue'
             size='m'
             onClick={() => {
+              applyFieldsSelection();
               store.setDisplays(activeTab, selectedDisplays);
               setIsModalOpen(false);
             }}
```
magicznyleszek, line 272: I've found a small bug :bug::
1. I open settings modal
2. I have all questions selected
3. I unselect some questions
4. I've decided I don't want to change it so I close modal (e.g. with ESC) 
5. I open settings modal again
6. I have some questions selected (it should be all questions selected, i.e. it should reflect what we have in store)

jsapp/js/components/processing/sidebar/sidebarDisplaySettings.tsx
```diff
@@ -22,6 +33,25 @@ export default function SidebarDisplaySettings() {
...
+      .map((question) => {
+        // We make an object to show the question label to the user but use the
+        // name internally so it works with duplicate question labels
+        return {name: question.name, label: question.label};
+      });
+
+    return questionsList;
+  }
+
+  const [selectedFields, setSelectedFields] = useState(getInitialFields());
```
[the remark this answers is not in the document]
magicznyleszek: This would possibly be a useless change for now, but it makes sense to always treat the store as the ultimate source of this information. These `selectedFields` here should only be treated as temporary state for when the modal is being open.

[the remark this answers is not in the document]
magicznyleszek: See my comment about small bug below (one of the last comments) for actual problem with current ways. It's possible the flow needs to be changed a bit, so that it's loading the hidden/selected fields from the store always when modal opens (and for clarity it might also clean up after closing)

jsapp/js/components/processing/sidebar/sidebarDisplaySettings.tsx
```diff
@@ -61,6 +91,57 @@ export default function SidebarDisplaySettings() {
...
+      .filter((question) => question.checked)
+      .map((question) => {
+        return {name: question.name, label: question.label};
+      });
+
+    const hiddenList = list
+      .filter((question) => !question.checked)
+      .map((question) => question.name);
+
+    setHiddenFields(hiddenList);
```
magicznyleszek, line 129: After you add code for building this list inside `applyFieldsSelection` you can simply remove this and the `const` above

magicznyleszek (changes requested):

jsapp/js/components/processing/sidebar/sidebarDisplaySettings.tsx
duvld: > I think though that this is a semi good thing. Ideally we would like to keep all the display settings between the submissions, so the selected questions too.
> 
> So for some reason the selected questions are being reset

This actually should be fixed now by implementing what you said underneath (https://github.com/kobotoolbox/kpi/pull/4889#discussion_r1557802317)

jsapp/js/components/processing/sidebar/sidebarDisplaySettings.tsx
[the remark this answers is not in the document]
duvld: Should be fixed now :)

jsapp/js/components/processing/sidebar/sidebarDisplaySettings.tsx
```diff
@@ -22,6 +33,25 @@ export default function SidebarDisplaySettings() {
...
+        // We make an object to show the question label to the user but use the
+        // name internally so it works with duplicate question labels
+        return {name: question.name, label: question.label};
+      });
+
+    return questionsList;
+  }
+
+  const [selectedFields, setSelectedFields] = useState(getInitialFields());
+  const [hiddenFields, setHiddenFields] = useState<string[]>();
```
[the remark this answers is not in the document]
duvld: > I have also noticed that I don't see a code that is reading the hidden fields from the store when opening display settings… (to be continued in different comment (above))

Should be added correctly now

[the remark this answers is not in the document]
duvld: > You don't need to store hidden fields at all. Simply move `const hiddenList = …` to inside `applyFieldsSelection` function.
> 
> It would be something like this:
> 
> ```ts
>   function applyFieldsSelection() {
>     const hiddenFields: string[] = getCheckboxes()
>       .filter((checkbox) => !checkbox.checked)
>       .map((checkbox) => checkbox.name) || [];
>     store.setHiddenSidebarQuestions(hiddenFields);
>   }
> ```
> 
> And `resetFieldsSelection` function would become:
> 
> ```ts
>   function resetFieldsSelection() {
>     setSelectedFields(getInitialFields());
>     applyFieldsSelection();
>   }
> ```

This fix mostly worked, I left a comment in the code about the second part and why I had to set it to `[]` explicitly instead. There may be an even better solution but I think that is fine too, lmk

jsapp/js/components/processing/sidebar/sidebarDisplaySettings.tsx
magicznyleszek: I think a fix would be to apply `setSelectedFields(getInitialFields());` inside the `onRequestClose` function. This way after closing and reopening, the state would be fresh. Alternatively we could do it in the same function that is opening the modal (doing `setIsModalOpen(true)`). So either "cleanup afterwards" or "always start fresh" approach :D After doing either of these, please check if one of the places that resets its already is not needed (e.g. inside `resetFieldsSelection`)

jsapp/js/components/processing/sidebar/sidebarDisplaySettings.module.scss
```diff
@@ -30,3 +34,16 @@
...
+  margin: sizes.$x12 0 sizes.$x12 sizes.$x36;
+}
+
+.checkbox {
+  padding: sizes.$x12;
+  margin-top: sizes.$x6;
+  max-height: sizes.$x140; // ~4.5 times the size of a checkbox, makes 4 checkboxes visible
+  border: 1px solid colors.$kobo-gray-92;
+  border-radius: sizes.$x2;
+  overflow: scroll;
```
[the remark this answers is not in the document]
magicznyleszek: This one is not fixed yet :)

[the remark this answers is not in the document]
magicznyleszek: And I also think that `margin: 12px 0 0 36px;` would be good enough for that `.questionList` above

jsapp/js/components/processing/sidebar/sidebarDisplaySettings.tsx
duvld: I think these are still needed in `resetFieldsSelection` as resetting is not the same as closing the modal--on reset, we need to tell the store to clear the hidden fields as well as reset them visually with `setSelectedFields(getInitialFields());`

duvld: > I think this is a bug with the modal in general. Maybe we never caught it before but the same steps above can be taken for the regular toggle switches (the displays). And it's due to this: https://github.com/kobotoolbox/kpi/pull/4889/files/be15a3938a5956f36531d3e2325e0774cecfe635#diff-4e96bc318c90b92e52221bc45959e508183d0babeb08d081a876bcccdcc3d542L146-L149

I cleared this up by adding `setSelectedDisplays(store.getDisplays(activeTab))` into `onRequestClose` as well

PR #4893, opened 2024-04-02
[TASK-560] [TASK-561] Reflect Processing View tab in url

magicznyleszek:
## Checklist

3. [x] Ensure the tests pass

## Description

Processing View tabs are now being added to the URL as part of the path. This allows opening given tab from the copied URL. Changing submission will not change currently active tab.

## Notes

This is relatively huge given the task title. I had to do a lot of changes, because the code was not routes-ready. Plus it was written in The Old Way. And because of that it turned out to be a real can of worms…

Things changed here:
- Created `…processing/routes.tsx` for keeping all route related code for Processing View
  - Moved things here from `js/router/router.tsx` and expanded
  - Had to use small "hack" to make sure route params are passed with `<Navigate />`. Maybe there is a better way 🤷 
  - Didn't want to refactor too much, so all tabs-routes use the same component (as previously)
- Created `…processing/routes.utils.ts` file for different navigation related utilites; all of them have comments, so code should be understandable. 
  - Moved here some functions from `routerUtils.ts`
  - Moved here and renamed `goToProcessing` (previously `openProcessing`)
- Now active tab is not being stored in `singleProcessingStore` but rather is extrapolated from the URL
  - This, together with new way of navigating to a processing tab, caused a few simple changes throughout the code
- Had to rework a lot `onRouteChange` in `singleProcessingStore` to make it work with new routes. It swallowed a lot of code from previous tab changing functions, but had to do some mind wrapping to it afterwards.
- Renamed tabs from the weird abbreviated enum into simple type
- Transcriptized `PermProtectedRoute` while I was debugging one elusive bug
  - ❗ I've added one improvement here that limits the number of (hopefully) unnecessary calls to `api/v2/assets/<uid>`. This change might require more attention than most changes in this PR
- After changing submission (up/down arrow) in processing view, user will remain on the same tab

## Related issues

Related to #4854


jsapp/js/components/processing/singleProcessingStore.ts
```diff
@@ -30,12 +25,12 @@ import type {
...
+  isAnyProcessingRouteActive,
+  getProcessingPathParts,
+} from 'js/components/processing/routes.utils';
 
-export enum SingleProcessingTabs {
-  Transcript = 'trc',
-  Translations = 'trl',
-  Analysis = 'an',
-}
+export type ProcessingTabName = 'transcript' | 'translations' | 'analysis';
```
LMNTL, line 33: This should stay an enum - that way you won't have to use these strings in `DefaultDisplays`, `getActiveTab()`, `onRouteChange()`, and `cleanUpDisplays()`.

LMNTL (changes requested):

jsapp/js/components/processing/singleProcessingStore.ts
magicznyleszek: When it was an `enum`, I had to use `ProcessingTabName.Translations`. With `type` I can use shorter `'translations'`. With both of these the type safety is similar, right? I was using `enum`s a lot few years back, but I found out that while they are more "correct" theoretically speaking, a simple type is more concise.

I guess we should have a small discussion which way do we prefer?

magicznyleszek: I brought back `enum` :)

magicznyleszek: Thought I could make that `enum` use routes strings and avoid using that `TabToRouteMap`, but turns out you can compute only numbers :shrug:

jsapp/js/components/processing/routes.tsx
```diff
@@ -0,0 +1,52 @@
...
+import {PERMISSIONS_CODENAMES} from 'js/components/permissions/permConstants';
+import {PROCESSING_ROUTES} from 'js/router/routerConstants';
+
+// This is needed so we have access to params :shrug:
+const ProcessingRootRedirect = () => {
+  const params = useParams();
+  return (
+    <Navigate to={generatePath(PROCESSING_ROUTES.TRANSCRIPT, params)} replace />
+  );
+};
```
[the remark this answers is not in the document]
magicznyleszek: This was related to the `onRouteChange` from `singleProcessingStore`. Turned out to be quite a trip, I would say more like mountain hiking than walk around the neighbourhood :P I ended up writing more complex/precise code and some tests

jsapp/js/components/processing/routes.utils.ts
```diff
@@ -0,0 +1,195 @@
...
+  qpath?: string;
+  submissionEditId?: string;
+  tab?: ProcessingTab;
+}
+
+/**
+ * For given processing path, returns all of it's params and parts. If path is
+ * not provided, function is working on current path.
+ */
+export function getProcessingRouteParts(path?: string): ProcessingRouteParts {
```
magicznyleszek, line 36: @LMNTL TBH I don't know how I feel about this function :D It does work, and is quite packed with safeness, and has tests. But I feel that there needs to be less stupid way to do this :D

jsapp/js/components/processing/routes.utils.ts
```diff
@@ -0,0 +1,195 @@
...
+    pathArray[formsPartIndex] !== 'forms' ||
+    pathArray[dataPartIndex] !== 'data' ||
+    pathArray[processingPartIndex] !== 'processing'
+  ) {
+    // Not processing path, so we return "empty" results
+    return {
+      assetUid: undefined,
+      qpath: undefined,
+      submissionEditId: undefined,
+      tab: undefined,
```
magicznyleszek, line 72: @LMNTL This is good and makes sense, but it causes some `|| ''` in `singleProcessingStore`…

jsapp/js/components/processing/singleProcessingStore.ts
```diff
@@ -173,20 +162,25 @@ class SingleProcessingStore extends Reflux.Store {
...
     this.data.source = undefined;
-    this.data.activeTab = SingleProcessingTabs.Transcript;
     this.data.isPristine = true;
   }
 
   public get currentAssetUid() {
-    return getSingleProcessingRouteParameters().uid;
+    // To avoid too much complexity in the code, we pass empty string instead
+    // of undefined
+    return getProcessingRouteParts().assetUid || '';
```
magicznyleszek, line 171: @LMNTL I don't like that I had to put `|| ''` here (and below), but I feel like this is better than returning `string | undefined` and adding many `if`s through the code. WYT?

jsapp/js/components/processing/routes.utils.ts
LMNTL: Hmmm ... you could get the path using [URL.pathname](https://developer.mozilla.org/en-US/docs/Web/API/URL/pathname).

For the rest of the path construction/matching, it might be best to just fall back on what React Router uses - [`matchPath()`](https://reactrouter.com/en/main/utils/match-path#matchpath) and/or [`generatePath()`](https://reactrouter.com/en/main/utils/generate-path) look like they could have promise.

LMNTL (changes requested):

jsapp/js/components/processing/routes.utils.ts
magicznyleszek: I managed to make that `matchPath` work :)

jsapp/js/components/processing/singleProcessingStore.ts
magicznyleszek: Empty strings work just fine, good idea 👍

jsapp/js/components/processing/singleProcessingStore.ts
```diff
@@ -983,10 +1007,10 @@ class SingleProcessingStore extends Reflux.Store {
...
-    const allTabs: ProcessingTabName[] = [
-      'transcript',
-      'translations',
-      'analysis',
+    const allTabs: ProcessingTab[] = [
+      ProcessingTab.Transcript,
+      ProcessingTab.Translations,
+      ProcessingTab.Analysis,
     ];
     allTabs.forEach((tab) => {
```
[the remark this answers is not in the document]
magicznyleszek: Nice 👍

LMNTL (approved):

PR #4901, opened 2024-04-10
Asset snapshot cleanup task

jamesrkiger:
## Checklist

3. [x] Ensure the tests pass

## Description

Adds regular asset snapshot cleanup maintenance task. Reworks existing test to make sure latest versioned snapshots are preserved.

## Notes

I would appreciate feedback on organization here. KPI already has a tasks.py file for celery tasks, and the new `remove_old_snapshots()` is not a celery task, so I didn't want to put it there. Currently it is in a new maintenance_tasks.py file. I was thinking of putting them both in a tasks folder. One could be celery_tasks.py and the other could stay as maintenance_tasks.py.

Down the line, I'm also wondering if I should be removing the existing management command code ([here](https://github.com/kobotoolbox/kpi/blob/asset-snapshot-cleanup/kpi/management/commands/delete_assets_snapshots.py) for snapshots and here for the [DeleteBaseCommand](https://github.com/kobotoolbox/kpi/blob/asset-snapshot-cleanup/kpi/management/delete_base_command.py) class) once I replace the other task (removing old import tasks) that relied on it. 

## Related issues

Partially replaces work in #2434


kobo/tasks.py
```diff
@@ -9,3 +10,4 @@ def perform_maintenance():
     Run daily maintenance tasks
```
bufke, line 10: We talked before about adding an explicit task specific timeout - let's go do that now as we're adding more to it. It would be good to make sure it works as expected too. If you make it 1 second, will it really time out as expected? We don't want them retried either.

kpi/maintenance_tasks.py
```diff
@@ -0,0 +1,39 @@
...
+
+from kpi.models import AssetSnapshot
+
+
+default_retention_period = 90
+
+def remove_old_assetsnapshots():
+    days = (
+        constance.config.ASSET_SNAPSHOT_DAYS_RETENTION
+        or default_retention_period
```
bufke, line 15: I believe the way constance works is that it uses a default already, 30 in this case. Maybe double check that. I suspect the 90 here will never be used.

kpi/maintenance_tasks.py
```diff
@@ -0,0 +1,39 @@
...
+    )
+    max_snapshots_qs = (
+        AssetSnapshot.objects.exclude(asset_version=None)
```
bufke, line 18: This could be a huge number. We need to chunk it. I get counts exceeding 4 million rows.

There was a distinct here that you removed, can you explain why it was unnecessary?

Why use the list?

bufke (changes requested):
> I'm also wondering if I should be removing the existing management command code ([here](https://github.com/kobotoolbox/kpi/blob/asset-snapshot-cleanup/kpi/management/commands/delete_assets_snapshots.py) for snapshots and here for the [DeleteBaseCommand](https://github.com/kobotoolbox/kpi/blob/asset-snapshot-cleanup/kpi/management/delete_base_command.py) class) once I replace the other task (removing old import tasks) that relied on it.

Yeah for sure. Either now or later. We may want to consider our alternative abstractions like RemoveBaseCommand but not until we have more code to actually abstract.

kobo/tasks.py
jamesrkiger: I have added this with a 23 hour timeout [here](https://github.com/kobotoolbox/kpi/pull/4901/files#diff-06dc42ce86b33dc5058a5293a872d437206226e3f09f6fbbe32217ed459136a4). Note that I added a soft time limit, as well. This was necessary because if I only set the hard time limit it gets overridden by the global timeout.

kpi/tasks.py
```diff
@@ -95,15 +95,13 @@ def enketo_flush_cached_preview(server_url, form_id):
...
-    Temporary task to delete old snapshots.
-    TODO remove when kpi#2434 is merged
+    Run daily maintenance tasks
     """
-    call_command(
-        'delete_assets_snapshots',
-        days=constance.config.ASSET_SNAPSHOT_DAYS_RETENTION,
-        asset_id=asset_id,
-    )
+    from kobo.apps.markdownx_uploader.tasks import remove_unused_markdown_files
```
bufke, line 103: Circular import? Might be good to note. Did you spend any time looking into why?

kpi/tasks.py
```diff
@@ -95,15 +95,13 @@ def enketo_flush_cached_preview(server_url, form_id):
     )
     response.raise_for_status()
 
-
-@celery_app.task
-def remove_asset_snapshots(asset_id: int):
+@celery_app.task(time_limit=82800, soft_time_limit=82800)
```
bufke, line 98: Maybe comment with a human readable time. Do we really need both? We won't need to catch SoftTimeLimitExceeded. If you do end up keeping both I would set a constant with a variable name that denotes the human readable time.

bufke (approved):
Approved pending consideration of my comments.

kpi/tasks.py
```diff
@@ -95,15 +93,13 @@ def enketo_flush_cached_preview(server_url, form_id):
     )
     response.raise_for_status()
 
+limit_hours_23 = 82800
```
bufke, line 96: CONSTANTS SHOULD BE CAPS

kpi/tasks.py
noliveleger: @jamesrkiger, you may want to look at `settings.CELERY_LONG_RUNNING_TASK_TIME_LIMIT` and `settings.CELERY_LONG_RUNNING_TASK_SOFT_TIME_LIMIT`

kpi/maintenance_tasks.py
```diff
@@ -0,0 +1,35 @@
+import constance
+from datetime import timedelta
+
+from django.db.models import Exists, OuterRef, Q
+from django.utils import timezone
+
+from kpi.models import AssetSnapshot
+
+
+def remove_old_assetsnapshots():
```
noliveleger, line 10: Let's be consistent and keep the same name convention. Please rename `remove_old_assetsnapshots` to `remove_old_asset_snapshots`

noliveleger (changes requested):

kpi/maintenance_tasks.py
```diff
@@ -0,0 +1,35 @@
...
+        .order_by('pk')
+    )
+
+    while True:
+        try:
+            deletion_delimiter = delete_queryset.values_list('pk', flat=True)[
+                1000:1001
+            ].get()
+            delete_queryset.filter(id__lte=deletion_delimiter).delete()
+        except AssetSnapshot.DoesNotExist:
```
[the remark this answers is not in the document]
bufke: Sadly the ORM does [not support this](https://docs.djangoproject.com/en/dev/ref/models/querysets/#delete), see my [SO post](https://stackoverflow.com/questions/60736901/how-to-fetch-data-in-chunks-from-db-in-django-and-then-delete-them/77158646#77158646). Feel free to suggest changes on my general approach as posted there. I usually will take a "break" over an extra variable assignment but that is stylistic. Python lack closures, so there is no way to isolate a variable to just the loop.

We use this workflow so many times, it's nice to have more eyes on it.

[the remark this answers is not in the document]
jamesrkiger: After discussion, we're going with @noliveleger's suggestion. I will use this method for subsequent cleanup tasks, as well.

PR #4904, opened 2024-04-15
[TASK-548] NLP add rtl script support

magicznyleszek:
## Checklist

3. [x] Ensure the tests pass

## Description

Any transcript or translation text written in RTL script will be displayed correctly.

## Notes

The `dir='auto'` attribute is magically doing all the job. I thought there would be a way to set it as default for every HTML node, but it doesn't seem to work. Adding it to each node with RTL text does the job.

### DIFF
--- a/kpi/templates/base_simple.html
+++ b/kpi/templates/base_simple.html
@@ -2,7 +2,7 @@
 
 {% load static %}
 <!doctype html>
-<html dir="auto">
+<html>
   <head>
     <!-- Setup -->
     <meta charset="utf-8">
--- a/kpi/templates/base_simple.html
+++ b/kpi/templates/base_simple.html
@@ -2,7 +2,7 @@
 
 {% load static %}
 <!doctype html>
-<html>
+<html dir="auto">
   <head>
     <!-- Setup -->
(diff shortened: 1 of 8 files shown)


kpi/templates/base_simple.html
```diff
@@ -2,7 +2,7 @@
 
 {% load static %}
 <!doctype html>
-<html>
+<html dir="auto">
```
LMNTL, line 5: Does this line do anything? As you noted in the PR description, `dir="auto"` doesn't appear to transfer down to child elements. I can't find a difference on `/accounts/login/` - the layout, text direction, and input text direction appear to all stay the same as on `beta`.

LMNTL (approved):
Works for me! I have a functional question about the Django templates, but it's not a blocker - the addition seems benign, just not sure what it's doing.

kpi/templates/base_simple.html
magicznyleszek: I thought that although it doesn't transfer down, it does something sometimes :P I tried to find any documentation or SO question that would confirm that, but there is nothing. I removed it

PR #4920, opened 2024-04-24
Add maintenance function to remove old import tasks

jamesrkiger:
## Checklist

3. [x] Ensure the tests pass

## Description

Adds maintenance task for cleaning up old import tasks. 

## Notes

I created a new test file for this. All other import task related tests are run in API test files and this didn't seem relevant to API testing. 

## Related issues

implements work from #2434

### DIFF
--- a/kpi/maintenance_tasks.py
+++ b/kpi/maintenance_tasks.py
@@ -5,6 +5,7 @@ from django.db.models import Exists, OuterRef, Q
 from django.utils import timezone
 
 from kpi.models import AssetSnapshot, ImportTask
+from kpi.utils.chunked_delete import chunked_delete
 
 
 def remove_old_asset_snapshots():
@@ -20,12 +21,7 @@ def remove_old_asset_snapshots():
         date_created__lt=timezone.now() - timedelta(days=days),
     ).filter(Exists(newer_snapshot_for_asset) | Q(asset_version=None))
 
-    while True:
-        count, _ = delete_queryset.filter(
-            pk__in=delete_queryset[:1000]
-        ).delete()
-        if not count:
-            break
+    chunked_delete(delete_queryset)
 
 
 def remove_old_import_tasks():
@@ -33,9 +29,4 @@ def remove_old_import_tasks():
         date_created__lt=timezone.now() - timedelta(days=90),
     )
 
-    while True:
-        count, _ = delete_queryset.filter(
-            pk__in=delete_queryset[:1000]
-        ).delete()
-        if not count:
-            break
+    chunked_delete(delete_queryset)
--- a/kpi/maintenance_tasks.py
+++ b/kpi/maintenance_tasks.py
@@ -25,8 +25,10 @@ def remove_old_asset_snapshots():
 
 
 def remove_old_import_tasks():
+    days = constance.config.IMPORT_TASK_DAYS_RETENTION
+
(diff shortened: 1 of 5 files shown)


kpi/maintenance_tasks.py
```diff
@@ -26,3 +26,16 @@ def remove_old_asset_snapshots():
         ).delete()
         if not count:
             break
+
+
+def remove_old_import_tasks():
+    delete_queryset = ImportTask.objects.filter(
+        date_created__lt=timezone.now() - timedelta(days=90),
```
bufke, line 33: @noliveleger in the original PR you wrote "(90 by default)" is there any reason to make this configurable? I like hard coding it, if not. Would a private server ever say want to make it longer?

kpi/maintenance_tasks.py
```diff
@@ -26,3 +26,16 @@ def remove_old_asset_snapshots():
...
+    )
+
+    while True:
+        count, _ = delete_queryset.filter(
```
bufke, line 37: Now that we're using this a few times, we could consider adding abstraction. I'm fine doing this right now or in the next batch of work. 

`chunked_delete(queryset: QuerySet)`

kpi/tests/test_import_tasks.py
```diff
@@ -0,0 +1,32 @@
...
+    def setUp(self):
+        self.user = User.objects.get(username='someuser')
+
+    def test_remove_old_import_tasks(self):
+        old_task = ImportTask.objects.create(
+            user=self.user,
+            data='{}',
+        )
+
+        old_task.date_created = timezone.now() - timedelta(days=95)
```
bufke, line 24: I assume we couldn't set date_created directly and this works around it. Does it work when using model bakery?

bufke (approved):
LGTM pending the review of my comments

kpi/maintenance_tasks.py
jamesrkiger: @bufke This will actually be the last PR for the KPI-portion of the Celery maintenance work. Should I just abstract this now or should we hold off until some theoretical future maintenance task tips it over into definite DRY territory?

kpi/tests/test_import_tasks.py
jamesrkiger: Correct, we can't set it on creation. Doesn't work with model bakery, either. I have added a note

kpi/maintenance_tasks.py
noliveleger: It's part of an old discussion of the back-end team. IIRC, yes some private servers wanted a different value. So please keep it configurable.

kpi/tests/test_import_tasks.py
```diff
@@ -0,0 +1,32 @@
+# coding: utf-8
+from datetime import timedelta
+
+from django.contrib.auth.models import User
```
noliveleger, line 4: Please don't import `User` directly.  

See https://chat.kobotoolbox.org/#narrow/stream/55-KoboCAT-Replacement/topic/KoboCAT.20as.20Django.20app/near/334486

noliveleger (changes requested):

kpi/maintenance_tasks.py
jamesrkiger: Added constance var

PR #4922, opened 2024-04-29
Fix tag entry behavior bug

RuthShryock:
## Checklist

3. [X] Ensure the tests pass

### DIFF
--- a/jsapp/js/components/common/koboTagsInput.tsx
+++ b/jsapp/js/components/common/koboTagsInput.tsx
@@ -80,6 +80,7 @@ class KoboTagsInput extends React.Component<KoboTagsInputProps> {
           inputProps={inputProps}
           onlyUnique
           addOnBlur
+          addOnPaste
           pasteSplit={this.pasteSplit}
         />
       </div>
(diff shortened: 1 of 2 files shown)


jsapp/js/components/common/koboTagsInput.tsx
```diff
@@ -80,7 +80,6 @@ class KoboTagsInput extends React.Component<KoboTagsInputProps> {
           inputProps={inputProps}
           onlyUnique
           addOnBlur
-          addOnPaste
```
magicznyleszek, line 83: Ok, note here that you are modifying `KoboTagsInput`. This component is a wrapper for `react-tagsinput`, and we use it in most of the places we render tags editor. But in NLP we don't use it, we use `react-tagsinput` directly (so that we can customize it more). The component you should look at is `TagsResponseForm`. And there is already [a PR](https://github.com/kobotoolbox/kpi/pull/4905/files) that modifies it - I've reassigned it to you right now :)

magicznyleszek: This way we fix both problems:
- pasting doesn't split by spaces
- pasting doesn't remove the already typed in text

RuthShryock: Sounds good!

PR #4928, opened 2024-05-03
Fix custom exports of grouped NLP questions

jnm:
## Checklist

3. [x] Ensure the tests pass

## Description

Solves the problem of transcripts and translations for grouped source questions not appearing when custom exports include only a subset of fields.

## Notes

Many questions remain about the original intent of the code, but given the comment in the original commit, d013bfe0f5, it's likely there's just extraneous stuff here that needs to be removed (after understanding that it's safe to do so):

    this should be pared down to just the atts that are used
    in the frontend and formpack


noliveleger:
Work as expected!

PR #4935, opened 2024-05-13
Track addon usage over time

jamesrkiger:
## Checklist

3. [x] Ensure the tests pass

## Description

The basic goal here is to enable the frontend to display how many units of addon consumption were used in the current billing period. For that, we need to track addon usage as it happens. This PR aims to do that by adding new addon usage keys to the NLPUsageCounter jsonfield dict whenever an addon's usage is incremented.  The aggregated data is served to the frontend via the service_usage/organization service_usage endpoints.

Related changes:
- Overhauled `is_organization_over_plan_limit` method to check not just whether an org is over its limit, but whether it will go over its limit with a specified usage amount.

- Updated project ownership transfer code to remove addon usage counts from transferred NLPUsageCounters, as these counts will not be relevant to a new owner.

Unrelated changes:
- Fixed tests that were already broken by earlier changes in `test_one_time_addons_api.py`

## Notes

There are still some deeper issues with how we handle usage caching that will need to be resolved before this project is completed. But I think the work I have done so far could be merged into the project branch so the PR for that cache work will be more manageable.

Credit to @noliveleger for the new `IncrementValues` ORM helper.

### DIFF
--- a/kobo/apps/stripe/constants.py
+++ b/kobo/apps/stripe/constants.py
@@ -27,10 +27,3 @@ USAGE_LIMIT_MAP = {
     'storage': 'storage_bytes',
     'submission': 'submission',
 }
-
-USAGE_LIMIT_MAP_STRIPE = {
-    'character': 'nlp_character',
-    'seconds': 'nlp_seconds',
-    'storage': 'storage_bytes',
-    'submission': 'submission',
-}
--- a/kobo/apps/stripe/constants.py
+++ b/kobo/apps/stripe/constants.py
@@ -27,3 +27,10 @@ USAGE_LIMIT_MAP = {
     'storage': 'storage_bytes',
     'submission': 'submission',
 }
+
+USAGE_LIMIT_MAP_STRIPE = {
+    'character': 'nlp_character',
+    'seconds': 'nlp_seconds',
+    'storage': 'storage_bytes',
+    'submission': 'submission',
+}
--- a/kobo/apps/stripe/tests/test_one_time_addons_tracking.py
+++ b/kobo/apps/stripe/tests/test_one_time_addons_tracking.py
@@ -135,6 +135,7 @@ class OneTimeAddonTrackingTestCase(BaseTestCase):
         self.charge.save()
 
     def test_increment_addon_usage(self):
+        # TODO: Remove once cache issues have been resolved
         # update_nlp_counter will error out if cache isn't updated first
         # Also note that the used asr_seconds_current_month has to be > subscription limit,
         # otherwise the addon will not be incremented (and neither will the counter, therefore)
@@ -178,6 +179,7 @@ class OneTimeAddonTrackingTestCase(BaseTestCase):
         )
 
     def test_increment_addon_usage_over_limit(self):
+        # TODO: Remove once cache issues have been resolved
         self.organization.update_usage_cache(
             {
                 'total_nlp_usage': {
--- a/kobo/apps/stripe/tests/test_one_time_addons_tracking.py
+++ b/kobo/apps/stripe/tests/test_one_time_addons_tracking.py
@@ -1,7 +1,5 @@
 from dateutil.relativedelta import relativedelta
 from django.contrib.auth.models import User
-from django.core.cache import cache
-from django.urls import reverse
 from django.utils import timezone
 from djstripe.models import (
     Charge,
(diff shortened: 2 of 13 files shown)


kobo/apps/stripe/constants.py
```diff
@@ -27,10 +27,3 @@
     'storage': 'storage_bytes',
     'submission': 'submission',
 }
-
-USAGE_LIMIT_MAP_STRIPE = {
```
jamesrkiger, line 31: If I'm not mistaken, this alternative mapping const is no longer needed. We use asr/mt rather than 'nlp' in stripe products now, right?

kobo/apps/stripe/tests/test_one_time_addons_tracking.py
```diff
@@ -0,0 +1,196 @@
...
+from kobo.apps.trackers.models import NLPUsageCounter
+from kobo.apps.trackers.submission_utils import (
+    create_mock_assets,
+)
+from kobo.apps.stripe.models import PlanAddOn
+from kobo.apps.trackers.utils import update_nlp_counter
+from kpi.tests.api.v2.test_api_service_usage import ServiceUsageAPIBase
+
+
+class OneTimeAddonTrackingTestCase(ServiceUsageAPIBase):
```
jamesrkiger, line 27: This is still an interim test class at the moment, but I do wonder how I should clean it up once it is ready. There is a fair amount of overlap with `ServiceUsageAPIBase` and `OrganizationUsageAPITestCase`. Ideally, I think I would create a base test class with methods flexible enough to work in all three classes. But that might be better left to another PR.

kpi/deployment_backends/kobocat_backend.py
```diff
@@ -1286,9 +1286,19 @@ def transfer_submissions_ownership(
 
     def transfer_counters_ownership(self, new_owner: 'auth.User'):
 
+        # Addon usage is not relevant to new owner, so we zero it
         NLPUsageCounter.objects.filter(
             asset=self.asset, user=self.asset.owner
-        ).update(user=new_owner)
+        ).update(
+            user=new_owner,
+            counters=UpdateJSONFieldAttributes(
```
jamesrkiger, line 1294: Ideally, I think it would be better to use a custom function that allows for removing these keys, rather than changing them to 0. But psql functions are still new territory for me, so it might take me a bit to figure out how. Let me know if you think it would be worth it.

kobo/apps/stripe/tests/test_one_time_addons_tracking.py
```diff
@@ -0,0 +1,196 @@
...
+            'quantity': quantity,
+            **(product.metadata),
+        }
+        self.charge.save()
+
+    def test_increment_addon_usage(self):
+        # update_nlp_counter will error out if cache isn't updated first
+        # Also note that the used asr_seconds_current_month has to be > subscription limit,
+        # otherwise the addon will not be incremented (and neither will the counter, therefore)
+        self.organization.update_usage_cache(
```
jamesrkiger, line 152: As I mention in the note in the code above, `update_nlp_counter()` does not work for asr/mt usage types if the org usage cache hasn't been updated previously. Unless I'm mistaken, the only time this cache gets updated currently is when the org usage endpoint is accessed. Should we initialize the org usage fields with data that can be used in the process of running `update_nlp_counter()` or should this function be changed to not depend only on this cache data?

kobo/apps/trackers/utils.py
```diff
@@ -43,17 +46,36 @@ def update_nlp_counter(
 
     # Update the total counters by the usage amount to keep them current
     kwargs = {}
+    addon_amount_used = 0
     if service.endswith('asr_seconds'):
         kwargs['total_asr_seconds'] = F('total_asr_seconds') + amount
         if asset_id is not None:
             # If we're not updating the catch-all counter, increment any NLP add-ons the user may have
-            PlanAddOn.increment_add_ons_for_user(user_id, 'seconds', amount)
+            remaining = PlanAddOn.increment_add_ons_for_user(
```
jamesrkiger, line 54: `increment_add_ons_for_user()` gets a list of relevant addons and, for each addon, checks [is_organization_over_plan_limit()](https://github.com/kobotoolbox/kpi/blob/06b50e2712272d53f92cddfa9bcf4cb43b533613/kobo/apps/organizations/models.py#L77) for that organization. Maybe I'm missing a step somewhere, but it seems to me like this only applies to cases where a user has reached their plan limit at some point in the past. But consider a case where I have a plan limit of 1000 asr minutes and an addon for 2000 additional asr minutes, then I use 2000 asr minutes.  I would expect my addon minutes limits_remaining to go down to 1000. Currently it doesn't change at all. 

If you are curious about this, the [addon tracking tests](kobo/apps/stripe/tests/test_one_time_addons_tracking.py) above can be changed slightly to test it out.

jamesrkiger: Changed `is_organization_over_plan_limit` to `check_usage_exceeds_plan_limit`, which can check against incoming usage.

kobo/apps/stripe/tests/test_one_time_addons_tracking.py
jamesrkiger: Planning to address this issue in a subsequent PR.

bufke (changes requested):
I asked about why we track addon_used_asr_seconds. @jamesrkiger is going to work on some documentation that will help me in this review. I'll leave these small comments for now and review again after.

jamesrkiger:
I am closing this because we have decided not to implement tracking of addon usage over time (as opposed to simply incrementing addon usage when addon minutes/characters have been used). Some of this code may still be useful, but most of it is now irrelevant.

PR #4941, opened 2024-05-20
Remove the Note qualitative analysis question type from exports 

RuthShryock:
## Checklist

3. [X] Ensure the tests pass

## Notes

To test: 

- Checkout `beta` and create a project with an audio question 
- Submit a response and add a note in the analysis view 
- Make an export and you can see an extra column created for the note
- Checkout this branch and restart your development environment 
- Make another export of the project and confirm that note doesn't generate an extra column

### DIFF
--- a/kpi/models/asset.py
+++ b/kpi/models/asset.py
@@ -502,6 +502,9 @@ class Asset(ObjectPermissionMixin,
                 settings='??',
                 path=[qpath, qual_question['uuid']],
             )
+            if field['type'] == 'qual_note':
+                # don't include note question type in exports
+                continue
             try:
                 field['choices'] = qual_question['choices']
             except KeyError:
--- a/kpi/models/asset.py
+++ b/kpi/models/asset.py
@@ -471,7 +471,10 @@ class Asset(ObjectPermissionMixin,
             # Remove newlines and tabs (they are stripped in front end anyway)
             self.name = re.sub(r'[\n\t]+', '', _title)
 
-    def analysis_form_json(self):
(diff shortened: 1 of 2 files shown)


kpi/models/asset.py
```diff
@@ -502,6 +502,9 @@ def analysis_form_json(self):
                 settings='??',
                 path=[qpath, qual_question['uuid']],
             )
+            if field['type'] == 'qual_note':
+                # don't include note question type in exports
+                continue
```
jnm, line 507: Hey @RuthShryock, thanks for working on this. Since the this method, `analysis_form_json()`, is not specific to exports and is also used to [generate API responses](https://github.com/kobotoolbox/kpi/blob/1fdd805e7733d0b6f12e3e20af6930383b82bc39/kpi/serializers/v2/asset.py#L474-L475), could you please:

1. Add a new parameter to this method; let's call it `omit_question_types`. This should be able to take a list of types that should be skipped.
2. Make the default value for the parameter `None`
    * It might seem like we should use an empty list (`[]`) as the default, but see https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments for reason why we shouldn't :)
3. Change your logic to `continue` if `field['type']` matches any of the types in `omit_question_types`
4. Find where `analysis_form_json()` is called by the export code, and pass `omit_question_types=['qual_note']` to it there.

Let me know if you have any questions :)

jnm (changes requested):

jnm (approved):
Awesome, thank you :tada:

PR #4943, opened 2024-05-21
Fix styling for add-ons

RuthShryock:
## Checklist

3. [X] Ensure the tests pass

### DIFF
--- a/jsapp/js/account/add-ons/addOnList.component.tsx
+++ b/jsapp/js/account/add-ons/addOnList.component.tsx
@@ -137,7 +137,7 @@ const AddOnList = (props: {
                         )}
                     </span>
                     <Badge color={color} size={'s'} label={badgeLabel} />
-                    <p className={styles.description}>
+                    <p className={styles.addonDescription}>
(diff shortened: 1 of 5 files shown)


jamesrkiger:
@RuthShryock Couple issues for you:

1. I don't know if there have been any explicit discussions of solutions for this, but I'm wondering if vertical scrolling is the best approach here. Could we get the elements of these addons to stack vertically so mobile users don't have to scroll?

2. I'm not sure if this is a new issue or not, but the combination of horizontal scrolling overflow and addon dropdowns is causing some erratic display issues. When I open a dropdown it expands into the overflow area and isn't visible without scrolling down and, at the same time, causes the container to jump slightly. Can we make it so that the dropdown is fully visible when we open it and doesn't cause any screen jumps?

RuthShryock:
@jamesrkiger I added a styling to make the add-ons responsive for smaller screens, they stack vertically and horizontal scroll is disabled so we don't have the dropdown display issues. More work can probably be done on this later after feedback from Tino and Tessa when this feature is being reviewed :)

Issue #4944, opened 2024-05-23
Improve broken translation strings


p2edwards:
## Description

Ideally, each string we send to Transifex is a cohesive translatable unit, with enough context to translate it well.

We've fallen short on this ideal in a few places. Let's collect related PR's under this issue.

#### Things to improve

- Replace fragments like `t('This is ') + isNot ? t('NOT') : t('') + t(' how we do it')` with proper templating.
  - These arrive all out of order in Transifex and are not grammatically friendly.
  - These are easy to find by sorting the keys in `extracted-strings.json` and manually reading top to bottom, or applying regex. We're tracking some of these [here](https://www.notion.so/kobotoolbox/Fix-broken-translations-strings-abc52b6cf9524f4ba213311f7147b18d?pvs=4) (internal link)
- Fix some strings that were missing from the build output entirely for some reason
  - e.g. `t('string ' + 'concatenation')` was failing silently for a while, causing these keys to never reach `extracted-strings.json`.
  - There's a fix for this coming.
- Follow established or predictable conventions when it comes to field interpolation, links, or formatting; don't use HTML tags in translated strings.
 
#### Upcoming PR's

- [ ] PR to **fix string concatenation** in t() calls
- [ ] PR to add a **translation templating helper** and fix some related fragment and web syntax related problems.

p2edwards:
### Detecting issues sooner 

Maybe we should consider un-ignoring `extracted-strings.json` at some point (at least on the frontend.)[^1] I know it's abnormal to check in generated assets or build artifacts,[^2] but one upside is we'd be more likely to spot potential translation issues earlier, like during code review.

[^1]: If we do this to make spot-checking easier then we should sort the output by key, and maybe simplify from json `{"key":"key"}` to array `["key"]` format, or plaintext, one per line with escaped `\n`'s.
[^2]: This might be a different idea but it sounds related: #4801

PR #4945, opened 2024-05-23
Allow concat in frontend t()-string sources

p2edwards:
## Description

Safely allow string concatenation in frontend translated strings.

## Purpose

- Restore some translation strings that were missing
- Improve code formatting flexibility in the future

## Details

Supports usage like 

```js
   const textContent = __(
     'This translation string is split ' +
     'across several lines to' + // to allow this kind of comment and
     'keep the column width low.'
   );
```

Without this fix, examples like the above would be **missing from the build output without warning.**[^1] If you compare `extracted-strings.json` before and afterwards you'll see maybe 8 or 9 translation strings turn up that were previously missing. Some of them can be seen here: [GitHub search :mag_right:](https://github.com/search?q=repo%3Akobotoolbox%2Fkpi%20%2F%5B%5Ea-zA-Z_%60%5Dt%5C(%5Cs*%5Cn%3F%5B%5E%5Cn)%5D*%5B%27%22%60%5D%5Cs%2B%5C%2B%2F&type=code)

[^1]: Previously discussed [here](https://chat.kobotoolbox.org/#narrow/stream/4-Kobo-Dev/topic/behavior.20of.20webpack-extract-translation-keys.20plugin) and [here](https://chat.kobotoolbox.org/#narrow/stream/4-Kobo-Dev/topic/Context.20in.20translations/near/298673) (internal links)

I wrote some more explanation about how this patch works in [my upstream PR :fork_and_knife:](https://github.com/grassator/webpack-extract-translation-keys/pull/25) on **grassator/webpack-extract-translation-keys**.

- **The diff there is likely more reviewable** than the one here, so read more there if you're curious.
- The version bump here (6.0.0->6.1.0) is because my other patch got merged. I finally got around to upgrading us so I could base the new patch on that.
- Right now we apply this patch with [patch-package](https://www.npmjs.com/package/patch-package), but if it gets merged upstream we can change this to just bumping the version number again.

## Related issues

Part of #4944 

## Checklist

2. [x] ~~If you've changed APIs, update (or create!) the documentation~~
3. [x] Ensure the tests pass
7. [x] ~~Open an issue in the [docs](https://github.com/kobotoolbox/docs/issues/new) if there are UI/UX changes~~


jnm (approved):
Tested and found several additional strings added to `extracted-strings.json`

PR #4949, opened 2024-05-24
Fix Cypress tests sufficiently for a proof of concept

jnm:
## Description

Update Cypress tests so that the first test, which logs in and creates a new project, works again.

## Notes

The tests were broken by Django upgrades, the switch to django-allauth, removal of valid webpack configuration from the testing environment (#4517), and UI layout changes.

`project.spec.js` now works, but `question.spec.js` and `delete.spec.js` are still broken due to UI changes.

NB: the `/api/v2/organizations/…/service_usage/` endpoint currently cannot be served by `./manage.py cypress_testserver`; it will log this failure:
```
django.db.utils.ProgrammingError: relation "logger_xform" does not exist
LINE 1: ...ttachment_storage_bytes"), 0) AS "bytes_sum" FROM "logger_xf...
```
There's a corresponding error message in the UI, but this is not the cause of any Cypress tests failing.

### DIFF
--- a/cypress/README.md
+++ b/cypress/README.md
@@ -2,8 +2,8 @@
 
 To run tests, you need to do 2 things:
 
-1. Start the **test server.** This is like a kpi server, but in a special cypress_testserver mode. (You will also need to restart it between Cypress test runs.)
-2. Use the **Cypress test runner.** You can run this in command-line mode, or open an interactive browser window.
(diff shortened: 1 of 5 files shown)


p2edwards (approved):
Looks good!

Ran the tests, following the instructions in `cypress/README.md`. Login works!

I made some light edits to the readme for clarity. Most notably, a reminder to install dev dependencies for your test server if you hadn't already.

PR #4950, opened 2024-05-24
Update logic for Stripe APIs

RuthShryock:
## Checklist

3. [X] Ensure the tests pass

## Description

Djstripe updated from using settings to database values, requiring an update to how we reference API keys and our logic for enabling Stripe.


bufke (approved):
Thanks, this looks good. @RuthShryock we should note that our public servers will need STRIPE_ENABLED set to true before this goes live. I'll proceed on setting that now, no need to wait.  Just be cognizant on changes like that generally - we don't want the left hand to break what the right hand is doing :smile:

PR #4964, opened 2024-06-03
Fix "Make public" button

magicznyleszek:
## Checklist

3. [x] Ensure the tests pass

## Description

Fixes "Make public"/"Make private" button for collection that was being frozen in pending state indefinitely.

## Notes

Reproduction:
1. Got to "My Library"
2. Enter a collection
3. Make sure collection has all required meta fields
4. Click "Make public"
5. Notice that button is grayed out (meaning pending) forever :bug:

Things changed:
- New `teal` color for `Button` (needed for public collections)
- Updated `AssetPublicButton` component:
  - Replaced deprecated buttons with the common `Button` component
  - Now button is clearly pending when doing things (previously it looked disabled)
  - Moved `isAwaitingFreshPermissions` flow from `actions.es6` into the component
  - Added some commments

The problem was because of the asset caching, as this component relies on getting fresh asset after the public status of collections is changed. Thus `, true)` was added to `actions.resources.loadAsset` call.

## Related issues

Built atop #4961

### DIFF
--- a/jsapp/js/components/library/assetPublicButton.tsx
+++ b/jsapp/js/components/library/assetPublicButton.tsx
@@ -4,6 +4,7 @@ import {actions} from 'js/actions';
 import assetUtils from 'js/assetUtils';
 import {ASSET_TYPES} from 'js/constants';
 import {notify} from 'js/utils';
+import Button from 'js/components/common/button';
 import type {AssetResponse} from 'js/dataInterface';
 
 interface AssetPublicButtonProps {
@@ -12,6 +13,10 @@ interface AssetPublicButtonProps {
 
 interface AssetPublicButtonState {
   isPublicPending: boolean;
+  /**
+   * After asset public state is changed, we wait for the asset to be loaded
+   * again, so that we know from the permissions `assetUtils.isAssetPublic`.
+   */
   isAwaitingFreshPermissions: boolean;
 }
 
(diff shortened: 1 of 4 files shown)


jsapp/js/components/library/assetPublicButton.tsx
```diff
@@ -51,8 +56,20 @@ export default class AssetPublicButton extends React.Component<
...
+
+      // We need to get fresh asset here, so that new permissions would be
+      // available for the button code. We rely on the fact that new asset
+      // would be passed through `props` and `componentWillReceiveProps` will
+      // unlock the button again.
+      //
+      // TODO: this flow should be improved, but it might require some more
+      // thought, as the asset data flow in the whole app should be redone
+      // (after very thorough planning). Unfortunately many places in the app
+      // have this problem.
```
p2edwards, line 71: Agreed, and there are some problems with state getting out of sync.

- It's a little too easy to trigger an infinite loading spinner just by clicking around (breadcrumbs, asset in collection, back button)
- If you create a new empty collection, add an item to it, and go back to collection, you'll see the item in it, but if you try to make it public, you'll get an error because some part of the js still thinks the collection is empty.

Hitting "refresh" usually clears these out, but it would be nice to fix these.

p2edwards (approved):

PR #4967, opened 2024-06-04
Add constance setting to enforce superuser authentication

RuthShryock:
## Checklist

3. [X] Ensure the tests pass

## Description

Require MFA for superusers with a usable password.

## Notes

`SUPERUSER_AUTH_ENFOREMENT` constance value is initially set to `False` so that server owners can customize superuser authentication requirements.


hub/admin/extend_user.py
```diff
@@ -310,3 +310,14 @@ def _get_message(self, singular: bool, grace_period: int) -> str:
             message += f'View <a href="{url}">trash.</a>'
 
         return mark_safe(message)
+
+    def save_model(self, request, obj, form, change):
+        if obj.is_superuser:
+            if config.SUPERUSER_AUTH_ENFORCEMENT:
+                if obj.has_usable_password() and not config.MFA_ENABLED:
```
bufke, line 317: I love how this doesn't actually need to check anything about SSO to work. If some admin wants to make a superuser account that has no authentication at all, then sure why not. :smile: :+1: 

> config.MFA_ENABLED

The docstring on that setting is 'Enable two-factor authentication'. This isn't checking if the user has MFA set up, it's only checking if the configuration enables MFA to potentially be set up. That's my understanding at least. We need to know specifically if this user (obj) has MFA set up.

Let's add a unit test to this logic. However, that may be a little hard in Django admin. There certainly is no need to test django admin itself. Could you either

- Test it via django admin - maybe I'm wrong and it's simple
- Abstract this out to some function and then unit test just that function. It seems reasonable we may end up reusing the logic anyway some day.

I think having this test, would reveal that checking config.MFA_ENABLED isn't sufficient. A good test should fail before the actual code change, and then pass with the code change (well that only relevant if you test with django admin, as the function you are making won't have existed prior)

hub/admin/extend_user.py
```diff
@@ -310,3 +310,14 @@ def _get_message(self, singular: bool, grace_period: int) -> str:
             message += f'View <a href="{url}">trash.</a>'
 
         return mark_safe(message)
+
+    def save_model(self, request, obj, form, change):
+        if obj.is_superuser:
+            if config.SUPERUSER_AUTH_ENFORCEMENT:
```
bufke, line 316: No need to nest

`if obj.is_superuser and config.SUPERUSER_AUTH_ENFORCEMENT and ...`

hub/admin/extend_user.py
```diff
@@ -310,3 +310,14 @@ def _get_message(self, singular: bool, grace_period: int) -> str:
...
+                if obj.has_usable_password() and not config.MFA_ENABLED:
+                    raise ValidationError(
+                        (
+                            'Superusers must use either MFA or SSO with an unusable password.'
```
bufke, line 320: Could we simplify this to say something like 'Superusers with a usable password must enable MFA' but maybe ask around. That term makes a lot of sense to me, but maybe not the general public.

bufke (changes requested):

hub/admin/extend_user.py
```diff
@@ -53,6 +63,11 @@ def clean(self):
                 f'User is in <a href="{url}">trash</a> and cannot be reactivated'
                 f' from here.'
             ))
+        is_superuser = cleaned_data.get('is_superuser', False)
+        if is_superuser and not validate_superuser_auth(self.instance):
```
bufke, line 67: Redundant  is_superuser

bufke (approved):
Looks good. This is ready to merge pending my one minor request.

PR #4972, opened 2024-06-07
Add "is_public" boolean to SocialAppCustomData 

RuthShryock:
## Checklist

3. [X] Ensure the tests pass

## Description

Allow display of the social login to be customizable when a SocialAppCustomData has been added. 

## Notes

`is_public` is default set to `False` which maintains the current behavior of hiding the social login. When set to `True`, users can view the social login. This allows us to add more settings in the future.


kpi/tests/api/test_api_environment.py
```diff
@@ -311,6 +312,11 @@ def test_social_apps(self):
             response = self.client.get(self.url, format='json')
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         app = baker.make('socialaccount.SocialApp')
+        custom_data = SocialAppCustomData.objects.create(
```
bufke, line 315: We need a test that ensure when there is no SocialAppCustomData, that it uses the current behavior - which is to show the login. If we didn't do that, then after this migration runs all social login buttons that aren't configured to have a SocialAppCustomData would just disappear. Or one could simplify that as to say - **all** social logins would disappear. That would be a breaking change that we should avoid.

bufke (commented):
LGTM, thank you!

PR #4975, opened 2024-06-10
Add `provision_server` management command

RuthShryock:
## Checklist

3. [X] Ensure the tests pass

## Description

Automates setup of social applications and constance configurations through running `./manage.py provision_server` along with specified arguments. 

## Notes

This command has two subcommands `socialapp` and `config`. 
Example: `./manage.py provision_server config MFA_ENABLED=False SUPERUSER_AUTH_ENFORCEMENT=True` will disable MFA and enable authentication for superusers. 
(The `config` subcommand can take an infinite number of key-value pairs.)


kpi/management/commands/provision_server.py
```diff
@@ -0,0 +1,184 @@
...
+                    else:
+                        # Specific handling for fields taking JSON arrays
+                        if key in [
+                            'PROJECT_METADATA_FIELDS',
+                            'USER_METADATA_FIELDS',
+                        ]:
+                            if value.startswith('[') and value.endswith(']'):
+                                try:
+                                    value = json.loads(value)
+                                    value = json.dumps(value)
```
bufke, line 152: Interesting way to validate json, I imagine it works fine. I don't think it's possible to error in the dumps line, so that can be moved outside of the try, except.

Or actually I think you could achieve the validation with 

```
try:
   json.loads(value)
except json.JSONDecodeError ...
```
you don't really need the value, right? You're just converting to and then from json to validate it I assume.

If you end up keeping this code - I might try to break it out so it's not so nested. I don't typically say one must "Don't repeat yourself" when repeating just twice. But in this case it might help reduce the nesting.

kpi/management/commands/provision_server.py
```diff
@@ -0,0 +1,184 @@
...
+                        value = True
+                    elif value.lower() == 'false':
+                        value = False
+                    else:
+                        # Specific handling for fields taking JSON arrays
+                        if key in [
+                            'PROJECT_METADATA_FIELDS',
+                            'USER_METADATA_FIELDS',
+                        ]:
+                            if value.startswith('[') and value.endswith(']'):
```
bufke, line 149: Is this needed? I know that json technically isn't supposed to start with an array. But this works for me

```python
>>> data = "[1, 2]"
>>> import json
>>> json.loads(data)
[1, 2]
>>> json.dumps(json.loads(data))
'[1, 2]'
```

bufke (changes requested):
I would expect this to work but it doesn't `/manage.py provision_server socialapp --provider goog --provider_id g --name name --client_id abc --settings "{}"` is it easy to allow keyword arguments? It would make it easier to omit optional values.

kpi/management/commands/provision_server.py
```diff
@@ -0,0 +1,186 @@
...
+                            value.startswith('{') and value.endswith('}')
+                        ):
+                            try:
+                                json.loads(value)
+                            except json.JSONDecodeError as e:
+                                self.stdout.write(
+                                    f'Invalid JSON value for key {key}. {e}'
+                                )
+                                continue
+                            json.dumps(value)
```
bufke, line 176: I don't think this does anything, does it? It serializes to a string but the result isn't saved anywhere.

kpi/management/commands/provision_server.py
```diff
@@ -0,0 +1,186 @@
...
+            self.handle_config(kwargs)
+        else:
+            self.stdout.write(self.style.ERROR('No valid subcommand provided'))
+
+    def handle_socialapp(self, kwargs):
+        provider = kwargs['provider']
+        provider_id = kwargs['provider_id']
+        name = kwargs['name']
+        client_id = kwargs['client_id']
+        secret = os.getenv('SOCIAL_APP_SECRET') or kwargs.get('secret', '')
```
bufke, line 107: Very minor but if you could do this

`kwargs.get('secret', os.getenv('SOCIAL_APP_SECRET', ''))`

That would prefer the kwarg over the env var (Seems right to me) :shrug:

kpi/management/commands/provision_server.py
```diff
@@ -0,0 +1,186 @@
...
+        key = kwargs.get('key', '')
+        settings_json = kwargs['settings']
+
+        try:
+            settings = json.loads(settings_json)
+        except TypeError:
+            raise json.JSONDecodeError
```
bufke, line 114: I think this is catching the same error and then re-raising it? Why not allow the original exception to be raised? If you want to catch it and add more context, do that. Or leave it alone. Catching and raising without any benefit is  obfuscating the error.

kpi/management/commands/provision_server.py
```diff
@@ -0,0 +1,186 @@
...
+
+        social_app_data = {
+            'provider': provider,
+            'provider_id': provider_id,
+            'name': name,
```
bufke, line 119: This line can be omitted, you don't need to repeat the args in get_or_create again in defaults. The usage of get_or_create makes a lot of sense though.

bufke (changes requested):

kpi/management/commands/provision_server.py
```diff
@@ -103,7 +103,7 @@ def handle_socialapp(self, kwargs):
         provider_id = kwargs['provider_id']
         name = kwargs['name']
         client_id = kwargs['client_id']
-        secret = kwargs.get('secret', os.getenv('SOCIAL_APP_SECRET', ''))
+        secret = kwargs.get('secret', '') or os.getenv('SOCIAL_APP_SECRET')
```
bufke, line 106: I know I'm harping on this but it could be more clear. 

The second arg in the get function is the default. Which defaults to `''` which it then considered "falsey" and so the os.getenv('SOCIAL_APP_SECRET') gets evaluated.

It can be simplified as ` kwargs.get('secret', os.getenv('SOCIAL_APP_SECRET'))`.

kpi/management/commands/provision_server.py
```diff
@@ -0,0 +1,183 @@
...
+        name = kwargs['name']
+        client_id = kwargs['client_id']
+        secret = kwargs.get('secret', '') or os.getenv('SOCIAL_APP_SECRET')
+        key = kwargs.get('key', '')
+        settings_json = kwargs.get('server_settings', '{}')
+
+        try:
+            settings = json.loads(settings_json)
+        except json.JSONDecodeError as e:
+            self.stdout.write(f"Invalid JSON for settings: {e}")
```
bufke, line 113: :+1:

bufke (changes requested):
Thanks, I'm pretty sure I can just merge after this one change.

bufke (approved):

PR #4986, opened 2024-06-26
Add django-import-export for organization users

bufke:
## Description

Django admin can now import and export organization users to and from formats including csv.

## Notes

It's a common need to import and export users. Django-import-export provides a customizable way to do so with minimal code. It's IMO a well maintained package but does occasionally include minor breaking changes, such as the 4.x release.

This change took about 45 minutes and was tested with importing Organization Users.

### DIFF
--- a/dependencies/pip/dev_requirements.txt
+++ b/dependencies/pip/dev_requirements.txt
@@ -184,6 +184,8 @@ django-extensions==3.2.3
     #   django-organizations
 django-filter==24.1
     # via -r dependencies/pip/requirements.in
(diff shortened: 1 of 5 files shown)


jnm (changes requested):
The changes look fine, but the PR description does not reflect well what the code actually does. See my comment on https://www.notion.so/kobotoolbox/Import-billing-users-in-bulk-c389e89af9b94606af6fef069facf850.

I have _not_ run the branch and tested the new functionality.

jnm (approved):
Thanks

PR #4989, opened 2024-06-28
Reject 2FA activation with API when option is disabled for that account #4556

### LINKED ISSUE #4556: Reject 2FA activation with API when option is disabled for that account
## Description
UI does not display 2FA section when option is disabled, thus back end should block any requests to 2FA endpoints when the option is disabled. 
Moreover, Token form is already disabled in back end in such situation.

Guitlle:
## Description

When the MFA activation API endpoint is accessed, it fails with a forbidden error if the user is not in the MfaAvailableToUser table (#4556). We had to include a check for user subscription and STRIPE_ENABLED.

## Notes

Trench provides a set of views for the MFA API. The view `trench.views.MFAMethodActivationView` was extended to check, before accepting a call to the activation endpoint, whether or not the user can use MFA by checking the model `kobo.apps.accounts.mfa.models.MfaAvailableToUser`.   The module `kobo.apps.accounts.adapter` was also changed to use the more centralized function at `kobo.apps.accounts.mfa.permissions.mfa_allowed_for_user`. 

## Related issues

Fixes #4556

### DIFF
--- a/kobo/apps/accounts/adapter.py
+++ b/kobo/apps/accounts/adapter.py
@@ -9,6 +9,7 @@ from django.template.response import TemplateResponse
 from django.utils import timezone
 from trench.utils import get_mfa_model, user_token_generator
 
+from .mfa.permissions import mfa_allowed_for_user
 from .mfa.forms import MfaTokenForm
 from .mfa.models import MfaAvailableToUser
 from .mfa.views import MfaTokenView
@@ -25,16 +26,9 @@ class AccountAdapter(DefaultAccountAdapter):
             # validated their email address
             return parent_response
 
-        mfa_allowed = True
-        if settings.STRIPE_ENABLED:
-            mfa_allowed = (
-                user_has_paid_subscription(user.username)
-                or MfaAvailableToUser.objects.filter(user=user).exists()
-            )
-
         # If MFA is activated and allowed for the user, display the token form before letting them in
         if (
-            mfa_allowed
+            mfa_allowed_for_user(user)
             and get_mfa_model()
             .objects.filter(is_active=True, user=user)
             .exists()
--- a/kobo/apps/accounts/adapter.py
+++ b/kobo/apps/accounts/adapter.py
@@ -13,7 +13,7 @@ from .mfa.forms import MfaTokenForm
 from .mfa.models import MfaAvailableToUser
 from .mfa.permissions import mfa_allowed_for_user
 from .mfa.views import MfaTokenView
-from .utils import user_has_paid_subscription
+from .utils import user_has_inactive_paid_subscription
 
 
 class AccountAdapter(DefaultAccountAdapter):
@@ -27,12 +27,14 @@ class AccountAdapter(DefaultAccountAdapter):
             return parent_response
 
         # If MFA is activated and allowed for the user, display the token form before letting them in
-        if (
-            mfa_allowed_for_user(user)
-            and get_mfa_model()
+        mfa_active = (
+            get_mfa_model()
             .objects.filter(is_active=True, user=user)
             .exists()
-        ):
+        )
+        mfa_allowed = mfa_allowed_for_user(user)
+        inactive_subscription = user_has_inactive_paid_subscription(user.username)
+        if  mfa_active and (mfa_allowed or inactive_subscription):
             ephemeral_token_cache = user_token_generator.make_token(user)
             mfa_token_form = MfaTokenForm(
                 initial={'ephemeral_token': ephemeral_token_cache}
--- a/kobo/apps/accounts/adapter.py
+++ b/kobo/apps/accounts/adapter.py
@@ -28,21 +28,20 @@ class AccountAdapter(DefaultAccountAdapter):
 
         # If MFA is activated and allowed for the user, display the token form before letting them in
         mfa_active = (
-            get_mfa_model()
-            .objects.filter(is_active=True, user=user)
-            .exists()
+            get_mfa_model().objects.filter(is_active=True, user=user).exists()
         )
         mfa_allowed = mfa_allowed_for_user(user)
-        inactive_subscription = user_has_inactive_paid_subscription(user.username)
-        if  mfa_active and (mfa_allowed or inactive_subscription):
+        inactive_subscription = user_has_inactive_paid_subscription(
+            user.username
+        )
+        if mfa_active and (mfa_allowed or inactive_subscription):
             ephemeral_token_cache = user_token_generator.make_token(user)
             mfa_token_form = MfaTokenForm(
                 initial={'ephemeral_token': ephemeral_token_cache}
             )
 
-            next_url = (
-                kwargs.get('redirect_url')
-                or resolve_url(settings.LOGIN_REDIRECT_URL)
+            next_url = kwargs.get('redirect_url') or resolve_url(
+                settings.LOGIN_REDIRECT_URL
             )
 
             context = {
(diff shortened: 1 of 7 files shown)


kobo/apps/accounts/mfa/tests/test_api.py
```diff
@@ -67,3 +71,21 @@ def test_mfa_activation_always_creates_new_secret(self):
...
+
+    def test_mfa_disabled(self):
+        anotheruser = User.objects.get(username='anotheruser')
+        self.client.login(username='anotheruser', password='anotheruser')
+        method = list(trench_settings.MFA_METHODS.keys())[0]
+
+        activate_response = self.client.post(
+            reverse('mfa-activate', args=(method,))
+        )
+        assert activate_response.status_code == 403
```
noliveleger, line 84: Let's use DRF constants instead, i.e. `status.HTTP_403_FORBIDDEN'. 
Please, change others accordingly.

kobo/apps/accounts/mfa/urls.py
```diff
@@ -1,7 +1,7 @@
 # coding: utf-8
 from django.urls import include, path
 
-from .views import MfaListUserMethodsView, MfaLoginView, MfaTokenView
+from .views import MfaListUserMethodsView, MfaLoginView, MfaTokenView, MfaMethodActivationView
```
noliveleger, line 4: Please (black) reformat, we are trying to respect the 80 characters limit (with up to 90 if variable names are way too long).

```python 
from .views import (
    MfaMethodActivationView,
    MfaListUserMethodsView,
    MfaLoginView,
    MfaTokenView,
)
```

kobo/apps/accounts/mfa/views.py
```diff
@@ -66,3 +70,15 @@ class MfaListUserMethodsView(ListAPIView):
...
         mfa_model = get_mfa_model()
         return mfa_model.objects.filter(user_id=self.request.user.id)
+
+
+class IsMfaEnabled(BasePermission):
+    def has_permission(self, request, view):
+        mfa_available = MfaAvailableToUser.objects.filter(
+            user=get_database_user(request.user)
+        ).exists()
+        return mfa_available
```
noliveleger, line 80: Let's try to be consistent with main app (`kpi`). Please the code in its own file `permissions.py` at the sub-app level.

kobo/apps/accounts/mfa/views.py
```diff
@@ -66,3 +70,15 @@ class MfaListUserMethodsView(ListAPIView):
...
+
+
+class MfaMethodActivationView(TrenchMFAMethodActivationView):
+    permission_classes = (IsAuthenticated, IsMfaEnabled)
```
noliveleger, line 84: ❤️

kobo/apps/accounts/mfa/tests/test_api.py
noliveleger: BTW, I'm wondering if the permission class should return a 400 instead of 403? I think it sounds more like a bad request than a forbidden access? 
What do you think?

noliveleger (changes requested):
The PR is great. I love your idea to add the permission class to avoid overloading trench classes. 
I'm only asking for "cosmetic" changes.

kobo/apps/accounts/mfa/tests/test_api.py
Guitlle: I think it's both. To me it looks like if you are not in the MFA available  to user table, then you don't have the permission to use MFA. I agree that 400 makes sense too, but I'm more inclined to the 403, as in: you are not allowed to activate MFA.

kobo/apps/accounts/mfa/urls.py
Guitlle: Ok, thanks, I'll set up the linters to avoid these,

kobo/apps/accounts/mfa/views.py
Guitlle: ok,

kobo/apps/accounts/mfa/tests/test_api.py
noliveleger: So let's keep the 403, then.

kobo/apps/accounts/adapter.py
```diff
@@ -9,6 +9,7 @@
 from django.utils import timezone
 from trench.utils import get_mfa_model, user_token_generator
 
+from .mfa.permissions import mfa_allowed_for_user
```
noliveleger, line 12: Please keep sorting alphabetically 🙏

kobo/apps/accounts/mfa/tests/test_api.py
```diff
@@ -45,8 +47,12 @@ def test_user_methods_with_date(self):
         for field in expected_fields:
             self.assertTrue(field in results[0])
 
+    @override_config(MFA_ENABLED=True)
     def test_mfa_activation_always_creates_new_secret(self):
         self.client.login(username='anotheruser', password='anotheruser')
+        anotheruser = User.objects.get(username='anotheruser')
+
+        mfa_availability = MfaAvailableToUser.objects.create(user=anotheruser)
```
noliveleger, line 55: As per discussion, I think this is not needed if `MfaAvailableToUser` list is empty. 
You can revert this change if it is case.

kobo/apps/accounts/mfa/tests/test_api.py
```diff
@@ -67,3 +73,36 @@ def test_mfa_activation_always_creates_new_secret(self):
...
+        activate_response = self.client.post(
+            reverse('mfa-activate', args=(method,))
+        )
+        assert activate_response.status_code == status.HTTP_200_OK
+
+        # Reset MFA whitelist state
+        mfa_availability.delete()
+        someuser_mfa_activation.delete()
+
+        # Test when whitelist is disabled
```
noliveleger, line 104: Would it be better to test if the whitelist is really empty before testing the activation?

kobo/apps/accounts/mfa/tests/test_api.py
```diff
@@ -67,3 +73,36 @@ def test_mfa_activation_always_creates_new_secret(self):
             )
             assert first_secret != second_secret
             assert first_response.json() != second_response.json()
+        mfa_availability.delete()
+
+    @override_config(MFA_ENABLED=True)
+    def test_mfa_disabled(self):
```
noliveleger, line 79: I think we should find a better name for this test. Having `@override_config(MFA_ENABLED=True)` just above `def test_mfa_disabled(self):` is confusing. 
What about something like `test_mfa_whitelisting`

noliveleger (changes requested):

kobo/apps/accounts/mfa/tests/test_api.py
Guitlle: Ok

kobo/apps/accounts/mfa/tests/test_api.py
Guitlle: Right, I removed it

kobo/apps/accounts/mfa/tests/test_api.py
Guitlle: Yes, moved it

kobo/apps/accounts/utils.py
```diff
@@ -4,12 +4,21 @@
...
 
 def user_has_paid_subscription(username):
+    return User.objects.filter(
+        username=username,
+        organizations_organization__djstripe_customers__subscriptions__status__in=ACTIVE_STRIPE_STATUSES,
+        organizations_organization__djstripe_customers__subscriptions__items__price__unit_amount__gt=0,
+    ).exists()
+
+
+def user_has_inactive_paid_subscription(username):
```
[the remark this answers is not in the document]
Guitlle: Ok

kobo/apps/stripe/tests/test_mfa_login.py
```diff
@@ -67,49 +63,130 @@ def _create_subscription(self, unit_amount=0, billing_status='active'):
...
+        )
+        self._assert_mfa_login(response)
+
+    @override_config(MFA_ENABLED=True)
+    def test_mfa_login_with_free_subscription_no_whitelist(self):
+        """
+        Validate MFA form is displayed after login when the user
+        has free subscription and whitelist is disabled
+        """
+        self._reset_whitelist([])
```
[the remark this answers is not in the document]
Guitlle: Right!

PR #5008, opened 2024-07-11
Fix project view table timeout

rgraber:
## Checklist

1. [ ] If you've added code that should be tested, add tests N/A
2. [ ] If you've changed APIs, update (or create!) the documentation N/A
3. [x] Ensure the tests pass
6. [ ] Mention any related issues in this repository (as #ISSUE) and in other repositories (as kobotoolbox/other#ISSUE) N/A
7. [ ] Open an issue in the [docs](https://github.com/kobotoolbox/docs/issues/new) if there are UI/UX changes N/A

## Description

Fixes a timeout that was occurring when trying to view the table for a project view that was set to look at all countries. 

## Notes

Updates the project_view/UID/assets endpoint to use a new FastAssetPagination class for pagination. The new class selects only primary keys for counting instead of every field. 
There was an existing class (AssetPagination) for querysets against the asset table but the new class is several times faster. The chart below shows the distribution of load times when cURLing the endpoint 100 times using each class. We should consider using the new class in more places, but those would be separate PRs.

Screenshot from 2024-07-12 09-32-50


kobo/apps/project_views/views.py
```diff
@@ -56,6 +57,7 @@ def get_queryset(self, *args, **kwargs):
     def assets(self, request, uid):
         if not user_has_view_perms(request.user, uid):
             raise Http404
+        self._paginator = FastAssetPagination()
```
noliveleger, line 60: I think we can specify the paginator class in the arguments passed to `@action` decorator. 


```python 
    @action(
        detail=True,
        methods=['GET'],
        filter_backends=[SearchFilter, AssetOrderingFilter],
        pagination_class=FastAssetPagination
    )
    def assets(self, request, uid):
        if not user_has_view_perms(request.user, uid):
```

kpi/paginators.py
```diff
@@ -31,6 +31,19 @@ def get_parent_url(self, obj):
         return reverse_lazy('api-root', request=self.context.get('request'))
 
 
+class FastAssetPagination(Paginated):
```
noliveleger, line 34: I would have given a name related to ProjectView (e.g. `ProjectViewAssetPagination` 😅 ) but if you think it can be reused somewhere else and you want to keep it as-is, please move it below `AssetPagination` to keep alphabetically sort 🙏 .

kpi/paginators.py
```diff
@@ -31,6 +31,19 @@ def get_parent_url(self, obj):
...
+    """
+    Pagination class optimized for faster counting for DISTINCT queries on large tables.
+
+    This class overrides the get_count() method to only look at the primary key field, avoiding expensive DISTINCTs
+    comparing several fields. This may not work for queries with lots of joins, especially with one-to-many or
+    many-to-many type relationships.
+    """
+
+    def get_count(self, queryset):
+        return queryset.only('pk').count()
```
noliveleger, line 44: I thought about that and maybe we want to optimize the count only when a distinct is present? 
Something like that. 

```python 
    def get_count(self, queryset):
        if queryset.query.distinct:
            return queryset.only('pk').count()
        return super().get_count()    
```

What do you think?

noliveleger: For the record, I wanted to test if we get some better performance with `values`, but it is uses with `.count()`.  It does not change anything. 

```python 
In [2]: import timeit

In [3]: timeit.timeit('Asset.objects.filter(owner__username="myusername").distinct().only("pk").count()', globals=globals(), number=10000)
Out[3]: 6.9069382949965075

In [4]: timeit.timeit('Asset.objects.filter(owner__username="myusername").distinct().values("pk").count()', globals=globals(), number=10000)
Out[4]: 7.383306294999784

In [5]: timeit.timeit('Asset.objects.filter(owner__username="myusername").distinct().values("pk").count()', globals=globals(), number=10000)
Out[5]: 6.69678862800356

In [6]: timeit.timeit('Asset.objects.filter(owner__username="myusername").distinct().only("pk").count()', globals=globals(), number=10000)
Out[6]: 7.368681169988122
```

noliveleger (changes requested):

kpi/paginators.py
rgraber: > I thought about that and maybe we want to optimize the count only when a distinct is present? Something like that.
> 
> ```python
>     def get_count(self, queryset):
>         if queryset.query.distinct:
>             return queryset.only('pk').count()
>         return super().get_count()    
> ```
> 
> What do you think?

I tested this locally by removing and re-adding the 'distinct' clause on the filter and it seemed to work

kpi/paginators.py
rgraber: I'm hoping it's reusable, so I'll keep the name and just move it. BTW the rest of the file is definitely not in alphabetical order

noliveleger (approved):

PR #5020, opened 2024-07-18
[TASK-823] Create an abstract model which provides `date_created` and `date_modified` fields

rajpatel24:
## Description
- Introduce an abstract timestamped model which provides date_created and date_modified fields.
- Refactor existing models to inherit from the abstract model, replacing the individual date_created and date_modified fields.

## Notes
The date_created and date_modified fields are common across several models in the project. To streamline their implementation, I created an abstract model similar to TimeStampedModel from django-extensions but without using auto_add_now=True, which does not fit our requirements. It reduces redundancy in the codebase by centralizing the timestamp logic in one abstract model.


kobo/apps/openrosa/apps/logger/models/instance.py
```diff
@@ -1,5 +1,8 @@
 # coding: utf-8
 from hashlib import sha256
+
+from kpi.models.abstract_models import AbstractTimeStampedModel
+
```
noliveleger, line 5: This import should be in the 3rd block, ordered alphabetically.

kobo/apps/openrosa/libs/data/tests/test_tools.py
```diff
@@ -38,7 +38,7 @@ def test_get_form_submissions_grouped_by_field(self, mock_time):
     @patch('kobo.apps.openrosa.apps.logger.models.instance.submission_time')
     def test_get_form_submissions_grouped_by_field_datetime_to_date(
             self, mock_time):
-        now = datetime(2014, 1, 1)
+        now = datetime.now()
```
noliveleger, line 41: Why did you change the date? I think you should keep the hard-coded value.

kobo/apps/trash_bin/models/__init__.py
```diff
@@ -3,6 +3,8 @@
```
noliveleger, line 3: Please remove the extra blank line `from django.utils.timezone import now` should be part of the same block of other django imports.

kpi/models/asset.py
```diff
@@ -157,10 +158,8 @@ class Asset(ObjectPermissionMixin,
             XlsExportableMixin,
             FormpackXLSFormUtilsMixin,
             StandardizeSearchableFieldMixin,
-            models.Model):
```
noliveleger, line 160: Let's format it with black ;-)

noliveleger (changes requested):

kpi/models/abstract_models.py
```diff
@@ -0,0 +1,27 @@
+from django.db import models
+from django.utils import timezone
+
+
+def get_current_time():
+    """
+    Return the current time so that the timestamp is evaluated at the time of model instantiation rather than
+    at the time of class definition
+    """
+    return timezone.now()
```
noliveleger, line 10: I would change the name to something like `_get_default_datetime`. 

Notes: The underscore is to indicate to other devs that it should not be outside this file. 

Try to keep the comment under the 80 characters per line and explain why we need to use this function instead of using `timezone.now` directly.

noliveleger (changes requested):

kpi/models/asset_export_settings.py
```diff
@@ -3,22 +3,21 @@
...
     uid = KpiUidField(uid_prefix='es')
     asset = models.ForeignKey('Asset', related_name='asset_export_settings',
                               on_delete=models.CASCADE)
-    date_modified = models.DateTimeField()
     name = models.CharField(max_length=255, blank=True, default='')
     export_settings = models.JSONField(default=dict)
 
     def save(self, *args, **kwargs):
-        self.date_modified = timezone.now()
         super().save(*args, **kwargs)
```
noliveleger, line 21: It can be removed entirely since it does nothing anymore.

kpi/models/asset_user_partial_permission.py
```diff
@@ -115,13 +116,8 @@ class Meta:
...
     )
     permissions = models.JSONField(default=dict)
-    date_created = models.DateTimeField(default=timezone.now)
-    date_modified = models.DateTimeField(default=timezone.now)
 
     def save(self, *args, **kwargs):
-        if self.pk is not None:
-            self.date_modified = timezone.now()
-
         super().save(*args, **kwargs)
```
noliveleger, line 121: It can be removed.

kpi/models/asset_version.py
```diff
@@ -78,5 +78,4 @@ def __str__(self):
             ' (deployed)' if self.deployed else '')
 
     def save(self, *args, **kwargs):
-        self.date_modified = timezone.now()
         super().save(*args, **kwargs)
```
noliveleger, line 81: It can be removed

noliveleger (changes requested):

kobo/apps/openrosa/libs/data/tests/test_tools.py
rajpatel24: I changed it to datetime.now() because we now store the current timestamp immediately upon the model being instantiated using an abstract model. Previously, date_created was not automatically being stored; instead, the value of submission_time was being stored in the date_created field. However, I have updated the test case by adding a mock datetime for date_created and by keeping datetime(2014, 1, 1).

kobo/apps/openrosa/apps/logger/models/instance.py
```diff
@@ -1,8 +1,6 @@
 # coding: utf-8
 from hashlib import sha256
 
-from kpi.models.abstract_models import AbstractTimeStampedModel
-
```
noliveleger: Remove extra lines, `zoneInfo` should be part of the first block of imports.

noliveleger (approved):
I won't merge it now. 
It will go in `beta` as soon as `django-kobocat-app` goes in the release branch.

PR #5021, opened 2024-07-18
Replace default values with (invalid) placeholders

### LINKED ISSUE #4364: Remove invalid default values that are not obvious placeholders
Something like `https://enketo.org` *looks* plausible as a setting for `ENKETO_URL`, but it could never work. We should remove all such values from this repository (and kobocat as well; feel free to simply open a kobocat issue if you're the person undertaking the kpi work).

https://github.com/kobotoolbox/kpi/blob/ee6d0ba6918e48e106bd894b34d594b7e4d94f92/kobo/settings/base.py#L600

noliveleger:
## Description

Avoid starting the app with wrong default and make admins realize the settings are wrong right away.

## Notes

Left alone `STRIPE_TEST_PUBLIC_KEY` and  `STRIPE_LIVE_PUBLIC_KEY` but IMO, we should be consistent and use env variables to pass their value and not hard-code their value.

## Related issues

Fixes #4364


RuthShryock (approved):
Looks good!

PR #5030, opened 2024-07-24
Fix bug affecting project deletion permissions in the project list table

RuthShryock:
## Checklist

3. [X] Ensure the tests pass

## Description

Allow only owners to be able to delete a shared project.

### DIFF
--- a/jsapp/js/projects/projectsTable/projectQuickActions.tsx
+++ b/jsapp/js/projects/projectsTable/projectQuickActions.tsx
@@ -105,7 +105,7 @@ export default function ProjectQuickActions(props: ProjectQuickActionsProps) {
 
       {/* Delete */}
       <Button
-        isDisabled={!isChangingPossible}
+        isDisabled={!isManagingPossible}
         type='bare'
         color='storm'
         size='s'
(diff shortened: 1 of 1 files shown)


jsapp/js/projects/projectsTable/projectQuickActions.tsx
```diff
@@ -120,7 +120,7 @@ export default function ProjectQuickActions(props: ProjectQuickActionsProps) {
           )
         }
         tooltip={
-          isChangingPossible ? t('Delete 1 project') : t('Delete project')
+          isManagingPossible ? t('Delete 1 project') : t('Delete project')
```
magicznyleszek, line 123: I think we could also update `projectSettings.es6` file (around line 1062). There is a delete project button there that currently is only available to the owner of the project, and I think it should be possible to use it as a `manage_asset` user too :)

magicznyleszek: Also I think there is a mistake in `assetActionButtons.tsx` (different file, similar name to this one). We (most probably I :wink:) use `userCan('delete_submissions'…` in a places that most probably (you need to verify it) should use `manage_asset`.

magicznyleszek (changes requested):

jsapp/js/projects/projectsTable/projectQuickActions.tsx
RuthShryock: I tested both `projectSettings.es6` and `assetActionButtons.tsx` and changed the permissions `manage_asset` but get the error from the backend: `{"detail":"You do not have permission to perform this action."}`. So I think they are both correct in those places.

magicznyleszek (approved):

PR #5032, opened 2024-07-25
Repeated group version fix

jamesrkiger:
## Checklist

3. [x] Ensure the tests pass

## Description

The immediate bug this PR is aiming to fix is that some nested repeated group objects were being passed as props to the TextModalCell component, resulting in an error that breaks the entire data table. This bug can be triggered by creating a form with repeated group/question `foo/bar`, making a submission, then renaming the group and question to `bar/foo`.

In the process of investigating this bug, I realized that repeated groups were broken in the data table even for responses on the current version of the form. They didn't usually throw errors, but they also didn't display because the wrong key was being used to grab them from the SurveyResponse. This PR fixes that underlying issue with repeated groups so they now render properly in the data table. It also applies a stopgap fix for the original issue by simply not attempting to render objects. 

## Notes

The solution here ultimately means repeated group answers from older versions of a form are not displayed at all on the data table, even though a column will still be displayed for the older group/question. So in the above example, there will be columns `bar/foo` (displaying any responses to the new version of the group/question) and `foo` (with no data displayed). This is an improvement over breaking the data table, but obviously not ideal. 

We *might* be able to get this data to display by recursively iterating over every submission response to generate the necessary flatpaths. But I was undecided on whether this is a good idea and, if so, whether it's something I should work on now or if we should leave it as a TODO, given that John has mentioned that we might be trying to integrate better version handling in the not-so-distant future. I would appreciate input on this question.

### DIFF
--- a/jsapp/js/components/submissions/submissionUtils.ts
+++ b/jsapp/js/components/submissions/submissionUtils.ts
@@ -554,7 +554,7 @@ export function getRepeatGroupAnswers(
 
     // Each level could be an array of repeat group answers or object with questions.
     if (levelKey === targetKey) {
-      if (targetKeyData !== undefined) {
+      if (targetKeyData !== undefined && typeof targetKeyData !== 'object') {
         answers.push(String(targetKeyData));
       }
     } else if (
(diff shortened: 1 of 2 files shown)


magicznyleszek (approved):
I don't think we should be doing any more work here at this moment. The case you described with going through all submissions is not worth it. Let's wait for better version handling :)

PR #5033, opened 2024-07-29
Update colors

magicznyleszek:
## Checklist

3. [x] Ensure the tests pass

## Description

We've updated most of the colors in our palette to more accessible shades, and added few new colors in preparation for upcoming improvements.

## Notes

Most of the colors were slightly updated. We had a major change in reds:
- `kobo-red` was renamed to `kobo-mid-red`
- `kobo-dark-red` was renamed to `kobo-red` (this is the part that might be confusing)
- we have new `kobo-dark-red` that is slightly darker

Another big change was with grays:
- All `kobo-gray-<lightness number>` were renamed to `kobo-neutral-<counter>`, e.g. `kobo-gray-98` is now `kobo-neutral-100`
- We use `-100`, `-200`, `-300`, …, `-900` as neutral color "types". This allows us to have flexibility of adding `-250` in future if needed
- The shades of neutrals were updated too

Other changes here:
- Migrated Button `dark-red` to `red` (this is kinda undoing recent change, but previous change #4993 was changing the color to darker shade, this one keeps the color, but the name is different)
- Migrated Icon component from `red` to `mid-red`
- Removed duplicated lines of code from `react-tagsinput.scss` file (I'm pretty sure I already did that)
- Updated the reds being used in various places to ones that make sense. Some were left as `kobo-red` (so will be darker than previously), and some were updated to be `kobo-mid-red` (so will appear as if nothing changed)

## Related issues

Followup to #4993
Made https://github.com/kobotoolbox/docs/issues/372


jsapp/scss/colors.scss
```diff
@@ -1,82 +1,82 @@
 // Kobo global palette of colors.
 
 /*
- * The Kobo "gray" scale
- *
- * NOTE: the list consist of colors with same hue (225) and saturation (16).
+ * The Kobo neutral scale
```
duvld, line 4: Just wondering, why did we change the name from gray to neutral? Gray to me is more intuitive as to what the color is trying to tell me

jsapp/scss/colors.scss
```diff
@@ -1,82 +1,82 @@
...
+$kobo-white: hsl(0deg 0% 100%); // #ffffff
 
 /*
  * Primary colours & shades
  */
 
-$kobo-dark-blue: hsl(206deg 100% 31%); // #00599d
-$kobo-hover-blue: hsl(213deg 81% 59%); // #408deb
+$kobo-dark-blue: hsl(207deg 77% 32%); // #135991
+$kobo-hover-blue: hsl(207deg 77% 43%); // #1977c2
```
duvld, line 45: Nitpick but I guess I never noticed "hover-someColor" as a color before, but the naming is a bit vague. Should we only use hover-x in conjunction with the mid-x colors? Does this mean we assume we will never have something that is lighter than mid-x use a hover color?

duvld (approved):
I did a quick look around the UI and it looks like your find and replace got all of them :) if someone spots a missed spot we can probably do individual touch ups

jsapp/scss/colors.scss
magicznyleszek: After Ruth and James voted they prefer "gray" too, I undid the change :)

jsapp/scss/colors.scss
magicznyleszek: I think that these `hover` colors are part of the palette for the simplicity case. We might decide in future that we would rather have some rule like "`hover` color is 10% darker than the idle color". I agree the name is not perfect, we might decide to change it too in future, if Tessa starts using it for other things :)

PR #5042, opened 2024-07-31
Code refactoring - Remove shadow models and KobocatDeploymentBackend class

noliveleger:
## Description

Part III of https://github.com/kobotoolbox/kpi/pull/5036. 

## Notes

With the new deployment backend class `OpenRosaDeploymentBackend` which can handle older projects too,  shadow models and `KobocatDeploymentBackend` class  become useless.

## Related issues

Part of #5036


kpi/management/commands/sync_kobocat_xforms.py
```diff
@@ -177,6 +177,7 @@ def _xform_to_asset_content(xform):
 
 
 def _get_kc_backend_response(xform):
+    # FIXME wrong backend info
```
jnm, line 180: :eyes:

jnm (commented):
My main concern is that having to remember `using()` every time `openrosa` models are accessed is more error-prone than the old KoboCAT `ShadowModel` approach, where the database routing was handled automatically.

What happens if someone forgets `using()`? Does their code blow up immediately because the KPI database does not have `openrosa` tables? That would probably be good enough. If not, we might want to consider some kind of safeguard (at the database router level?) to wail and moan if someone attempts to access an `openrosa` model using the KPI database.

(I realize we'll be getting rid of the separate database situation before too long, but we don't have a definite timeline yet)

noliveleger:
> My main concern is that having to remember `using()` every time `openrosa` models are accessed is more error-prone than the old KoboCAT `ShadowModel` approach, where the database routing was handled automatically.
> 
> What happens if someone forgets `using()`? Does their code blow up immediately because the KPI database does not have `openrosa` tables? That would probably be good enough. If not, we might want to consider some kind of safeguard (at the database router level?) to wail and moan if someone attempts to access an `openrosa` model using the KPI database.
> 
> (I realize we'll be getting rid of the separate database situation before too long, but we don't have a definite timeline yet)


AFAIK, every openrosa model is handled by the router still? [1](https://github.com/kobotoolbox/kpi/blob/b4aba7b7be182b6fdd8e6ad61be7e8b1d1ebd1d5/kpi/db_routers.py#L36), [2](https://github.com/kobotoolbox/kpi/blob/b4aba7b7be182b6fdd8e6ad61be7e8b1d1ebd1d5/kpi/db_routers.py#L20).

`using()` is only of shared model like `Permission` or `User`. Forgetting `using()` on the User table is not a big deal because it is synchronized, but on `Permission`, pk would not match and errors could happen.

jnm (approved):

PR #5046, opened 2024-08-02
[TASK-915] Simplify button component (even more)

magicznyleszek:
## Checklist

3. [x] Ensure the tests pass

## Description

Minimizes the amount of button types we use in the UI. There are still some legacy buttons being used (that don't look properly), but the vast majority of buttons will now fall into 5 distinct types. This is an improvement from current 9 types, and from previous 15 types we've had for the last few years.

Improved the accessibility of Button: pending state is no longer focusable and keyboard-clickable, disabled button is no longer TAB focusable

## Notes

Things changed here:
- Changed Button component to not use `type` + `color` mix, but rather unified and simplified `type` (we have 5 types: `primary`, `secondary`, `danger`, `secondary danger`, and `text`)
- Updated **all** instances of `<Button>` to use the new type
- Updated Button stories to have a view with all the buttons types, sizes and states being displayed at the same time

## Related issues

Includes work from #5033

After merging this https://github.com/kobotoolbox/docs/issues/372 can be done


jsapp/js/components/common/button.tsx
```diff
@@ -10,20 +10,18 @@ import cx from 'classnames';
...
+ * 3. danger - white text on red background
+ * 4. secondary danger - dark red text on light red background
+ * 5. text - dark blue text with no background
  */
-export type ButtonType = 'bare' | 'frame' | 'full';
-export type ButtonColor =
-  | 'blue'
-  | 'red'
-  | 'dark-blue';
+export type ButtonType = 'primary' | 'secondary' | 'danger' | 'secondary-danger' | 'text';
```
duvld, line 24: Tiny nitpick. Is there a way to add a jsdoc to this type such that it shows up when auto-completing? At first glance these names are a bit vague and requires the user to read this component whereas before it was a bit more descriptive. 
image

duvld: image
(Just checking it is the same on vscode also)

duvld (approved):
Everything else looks good to me :)

jsapp/js/components/common/button.tsx
magicznyleszek: There is a JSDoc comment for the line where type is defined, but I will move it to the line where button prop is - it should do the trick :ok:

PR #5047, opened 2024-08-05
fix!: reject duplicate submissions

rajpatel24:
## Summary
Implemented logic to detect and reject duplicate submissions.

## Description

We have identified a race condition in the submission processing that causes duplicate submissions with identical UUIDs and XML hashes. This issue is particularly problematic under conditions with multiple remote devices submitting forms simultaneously over unreliable networks.

To address this issue, a PR has been raised with the following proposed changes:

- Race Condition Resolution: A locking mechanism has been added to prevent the race condition when checking for existing instances and creating new ones. This aims to eliminate duplicate submissions.

- UUID Enforcement: Submissions without a UUID are now explicitly disallowed. This ensures that every submission is uniquely identifiable and further mitigates the risk of duplicate entries.

- Introduction of `root_uuid`:

    - To ensure a consistent submission UUID throughout its lifecycle and prevent duplicate submissions with the same UUID, a new `root_uuid` column has been added to the `Instance` model with a unique constraint (`root_uuid` per `xform`).

        - If the `<meta><rootUuid>` is present in the submission XML, it is stored in the `root_uuid` column.

        - If `<meta><rootUuid>` is not present, the value from `<meta><instanceID>` is used instead.

    - This approach guarantees that the `root_uuid` remains constant across the lifecycle of a submission, providing a reliable identifier for all instances.

- UUID Handling Improvement: Updated the logic to strip only the `uuid:` prefix while preserving custom, non-UUID ID schemes (e.g., domain.com:1234). This ensures compliance with the OpenRosa spec and prevents potential ID collisions with custom prefixes.

- Error Handling:
    - 202 Accepted: Returns when content is identical to an existing submission and successfully processed.
    - 409 Conflict: Returns when a duplicate UUID is detected but with differing content.

These changes should improve the robustness of the submission process and prevent both race conditions and invalid submissions.

## Notes

- Implemented a fix to address the race condition that leads to duplicate submissions with the same UUID and XML hash.
- Incorporated improvements from existing work, ensuring consistency and robustness in handling concurrent submissions.
- The fix aims to prevent duplicate submissions, even under high load and unreliable network conditions.

## Related issues

Supersedes [kobotoolbox/kobocat#876](https://github.com/kobotoolbox/kobocat/pull/876)
and https://github.com/kobotoolbox/kobocat/pull/859


kobo/apps/openrosa/apps/api/tests/viewsets/test_xform_submission_api.py
```diff
@@ -438,3 +455,105 @@ def test_submission_blocking_flag(self):
...
+                    survey_=survey,
+                    username_=username,
+                    token_=self.token.key
+                ),
+                range(DUPLICATE_SUBMISSIONS_COUNT),
+            ):
+                results[result] += 1
+
+        assert results[status.HTTP_201_CREATED] == 1
+        assert results[status.HTTP_409_CONFLICT] == DUPLICATE_SUBMISSIONS_COUNT - 1
```
noliveleger, line 516: Pasting @jnm's comment over here:

> does the OpenRosa spec allow returning a 409? and do Enketo and Collect handle a 409 properly? i can't find the code, but i remember wanting to return a 40x that wasn't 400 but being forced to use 400 only because without it Collect wouldn't display the error message i was sending. could've been Enketo, though, or i might be misremembering entirely

https://github.com/kobotoolbox/kobocat/pull/876#discussion_r1210297501

kobo/apps/openrosa/apps/logger/exceptions.py
```diff
@@ -10,6 +10,15 @@ class BuildDbQueriesNoConfirmationProvidedError(Exception):
     pass
 
 
+class ConflictingXMLHashInstanceError(Exception):
+    pass
+
```
noliveleger, line 15: We may get rid of this class and at least rename it. 

Please see @jnm's comment 

https://github.com/kobotoolbox/kobocat/pull/876#discussion_r1253934049

kobo/apps/openrosa/libs/utils/logger_tools.py
```diff
@@ -577,8 +601,13 @@ def safe_create_instance(
...
         error = OpenRosaResponseBadRequest(t("Improperly formatted XML."))
-    except DuplicateInstance:
-        response = OpenRosaResponse(t("Duplicate submission"))
+    except ConflictingXMLHashInstanceError:
+        response = OpenRosaResponse(t('Conflict with already existing instance'))
+        response.status_code = 409
+        response['Location'] = request.build_absolute_uri(request.path)
+        error = response
+    except DuplicateInstanceError:
+        response = OpenRosaResponse(t('Duplicate instance'))
```
noliveleger, line 610: Copying @jnm's comment 

https://github.com/kobotoolbox/kobocat/pull/876#discussion_r1229010121

kobo/apps/openrosa/apps/api/tests/viewsets/test_xform_submission_api.py
```diff
@@ -438,3 +455,105 @@ def test_submission_blocking_flag(self):
                 )
                 response = self.view(request, username=username)
                 self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+
+
+class ConcurrentSubmissionTestCase(RequestMixin, LiveServerTestCase):
```
noliveleger, line 460: As @jnm's asked in former PR [[1](https://github.com/kobotoolbox/kobocat/pull/876#discussion_r1253934049)], ][2](https://github.com/kobotoolbox/kobocat/pull/876#discussion_r1210308268)], do we have tests that cover 

- client uploads XML with no attachment -> accept
- client uploads same XML again -> reject
- client uploads same XML but now with an attached file -> accept; attach file to existing submission

noliveleger (changes requested):

kobo/apps/openrosa/libs/utils/logger_tools.py
```diff
@@ -165,50 +169,54 @@ def create_instance(
...
+        if xform.has_start_time or new_uuid is not None:
+            # XML matches are identified by identical content hash OR, when a
+            # content hash is not present, by string comparison of the full
+            # content, which is slow! Use the management command
+            # `populate_xml_hashes_for_instances` to hash existing submissions
+            existing_instance = Instance.objects.filter(
+                Q(xml_hash=xml_hash) | Q(xml_hash=Instance.DEFAULT_XML_HASH, xml=xml),
+                xform__user=xform.user,
+            ).first()
         else:
```
noliveleger, line 196: We should get rid of this. https://github.com/kobotoolbox/kobocat/pull/876#discussion_r1210322404
If submission does not have a submission uuid, it has to be rejected. It means probably before the lock.

noliveleger: Already discussed previously, but does not the lock let another submission with same xml_hash, root_uuid but new or different attachment come in? Do we have a test for that?

kobo/apps/openrosa/apps/api/tests/viewsets/test_xform_submission_api.py
rajpatel24: Yes we have in `TestFormSubmission`

kobo/apps/openrosa/apps/logger/tests/test_parsing.py
```diff
@@ -140,6 +144,15 @@ def test_get_uuid_from_xml(self):
         instanceID = get_uuid_from_xml(xml_str)
         self.assertEqual(instanceID, "729f173c688e482486a48661700455ff")
 
+        # Additional test case for a custom prefixed UUID
+        submission = """<?xml version="1.0" encoding="UTF-8" ?>
+        <submission xmlns:orx="http://openrosa.org/xforms">
+            <meta><instanceID>uuid:getodk.org:123456789</instanceID></meta>
```
noliveleger, line 150: let's use `kobotoolbox.org` instead of `getodk.org` ;-)

kobo/apps/openrosa/apps/logger/tests/test_parsing.py
```diff
@@ -140,6 +144,15 @@ def test_get_uuid_from_xml(self):
...
+        </submission>
+        """
+        custom_instance_id = get_uuid_from_xml(submission)
+        self.assertEqual(custom_instance_id, "getodk.org:123456789")
```
noliveleger, line 154: Single quotes 🙏

kobo/apps/openrosa/apps/logger/xform_instance_parser.py
```diff
@@ -55,15 +55,19 @@ def get_meta_from_xml(xml_str: str, meta_name: str) -> str:
...
 
-    def _uuid_only(uuid, regex):
-        matches = regex.match(uuid)
-        if matches and len(matches.groups()) > 0:
-            return matches.groups()[0]
-        return None
+    def _uuid_only(uuid):
+        """
+        Strips the 'uuid:' prefix from the provided identifier if it exists.
+        This preserves any custom ID schemes (e.g., 'getodk.org:123456789')
```
noliveleger, line 61: replace `getodk.org` with `kobotoolbox.org`

kpi/tests/api/v2/test_api_submissions.py
```diff
@@ -1584,8 +1595,6 @@ def test_edit_submission_with_xml_missing_uuids(self):
         submission_xml_root = fromstring_preserve_root_xmlns(submission_xml)
         assert submission_json['_id'] == submission['_id']
         assert submission_xml_root.find('./find_this').text == 'hello!'
-        assert submission_xml_root.find('./meta/instanceID') is None
-        assert submission_xml_root.find('./formhub/uuid') is None
```
noliveleger, line 1588: We are rejecting submissions without `meta/instanceID` but the assertion about `formub/uuid` should be still there I think, no?

kobo/apps/openrosa/libs/utils/logger_tools.py
```diff
@@ -601,11 +607,6 @@ def safe_create_instance(
...
     except ExpatError as e:
         error = OpenRosaResponseBadRequest(t("Improperly formatted XML."))
-    except ConflictingXMLHashInstanceError:
-        response = OpenRosaResponse(t('Conflict with already existing instance'))
-        response.status_code = 409
-        response['Location'] = request.build_absolute_uri(request.path)
-        error = response
     except DuplicateInstanceError:
         response = OpenRosaResponse(t('Duplicate instance'))
         response.status_code = 202
```
noliveleger, line 612: We should return a 409 when it's a duplicate. 
Let confirm that with the rest of the team.

noliveleger (changes requested):

kobo/apps/openrosa/apps/logger/tests/test_form_submission.py
```diff
@@ -244,12 +244,45 @@ def test_duplicate_submission_with_different_content(self):
         self.assertEqual(self.response.status_code, 201)
         self.assertEqual(Instance.objects.count(), pre_count + 1)
         inst = Instance.objects.order_by('pk').last()
-        self._make_submission(duplicate_xml_submission_file_path)
-        self.assertEqual(self.response.status_code, 201)
-        self.assertEqual(Instance.objects.count(), pre_count + 2)
+        with self.assertRaises(Exception):
+            self._make_submission(duplicate_xml_submission_file_path)
+        self.assertEqual(self.response.status_code, 401)
```
noliveleger, line 249: This should return a "duplicate instance" error (which is a 202) not a 401.

noliveleger (changes requested):

kobo/apps/openrosa/libs/utils/logger_tools.py
noliveleger: Internal discussion: https://chat.kobotoolbox.org/#narrow/stream/4-Kobo-Dev/topic/Duplicated.20submission.20UUIDs/near/445782

rajpatel24: Will wait for John's comment

kobo/apps/subsequences/tests/test_submission_stream.py
```diff
@@ -257,7 +257,8 @@ def test_stream_with_extras_handles_duplicated_submission_uuids(self):
                 '_uuid': '1c05898e-b43c-491d-814c-79595eb84e81',
             },
         ]
-        self.asset.deployment.mock_submissions(submissions)
+        with self.assertRaises(Exception) as ex:
+            self.asset.deployment.mock_submissions(submissions)
```
noliveleger, line 261: Let's use the expected Exception with `self.assertRaises()`

kpi/deployment_backends/mock_backend.py
```diff
@@ -118,6 +120,14 @@ class FakeRequest:
...
             )
             if error:
+                error_dict = error.__dict__
+                if '_container' in error_dict and len(error_dict['_container']) > 0:
+                    xml_response = error_dict['_container'][0]
+                    message_text = DET.fromstring(xml_response).findtext(
+                        './/{http://openrosa.org/http/response}message'
+                    )
+                    if message_text == 'Instance ID is required':
+                        raise InstanceIdMissingError()
```
noliveleger, line 130: Let's talk about this. It should be raised automatically without having this whole block

rajpatel24: In the test, we use `mock_submissions()`, which in turn calls `safe_create_instance()` for submission processing. If `safe_create_instance()` encounters an issue, it returns an `OpenRosaResponse` with an error code and message, rather than a specific exception class. This is why I added code to re-raise the specific exception based on the error content. However, I'm currently looking into it to improve this handling and make it more efficient

kobo/apps/openrosa/apps/api/tests/viewsets/test_xform_submission_api.py
noliveleger: For the records, the tests below are covering different scenarios:

1) `kobo/apps/openrosa/apps/api/tests/viewsets/test_xform_submission_api.py::ConcurrentSubmissionTestCase::test_post_concurrent_same_submissions` blocks submissions with same data `POST`ed concurrently

2) `test_duplicate_submission_with_different_content` and `test_duplicate_submission_with_same_content_but_with_attachment`

kobo/apps/openrosa/libs/utils/logger_tools.py
```diff
@@ -833,7 +854,10 @@ def _get_instance(
         instance.xml = xml
         instance._populate_xml_hash()
         instance.uuid = new_uuid
-        instance.save()
+        try:
```
noliveleger, line 861: I cannot comment on line 845, but I think we want to filter on `root_uuid` instead. 
```python 
if instance := Instance.objects.filter(root_uuid=old_uuid).first():
        check_edit_submission_permissions(request, xform)
        InstanceHistory.objects.create(
            xml=instance.xml, xform_instance=instance, uuid=old_uuid
        )
        instance.xml = xml
        instance._populate_xml_hash()
        instance.uuid = new_uuid
        ...
```

noliveleger (changes requested):

kobo/apps/subsequences/tests/test_submission_stream.py
```diff
@@ -257,7 +258,8 @@ def test_stream_with_extras_handles_duplicated_submission_uuids(self):
                 '_uuid': '1c05898e-b43c-491d-814c-79595eb84e81',
             },
         ]
-        self.asset.deployment.mock_submissions(submissions)
+        with self.assertRaises(ConflictingSubmissionUUIDError):
+            self.asset.deployment.mock_submissions(submissions)
```
noliveleger, line 262: if `self.asset.deployment.get_submissions` returns only submission, the test is useless.
We may need to find another way to make the test pass. We may need to mock the data only for this test to bypass the unique constraint.

kobo/apps/openrosa/apps/logger/tests/data_from_sdcard/bulk_submission.zip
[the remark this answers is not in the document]
rajpatel24: Yup!

kobo/apps/openrosa/libs/utils/logger_tools.py
```diff
@@ -165,51 +169,43 @@ def create_instance(
 
     # get new and deprecated uuid's
     new_uuid = get_uuid_from_xml(xml)
+    if not new_uuid:
+        raise InstanceIdMissingError()
```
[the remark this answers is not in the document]
rajpatel24: Will keep this in mind ;-)

kobo/apps/openrosa/apps/logger/tests/test_form_submission.py
```diff
@@ -244,12 +244,45 @@ def test_duplicate_submission_with_different_content(self):
...
+        initial_instance = Instance.objects.last()
+        self.assertEqual(self.response.status_code, 201)
+        self.assertEqual(Instance.objects.count(), initial_instance_count + 1)
+
+        # Test duplicate submission with attachment
+        with open(media_file_path, 'rb') as media_file:
+            self._make_submission(xml_submission_file_path, media_file=media_file)
+        self.assertEqual(self.response.status_code, 201)
+        self.assertEqual(Instance.objects.count(), initial_instance_count + 1)
+        self.assertEqual(Attachment.objects.filter(instance=initial_instance).count(), 1)
```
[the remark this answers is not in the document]
rajpatel24: I have added a new test for this. I had to remove `existing_attachment.file_hash == hash_attachment_contents(f.read())` from the logic so that it does not create a new attachment if the file name is the same but the content is different. I'm thinking about the edit scenario (with deprecatedID). Feeling a little sleepy as it’s quite late on my end, but I will definitely test this thoroughly :-)

kobo/apps/openrosa/apps/logger/management/commands/clean_duplicated_submissions.py
```diff
@@ -39,99 +38,158 @@ def add_arguments(self, parser):
             help="Specify a XForm's `id_string` to clean up only this form",
         )
 
+        parser.add_argument(
+            '--delete-unique-uuids',
+            action='store_true',
+            default=False,
+            help='Delete duplicates with identical uuid',
+        )
+
```
noliveleger, line 47: As per our internal discussion, remove this option.

kobo/apps/openrosa/apps/logger/management/commands/clean_duplicated_submissions.py
```diff
@@ -39,99 +38,158 @@ def add_arguments(self, parser):
...
     def handle(self, *args, **options):
         username = options['user']
         xform_id_string = options['xform']
+        self._delete_unique_uuids = options['delete_unique_uuids']
```
noliveleger, line 51: To be removed

kobo/apps/openrosa/apps/logger/management/commands/clean_duplicated_submissions.py
```diff
@@ -39,99 +38,158 @@ def add_arguments(self, parser):
...
         if username:
             query = query.filter(xform__user__username=username)
 
-        query = query.values_list('uuid', flat=True)\
-            .annotate(count_uuid=Count('uuid'))\
-            .filter(count_uuid__gt=1)\
+        query = (
+            query.values_list('xml_hash', flat=True)
+            .annotate(count_xml_hash=Count('xml_hash'))
+            .filter(count_xml_hash__gt=1)
```
noliveleger, line 65: As per our internal discussion, `query` should be on `uuid` (as it was before) because we will compare `xml_hash` later.

kobo/apps/openrosa/apps/logger/management/commands/clean_duplicated_submissions.py
```diff
@@ -39,99 +38,158 @@ def add_arguments(self, parser):
...
-            # Update number of submissions for each user.
-            for user_ in list(self.__users):
-                result = XForm.objects.filter(user_id=user_.id)\
-                    .aggregate(count=Sum('num_of_submissions'))
-                user_.profile.num_of_submissions = result['count']
-                self.stdout.write(
-                    "\tUpdating `{}`'s number of submissions".format(
-                        user_.username))
-                user_.profile.save(update_fields=['num_of_submissions'])
+        for xml_hash in query.iterator():
```
noliveleger, line 69: As per our internal discussion, loop should be on `uuid` now.

kobo/apps/openrosa/apps/logger/management/commands/clean_duplicated_submissions.py
```diff
@@ -39,99 +38,158 @@ def add_arguments(self, parser):
...
+            duplicates_by_xform = self._get_duplicates_by_xform(
+                instances_with_same_xml_hash
+            )
+
+            for (
+                xform_id,
+                instances_with_same_xml_hash,
+            ) in duplicates_by_xform.items():
+                instance_ref = instances_with_same_xml_hash.pop(0)
+                self._clean_up(instance_ref, instances_with_same_xml_hash)
```
noliveleger, line 86: Refactor the code to get all duplicates with same XML hash, and all duplicates with a different XML hash. 
All duplicates with the same XML hash should removed (and their attachments linked to `instance_ref`) and all duplicates with a different XML hash should have only their `uuid` replaced with the new string at line ...

noliveleger (changes requested):

kobo/apps/openrosa/apps/logger/models/instance.py
```diff
@@ -89,6 +90,7 @@ class Instance(AbstractTimeStampedModel):
     # do not use it anymore.
     deleted_at = models.DateTimeField(null=True, default=None)
 
+    root_uuid = models.CharField(max_length=249, null=True, unique=True, db_index=True)
```
noliveleger, line 93: As per our discussion, let's using a unique together constraint with `root_uuid` and `xform_id`.

kobo/apps/openrosa/libs/utils/logger_tools.py
```diff
@@ -225,6 +217,26 @@ def dict2xform(submission: dict, xform_id_string: str) -> str:
...
 
 
+@contextlib.contextmanager
+def get_instance_lock(xml_hash: str, submission_uuid: str, xform_id: int) -> bool:
+    int_lock = int.from_bytes(
+        hashlib.shake_128(
+            f'{xform_id}!!{submission_uuid}!!{xml_hash}'.encode()
+        ).digest(7), 'little'
+    )
+    acquired = False
```
noliveleger, line 227: @rajpatel24 please write a comment to explain why we use this thing.

noliveleger (approved):
@rajpatel24 I'll let you have the honors of merging this PR.

FYI: I pushed one tiny commit for the management command to fix it to avoid another back and forth.

PR #5044, opened 2024-08-01
Get source from requests

rgraber:
## Checklist

1. [x] If you've added code that should be tested, add tests
2. [ ] If you've changed APIs, update (or create!) the documentation N/A
3. [x] Ensure the tests pass
4. [x] Make sure that your code lints and that you've followed [our coding style](https://github.com/kobotoolbox/kpi/blob/master/CONTRIBUTING.md)
5. [x] Write a title and, if necessary, a description of your work suitable for publishing in our [release notes](https://community.kobotoolbox.org/tag/release-notes)
6. [ ] Mention any related issues in this repository (as #ISSUE) and in other repositories (as kobotoolbox/other#ISSUE) N/A
7. [ ] Open an issue in the [docs](https://github.com/kobotoolbox/docs/issues/new) if there are UI/UX changes N/A

## Description

Enable parsing the source (browser and OS) from a request.

## Notes

Adds a new method for parsing out the browser and OS from the HTTP_USER_AGENT header. Relies on a new ua_parser library. Also adds ddt to dev requirements for easier unit tests.

PR #5035, opened 2024-07-30
Update audit logs for faster querying

rgraber:
## Checklist

1. [ ] If you've added code that should be tested, add tests N/A
2. [ ] If you've changed APIs, update (or create!) the documentation N/A
3. [ ] Ensure the tests pass
4. [x] Make sure that your code lints and that you've followed [our coding style](https://github.com/kobotoolbox/kpi/blob/master/CONTRIBUTING.md)
5. [x] Write a title and, if necessary, a description of your work suitable for publishing in our [release notes](https://community.kobotoolbox.org/tag/release-notes)
6. [ ] Mention any related issues in this repository (as #ISSUE) and in other repositories (as kobotoolbox/other#ISSUE) N/A
7. [ ] Open an issue in the [docs](https://github.com/kobotoolbox/docs/issues/new) if there are UI/UX changes N/A

## Description

Add some indexes on audit logs for faster querying by asset, and add a new 'auth' type for access logs.

## Notes
Adds a new indexes for the AuditLog table on (metadata__asset_uid) and (metadata__asset_uid, action) as well as a new AUTH AuditAction

PR #5050, opened 2024-08-06
Create access logs from requests

rgraber:
## Checklist

3. [x] Ensure the tests pass

## Description

Enable the creation of access logs when users are authenticated.

## Notes

Adds a method to the AuditLog class to create from a request, with logic to handle regular logins, non-persisted authenticated requests,  and logins via django-loginas. This is meant to be called from a user_logged_in signal handler, which passes the request along, or an authentication class. Technically it can be called with any request but it will only log the authentication of the user, not any action they performed. 

For any request, it will store the ip, OS, and browser of the request (taken from the headers) as well as information about the user in request.user. The user id is also stored as the object_id, with the object_type being the name of the user model.

If the url_name of the resolver is 'loginas-user-login,' we also store information about the _cached_user, which represents the user making the original request. This should work even in the case of cURL requests since they require a CSRF token and session id, so the original user gets fully authenticated.

The authorization type can be set by the caller, otherwise it will default to 'django-loginas' if it's a loginas request and the name of the authenticating backend if it's a normal login request.

## Related issues
Blocked by https://github.com/kobotoolbox/kpi/pull/5035 and https://github.com/kobotoolbox/kpi/pull/5044


rgraber:
Closed in favor of https://github.com/kobotoolbox/kpi/pull/5055

PR #5055, opened 2024-08-08
Record all logins as access logs

rgraber:
## Checklist

3. [x] Ensure the tests pass

## Description

Create an access log every time someone logs in.

## Notes
Adds a method to the AuditLog class to create a new access log with details about the authentication of a request. It will store the usual AuditLog fields, plus metadata about the source of the request and, in the case of django-loginas, the super user who actually made the request (this is stored in request._cached_user). The method can also take a user, which is useful for rare cases when request.user is empty (like in tests). The last optional parameter is an authentication type, which can be customized to identify the authentication as Token or Digest or similar.

Also adds a receiver to the user_logged_in signal to call the new method. The signal automatically passes the request along so we don't have to worry about that. There are tests for a simple login flow, loginas, MFA, and SSO login. A future PR will handle creating access logs for one-time authentication like token, Oauth, digest, etc.

To test:
1. login using normal login, sso, or mfa
2. in a python shell or in postgres, check that a new audit log has been created of type 'auth'
3. using a superuser account, login as another user from django-admin
4. check that a new audit log has been created recording the login of the new user, but with the super user name  and uid in the metadata


kobo/apps/audit_log/models.py
```diff
@@ -1,10 +1,22 @@
+import logging
+
+from django.conf import settings
+from django.contrib.auth import get_user_model
```
noliveleger, line 4: FYI: you can import this instead:

`from kobo.apps.kobo_auth.shortcuts import User`

kobo/apps/audit_log/tests/test_models.py
```diff
@@ -0,0 +1,121 @@
...
+        self.assertEqual(audit_log.user.id, user.id)
+        self.assertEqual(audit_log.app_label, KOBO_AUTH_APP_LABEL)
+        self.assertEqual(audit_log.model_name, 'User')
+        self.assertEqual(audit_log.object_id, user.id)
+        self.assertEqual(audit_log.user_uid, user.extra_details.uid)
+        self.assertEqual(audit_log.action, AuditAction.AUTH)
+
+    def test_basic_create_auth_log_from_request(self, patched_ip, patched_source):
+        request = self._create_request(
+            '/accounts/login/', AnonymousUser(), AuditLogModelTestCase.super_user
```
noliveleger, line 50: I'd rather prefer to use `reverse('view_name') than hard-coded endpoint.

kobo/apps/audit_log/tests/test_models.py
```diff
@@ -0,0 +1,121 @@
...
+            },
+        )
+
+    def test_create_auth_log_from_loginas_request(self, patched_ip, patched_source):
+        second_user = get_user_model().objects.create_user(
+            'second_user', 'second@example.com', 'pass'
+        )
+        second_user.save()
+        request = self._create_request(
+            f'/admin/login/user/{second_user.id}/',
```
noliveleger, line 69: Same comment for hard-coded endpoint. Let's use `reverse('loginas-user-login', args=(second_user.id,)` instead.

kobo/apps/audit_log/tests/test_models.py
```diff
@@ -0,0 +1,121 @@
...
+        )
+
+    def test_create_auth_log_unknown_authenticator(self, patched_ip, patched_source):
+        # no backend attached to the user object
+        second_user = get_user_model().objects.create_user(
+            'second_user', 'second@example.com', 'pass'
+        )
+        second_user.save()
+        request = self._create_request(
+            f'/api/v2/assets/', AuditLogModelTestCase.super_user, second_user
```
noliveleger, line 110: 👀

kobo/apps/audit_log/tests/test_signals.py
```diff
@@ -0,0 +1,140 @@
+import contextlib
+from unittest.mock import patch
+
+from allauth.account.models import EmailAddress
+from django.contrib.auth import get_user_model
```
noliveleger, line 5: Let's use `from kobo.apps.kobo_auth.shortcuts import User` instead

kobo/apps/audit_log/tests/test_signals.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def test_simple_login(self):
+        count = AuditLog.objects.count()
+        self.assertEqual(count, 0)
+        user = AuditLogSignalsTestCase.user
+        data = {
+            'login': 'user',
+            'password': 'pass',
+        }
+        with self.settings(ACCOUNT_EMAIL_VERIFICATION='none'):
```
noliveleger, line 62: Is this `this.settings` documented somewhere? 

We usually use  [override_settings](https://docs.djangoproject.com/en/4.2/topics/testing/tools/#django.test.override_settings) 

```python 
@override_settings(ACCOUNT_EMAIL_VERIFICATION='none')
def test_simple_login(self):
     ... 
```

noliveleger (changes requested):

kobo/apps/audit_log/tests/test_signals.py
rgraber: It is documented here: https://docs.djangoproject.com/en/5.1/topics/testing/tools/#overriding-settings
But I'm perfectly happy to use override_settings

noliveleger: Nope, that's great ;-) keep it! Sorry about my ignorance.

kobo/apps/audit_log/tests/test_signals.py
```diff
@@ -0,0 +1,142 @@
...
+    @override_settings(ACCOUNT_EMAIL_VERIFICATION='none')
+    def test_simple_login(self):
+        count = AuditLog.objects.count()
+        self.assertEqual(count, 0)
+        user = AuditLogSignalsTestCase.user
+        data = {
+            'login': 'user',
+            'password': 'pass',
+        }
+        with self.settings(ACCOUNT_EMAIL_VERIFICATION='none'):
```
noliveleger, line 64: Maybe it's redundant because of the decorator?

noliveleger (changes requested):
LGTM.
Just pick whatever you prefer between `self.settings` or `@override_settings` , then you can merge (and delete the branch).

kobo/apps/audit_log/tests/test_signals.py
rgraber: My b. Good catch.

PR #5059, opened 2024-08-12
Fix bug causing large asynchronous translations to fail

Guitlle:
## Description

Fix the backend to handle large (async) translations using google cloud integration

## Notes

This includes a lot of refactoring of old code as well as deleting some old code that is not being used anywhere on the code base. There is still one class that I don't know yet if it'll be used. Further discussion to delete it is needed. This integration was tested for all 4 cases: sync/async and transcriptions/translations.

## Related issues

Fixes [Notion task 709](https://www.notion.so/kobotoolbox/Fix-bug-causing-large-asynchronous-translations-to-fail-ca1ba10181bb46659421be4294b764fc)


jsapp/js/components/processing/singleProcessingStore.ts
```diff
@@ -653,22 +658,22 @@ class SingleProcessingStore extends Reflux.Store {
...
   }
 
-  private isAutoTranscriptionEventApplicable(event: AutoTranscriptionEvent) {
+  private isAutoTranscriptionEventApplicable(event: AutoTransXEvent) {
     // Note: previously initiated automatic transcriptions may no longer be
     // applicable to the current route
-    const googleTsResponse =
-      event.response[this.currentQuestionQpath]?.googlets;
+    const googleTxResponse =
+      event.response[this.currentQuestionQpath]?.googletx;
```
magicznyleszek, line 665: @Guitlle shouldn't this be `.googlets` and `googleTsResponse`?

jsapp/js/components/processing/singleProcessingStore.ts
```diff
@@ -105,7 +105,7 @@ interface SubmissionsEditIds {
   }>;
 }
 
-interface AutoTranscriptionEvent {
+interface AutoTransXEvent {
```
magicznyleszek, line 108: @Guitlle  We already use `TransxObject` and `TransxLanguages` etc. (lowercase "x"). Let's keep this consistent

jsapp/js/components/processing/singleProcessingStore.ts
Guitlle: Yes, this is the missing part for the frontend, it's still WIP for the frontend, maybe we can talk about it later, I think I could handle it but I have some questions

jsapp/js/components/processing/singleProcessingStore.ts
Guitlle: Ok

magicznyleszek: I've fixed that :ok:

kobo/apps/subsequences/exceptions.py
```diff
@@ -10,3 +10,11 @@ class TranscriptionResultsNotFound(Exception):
...
     No results returned by specified transcription service
     """
+
+
+class TranslationResultsNotFound(Exception):
+    pass
+
+
+class TranslationAsyncResultAvailable(Exception):
+    pass
```
noliveleger, line 20: Keep alphabetical sorting :pray:

kobo/apps/subsequences/constants.py
```diff
@@ -9,5 +12,5 @@
 GOOGLE_CACHE_TIMEOUT = 28800  # 8 hours
 
 
-def make_async_cache_key(user_id: int, submission_id: str, xpath: str, source: str):
-    return f'{SUBSEQUENCES_ASYNC_CACHE_KEY}-{user_id}-{submission_id}-{xpath}-{source}'
+def make_async_cache_key(*args: List[Any]) -> str:
```
noliveleger, line 15: 🤔 , any reason why you changed this? 
I think I preferred when what is composing the key was explicit. 
Could you at least leave a comment what we are expecting here? Other devs may not know what it is needed to retrieve the key.

kobo/apps/subsequences/integrations/google/base.py
```diff
@@ -0,0 +1,128 @@
...
+from django.core.cache import cache
+
+from kobo.apps.trackers.utils import update_nlp_counter
+from kpi.utils.log import logging
+from .utils import google_credentials_from_constance_config
+from ...models import SubmissionExtras
+from ...constants import GOOGLE_CACHE_TIMEOUT, make_async_cache_key
+from ...exceptions import SubsequenceTimeoutError
+
+REQUEST_TIMEOUT = 10  # seconds
```
noliveleger, line 20: What about making a setting with this constant? It could be also a Constance setting.
What do you think?

kobo/apps/subsequences/constants.py
```diff
@@ -1,5 +1,8 @@
+from typing import Any, List
```
noliveleger, line 1: `List` is deprecated. Use `from __future__ import annotations` and use lowercase type everywhere instead.
See https://docs.python.org/3/library/typing.html#deprecated-aliases

kobo/apps/subsequences/integrations/google/base.py
```diff
@@ -0,0 +1,128 @@
+from abc import ABC, abstractmethod
+from concurrent.futures import TimeoutError
+from typing import Any, Tuple
```
noliveleger, line 3: `Tuple` is deprecated. Use `from __future__ import annotations` and use lowercase type everywhere instead.
See https://docs.python.org/3/library/typing.html#deprecated-aliases

kobo/apps/subsequences/integrations/google/google_transcribe.py
```diff
@@ -3,63 +3,182 @@
 import posixpath
 from concurrent.futures import TimeoutError
 from datetime import timedelta
+from typing import Union, Tuple, Any
```
noliveleger, line 6: `Tuple` is deprecated. Use `from __future__ import annotations` and use lowercase type everywhere instead.
See https://docs.python.org/3/library/typing.html#deprecated-aliases

kobo/apps/subsequences/integrations/google/google_translate.py
```diff
@@ -41,92 +52,207 @@ def __init__(self):
...
+                blob.delete()
+
+        return text
+
+    def get_unique_paths(
+        self,
+        xpath: str,
+        source_lang: str,
+        target_lang: str,
+    ) -> Tuple[str, str]:
```
noliveleger, line 156: `tuple[str, str]`

kobo/apps/subsequences/integrations/google/google_translate.py
```diff
@@ -41,92 +52,207 @@ def __init__(self):
...
-        self.submission_uuid = submission_uuid
-        self.username = username
-        self.xpath = xpath
-        _uniq_path = _hashed_strings(self.submission_uuid, self.xpath)
-        _uniq_dir = f'{self.date_string}/{_uniq_path}'
-        bucket_prefix = constance.config.ASR_MT_GOOGLE_STORAGE_BUCKET_PREFIX
-        source_path = posixpath.join(bucket_prefix, _uniq_dir, 'source.txt')
-        output_dir = posixpath.join(bucket_prefix, _uniq_dir, 'completed/')
+        content: Any,
+    ) -> Tuple[object, int]:
```
noliveleger, line 81: `tuple[object, int]`

kobo/apps/subsequences/integrations/google/base.py
```diff
@@ -0,0 +1,128 @@
...
+        pass
+
+    @abstractmethod
+    def begin_google_operation(
+        self,
+        xpath: str,
+        source_lang: str,
+        target_lang: str,
+        content: Any,
+    ) -> Tuple[object, int]:
```
noliveleger, line 56: `tuple[object, int]`

kobo/apps/subsequences/integrations/google/google_transcribe.py
```diff
@@ -3,63 +3,182 @@
...
+
+        return {
+            'status': 'complete',
+            'value': value,
+            'languageCode': language_code,
+            'regionCode': region_code,
+        }
+
+    def transcribe_file(
+        self, xpath: str, source_lang: str, content: Tuple[object, int]
```
noliveleger, line 164: `tuple[object, int]`

kobo/apps/subsequences/models.py
```diff
@@ -34,124 +28,21 @@ class Meta:
...
-                            'fullResponse': results,
-                            'languageCode': language_code,
-                            'regionCode': region_code,
-                        }
-                    else:
-                        continue
-                except (KeyError, TypeError) as err:
-                    continue
+        from .integrations.google.google_transcribe import GoogleTranscriptionService
+        from .integrations.google.google_translate import GoogleTranslationService
```
noliveleger, line 33: Why here? Because of circular imports? This could not be avoided? 
Please add a comment if not.

kobo/apps/subsequences/tests/test_nlp_integration.py
```diff
@@ -0,0 +1,130 @@
...
+        assert 'services' in built_params
+
+
+    def test_instantiate_action_with_params(self):
+        survey = self.asset.content
+        action_params = AutomaticTranscriptionAction.build_params({}, survey)
+        action_instance = AutomaticTranscriptionAction(action_params)
+        assert action_instance is not None
+
+    @pytest.mark.skip(reason='transcription currently does not depend on this working')
```
noliveleger, line 63: Is it something that comes from former PR? 
If not, I don't understand the reason why this test is skipped?

noliveleger (changes requested):

kobo/apps/subsequences/constants.py
Guitlle: Yeah, I agree. This was from previous commits. I'll fix it

kobo/apps/subsequences/exceptions.py
Guitlle: ok

Guitlle: ok

kobo/apps/subsequences/integrations/google/base.py
Guitlle: Sure, sounds good

kobo/apps/subsequences/constants.py
Guitlle: Ok

kobo/apps/subsequences/tests/test_nlp_integration.py
Guitlle: Yes, this is testing a feature that is not really implemented. The Actions architecture that was made for the subsequences app is supposed to handle pending actions through Actions classes. So what I understand is this: instead of handling the processing on the save override for the submission extras model, they could be handled automatically through these actions architecture. However, the translation and transcription actions are not really implemented that way. And the save override doesn't use the actions interface to execute any pending actions. Right now the actions classes are being used only to specify the JSON schema modifications done to the submission extra data. 

So I think I could adapt the current solution to this actions architecture in order to avoid having this tech debt hanging in there.

noliveleger: sounds a good idea ;-) 
@jnm  what do you think?

kobo/apps/subsequences/integrations/google/base.py
```diff
@@ -1,3 +1,5 @@
+# coding: utf-8
+from __future__ import annotations
```
noliveleger, line 2: No need to add `coding: utf-8` anymore since we don't support Python 2 anymore.

kobo/apps/subsequences/integrations/google/google_transcribe.py
```diff
@@ -1,6 +1,7 @@
 # coding: utf-8
 import uuid
 import posixpath
+from __future__ import annotations
```
noliveleger, line 4: `from __future__ import annotations` should be the first line of the imports and should be in its own block 

```python 
from __future__ import annotations 

import uuid
imporrt posixpath 
```

kobo/apps/subsequences/integrations/google/base.py
Guitlle: ok, removed it

kobo/apps/subsequences/models.py
Guitlle: ok

kobo/apps/subsequences/tests/test_nlp_integration.py
jnm: We had a chat about this and decided to leave it alone for now. We will have to pay off the tech debt eventually, but it's a higher priority to address other NLP-related tech debt like https://www.notion.so/kobotoolbox/Pick-different-delimiter-for-qpath-probably-with-URL-encoding-e1271ab8ec4442bcb7c5af231f751127

Guitlle: Ok, thanks John and Olivier for the feedback.

PR #5064, opened 2024-08-19
Add created_by and last_modified_by fields to asset model

rajpatel24:
## Checklist

3. [x] Ensure the tests pass

## Description

This PR introduces two new fields to the Asset model:
- `created_by`: A CharField that stores the username of the asset's creator. This field is indexed to allow efficient querying. If an asset has never been transferred, the `created_by` field is populated with the username of the asset's owner during migration.
If an asset has been transferred, the migration checks the Invite model to determine the first transfer and uses the username of the sender of the invite to populate the `created_by` field.

- `last_modified_by`: A CharField that stores the username of the last user who modified the asset. This field is also indexed.
Since accurate historical data for this field is not available, it remains null.

## Notes

- In this PR, I have added two new fields to the Asset model, `created_by` and `last_modified_by`. The addition of these fields aims to enhance traceability by capturing the user responsible for creating an asset and the user who last modified it.

- The `created_by` field will help identify the original creator of the asset, which is useful for auditing and ownership tracking.
- The `last_modified_by` field is essential for understanding the history of changes made to an asset, allowing for better accountability.


kpi/models/asset.py
```diff
@@ -859,6 +861,11 @@ def save(
     ):
         is_new = self.pk is None
 
+        if self.owner:
+            if is_new:
+                self.created_by = self.owner.username
+            self.last_modified_by = self.owner.username
+
```
noliveleger, line 868: You better to do it at the API level. 
The asset could be modified by someone else with correct permissions.

noliveleger (changes requested):
Please add unit tests and a management command to populate fields as per our discussion ;-)

kpi/management/commands/populate_asset_created_and_modified_by.py
```diff
@@ -0,0 +1,67 @@
...
+                # Set created_by based on the first transfer's invite sender or fallback to the owner
+                transfers = asset.prefetched_transfers
+                owner = asset.owner
+                if transfers and transfers[0].invite:
+                    first_transfer = transfers[0]
+                    asset.created_by = first_transfer.invite.sender.username
+                elif owner:
+                    asset.created_by = owner.username
+                else:
+                    asset.created_by = None
```
noliveleger, line 60: Is there a case where `asset.owner` is None?

kpi/management/commands/populate_asset_created_and_modified_by.py
```diff
@@ -0,0 +1,67 @@
...
+            assets = paginator.page(page).object_list
+
+            for asset in assets:
+                if self._verbosity >= 1:
+                    self.stdout.write(f'\tAsset {asset.uid}...')
+
+                # Set created_by based on the first transfer's invite sender or fallback to the owner
+                transfers = asset.prefetched_transfers
+                owner = asset.owner
+                if transfers and transfers[0].invite:
```
noliveleger, line 54: We should be sure that `transfers[0]` is really the first transfer for this asset. 
E.g.: 
- Raj transfers project A to Olivier (Transfer n°1)
- Olivier transfers project A to John (Transfer n°2) 

In the `prefetch_related`, we should add an `order_by('date_created')` to be sure that `transfers[0]` equals `Transfer n°1`

noliveleger (changes requested):

kpi/management/commands/populate_asset_created_and_modified_by.py
rajpatel24: Thanks for pointing this out! I've updated the code to order the transfers by their creation date, ensuring that transfers[0] is indeed the first transfer for the asset

kpi/management/commands/populate_asset_created_and_modified_by.py
rajpatel24: There is a case where `asset.owner` can be `None` because the `owner` field in the `Asset` model allows `null` values `(null=True)`. Although the chances are less, it's possible that a record might contain `null` data for the `owner` field. To handle this and be on the safe side, I've added checks.

noliveleger (approved):

PR #5067, opened 2024-08-20
Create new endpoints for viewing access logs

rgraber:
## Checklist

3. [x] Ensure the tests pass

## Description

Creates two new endpoints for listing access logs, one for a specific user, and the other for all users.

## Notes

Adds /access-logs and /access-logs/all endpoints. The former will show all access logs for the currently logged-in user. The latter will show all access logs for all users, and is only available to superusers. The latter also can be searched the same way as the /audit-log endpoint (ie using ?q=...). Both endpoints can be paginated by offset and limit.

### DIFF
--- a/kobo/apps/audit_log/tests/api/v2/test_api_audit_log.py
+++ b/kobo/apps/audit_log/tests/api/v2/test_api_audit_log.py
@@ -255,7 +255,7 @@ class AllApiAccessLogsTestCase(BaseAuditLogTestCase):
     def test_show_all_access_logs_succeeds_for_superuser(self):
         self.force_login_user(User.objects.get(username='admin'))
         response = self.client.get(self.url)
-        self.assert_audit_log_nowresults_equal(
+        self.assert_audit_log_results_equal(
             response=response, expected_kwargs={'action': AuditAction.AUTH}
         )
 
--- a/kobo/apps/audit_log/tests/api/v2/test_api_audit_log.py
+++ b/kobo/apps/audit_log/tests/api/v2/test_api_audit_log.py
@@ -1,7 +1,7 @@
 from datetime import timedelta
 
 from django.contrib.auth import get_user_model
-from django.utils.timezone import now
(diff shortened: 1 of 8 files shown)


kobo/apps/audit_log/tests/api/v2/test_api_audit_log.py
```diff
@@ -106,11 +156,132 @@ def test_filter_list(self):
...
+            user=User.objects.get(username='someuser'),
+            app_label='foo',
+            model_name='bar',
+            object_id=1,
+            action=AuditAction.DELETE,
+        )
+        log.save()
+        self.assertEqual(AuditLog.objects.count(), 4)
+
+    def test_list_as_anonymous(self):
```
noliveleger, line 245: Not specific to this test only, what do you think about having more explicit names?
For example for this test,  something like `test_list_not_authorized_as_anonymous` . IMO, It makes the behaviour more obvious to the reader.

kobo/apps/audit_log/tests/api/v2/test_api_audit_log.py
```diff
@@ -106,11 +156,132 @@ def test_filter_list(self):
...
+        self.assert_audit_log_results_equal(
+            response=response,
+            expected_kwargs={'action': AuditAction.AUTH, 'user': another_user},
+        )
+
+    def test_search_by_date(self):
+        another_user = User.objects.get(username='anotheruser')
+        with skip_login_access_log():
+            self.client.force_login(User.objects.get(username='admin'))
+        tomorrow = now() + timedelta(days=1)
```
noliveleger, line 275: Not related to your code but we are trying to use `timezone.now()` (and import `from django.utils import timezone`) instead of just using `now()` which could be ambiguous. If you see former import, do not hesitate to correct them.

noliveleger (changes requested):

noliveleger (approved):

PR #5073, opened 2024-08-22
Code refactoring: ServiceUsageSerializer

Guitlle:
## Description

This PR contains a refactor of the ServiceUsageSerializer. The calculations and counts were moved to a utility class that can be reused on other places besides the serializer. It includes the corresponding unit tests.

## Related issues

Fixes [TASK-847](https://www.notion.so/kobotoolbox/Refactor-ServiceUsageSerializer-af0dca4964234135a68c139d8347a256)


kpi/tests/test_usage_calculator.py
```diff
@@ -0,0 +1,220 @@
...
+                        'id': self.attachment_id + 1,
+                        'download_url': 'http://testserver/anotheruser/audio_conversion_test_image.jpg',
+                        'filename': 'anotheruser/audio_conversion_test_image.jpg',
+                        'mimetype': 'image/jpeg',
+                    },
+                ],
+                '_submitted_by': 'anotheruser',
+            }
+            # increment the attachment ID for each attachment created
+            self.attachment_id = self.attachment_id + 2
```
noliveleger, line 146: We can get rid of `self.attachment_id` since `_attachments[index]['id']` is set directly in `mock_submissions()`. 
If we use in somewhere in the tests, we need to use something else to get the id of the attachment (e.g. `submission[0]['_attachments'][idx]`)

Guitlle: Ok, also removed the self.counter property

PR #5080, opened 2024-08-26
Group submission access logs

rgraber:
## Checklist

3. [x] Ensure the tests pass

## Description

Group together submissions that occurred within an hour of each other when reporting in the API.

## Notes

Create proxy models for submission access logs and submission group access logs. Using a post_save signal, whenever a submission access log is saved, look for any submission groups that were last updated within the past hour. If it finds one, adds to the count and updates the last updated date. If not, start a new group.
This PR also contains some refactoring of many the tests to extract out some common logic and use a common fixture.

This is very much a quick and dirty solution. It assumes we never need to go back and re-link the groups to the submissions that are included, or backport  older submission logs, or group existing logs. It also means that if we have 1 submission every 59 minutes for a while they will all be grouped together, although that is easily changed if we want to change it to all submission logs from 6-7, 7-8, 8-9, etc.
The benefit is that it keeps the view and serializer very simple, and means we don't have to join the audit log table against itself.


rgraber:
Putting this back in draft so I can simplify the decoder

rgraber:
Replaced by https://github.com/kobotoolbox/kpi/pull/5094

PR #5082, opened 2024-08-27
Add GitHub Actions for ruff for Python linting

noliveleger:
## Description

For developers interest, GitHub Actions is used to raise errors when the python code does not lint

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,18 +6,28 @@
 # verbose = true
 line-length = 80
 skip-string-normalization = true
+
 [tool.isort]
 profile = "black"
 known_first_party = "kobo"
 no_lines_before = ["LOCALFOLDER"]
 
 [tool.ruff]
-line-length = 80
+line-length = 88
 [tool.ruff.format]
 quote-style = "single"  # Preserve is coming soon to ruff
 [tool.ruff.lint]
-# Enable ruff isort
-extend-select = ["I"]
+extend-select = [
+    "I", # Enable ruff isort
+    "Q", # Flake quotes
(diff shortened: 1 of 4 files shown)


Guitlle (approved):

pyproject.toml
```diff
@@ -6,18 +6,28 @@
...
 skip-string-normalization = true
+
 [tool.isort]
 profile = "black"
 known_first_party = "kobo"
 no_lines_before = ["LOCALFOLDER"]
 
 [tool.ruff]
-line-length = 80
+line-length = 88
```
rgraber, line 16: Why are we making the line-length longer for ruff than for black?

rgraber (commented):
Question about line-length

pyproject.toml
noliveleger: Because we use an extra buffer of ~10 characters. If your line is 85 characters and you don't want to break it for 5 characters, we don't want the linter to complain about it. Moreover, that's the default configuration (see https://docs.astral.sh/ruff/configuration/)

> # Same as Black.
> line-length = 88
> indent-width = 4

I would leave black as-is to format with 80 characters. Does not mean the dev cannot bring a line up to 88 characters after Black formatting.

noliveleger: In PyCharm, I added two visual limits. One at 80 (kinda soft-limit) and the other one at 90 (kind hard-limit).

Screenshot 2024-08-28 at 09 36 09

rgraber: I'm still a little confused. Why do we have black at 80 characters then?

rgraber: Point still holds. Why do we cap black at 80?

PR #5083, opened 2024-08-28
Add Python linter to GitHub Actions

noliveleger:
## Description 

Use `darker` with `flake8` to lint changes (only) in Python code. 

## Notes 

An utility has been created in this PR to be used before submitting new commits. It has not been added to pre-commit hook to give the developer the liberty to edit changes make by the formatter before pushing their changes. 

It uses `ruff` first, then `darker`. `ruff` should only check (unused and/or unsorted) imports, single vs double quotes and some misc other rules. This should apply on the whole content of every edited or new file of a PR.

The main formatting is delegated to `darker`. This should apply only on parts that changed of every edited or new file of a PR.

To run: 

`bash format-python.sh [revision]` 

By default, it looks for changes with the `beta` branch if `revision` is not specified. `-l` (L in lowercase) or `--last` option can be used to make the diff with last commit in the history log. 

**Examples:** 
- `bash format-python.sh`  # use diff with `origin/beta`
- `bash format-python.sh main` # use diff with `main`
- `bash format-python.sh abcd1234` # use diff with commit `abcd1234`
- `bash format-python.sh -l` # use diff with last commit


pyproject.toml
```diff
@@ -6,22 +6,51 @@
...
+
 [tool.isort]
 profile = "black"
 line_length = 80 # same as black
 known_first_party = "kobo"
 no_lines_before = ["LOCALFOLDER"]
 
 [tool.ruff]
-line-length = 80
+line-length = 88
```
rgraber, line 25: If we're capping this at 88 we should cap black and isort at 88 as well, or put a comment explaining why they are different.

rgraber (changes requested):
Can we update PULL_REQUEST_MD as well? Something like 
`4. [ ] Run ./python-format.sh to make sure that your code lints and that you've followed [our coding style](https://github.com/kobotoolbox/kpi/blob/master/CONTRIBUTING.md).`

pyproject.toml
rgraber: I know we discussed this on an earlier PR, but I still don't understand why we kept our black config at 80

dependencies/pip/dev_requirements.txt
```diff
@@ -642,4 +682,7 @@ yarl==1.9.4
     # via aiohttp
 yubico-client==1.13.0
     # via django-trench
+
+# The following packages are considered to be unsafe in a requirements file:
```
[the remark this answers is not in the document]
noliveleger: Nope, produced by `./pip-compile.sh`

pyproject.toml
noliveleger: We want the formatter to break at 80 but we don't want the linter to complain if we manually keep in purpose at line between 80 and 90 (as explained before) without adding `#noqa` or even worse `#fmt off/on`.

dependencies/pip/dev_requirements.txt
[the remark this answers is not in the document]
rgraber: Weird. It seems unrelated to the changes, which is a little worrisome.

pyproject.toml
rgraber: We should discuss this more at a later date. I think we should add `#fmt off` anywhere we are purposefully ignoring formatting rules

rgraber: To put it another way: I think we need to decide firmly whether 88 is an acceptable limit. If it is, it should be the limit in all cases and we don't need to warn at 80. If not, we should standardize the line-length at 80 and explicitly call out lines that break that rule with `#fmt off` and an explanation

noliveleger: Let's not waste our time on this. Let's use 88 everywhere since that's the default for Black and ruff. Moreover, I really don't want to add `#fmt on/off` each time my line is just 2 characters above 80.

format-python.sh
```diff
@@ -0,0 +1,30 @@
...
+#  - Indentation warning (--select W1)
+#  - No newline at end of file (--select W292)
+PYTHON_CHANGES=$(git diff --name-only "$BASE_REVISION" | grep '\.py')
+if [ -n "$PYTHON_CHANGES" ]; then
+    echo "Using ruff..."
+    git diff --name-only "$BASE_REVISION" | grep '\.py' | xargs ruff check --select Q --select I --select F401 --select UP026 --select UP034 --select UP039 --select W292 --fix
+
+    # Applying Black format on changes (since $BASE_REVISION)
+    echo "Using darker..."
+    darker -i -r "$BASE_REVISION"
```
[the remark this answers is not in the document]
noliveleger: I replaced with long name options to make it more obvious.

rgraber (approved):
LGTM

PR #3869, opened 2022-06-14
Permission assignments bug fix

noliveleger:
## Checklist

1. [x] If you've added code that should be tested, add tests
2. [ ] If you've changed APIs, update (or create!) the documentation
3. [x] Ensure the tests pass
4. [x] Make sure that your code lints and that you've followed [our coding style](https://github.com/kobotoolbox/kpi/blob/master/CONTRIBUTING.md)
5. [x] Write a description of your work suitable for publishing on [our forum](https://community.kobotoolbox.org/tag/release-notes)
6. [x] Mention any related issues in this repository (as #ISSUE) and in other repositories (as kobotoolbox/other#ISSUE)
7. [ ] Open an issue in the [docs](https://github.com/kobotoolbox/docs/issues/new) if there are UI/UX changes

## Description

Save permission assignments correctly when usernames start with the same letter but contain non alphanumeric characters (such as `_`) and have a mix of upper and lower case. 

## Related issues

Fixes #3864

PR #5087, opened 2024-08-29
Reduce queries in bulk permission assignment api

RuthShryock:
## Checklist

3. [X] Ensure the tests pass

## Description

Refactored `assign_perm()` and `remove_perm()` to support bulk assignment and removal of permissions instead of processing them individually

### Some background 
This was the original task: https://github.com/kobotoolbox/kpi/pull/3869 added more queries to the endpoint. We should review and either reduce the number of queries or prove why this is necessary. Of course we should also inspect average load times - the goal is to make it faster, not just reduce queries. 
After some clarification, the issue is that we call `assign_perm()` and `remove_perm()` for each permission assignment, one by one. This is costly  and we want to reduce the number of queries and speed-up the endpoint so we should rewrite `assign_perm()` and `remove_perm()` to support bulk assignment.

### Testing

Test cases have been added but if you would like to test manually:
1. Pick a test in `asset_permission_assignment.py` such as `test_partial_permission_grants_implied_view_asset`
2. Add `self.assertNumQueries`
3. Compare before and after the PR


kpi/tests/api/v2/test_api_asset_permission_assignment.py
```diff
@@ -938,3 +940,48 @@ def test_partial_permission_invalid(self):
...
+
+        # Assign permissions one by one
+        with CaptureQueriesContext(connection) as old_context:
+            for user in [self.someuser, self.anotheruser]:
+                for perm in permissions:
+                    self.asset.assign_perm(user, perm)
+        old_method_query_count = len(old_context.captured_queries)
+
+        # Bulk assign permissions
+        with CaptureQueriesContext(connection) as new_context:
```
rgraber, line 955: We should test this with an assertNumQueries so we'll know if something makes the numbers creep up again

kpi/tests/api/v2/test_api_asset_permission_assignment.py
```diff
@@ -938,3 +940,48 @@ def test_partial_permission_invalid(self):
...
+
+        # Remove permissions one by one
+        with CaptureQueriesContext(connection) as old_context:
+            for user in [self.someuser, self.anotheruser]:
+                for perm in permissions:
+                    self.asset.remove_perm(user, perm)
+        old_method_query_count = len(old_context.captured_queries)
+
+        # Bulk remove permissions
+        with CaptureQueriesContext(connection) as new_context:
```
rgraber, line 981: Ditto above

rgraber (changes requested):
Mostly style comments and a thought about testing

noliveleger (changes requested):

kpi/mixins/object_permission.py
```diff
@@ -442,118 +442,158 @@ def get_all_implied_perms(cls, for_instance=None):
...
         post_assign_perm.send(
             sender=self.__class__,
             instance=self,
             user=user_obj,
-            codename=codename,
+            codename=', '.join(perm),
         )
 
         # Recalculate all descendants
         self.recalculate_descendants_perms()
```
noliveleger, line 595: If there a way to improve this method too?

noliveleger (changes requested):
Do we have a test that checks expected permission assignments are there when we use `assign_perms()` with many multiple permissions at a time. 

I would like to be sure that assigning permissions with collections for example we still work the way we expect. 

By the way, I would keep`assign_perm()` (singular) and it would call your new `assign_perms` behind the scene. We would avoid changing `assign_perm` everywhere. 
We can use typing annotation to know which type each method is expecting.

kpi/mixins/object_permission.py
```diff
@@ -442,118 +442,158 @@ def get_all_implied_perms(cls, for_instance=None):
...
+            implied_permissions.update(
+                self.get_implied_perms(
+                    codename, reverse=deny, for_instance=self
+                ).intersection(assignable_permissions)
+            )
+
+        if new_permissions:
+            ObjectPermission.objects.bulk_create(new_permissions)
+            if len(new_permissions) == 1:
+                new_permissions = new_permissions[0]
```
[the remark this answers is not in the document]
RuthShryock: This was needed because a lot of code calling `assign_perm()`` was expecting a single string in return. So I wanted to make sure that if we just assigned one permission, it would return one string as expected. But now, since we are keeping the original `assign_perm()` function and that will still be used when assigning a single permission, it is no longer needed.

jdimeo:
We're very excited about this PR because what we want to do (use APIs to create shared questions in the library and then automatically grant users access based on their country) is blocked by the inability of Kobo to grant the permissions before our API gateway kills the request.

noliveleger (changes requested):
Unfortunately, there is something wrong with the PR. @RuthShryock 
When assigning a permission to user on a collection, it does not propagate to children of the collection. 

But if I use the shell, it does work. My first theory could be some methods cached with `@cache_request` which does not return an accurate state of the DB. I did not test it, but that the first explanation I have why it does work in shell but not through the API.

noliveleger:
> We're very excited about this PR because what we want to do (use APIs to create shared questions in the library and then automatically grant users access based on their country) is blocked by the inability of Kobo to grant the permissions before our API gateway kills the request.

Hello @jdimeo sorry to hear that, if you are admin of your server, you may achieve what you want with `ProjectView` instead.

rgraudin:
> Hello @jdimeo sorry to hear that, if you are admin of your server, you may achieve what you want with `ProjectView` instead.

Hi @noliveleger, I'm in John's org. We are definitely willing to use a workaround if there is one. I have tested creating a `ProjectView` but could not quite figure out how to make use of it. First of all, I can't figure out how to add projects to a project view in the console or using the API by declaring the project view as the parent. It would be helpful if we could share content this way. Of course, I'm not sure Project Views would help us get all the way there because we have content divided in `Collections` that are only allowed in the `Library` section, not the `Projects` section. It would be helpful if a project view could be used like a referenceable project 'group' so we can share content with the members of the `ProjectView` but that may not be implemented.

Appreciate any help or even somewhere to look documenting Project Views or other potentially helpful functionalities.

PR #4925, opened 2024-05-01
[TASK-659] [TASK-556] Add NLP support for background-audio questions

jnm:
## Description

Add support for transcripts, translations, and qualitative analysis of `background-audio` questions, and improve display of `background-audio` questions and groups in the data table.

## Notes

**UX/UI changes**

- Allow `background-audio` as a valid question type in NLP view
- Fix NLP view UI crash when opening a question inside a nested group
- Fix xml/labels switcher in the Display Settings modal when switching back to "Labels" option
- Rename Display Settings modal option from "XML names" to "XML values" to match nomenclature in Data Table and Data Downloads
- Change display for "audio" and "background audio" from xml values to proper translation during the first step of creating a transcript/translation
- Remove extra empty space caused by `NlpUsageLimitBlockModal` component in transcript/translation steps

**Front-end code changes**
- Treat `background-audio` as a regular question type
- Add tests for `tableUtils.getColumnLabel`
- Improve `tableUtils.getColumnLabel` to handle `background-audio` and groups properly
- Add more tests for `getSupplementalPathParts`
- `getSupplementalPathParts` now returns `sourceRowPath` in addition to the name
- Remove unused icon `background-rec.svg`

**Back-end code changes**
- Add back-end support for transcripts, translations, and qualitative analysis of `background-audio` questions

## Testing

Since Enketo doesn't collect `background-audio`, you'll need your development environment (or staging server) to be reachable from an **Android device** using KoboCollect or ODK Collect.

### Testing locally with kobo-compose

For this review, I used [**kobo-compose**](https://github.com/kobotoolbox/kobo-compose), since without any modification it accepts connections on `0.0.0.0`. I could then test locally without setting up any DNS or domain name settings. [^1]

- Connect both **dev machine** and **Android device** to the **same local area network**.
- In **KoboCollect**, enter the the dev machine's local IP address and OpenRosa port on that network.
	- Example: `http://192.168.0.123:8080`

[^1]: It's possible to do this with kobo-install too, but some modification is required. [Internal discussion](https://chat.kobotoolbox.org/#narrow/stream/4-Kobo-Dev/topic/Tools.2C.20tips.20and.20tricks/near/478511)

## Additional PR context

This PR started out as a back-end support for NLP background-audio questions by @jnm and was subsequently picked up by @magicznyleszek for additional frontend enhancements. Original description:

[...]

Issue #4407, opened 2023-04-10
Background audio is not handled properly by the NLP feature


magicznyleszek:
## Description

You can't preview audio responses from background audio sources in Data Table, and the NLP features are not working for it either.

## Steps to Reproduce

If the form has `Background audio` feature on, it can be seen in **DATA>Table** view. See the image below:

![image1](https://user-images.githubusercontent.com/2521888/230879450-01715d68-9d0a-4638-b7f4-099dea138750.png)

But if you click the play button, you will see this …

![image2](https://user-images.githubusercontent.com/2521888/230879473-3188691a-eb21-4092-8eb1-65240a01491a.png)

If you click the Open button, you will get this screen …

![image3](https://user-images.githubusercontent.com/2521888/230879499-e4959c93-f10b-4eff-a3f0-2577a92dca6f.png)

But if you check the background audio under the **DATA>Downloads>Media Attachments (ZIP)**, it’s safe there. You can open it and play the background audio without any issues.

![image4](https://user-images.githubusercontent.com/2521888/230879523-ab470a3b-fa22-476b-a1af-e0b420d19057.png)

## Expected behavior 

Background audio should be handled properly by NLP feature.

PR #5091, opened 2024-09-03
Do not use a dash as a group delimiter anymore

noliveleger:
## Description

Replace `qpath` with XPath

## Related issues

Part of #4407
Part of #4925
Related to kobotoolbox/formpack#324


kobo/apps/subsequences/utils/deprecation.py
```diff
@@ -0,0 +1,114 @@
...
+            if xpath == old_xpath:
+                continue
+
+            del content[old_xpath]
+            content[xpath] = values
+            changed = True
+
+    if changed:
+        submission_extra.content = content
+        # TODO Should we save submission_extra?
```
Guitlle, line 93: Maybe it's better to pass the content dictionary and return a sanitized dictionary, instead of receiving the ORM instance, It's be simpler and you would avoid mutating the instance in the function

kobo/apps/subsequences/utils/deprecation.py
```diff
@@ -0,0 +1,114 @@
...
+        # Old `qpath` should not contain "/", but could contain "-".
+        # If the question does not belong to a group but does contain "-",
+        # it will enter this condition - which is not a problem except extra
+        # CPU usage for nothing.
+        if '-' in xpath and '/' not in xpath:
+            xpath = qpath_to_xpath(xpath, asset)
+            rest.insert(0, xpath)
+            asset.known_cols[idx] = ':'.join(rest)
+
+    # TODO Should we save asset.known_cols if it has changed?
```
Guitlle, line 70: Same as below, pass the dictionary and return the new dictionary, instead of passing and mutating the asset instance

kobo/apps/subsequences/utils/deprecation.py
```diff
@@ -0,0 +1,114 @@
...
+    """
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self._model_instance = None
+
+    def to_representation(self, value):
+        self._model_instance.validate_advanced_features()
+        return value
+
+    def get_attribute(self, instance):
```
Guitlle, line 112: Is it guaranteed that this method will be called to set _model_instance, before calling to_representation?

kobo/apps/subsequences/utils/deprecation.py
```diff
@@ -0,0 +1,114 @@
...
+            schema=ADVANCED_FEATURES_PARAMS_SCHEMA,
+        )
+    except jsonschema.exceptions.ValidationError as e:
+        if "'qpath' was unexpected" not in str(e):
+            raise
+
+        qual_survey_orig = asset.advanced_features['qual']['qual_survey']
+        qual_survey_iter = deepcopy(qual_survey_orig)
+        for idx, qual_q in enumerate(qual_survey_iter):
+            qpath = qual_survey_orig[idx]['qpath']
```
Guitlle, line 27: Should check if 'qpath' key exists and continue if not. If it was sanitized before, qpath won't be present, right?

Guitlle (changes requested):

kpi/models/asset.py
```diff
@@ -1142,10 +1154,16 @@ def update_languages(self, children=None):
     def validate_advanced_features(self):
         if self.advanced_features is None:
             self.advanced_features = {}
-        jsonschema_validate(
-            instance=self.advanced_features,
-            schema=ADVANCED_FEATURES_PARAMS_SCHEMA,
-        )
+
+        # TODO uncomment the 4 lines below…
```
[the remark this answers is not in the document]
noliveleger: That's a good question. The migration would take too long and there is already a management command (or a script) for that. Let's see with the rest of the team what they think.

kobo/apps/subsequences/utils/deprecation.py
noliveleger: Indeed. When the serializer calls its own method `to_representation()`, it calls `field.get_attribute()` then `field.to_representation()`. 

See https://github.com/encode/django-rest-framework/blob/master/rest_framework/serializers.py#L516-L538

kobo/apps/subsequences/utils/deprecation.py
noliveleger: Indeed, `jsonschema.validate()` does raise an error if `qpath` is present because it does not belong to `ADVANCED_FEATURES_PARAMS_SCHEMA`. 

Line 24 cannot be reached if `qpath` is not present.

kobo/apps/subsequences/utils/deprecation.py
noliveleger: Unfortunately, we need the asset object in `qpath_to_xpath()`. It would be a little bit redundant to pass `asset.known_cols` and `asset` to `sanitize_known_columns()`. 

What do you think about something like that? 

```python
def get_sanitize_known_columns(asset: 'Asset') -> dict | None:
    known_cols = deepcopy(asset.known_cols)
    for idx, known_column in enumerate(known_cols):
        ...
    return knows_cols
```

Guitlle: yeah, it works

Guitlle (approved):

PR #5092, opened 2024-09-03
Refactor access logs 

rgraber:
## Checklist

3. [x] Ensure the tests pass

## Description

Developer-facing changes only.

## Notes

Make AccessLogs a proxy model of AuditLogs to eliminate duplicate code and reduce the possibility of an access log being saved with bad data. All Access Logs should have the same log_type, model_name, action, and app_label, and the object_id should always be set to the id of the user being logged in. To that end, overwrite the `create()` method on the AccessLog class to always set those fields and ignore any attempt to set them otherwise. Also, give it an updated queryset to only get logs with log_type='access.'


kobo/apps/audit_log/models.py
```diff
@@ -94,12 +95,54 @@ def save(
...
+        log_type = kwargs.pop('log_type', None)
+        if log_type is not None:
+            logging.warning(
+                f'Ignoring attempt to set {log_type=} on access log'
+            )
+        user = kwargs.pop('user')
+        return super().create(
+            # set the fields that are always the same for access logs, pass along the rest to the original constructor
+            app_label=ACCESS_LOG_KOBO_AUTH_APP_LABEL,
+            model_name=User.__qualname__,
```
noliveleger, line 127: `User.__qualname__` returns `User`, but model_name should be in lowercase. 
At the first glance, the idea was to be consistent with `ContentType` model and `Meta` class attributes. 

```python
from django.contrib.contenttypes.models import ContentType

c = ContentType.objects.get_for_model(User)
c.app_label  # kobo_auth 
c.model  # user <-- lowercase

User._meta.app_label  # kobo_auth 
User._meta.model_name  # user
```

Thinking about that, do we need really need a constant for `ACCESS_LOG_KOBO_AUTH_APP_LABEL = 'kobo_auth'`  since we can get it directly with the `Meta` class attributes of User class ?

noliveleger (changes requested):
Except my comments on the lowercase model name, I really like the idea of the proxy model.

kobo/apps/audit_log/models.py
rgraber: Should I add a migration for old audit logs to use lowercase?

noliveleger (approved):

PR #5094, opened 2024-09-04
Group submission access logs

rgraber:
## Checklist

3. [x] Ensure the tests pass

## Description
Group submission access logs together to make it easier to view and understand large collections of access logs.

## Notes

Model changes:
1. Adds a submission_group field to the AuditLog model, used for grouping submission logs together.
2. Adds 2 new proxy models based on AuditLogs: SubmissionAccessLog and SubmissionGroup. These proxy models are in charge of creating instances with the correct auth_type. 
3. The SubmissionGroup proxy model is also responsible for returning only those access logs with the 'submission-group' auth type when queried. This was not necessary for the SubmissionAccessLog so was left off.
4. The SubmissionAccessLog also has a method for adding itself to an existing SubmissionGroup or creating a new one

Signal changes:
Adds 2 new signals. 
1. SubmissionAccessLog: On saving a new SubmissionAccessLog, we look for the most recent submission group for the same user if it exists. If that submission group contains a log whose creation date is within a specified number of minutes of the creation time of the calling log, we add the calling log to the existing group. Otherwise, create a new one.
2. SubmissionGroups: On saving a new SubmissionGroup, we assign it to itself (ie set submission_group=self). This makes the grouping query easier. 

Api changes:
1. Replaces the queryset for the AccessLog endpoints with one that uses `fields` and `annotate` to group access logs by the `submission_group`, which should only be set on SubmissionAccessLogs and SubmissionGroups. If there is no `submission_group`, it groups by the id instead, so all other logs are effectively grouped with themselves.
* In order to improve query performance, this also removes fields that will be the same on all AccessLogs (`model_type`,`app_name`,`action`,and `log_type`). Removing these fields means the grouping clause doesn't have to check their equality each time.
2. Adds a new serializer to deal with the new query, which returns `dict`s instead of `AuditLog`s.
2. Sets the default search field on the `/all` endpoint to be the username since that's the only really searchable field the requirements call for (besides date, which doesn't make sense for a text search)

Replaces https://github.com/kobotoolbox/kpi/pull/5080
Depends on https://github.com/kobotoolbox/kpi/pull/5092


kobo/apps/audit_log/tests/test_signals.py
```diff
@@ -137,3 +130,65 @@ def test_loginas(self):
...
+    def test_new_group_for_new_submission_if_existing_groups_too_old(self):
+        user = User.objects.get(username='someuser')
+        ten_minutes_ago = timezone.now() - timedelta(minutes=10)
+        old_submission = SubmissionAccessLog.objects.create(
+            user=user, date_created=ten_minutes_ago
+        )
+
+        # make sure that created a new group
+        group_query = SubmissionGroup.objects.filter(user=user)
+        self.assertEqual(group_query.count(), 1)
```
noliveleger, line 186: Maybe we should assert it equals 0 before submission access log is created?

kobo/apps/audit_log/tests/test_signals.py
```diff
@@ -137,3 +130,65 @@ def test_loginas(self):
...
+        old_group = group_query.first()
+        self.assertEqual(old_submission.submission_group.id, old_group.id)
+
+        # new submission
+        new_submission = SubmissionAccessLog.objects.create(user=user)
+        self.assertEqual(group_query.count(), 2)
+        new_group = group_query.order_by('-date_created').first()
+        self.assertEqual(new_submission.submission_group.id, new_group.id)
```
noliveleger, line 194: even if `group_query.count() == 2`, I think it could be great to validate ` new_group.id != old_group.id` just in case.

kobo/apps/audit_log/views.py
```diff
@@ -88,108 +90,85 @@ class AuditLogViewSet(mixins.ListModelMixin, viewsets.GenericViewSet):
...
-    >                    "action": "AUTH",
-    >                    "metadata": {
-    >                        "source": "Firefox (Ubuntu)",
-    >                        "auth_type": "Digest",
-    >                        "ip_address": "172.18.0.6"
-    >                   },
-    >                    "date_created": "2024-08-19T16:48:58Z",
-    >                    "log_type": "access"
-    >                },
-    >                ...
```
noliveleger, line 128: Why did you remove the example?

noliveleger:
While I was testing the logs with `mockobo`, (which uses `/api/v1/submissions` with auth-token).  I have noticed that it only logs one event even if I called the endpoints 40 times in less than a minute.

kobo/apps/audit_log/tests/test_models.py
```diff
@@ -192,3 +202,120 @@ def test_create_auth_log_with_extra_metadata(
...
+        user2 = User.objects.get(username='anotheruser')
+        with skip_all_signals():
+            group: SubmissionGroup = SubmissionGroup.objects.create(user=user1)
+            log: SubmissionAccessLog = SubmissionAccessLog.objects.create(
+                user=user2
+            )
+            log.add_to_existing_submission_group(group)
+            log.save()
+        # make sure we reported an error
+        patched_error.assert_called_once()
```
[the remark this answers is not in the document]
rgraber: This should never happen in a real scenario. It's for future-proofing in case other code needs to use this method. I wasn't sure not adding to a group would be a big enough problem to raise an exception but I'm pretty ambivalent about it

kobo/apps/audit_log/models.py
```diff
@@ -204,4 +211,63 @@ def create_from_request(
...
+
+    def add_to_existing_submission_group(
+        self, submission_group: SubmissionGroup
+    ):
+        if self.user_uid != submission_group.user_uid:
+            logging.error(
+                f'Cannot add submission log with user {self.user_uid} to group with user {submission_group.user_uid}'
+            )
+            return
+        self.submission_group = submission_group
```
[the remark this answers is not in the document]
rgraber: I go back and forth on who should be responsible for saving the model. There's a reasonable chance any future code that calls this method would make multiple changes and it would be a waste of db work to save every time. I don't feel strongly about it though.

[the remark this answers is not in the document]
rgraber: On further reflection, after a refactor, ended up deleting this method all together

kobo/apps/audit_log/views.py
rgraber: Accident from moving things around

PR #5104, opened 2024-09-11
TASK 1027 - Conditional label for 'Delete Project and Data' button

pauloamorimbr:
## Checklist

3. [ ] Ensure the tests pass
5. [ ] Write a title and, if necessary, a description of your work suitable for publishing in our [release notes](https://community.kobotoolbox.org/tag/release-notes)

## Description

This PR adds a conditional label for the 'Delete Project and Data' button, where it will only show 'Delete Project' when there's no submissions to the project.

No data:
image

With data:
image
## Notes

I opted to isolate the label in a separated variable instead of adding the conditional logic into the button for better readability.

### DIFF
--- a/jsapp/js/components/modalForms/projectSettings.es6
+++ b/jsapp/js/components/modalForms/projectSettings.es6
@@ -873,6 +873,10 @@ class ProjectSettings extends React.Component {
     const isSelfOwned = assetUtils.isSelfOwned(this.state.formAsset);
     const descriptionField = envStore.data.getProjectMetadataField('description');
 
+    const deleteButtonLabel = this.state.formAsset.deployment__submission_count > 0 ?
+      t('Delete Project and Data') :
+      t('Delete Project');
+
     return (
       <bem.FormModal__form
         onSubmit={this.handleSubmit}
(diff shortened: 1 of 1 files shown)


jsapp/js/components/modalForms/projectSettings.es6
```diff
@@ -873,6 +873,10 @@ class ProjectSettings extends React.Component {
     const isSelfOwned = assetUtils.isSelfOwned(this.state.formAsset);
     const descriptionField = envStore.data.getProjectMetadataField('description');
 
+    const deleteButtonLabel = this.state.formAsset.deployment__submission_count > 0 ?
```
p2edwards, line 876: I wanted to check this real quick, since it's possible for `deployment__submission_count` to be null when viewed by certain users (see [kpi#4739](https://github.com/kobotoolbox/kpi/pull/4739)). But thanks to the magic of JS, `null > 0` is false, so no problem here.

The actual delete button below has `isSelfOwned && this.props.context === PROJECT_SETTINGS_CONTEXTS.EXISTING` guarding it.

p2edwards (approved):
:+1: Tested and it works!

jsapp/js/components/modalForms/projectSettings.es6
pauloamorimbr: Thanks. I changed it and moved the condition to the element. Since it may not be rendered it does make more sense to leave the conditional label inside it as well. 🙇🏻

PR #5105, opened 2024-09-11
Fix bad performance introduced in `2.024.25`

noliveleger:
## Description

Fix 504 errors that are returned while performing random actions in the app.

## Notes

One of the main cause is - with the merge of Kobocat base code in `openrosa` django app in KPI - there is only one app left that manages the disclaimer in Enketo. We need to get the Asset object from the XForm object. 
Both queries that are used for this purpose do not use indexes and are very slow.


kpi/migrations/0057_alter_assetfile_content_alter_exporttask_result_and_more.py
```diff
@@ -0,0 +1,55 @@
...
+    else:
+        operations = [
+            migrations.AddIndex(
+                model_name='asset',
+                index=django.contrib.postgres.indexes.BTreeIndex(
+                    models.F('_deployment_data__formid'),
+                    name='deployment_data__formid_idx',
+                ),
+            ),
+        ]
```
jnm, line 55: The name of this file, `0057_alter_assetfile_content_alter_exporttask_result_and_more.py`, is strange given the operations :thinking:

kobo/apps/openrosa/apps/logger/migrations/0035_alter_attachment_media_file_and_more.py
```diff
@@ -0,0 +1,56 @@
...
+            )
+        ]
+    else:
+        operations = [
+            migrations.AlterField(
+                model_name='xform',
+                name='kpi_asset_uid',
+                field=models.CharField(db_index=True, max_length=32, null=True),
+            ),
+        ]
```
jnm, line 56: The name of this file, `0035_alter_attachment_media_file_and_more.py`, is strange given the operations :thinking:

jnm (changes requested):
No issues with the content, just questions about the filenames of the migrations

kpi/migrations/0057_alter_assetfile_content_alter_exporttask_result_and_more.py
noliveleger: me fixing conflicts = Fail 🤦

PR #5111, opened 2024-09-16
Fix error when deleting a user from admin interface

noliveleger:
## Description

Fix error `relation "logger_xform" does not exist ... ` when user is being deleted

## Notes

This PR uses monkey patching to replace a Django utility to remove models which have a relationship with a model but from an other database. E.g. When user1 - User object - is deleted from KPI database, XForm related models should not be deleted in cascade within this transaction. They should be only when user1 is deleted from KoboCAT database.

A new migration has been created to delete tables not used anymore. It has been added to `api` app in purpose, to avoid other conflicts with migrations of the next release.


rgraber (commented):
One big question: is there any way to construct actual reverse sqls so we get the constraints/table back? I worry about having no way to reconstruct the tables if we need to revert the migration for some reason.

noliveleger:
> One big question: is there any way to construct actual reverse sqls so we get the constraints/table back? I worry about having no way to reconstruct the tables if we need to revert the migration for some reason.

@rgraber it is a really good question, but, IMO, in that case it is not worth it because the models have been removed from current release. In the previous release, they were a copy of KPI tables. If we had to revert to the previous release, we would apply migrations from `auth` and `django_digest` to restore the tables and their constraints. Then we would just need to copy the data from KPI (but it was auto synchronizing data).

kpi/utils/monkey_patching.py
```diff
@@ -23,4 +27,42 @@ def create_permissions(app_config, using=DEFAULT_DB_ALIAS, **kwargs):
...
+    """
+
+    db_connection = router.db_for_write(opts.model)
+
+    return (
+        f
+        for f in opts.get_fields(include_hidden=True)
+        if f.auto_created
+        and not f.concrete
+        and (f.one_to_one or f.one_to_many)
```
rgraber, line 49: Can you add a comment specifying that here and below is the actual new stuff, and the bit above comes directly from the original source code?

kobo/apps/openrosa/apps/api/migrations/0006_delete_digest_and_auth_token_table.py
```diff
@@ -0,0 +1,79 @@
...
+    if settings.TESTING or settings.SKIP_HEAVY_MIGRATIONS:
+        # Skip this migration if running in test environment or because we want
+        # to voluntarily skip it.
+        return []
+
+    tables = KC_TABLES_TO_DELETE
+    operations = []
+
+    sql = """
+        SELECT con.conname
```
rgraber, line 24: More comments about what this does would be helpful

rgraber (changes requested):
Requests for more comments

rgraber:
> > One big question: is there any way to construct actual reverse sqls so we get the constraints/table back? I worry about having no way to reconstruct the tables if we need to revert the migration for some reason.
> 
> @rgraber it is a really good question, but, IMO, in that case it is not worth it because the models have been removed from current release. In the previous release, they were a copy of KPI tables. If we had to revert to the previous release, we would apply migrations from `auth` and `django_digest` to restore the tables and their constraints. Then we would just need to copy the data from KPI (but it was auto synchronizing data).

That's still a little scary but as long as we have some kind of reversion plan we can go with it.

rgraber (approved):
LGTM

PR #5120, opened 2024-09-19
[TASK-1046] Introduce unit tests for UI components

pauloamorimbr:
## Checklist

3. [x] Ensure the tests pass

## Description

Create the base configuration and samples for adding unit tests with Jest.
Create simple tests for the `Button` and `KoboSelect3` components as samples.

## Notes

- Some new packages were added for `jest` to work
- Configuration for jest was added into `/jsapp/jest`
- 2 new scripts were created on `package.json`:
    - `npm run jest`
    - `npm run jest-watch` (Very useful when writing tests)
    -  Both scripts can filter test files as in: `npm run jest-watch -- Button`
- To avoid confusion with mocha `.tests.*` file, we're using the `.spec.*` extension for tests.
- A test for the `Button` component was added checking for its interaction when enabled and disabled
- A test for the `KoboSelect3` component was added checking for its mouse and keyboard interaction


jsapp/js/components/common/button.spec.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+    // Mock
+    const handleClickFunction = jest.fn();
+
+    // Render
+    render(
+      <Button
+        type={'primary'}
+        size={'l'}
+        label='Button Label'
```
Akuukis, line 19: Do we have a linter rule to convert `{'l'}` into `'l'`?

jsapp/js/components/special/koboAccessibleSelect.spec.tsx
```diff
@@ -0,0 +1,150 @@
...
+    // List should be collapsed after selection
+    expect(list.dataset.expanded).toBe('false');
+  });
+
+  it('Should respond to keyboard interaction', async () => {
+    // Mock
+    const onChangeMock = jest.fn();
+
+    // Render
+    render(<Wrapper onChange={onChangeMock} />);
```
Akuukis, line 86: Most of comments are useful, but IMHO this one is excessive because it adds no information at cost of noise/lines - it's very easy to read the next line.

jsapp/js/components/common/button.spec.tsx
```diff
@@ -4,55 +4,65 @@ import {render, screen} from '@testing-library/react';
...
 
-    // Mock
-    const handleClickFunction = jest.fn();
+// Mock
+const handleClickFunction = jest.fn();
 
-    // Render
+
+describe('Enabled button', () => {
+  beforeEach(() => {
```
Akuukis, line 14: let's move `handleClickFunction` into describe block to use as narrow scope as possible?

jsapp/js/components/common/button.spec.tsx
pauloamorimbr: It was my mistake to leave it like that. Fixed now!

jsapp/js/components/common/button.spec.tsx
pauloamorimbr: Will do! 
I did it like that to avoid duplication, since the same function is used on both `describes` there. 
Since it's a simple function we can redeclare it, no problems, but in the cases where we can have a more complex mocked implementation it'd be good to have it declared and implemented only once!

package.json
```diff
@@ -201,7 +211,9 @@
     "copy-fonts": "python3 ./scripts/copy_fonts.py && npm run generate-icons",
     "hint": "node ./scripts/hints.js",
     "storybook": "storybook dev -p 6006 --no-open",
-    "build-storybook": "storybook build"
+    "build-storybook": "storybook build",
+    "jest": "jest --config ./jsapp/jest/jest.config.ts",
```
p2edwards, line 215: We could add this to our GitHub CI (.github/workflows/npm-test.yml)
```yml
      # Run Jest component tests
      - name: Run Jest component tests
      - run: npx jest --config ./jsapp/jest/jest.config.ts --ci
```
Here, or in a separate PR.

jsapp/js/components/special/koboAccessibleSelect.spec.tsx
```diff
@@ -0,0 +1,136 @@
...
+import KoboSelect3 from './koboAccessibleSelect';
+import {useState} from 'react';
+
+const options: KoboSelectOption[] = [
+  {value: '1', label: 'Apple'},
+  {value: '2', label: 'Banana'},
+  {value: '3', label: 'Avocado'},
+];
+
+// A wrapper is needed for the component to retain value changes
```
p2edwards, line 14: 📓 Good point — many of our Kobo components are like this — 'controlled'-only. It'd be possible to make them more flexible.

- Cool article: [Creating React Components that can be Controlled and Uncontrolled](https://maxschmitt.me/posts/react-components-controlled-uncontrolled)

'Course, then there would be more to test :sweat_smile:

jsapp/js/components/special/koboAccessibleSelect.spec.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+      // No item selected
+      expect(triggerLabel).toHaveTextContent('Select…');
+
+      // Type 'b' to select Banana
+      fireEvent.keyDown(trigger, {key: 'b'});
+      expect(onChangeMock).lastCalledWith(options[1].value);
+      expect(triggerLabel).toHaveTextContent(options[1].label);
+  });
+});
```
p2edwards, line 136: Awesome! Great demo.

Seeing this makes me think of a bunch of other test cases. :) 

Looking forward to using jest tests to get more safety for stuff we'd test manually, including complicated components like the data table, and query logic if we need it (with [msw](https://github.com/testing-library/react-testing-library#:~:text=We%20recommend%20using%20Mock%20Service%20Worker%20library%20to%20declaratively%20mock%20API%20communication%20in%20your%20tests%20instead%20of%20stubbing%20window.fetch%2C%20or%20relying%20on%20third%2Dparty%20adapters.))

p2edwards (changes requested):

package.json
Akuukis: Good catch, I think that's in the scope of adding jest to project, because normally all tests are always run on CI.

package.json
```diff
@@ -147,6 +155,7 @@
     "prettier-eslint": "^16.3.0",
     "react-docgen-typescript-plugin": "^1.0.8",
     "react-refresh": "^0.14.2",
+    "react-test-renderer": "^18.3.1",
```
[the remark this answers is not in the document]
pauloamorimbr: Nice catch! 🎣 
I had to add them at some moment in the process but then I changed it up and forgot about it. Cleaned.

package.json
pauloamorimbr: I was thinking about doing this in a different PR, after some more tests were added. But I believe there's no problem doing so here anyways. 👍🏻

jsapp/js/components/special/koboAccessibleSelect.spec.tsx
pauloamorimbr: It's a common practice I found for single component testing, since most of the components will depend on external context anyways. So I really don't think it's the case of making them 'uncontrolled' just for the sake of tests. We only need to be careful to not create biased tests that could influence the component behavior. (but yeah, cool article! 😄)

pauloamorimbr:
Cool. Checks added the new tests already!

PR #5128, opened 2024-09-27
Regenerate snapshots when downloading XForm XML

jnm:
## Checklist

3. [x] Ensure the tests pass

## Description

Invokes pyxform every time "Download XML" is clicked, which avoids serving stale XForm XML in certain situations. For example, prior to this work, if a form had *not* been changed since its XForm XML was last generated, but the behavior of pyxform *had* changed, clicking "Download XML" would have ignored the pyxform changes.

Please note that the XForm XML obtained from "Download XML" is *not* identical to the XForm XML generated when deploying a form. Deploying or redeploying a form was *not* affected by this stale XML issue, even prior to this change.

## Notes

I did not add a unit test or attempt to mock the response by pyxform to simulate changes to its behavior. The reviewer may make the determination about the sufficiency of my approach.

Internal discussion: https://chat.kobotoolbox.org/#narrow/stream/4-Kobo-Dev/topic/Pyxform.20updates/near/458832

### DIFF
--- a/kpi/renderers.py
+++ b/kpi/renderers.py
@@ -237,6 +237,8 @@ class XMLRenderer(DRFXMLRenderer):
         accepted_media_type=None,
         renderer_context=None,
         relationship=None,
+        relationship_args=None,
+        relationship_kwargs=None,
     ):
         if hasattr(renderer_context.get('view'), 'get_object'):
             obj = renderer_context.get('view').get_object()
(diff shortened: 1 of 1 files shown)


tiritea:
Curious, what is/was the reason behind needing to persist snapshots in the first place, given they're not the XML XForm definition that is going to get deployed in any case?

Are snapshots just being done (and basically cached) for Enketo-based preview functionality? If so, then we would seen to be in much the same situation with Preview: unless you happen to have done an "XML Download" on all your forms _after_ a pyxform change, the saved snapshot that Enketo renders will still be effectively out-of-date [and its not intuitive that "XML Download" would be a prerequisite for Preview...]

noliveleger (changes requested):
As per our discussion, let's add a minimal unit test for that 

https://chat.kobotoolbox.org/#narrow/stream/4-Kobo-Dev/topic/Pyxform.20updates/near/466261

PR #5131, opened 2024-09-28
NLP usage tracking and limits based on subscription plan and one time add ons

Guitlle:
## Description

This PR provides the logic for the NLP usage counts tracking and also limits enforcing. Limits are based on the subscription plan and one time add ons. Some refactoring was involved to organize the utilities, comments are welcome to improve the organization of the code.

## Notes

Changes on the following components:
* kobo/apps/organizations 
  * Added a function to get the organization for a user id
* kobo/apps/stripe
  * Logic for the PlanAddOn model to get the total counters for organization as well as modifying the increment function to receive an organization instead of an user.
* kobo/apps/trackers
  * Moved some general usage tracking utility functions to this app in order to organize things better, updated the NLP update counter function to use the new logic.
* kobo/apps/subsequences 
  * The google integrations have been modified to check that the remaining organization usage is enough to process the amount for a new usage request

## Related issues

Notion TASK-1014
Notion TASK-1004


kobo/apps/trackers/utils.py
```diff
@@ -45,15 +52,65 @@ def update_nlp_counter(
...
+    """
+    PlanAddOn = apps.get_model('stripe', 'PlanAddOn')  # noqa
+
+    plan_limit = get_organization_plan_limit(organization, usage_type)
+    usage = get_organization_usage(organization, usage_type)
+    addon_limit, addon_remaining = PlanAddOn.get_organization_totals(
+        organization,
+        usage_type,
+    )
+    remaining = addon_limit + plan_limit - usage
```
jamesrkiger, line 96: Shouldn't we be using addon_remaining here?

Guitlle: Yes,  that's right! 

Please validate this example case for me: User A on Organization X has used 1000 minutes for NLP transcription this month, a while ago they bought 1 add on with 2000 minutes that has 1000 remaining because they used 500 on the past month and 500 this month, their plan has 500 minutes per month. 

So if their usage is 1000, that means they went beyond their plan limit and therefore used the add ons. So the remaining usage should only include the add on remaining. But if their usage was 300, for instance, that would mean they have not passed their plan limit and therefore they still have 200 plan minutes left. And that should be summed to the add on remaining. 

Is this correct?

jamesrkiger: I believe so. Let me write it out in another way so we can make sure we are on the same page.

Organization X is on the community plan. They get 10 transcription minutes per month. In September they bought an addon for 1,000 transcription minutes. That month they used 510 minutes. Therefore, their remaining addon limit at the end of September was 500 minutes.

They use 10 minutes of transcription on October 1. This gets taken from their monthly limit, so the remaining addon limit stays at 500. On October 2, they use 200 minutes. Their addon remaining limit goes down to 300. On October 3, they use 300 minutes. Their addon remaining limit goes down to 0.

kobo/apps/subsequences/integrations/google/google_transcribe.py
```diff
@@ -74,6 +74,11 @@ def begin_google_operation(
         submission_uuid = self.submission.submission_uuid
         flac_content, duration = content
         total_seconds = int(duration.total_seconds())
+        # Check if the user's organization has available usage credits
+        org = Organization.get_from_user_id(self.user.pk)
+        if get_organization_remaining_usage(org, 'seconds') < total_seconds:
+            raise UsageLimitExceeded
```
jamesrkiger, line 80: I think we said we're going to leave blocking excess usage for later, right? Same question goes for google translate below.

Guitlle: Yes, right, I can revert these and bring them back later

kobo/apps/trackers/utils.py
Guitlle: Ok, good, the change I made for this handles this logic. It checks if there are remaining minutes in the plan and if there are it takes that into account,  if there are still more minutes to be used it takes them from the add ons.

kobo/apps/stripe/models.py
```diff
@@ -156,34 +158,69 @@ def make_add_ons_from_existing_charges():
...
+            total_remaining=Coalesce(
+                Sum(Cast(limit_field, output_field=IntegerField())),
+                0,
+            ),
+        )
+
+        return totals['total_usage_limit'], totals['total_remaining']
+
+    @staticmethod
+    def increment_add_ons_for_organization(
```
noliveleger, line 200: `increment` is confusing. We are subtracting amount from AddOn limit. 
Let's use another word. 

- decrement_add_ons_for_organization
- deduct_add_ons_for_organization
- subtract_add_on_minutes_for_organization

noliveleger: Let's change it everywhere we use the concept of `increment`ing an add-on.

kobo/apps/trackers/utils.py
```diff
@@ -45,15 +52,72 @@ def update_nlp_counter(
...
+
+@cache_for_request
+def get_organization_usage(organization: Organization, usage_type: UsageType) -> int:
+    """
+    Get the used amount for a given organization and usage type
+    """
+    usage_calc = ServiceUsageCalculator(
+        organization.owner.organization_user.user, organization
+    )
+    usage_calc._clear_cache()  # Do not use cached values
```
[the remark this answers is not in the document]
noliveleger: As per our discussion, we are going to modify `ServiceUsageCalculator` without initializing the cache. 
It means the cache could be stale for a few minutes, but it's ok.

noliveleger (changes requested):

kobo/apps/organizations/tests/test_organizations_model.py
```diff
@@ -0,0 +1,22 @@
...
+        self.anotheruser = User.objects.get(username='anotheruser')
+        self.organization = baker.make(Organization, id='org_abcd1234')
+        self.organization.add_user(user=self.user, is_admin=True)
+
+    def test_get_from_user_id(self):
+        org = Organization.get_from_user_id(self.user.pk)
+        assert org.pk == self.organization.pk
+
+        org = Organization.get_from_user_id(self.anotheruser.pk)
+        assert org is None
```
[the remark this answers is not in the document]
Guitlle: Ok, so for now it stay as is

kobo/apps/trackers/utils.py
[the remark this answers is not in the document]
Guitlle: Yes, added the disable_cache init parameter and a unit test for that too

kobo/apps/stripe/models.py
Guitlle: I like "deduct" more.

kobo/apps/stripe/constants.py
```diff
@@ -22,7 +22,7 @@
 ORGANIZATION_USAGE_MAX_CACHE_AGE = timedelta(minutes=15)
 
 USAGE_LIMIT_MAP = {
-    'character': 'mt_character',
+    'character': 'mt_characters',
```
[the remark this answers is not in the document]
Guitlle: This won't be a problem on existing data. The API for usage already uses mt_characters in plural, the model property also uses plural and the google translate integration uses plural as well. The problem is that this USAGE_LIMIT_MAP had not been used yet for translation feature until now, with the add ons works, so it had the wrong constant

kobo/apps/stripe/models.py
```diff
@@ -156,34 +158,69 @@ def make_add_ons_from_existing_charges():
...
         ).order_by(metadata_key)
         remaining = amount
         for add_on in add_ons.iterator():
-            if not add_on.organization.is_organization_over_plan_limit(add_on_type):
-                return remaining
             if add_on.is_available():
-                remaining -= add_on.increment(limit_type=usage_type, amount_used=remaining)
+                remaining -= add_on.increment(
+                    limit_type=limit_key, amount_used=remaining
+                )
```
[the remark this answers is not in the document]
Guitlle: I made a special django orm utility to make this better. Let me know what you think.

kobo/apps/stripe/constants.py
```diff
@@ -22,14 +22,14 @@
...
-    'character': 'mt_characters',
+    'characters': 'mt_characters',
     'seconds': 'asr_seconds',
     'storage': 'storage_bytes',
     'submission': 'submission',
 }
 
 USAGE_LIMIT_MAP_STRIPE = {
-    'character': 'nlp_character',
+    'characters': 'nlp_character',
```
noliveleger, line 32: You're going to find me annoying but, is it possible to also use `nlp_character`**s** (plural), to match other strings?

noliveleger: This will be covered by https://www.notion.so/kobotoolbox/Switch-to-mt_characters-and-asr_minutes-in-code-stripe-2efeaf1173404bca8595e675440abb84

PR #5135, opened 2024-09-30
Fix project transfer failure in some circumstances

noliveleger:
## Description

Fix an error when project is not transferred entirely to the new owner and data is missing under the account of the new owner. 
Project returns the correct number of submissions, but the table view only returns a subset of it.

## Notes 
A new management command has been added to resume broken transfers done after `2.024.25` release has been deployed: `resume_failed_transfers_2_024_25_fix` 

The main problem was a race condition between uWSGI process and Celery process. The PR ensures that Celery tasks are triggered IFF data in transaction is saved in DB, thanks to `transaction.on_commit` hook. 

Moreover, `media_file` field of `Attachment` model is changed to `ExtendedField` like it is in its shadow model counterpart (i.e.: `KobocatAttachment`).  It provides the method `move()` which is required by transfer asynchronous tasks. See https://github.com/kobotoolbox/kpi/blob/834ec518a9487dcfacd51e2476f1ad6fafda041c/kpi/fields/file.py for more details.

### DIFF
--- a/kobo/apps/openrosa/apps/logger/models/attachment.py
+++ b/kobo/apps/openrosa/apps/logger/models/attachment.py
@@ -1,4 +1,3 @@
-# coding: utf-8
 import mimetypes
 import os
 
@@ -10,6 +9,7 @@ from kobo.apps.openrosa.libs.utils.hash import get_hash
 from kpi.deployment_backends.kc_access.storage import (
     default_kobocat_storage as default_storage,
 )
+from kpi.fields.file import ExtendedFileField
 from .instance import Instance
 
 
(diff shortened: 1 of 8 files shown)


kobo/apps/openrosa/apps/logger/models/attachment.py
```diff
@@ -1,4 +1,3 @@
-# coding: utf-8
```
rgraber, line 1: Why was this removed?

rgraber (approved):
LGTM. One non-blocking question

kobo/apps/openrosa/apps/logger/models/attachment.py
noliveleger: When I was adding the linter to CI, I came across [this rule](https://docs.astral.sh/ruff/rules/utf8-encoding-declaration/). I did not want to enforce it in the CI but I decided to remove them manually when I was touching a file.

PR #5137, opened 2024-10-01
feat(OpenRosa): support XLSForm `name` setting TASK-1074

rajpatel24:
## Checklist

3. [x] Ensure the tests pass

## Description

This is an advanced feature for unusual cases where it's necessary to [override the root XML node name](https://xlsform.org/en/#specify-xforms-root-node-name) used in submissions. This may be used for an upcoming feature to allow customizable confirmation messages in Enketo.

## Notes

- Implemented logic to check and set the name attribute from XLSForms in the survey object.
- The existing behavior where name was not utilized has been modified to honor the name setting, defaulting to the id_string if name is not provided or is invalid.

## Related issues

Part of #5112


kobo/apps/openrosa/apps/viewer/models/data_dictionary.py
```diff
@@ -157,9 +157,8 @@ def add_instances(self):
     def save(self, *args, **kwargs):
         if self.xls:
             survey = create_survey_from_xls(self.xls)
-            survey.update({
-                'name': survey.id_string,
-            })
+            if not survey.name or survey.name == 'None':
```
jnm, line 160: Hi @rajpatel24, thanks for working on this! I'm curious: were there conditions where the string `None` gets passed in as the `name`? It seems a bit surprising to me

jnm (changes requested):

jnm (changes requested):
Could we have a unit test for this? I think it would ideally:
1. Create an `Asset` with the `name` setting (this can be in the `Asset.content`; we do not need to mess around with importing an XLSX file)
2. Deploy that `Asset`
3. Inspect the resulting `XForm.xml` to verify that the first child of the [primary instance](https://getodk.github.io/xforms-spec/#primary-instance) (the first `<instance>`) has a node name that matches the `name` setting set in the `Asset.content`.

Please ping me on Zulip if you have any questions :pray:

kobo/apps/openrosa/apps/viewer/models/data_dictionary.py
rajpatel24: Hi @jnm, When the form lacked a 'name' setting, the name was being set as the string 'None'.

Previously, we ignored the name setting value, whether it was 'None' or an actual string, and always assigned survey.id_string as the default name.

Now, if a valid name string is present, we leave it unchanged. If the name setting is absent, we assign survey.id_string as the default name.

jnm: Sorry to hassle on this, but I wanted to know where the string `None` was coming from. It bothers me because someone could have their XLSForm `settings` sheet look like this:
| name |
| - |
| None|

It would be weird, but it'd be valid, and we wouldn't be able to tell the difference between a user-provided `None` and the one that gets generated when the `settings` sheet lacks a `name` altogether.

I did some digging and found that the `None` string is set by pyxform:
* It's a string because of this: https://github.com/XLSForm/pyxform/blob/9d20bfec5128f53ec0d555f7187f3bc4881f667e/pyxform/xls2json.py#L396-L397
* `None` itself comes from `default_name=None` in https://github.com/XLSForm/pyxform/blob/9d20bfec5128f53ec0d555f7187f3bc4881f667e/pyxform/builder.py#L327
* `default_name` gets overridden by the `name` value in the `settings` sheet of the XLSForm here: https://github.com/XLSForm/pyxform/blob/9d20bfec5128f53ec0d555f7187f3bc4881f667e/pyxform/xls2json.py#L469-L470

I see that even though `create_survey_from_xls()` has a handy `default_name` parameter, `create_survey_from_xls(self.xls, default_name=self.id_string)` won't work because our logic relies on `create_survey_from_xls()` pulling the `id_string` out of the XLS file in the first place. This is a historical artifact; nowadays, we're always the ones generating the XLS file internally¹, and [we always know the `id_string` in advance](https://github.com/kobotoolbox/kpi/blob/8564da5c63fdd1aa9a6e0b3335ec2fe08f81e492/kpi/deployment_backends/openrosa_backend.py#L129-L139).

Let's do a hack, though, instead of blowing up the scope of this PR. Please invent some constant that's unlikely ever to be a user-provided value, like `PLACEHOLDER_NAME = '__kobo_placeholder_name_value__'`, call `survey = create_survey_from_xls(self.xls, default_name=PLACEHOLDER_NAME)`, and then overwrite `survey.name` only if it's equal to `PLACEHOLDER_NAME`.

¹ Even when someone uploads their own XLSForm, we convert that to JSON, and then we convert it back to XLS again when deploying.

kpi/tests/test_assets.py
```diff
@@ -779,3 +781,85 @@ def test_anonymous_as_baseline_for_authenticated(self):
...
+                {
+                    'type': 'select_one',
+                    'label': 'q1',
+                    'select_from_list_name': 'iu0sl99',
+                },
+            ],
+            'choices': [
+                {'name': 'a11', 'label': ['a11'], 'list_name': 'iu0sl99'},
+                {'name': 'a3', 'label': ['a3'], 'list_name': 'iu0sl99'},
+            ],
```
jnm, line 805: Please change the `type` to something simple like `text`, and then you can eliminate `select_from_list_name` as well as `choices`. It'll just make the test easier to read since we're not testing any features related to `select` questions

kpi/tests/test_assets.py
```diff
@@ -779,3 +781,85 @@ def test_anonymous_as_baseline_for_authenticated(self):
...
+        # Get the deployed XForm and parse it
+        xform = XForm.objects.get(id_string=asset.uid)
+        parser = XFormInstanceParser(xform.xml, xform.data_dictionary())
+
+        # Access the first child element of the <instance> node
+        instance_node = parser.get_root_node().getElementsByTagName('instance')[0]
+        root_element = instance_node.firstChild
+
+        # Assert that the name setting matches the root node name
+        self.assertEqual(asset.content['settings']['name'], root_element.nodeName)
```
jnm, line 826: New tests can use the simple `assert` syntax. I prefer to see the hard-coded value in the same line; we're not generating the value `custom_root_node_name` dynamically, so it can be written here directly for clarity. Suggestion:
```python
assert root_element.nodeName == 'custom_root_node_name`
```

kpi/tests/test_assets.py
```diff
@@ -779,3 +781,85 @@ def test_anonymous_as_baseline_for_authenticated(self):
...
+                {
+                    'type': 'select_one',
+                    'label': 'q1',
+                    'select_from_list_name': 'iu0sl99',
+                },
+            ],
+            'choices': [
+                {'name': 'a11', 'label': ['a11'], 'list_name': 'iu0sl99'},
+                {'name': 'a3', 'label': ['a3'], 'list_name': 'iu0sl99'},
+            ],
```
jnm, line 844: Please see comment above about simplifying the survey by using a non-select type

kpi/tests/test_assets.py
```diff
@@ -779,3 +781,85 @@ def test_anonymous_as_baseline_for_authenticated(self):
...
+        instance_node = parser.get_root_node().getElementsByTagName('instance')[0]
+        root_element = instance_node.firstChild
+
+        # Assert that the name setting matches the root node name
+        self.assertEqual(asset.content['settings']['name'], root_element.nodeName)
+
+    def test_asset_without_name_setting(self):
+        """
+        Test if 'name' setting is not provided, the root node should fall back
+        to xform.id_string
```
jnm, line 831: Let's say "fall back to the asset UID"

kpi/tests/test_assets.py
```diff
@@ -779,3 +781,85 @@ def test_anonymous_as_baseline_for_authenticated(self):
...
+        # Get the deployed XForm and parse it
+        xform = XForm.objects.get(id_string=asset.uid)
+        parser = XFormInstanceParser(xform.xml, xform.data_dictionary())
+
+        # Access the first child element of the <instance> node
+        instance_node = parser.get_root_node().getElementsByTagName('instance')[0]
+        root_element = instance_node.firstChild
+
+        # Assert that the root node name is the xform.id_sting
+        self.assertEqual(xform.id_string, root_element.nodeName)
```
jnm, line 865: Let's actually assert that the `root_element.nodeName == asset.uid` instead of `xform.id_string`. That's the application behavior we want. It's true that `xform.id_string` ends up matching the asset UID, but that's more of a side-effect than the thing we primarily care about.

See also previous note about new `assert` syntax.

jnm (changes requested):

kpi/tests/test_assets.py
```diff
@@ -779,3 +781,75 @@ def test_anonymous_as_baseline_for_authenticated(self):
...
+        """
+        Test if 'name' setting is provided, it should match with the root node.
+        """
+        content = {
+            'survey': [
+                {
+                    'text': 'select_one',
+                    'label': 'q1'
+                },
+            ],
```
jnm, line 800: Interesting! I'm surprised the unit tests don't fail since there's no `type`. I'll look into it.

kpi/tests/test_assets.py
```diff
@@ -779,3 +781,75 @@ def test_anonymous_as_baseline_for_authenticated(self):
...
+        Test if 'name' setting is not provided, the root node should fall back
+        to asset UID
+        """
+        content = {
+            'survey': [
+                {
+                    'text': 'select_one',
+                    'label': 'q1',
+                },
+            ],
```
jnm, line 834: As above

kpi/tests/test_assets.py
jnm: It turns out that the entire question gets silently discarded (!)
```
(Pdb) l
809             )
810             asset.deploy(backend='mock', active=True)
811             breakpoint()
812  
813             # Get the deployed XForm and parse it
814  ->         xform = XForm.objects.get(id_string=asset.uid)
815             parser = XFormInstanceParser(xform.xml, xform.data_dictionary())
816  
817             # Access the first child element of the <instance> node
818             instance_node = parser.get_root_node().getElementsByTagName('instance')[0]
819             root_element = instance_node.firstChild
(Pdb) import json
(Pdb) print(json.dumps(asset.content, indent=2))
{
  "survey": [],
  "settings": {
    "name": "custom_root_node_name"
  },
  "translations": [
    null
  ],
  "translated": [
    "label"
  ],
  "schema": "1"
}
```

rajpatel24: > Interesting! I'm surprised the unit tests don't fail since there's no `type`. I'll look into it.

I'm ashamed. I completely misunderstood this and replaced `type` with `text` 🤦
Thanks @jnm  for the correction :-)

jnm: Made an internal thread to discuss the weird (existing) behavior of silently discarding invalid rows:
https://chat.kobotoolbox.org/#narrow/stream/4-Kobo-Dev/topic/Invalid.20asset.20content.20rows.20silently.20disarded

PR #5139, opened 2024-10-01
[TASK-883] Create route for form activities

pauloamorimbr:
## Checklist

3. [ ] Ensure the tests pass

## Description

This PR is being merged into a project branch.

Creation of the route: `/forms/{id}/settings/activity`
Only a dummy page is in place right now

## Notes

~I noticed we'll need to setup permissions for the route elements and that's probably gonna take me some time, so I've created a [new task](https://www.notion.so/kobotoolbox/Setup-history-logs-permissions-fe-be-1127e515f65480a4b51af7ddc0765a13?pvs=4) for that.~

~For now, I'm leaving some existing permissions from other elements as a placeholder.~

I just checked with John and Jackie and fixed the permissions. I'm deleting the other task.


jsapp/js/components/formActivity.es6
```diff
@@ -0,0 +1,7 @@
+export default function FormActivity() {
```
magicznyleszek, line 1: Why not `formActivity.tsx`? Also I think it would be a good idea to make a separate directory for that file, e.g. `activity/formActivity.tsx`. I imagine we would need some styles file for it in near future.

`jsapp/js/components` directory was historically a bottomless pit for all js files, and we still have quite a lot of mess in there. So take it into account when trying to find stuff to mimic :D

magicznyleszek (changes requested):

jsapp/js/components/formViewSideTabs.es6
```diff
@@ -162,6 +162,20 @@ class FormViewSideTabs extends Reflux.Component {
           path: ROUTES.FORM_REST.replace(':uid', this.state.asset.uid),
         });
       }
+
+      // TODO: Fix permissions
```
magicznyleszek, line 166: > I noticed we'll need to setup permissions for the route elements and that's probably gonna take me some time, so I've created a [new task](https://www.notion.so/kobotoolbox/Setup-history-logs-permissions-fe-be-1127e515f65480a4b51af7ddc0765a13?pvs=4) for that.

I think you can just ask John/Jackie/Tino about what permissions should be applied here. `userCan` is a bit complex inside, but the usage is pretty straightforward. Just swap these with whatever they say we need here. Also take a look at `/jsapp/js/components/permissions/permConstants.ts` - there is a list of all permissions, each with a comment describing in some detail what is it about :)

magicznyleszek (changes requested):

jsapp/js/components/formViewSideTabs.es6
magicznyleszek: But it's ok to have this merged with the TODO comments - if it would make your work easier - as you will wait for the response :)

PR #5142, opened 2024-10-02
Reduce number of DB queries while transferring a project to another user

noliveleger:
## Description

Try to avoid running one update query per attachment while maintaining the atomicity of the transaction.

## Notes

The former choice to run `Attachment.save()` right after `.move()` was done in purpose. The idea was to be sure that the new path of the file was saved in the DB before processing another file. Unfortunately, this created as many DB queries as there were existing attachments (or media files), which was slowing down the process even more. Using `bulk_update` is (way) more efficient. Using the `try/finally` block should ensure the atomicity if an error occurs while looping on attachments (or media files).

`resume_failed_transfers_2_024_25_fix` has been revisited. The call of the command `update_attachment_storage_bytes` was useless because counters were not affected by the bug and were still accurate. 
Moving files are now wrapped in try/except to avoid the command to crash if one transfer cannot be fixed

## Testing

1. Test the management command

- Create a project with `2.024.25b`. It should include attachments and media files. 
- Deploy it
- Submit data with Enketo
- Go to Project settings > Sharing and transfer to another user
- Open another browser and log in as the other user
- Open the invite link you received in the other user browser and click accept.
- Look at the celery logs and see that at least of the celery tasks failed
- Reload other user browser page and see the project is there but data is missing
- Checkout the PR branch (don't forget to restart workers)
- Run the management command
- See data is back
- Transfer it back to former owner

Every should work as expected.

2. Ensure attachment or media_file move and DB update atomicity

- use `mockobo` to add several submissions to project created in 1), assuming the project has one "image" question.
```
./mockobo.py -a <asset_uid> -c <number_submissions_to_add> -mf image:<local_path_to_image>
```
- Force `move_attachments()` and/or `move_media_files()` to raise an error in the middle of the process (before the `FILE_MOVE_CHUNK_SIZE` is reached)
- Confirm that every object moved before the crash has a path that starts with correct owner username.

For example: 
previous_owner: `user_a`
new_owner: `user_b`

All object paths moved before crashed should start with `user_b`. All others are still starting with `user_a`.

### DIFF
--- a/kobo/apps/project_ownership/utils.py
+++ b/kobo/apps/project_ownership/utils.py
@@ -1,9 +1,11 @@
 import os
+import time
 from typing import Optional
 
 from django.apps import apps
 from django.utils import timezone
 
+from kobo.apps.openrosa.apps.logger.models.attachment import Attachment
 from kpi.deployment_backends.kc_access.shadow_models import (
     KobocatAttachment,
     KobocatMetadata,
@@ -11,6 +13,7 @@ from kpi.deployment_backends.kc_access.shadow_models import (
 from kpi.models.asset import AssetFile
 from .models.choices import TransferStatusChoices, TransferStatusTypeChoices
 from .exceptions import AsyncTaskException
+from .constants import ASYNC_TASK_HEARTBEAT, FILE_MOVE_CHUNK_SIZE
 
 
 def get_target_folder(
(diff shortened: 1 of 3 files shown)


kobo/apps/project_ownership/utils.py
```diff
@@ -102,41 +117,53 @@ def move_media_files(transfer: 'project_ownership.Transfer'):
...
+    finally:
+        # No need to use chunk size for media files like we do for attachments,
+        # because the odds are pretty low that more than 100 media files are
+        # linked to the project.
+        if metadata_to_update:
+            media_files_to_update.append(media_file)
+
+        if media_files_to_update:
+            AssetFile.objects.bulk_update(
+                media_files_to_update, update_fields=['content', 'metadata']
```
rgraber, line 165: this should be `fields` not `update_fields`

rgraber (changes requested):
Syntax error when moving media files

PR #5145, opened 2024-10-03
[TASK-1110] Implement basic UI for form activity

pauloamorimbr:
## Checklist

3. [x] Ensure the tests pass

## Description

This PR implements the UI for the `.../settings/activity` view with mocked data, based on the Figma docs and code standards.

## Notes

- For the mock data we've implemented a generator inside the react query hook
- The paginated table component is being used as a standard
- Some changes for the paginated table component were needed so that it could have a renderer for the data

View:
image


jsapp/js/components/activity/formActivity.tsx
```diff
@@ -1,7 +1,116 @@
...
+  who,
+  action,
+  what,
+}: {
+  who: string;
+  action: string;
+  what: string;
+}) => (
+  <div className='event-description'>
+    <div className='profile-picture'>{who.slice(0, 1)}</div>
```
magicznyleszek, line 47: In `accountMenu.tsx` there is a similar "avatar" thing. The interesting function in there is `stringToColor`. Up until now, we didn't need to render avatars anywhere else. Could you please create some avatar component in `js/components/common` and use it in both places? I'm ok with doing this in this PR.

jsapp/js/components/activity/formActivity.scss
```diff
@@ -0,0 +1,91 @@
...
+      width: sizes.$x150;
+    }
+  }
+
+  .table-container {
+    flex: 1;
+    overflow: auto;
+    padding: sizes.$x20;
+
+    .loading {
```
magicznyleszek, line 40: There is no `.loading` anywhere, should we add it or delete this? I think it's something I've omitted in `UniversalTable`. It's a good moment to add it :) From my conversation with Tessa the loading indicator should be similar to Data Table one: 
next-page-loader-action
There is an existing `LoadingSpinner` component that we should use here.

jsapp/js/components/activity/formActivity.scss
```diff
@@ -0,0 +1,91 @@
...
+      width: 100%;
+      height: 100%;
+      display: flex;
+      justify-content: center;
+      align-items: center;
+    }
+
+    td {
+      color: colors.$kobo-gray-700;
```
magicznyleszek, line 49: I would say that any override to `UniversalTable` styles should really be incorporated into `UniversalTable` itself. We are generally trying to have the least amount of overrides to existing components. It is better to unify the looks even if this means not following designs in 100%.

But in this case, I think that Figma file says it should be gray-800, so the default text color. Remove this and either do nothing, or add `color: colors.$kobo-gray-800` to `universalTable.module.scss` :)

jsapp/js/components/activity/formActivity.scss
```diff
@@ -0,0 +1,91 @@
...
+  flex-direction: column;
+  overflow: hidden;
+  background-color: colors.$kobo-white !important;
+
+  .header {
+    width: 100%;
+    display: flex;
+    gap: sizes.$x8;
+    align-items: center;
+    padding: 0 sizes.$x20;
```
magicznyleszek, line 17: We should also have `padding-top: 20px` to match Figma design.

jsapp/js/components/activity/formActivity.scss
```diff
@@ -0,0 +1,91 @@
...
+
+    h1 {
+      color: colors.$kobo-gray-800;
```
magicznyleszek, line 20: It should have `margin: 0` to reset the `normalize.css` value

magicznyleszek: And the color doesn't really change anything, as it's the same as the color for `html`

jsapp/js/components/activity/formActivity.tsx
```diff
@@ -1,7 +1,116 @@
...
+    <span className='description'>
+      <span className='who'>{who}</span>{' '}
+      <span className='action'>{action}</span> {what}
+    </span>
+    <button className='see-details'>{t('See details')}</button>
```
magicznyleszek, line 52: Just to be sure it's clear (for further PRs) - we are not doing the "expanded row" kind of details. We are going with a modal solution :)

jsapp/js/components/activity/formActivity.scss
```diff
@@ -0,0 +1,91 @@
...
+.main-container {
+  display: flex;
+  flex-direction: column;
+  overflow: hidden;
+  background-color: colors.$kobo-white !important;
+
+  .header {
+    width: 100%;
+    display: flex;
+    gap: sizes.$x8;
```
magicznyleszek, line 15: That `scss/sizes` is deprecated and I'm going to remove it next few days. We should use regular `px` values going forward :) So please change this file here to `px`, and I will have one file less to fix :P

magicznyleszek (changes requested):

jsapp/js/components/activity/formActivity.scss
pauloamorimbr: Oh...I had that in my first implementation, using the class to display the loading spinner in the place of the universal table during the loading. Looks like I forgot to remove the class. Sorry about that.

jsapp/js/components/activity/formActivity.tsx
```diff
@@ -1,7 +1,116 @@
...
+
+const columns: UniversalTableColumn[] = [
+  {
+    key: 'description',
+    label: t('Event description'),
+  },
+  {
+    key: 'date',
+    label: t('Date'),
+    size: 100,
```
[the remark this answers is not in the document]
pauloamorimbr: 🤔 After taking a better look at it, I don't think the `cellFormatter` will gonna be enough. The idea of the `rowRenderer` is that it takes the whole object and generate each cell because, in this case, I the cell will need to aggregate multiple values from the object to generate the complex rendering, even adding the Avatar. We could extract the avatar to another cell, but we'd still need to join multiple values.

For the universal table it's easier to implement something like this, but since the paginated takes care of fetching data we need a way to process it, that's why I added types to it, so we can indicate which is the data type and which is the cell type, because I don't think the types will match frequently.

WDYT?

jsapp/js/components/activity/formActivity.tsx
```diff
@@ -1,7 +1,116 @@
+import type {ReactNode} from 'react';
+import {useState} from 'react';
+import '../../../scss/components/_kobo.form-view.scss';
```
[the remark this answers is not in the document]
pauloamorimbr: It took me a while to find this file by myself, so maybe it's good to leave it there so the next to work on it know it's being used... WDYT?

jsapp/js/components/activity/formActivity.tsx
pauloamorimbr: I've added a local component using the stringToColor already. Maybe it's just a matter of copying it to another file now?

jsapp/js/components/activity/formActivity.scss
pauloamorimbr: I cleaned up the CSS and I'm creating a task for us to add it to the UniversalTable

jsapp/js/components/activity/formActivity.tsx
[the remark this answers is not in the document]
magicznyleszek: Sure let's leave it, I would just say make it absolute path :)

jsapp/js/components/activity/formActivity.tsx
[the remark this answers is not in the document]
pauloamorimbr: Just to close this up: The new version of PaginatedQueryUniversalTable has a different cellFormatter, which takes the whole DataItem as a parameter. I rebased this branch, my addition was removed and I'm now using the cellFormatter.
Thank you.

pauloamorimbr:
@magicznyleszek I've added a `QueryKeys` enum for us to use with react query and also created a new query for the filter options, this way I'm already keeping all the mock data into the Query file. I was planning to do that later on another task, but it seems the opportunity was just here.

pauloamorimbr:
@magicznyleszek To keep following Figma's layout I've suggested a change in the `avatar` component to allow for the label to be hidden

jsapp/js/query/queries/activityLog.query.ts
```diff
@@ -0,0 +1,96 @@
...
+export const useActivityLogsQuery = (itemLimit: number, pageOffset: number) =>
+  useQuery({
+    queryKey: [QueryKeys.activityLog, itemLimit, pageOffset],
+    queryFn: () => getActivityLogs(itemLimit, pageOffset),
+    placeholderData: keepPreviousData,
+  });
+
+/**
+ * This is a hook to fetch the filter options for the activity logs.
+ * @returns
```
magicznyleszek, line 90: Do we need all these empty `@returns`?

magicznyleszek (changes requested):

magicznyleszek (approved):

PR #5151, opened 2024-10-04
[TASK-1117] Fix alertify elusive crash when (un)archiving a project from My Projects

magicznyleszek:
## Checklist

3. [x] Ensure the tests pass

## Notes

I was able to reproduce the bug __sometimes__. I no longer can reproduce it now, so I am not sure this PR fixes it. It's an educated guess. 

This particular alertify instance has an async function insid `onok` callback, and only then is closing the dialog. But the `onok` is closing the dialog by default - unless the callback returns `false`. My guess is that without `return false;` alertify was trying to close the dialog and cleanup some inner things. And then the callback was resolved and it also tried to close the dialog. And sometimes this works fine, and sometimes it doesn't - this is my theory :wink:


jsapp/js/assetQuickActions.tsx
```diff
@@ -176,6 +176,12 @@ export function archiveAsset(
           dialog.destroy();
         }
       );
+      // Prevent the default alertify behaviour of trying to close the dialog.
+      // I suspect (a guess) that there is some race condition and that
+      // sometimes when `actions.resources.setDeploymentActive` finishes,
+      // the `dialog.destroy` can't find something and crashes with:
+      // `Uncaught TypeError: instance.__internal is undefined`
+      return false;
```
p2edwards, line 184: Works for me! Add this to `unarchiveAsset` too please :+1:

magicznyleszek: Added :)

PR #5167, opened 2024-10-14
Fix PlanAddOn creation logic

Guitlle:
## Description

Fix the PlanAddOn creation logic to avoid unnecessary instances when the payment is not successful 

## Notes

The stripe PlanAddOn model was modified.

## Testing

This is covered in unit tests, no manual testing necessary.

## Related issues

Notion [TASK-1118](https://www.notion.so/kobotoolbox/Fix-PlanAddon-creation-logic-1147e515f65480e399f5cccbd74095da)


jamesrkiger:
@Guitlle Can you add back the change to the is_available method that got dropped with the latest merge?

PR #5196, opened 2024-10-24
feat(organizations): Add `mongo_uuid` field to XForm model and populate it TASK-957

rajpatel24:
## Checklist

3. [x] Ensure the tests pass
8. [ ] Create a testing plan for the reviewer and add it to the Testing section

## Description

This PR introduces a new field, mongo_uuid, to the XForm model in KoboToolbox. This unique identifier is designed to enhance the ownership transfer process of XForms by ensuring that each form is linked to a distinct UUID in MongoDB. The addition of mongo_uuid will streamline how XForms are managed, especially during ownership transfers, by reducing unnecessary updates to the MongoDB documents related to the XForm.

## Notes

- A new column (mongo_uuid) has been added to the XForm model, which will serve as a unique identifier.
Ownership Transfer: The logic for transferring an XForm's ownership has been updated to ensure that the mongo_uuid is populated if it is currently null.
- All existing MongoDB documents that match the _userform_id will be updated to include the new mongo_uuid value, ensuring consistency across the database.
- The ParsedInstance.to_dict_for_mongo() method has been modified to prioritize the mongo_uuid over _userform_id when exporting data. If mongo_uuid is not null, _userform_id will be set to mongo_uuid; otherwise, the original logic will be maintained.
- The mongo_userform_id property in the OpenRosaDeploymentBackend class has been updated to return the mongo_uuid when available, ensuring that the correct ID is used during operations.


kobo/apps/project_ownership/tests/api/v2/test_api.py
```diff
@@ -500,17 +501,79 @@ def test_data_accessible_to_new_user(self):
         for attachment in response.data['results'][0]['_attachments']:
             assert attachment['filename'].startswith('anotheruser/')
 
+        # Get the mongo_uuid for the transferred asset (XForm)
+        xform = XForm.objects.get(kpi_asset_uid=self.asset.uid)
```
noliveleger, line 505: You can use the `XForm` object from the deployment, i.e.: `self.asset.deployment.xform`

kpi/deployment_backends/openrosa_backend.py
```diff
@@ -731,7 +733,13 @@ def get_validation_status(
 
     @property
     def mongo_userform_id(self):
-        return '{}_{}'.format(self.asset.owner.username, self.xform_id_string)
+        try:
+            return (
+                self.xform.mongo_uuid
+                or f'{self.asset.owner.username}_{self.xform_id_string}'
+            )
+        except KeyError:
```
noliveleger, line 741: `KeyError`? when can this error happen?

kpi/deployment_backends/openrosa_backend.py
```diff
@@ -1234,6 +1243,20 @@ def xform(self):
...
+        return ''.join(secrets.choice(characters) for _ in range(length))
+
+    def set_mongo_uuid(self):
+        """
+        Set the `mongo_uuid` for the associated XForm if it's not already set.
+        """
+        if not self.xform.mongo_uuid:
+            self.xform.mongo_uuid = self.generate_unique_key()
+            self.xform.save(update_fields=['mongo_uuid'])
+
```
noliveleger, line 1259: Please keep methods sorted alphabetically and let's use `KpiUidField.generate_unique_id()` which already generates a random string (don't forget to clean imports after).

noliveleger (changes requested):

kpi/deployment_backends/openrosa_backend.py
rajpatel24: Thanks for the review Olivier. The `KeyError` can occur when the `formid` key is missing from the `backend_response` dictionary. This is typically the case in certain test scenarios, such as `ConflictingVersionsMockDataExports`, where mock data doesn't include the `formid` key.

When `self.xform` is accessed, it attempts to retrieve the `formid` from `backend_response`, which raises a `KeyError` if it's not present.

noliveleger: As our per discussion, we don't want the `KeyError` to be silently ignored. 
It should never happen but if it does, retrieving `xform_id_string` should be also an issue. 

Fixture has been fixed to match production flow

noliveleger (approved):
LGTM

PR #5164, opened 2024-10-10
refactor(account): query-ify OrganizationContext TASK-1194

duvld:
## Checklist

1. [ ] If you've added code that should be tested, add tests
2. [ ] If you've changed APIs, update (or create!) the documentation
3. [ ] Ensure the tests pass
4. [ ] Run `./python-format.sh` to make sure that your code lints and that you've followed [our coding style](https://github.com/kobotoolbox/kpi/blob/main/CONTRIBUTING.md)
5. [x] Write a title and, if necessary, a description of your work suitable for publishing in our [release notes](https://community.kobotoolbox.org/tag/release-notes)
6. [ ] Mention any related issues in this repository (as #ISSUE) and in other repositories (as kobotoolbox/other#ISSUE)
7. [ ] Open an issue in the [docs](https://github.com/kobotoolbox/docs/issues/new) if there are UI/UX changes
8. [x] Create a testing plan for the reviewer and add it to the Testing section

## Description
Refactor some react context into react query.

## Notes

Basic refactoring example from our own implemented react context to react query. Using `OrganizatinContext` as the example.

## Testing

Go through functionality of billing/usage to see if the functionality remains the same. Look at code to see if react query is implemented correctly *and* if old react context has been removed correctly.

PR #5197, opened 2024-10-24
[TASK-1096] Implementation of queries for organization

pauloamorimbr:
## Checklist

3. [x] Ensure the tests pass

## Description

This PR implements reactQuery hooks to be used on organization development.

## Notes

The `useOrganizationQuery` hook was modified:
- A separated function was created to process organization data before returning it
- A `enabled` property was added to the query, to make sure it can only run after the `sessionStore` is loaded, since we need  the `currentAccount.organization.url` to load the organization details in the new implementation.
- We're forcing a `is_mmo` false when the feature flag is off while in development to ensure we're not displaying any in development functionality
- We are now using the new API implementation, where we first get the organization API url from `sessionStore.currentAccount.organization.url` and then we use that URL to fetch the organization data.

### Check for the mock data:
`useOrganizationQuery` is currently being used in `accountSidebar`, so one way of testing this implementation is to use:
``` typescript
  const orgQuery = useOrganizationQuery();
  console.log(orgQuery.data);
```
This way one can see the result of the data mocking with and without the `mmosEnabled` feature flag.

### Check for the conditional query
On can comment out or add a `setTimeout` in `session.tsx` to prevent the `sessionStorage.isPending` from being set to false, causing the organization query to not run, waiting for it.

Other than code-related tests, there should be no changes in the current functionality.


Akuukis (changes requested):
- [x] As funny as they are, I'm ok with naming because they are consistent and descriptive.
- [x] blocked by #5164 
- [ ] blocked by decision in [internal chat](https://chat.kobotoolbox.org/#narrow/stream/4-Kobo-Dev/topic/FE.3A.20react-query.20conventions)

jsapp/js/dataInterface.ts
```diff
@@ -786,6 +786,8 @@ export interface AccountResponse {
     tag: string | boolean;
   };
   social_accounts: SocialAccount[];
+  // Organization details
+  organizationPath?: string;
```
jamesrkiger, line 790: This should be structured as `organization?: {url: string}`

jsapp/js/account/stripe.api.ts
```diff
@@ -48,13 +50,62 @@ export async function changeSubscription(
...
+    // Once we have the full data from backend, this function should be modified to return
+    // the actual data.
+
+    if (isMmoEnabled) {
+
+      const currentAccount = sessionStore.currentAccount as AccountResponse;
+
+      if (!currentAccount) {throw new Error('No account data found');}
+
+        const organizationUrl = currentAccount.organization?.url;
```
duvld, line 68: I think this can be indented back to match the `if` above it?

jsapp/js/account/stripe.api.ts
```diff
@@ -48,13 +50,62 @@ export async function changeSubscription(
...
+
+        // Here we would fetch organization data from organizationSlug
```
duvld, line 70: Did you mean `organizationUrl` instead of `organizationSlug` (or vice versa for the variables)?

jsapp/js/account/stripe.api.ts
```diff
@@ -48,13 +50,62 @@ export async function changeSubscription(
...
+        // For now, we will return mocked data
+        console.log('Fetching organization data from:', organizationUrl);
+
+        // Return mocked data for MMO
+        return {
```
duvld, line 75: I think in the past we added some artificial delay to make it more "realistic", what do you think about that?

jsapp/js/account/stripe.api.ts
```diff
@@ -48,13 +50,62 @@ export async function changeSubscription(
...
+  // feature flag dependencies from being added to the hook
+  const fetchOrganization = async (): Promise<Organization> => {
+    // Once we have the full data from backend, this function should be modified to return
+    // the actual data.
+
+    if (isMmoEnabled) {
+
+      const currentAccount = sessionStore.currentAccount as AccountResponse;
+
+      if (!currentAccount) {throw new Error('No account data found');}
```
duvld, line 66: React query question, would this be sufficient for, later in some component, for the check `orgQuery.isError` to return true?

jsapp/js/account/stripe.types.ts
```diff
@@ -147,6 +147,8 @@ export interface Organization {
   modified: string;
   slug: string;
   is_owner: boolean;
+  is_mmo: boolean;
```
duvld, line 150: Should these be optional until the backend will actually return it? Or just returning false in the query is safe enough

duvld (commented):
Just some questions

jsapp/js/account/stripe.api.ts
pauloamorimbr: Exactly. React query catches any error and sets its error flag automatically.

jsapp/js/account/stripe.types.ts
pauloamorimbr: As far as I know, and from my tests, backend is already returning these fields. I'm setting it to false in the current query because when talking to @jamesrkiger we thought it would be a good idea to force the default "false" values until we have the front end properly implemented to avoid issues.

jsapp/js/account/stripe.api.ts
pauloamorimbr: Forgot to fix that comment. Thank you!

jsapp/js/account/stripe.api.ts
pauloamorimbr: Good one. Just added it.

duvld:
Look good to me now :)

jamesrkiger:
@pauloamorimbr I'm testing this branch out locally (with Stripe enabled, though I'm not sure that's pertinent) and I am seeing a redundant organization query when I navigate from the projects page (home) to account settings. Doesn't seem to happen on main. I suspect it's because the profile page calls `sessionStore.refreshAccount()`.

jamesrkiger:
@pauloamorimbr It looks like the work for adding orgs to the /me endpoint got [merged yesterday](https://github.com/kobotoolbox/kpi/pull/5204/files). I'm inclined to say we should implement the real fetch now, at least behind the feature flag (though maybe we don't even need the feature flag?). What do you think?

PR #5205, opened 2024-10-28
Fix SSRFProtect validation options to properly handle Constance settings

noliveleger:
## Description :
This PR addresses an issue where SSRF Protect validation fails to consider the Constance settings correctly. It now strips line break characters from the IP input before performing the SSRF Protect validation.

## Testing

- Create a project (deploy it first) with RestService (Project > Settings > Rest Services)
- Use local ip (of your host machine) for the endpoint URL (e.g, http://192.168.0.10)
- Add `192.168.0.10` and `192.168.0.11` in `SSRF_ALLOWED_IP_ADDRESS` in Constance
- Submit data
- SSRFProtect Validation:
   1. Without this PR, it should raise error `ssrf_protect.exceptions.SSRFProtectException: URL "http://192.168.0.10" is not allowed because it resolves to a private IP address` 
   2. With this PR, it should not raise the error (but could still fail because you don't have a webserver which listens on local computer).

### DIFF
--- a/kpi/utils/strings.py
+++ b/kpi/utils/strings.py
@@ -1,4 +1,3 @@
-# coding: utf-8
 import base64
 
 
@@ -13,5 +12,5 @@ def to_str(obj):
 
 
 def split_lines_to_list(value: str) -> list:
-    ip_addresses = value.strip().split('\n')
-    return [ip.strip() for ip in ip_addresses if ip.strip()]
+    values = value.strip().split('\n')
+    return [ip.strip() for ip in values if ip.strip()]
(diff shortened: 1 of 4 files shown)


rgraber (changes requested):
Could we separate out the parsing into another method for more contained testing? In this case the test could fail if anything else in the send() goes wrong, and it requires more setup (making it harder to maintain)

kpi/utils/strings.py
```diff
@@ -10,3 +10,8 @@ def to_str(obj):
     if isinstance(obj, bytes):
         return obj.decode()
     return obj
+
+
+def split_lines_to_list(value: str) -> list:
```
rgraber, line 15: nit: i would use `values` instead of `ip_addresses` since this could be used for any strings. not blocking

rgraber (approved):
LGTM

kpi/utils/strings.py
noliveleger: 💯

PR #5207, opened 2024-10-29
Limit Celery Beat event updates on bulk trash-bin actions

noliveleger:
## Description

This PR restricts the number of Celery Beat event updates triggered during bulk actions within the trash bin. By reducing the frequency of updates, we prevent the stacking of multiple updates that can create a domino effect, leading to a bottleneck. 

This allows projects to be deleted without delays and helps to prevent 504-Timeout errors.

## Notes
Use a context manager to disconnect and reconnect Celery Beat signals on save and delete.
Only update Celery Beat event scheduler if it has not been updated for the last 5 minutes.

### DIFF
--- a/kobo/apps/trash_bin/utils.py
+++ b/kobo/apps/trash_bin/utils.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import json
+from contextlib import contextmanager
 from copy import deepcopy
 from datetime import timedelta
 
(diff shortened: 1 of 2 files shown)


kobo/apps/trash_bin/utils.py
```diff
@@ -298,6 +290,35 @@ def replace_user_with_placeholder(
     return placeholder_user
 
 
+@contextmanager
+def signals_temporarily_disconnected(save=False, delete=False):
```
jnm, line 294: Nitpick, but let's make this `temporarily_disconnect_signals()`, in the imperative mood

jnm (changes requested):

jnm (approved):

PR #5219, opened 2024-11-01
feat(organizations): update page access permissions TASK-977

jamesrkiger:
## Checklist

3. [x] Ensure the tests pass
9. [x] Add frontend or backend tag and any other appropriate tags to this pull request

## Description

For various pages in account settings, we previously used a wrapper component to verify whether the user was the owner of their org before allowing them to access the page. With the organizations project, this permissions check needs to allow for checking multiple role options. We also need to be able to check whether an org is an mmo for the members page. This PR overhauls the old wrapper component to allow for these checks.

## Testing

With Stripe enabled (and products synced), create a new user. Check the routes in the sidenav that were previously wrapped with `RequireOrgOwner`. There should be no difference with current behavior on main. 

Then turn on mmo_override for the org. Create a second user and then add them to the org via django admin. Logged in as the second user, check the routes again. You will have to enter the url manually for members, usage and org settings. You should be redirected to the account settings page, because the second user only has the org role of `member`. Finally, open the organization in django admin  and check "is_admin" for the second user. Check the routes again. The admin should be able to view the members, usage and settings pages but still not the plans page.

### DIFF
--- a/jsapp/js/account/organizations/validateOrgPermissions.component.tsx
+++ b/jsapp/js/account/organizations/validateOrgPermissions.component.tsx
@@ -30,14 +30,13 @@ export const ValidateOrgPermissions = ({
   ) : true;
   const hasValidOrg = mmoOnly ? orgQuery.data?.is_mmo : true;
 
-  // Redirect to Account Settings if you're not the owner
+  // Redirect to Account Settings if conditions not met
   useEffect(() => {
     if (
       redirect &&
       !orgQuery.isPending &&
       orgQuery.data &&
-      !hasValidRole &&
-      !hasValidOrg
+      (!hasValidRole || !hasValidOrg)
     ) {
       navigate(ACCOUNT_ROUTES.ACCOUNT_SETTINGS);
     }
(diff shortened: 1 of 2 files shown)


jsapp/js/account/organizations/validateOrgPermissions.component.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const hasValidRole = validRoles ? validRoles.includes(
+    orgQuery.data?.request_user_role ?? OrganizationUserRole.member
+  ) : true;
+  const hasValidOrg = mmoOnly ? orgQuery.data?.is_mmo : true;
+
+  // Redirect to Account Settings if conditions not met
+  useEffect(() => {
+    if (
+      redirect &&
+      !orgQuery.isPending &&
```
pauloamorimbr, line 37: I believe its redundant to check for `isPending` if you're also checking for data existence.
According to the docs `isPending` is direct related to data existing (isPending or status === 'pending' - The query has no data yet -> [here](https://tanstack.com/query/latest/docs/framework/react/guides/queries))

pauloamorimbr (approved):
Really nice! 💯 

I was able to test after some issues with org creation and setting users! 🙌🏻 

Left a nitpick comment.

PR #5220, opened 2024-11-04
fix(Library): Fix collection child route not loading properly TASK-1216

magicznyleszek:
## Checklist

3. [x] Ensure the tests pass
9. [x] Add frontend or backend tag and any other appropriate tags to this pull request

## Description

Fix collection child not loading (UI displays infinite spinner) when trying to navigate into the collection child consecutive times.

## Notes

For cached asset `PermProtectedRoute` was skipping the Reflux action call, and using the callback function directly. Unfortunately `AssetRoute` is expecting the Reflux action to happen (regardless if response comes from cache). This is sign of a bigger issue in our codebase - #4835 brought asset cache that is causing bugs, but the fact that we have a mess in the ways of loading the asset is the real culprit.

## Testing

Given the fact that `PermProtectedRoute` is being used for many (27!) different routes, so it would be best if tester would visit each of them, just to verify they still load properly :)

Besides that, please follow reproduction steps from Notion task description.

## Related issues

This is caused by the dreaded #4835

### DIFF
--- a/jsapp/js/router/permProtectedRoute.tsx
+++ b/jsapp/js/router/permProtectedRoute.tsx
@@ -61,6 +61,17 @@ class PermProtectedRoute extends React.Component<
       return;
     }
 
+    // Listen to incoming load of asset
+    this.unlisteners.push(
+      actions.resources.loadAsset.completed.listen(
+        this.onLoadAssetCompleted.bind(this)
+      ),
+      actions.resources.loadAsset.failed.listen(
+        this.onLoadAssetFailed.bind(this)
+      )
+    );
+
+    // See if the asset is already loaded in the store
     const assetFromStore = assetStore.getAsset(this.props.params.uid);
     if (assetFromStore) {
       // If this asset was already loaded before, we are not going to be picky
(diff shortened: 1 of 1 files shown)


Akuukis (approved):
code LGTM, not tested. Nice use of JSDocs on the prop.

According to the new [PR workflow](https://www.notion.so/kobotoolbox/KIP-Pull-Requests-1147e515f6548094bf2ec594ed1f5617#1287e515f65480e08069e274595d8333), few comments about the title:
- "fix(Library): ~~Fix~~ collection ..." -- don't repeat yourself & lower case title
- "fix(~~L~~library): ..." -- use lower case for groups/features/routes.
- "fix(Library): ~~Fix collection child route not loading properly~~ load collection child route properly TASK-1216" -- use imperative mood & fit in the 50ish chars

jsapp/js/router/permProtectedRoute.tsx
```diff
@@ -27,6 +27,10 @@ interface PermProtectedRouteState {
   userHasRequiredPermissions: boolean | null;
   errorMessage?: string;
   asset: AssetResponse | null;
+  /**
+   * Tells the `dmix` mixin (from `mixins.tsx`) that this route component
+   * already handled asset load, so `dmix` doesn't have to.
+   */
```
p2edwards, line 33: good comment :heart:

PR #5233, opened 2024-11-06
feat(organizations): disable email updates for regular organization members TASK-997

duvld:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

UI for changing emails is disabled for `Members` of a multi-member organization.

### 👀 Preview steps
1. Make sure your user's email is correct in both the user section of the django admin as well as the email address section. If there is a difference it would only show the former here. See https://www.notion.so/kobotoolbox/task-1236
2. Make an MMO that has an owner, admin, and a member, as well as a regular user
3. Navigate to the account settings page
4. Visit the security section
5. If the user is a regular user, or an MMO owner or admin, they should be able to change their email address as normal
6. If the user is a member of an MMO the text box and button should be gone

### 📖 Description

Remove the text box and button entirely if the user is a `member`. Replaced with just text of the email returned from `/me`.

### 💭 Notes

Make sure the user either has the proper email reflected in the user section of the django admin or was created through an email confirmation link. See https://www.notion.so/kobotoolbox/task-1236


jsapp/js/account/security/email/emailSection.component.tsx
```diff
@@ -55,15 +58,20 @@ export default function EmailSection() {
   }, []);
 
   function setNewUserEmail(newEmail: string) {
-    setUserEmail(newEmail).then(() => {
-      getUserEmails().then((data) => {
-        setEmail({
-          ...email,
-          emails: data.results,
-          newEmail: '',
+    setUserEmail(newEmail).then(
```
jamesrkiger, line 61: Unless my eyes are missing something, I don't think you actually touched the code here. Looks like an automatic lint change. We want to avoid noise like this in PRs so the code changes are easier to read. Can you revert it?

jsapp/js/account/security/email/emailSection.component.tsx
```diff
@@ -112,6 +120,14 @@ export default function EmailSection() {
     }
   }
 
+  function userCanChangeEmail() {
```
jamesrkiger, line 123: Can we make this a variable instead of a function? I think you could also use a ternary to shorten it up.

jsapp/js/account/security/email/emailSection.component.tsx
```diff
@@ -120,20 +136,24 @@ export default function EmailSection() {
   return (
     <section className={securityStyles.securitySection}>
       <div className={securityStyles.securitySectionTitle}>
-        <h2 className={securityStyles.securitySectionTitleText}>{t('Email address')}</h2>
+        <h2 className={securityStyles.securitySectionTitleText}>
+          {t('Email address')}
```
jamesrkiger, line 140: See note above about unrelated lint changes

jsapp/js/account/security/email/emailSection.component.tsx
```diff
@@ -120,20 +136,24 @@ export default function EmailSection() {
...
-          session.isInitialLoadComplete &&
-          'email' in currentAccount && (
-            <TextBox
-              value={email.newEmail}
-              placeholder={t('Type new email address')}
-              onChange={onTextFieldChange.bind(onTextFieldChange)}
-              type='email'
-            />
-          )}
+        session.isInitialLoadComplete &&
```
jamesrkiger, line 145: I think there was too much logic in this template even before adding another boolean check. But adding in userCanChangeEmail really pushes it over the edge for me. I'd say you can abstract these out as a single variable: 
```
!session.isPending &&
session.isInitialLoadComplete &&
'email' in currentAccount &&
```
While we're at it, do we actually need to check both isInitialLoadComplete and isPending? Maybe we do, but it smells redundant to me.

And those checks are actually duplicated in two places right next to each other (see lines 159-161). At a glance, I'd say we can combine these and then just nest the check for unverifiedEmail?.email.

jsapp/js/account/security/email/emailSection.component.tsx
```diff
@@ -185,21 +205,24 @@ export default function EmailSection() {
...
+              handleSubmit();
+            }}
+          >
+            <Button
+              label='Change'
+              size='m'
+              type='primary'
+              onClick={handleSubmit}
+            />
+            )
```
jamesrkiger, line 222: Looks like a stray parenthesis

jsapp/js/account/security/email/emailSection.module.scss
```diff
@@ -34,3 +34,8 @@
   display: flex;
   gap: 10px;
 }
+
+.emailText {
```
jamesrkiger, line 38: The text is black in the designs:
image

Also, can we use a more descriptive class name? Perhaps "readOnlyEmailText" or something along those lines.

jsapp/js/account/security/email/emailSection.component.tsx
```diff
@@ -124,82 +132,81 @@ export default function EmailSection() {
...
+                {t('Email was sent again: ##TIMESTAMP##').replace(
+                  '##TIMESTAMP##',
+                  email.refreshedEmailDate
+                )}
+              </label>
+            )}
+          </>
+        )}
+      </div>
+      <div className={styles.options}>
```
jamesrkiger, line 193: Should we move this container div inside the userCanChangeEmail check?

jsapp/js/account/security/email/emailSection.module.scss
jamesrkiger: Just a reminder about the class name change here.

jsapp/js/account/security/email/emailSection.component.tsx
```diff
@@ -116,6 +119,11 @@ export default function EmailSection() {
   const unverifiedEmail = email.emails.find(
     (userEmail) => !userEmail.verified && !userEmail.primary
   );
+  const isReady = session.isInitialLoadComplete && 'email' in currentAccount;
+  // Only users who are members in a MMO cannot change their email
```
jamesrkiger, line 123: I think we can drop this comment. The code explains itself, imo.

jsapp/js/account/security/email/emailSection.module.scss
duvld: Missed this one, thanks for reminder

jsapp/js/account/security/email/emailSection.component.tsx
duvld: This is a bit of a hack, if the container div is inside of the check then the rest of the row CSS gets unaligned. The only way I saw to fix this was to add conditional styling which I thought looked worse than this. What do you think?

jamesrkiger: Ah, I wondered if that might be the reason. I'd say we should go with conditional CSS. You're right that it will add some more complexity, but I think it will make it easier for other coders working on this section in the future to know what's going on.

jamesrkiger (approved):
@duvld There was an import error, so I pushed a fix for that. I also made a small change to simplify the logic in that cx() function. You can go ahead and merge unless you have any objections to my changes.

PR #5218, opened 2024-10-31
feat(organizations): add new asset endpoint for owners and admins TASK-960

noliveleger:
### Summary:
Added a new endpoint to retrieve organization assets, accessible to both owners and admins.

### Description:
This update introduces a dedicated endpoint that allows organization owners and admins to access a complete list of assets associated with their organization. Owners and admins have unrestricted access to these assets by virtue of their roles, without needing explicit permissions. 

### Notes:
As part of this update, a refactor was implemented in the unit test suite to minimize code duplication when testing organization admin access across various endpoints. This refactor allows us to test admin access without explicitly assigning permissions, ensuring that the tests more accurately reflect role-based access.

## Testing

- Register three different users
- From the shell, add user2 and user3 to user1 org 
```python
organization = user1.organization
organization.mmo_override = True
organization.save()
organization.add_user(user2, is_admin=True)
organization.add_user(user3, is_admin=False)
```
- The UI is not aware yet about org's admin role, so you will receive Access Denied until you patch the JS code (see below) when you try to access project details but the API call should work.

user2 is now an admin and should be able to see and to do whatever they want on user1's organization projects, they should be able to access the org assets list: `/api/v2/organizations/<organization_uid>/assets/`
user3 is a regular member, they cannot access `/api/v2/organizations/<organization_uid>/assets/` 

`organization_uid` can be retrieve from `/me` in `organization` property. 

- Create a project with user3
   1.  user3 should see it in the regular asset list (i.e. : `/api/v2/assets/`) and should have been assigned `manage_asset` permission
   2.  user2 should see it in the org asset list but not in the regular asset list. No permissions assigned.
   3.  user1 should see it in the org asset list and in the regular asset list (until the latter is filtered in a future PR see TASK-1187). They are the owner.

- Create a project with user2
   1.  user3 should not see it in the regular asset list (and should be able to access its details either)
   2.  user2 should see it in the org asset list and in the regular asset list and should have been assigned `manage_asset` permission
   3.  user1 should see it in the org asset list and in the regular asset list. They are the owner.
   
- Create a project with user1
   1.  user3 should not see it in the regular asset list (and should be able to access its details either)
   2.  user2 should see it in the org asset list but not in the regular asset list. No permissions assigned.
   3.  user1 should see it in the org asset list and in the regular asset list. They are the owner.

**By-pass front-end checks to allow org admins access project details**
[...]

PR #5235, opened 2024-11-06
feat(organizations): add endpoints to handle organization members TASK-963

rajpatel24:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
- Enhance the organization member management API to support role updates and member removal via PATCH and DELETE requests, excluding member creation.

### 📖 Description
- This update introduces endpoints for managing organization members:

    - GET `/api/v2/organizations/<org-id>/members/` to list all members in an organization.
    - PATCH `/api/v2/organizations/<org-id>/members/<username>/` to update a member's role (e.g., promote to admin).
    - DELETE `/api/v2/organizations/<org-id>/members/<username>/` to remove a member from the organization.

Note: Creating members is not supported via this endpoint. Roles are restricted to 'admin' or 'member'. Including the details of invited members (those who have not yet joined the organization) is not covered in this update.

- This update provides new functionality to manage organization members:

    - Role updates for members (promotion/demotion between 'admin' and 'member').
    - Member removal from an organization.
    - The endpoints are optimized to use database joins to fetch member roles efficiently without excessive database queries, ensuring minimal load.

### 👀 Preview steps
1. ℹ️ Have an account and an organization with multiple members.
2. Use the GET method to list the members of the organization.
3. Use the PATCH method to update a member's role (e.g., promote a user to admin).
4. Use the DELETE method to remove a member from the organization.
5. 🟢 Notice the endpoints behave as expected, with valid role updates and member removal.

### 💭 Notes
- The code uses database joins to optimize role determination for members, avoiding inefficient role lookups.
- Role validation restricts role assignments to 'admin' or 'member' only.
- Including invited members' details is not covered in this update. This will be implemented in a future update.


kpi/paginators.py
```diff
@@ -142,3 +142,10 @@ class TinyPaginated(PageNumberPagination):
     Same as Paginated with a small page size
     """
     page_size = 50
+
+
+class OrganizationPagination(PageNumberPagination):
```
noliveleger, line 147: I would rename the class just to match what it is paginating (i.e.: organization members)

kpi/paginators.py
```diff
@@ -142,3 +142,10 @@ class TinyPaginated(PageNumberPagination):
...
+    """
+    Pagination class for Organization
+    """
+    page_size = 10
```
noliveleger, line 151: Is `10` a requirement?

kobo/apps/organizations/serializers.py
```diff
@@ -11,10 +15,56 @@
...
+            'user__email',
+            'user__name',
+            'role',
+            'has_mfa_enabled',
+            'date_joined',
+            'is_active'
+        ]
+
+    def get_has_mfa_enabled(self, obj):
+        return config.MFA_ENABLED
```
noliveleger, line 49: Actually, you need to get if the user has MFA enabled. (e.g. `MfaMethod.objects.filter(user=obj, is_active=True).exists()`). Ensure to optimize this for a list like you did for role ;-)

kobo/apps/organizations/tests/test_organization_members_api.py
```diff
@@ -0,0 +1,80 @@
...
+from kpi.tests.kpi_test_case import BaseTestCase
+from kpi.urls.router_api_v2 import URL_NAMESPACE
+
+
+class OrganizationMemberAPITestCase(BaseTestCase):
+    fixtures = ['test_data']
+    URL_NAMESPACE = URL_NAMESPACE
+
+    def setUp(self):
+        self.organization = baker.make(Organization, id='org_12345')
```
noliveleger, line 16: For future-proof validaiton, please add `mmo_override=True` 
This can be simplied with

```python
self.organization = baker.make(Organization, id='org_12345', mmo_override=True)
self.owner_user = baker.make(User, username='owner')
self.member_user = baker.make(User, username='member')
self.invited_user = baker.make(User, username='invited')

self.organization.add_user(self.owner_user)  # first user added to org, is always the owner and an admin
self.organization.add_user(self.member_user, is_admin=False) 
```

kobo/apps/organizations/tests/test_organization_members_api.py
```diff
@@ -0,0 +1,80 @@
...
+        self.owner_user = baker.make(User, username='owner')
+        self.member_user = baker.make(User, username='member')
+        self.invited_user = baker.make(User, username='invited')
```
noliveleger, line 19: `invited_user` is not used at all for the moment, right? Let's remove until we implement invitations.

kobo/apps/organizations/tests/test_organization_members_api.py
```diff
@@ -0,0 +1,80 @@
...
+        response = self.client.patch(self.detail_url('member'), data)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.data['role'], 'admin')
+
+    def test_delete_member(self):
+        response = self.client.delete(self.detail_url('member'))
+        self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
+        # Confirm deletion
+        response = self.client.get(self.detail_url('member'))
+        self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
```
noliveleger, line 75: All your tests are done with `self.client == self.owner` , but we need to validate permissions for regular members, admins and anonymous. Let's those tests as soon as #5218 is merged. I will tag this PR `blocked by`

kobo/apps/organizations/serializers.py
```diff
@@ -11,10 +15,56 @@
...
+    role = serializers.CharField()
+    has_mfa_enabled = serializers.SerializerMethodField()
+    url = serializers.SerializerMethodField()
+    date_joined = serializers.DateTimeField(
+        source='user.date_joined', format='%Y-%m-%dT%H:%M:%SZ'
+    )
+    user__username = serializers.ReadOnlyField(source='user.username')
+    user__name = serializers.ReadOnlyField(source='user.get_full_name')
+    user__email = serializers.ReadOnlyField(source='user.email')
+    is_active = serializers.ReadOnlyField(source='user.is_active')
```
noliveleger, line 32: I think we should use `user__is_active` to keep the same convention as other fields.

kobo/apps/organizations/serializers.py
```diff
@@ -11,10 +15,56 @@
 
 
 class OrganizationUserSerializer(serializers.ModelSerializer):
+    user = serializers.HyperlinkedRelatedField(
+        queryset=get_user_model().objects.all(),
+        lookup_field='username',
+        view_name='user-kpi-detail',
+    )
+    role = serializers.CharField()
+    has_mfa_enabled = serializers.SerializerMethodField()
```
noliveleger, line 24: I think we should use `user__has_mfa_enabled` to keep the same convention as other fields.

noliveleger (changes requested):

kobo/apps/organizations/views.py
```diff
@@ -194,3 +202,173 @@ def asset_usage(self, request, pk=None, *args, **kwargs):
...
+        instance = self.get_object()
+        serializer = self.get_serializer(
+            instance, data=request.data, partial=True
+        )
+        serializer.is_valid(raise_exception=True)
+        role = serializer.validated_data.get('role')
+        if role:
+            instance.is_admin = (role == 'admin')
+            instance.save()
+        return super().partial_update(request, *args, **kwargs)
```
noliveleger, line 374: Let's move the logic to the serializer to avoid calling the "update" twice. something like

```python 
def update(self, instance, validated_data):
     if role := validated_data.pop('role', None)
         validated_data['is_admin'] = role == 'admin'
     return super().update(instance, validated_data)
```

noliveleger (changes requested):

noliveleger:
This PR is blocked by #5218

magicznyleszek:
@noliveleger @rajpatel24 this is no longer blocked, right?

kpi/paginators.py
rajpatel24: 10 is not a requirement defined in the task. I need to check with you or the team to determine the appropriate value for it.

kobo/apps/organizations/tests/test_organization_members_api.py
rajpatel24: Thanks for pointing that out! I've added the role-based validation tests for organization member permissions

kobo/apps/organizations/serializers.py
```diff
@@ -11,10 +15,56 @@
 
 
 class OrganizationUserSerializer(serializers.ModelSerializer):
```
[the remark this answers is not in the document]
rajpatel24: Achieved this using `http_method_names`. Added a test for it as well

kobo/apps/organizations/views.py
rajpatel24: Done! I made a slight adjustment to your suggested implementation. I opted to use `get` instead of `pop` when accessing the `role` value from `validated_data`, as this allows the updated value to be properly reflected in the API response :-)

kpi/paginators.py
jamesrkiger: @rajpatel24 Can we change this to limit offset pagination? Our table components on the frontend currently assume limit offset pagination and @noliveleger said it shouldn't cause any problems to use that here.

kpi/paginators.py
```diff
@@ -137,6 +137,13 @@ def get_count(self, queryset):
         return super().get_count(queryset)
 
 
+class OrganizationMembersPagination(PageNumberPagination):
+    """
+    Pagination class for Organization Members
+    """
+    page_size_query_param = 'page_size'
+
+
```
noliveleger, line 146: You can delete this class, not needed anymore

kobo/apps/organizations/serializers.py
```diff
@@ -12,10 +15,60 @@
...
+    role = serializers.CharField()
+    user__has_mfa_enabled = serializers.BooleanField(
+        source='has_mfa_enabled', read_only=True
+    )
+    url = serializers.SerializerMethodField()
+    date_joined = serializers.DateTimeField(
+        source='user.date_joined', format='%Y-%m-%dT%H:%M:%SZ'
+    )
+    user__username = serializers.ReadOnlyField(source='user.username')
+    user__name = serializers.ReadOnlyField(source='user.get_full_name')
```
noliveleger, line 32: I think that's the wrong field. You need to retrieve  `User.extra_details.data['name']` . BTW in that case, the field should be `user__extra_details__name` . 

cc @jamesrkiger

kobo/apps/organizations/views.py
```diff
@@ -79,10 +86,12 @@ class OrganizationViewSet(viewsets.ModelViewSet):
     queryset = Organization.objects.all()
     serializer_class = OrganizationSerializer
     lookup_field = 'id'
-    permission_classes = (IsAuthenticated, IsOrgAdminOrReadOnly)
+    permission_classes = [HasOrgRolePermission]
     pagination_class = AssetUsagePagination
```
noliveleger, line 90: Asked by the front-end team, let's use `LimitOffsetPagination` instead.

noliveleger (changes requested):

kobo/apps/organizations/views.py
```diff
@@ -263,3 +272,195 @@ def asset_usage(self, request, pk=None, *args, **kwargs):
...
+
+    def destroy(self, request, *args, **kwargs):
+        """
+        Delete an organization member and their associated user account
+        """
+        instance = self.get_object()
+        user = instance.user
+        instance.delete()
+        user.delete()
+        return Response(status=status.HTTP_204_NO_CONTENT)
```
noliveleger, line 466: You can remove this method entirely. Right now, we want to delete the `OrganizationUser` only.  Do not touch the `User` object.

kpi/paginators.py
rajpatel24: @jamesrkiger I have removed the custom pagination class for the organization members API. It will now use `LimitOffsetPagination` by default.

kpi/paginators.py
rajpatel24: Done. It will use `LimitOffsetPagination` by default

noliveleger (approved):
LGTM 

For the record: The complex query took only few millisecond on a 500 members org.

PR #5240, opened 2024-11-07
feat(organization): add new route for organization projects TASK-975

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Make it possible to view your organization projects in a table.

### 📖 Description
Add new route for viewing your organization projects at `#/projects/organization`. Make it possible to switch into the new route through the View Switcher UI element.

### 👀 Preview steps

1. ℹ️ have a non admin account (e.g. "joe") with couple of projects
2. enable feature flag `mmosEnabled`
3. navigate to "My projects"
4. 🟢 notice that "My projects" is just a text, not a View Switcher (kind of a dropdown)
5. log in as admin and go to http://kf.kobo.local/admin/organizations/organization/
6. choose non admin account oganization (e.g. "joe's organization")
7. enable "Multi-members override"
8. log in as non admin account ("joe")
9. navigate to "My projects"
10. 🟢 notice that "My projects" is a View Switcher (kind of a dropdown), and one of the options would be the non admin account organization name ("joe's organization")
11. select organization
12. 🟢 notice that in the Projects Table the organizations projects load properly

### 💭 Notes

This is build atop #5218 so it would be best to wait with merge until the BE PR is done. Still, this one is totally reviewable :heart:


jsapp/js/projects/myOrgProjectsRoute.tsx
```diff
@@ -0,0 +1,256 @@
...
+  HOME_EXCLUDED_FIELDS,
+} from './projectViews/constants';
+import {ROOT_URL} from 'js/constants';
+import {endpoints} from 'js/api.endpoints';
+import {PROJECTS_ROUTES} from 'js/router/routerConstants';
+
+// Styles
+import styles from './projectViews.module.scss';
+
+interface InviteState {
```
magicznyleszek, line 47: @jamesrkiger the whole invite related code is copy-pasted from `myProjectsRoute.tsx`. I wonder if I should componentize it and reuse in both, or should I rather remove id from `myOrgProjectsRoute.tsx`, and have the invite feature accessible in "My Projects" only.

jamesrkiger: @magicznyleszek I think the invite code should be abstracted out. But on a broader note: Would it be possible to refactor more of these very similar views (`myProjectsRoute`, `myOrgProjectsRoute` and `customViewRoute`)?  At a glance, it looks like there's a lot of duplicated code we could reuse.

magicznyleszek: I will see how possible it would be :)

magicznyleszek: @jamesrkiger These are the unique pieces of each of the three components:

- dropzone: `MyProjectsRoute`
- invites: `MyProjectsRoute`, `MyOrgProjectsRoute`
- loading spinner: `MyOrgProjectsRoute`
- organization query: `MyOrgProjectsRoute`
- useEffect to navigate if org not MMO: `MyOrgProjectsRoute`
- export button: `CustomViewRoute`
- projectViewsStore: `CustomViewRoute`

magicznyleszek: I think it would be possible to create some shared component, either for just `MyProjectsRoute` + `MyOrgProjectsRoute`, or maybe for all three. I would need to separate dropzone, invites, and export button first, so that each of these could be a component to be imported

jamesrkiger: Sounds good, thank you. If you end up doing this, can you put the initial refactor in a separate PR to go along with our new PR guidelines?

magicznyleszek:
@jamesrkiger because this is built atop #5218 and requires #5251, I will update it and make it review-ready after either of them is merged.

jsapp/js/projects/myOrgProjectsRoute.tsx
```diff
@@ -0,0 +1,63 @@
...
+  // We need to load organization data to build the api url.
+  useEffect(() => {
+    if (orgQuery.data) {
+      setApiUrl(endpoints.ORG_ASSETS_URL.replace(':organization_id', orgQuery.data.id));
+    }
+  }, [orgQuery.data]);
+
+  // If user visits organization projects route and the organization is not MMO
+  // we redirect user out of here.
+  useEffect(() => {
```
jamesrkiger, line 41: I would prefer if we handled this redirect logic at the route level. There is some redirect functionality in the `ValidateOrgPermissions` component I overhauled recently. What would you think about me updating that to include a prop for redirect route. Then you could remove this useEffect and use that wrapper component [here](https://github.com/kobotoolbox/kpi/pull/5240/files#diff-862f1d1178ac9c0f744ac0714f9d16de4e3f313754d88168cc82970cd9e99f2bR31)? Let me know what you think.

magicznyleszek: Agree, let's do it at the route level :)

magicznyleszek: Note to self: will fix this after #5262 gets merged

PR #5241, opened 2024-11-07
feat(projectHistoryLog): log exports TASK-944

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Record project history logs for when users export an asset.

### 👀 Preview steps

Bug template:
1. ℹ️ Log in as a super user. Create and deploy a project
2. Add at least 1 submission to the project
3. Go to Data > Downloads and hit Export. Note: the export may get stuck in processing. This is due to an unrelated bug (https://www.notion.so/kobotoolbox/Exports-fail-with-ExportTask-matching-query-does-not-exist-1377e515f65480328422e0c8fcfed1c0) but doesn't affect this work since we want to log every attempt to export, not just successful ones
5. 🟢 Go to `api/v2/audit-logs/?q=metadata__asset_uid:<asset_uid> AND log_type='project-history'`
6. There should be a new PH log with action=`export` (usual metadata)

### 💭 Notes
Uses the AuditLoggedViewSet and `create_from_related_request` flows (similar to asset files, paired data, etc.)

### DIFF
--- a/kobo/apps/audit_log/audit_actions.py
+++ b/kobo/apps/audit_log/audit_actions.py
@@ -14,6 +14,7 @@ class AuditAction(models.TextChoices):
     DISABLE_SHARING = 'disable-sharing'
     DISCONNECT_PROJECT = 'disconnect-project'
     ENABLE_SHARING = 'enable-sharing'
+    EXPORT = 'export'
     IN_TRASH = 'in-trash'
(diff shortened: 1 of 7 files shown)


Guitlle (approved):
cool, now that the base code is in place, these new logs are straight forward

PR #5246, opened 2024-11-08
feat(organizations): always return `use_team_label` config value TASK-1167

jamesrkiger:
### 📣 Summary
Always return the configured Constance value of `use_team_label` regardless of Stripe status.

### 👀 Preview steps
No need for a preview, as this is covered in the unit tests for the environment API. 

### 💭 Notes
I also removed the unit tests that were assuming the API would tie this return value to Stripe status. The initial tests cover simple config returns.


noliveleger (approved):
LGTM

PR #5249, opened 2024-11-08
feat(organizations): update usage limits modals and banners TASK-996

pauloamorimbr:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Usage limits modal and banner descriptions were changed to reflect the individual, organization or team subscription.

### 📖 Description

Information for usage limits modal and banner was changed:
- The subject changes between "you", "your organization" and "your team" depending on the user subscription
- Texts were updated to fix translation problems, isolated in longer sentences (even if it makes the code look a bit redundant)
- A Bug was fixed where the warning limits would be displayed repeated times. (Was already fixed for non-warning messages)
- The `react-markdown` library was added to format markdown-like strings

### 👀 Preview steps

2. have Stripe [setup](https://www.notion.so/kobotoolbox/Setting-Up-Stripe-9a958a84d6324ec8a4ca887b820eb8cf)
2. Reach the usage limit or get close to the usage limit. Easiest way is to override values in [useUsage.hook.ts](https://github.com/kobotoolbox/kpi/blob/3ede0c8fc64b7ca863869b44d54e99cb818caa31/jsapp/js/account/usage/useUsage.hook.ts)
3. Open projects view.
4. 🟢 Banner should show up informing about the usage limit. The message will vary depending on the `is_mmo` flag from the organization and on the `shouldUseTeamLabel` from environment.
5. 🟢 For exceeded submissions limits a modal will show up having the same variants as the previous step.
6. 🟢 For Members of an MMO, the warning is not displayed
7. 🟢 For Admins of an MMO, the Call to action is not displayed
8. 🟢 For Owners (MMO or not) the call to action is displayed
9. 🟢 Each role, including being or not an MMO, has a different second part for the text.

Reference text/rules in can be seen on [Figma](https://www.figma.com/design/VIn6Fv2oEpqDHaGoqs6Zt9/%E2%9A%99%EF%B8%8F-KPI-%2F-Teams-(Organization)-Feature?node-id=1850-28029&node-type=frame&t=uvbnnHzIapMRo8v3-0)

### 💭 Notes

- The full idea behind some of the changes is to improve translation process, converting broken down texts to longer sentences with context and formatting where needed.
- At first a simpler markdown parser was implemented, but after some discussion in [its PR](https://github.com/kobotoolbox/kpi/pull/5257) we've decided to use `react-markdown` instead.
- We could now consider updating the places in Kobo where we're using HTML and it could be dangerous, so @magicznyleszek created a task for that refactor: [TASK-1257](https://www.notion.so/kobotoolbox/Refactor-some-dangerouslySetInnerHTML-code-to-use-react-markdown-13e7e515f65480c5a95ec9b1a7a7477c?pvs=4) 🙇🏻


jamesrkiger:
@pauloamorimbr I think this would probably benefit from being two PRs: One for the new tooling and one for the actual changes to the limit banners.

Also, how would you feel about putting together some unit tests for this? I see we've got a story in there (thanks for that), but this feels like the kind of thing it would be good to have actual tests for to make sure we catch edge cases.

jamesrkiger:
We need to filter out admins and members so that only owners see the over limit modal.

jamesrkiger:
This banner seems to have been moved to the bottom of the projects page. 
image

jamesrkiger:
Banner padding needs adjustment
image

jsapp/js/components/usageLimits/overLimitBanner.component.tsx
```diff
@@ -13,40 +20,166 @@ interface OverLimitBannerProps {
...
+  let message2 = '';
+  if (isMmo) {
+    if (userRole === OrganizationUserRole.owner) {
+      if (warning) {
+        message2 = t('Purchase additional submissions add-ons to continue collecting and submitting data.');
+      } else {
+        message2 = t('Please purchase an add-on to increase your submission limits.');
+      }
+    } else if (warning) {
+      message2 = t("Once the limit has been reached, you won't be able to collect or submit any new data until the team owner has purchased additional submissions.");
```
jamesrkiger, line 155: 'Team' shouldn't be hardcoded here. This text is already quite complex, though. I will leave it up to you whether you want to add in more logic or just cut out the term and say "until the owner".

jamesrkiger:
Would you mind taking one more look at the padding/margins here? The <p> element has its own margin so it's making the box bigger than it ought to be:
image

jamesrkiger (approved):
lgtm

PR #5251, opened 2024-11-11
refactor(projects): deduplicate code via UniversalProjectsRoute

magicznyleszek:
### 📣 Summary
Use single universal component for My Projects (`#/projects/home`) and Custom Project Views (`#/projects/<uid>`) routes.

### 💭 Notes
Move transfer project ownership code from `myProjectsRoute` to separate component - logic gathered in `ProjectOwnershipTransferModalWithBanner` component and banner in `ProjectTransferInviteBanner` component.

Create `UniversalProjectsRoute` component that gathers similar logic from both `MyProjectsRoute` and `CustomViewRoute`. To not increase complexity of the code deduplication is done under assumption that all routes wants (almost) all the same functionalities.

This change is done for upcoming organizations feature.


jsapp/js/components/permissions/transferProjects/inviteBanner.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+interface InviteBannerProps {
+  invite: InviteState;
+  onRequestClose: () => void;
+}
+
+/**
+ * Displays a banner about accepting or declining project transfer invitation.
+ */
+export default function InviteBanner(props: InviteBannerProps) {
```
jamesrkiger, line 23: For the organizations project we are going to have organization membership invites that will be very similar to project ownership transfer invites. I haven't thought about it too much, but I don't think we will reuse this component for that. With that in mind, do you think we should give the components in the transfer projects folder more specific names so we don't confuse them with other invite flows down the line? So in this case maybe something like ProjectTransferInviteBanner. It's a bit wordy, but might be good to avoid confusion.

jsapp/js/components/permissions/transferProjects/transferModalWithBanner.tsx
```diff
@@ -0,0 +1,74 @@
...
+} from './transferProjects.api';
+
+// Constants and types
+import type {InviteState} from './inviteBanner';
+
+/**
+ * This is a glue component that displays a modal from `TransferProjectsInvite`
+ * and a banner from `InviteBanner` as an outcome of the modal action.
+ */
+export default function TransferModalWithBanner() {
```
jamesrkiger, line 22: Similar to my comment about invitations above, we are going to have organization ownership transfers down the line. Should we rename this something like ProjectOwnershipTransferModalWithBanner?

PR #5253, opened 2024-11-11
feat(organizations): add badge to profile dropdown TASK-979

duvld:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Add the organization badge to the profile dropdown.

### 📖 Description

Tiny reshuffle of elements to make it match the designs.

### 👀 Preview steps

1. Have an MMO with some users in it
2. Ensure the feature flag `mmosEnabled` is true
3. While logged in with a user in an MMO, clicking the profile icon should show the organization badge (in blue) under the user's email


jsapp/js/components/header/organizationBadge.module.scss
```diff
@@ -1,12 +1,20 @@
...
   padding: 6px 10px;
   border-radius: 48px;
   font-weight: 600;
   font-size: .85em;
-  line-height: 12px;
   margin-right: 20px;
+  width: fit-content;
+}
+
+.header {
```
jamesrkiger, line 12: Since the only thing we are changing based on where this component is being used is color, let's just have the component take 'badgeColor' as a prop. That way if someone ends up using a light blue org badge elsewhere in the app they won't need to make a new class for it.

jamesrkiger (approved):
lgtm

PR #5257, opened 2024-11-12
refactor(TextFormatter): create TextFormatter component TASK-996

pauloamorimbr:
### 🗒️ Checklist

1. [x] run linter locally

### 👀 Preview steps
- The component is available for visualization and tests in storybook

### 💭 Notes

A `TextFormatter` component is being added to support long strings formatted in a markdown-like fashion.
Such component is needed to support long sentences to be properly translated without having to break out words just to add simple styles like bold, italic or links.
- The `TextFormatter` component suports an input of a `text` prop containing an string to be processed and generates the output based on the processing of the string.
- Formatting can be nested if needed

### DIFF
--- a/jsapp/js/components/textFormatter/textFormatter.component.tsx
+++ b/jsapp/js/components/textFormatter/textFormatter.component.tsx
@@ -1,0 +1,56 @@
+import type {ReactNode} from 'react';
+
+const regExpMatchParts = /[^*[\]]+|(\*\*[^*]*(?:\*[^*]+)*\*\*)|(\*[^*]+\*)|(\[.+?\]\(.+?\)({:.+})?)/g;
+
+const processText = (text: string): ReactNode[] => {
+  const parts = text.match(regExpMatchParts);
+
+  if (!parts) {return [text];}
+
+  const formattedParts: ReactNode[] = [];
+
+  for (const part of parts) {
+    if (part.startsWith('**')) {
+      formattedParts.push(<strong>{processText(part.slice(2, -2))}</strong>);
+    } else if (part.startsWith('*')) {
+      formattedParts.push(<em>{processText(part.slice(1, -1))}</em>);
+    } else if (part.startsWith('[')) {
+      const [label, link, target] = part
+        .slice(1, -1)
+        .split(/\]\(|\)\{:target=/);
+      formattedParts.push(
+        <a href={link} target={target ? target.slice(1, -1) : '_self'}>
+          {processText(label)}
+        </a>
+      );
+    } else {
+      formattedParts.push(part);
+    }
+  }
+
+  return formattedParts;
+};
+
+interface TextFormatterProps {
+  /**
+   * Text will be processed and formatted with markdown-like syntax. It accepts the following:
+   * - *<text>* for italic
+   * - **<text>** for bold
+   * - [text](link) for links with target _self
+   * - [text](link){:target="_blank"} for links with target _blank or other target
+   * - Formatting can be nested to be combined
+   */
+  text: string;
+}
+
+/**
+ * This component process text and applies simple formatting rules with markdown-like syntax.
+ * This is meant to be used with long sentences that need formatting and cannot be broken down
+ * into smaller components duo to translation issues that broken down components would cause.
+ * The formatting options mimics markdown syntax for italic, bold and links.
+ *
+ * @returns ReactNode[]
+ */
+export default function TextFormatter({text}: TextFormatterProps) {
+  return processText(text);
+}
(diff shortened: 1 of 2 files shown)


jsapp/js/components/textFormatter/textFormatter.component.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+const processText = (text: string): ReactNode[] => {
+  const parts = text.match(regExpMatchParts);
+
+  if (!parts) {return [text];}
+
+  const formattedParts: ReactNode[] = [];
+
+  for (const part of parts) {
+    if (part.startsWith('**')) {
```
jamesrkiger, line 13: Just a style issue, but we generally like to avoid if/else chains in favor of multiple simple if statements:

```
if (x) {
    return a
}
if (y) {
    return b
}
return c
```

jamesrkiger:
Left one minor comment, but other than that lgtm.

magicznyleszek:
@pauloamorimbr do we have plans to use it in all the places that currently go with `dangerouslySetInnerHTML`? I think there are around 11 of them (and not all apply).

The list of components that could use `TextFormatter`:
- `ActivityMessage`
- `SharingForm` (uses `replaceBracketsWithLink`)
- `TransferProjects`
- `AccessDenied` (uses `replaceBracketsWithLink`)
- `UniversalTable`

Also do we have plans to replace our old way of handling links (`replaceBracketsWithLink` from `textUtils.ts`) with this new way? :)

magicznyleszek:
Also I think that this should be discussed with @JacquelineMorrissette (unless it already was) given the fact that we are kinda setting some new rules for how translation text is written.

jsapp/js/components/textFormatter/textFormatter.component.tsx
```diff
@@ -0,0 +1,56 @@
+import type {ReactNode} from 'react';
+
+const regExpMatchParts = /[^*[\]]+|(\*\*[^*]*(?:\*[^*]+)*\*\*)|(\*[^*]+\*)|(\[.+?\]\(.+?\)({:.+})?)/g;
```
magicznyleszek, line 3: Wouldn't some existing library satisfy our needs here? Like https://www.npmjs.com/package/markdown-to-jsx or https://www.npmjs.com/package/react-markdown or https://www.npmjs.com/package/marked ?

Akuukis: I've used `react-markdown` before and suggest it. I like most the ability to override all components with custom implementation, that's useful especially with Link-s.

pauloamorimbr:
> Also I think that this should be discussed with @JacquelineMorrissette (unless it already was) given the fact that we are kinda setting some new rules for how translation text is written.

I did talk to @JacquelineMorrissette about this. 👍🏻

pauloamorimbr:
> @pauloamorimbr do we have plans to use it in all the places that currently go with `dangerouslySetInnerHTML`? I think there are around 11 of them (and not all apply).
> 
> The list of components that could use `TextFormatter`:
> 
> * `ActivityMessage`
> * `SharingForm` (uses `replaceBracketsWithLink`)
> * `TransferProjects`
> * `AccessDenied` (uses `replaceBracketsWithLink`)
> * `UniversalTable`
> 
> Also do we have plans to replace our old way of handling links (`replaceBracketsWithLink` from `textUtils.ts`) with this new way? :) Not sure if these two are really compatible, as `replaceBracketsWithLink` has a bit different functionality (there is no URL in the translation string, e.g. `foo [link] bar`, FE code provides the link) - but I think at least we should acknowledge one and the other with a comment. I haven't looked too deep into this :P

I didn't think of migrating other stuff. The idea is to apply this to the changes I'm working on right now.
If you think we could benefit from applying the same to other pieces of code I believe we could create another task to handle that. WDYT?

Regarding replacing the `replaceBracketsWithLink`, using markdown allow us to have better text for translation. I've taken a look at it and it's not exactly the same. I believe it falls in the same answer as above. We could create a new task to refactor what's being done currently. 
I didn't consider using HTML directly due to safety reasons and also because it's a less recognizable text for translatos...that's why I'm suggesting the new approach.

pauloamorimbr:
I've done tests with the `react-markdown` and it seems to work pretty well. For now I don't think we will benefit from wrapping it in our own component, so I will close this PR if no one disagrees in favor of just using the `react-markdown` on https://github.com/kobotoolbox/kpi/pull/5249 directly.

magicznyleszek:
@pauloamorimbr made a task for the refactor proposition: https://www.notion.so/kobotoolbox/Refactor-some-dangerouslySetInnerHTML-code-to-use-react-markdown-13e7e515f65480c5a95ec9b1a7a7477c?pvs=4

PR #5258, opened 2024-11-12
feat(accessLogExport)!: create new AccessLogExportTask to generate a csv of access logs TASK-871

RuthShryock:
### 👀 Preview steps

1. Ensure that the `Export all data` button for `ProjectViews` still works as normal and that the email header has the appropriate title: "Project View Report Complete"
2. Ensure that exports for submission data works correctly:
- make submissions to a project
- navigate to the _Project < Data < Downloads_ 
- 🟢 ensure the `Export` button is exporting data correctly
3. Preview the AccessLogExportTask from the shell:
- ℹ️ have a user 
- enter the Django shell: `./manage.py shell_plus`
- get your user: `test_user = User.objects.get(username='test')`
- create an AccessLogExportTask: 
``` 
task = AccessLogExportTask.objects.create(
           user=test_user,
           get_all_logs=False,
           data={'type': 'access_logs_export'},
  )
```
- 🟢 print the task object and notice that an AccessLogExportTask object has been created: `print(task)`
- run the task: `task.run()`
- 🟢 print the result and notice that it generated a link to the csv export: `print(task.result)`

### 💭 Notes
- Changed the field named `user` to `user_url` because naming it  `user` conflicts with an existing field in the `AccessLog` model
- Renamed the file titled `project_view_exports.py` to `data_exports.py` because it now generates exports for both project view and access log exports
- Refactored the `ProjectViewExportTask` to have a base `CommonExportTask` which both the project view and access log export classes inherit from 
- Updated the email (in `kpi/tasks.py`) to have the correct subject title based on the type of export report it is sending
- Alphabetized the functions in `data_exports.py`


jnm (changes requested):
A nitpick that I need to somehow get into our organization-wide development guidelines: `SubmissionsExport…` should be `SubmissionExport`. Another example of something similar would be that we use `AssetSerializer` instead of `AssetsSerializer`

jamesrkiger (changes requested):
We should figure out what is going on with the diff here (208 files changed) before we proceed with this. I'm sure whoever else looks at this will notice, but I'm marking it as change requested out of an abundance of caution.

RuthShryock:
@jamesrkiger Yes sorry I'm working on cleaning this up, something with my force push went wrong.

kpi/utils/data_exports.py
```diff
@@ -79,9 +94,72 @@
...
     },
+    'access_logs_export': {
+        'queryset': lambda: apps.get_model('audit_log', 'AccessLog').objects.all(),
+        'key': 'metadata',
+        'columns': ACCESS_LOGS_EXPORT_FIELDS,
+    },
 }
 
 
+def create_data_export(
```
rgraber, line 105: If we do it this way we'll have to add to this method every time we create a new kind of export task. I think it would make more sense to have this method take just take data and headers and create a csv from it. The AccessLogExportTask and ProjectViewExportTask classes would do whatever they had to do to get the data and then call this method with the data they collected.

kpi/utils/data_exports.py
```diff
@@ -98,12 +176,48 @@ def flatten_settings_inplace(settings: dict) -> None:
             settings[k] = ''
 
 
-def get_row_value(row: dict, col: str) -> Union[str, int, float, bool, None]:
-    val = row.get(col, '')
-    # remove any new lines from text
-    if isinstance(val, str):
-        val = val.replace('\n', '')
-    return val
+def get_data(filtered_queryset: QuerySet, export_type: str) -> QuerySet:
```
rgraber, line 179: now that we have more cases, we should just have each type of ExportTask have its own method for getting data so we don't just keep adding `elif`s for every new kind of export task.

rgraber (changes requested):
I think we can simplify this refactor a bit. 
If we make CommonExportTask a mixin we won't need to do the difficult migration.
Also, I think instead of  having a lot of methods that are like
```
do_thing(task):
    if task is project_view:
        do_stuff
    elif task is export:
        do other stuff
```
we can make the task classes themselves responsible for generating the data. 
Then `export_task_in_background` would look like (pseudocode)
```
model = get_model()
task = model.get(uid=uid)
task.run()
if status is complete:
      subject = model.default_email_subject
       ...get the result and send the email
```
and in CommonExportTask, `run_task` could do something like
```
data = self.get_data()
url = create_export_file()
return url
```
and `AccessLogExportTask` and `ProjectViewExportTask` would each have their own implementations of `get_data()` and a value for `default_email_subject`.
Usually inheritance is preferable to if-else chains because the code it's easier to change one subclass without changing methods that affect other subclasses.

kpi/migrations/0060_commonexporttask_remove_projectviewexporttask_data_and_more.py
```diff
@@ -0,0 +1,169 @@
...
+                        on_delete=django.db.models.deletion.CASCADE,
+                        to=settings.AUTH_USER_MODEL,
+                    ),
+                ),
+            ],
+            options={
+                'abstract': False,
+            },
+        ),
+        migrations.RunPython(populate_common_export_tasks),
```
rgraber, line 89: Forgot: if we end up keeping the migration, this needs to be reversible, even if the reverse is RunPython.noop. Otherwise unapplying migrations breaks

kpi/models/import_export_task.py
```diff
@@ -477,27 +488,26 @@ def export_upload_to(self, filename):
...
-    uid = KpiUidField(uid_prefix='pve')
-    result = PrivateFileField(upload_to=export_upload_to, max_length=380)
+class ExportTaskMixin:
+
+    @property
+    def default_email_subject(self) -> str:
+        return 'Report Complete'
+
+    def _get_export_details(self) -> tuple:
+        return self.data['type'], self.data['view']
```
rgraber, line 498: This should account for `data` not having a `view` key since we don't need one for access logs exports. the filename can just be `export_type-username-time` for tasks without a view.

kpi/models/import_export_task.py
```diff
@@ -477,27 +488,26 @@ def export_upload_to(self, filename):
...
 
     def _build_export_filename(
         self, export_type: str, username: str, view: str
     ) -> str:
         time = timezone.now().strftime('%Y-%m-%dT%H:%M:%SZ')
         return f'{export_type}-{username}-view_{view}-{time}.csv'
```
rgraber, line 504: Same as previous comment. This should be updated to allow an empty/None `view` parameter

kpi/models/import_export_task.py
```diff
@@ -510,7 +520,95 @@ def delete(self, *args, **kwargs) -> None:
...
+        if not self.get_all_logs:
+            queryset = queryset.filter(user__username=self.user.username)
+        data = self.get_data(queryset)
+        accessed_metadata_fields = [
+            'auth_type',
+            'source',
+            'ip_address',
+            'initial_user_username',
+            'initial_user_uid',
+            'auth_app_name',
```
rgraber, line 568:
```suggestion
            'authorized_app_name',
```

rgraber (changes requested):
some small things

kpi/tests/test_access_logs_export_task.py
```diff
@@ -0,0 +1,124 @@
...
+import os
+
+from django.conf import settings
+from django.contrib.auth import get_user_model
+from django.test import TestCase
+
+from kobo.apps.audit_log.models import AccessLog
+from kpi.models.import_export_task import AccessLogExportTask
+
+User = get_user_model()
```
rgraber, line 11: nit (non-blocking)
If you want to avoid this boilerplate, you can use `from kobo.apps.kobo_auth.shortcuts import User`

kpi/tests/test_access_logs_export_task.py
```diff
@@ -0,0 +1,124 @@
...
+            str(context.exception), 'Only superusers can export all access logs.'
+        )
+
+    def test_run_task_creates_csv(self):
+        task = self.create_export_task(self.superuser)
+        task.run()
+
+        self.assertIsNotNone(task.result, 'The task.result should not be None.')
+        expected_pattern = (
+            rf'{self.superuser.username}/exports/access_logs_export-'
```
rgraber, line 56: you are way more patient than I am to write this regex. respect.

kpi/utils/data_exports.py
```diff
@@ -79,9 +92,43 @@
         'key': METADATA,
         'columns': USER_FIELDS + METADATA_FIELDS,
     },
+    'access_logs_export': {
```
rgraber, line 95: We should order by `'-date-created` to be consistent with the endpoint

rgraber (changes requested):
Just the ordering thing and then I think we're good to go!

rgraber (approved):
LGTM

PR #5261, opened 2024-11-13
feat(organizations): members table TASK-980

magicznyleszek:
### 📣 Summary
Add MembersRoute component that renders things at `#/account/organization/members` (a header and a table of organization members). 

### 📖 Description
Add `membersQuery` that defines `OrganizationMember` TS interface and adds way for getting organization members.
Update `PaginatedQueryUniversalTableProps` to allow passing custom options to the hook.
Change minimum column size in UniversalTable props to 60 (smaller then previously).

### 👀 Preview steps

3. enable "Multi-members override" for joe's organization
6. navigate to `#/account/organization/members`
7. 🟢 notice that a table of organization members displays a list of users added in step 2

### 💭 Notes
Build atop #5235


jamesrkiger:
There is an issue with the layout when the horizontal page size gets too small:
image

jsapp/js/universalTable/paginatedQueryUniversalTable.component.tsx
```diff
@@ -9,12 +9,25 @@ import type {UseQueryResult} from '@tanstack/react-query';
 import type {PaginatedResponse} from 'js/dataInterface';
 import type {UniversalTableColumn} from './universalTable.component';
 
+/**
+ * A way of passing custom options, e.g. a parameter that needs to be included
+ * in the API url.
+ */
+export interface PaginatedQueryHookOptions {
```
jamesrkiger, line 16: Can we import `UseQueryOptions` from react-query instead?

jamesrkiger: We have come up with an alternative approach that doesn't require passing options.

jamesrkiger (approved):
lgtm. OK to merge into main after dependencies are merged.

PR #5272, opened 2024-11-14
fix(OpenRosa): correct detection of root node tag name for XForm surveys TASK-1258

noliveleger:
### 📣 Summary
Fixes the detection of the root node tag name of the survey in the XForm XML, respecting the name field in XLSForm settings

### 📖 Description
This bugfix addresses two issues caused by wrong detection of the root node tag name in the "survey" section of XForm XML: 
- an empty `formhub/uuid` on submission creation
- a 500 error when a form is open in Enketo (preview/collection) when a disclaimer is set

The update ensures that the name field, if provided in the XLSForm settings, is respected as the root node's tag name.

### 👀 Preview steps

Bug template:
1. Log in as a regular user
2. Create a project (add `name` field with custom value in the  `settings` worksheet)
3. Deploy and submit data
  1. 🔴 [on main] notice that this `formhub/uuid` is empty
  2. 🟢 [on PR] notice that this is `formhub/uuid` matches XForm uuid
  
1. Go to admin and add a disclaimer
2. Try to open the project in Enketo (preview and collection)
  1. 🔴 [on main] notice that it raises a 500 error
  2. 🟢 [on PR] notice that the form opens

### 💭 Notes
This PR relies on the fact that `id_string` and `name` are always provided in the XForm.json field (according to existing data on production servers).
In case of preview, If the `name` field is not supplied, the fallback mechanism uses the asset's UID to determine the tag name (as it is in `main`).


jnm (changes requested):
Great! Just a few minor things

jnm (approved):

tiritea:
let me know if you need help testing this

PR #5277, opened 2024-11-19
chore(eslint): allow leading underscore and in-between double underscore in snake_case

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### Description
Allow leading and trailing underscore (or double underscore) and in-between double underscore in snake_case. This is needed to not display warnings on things that we receive from API calls.


Akuukis (changes requested):
This rule allows it, but we shouldn't normally. The API is exception to the rule. I suggest to add linter ignore to the API file

PR #5280, opened 2024-11-19
feat(organizations): hide organization fields from settings for mmo organizations - TASK-978

pauloamorimbr:
### 📣 Summary
Organization fields are not displayed anymore in setting screens for MMOs.

### 👀 Preview steps
Feature/no-change template:
1. ℹ️ have account
2. Go to "settings" view at `/#/account/settings`
3. Use the feature flag `?ff_mmoEnabled=true`
4. Set the organization as MMO
5. 🟢 Fields should not be displayed
6. Set the feature flag to false as in `?ff_mmoEnabled=false`
7. 🟢 Fields should be displayed again

### 💭 Notes
The 'organization', 'organization_type' and 'organization_website' are now being hidden when the current organization has the `is_mmo` flag.

### DIFF
--- a/jsapp/js/account/accountSettingsRoute.tsx
+++ b/jsapp/js/account/accountSettingsRoute.tsx
@@ -20,7 +20,7 @@ import type {
 } from './account.constants';
 import {HELP_ARTICLE_ANON_SUBMISSIONS_URL} from 'js/constants';
 import {useSession} from '../stores/useSession';
-import {useOrganizationQuery} from './stripe.api';
+import {useOrganizationQuery} from './organization/organizationQuery';
 
 bem.AccountSettings = makeBem(null, 'account-settings', 'form');
 bem.AccountSettings__left = makeBem(bem.AccountSettings, 'left');
@@ -45,6 +45,7 @@ const AccountSettings = () => {
   const orgQuery = useOrganizationQuery();
 
   useEffect(() => {
+
     if (!currentLoggedAccount || !orgQuery.data) {
       return;
     }
--- a/jsapp/js/account/accountSettingsRoute.tsx
+++ b/jsapp/js/account/accountSettingsRoute.tsx
@@ -74,8 +74,8 @@ const AccountSettings = () => {
 
     const organization = orgQuery.data;
 
-    // We will not display organization fields if it's and MMO organization
-    // in favor of only displaying those fields in organization settings view
+    // We will not display organization fields if user is a member of an MMO,
+    // only displaying these fields in organization settings view
     setDisplayedFields(
       !organization?.is_mmo
         ? fieldKeys
--- a/jsapp/js/account/accountSettingsRoute.tsx
+++ b/jsapp/js/account/accountSettingsRoute.tsx
@@ -20,6 +20,7 @@ import type {
 } from './account.constants';
 import {HELP_ARTICLE_ANON_SUBMISSIONS_URL} from 'js/constants';
 import {useSession} from '../stores/useSession';
+import {useOrganizationQuery} from './stripe.api';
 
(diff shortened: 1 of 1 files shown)


jsapp/js/account/accountSettingsRoute.tsx
```diff
@@ -21,6 +22,8 @@ import type {
   AccountFieldsErrors,
 } from './account.constants';
 import {HELP_ARTICLE_ANON_SUBMISSIONS_URL} from 'js/constants';
+import {keys} from 'mobx';
```
jamesrkiger, line 25: Did this get added on accident? Looks unused.

jsapp/js/account/accountSettingsRoute.tsx
```diff
@@ -60,36 +67,62 @@ const AccountSettings = observer(() => {
...
   useEffect(() => {
     const currentAccount = sessionStore.currentAccount;
+
     if (
-      !sessionStore.isPending &&
-      sessionStore.isInitialLoadComplete &&
-      'email' in currentAccount
+      sessionStore.isPending ||
+      !sessionStore.isInitialLoadComplete ||
+      !('email' in currentAccount) ||
```
jamesrkiger, line 74: Since we're touching this section, I have to ask: Do we need these session store readiness checks? My impression is we could just use `'email' in currentAccount` (which I suspect is our way of making sure the store has a real user and not an anonymous user stub (extra credit if you find a better way to do that, too).

jamesrkiger:
@pauloamorimbr This isn't an issue with your code, but I am concerned about the way we are submitting the PATCH request in this form. It looks to me like we are basically just always sending the data for every form field. It's a PATCH request but we're essentially treating it like a PUT. This gets uglier as a result of the changes we are making here: Now we are submitting the org detail fields even though we aren't displaying them.

At the very least, I believe we need to stop submitting org data this way. If it didn't take too long, I would say it would be better if we only submitted data for "dirty" form fields (ones that have been edited by the user). But if that second goal is going to take too long perhaps we can ignore that for the moment. Can you spend 20 minutes or so to determine what the best approach is here?

pauloamorimbr:
> @pauloamorimbr This isn't an issue with your code, but I am concerned about the way we are submitting the PATCH request in this form. It looks to me like we are basically just always sending the data for every form field. It's a PATCH request but we're essentially treating it like a PUT. This gets uglier as a result of the changes we are making here: Now we are submitting the org detail fields even though we aren't displaying them.
> 
> At the very least, I believe we need to stop submitting org data this way. If it didn't take too long, I would say it would be better if we only submitted data for "dirty" form fields (ones that have been edited by the user). But if that second goal is going to take too long perhaps we can ignore that for the moment. Can you spend 20 minutes or so to determine what the best approach is here?

After the refactors, I'm back here...and now it behaves the proper way, sending only the edited field, thus not sending organization data since the fields are not visible for editing.

jsapp/js/account/accountSettingsRoute.tsx
```diff
@@ -60,8 +66,29 @@ const AccountSettings = () => {
...
-  }, [currentLoggedAccount]);
+    };
+
+    setFormFields(fields);
+
+    const fieldKeys = Object.keys(fields) as Array<keyof AccountFieldsValues>;
+
+    const organization = orgQuery.data;
+
+    // We will not display organization fields if it's and MMO organization
```
jamesrkiger, line 77: Typo: should be "an MMO organization".

It might be more accurate to say "We will not display organization fields if user is a member of an MMO ..."

jamesrkiger (approved):
Left one minor comment. Other than that, lgtm

PR #5281, opened 2024-11-20
feat(organizations): add functions for managing members TASK-985

magicznyleszek:
### 💭 Notes
Not testable per se until further PR is built atop this one and uses it :)


jsapp/js/account/organizations/membersQuery.ts
```diff
@@ -1,10 +1,21 @@
...
-import {fetchGet} from 'js/api';
+import {fetchGet, fetchPatch, fetchDelete} from 'js/api';
 import {QueryKeys} from 'js/query/queryKeys';
 import {useOrganizationQuery} from '../stripe.api';
 
+/**
+ * Note that it's only possible to update the role via API to either `admin` or
+ * `member`.
+ */
+export enum OrganizationMemberRole {
```
magicznyleszek, line 12: Note to self @magicznyleszek: there is already `OrganizationUserRole` with same-ish values

jsapp/js/account/organization/membersQuery.ts
```diff
@@ -34,6 +34,34 @@ export interface OrganizationMember {
...
+export async function removeOrganizationMember(
+  organizationId: string,
+  username: string
+) {
+  const apiUrl = endpoints.ORGANIZATION_MEMBER_URL
+    .replace(':organization_id', organizationId)
+    .replace(':username', username);
+  return fetchDelete(apiUrl);
+}
+
```
pauloamorimbr, line 64: We will probably need to add to these file the mutation hooks for the functions  and do not export the actual functions above.

``` typescript

/**
 * Mutation hook to patch and organization member.
 * This receives the current members query result to refetch the list after the
 * mutation is successful instead of using query keys duo to pagination parameters.
 */
export const usePatchOrganizationMember = (
  membersQuery: UseQueryResult<PaginatedResponse<OrganizationMember>>
) => {
  const orgQuery = useOrganizationQuery();
  const orgId = orgQuery.data?.id;

  return useMutation({
    mutationFn: ({
      username,
      data,
    }: {
      username: string;
      data: Partial<OrganizationMember>;
    }) => {
      if (!orgId) {
        throw new Error('Organization not found');
      }
      return patchOrganizationMember(orgId, username, data);
    },
    onSuccess: () => {
      // This will refetch the members list
      membersQuery.refetch();
    },
  });
};

/**
 * Mutation hook to remove an organization member.
 * This receives the current members query result to refetch the list after the
 * mutation is successful instead of using query keys duo to pagination parameters.
 */
export const useRemoveOrganizationMember = (
  membersQuery: UseQueryResult<PaginatedResponse<OrganizationMember>>
) => {
  const orgQuery = useOrganizationQuery();
  const orgId = orgQuery.data?.id;

  return useMutation({
    mutationFn: (username: string) => {
      if (!orgId) {
        throw new Error('Organization not found');
      }
      return removeOrganizationMember(orgId, username);
    },
    onSuccess: () => {
      // This will refetch the members list
      membersQuery.refetch();
    },
  });
};

```

Usage would be something like: 
``` typescript
const membersQuery = useOrganizationMembersQuery()
const membersPatch = usePatchOrganizationMember(membersQuery)

const onPatch = (username: string, data:Partial<OrganizationMember>) => {
    membersPatch.mutateAsync({username, data})
}
```

magicznyleszek: @pauloamorimbr Thansk for your guidance on this! I did it a bit differently though - not requiring the `membersQuery` to be passed, and using invalidation rather than refetch (see comment link)

magicznyleszek: @Akuukis wouldn't this be the options we want?:

```ts
/**
 * Mutation hook for updating organization member. It ensures that all related
 * queries refetch data (are invalidated).
 */
export function usePatchOrganizationMember(options?: Omit<UseMutationOptions<unknown, Error, PatchOrgMemberVars>, 'mutationFn' | 'onSettled'>) {
  const queryClient = useQueryClient();

  return useMutation({
    mutationFn: (vars: PatchOrgMemberVars) => patchOrganizationMember(vars),
    onSettled: () => {
      // We invalidate query, so it will refetch (instead of refetching it
      // directly, see: https://github.com/TanStack/query/discussions/2468)
      queryClient.invalidateQueries({queryKey: [QueryKeys.organizationMembers]});
    },
    ...options,
  });
}
```

Not arguing, just verifying if I understand correctly :) Changes:
- I removed the 4th `unknown` as I don't know why it's there.
- I changed the 1st to `unknown` as I don't use the mutation fn output in any way.
- I changed `'mutationFn'` to `'mutationFn' | 'onSettled'`, as both are predefined.

And general question - do we even need to have `options` here given the fact we are not using them? It adds some "noise" to the code. I don't have an opinion on this, just asking

Akuukis:
```ts
/**
 * Mutation hook for updating organization member. It ensures that all related
 * queries refetch data (are invalidated).
 */
export function usePatchOrganizationMember(defaultVars?: Partial<PatchOrgMemberVars>) {
  const queryClient = useQueryClient();

  return useMutation({
    mutationFn: (vars: PatchOrgMemberVars) => patchOrganizationMember({...defaultVars, vars}),
    onSettled: () => {
      // We invalidate query, so it will refetch (instead of refetching it
      // directly, see: https://github.com/TanStack/query/discussions/2468)
      queryClient.invalidateQueries({queryKey: [QueryKeys.organizationMembers]});
    },
    ...options,
  });
}
```

jsapp/js/account/organization/membersQuery.ts
```diff
@@ -34,6 +47,79 @@ export interface OrganizationMember {
...
+/**
+ * Mutation hook for updating organization member. It ensures that all related
+ * queries refetch data (are invalidated).
+ */
+export function usePatchOrganizationMember() {
+  const queryClient = useQueryClient();
+
+  return useMutation({
+    mutationFn: (vars: PatchOrgMemberVars) => patchOrganizationMember(vars),
```
p2edwards, line 79:
```suggestion
    mutationFn: patchOrganizationMember,
```

Paulo's suggestion

jsapp/js/account/organization/membersQuery.ts
Akuukis:
```ts
export default (props: {orgId}) => {
  const patchOrganizationMember = usePatchOrganizationMember({orgId})
  const handleCSubmit = useCallback((event) => patchOrganizationMember({name: event.target.value}), [])
  return <input onSubmit={handleCSubmit} />
  }

**Review by Akuukis — APPROVED** [review]
LGTM, not tested. I like how simple `getMemberEndpoint` approach is, for now let's stick to that.

Once we get to API schema auto-generation and auto-generated types, preferrably with auto-generated helpers, then let's worry later about what concerns and restrictions that will bring.

**Review by pauloamorimbr — APPROVED** [review]
LGTM! 👌🏻 

Just left a non blocking comment.

File: jsapp/js/account/organization/membersQuery.ts [hunk 4]
```diff
@@ -34,6 +47,62 @@ export interface OrganizationMember {
...
+  return endpoints.ORGANIZATION_MEMBER_URL
+    .replace(':organization_id', orgId)
+    .replace(':username', username);
+}
+
+/**
+ * Mutation hook for updating organization member. It ensures that all related
+ * queries refetch data (are invalidated).
+ */
+export function usePatchOrganizationMember(username: string) {
```
pauloamorimbr, line 60: In the hypothetical situation where we could need to call multiple `usePatch...` in the same view we might want to have the username as a parameter of the `mutationFn` itself instead of the hook.

But that's only to stay in our minds for the future. We can change this is needed.

pauloamorimbr (approved):
LGTM! Nice!!! 👏🏻

PR #5285, opened 2024-11-20
feat(organizations): auto-refresh on orgId change TASK-1247

Akuukis:
### 📣 Summary

On organization change auto-refresh the new organization for logged-in user.

### 👀 Preview steps

1. ℹ️ have an account and MMO enabled organization (belonging to another user)
2. log in, keep tab open
3. switch to admin panel, add account to the organization
4. switch back to account tab, notice "No Organization matches the given query." error toast.
5. 🔴 [on main] notice that toast keep going forever once a second or so.
6. 🟢 [on PR] notice that toast stops after 3-4 instances.


jamesrkiger (approved):
lgtm

PR #5289, opened 2024-11-21
feat(organizations): update is_mmo to identify mmo subscriptions TASK-1231

jamesrkiger:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Change the logic of the `is_mmo` property on the Organization model to check Stripe product metadata for the `'mmo_enabled'` field.

### 👀 Preview steps
This change is covered by a unit test.

For a manual test:
1. Create an organization and purchase a storage addon (this is a subscription in Stripe even if we don't describe it as such in the UI).
2. Check org's `is_mmo` property in shell. On `main`, this will be `True` because it is only checking to see whether or not an org has a plan. On this branch, however, it will correctly be `False` because storage addons shouldn't enable MMO features.
3. Upgrade to an enterprise plan, which includes MMO support. 
4. Check `is_mmo` property in shell. On this branch, it should now be `True`.


kobo/apps/stripe/tests/utils.py
```diff
@@ -30,14 +29,12 @@ def generate_plan_subscription(
         product_metadata = {**product_metadata, **metadata}
     product = baker.make(Product, active=True, metadata=product_metadata)
 
-    if not (price := Price.objects.filter(id=price_id).first()):
```
jamesrkiger, line 33: This was making it difficult to test an org with multiple subscriptions and I couldn't figure out why it was written this way with a hardcoded id. Removing it doesn't seem to have broken any unit tests.

kobo/apps/stripe/tests/test_organization_usage.py
```diff
@@ -438,31 +438,23 @@ def test_user_not_member_of_organization(self):
...
+    def test_aggregates_usage_for_mmo(self):
+        generate_mmo_subscription(self.organization)
         self.organization.add_user(self.newuser)
         # create 2 additional assets, one per user
         create_mock_assets([self.anotheruser, self.newuser])
         response = self.client.get(self.detail_url)
         assert response.status_code == status.HTTP_200_OK
         assert response.data['count'] == 2
 
-    def test_users_without_enterprise_see_only_their_usage(self):
```
jamesrkiger, line 458: With the introduction of the MMO feature, it is no longer possible to have an organization with multiple users that doesn't aggregate usage. Therefore, this test is irrelevant.

kobo/apps/stripe/tests/test_organization_usage.py
```diff
@@ -509,3 +501,36 @@ def test_get_suscription_limit_unlimited(self, usage_type):
...
+            self.organization, metadata=product_metadata
+        )
+        assert self.organization.is_mmo is False
+        subscription.status = 'canceled'
+        subscription.ended_at = timezone.now()
+        subscription.save()
+
+        product_metadata['mmo_enabled'] = 'true'
+        generate_plan_subscription(self.organization, metadata=product_metadata)
+        assert self.organization.is_mmo is True
```
noliveleger, line 536: Is there any case where - in the same request - `is_mmo` could change from False to True? 
I'm wondering if `cache_for_request` could be a problem and we would need to void the cache for request when new subscription is created?

kobo/apps/stripe/tests/utils.py
noliveleger: Probably because sometimes `Baker` uses super long strings and gets truncated when inserting data in the DB. 
So, hardcoding the ID could be helpful when you need to retrieve the object.

noliveleger (changes requested):

kobo/apps/stripe/tests/test_organization_usage.py
jamesrkiger: Currently, as far as I know the only way subscription status is going to change would be via Stripe webhooks. There is no way to change or cancel a subscription directly through our app. You have to go to the Stripe platform and make changes there, then those changes are reported to us via webhook.

kobo/apps/organizations/models.py
```diff
@@ -179,7 +179,14 @@ def is_mmo(self):
...
         """
-        return self.mmo_override or bool(self.active_subscription_billing_details())
+        if self.mmo_override:
+            return True
+
+        if billing_details := self.active_subscription_billing_details():
+            if product_metadata := billing_details.get('product_metadata'):
+                return product_metadata.get('mmo_enabled') == 'true'
+
+        return False
```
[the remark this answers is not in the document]
jamesrkiger: Good catch. I have updated the text.

kobo/apps/stripe/tests/test_organization_usage.py
```diff
@@ -509,3 +501,36 @@ def test_get_suscription_limit_unlimited(self, usage_type):
...
+        self.someuser = User.objects.get(username='someuser')
+        self.organization = self.someuser.organization
+
+    def test_is_mmo_subscription_logic(self):
+        self.organization.mmo_override = True
+        self.organization.save()
+        assert self.organization.is_mmo is True
+
+        self.organization.mmo_override = False
+        self.organization.save()
```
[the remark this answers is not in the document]
jamesrkiger: Yeah, I was probably being overly thorough here. We don't need these checks since they basically already happen in the tests in `test_organizations_api.py`. I have removed them so this test only checks the Stripe integration.

PR #5290, opened 2024-11-21
refactor(organizations): code cleanup and patch data cleanup in account settings and tos view TASK-1292

pauloamorimbr:
### 👀 Preview steps
2. Go into account settings view
3. Change information on some field(s)
4. Save the changes
5. Check in the network tab the call to the `/me` endpoint. The payload should only contain the changed fields.
6. Enable a TOS view by following [this document](https://www.notion.so/kobotoolbox/Setting-up-TOS-1447e515f65480f8ab34f69d42d806f8)
7. In the TOS view, fill in the information and save (you may want to `preserve log` before saving data)
8. Check the network tab. Payload should contain only the changed field.

### 💭 Notes
We're refactoring accountDetails and TOS view on the way of implementing some changes on displayed fields:
- Removed the "global" observer from mob-x which was causing linting issues in favor of using the `useSession` hook created at https://github.com/kobotoolbox/kpi/pull/5303
- Cleaned up the code
- Changed in a way that only changed fields are sent to the `PATCH` request to update data


jsapp/js/account/routes.constants.ts
```diff
@@ -14,7 +14,7 @@ export const AddOnsRoute = React.lazy(
   () => import(/* webpackPrefetch: true */ './add-ons/addOns.component')
 );
 export const AccountSettings = React.lazy(
-  () => import(/* webpackPrefetch: true */ './accountSettingsRoute')
+  () => import(/* webpackPrefetch: true */ './accountSettings.component')
```
magicznyleszek, line 17: @Akuukis what's our current idea for naming these?

Akuukis: Currently no idea, just stick to whatever is more consistent, simpler or faster. It's true that I'm about to suggest different naming in Linter Rules, BUT it's not there yet, so let's worry about it later. And I will try to get it done sooner.

pauloamorimbr: I'm reverting this for now. If needed I can rename this in another PR, or else, like @magicznyleszek mentioned, it's too difficult to review a whole new file

jamesrkiger (approved):
I left one comment, but otherwise lgtm

PR #5296, opened 2024-11-25
feat(organizations): org role logic for usage page plan update button TASK-999

jamesrkiger:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Adds logic to only display "See plans" button for organization owners. Role check is behind a feature flag.

### 👀 Preview steps

1. Create an MMO, add an admin.
2. View usage page with `ff_mmosEnabled=true` as admin. You should not see the "See plans" button.
3. View the usage page as organization owner. You should see the "See plans" button".

### DIFF
--- a/jsapp/js/account/usage/yourPlan.component.tsx
+++ b/jsapp/js/account/usage/yourPlan.component.tsx
@@ -14,6 +14,8 @@ import {
 import {ProductsContext} from '../useProducts.hook';
 import {getSubscriptionChangeDetails} from '../stripe.utils';
 import {ACCOUNT_ROUTES} from 'js/account/routes.constants';
+import {useOrganizationQuery} from '../organization/organizationQuery';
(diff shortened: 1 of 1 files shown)


jsapp/js/account/usage/yourPlan.component.tsx
```diff
@@ -69,6 +73,10 @@ export const YourPlan = () => {
     }
   }, [env.isReady, subscriptions.isInitialised]);
 
+  const showPlanUpdateLink = areMmosEnabled
```
pauloamorimbr, line 76: Thinking about this... it would help us to have the organization utils as a hook, even encapsulating the `useOrganizationQuery` and the feature flag verification so that we could have the conditions named as in:

``` typescript
const {organizationData, isOwner} = useOrganization()
```
WDYT?

pauloamorimbr (approved):
LGTM!
I left a non-blocking comment for an implementation idea to see WYT about it!

PR #5299, opened 2024-11-25
feat(organizations): add Organization Settings route TASK-981

magicznyleszek:
### 📣 Summary
In Organization Settings route (`#/account/organization/settings`) display editable organization name, and website. Also display non-editable organization type.

### 📖 Description
Only some roles allow editing the fields, and there is some logic in regards to fields being visible.

### 👀 Preview steps

1. for one of the users (e.g. "joe"), use http://kf.kobo.local/admin/organizations/organization/ to enable "Multi-members override" (for "joe's organization")
2. enable feature flag `mmosEnabled`
3. navigate to `#/account/organization/settings`
4. 🟢 notice that at least "Team name"/"Organization name" field is being displayed and is editable(-ish)
  - Editing the field doesn't do anything yet - this will be implemented in next PR
5. 🟢 if you have enabled Stripe - notice that "Team website"/"Organization website" field and "Team type"/"Organzation type" (non editable) fields are also being displayed

Next steps:
1. for another user (e.g. "zoe"), use http://kf.kobo.local/admin/organizations/organization/ to add "zoe" into "joe's organization"
2. logged in as "zoe", navigate to `#/account/organization/settings`
3. 🟢 notice that at least "Team name"/"Organization name" field is being displayed (non editable)
4. 🟢 if you have enabled Stripe - notice that "Team website"/"Organization website" field and "Team type"/"Organzation type" fields are also being displayed (both non editable)


jsapp/js/account/accountFieldsEditor.component.tsx
```diff
@@ -13,7 +13,7 @@ import type {
 } from './account.constants';
 
 // See: kobo/apps/accounts/forms.py (KoboSignupMixin)
-const ORGANIZATION_TYPE_SELECT_OPTIONS = [
+export const ORGANIZATION_TYPE_SELECT_OPTIONS = [
```
pauloamorimbr, line 16: IMO it's a bit off to have this exported from "accountFieldsEditor", maybe we should move this const somewhere else more directed to organizations?

jsapp/js/account/organization/OrganizationSettingsField.tsx
```diff
@@ -0,0 +1,36 @@
...
+  onChange?: (newValue: string) => void;
+  /**
+   * Function that ensures that field value is valid. If invalid will cause
+   * an error to be displayed.
+   */
+  validateValue?: (currentValue: string) => string | boolean | string[] | undefined;
+}
+
+/**
+ * A `TextBox` wrapper componet for `OrganizationSettingsRoute` that makes code
```
pauloamorimbr, line 18: Typo

```suggestion
 * A `TextBox` wrapper component for `OrganizationSettingsRoute` that makes code
```

jsapp/js/account/organization/OrganizationSettingsRoute.tsx
```diff
@@ -1,7 +1,156 @@
...
+
+  const mmoLabel = getSimpleMMOLabel(
+    envStore.data,
+    subscriptionStore.activeSubscriptions[0],
+    false,
+    true
+  );
+  const mmoLabelLowercase = mmoLabel.toLowerCase();
+
+  if (!orgQuery.data) {
```
pauloamorimbr, line 95: I believe we should use the loading flag instead.

```suggestion
  if (orgQuery.data.isLoading) {
```

jsapp/js/account/organization/OrganizationSettingsField.tsx
```diff
@@ -0,0 +1,36 @@
...
+) {
+  return (
+    <div className={organizationSettingsStyles.field}>
+      <TextBox
+        label={label}
+        value={value}
+        required
+        onChange={onChange}
+        disabled={!onChange || isDisabled}
+        errors={validateValue ? validateValue(value) : undefined}
```
pauloamorimbr, line 32: 🤔 

the logic here seems to be reversed, or the naming may be confusing. 

`validateValue` seems like it should return `true` if the field is valid.

Maybe we could just have the same `errors` prop as boolean being set in the parent and passed through?

jsapp/js/account/organization/OrganizationSettingsRoute.tsx
```diff
@@ -1,7 +1,156 @@
...
+
+/**
+ * Renders few fields with organization related settings, like name or website
+ * (with some logic in regards to their visibility). If user has necessary role,
+ * they can edit available fields.
+ */
 export default function OrganizationSettingsRoute() {
+  const orgQuery = useOrganizationQuery();
+  const [subscriptions] = useState(() => subscriptionStore);
+  const [state, setState] = useState<State>({name: ''});
```
pauloamorimbr, line 37: I'm not really a fan of this pattern of having an object in state. 
Usually when that's the case it's common to use `useReducer`, but in this case, just for a few fields I'd go with multiple states.

jsapp/js/account/organization/OrganizationSettingsRoute.tsx
```diff
@@ -1,7 +1,156 @@
...
+      <section className={styles.fieldsRow}>
+        <OrganizationSettingsField
+          label={t('##team/org## name').replace('##team/org##', mmoLabel)}
+          onChange={handleChangeName}
+          value={state.name}
+          validateValue={isNameValueValid}
+          isDisabled={!isUserAdminOrOwner || isPendingOrgPatch}
+        />
+
+        {isStripeEnabled && state.website && (
```
pauloamorimbr, line 116: For me, who doesn't have the `website` filled in yet, the field is not displayed for editing. Is this on purpose? 
Same thing for the `type`

jsapp/js/account/organization/OrganizationSettingsField.tsx
Akuukis: I've used to `validate*` returning list of errors because that's what form libs (e.g. [formik](https://formik.org/docs/guides/validation)) I worked with before did. This TextBox also displays error so that seems right to me. Do you have examples or suggestions for a better naming?

jsapp/js/account/organization/OrganizationSettingsRoute.tsx
Akuukis: Likely yes, but depends how you would handle query errors.

magicznyleszek: Good idea, but you've probably meant `orgQuery.isLoading` :)

jsapp/js/account/organization/OrganizationSettingsField.tsx
magicznyleszek: Changed :+1:

jsapp/js/account/organization/OrganizationSettingsRoute.tsx
```diff
@@ -1,7 +1,173 @@
...
+    return <LoadingSpinner />;
+  }
+
+  let deletionMessage = t('To delete this ##team/org##, please contact the server administrator.')
+    .replaceAll('##team/org##', mmoLabelLowercase);
+  if (isStripeEnabled) {
+    deletionMessage = t("To delete this ##team/org##, you need to cancel your current ##plan name## plan. At the end of the plan period your ##team/org##'s projects will be converted to projects owned by your personal account.")
+    .replaceAll('##team/org##', mmoLabelLowercase)
+    .replace('##plan name##', subscriptions.planName);
+  }
```
Akuukis, line 89: more of a linter thing, but don't use `let` and mutation if possible. Here one way to rewrite it
```suggestion
  let deletionMessage = isStripeEnabled
    ? t("To delete this ##team/org##, you need to cancel your current ##plan name## plan. At the end of the plan period your ##team/org##'s projects will be converted to projects owned by your personal account.")
      .replaceAll('##team/org##', mmoLabelLowercase)
      .replace('##plan name##', subscriptions.planName)
    : t('To delete this ##team/org##, please contact the server administrator.')
      .replaceAll('##team/org##', mmoLabelLowercase);
```
or 
```ts
  let deletionMessage = t(isStripeEnabled
    ? "To delete this ##team/org##, you need to cancel your current ##plan name## plan. At the end of the plan period your ##team/org##'s projects will be converted to projects owned by your personal account."
    : 'To delete this ##team/org##, please contact the server administrator.'
    )
      .replaceAll('##team/org##', mmoLabelLowercase)
      .replace('##plan name##', subscriptions.planName)
```

jsapp/js/account/organization/OrganizationSettingsRoute.tsx
```diff
@@ -1,7 +1,173 @@
...
+      */}
+      {isStripeEnabled && orgType && (
+        <section className={styles.fieldsRow}>
+          <KoboSelect
+            className={styles.fieldLong}
+            name='org-settings-type'
+            type='outline'
+            size='l'
+            isDisabled // always disabled
+            label={t('##team/org## type').replace('##team/org##', mmoLabel)}
```
Akuukis, line 148: Non-blocking styling question - why runtime `replace()` when string templates are simpler?

```suggestion
            label={t(`${mmoLabel} type`)}
```

Akuukis (approved):
LGTM, not tested, see 2 nitpicks inline.

jsapp/js/account/organization/OrganizationSettingsRoute.tsx
pauloamorimbr: > Non-blocking styling question - how it impacts translations using runtime `replace()` instead of simpler string templates?

I believe this is not styling. The translation function needs to be clear, no concatenations, so it can be identified for auto generation of translation strings. That's why we do a full sentence with a replace instead of a string template with code in it, which the `t` wouldn't recognize due to the dynamic string being passed to it.

pauloamorimbr (approved):
LGTM! 
Tested and now it works like a charm! 👏🏻

PR #5307, opened 2024-11-27
chore(attachmentStorage): Update sync management command to support no lock during execution

noliveleger:
### 📣 Summary
Added an option to the `update_attachment_storage_bytes` management command to prevent user submission suspension. Updated logic to handle the `submissions_suspended` flag conditionally.

### 📖 Description
This PR updates the `update_attachment_storage_bytes` management command to support a "no lock" mode. Previously, the command would set the `submissions_suspended` flag to True during its execution, preventing users from submitting data until the process was complete.
Introduced a new argument/option to skip toggling the `submissions_suspended` flag.
Default behavior (suspending submissions) remains unchanged to avoid breaking existing workflows.
When the "no lock" option is enabled, the command synchronizes attachment storage without blocking user submissions.

The `{"no_lock": true}` option can now be passed to the `sync_storage_counters` periodic task (if it was previously added).  
If `sync_storage_counters` has already been added as a periodic task, you should also add `fix_stale_submissions_suspended_flag` to ensure that accidentally locked accounts are unlocked.

### 💭 Notes
⚠️  Using `no_lock` option could result in discrepancies since submissions could continue to come in while the process is still counting storage usage.

### DIFF
--- a/kobo/apps/openrosa/apps/logger/constants.py
+++ b/kobo/apps/openrosa/apps/logger/constants.py
@@ -1,0 +1,2 @@
+
+SUBMISSIONS_SUSPENDED_HEARTBEAT_KEY = 'kobo:update_attachment_storage_bytes:heartbeat'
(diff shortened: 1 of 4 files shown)


rajpatel24 (approved):
LGTM

rajpatel24 (approved):
LGTM

PR #5309, opened 2024-11-28
feat(MembersRoute): add actions dropdown to the table of organization members TASK-987 TASK-990

magicznyleszek:
### 📣 Summary
In Members Table (from Account → Team/Organization → Members) add actions dropdown with one action. The action allows removing a member or leaving the organization.

### 📖 Description
The availability of "Remove" action is based on the role of the logged in user. When trying to remove user (or leave organization) a confirmation prompt is being shown.

### 👀 Preview steps
3. For one of the users (e.g. "sue") set the role to "admin"
4. enable "Multi-members override" for joe's organization
6. navigate to `#/account/organization/members`
7. 🟢 notice that a table of organization members displays a list of users added in step 2
8. 🟢 notice that in the table "joe" (as owner) has an ability to remove all organization members (excluding themselves) including "sue" who's an admin

As continuation for above steps:
1. log in as "sue"
2. navigate to `#/account/organization/members`
3. 🟢 notice that in the table "joe" (the owner) has no actions available
4. 🟢 notice that any "member" user in the table has actions available, and if you click "Remove" a confirmation prompt will appear
5. 🟢 notice that for "sue" the action available is "Leave team"/"Leave organization", and if you click it, a confirmation prompt will appear - and it will have different content than confirmation prompt for removing a member

As continuation for above steps:
1. pick a "member" user and use "Remove" action
2. confirm prompt
3. 🟢 verify that user was removed successfuly

### 💭 Notes
Build atop #5281


jsapp/js/account/organization/MemberRemoveModal.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+// Stores, hooks and utilities
+import {getSimpleMMOLabel} from './organization.utils';
+import envStore from 'jsapp/js/envStore';
+import subscriptionStore from 'jsapp/js/account/subscriptionStore';
+import {useRemoveOrganizationMember} from './membersQuery';
+import {notify} from 'alertifyjs';
+
+export const REMOVE_SELF_TEXT = {
+  title: t('Leave this ##team/org##'),
```
jamesrkiger, line 17: I think ##TEAM_OR_ORGANIZATION## won your recent poll, so should we change these?

jsapp/js/account/organization/MemberActionsDropdown.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+// Constants and types
+import {OrganizationUserRole} from './organizationQuery';
+
+// Styles
+import styles from './memberActionsDropdown.module.scss';
+
+interface MemberActionsDropdownProps {
+  /** Target member username. */
+  username: string;
```
jamesrkiger, line 24: Might be nice to rename this to something like `targetUsername` and then we won't need the comment.

jsapp/js/account/organization/MemberRemoveModal.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const removeMember = useRemoveOrganizationMember();
+  const mmoLabel = getSimpleMMOLabel(
+    envStore.data,
+    subscriptionStore.activeSubscriptions[0],
+    false,
+    false
+  );
+
+  // Choose proper text
+  let title = isRemovingSelf ? REMOVE_SELF_TEXT.title : REMOVE_MEMBER_TEXT.title;
```
jamesrkiger, line 71: 1. I think we could merge all this into something like `let textToDisplay = isRemovingSelf ? REMOVE_SELF_TEXT : REMOVE_MEMBER_TEXT`. 

2. Below, instead of calling replacePlaceholders() four times individually you could iterate over the object.

jsapp/js/account/organization/MemberActionsDropdown.tsx
```diff
@@ -0,0 +1,104 @@
...
+  let removeButtonLabel = t('Remove');
+  if (isAdminRemovingSelf) {
+    const mmoLabel = getSimpleMMOLabel(
+      envStore.data,
+      subscriptionStore.activeSubscriptions[0],
+      false,
+      false
+    );
+    removeButtonLabel = REMOVE_SELF_TEXT.confirmButtonLabel
+      .replace('##team/org##', mmoLabel);
```
jamesrkiger, line 69: I don't think we should be importing this text. It's very simple and I don't think there's any reason we need to force consistency between this button and the modal button.

jamesrkiger (changes requested):
Looks good, works good. I just left some code style comments.

jamesrkiger:
The dropdown here is running into trouble when it is near the bottom of the list:
image

(Same applies to the org member role selector, fyi)

jsapp/js/account/organization/MemberRemoveModal.tsx
magicznyleszek: I simplified this following your both of your advices

jsapp/js/account/organization/MemberRemoveModal.tsx
```diff
@@ -0,0 +1,130 @@
...
+import KoboModalFooter from 'jsapp/js/components/modals/koboModalFooter';
+
+// Stores, hooks and utilities
+import {getSimpleMMOLabel} from './organization.utils';
+import envStore from 'jsapp/js/envStore';
+import subscriptionStore from 'jsapp/js/account/subscriptionStore';
+import {useRemoveOrganizationMember} from './membersQuery';
+import {notify} from 'alertifyjs';
+
+interface TextToDisplay {
```
jamesrkiger, line 19: Is this interface necessary? I think we could skip it without TS complaining.

jsapp/js/account/organization/MemberRemoveModal.tsx
```diff
@@ -0,0 +1,130 @@
...
+/**
+ * @returns one of `TextToDisplay` objects with all placeholders replaced with
+ * proper values.
+ */
+function getTextToDisplay(
+  isRemovingSelf: boolean,
+  username: string,
+  mmoLabel: string
+) {
+  const text = clonedeep(isRemovingSelf ? REMOVE_SELF_TEXT : REMOVE_MEMBER_TEXT);
```
jamesrkiger, line 49: If we move the REMOVE... texts and the replace function into the component itself could we skip this cloning? No big deal if not, just thought I'd ask.

jamesrkiger (changes requested):
Left two minor comments

jamesrkiger (approved):
lgtm

PR #5315, opened 2024-12-02
feat(organizations): member role selector TASK-986

magicznyleszek:
### 📣 Summary
In Members Table (from Account → Team/Organization → Members) add UI for changing member's role.

### 📖 Description
Only owner and admins can modify the role.

### 👀 Preview steps
3. For one of the users (e.g. "sue") set the role to "admin"
4. enable "Multi-members override" for joe's organization
6. navigate to `#/account/organization/members` (as "joe")
7. 🟢 notice that in the table there is no way to change "joe" (owner) role
8. 🟢 notice that in the table you can change any other user role

As continuation for above steps:
1. log in as "sue"
2. navigate to `#/account/organization/members`
3. 🟢 notice (again) that in the table there is no way to change "joe" (owner) role (even though you're an "admin")
4. 🟢 notice that for "sue" you can change the role (to stop being an admin)

### 💭 Notes
Build atop #5309


jsapp/js/account/organization/MemberRoleSelector.tsx
```diff
@@ -0,0 +1,48 @@
...
+          label: t('Admin'),
+        },
+        {
+          value: OrganizationUserRole.member,
+          label: t('Member'),
+        },
+      ]}
+      selectedOption={role}
+      onChange={(newRole: string | null) => {
+        if (newRole !== null) {
```
jamesrkiger, line 40: Could we just do `if (newRole)` here? We don't want to allow empty strings anyways, right?

jamesrkiger (approved):
Left one minor comment. This is good to merge after reading and/or addressing it.

jamesrkiger (approved):
lgtm

PR #5318, opened 2024-12-03
feat(organizations): mutation hook for updating organization settings TASK-1219

magicznyleszek:
### 💭 Notes
Cleanup imports. Cleanup organization types typings (preparation for usage in other PR). Add `usePatchOrganization` mutation hook. Update comments.

### 👀 Preview steps
Not testable by itself - see #5318.

### DIFF
--- a/jsapp/js/account/organization/organizationQuery.ts
+++ b/jsapp/js/account/organization/organizationQuery.ts
@@ -1,19 +1,36 @@
-import type {FailResponse} from 'js/dataInterface';
-import {fetchGetUrl} from 'jsapp/js/api';
-import type {UndefinedInitialDataOptions} from '@tanstack/react-query';
-import {useQuery} from '@tanstack/react-query';
-import {QueryKeys} from 'js/query/queryKeys';
+// Libraries
+import {useMutation, useQuery, useQueryClient, type UndefinedInitialDataOptions} from '@tanstack/react-query';
+import {useEffect} from 'react';
+
+// Stores, hooks and utilities
+import {fetchGetUrl, fetchPatch} from 'jsapp/js/api';
 import {FeatureFlag, useFeatureFlag} from 'js/featureFlags';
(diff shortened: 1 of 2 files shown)


jsapp/js/account/organization/organizationQuery.ts
```diff
@@ -25,6 +42,22 @@ export enum OrganizationUserRole {
...
+/**
+ * Mutation hook for updating organization. It ensures that all related queries
+ * refetch data (are invalidated).
+ */
+export function usePatchOrganization(orgUrl: string) {
+  const queryClient = useQueryClient();
+  return useMutation({
+    mutationFn: async (data: Partial<Organization>) => (
+      fetchPatch<Organization>(orgUrl, data)
+    ),
```
Akuukis, line 54: await or return the fetchPatch to avoid an unhandled promise

jsapp/js/account/organization/organizationQuery.ts
```diff
@@ -25,6 +42,22 @@ export enum OrganizationUserRole {
   owner = 'owner',
 }
 
+/**
+ * Mutation hook for updating organization. It ensures that all related queries
+ * refetch data (are invalidated).
+ */
+export function usePatchOrganization(orgUrl: string) {
```
Akuukis, line 49: `orgUrl` seems like something that's a singleton and always available in the global state/cache, so let's simplify and grab it within the hook to avoid grabbing it in parent and passing around as parameter.

magicznyleszek: @Akuukis I tried adding this previously:

```ts
const session = useSession();
const orgUrl = session.currentLoggedAccount?.organization?.url;
```

But `orgUrl` is not always `string` (TypeScript says so), so what's the best approach here?

pauloamorimbr (approved):
LGTM! Good job! 💯

PR #5321, opened 2024-12-03
feat(OpenRosa): Block submissions when owning account is inactive TASK-1323

jnm:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Previously, projects owned by deactivated accounts could still receive submissions, e.g. if the project allowed anonymous submissions, or if a still-active user was granted access to submit to the project.

This change categorically prevents all submissions to projects owned by inactive users.

### 👀 Preview steps

1. Log in as a normal user (let's call them "Psy")
2. Create a project and deploy it
3. Set the project to allow anonymous submissions
4. Open Enketo, create a submission, and keep the Enketo tab open
5. In a separate browser session, use a superuser account to deactivate Psy's account (uncheck the "active" box in Django admin)
6. Go back to Enketo and attempt to submit
7. See submission fail with error message XXX

### DIFF
--- a/kobo/apps/openrosa/apps/api/tests/viewsets/test_xform_submission_api.py
+++ b/kobo/apps/openrosa/apps/api/tests/viewsets/test_xform_submission_api.py
@@ -18,7 +18,10 @@ from kobo.apps.openrosa.apps.logger.models import Attachment
 from kobo.apps.openrosa.libs.constants import (
     CAN_ADD_SUBMISSIONS
 )
-from kobo.apps.openrosa.libs.utils.logger_tools import OpenRosaTemporarilyUnavailable
+from kobo.apps.openrosa.libs.utils.logger_tools import (
+    OpenRosaResponseNotAllowed,
+    OpenRosaTemporarilyUnavailable,
+)
 
 
 class TestXFormSubmissionApi(TestAbstractViewSet):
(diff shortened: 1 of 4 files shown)


kobo/apps/openrosa/apps/api/tests/viewsets/test_xform_submission_api.py
```diff
@@ -458,6 +461,42 @@ def test_edit_submission_with_service_account(self):
                 response['Location'], 'http://testserver/submission'
             )
 
+    def test_submission_account_inactive(self):
```
jnm, line 464: Sorry this is so long :neutral_face: I copied `test_submission_blocking_flag()` and did slim it down somewhat

kobo/apps/openrosa/libs/utils/logger_tools.py
```diff
@@ -546,6 +547,8 @@ def safe_create_instance(username, xml_file, media_files, uuid, request):
         error = OpenRosaResponseNotAllowed(t("Form is not active"))
     except TemporarilyUnavailableError:
         error = OpenRosaTemporarilyUnavailable(t("Temporarily unavailable"))
+    except AccountInactiveError:
+        error = OpenRosaResponseNotAllowed(t('Account is not active'))
```
jnm, line 551: Seems kind of weird to return HTTP 405 for this, but I'm following the behavior of inactive (archived) forms as shown above

jnm:
@noliveleger said LGTM, but neither he nor @jnm could do anything without getting 500s from GitHub, so @pauloamorimbr saved the day

PR #5328, opened 2024-12-04
feat(accessLogsExport): add export button TASK-860

RuthShryock:
### 📣 Summary
Add an Export button to the Security page that allows users to initiate an access logs export.

### 👀 Preview steps

1. ℹ️ have account
2. navigate to _Account Settings < Security_ 
4. 🟢 notice that there is an "Export log data" button by the _Recent Account Activity_ 
5. click the button
6. 🟢 notice that you receive an email with a link to your export

### DIFF
--- a/jsapp/js/account/security/accessLogs/accessLogs.query.ts
+++ b/jsapp/js/account/security/accessLogs/accessLogs.query.ts
@@ -55,9 +55,8 @@ export default function useAccessLogsQuery(
  * Starts the exporting process of the access logs.
  * @returns {Promise<void>} A promise that starts the export.
  */
-export const startAccessLogsExport = (): Promise<void> =>
+export const startAccessLogsExport = () =>
   fetchPost(endpoints.ACCESS_LOGS_EXPORT_URL, {notifyAboutError: false})
-    .then(() => {})
     .catch((error) => {
       const failResponse: FailResponse = {
         status: 500,
--- a/jsapp/js/account/security/accessLogs/accessLogs.query.ts
+++ b/jsapp/js/account/security/accessLogs/accessLogs.query.ts
@@ -55,24 +55,18 @@ export default function useAccessLogsQuery(
  * Starts the exporting process of the access logs.
  * @returns {Promise<void>} A promise that starts the export.
  */
-const startAccessLogsExport = (): Promise<void> =>
+export const startAccessLogsExport = (): Promise<void> =>
   new Promise<void>((resolve, reject) => {
-
     fetchPost(endpoints.ACCESS_LOGS_EXPORT_URL, {})
       .then(() => {
-          resolve();
+        resolve();
       })
       .catch((error) => {
         const failResponse: FailResponse = {
           status: 500,
-          statusText: error.message || 'An error occurred while exporting the logs.',
+          statusText:
+            error.message || 'An error occurred while exporting the logs.',
         };
         reject(failResponse);
       });
   });
-
-/**
- * This is a hook to start the exporting process of the access logs.
- * @returns {() => void} The function to start the export
- */
-export const useExportAccessLogs = () => startAccessLogsExport;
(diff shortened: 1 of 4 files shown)


jsapp/js/account/security/accessLogs/accessLogsSection.component.tsx
```diff
@@ -4,15 +4,18 @@ import React from 'react';
...
 // Utilities
-import useAccessLogsQuery, {type AccessLog} from './accessLogs.query';
+import useAccessLogsQuery, {useExportAccessLogs, type AccessLog} from './accessLogs.query';
 import {formatTime} from 'js/utils';
 // import sessionStore from 'js/stores/session';
 
 // Styles
 import securityStyles from 'js/account/security/securityRoute.module.scss';
 
+const exportData = useExportAccessLogs();
```
pauloamorimbr, line 17: I believe it could be better to not use a hook here, since we're not getting any benefit from it being a hook, and just export the `startAccessLogsExport`. Then we don't even need it here, we could just use it directly in the button's `onclick`

pauloamorimbr (commented):
Left a comment to think about! 🤔

jsapp/js/account/security/accessLogs/accessLogs.query.ts
```diff
@@ -50,3 +50,23 @@ export default function useAccessLogsQuery(
...
   });
 }
+
+/**
+ * Starts the exporting process of the access logs.
+ * @returns {Promise<void>} A promise that starts the export.
+ */
+export const startAccessLogsExport = (): Promise<void> =>
+  new Promise<void>((resolve, reject) => {
+    fetchPost(endpoints.ACCESS_LOGS_EXPORT_URL, {})
```
magicznyleszek, line 60: I see you're handling errors here, so you should also pass options `{notifyAboutError: false}` to not have duplicate notifications :)

jsapp/js/account/security/accessLogs/accessLogs.query.ts
```diff
@@ -50,3 +50,23 @@ export default function useAccessLogsQuery(
     refetchOnWindowFocus: true,
   });
 }
+
+/**
+ * Starts the exporting process of the access logs.
+ * @returns {Promise<void>} A promise that starts the export.
+ */
+export const startAccessLogsExport = (): Promise<void> =>
+  new Promise<void>((resolve, reject) => {
```
magicznyleszek, line 59: Do I understand this correctly - `fetchPost` is returning a promise, but you need to wrap it in `new Promise` here to handle the error, right?

jsapp/js/account/security/accessLogs/accessLogs.query.ts
```diff
@@ -50,3 +50,23 @@ export default function useAccessLogsQuery(
...
+    fetchPost(endpoints.ACCESS_LOGS_EXPORT_URL, {})
+      .then(() => {
+        resolve();
+      })
```
magicznyleszek, line 63: Can this be `.then(resolve)`?

magicznyleszek (changes requested):

jsapp/js/account/security/accessLogs/accessLogs.query.ts
Akuukis: it can be simplified without wrapping, this should be equivalent:
```js
fetchPost(endpoints.ACCESS_LOGS_EXPORT_URL, {})
  .catch((error) => {
        const failResponse: FailResponse = {
          status: 500,
          statusText:
            error.message || 'An error occurred while exporting the logs.',
        };
        throw failResponse;
      });

File: jsapp/js/account/security/accessLogs/accessLogs.query.ts [hunk 5]
```diff
@@ -50,3 +50,18 @@ export default function useAccessLogsQuery(
...
+ * Starts the exporting process of the access logs.
+ * @returns {Promise<void>} A promise that starts the export.
+ */
+export const startAccessLogsExport = () =>
+  fetchPost(endpoints.ACCESS_LOGS_EXPORT_URL, {notifyAboutError: false})
+    .catch((error) => {
+      const failResponse: FailResponse = {
+        status: 500,
+        statusText:
+          error.message || 'An error occurred while exporting the logs.',
```
magicznyleszek, line 64:
```suggestion
          error.message || t('An error occurred while exporting the logs'),
```

magicznyleszek (approved):

PR #5329, opened 2024-12-05
Expose organization user created timestamp in members endpoint

rajpatel24:
### 📣 Summary
Expose organization user created timestamp instead of `date_joined` in members endpoint.

### 📖 Description
The created timestamp for organization users is now exposed in the members endpoint, providing a clear indication of when each user was added to the organization. This timestamp is formatted as `%Y-%m-%dT%H:%M:%SZ`.


noliveleger (approved):
LGTM

PR #5334, opened 2024-12-05
fix(sso): 500 on admin login

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Fixes a 500 being thrown when a non-superuser goes to /admin and then tries to log in as a superuser.

Upgrade django-allauth to 65.1.0.

### 💭 Notes
The simplest solution seemed to be just to upgrade django-allauth. It's unclear which exact commit fixes the bug and we are very behind in versions so upgraded to the latest. Nothing obvious broke except one unit test.

### 👀 Preview steps

Bug template:
1. ℹ️ Have 2 accounts, one super and one not
2. Log in as the non-superuser
3. Go to /admin
4. You should be redirected to the login page.
5. Log in as the super user
6. 🔴 [on main] 500 error
7. Checkout the PR branch. Make sure to reinstall requirements.
8. Do 1-5 again.
9. 🟢 [on PR] You should be correctly logged in and sent to the admin page


noliveleger (approved):
LGTM

PR #5343, opened 2024-12-11
feat(projectHistoryLogs): fetch data from API TASK-881 TASK-1156

pauloamorimbr:
### 💭 Notes
This PR fetches data from the backend:
- Some modifications in the `PaginatedQueryUniversalTable` were needed to accommodate passing extra data for the query, that affected existent queries. I tried making it a minimal impact though.
- Filter selector and export button actions will be finished in a future task

### 👀 Preview steps
2. Have the activity logs feature flag on with: `?ff_activityLogsEnabled=true`
3. Make some changes to the project, like editing the form and redeploying
4. Navigate to Project > Settings > Activity
5. 🟢 Changes made to the project should appear in the list

### DIFF
--- a/jsapp/js/api.endpoints.ts
+++ b/jsapp/js/api.endpoints.ts
@@ -1,5 +1,5 @@
 export const endpoints = {
-  ASSET_HISTORY: '/api/v2/assets/:asset_id/history/',
+  ASSET_HISTORY: '/api/v2/assets/:asset_uid/history/',
   ASSET_URL: '/api/v2/assets/:uid/',
   ORG_ASSETS_URL: '/api/v2/organizations/:organization_id/assets/',
   ME_URL: '/me/',
--- a/jsapp/js/api.endpoints.ts
+++ b/jsapp/js/api.endpoints.ts
@@ -1,4 +1,5 @@
 export const endpoints = {
+  ASSET_HISTORY: '/api/v2/assets/:asset_id/history/',
   ASSET_URL: '/api/v2/assets/:uid/',
   ORG_ASSETS_URL: '/api/v2/organizations/:organization_id/assets/',
   ME_URL: '/me/',
@@ -11,6 +12,7 @@ export const endpoints = {
   CHECKOUT_URL: '/api/v2/stripe/checkout-link',
   /** Expected parameter: organization_id  **/
   PORTAL_URL: '/api/v2/stripe/customer-portal',
+  PROJECT_HISTORY_LOGS: '/api/v2/project-history-logs/',
   /** Expected parameters: price_id and subscription_id **/
   CHANGE_PLAN_URL: '/api/v2/stripe/change-plan',
   ACCESS_LOGS_URL: '/api/v2/access-logs/me',
--- a/jsapp/js/components/activity/activity.constants.ts
+++ b/jsapp/js/components/activity/activity.constants.ts
@@ -174,7 +174,7 @@ export const AUDIT_ACTION_TYPES: AuditActionTypes = {
   // See: /kpi/kobo/apps/audit_log/views.py
   'update-form': {
     name: AuditActions['update-form'],
-    message: t('##username## edited the form in the formbuilder'),
+    message: t('##username## edited the form in the form builder'),
   },
   'add-user': {
     name: AuditActions['add-user'],
--- a/jsapp/js/components/activity/activity.constants.ts
+++ b/jsapp/js/components/activity/activity.constants.ts
@@ -3,33 +3,45 @@
  * @see `AuditAction` class from {@link kobo/apps/audit_log/models.py} (BE code)
  */
 export enum AuditActions {
+  'add-media' = 'add-media',
+  'allow-anonymous-submissions' = 'allow-anonymous-submissions',
+  'archive' = 'archive',
+  'connect-project' = 'connect-project',
+  'delete-media' = 'delete-media',
+  'delete-service' = 'delete-service',
+  'deploy' = 'deploy',
+  'disable-sharing' = 'disable-sharing',
+  'disallow-anonymous-submissions' = 'disallow-anonymous-submissions',
+  'disconnect-project' = 'disconnect-project',
+  'enable-sharing' = 'enable-sharing',
+  'export' = 'export',
+  'modify-imported-fields' = 'modify-imported-fields',
+  'modify-service' = 'modify-service',
+  'modify-sharing' = 'modify-sharing',
+  'modify-user-permissions' = 'modify-user-permissions',
+  'redeploy' = 'redeploy',
+  'register-service' = 'register-service',
+  'replace-form' = 'replace-form',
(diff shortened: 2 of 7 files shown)


jsapp/js/api.endpoints.ts
```diff
@@ -1,4 +1,5 @@
 export const endpoints = {
+  ASSET_HISTORY: '/api/v2/assets/:asset_id/history/',
```
magicznyleszek, line 2:
```suggestion
  ASSET_HISTORY: '/api/v2/assets/:asset_uid/history/',
```
I don't think we have such thing as asset **id** :)

jsapp/js/components/activity/activityLogs.query.ts
```diff
@@ -1,116 +1,85 @@
...
+}) => {
+  const params = new URLSearchParams({
+    limit: limit.toString(),
+    offset: offset.toString(),
   });
+  if (actionFilter) {
+    params.append('q', `action:${actionFilter}`);
+  }
+
+  const endpointUrl = endpoints.ASSET_HISTORY.replace(':asset_id', projectId);
```
magicznyleszek, line 33:
```suggestion
  const endpointUrl = endpoints.ASSET_HISTORY.replace(':asset_uid', projectId);
```
Matches the change from other suggestion

jsapp/js/components/activity/activityLogs.query.ts
```diff
@@ -1,116 +1,85 @@
...
+    {
+      errorMessageDisplay: t('There was an error getting one-time add-ons.'),
+    }
+  );
+};
 
 /**
  * Fetches the filter options for the activity logs.
+ *
+ * Filter options, for now, comes from ../api/v2/assets/[ assetId]]/history
```
magicznyleszek, line 46:
```suggestion
 * Filter options, for now, comes from `endpoints.ASSET_HISTORY`
```
I think this would be good enough. And if you want to keep the url string here, it should be exactly the same as the one from endpoints file

jsapp/js/components/activity/activityLogs.query.ts
```diff
@@ -1,116 +1,85 @@
...
+    'share-data-publicly',
+    'share-form-publicly',
+    'transfer',
+    'unarchive',
+    'unshare-data-publicly',
+    'unshare-form-publicly',
+    'update_content',
+    'update-name',
+    'update-settings',
+    'update-qa',
```
magicznyleszek, line 80: This array of strings seems like should be a `const` somewhere. And I think it would be great to locate the BE file that also has these and add the path to that BE file in the FE comment

jsapp/js/components/activity/activityLogs.query.ts
```diff
@@ -139,10 +108,21 @@ const startActivityLogsExport = async () =>
...
+  projectId,
+  actionFilter,
+}: PaginatedQueryHookParams) =>
   useQuery({
-    queryKey: [QueryKeys.activityLogs, itemLimit, pageOffset],
-    queryFn: () => getActivityLogs(itemLimit, pageOffset),
+    queryKey: [QueryKeys.activityLogs, projectId, actionFilter, limit, offset],
+    queryFn: () =>
+      getActivityLogs({
+        projectId: projectId as string,
```
magicznyleszek, line 121: Is `projectId` actually an `assetUid`? We always use `assetUid` name and never use `projectUid` or `projectId` or `assetId` (ok, the last one can be found in some ancient code).

jsapp/js/universalTable/paginatedQueryUniversalTable.component.tsx
```diff
@@ -8,13 +8,24 @@ import UniversalTable from './universalTable.component';
...
+
+interface PaginatedQueryHook<DataItem> {
+  (
+    params: PaginatedQueryHookParams
+  ): UseQueryResult<PaginatedResponse<DataItem>>;
 }
 
 interface PaginatedQueryUniversalTableProps<DataItem> {
   queryHook: PaginatedQueryHook<DataItem>;
+  queryHookData?: PaginatedQueryHookData;
```
magicznyleszek, line 28: I would love to see a comment explaining what `queryHookData` is and what is it's main purpose 🫶

jsapp/js/components/activity/formActivity.tsx
```diff
@@ -88,25 +95,28 @@ export default function FormActivity() {
...
               <pre>{JSON.stringify(detailsModalData, null, '  ')}</pre>
             </section>
           </KoboModal>
-        }
+        )}
 
         <PaginatedQueryUniversalTable<ActivityLogsItem>
           columns={columns}
           queryHook={useActivityLogsQuery}
+          queryHookData={queryData}
```
magicznyleszek, line 119: Tested doing some actions and all of them appear successfuly. There is problem with `ActivityMessage` not supporting all of them (see other review comment)

jsapp/js/components/activity/activityLogs.query.ts
magicznyleszek: Wait… Is this the same thing as the list from `jsapp/js/components/activity/activity.constants.ts`? It is! I think that list might require some updating (see other comment about finding BE file with the list of possible activities).

magicznyleszek: For example `allow-anonymous-submissions` is missing (as table displays fallback message "user x did action y" through `ActivityMessage` component)

magicznyleszek (changes requested):

jsapp/js/components/activity/activityLogs.query.ts
pauloamorimbr: I was trying to mean that it's documented in the API... maybe I need to rewrite better! 🤔

jsapp/js/components/activity/activityLogs.query.ts
pauloamorimbr: At least in code I didn't find a place in BE with constants. Even in the [documentation page](https://github.com/kobotoolbox/kpi/blob/0cc5cd147415ffdf4ffd58c87a568f87bb97eced/kobo/apps/audit_log/views.py#L424) where it's listed it's a fixed string. I'll compare and add the missing ones, and in the future we gonna get these values from BE for the filter. Jackie is creating a task for that.

jsapp/js/components/activity/formActivity.tsx
pauloamorimbr: Added everything that was missing, now we have more keys than the BE. We will need to find a way to sync these keys.

jsapp/js/components/activity/activity.constants.ts
```diff
@@ -3,33 +3,45 @@
  * @see `AuditAction` class from {@link kobo/apps/audit_log/models.py} (BE code)
```
pauloamorimbr, line 3: I've sorted the keys for better comparing with BE list.

magicznyleszek (approved):
Good to go :shipit:

PR #5349, opened 2024-12-12
feat(MembersRoute): fix table overflow issues TASK-1349

magicznyleszek:
### 🗒️ Checklist

1. [ ] run linter locally

### 👀 Preview steps

Bug template:
2. do this
3. do that

Feature/no-change template:
2. do this
3. do that
5. do that another thing

### DIFF
--- a/jsapp/js/account/organization/MemberRoleSelector.tsx
+++ b/jsapp/js/account/organization/MemberRoleSelector.tsx
@@ -25,6 +25,7 @@ export default function MemberRoleSelector(
       name={`member-role-selector-${username}`}
       type='outline'
       size='m'
+      placement='up-center'
       options={[
         {
           value: OrganizationUserRole.admin,
--- a/jsapp/js/account/organization/MemberActionsDropdown.tsx
+++ b/jsapp/js/account/organization/MemberActionsDropdown.tsx
@@ -83,7 +83,7 @@ export default function MemberActionsDropdown(
 
       <KoboDropdown
         name={`member-actions-dropdown-${targetUsername}`}
-        placement='down-right'
+        placement='middle-left'
         hideOnMenuClick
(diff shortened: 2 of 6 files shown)


jsapp/js/account/organization/MemberRoleSelector.tsx
```diff
@@ -25,6 +25,10 @@ export default function MemberRoleSelector(
       name={`member-role-selector-${username}`}
       type='outline'
       size='m'
+      // We assume that the row with the owner (who doesn't have
+      // `MemberRoleSelector`) will be first in the table. This means that all
+      // the other rows can open the selector up.
+      placement='up-center'
```
jamesrkiger, line 31: I'm not sure we can assume this, unfortunately. It doesn't appear to be true on my local organization, where I have a member appearing above the owner. Additionally, we need to account for multiple pages. Even if the owner was always the first in the list of members, if an org 11 members, for example, the second page of results will only have one member and they will not be the owner.

jamesrkiger: I think this is OK for now. There is still one annoying case where we run into trouble: An org of X + 1 members, where the table page size is set at X. Neither up nor down is going to work there. I am OK with this, however. We're really just doing a bandaid here pending implementation of a component library.

jamesrkiger (approved):
Left one comment about an issue that I don't know if it's worth fixing. This is OK to merge.

magicznyleszek:
@jamesrkiger I am triaging tasks for Orgs [4] and I see that the `MemberActionsDropdown` needs to be handled better here (now it has 1 option, but it would be possible to have 2 - and more in future). I will do similar thing as for `MemberRoleSelector`, so that whoevers works on Orgs [4] task doesn't need to fix this again

PR #5351, opened 2024-12-12
fix(permissions): restore edit functionality for users with partial permissions TASK-1366

noliveleger:
### 📣 Summary
Fixed an issue where users with partial permissions could no longer edit submissions they were allowed to.

### 📖 Description
Users with partial permissions faced an issue where they were unable to edit submissions they had proper access to. This fix resolves the problem, allowing these users to edit submissions within their permissions scope, while maintaining the integrity of restricted resources.

### 👀 Preview steps

Bug template:
1. Create a project
2. Give another user partial permissions: edit their data  
3. Submit data with the other user
4. Edit the data the other user
5. 🔴 [on release branch] Receive a 403 in enketo
6. 🟢 [on PR] Edit is successful


Guitlle (approved):
LGTM

PR #5356, opened 2024-12-13
fix(billing): adjust frontend service usage billing period handling TASK-1328

jamesrkiger:
### 📣 Summary
Adjusts frontend types and handling of data service/asset usage data according to the changes in #5326.

### 📖 Description
Previously, the ServiceUsageSerializer and AssetUsageSerializer returned usage information calculated according to a monthly billing cycle and a yearly billing cycle, regardless of what kind of billing cycle a user was actually subscribed to. So if a user was on a monthly plan, we made up a hypothetical yearly cycle and returned data for that in addition to the actual monthly cycle they are on. The frontend, in turn, selected the correct cycle information to display based on a separate request fetching the user’s subscription information. The hypothetical periods have been removed from the backend, which now returns only data for "current_period". This PR updates the frontend to handle this new data format.

### 💭 Notes
This currently targets the relevant backend branch so it can be tested, but I will retarget it to main after it is approved and the backend work (already approved) is merged.

### 👀 Preview steps
1. Create a new user on a Stripe-enabled instance

2. Navigate to usage page. Observe that user is on community plan with start date corresponding to account creation date. Displayed usage period should correspond to either first day of this month or last day of previous month (most likely the latter, though it currently depends on one’s timezone) and the last day of this month.

3. Purchase a monthly plan and sync djstripe Subscriptions, navigate to usage page. Observe that start date is still account creation date. Displayed usage period should now correspond to start date and one month after start date.

4. Cancel plan as the user, then cancel the plan via Stripe (or ask someone to do this), and sync subscriptions. Visit 
usage page and observe dates have not changed (because your community plan dates now correspond to date of cancellation and one month later).

5. Sign up for a yearly plan and sync subscriptions. Visit usage page and observe that start date remains the same, while displayed usage period ends a year from today.

### DIFF
--- a/jsapp/js/account/usage/usage.api.ts
+++ b/jsapp/js/account/usage/usage.api.ts
@@ -14,11 +14,7 @@ export interface AssetWithUsage {
   asset: string;
   uid: string;
   asset__name: string;
-  nlp_usage_current_month: {
-    total_nlp_asr_seconds: number;
-    total_nlp_mt_characters: number;
-  };
-  nlp_usage_current_year: {
+  nlp_usage_current_period: {
     total_nlp_asr_seconds: number;
     total_nlp_mt_characters: number;
   };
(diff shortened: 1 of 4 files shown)


pauloamorimbr (approved):
LGTM!

Codewise looks good!
For all I've tested the behavior is the expected and everything works. 👍🏻

PR #5357, opened 2024-12-13
feat(organizations): prevent creation of invites for ongoing transfers

noliveleger:
### 📣 Summary
Added validation to block the creation of invites for already existing transfers.

### 📖 Description
This update prevents the creation of duplicate invites when a transfer for the same resource already exists. By avoiding redundant invites, the system ensures better consistency and prevents confusion or potential conflicts arising from simultaneous transfers.


rajpatel24 (approved):
LGTM

PR #5358, opened 2024-12-13
feat(organizations): disable sorting for owner on projects table TASK-1336

RuthShryock:
### 📣 Summary
Removes option to sort by owner on project lists due to performance concerns.

### 👀 Preview steps

2. navigate to the projects table
3. 🔴[on main] notice that sorting by the **Owner** column is possible
4. 🟢 [on pr] notice that you can only hide the **Owner** column, no sorting is possible


jamesrkiger:
I think we should remove the arrow sort indicator from the owner column header here:
image

jamesrkiger:
This is how the permissions list appears to me as a non-member of "Ent Enterprise's":
image

That doesn't seem right.

jamesrkiger (changes requested):
Left a couple comments. 

Since the permission list fix looks a little more complicated than originally planned, let's move that fix to a separate PR.

PR #5362, opened 2024-12-17
feat(organizations): update owner filtering on projects table TASK-1335

RuthShryock:
### 📣 Summary
Enable filtering by both the organization name and the owner's name when selecting 'Owner' as a filter on the projects table. 

## 💭 Notes 
Known issues (Olivier's notes from Notion task)

Because we are searching on both field at the same time (it would be too difficult to do conditional search on whether the organization is MMO). Returned results could be false positive. 
Example: 
Tino sees two projects on in “My Project” list 

- Project 1 owns by olivierleger the user 
- Project 2 owns by jnm the user, but filled out the registration form with  “leger inc” as his the organization. 

If Tino searches for “owner” which should contain “leger”, both projects will appear, even if Project 2 shows “jnm” as the owner.

### 👀 Preview steps

Feature/no-change template:
1. ℹ️ have account (someuser) and several projects 
2. have user (kobouser) with organization name (Some Org) share a project with your account 
3. have  user (anotheruser) who is part of an MMO (Kobo Inc) share a project with your account 
4. 🟡 notice that projects by **_someuser_** and **_kobouser_** have the owner name as the owner label, but projects by **_anotheruser_** have **_Kobo Inc_** as the owner label (not part of this pr but important)
5. ➡️  filter by "Owner" contains "user"
6. 🟢 notice that all projects are shown
7. ➡️  filter by "Owner" contains "kobo"
8. 🟢 notice that projects from **_kobouser_** and **_Kobo Inc_** are shown
9. ➡️  filter by "Owner" contains "some"
10. 🟢 notice that projects from **_someuser_** and **_kobouser_** are shown. (Even though **_kobouser_** is shown as the owner name, we are filtering by the organization name as well, more on false positive in Notes above).


jsapp/js/projects/projectViews/utils.ts
```diff
@@ -69,6 +69,17 @@ export function buildQueriesFromFilters(filters: ProjectsFilterDefinition[]) {
     const fieldDefinition = PROJECT_FIELDS[filter.fieldName];
     const conditionDefinition = FILTER_CONDITIONS[filter.condition];
 
+    // Filtering by `ownerUsername` should filter on both the owner and organization name
+    if (filter.fieldName === 'ownerUsername') {
+      const ownerNameQuery = conditionDefinition.filterQuery
+        .replace('<field>', 'search_field__owner_name')
```
noliveleger, line 75: Should be `search_field__owner_username`

noliveleger (changes requested):

jsapp/js/projects/projectViews/utils.tests.ts
```diff
@@ -71,12 +71,10 @@ describe('projectViewsUtils', () => {
...
-          in: {
-            fieldName: 'ownerUsername',
-            condition: 'doesNotContain',
-            value: 'foo',
-          },
-          out: 'NOT owner__username__icontains:"foo"',
+          in: {fieldName: 'ownerUsername', condition: 'contains', value: 'foo'},
+          out:
+            '(search_field__owner_username__icontains:"foo" ' +
+            'OR search_field__organization_name__icontains:"foo")',
```
p2edwards, line 77: in this update, it looks like "owner does not contain" was removed and "owner does contain" was added.

- does the frontend support **exact match** on owner name? if so, should there be a test for this one, too?
- (less important) should we add back the does not contain test?

pauloamorimbr:
Frontend looks good. My only issue here is that I have a user called `dd` and 2 projects that belong to me, but when filtering by Owner I just can't make the projects appear. Either using `d`, `dd` or `me`, they never show up.

If it's expected behavior (like if it needs at least 3 characters or something like that), then it seems it's good to go.

pauloamorimbr (approved):
LGTM! 👏🏻

PR #5367, opened 2024-12-18
fix(billing): check default plan limits on backend TASK-1364

jamesrkiger:
### 📣 Summary
Updates backend code to check Stripe products for default "community plan" limits when determining organization usage.

### 📖 Description
Currently, the frontend gets billing usage limits for the free Community Plan from the corresponding Stripe product (returned by the products endpoint). The backend, however, does not reference these limits at all. This has not been a problem previously, because the backend currently does not enforce usage limits (only NLP limits are enforced, and this is handled by the frontend at present) and had no other need for these limits. 

Now that the backend is tracking addon usage, however, it does need to be aware of community plan usage limits. This PR updates the get_organization_plan_limit util function to use the limits from the metadata of the default community plan when an organization does not have a subscription.

### 👀 Preview steps
Change covered by unit test


Guitlle (approved):
good

PR #5379, opened 2024-12-18
feat(migrations): introduce long-running migration system using Celery

noliveleger:
### 📣 Summary
Added a new system for handling long-running migrations using Celery.

### 📖 Description
A new long-running migration system has been implemented, leveraging Celery to process migrations in the background. This system is designed to handle large-scale data updates that would otherwise cause require significant downtime with regular Django migrations

### 💭 Notes

This long-running migration system leverages Celery for asynchronous task processing. However, a similar (and likely more robust) feature is being developed in Django, as detailed in [DEP 14 - Background Workers](https://github.com/django/deps/blob/main/accepted/0014-background-workers.rst). Once Django officially releases its built-in background worker system, this custom solution will be phased out in favor of the native implementation to better align with the Django ecosystem.


kobo/apps/long_running_migrations/models.py
```diff
@@ -0,0 +1,85 @@
...
+
+        self.status = LongRunningMigrationStatus.COMPLETED
+        self.save(update_fields=['status'])
+
+    def save(self, **kwargs):
+
+        self.clean()
+
+        if self._state.adding:
+            file_path = os.path.join(
```
Guitlle, line 80: Isn't this missing the "/jobs/" subdir? it will look for the migration job file on the long_running_migrations folder.

kobo/apps/long_running_migrations/tasks.py
```diff
@@ -0,0 +1,41 @@
...
+            # clock drift between the Celery workers and the database, ensuring tasks
+            # are not prematurely skipped.
+            offset = 5 * 60
+            task_expiry_time = timezone.now() - relativedelta(
+                seconds=settings.CELERY_LONG_RUNNING_TASK_TIME_LIMIT + offset
+            )
+            # Run tasks that were just created or are in progress but have exceeded
+            # the maximum runtime allowed for a Celery task, causing Celery to terminate
+            # them and raise a SoftTimeLimitExceeded exception.
+            for migration in LongRunningMigration.objects.filter(
```
Guitlle, line 34: Should we have another status for skipping a job? Just to have more flexibility. Maybe it's a bad idea because then users can produce a bad db state, but what do you think?

kobo/apps/long_running_migrations/tasks.py
```diff
@@ -0,0 +1,41 @@
...
+                Q(status=LongRunningMigrationStatus.CREATED)
+                | Q(status=LongRunningMigrationStatus.IN_PROGRESS)
+                & Q(date_modified__lte=task_expiry_time)
+            ):
```
Guitlle, line 38: Should you explicitly sort the jobs? Not sure if they should be sorted by name, or date created, or maybe have a custom sorting column

Guitlle (changes requested):

kobo/apps/long_running_migrations/models.py
noliveleger: D'oh. Good catch!

kobo/apps/long_running_migrations/tasks.py
noliveleger: Indeed. Good idea ;-)

kobo/apps/long_running_migrations/tasks.py
noliveleger: Right, we can think about a more flexible process later in another PR but I would keep this one simple.

Guitlle (approved):

PR #5388, opened 2024-12-19
fix(organization): expose `owner_label` in ProjectView endpoint response TASK-1384

noliveleger:
### 📣 Summary
The Project view endpoint has been updated to include the owner label in its response. 

### 📖 Description
The Project view endpoint has been updated to include the owner label in its response. This fix ensures that the front end can retrieve the owner information directly from the endpoint, improving usability across the application.

### DIFF
--- a/kobo/apps/project_views/views.py
+++ b/kobo/apps/project_views/views.py
@@ -1,4 +1,3 @@
-# coding: utf-8
 from typing import Union
 
 from django.conf import settings
@@ -14,6 +13,7 @@ from kpi.filters import (
     AssetOrderingFilter,
     SearchFilter,
 )
+from kpi.mixins.asset import AssetViewSetListMixin
 from kpi.mixins.object_permission import ObjectPermissionViewSetMixin
 from kpi.models import Asset, ProjectViewExportTask
 from kpi.paginators import FastAssetPagination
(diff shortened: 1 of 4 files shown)


kobo/apps/project_views/views.py
```diff
@@ -134,9 +134,17 @@ def users(self, request, uid):
             queryset, serializer_class=UserListSerializer
         )
 
-    def get_serializer_context(self):
+    def get_serializer_context(self, data: list = None):
```
Guitlle, line 137: The type hint should be Optional[list]

kpi/mixins/asset.py
```diff
@@ -0,0 +1,33 @@
+from collections import defaultdict
+from django.db.models import Prefetch
+
+from kobo.apps.organizations.models import Organization
+from kpi.models.asset import Asset
+
+
+class AssetViewSetListMixin:
+
+    def get_organizations_per_asset(self, asset_ids: list) -> dict:
```
Guitlle, line 10: I'd prefer having a slightly more explicit function name like get_organizations_per_asset_id, because this implies you'll pass asset objects. Though it's not really important. What do you think?

Guitlle (changes requested):

Guitlle (approved):

PR #5390, opened 2024-12-19
fix: increase `FileField` maximum length to 380 characters

noliveleger:
### 📣 Summary
Increased the maximum length of FileField models from the default to 380 characters.

### 📖 Description
The maximum length for FileField objects has been increased to 380 characters (100 by default). This change allows longer file paths to be stored, ensuring compatibility with deeply nested directories or longer file names that might have been truncated previously.


kobo/apps/openrosa/apps/logger/migrations/0040_increase_file_fields_max_length.py
```diff
@@ -0,0 +1,39 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+        ('logger', '0039_populate_counters'),
+    ]
+
+    operations = [
+        migrations.AlterField(
```
Guitlle, line 19: What would happen if there exist values with more than 380 characters in this field?

Guitlle (approved):
Looks good, but not sure about the risk that the migrations can have

kobo/apps/openrosa/apps/logger/migrations/0040_increase_file_fields_max_length.py
noliveleger: they would be truncated, but default is 100. so we are safe

PR #5392, opened 2024-12-19
feat(organizations): remove mmos feature flag TASK-1397

pauloamorimbr:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Now the Organizations feature implemented up to this point is no longer hidden behind a feature flag

### 📖 Description
- `isMMosEnabled` feature flag was removed
- routes and menus hidden behind it are now enabled by default
- organizationQuery no longer returns fixed result for `is_mmo`

### 👀 Preview steps
Feature/no-change template:
1. ℹ️ have account
2. ℹ️ have or be part of an organization
3. ℹ️ Make sure to disable the active feature flag with `?ff_isMMosEnabled=false` or cleaning your session storage
4. Navigate to account settings
5. 🟢 notice that the organization links are visible
6. 🟢 notice that you can navigate and operate organization feature freely without the feature flag

### DIFF
--- a/jsapp/js/account/organization/organizationQuery.ts
+++ b/jsapp/js/account/organization/organizationQuery.ts
@@ -12,10 +12,15 @@ import {QueryKeys} from 'js/query/queryKeys';
 import {queryClient} from 'jsapp/js/query/queryClient';
 
 // Comes from `kobo/apps/accounts/forms.py`
-export type OrganizationTypeName = 'non-profit' | 'government' | 'educational' | 'commercial' | 'none';
+export type OrganizationTypeName =
+  | 'non-profit'
+  | 'government'
+  | 'educational'
+  | 'commercial'
+  | 'none';
 
 export const ORGANIZATION_TYPES: {
-  [P in OrganizationTypeName]: {name: OrganizationTypeName; label: string}
+  [P in OrganizationTypeName]: {name: OrganizationTypeName; label: string};
 } = {
   'non-profit': {name: 'non-profit', label: t('Non-profit organization')},
   government: {name: 'government', label: t('Government institution')},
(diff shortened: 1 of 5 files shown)


jsapp/js/account/organization/organizationQuery.ts
```diff
@@ -95,25 +93,12 @@ export const useOrganizationQuery = (params?: OrganizationQueryParams) => {
 
   // Using a separated function to fetch the organization data to prevent
   // feature flag dependencies from being added to the hook
-  const fetchOrganization = async (): Promise<Organization> => {
+  const fetchOrganization = async (): Promise<Organization> =>
     // `organizationUrl` is a full url with protocol and domain name, so we're
```
jamesrkiger, line 97: Just a style issue: I tend to think we should move all the doc text for fetchOrganization above it. It's essentially a single line of code broken for screen length purposes. Seeing that solo await without brackets around it feels weird.

jamesrkiger (approved):
left one minor comment on style, lgtm otherwise

jsapp/js/account/organization/organizationQuery.ts
p2edwards: So, combining both comments... Revisiting the existing comment above `fetchOrganization`, seems if we're removing the feature flag, we can simplify it even more by putting `fetchGet<...>` into queryFn below

pauloamorimbr: Thank you...nice catch. I visually ignored those comments and should've just removed everything at once. Just did it now. Also some changes to the query were required due to dependencies, since now we're using the `organizationUrl` inside of the query.

PR #5393, opened 2024-12-19
feat(organizations): handle user removes itself flow TASK-1354

pauloamorimbr:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
The navigation and data flow is improved when an MMO organization admin removes itself from the organization.

### 📖 Description
The objective of this PR is to:
- Minimize wrong API calls due to permissions and session data that needs to be updated
- Present a better flow for the user, redirecting them to the correct view after the removal
There's still a big margin for improving, mainly due to some work needed in `useUsage`, but it unfolds into several other files and flows, so it's better to be handled in a separated PR

### 👀 Preview steps
2. Be the admin in an MMO
3. Navigate to Account Settings > Members
4. Remove yourself from the organization
5. 🔴 [on main] notice that the user is not redirected and there are some failed attempts to API calls due to permissions
6. 🟢 [on PR] notice that the user is redirected and there's only 1 failed attempt to usage API

### DIFF
--- a/jsapp/js/account/organization/MemberRemoveModal.tsx
+++ b/jsapp/js/account/organization/MemberRemoveModal.tsx
@@ -12,8 +12,6 @@ import envStore from 'jsapp/js/envStore';
 import subscriptionStore from 'jsapp/js/account/subscriptionStore';
 import {useRemoveOrganizationMember} from './membersQuery';
 import {notify} from 'alertifyjs';
-import {ROUTES} from 'jsapp/js/router/routerConstants';
-import router from 'jsapp/js/router/router';
 
 interface MemberRemoveModalProps {
   username: string;
@@ -37,7 +35,7 @@ export default function MemberRemoveModal(
     onCancel,
   }: MemberRemoveModalProps
 ) {
-  const removeMember = useRemoveOrganizationMember({isRemovingSelf});
+  const removeMember = useRemoveOrganizationMember();
   const mmoLabel = getSimpleMMOLabel(
     envStore.data,
     subscriptionStore.activeSubscriptions[0],
--- a/jsapp/js/account/organization/MemberRemoveModal.tsx
+++ b/jsapp/js/account/organization/MemberRemoveModal.tsx
@@ -12,6 +12,8 @@ import envStore from 'jsapp/js/envStore';
 import subscriptionStore from 'jsapp/js/account/subscriptionStore';
 import {useRemoveOrganizationMember} from './membersQuery';
 import {notify} from 'alertifyjs';
+import {ROUTES} from 'jsapp/js/router/routerConstants';
+import router from 'jsapp/js/router/router';
 
 interface MemberRemoveModalProps {
   username: string;
@@ -35,7 +37,7 @@ export default function MemberRemoveModal(
     onCancel,
   }: MemberRemoveModalProps
 ) {
-  const removeMember = useRemoveOrganizationMember();
+  const removeMember = useRemoveOrganizationMember({isRemovingSelf});
   const mmoLabel = getSimpleMMOLabel(
     envStore.data,
     subscriptionStore.activeSubscriptions[0],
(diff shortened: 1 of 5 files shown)


p2edwards:
Skimmed, will test soon.

jsapp/js/account/organization/membersQuery.ts
```diff
@@ -84,22 +85,36 @@ export function usePatchOrganizationMember(username: string) {
...
       // query (`useOrganizationMembersQuery`) wouldn't be enabled without it.
       // Plus all the organization-related UI (that would use this hook) is
       // accessible only to logged in users.
-      fetchDelete(getMemberEndpoint(orgId!, username))
-    ),
+      return fetchDelete(getMemberEndpoint(orgId!, username));
+    },
     onSettled: () => {
-      queryClient.invalidateQueries({queryKey: [QueryKeys.organizationMembers]});
+      if (params?.isRemovingSelf) {
```
Akuukis, line 113: IMHO this hook can be simplified by removing the `isRemovingSelf` parameter and accessing variables within `onSettled` instead.
```ts
    onSettled: (_data, _error, username) => {
      if (session.currentLoggedAccount?.username === username) {
```

jsapp/js/account/organization/membersQuery.ts
```diff
@@ -84,22 +85,36 @@ export function usePatchOrganizationMember(username: string) {
...
+
       // We're asserting the `orgId` is not `undefined` here, because the parent
       // query (`useOrganizationMembersQuery`) wouldn't be enabled without it.
       // Plus all the organization-related UI (that would use this hook) is
       // accessible only to logged in users.
-      fetchDelete(getMemberEndpoint(orgId!, username))
-    ),
+      return fetchDelete(getMemberEndpoint(orgId!, username));
+    },
     onSettled: () => {
```
Akuukis, line 112: perhaps use `onSuccess` instead? It would be confusing to get logged out on a failure to remove oneself.

Akuukis (changes requested):
questions inline, didn't test. Also
- [ ] the title is confusing to have both "feat" and "fix" keywords in it, how about "feat(organizations): handle user removes itself flow" or similar?

jsapp/js/account/organization/MemberRemoveModal.tsx
```diff
@@ -12,6 +12,8 @@ import envStore from 'jsapp/js/envStore';
 import subscriptionStore from 'jsapp/js/account/subscriptionStore';
 import {useRemoveOrganizationMember} from './membersQuery';
 import {notify} from 'alertifyjs';
+import {ROUTES} from 'jsapp/js/router/routerConstants';
+import router from 'jsapp/js/router/router';
```
jamesrkiger, line 16: unused imports

jamesrkiger (changes requested):
Left a couple minor notes. On testing, I see the remaining issues from usage, but this is certainly an improvement. I will test again once you go through Kalvis' requested changes.

jsapp/js/account/organization/membersQuery.ts
```diff
@@ -84,22 +85,36 @@ export function usePatchOrganizationMember(username: string) {
...
-      queryClient.invalidateQueries({queryKey: [QueryKeys.organizationMembers]});
+      if (params?.isRemovingSelf) {
+        session.refreshAccount();
+      } else {
+        queryClient.invalidateQueries({queryKey: [QueryKeys.organizationMembers]});
+      }
```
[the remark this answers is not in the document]
pauloamorimbr: I'm adding that parameter there because the user, upon removal (self or not), will loose access to the members query, and react-query will try to refetch it, since it has done before, causing a error in the API call. 
I'm preventing that wrong call here by not running the invalidation. 

Right after this process the user is redirected to another view and the query is killed, but that only happens after the attempt to call the invalidation.

jsapp/js/account/organization/membersQuery.ts
pauloamorimbr: Really good one! 👏🏻

pauloamorimbr:
> Left a couple minor notes. On testing, I see the remaining issues from usage, but this is certainly an improvement. I will test again once you go through Kalvis' requested changes.

I changed it up a bit and seems like I got rid of the usage call on this.  🙌🏻 
Still, there are 2 tasks created related to this:
- migrate useUsage to react-query
- handle the flow for admin users who turn themselves into members

jamesrkiger (approved):
lgtm

PR #5405, opened 2025-01-02
refactor(library locking): migrate files to TypeScript

magicznyleszek:
### 💭 Notes
Migrating files to TS introduced some non-trivial changes, like:
- `enum` introduction required some parts of the code to be rewritten (more `if`'s etc.)
- mocks file had to be rewritten from scratch, as the asset responses there were extremely outdated and incomplete

### 👀 Preview steps
There is a nice long [Library Locking Technical Reference article](https://kobotoolbox.github.io/articles/library-locking-technical-reference) that I feel would be very helpful here :)

Some useful templates for testing:
- [locking template 1 (lock_all).xls](https://github.com/user-attachments/files/18306729/locking.template.1.lock_all.xls)
- [locking template 2 (profile, language_edit).xls](https://github.com/user-attachments/files/18306730/locking.template.2.profile.language_edit.xls)
- [locking template 3 (profile).xlsx](https://github.com/user-attachments/files/18306731/locking.template.3.profile.xlsx)

Testing:
1. Upload one of the above templates (__no drag&drop!__) using "My Library" > "NEW" > "Upload file" with "Upload as template" checked (__important!__ as without this, all locking features would be stripped by BE code).
6. Open the template in Form Builder
7. 🟢 notice that locking features work as expected (see Technical Reference linked above)
8. Go to "My Library".
9. For the locked template in the list use "Create project" button.
10. Open the newly created project in Form Builder.
11. 🟢 notice that locking features work as expected

There are three __fixes__ in this PR (_mea culpa_ for not splitting).

__First fix__: I changed how `isAssetLockable` works, by extending it to also include `template`s (previously only allowed `survey`s). This function works like a feature flag -ish, and I've noticed that with a template that is fully locked (has `lock_all: true`), the Form Builder sidebar wasn't all disabled.

__Second fix__: in the same area as above, Background audio setting was not being disabled, and it should be with `lock_all`. I suspect we've added locking before adding Background audio in there and forgot to include it.

__Third fix__: I no longer disabled "Add from Library" and "Layout & Settings" buttons for `lock_all` forms. There is an annoying tiny UX bug that was happening with current code:
1. Open a non-locked form in Form Builder and open "Layout & Settings" sidebar
2. The sidebar being opened is being remembered by FE code
3. Open a locked form in Form Builder and notice the sidebar is open
4. 🔴 Notice you can't close the sidebar, because the toggle button is disabled
5. 🟢 Notice that with this PR it is now possible to close sidebar (and everything inside the sideabar is already disabled, so no harm if user opens it)


Akuukis (approved):
LGTM, not tested. See 3 pushed commits - I started a review but quickly figured that it's much easier to show than tell. In short,
- 1st commit - fully used enum and ended up removing the object
- 2nd commit - don't mutate
- 3rd commit - see how `satisfies AssetResponse` instead of `: AssetResponse` yields more appropriate typings for the mocks, and eliminates all unnecessary type-guards.

pauloamorimbr (approved):
LGTM! 👌🏻

PR #5407, opened 2025-01-04
feat(invitation system): query file for handling API TASK-1092 TASK-991

magicznyleszek:
### 🗒️ Checklist

1. [ ] run linter locally

### 👀 Preview steps

Bug template:
2. do this
3. do that

Feature/no-change template:
2. do this
3. do that
5. do that another thing


jsapp/js/account/organization/membersInviteQuery.ts
```diff
@@ -0,0 +1,136 @@
...
+/*
+ * NOTE: `invites` - `membersQuery` holds a list of members, each containing
+ * an optional `invite` property (i.e. invited users that are not members yet
+ * will also appear on that list). That's why we have mutation hooks here for
+ * managing the invites. And each mutation will invalidate `membersQuery` to
+ * make it refetch.
+ */
+
+/*
+ * NOTE: `orgId` - we're asserting it is not `undefined` in code below,
```
Akuukis, line 23:
```suggestion
 * NOTE: `orgId` - we're assuming it is not `undefined` in code below,
```

jsapp/js/account/organization/membersInviteQuery.ts
```diff
@@ -0,0 +1,136 @@
...
+ * because the parent query (`useOrganizationMembersQuery`) wouldn't be enabled
+ * without it. Plus all the organization-related UI (that would use this hook)
+ * is accessible only to logged in users.
+ */
+
+/**
+ * This is the same list as `OrganizationInviteStatusChoices` from
+ * `kobo/apps/organizations/models.py`.
+ */
```
Akuukis, line 32:
```suggestion
/**
 * The source of truth of statuses are at `OrganizationInviteStatusChoices` in
 * `kobo/apps/organizations/models.py`. This enum should be kept in sync.
 */
```

Let's be clear on which of two is the source of truth

Akuukis (approved):
LGTM, not tested. Few doc nitpicks inline, and:
- use camel/pascal case in scope: "invitation system" -> "invitationSystem", or simply "invites"

jamesrkiger:
Given that none of this work is specific to the mantine POC and would be used even if we ended up not using mantine, should we just target `main` for this PR and then update the `mantine-setup` branch from main?

jsapp/js/api.endpoints.ts
```diff
@@ -3,6 +3,8 @@ export const endpoints = {
   ASSET_HISTORY_ACTIONS: '/api/v2/assets/:asset_uid/history/actions',
   ASSET_URL: '/api/v2/assets/:uid/',
   ORG_ASSETS_URL: '/api/v2/organizations/:organization_id/assets/',
+  ORG_INVITES_URL: '/api/v2/organizations/:organization_id/invites/',
```
jamesrkiger, line 6: 1. It might be a bit wordy, but given that we will also have org ownership invites, I'm inclined to say we should use the prefix `ORG_MEMBER_INVITE`

2. What would you think of something like `ORG_MEMBER_INVITE_DETAIL_URL` for the second url here? It would if the names were a bit more descriptive of the difference between the two endpoints beyond just singular vs. plural.

magicznyleszek:
Closing in favor of #5409 (the other PR has branch based in `main`)

PR #5412, opened 2025-01-08
feat(organizations): add theming/story for Mantine ActionIcon TASK-1448

jamesrkiger:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Adds theming and storybook entry for a custom ActionIcon component that automatically includes an appropriate sized icon.

### 💭 Notes
We previously used our custom in-house button component to display icon buttons. With Mantine, however, these kinds of buttons are rendered with the ActionIcon component.

### 👀 Preview steps
Use storybook stories and compare to existing icon buttons in the repo. Existing examples include:

- archive/share/delete options in project list view
- edit/delete options in user permissions section of sharing view
- member actions dropdown trigger in organizations member table


jsapp/js/theme/kobo/ActionIcon.ts
```diff
@@ -0,0 +1,38 @@
...
+      root: {
+        '--ai-size-sm': rem(28),
+        '--ai-size-md': rem(32),
+        '--ai-size-lg': rem(38),
+
+        ...(props.variant === 'filled' && {
+          '--ai-hover': theme.colors.blue[5],
+        }),
+        ...(props.variant === 'light' && {
+          '--ai-color': theme.colors.blue[5],
```
magicznyleszek, line 17: I think that the blue color of the icon is lighter than our current icon buttons.

jsapp/js/components/common/actionIcon.stories.tsx
```diff
@@ -0,0 +1,148 @@
...
+    },
+    size: {
+      description: 'Size of action icon',
+      options: actionIconSizes,
+      control: 'radio',
+    },
+    iconS: {
+      description: 'Icon',
+      options: Object.keys(IconNames),
+      mapping: generateSizedIconElementMap('s'),
```
magicznyleszek, line 55: There is something weird happening when I switch the size - the icon that I've set is gone and instead no icon or some other icon appears. It seems it's somehow connected to this function. It's not a bug per se, but it works differently from what we have

magicznyleszek (changes requested):

jsapp/js/theme/kobo/ActionIcon.module.css
```diff
@@ -0,0 +1,7 @@
+.root {
```
magicznyleszek, line 1: Shouldn't this file be `.scss`? I see that it does work in the build, but the code in it is not really CSS given the nesting and `&`

magicznyleszek (changes requested):

jsapp/js/theme/kobo/ActionIcon.ts
Akuukis: > let's prefer CSS Modules file over styles prop, because that edits CSS class once, instead of adding `style="font-size: calc(0.875rem * var(--mantine-scale)); border-top: 1px solid rgb(237, 238, 242);"` to every cell individually (see DOM of storybook). Also, that's officially recommended.
> 
> Furthermore, would you agree to prefer vars prop over CSS Modules file, because that's a simple parameterization of the CSS Modules? Although performance is unknown and official docs lacks an opinion on using vars prop vs "hardcoded" values in CSS Modules.

Preferring vars over css is an option, let's talk in the meetup about our policy (options: prefer vars, prefer css, undefined).

jsapp/js/theme/kobo/ActionIcon.module.css
Akuukis: Weird indeed, Mantine officially does the same, see https://mantine.dev/styles/css-modules/

jsapp/js/components/common/actionIcon.stories.tsx
jamesrkiger: Yeah, this is a little annoying but it's because currently button size and icon size are independent of each other in our code so this a side effect of how I'm enforcing "only small icons in small buttons", etc. in the Storybook controls.

jamesrkiger: Fixed with wrapped custom component.

jsapp/js/components/common/ActionIcon.tsx
```diff
@@ -0,0 +1,30 @@
...
+    const iconSize = props.size[0] as IconSize;
+    return (
+      <ActionIconMantine {...props} ref={ref}>
+        <Icon name={iconName} size={iconSize} />
+      </ActionIconMantine>
+    );
+  }
+);
+
+export default createPolymorphicComponent<'button', ActionIconProps>(
```
pauloamorimbr, line 28: Do we need this? 🤔 

Just wondering, since the default component for the icon is already the button, we don't need to create a polyformic...I believe that simply exporting the ActionIcon wiill do it

pauloamorimbr (approved):
LGTM! 

Left 1 simple comment with an opinion, but it's up to you.

jsapp/js/theme/kobo/ActionIcon.module.css
jamesrkiger: Planning to stick with this since it's how Mantine does it. We can always revisit later if we want to make a new rule for theme file types.

magicznyleszek (approved):

PR #5414, opened 2025-01-09
feat(organizations): add theming/story for Mantine Select TASK-1382 

pauloamorimbr:
### 🗒️ Checklist

1. [ ] run linter locally

### 📣 Summary
Theming for Mantine's Select was added to match our KoboSelect visual

### 📖 Description
- Story was added in storybook under Mantine/Select
- Theming was added for Select component
- Some "hacking" was needed on the CSS to change the arrow
- ⚠️ `clearable` is not working due to Theming setting the `rightSection` prop

Also:
- Members table role selector was updated to use the Select component as a POC
- Some logic was changed to be simplified and match the new implementation
- ⚠️ a test `LoadingOverlay` is being used in the members table to keep the user from doing other actions while processing the role change. The functionality seems good, but we need to setup the Loader theme as well for it to be used like that.

### 👀 Preview steps
- Use storybook to check the visual and basic functionality
- Check Organization's Members table. It's recommended to change other user's role for testing.


Akuukis (approved):
LGTM :ok_hand:  not tested

jsapp/js/theme/kobo/Select.module.css
```diff
@@ -0,0 +1,39 @@
...
+.options:hover .option[data-checked] {
+  background-color: transparent;
+}
+
+.option {
+  border-radius: 0;
+}
+
+.option:hover {
+  background-color: var(--mantine-color-blue-8) !important;
```
magicznyleszek, line 24: A comment why `!important` would be nice

jsapp/js/theme/kobo/Select.module.css
```diff
@@ -0,0 +1,39 @@
...
+    opacity: 1;
+  }
+}
+
+.dropdown {
+  border-color: var(--mantine-color-gray-6);
+  padding: 0;
+}
+
+.options:hover .option[data-checked] {
```
magicznyleszek, line 15: The options don't look identical to current `KoboSelect`, which is fine by me but ask @tesster7 what she thinks. What I have an issue with is the uneven spacing:
Screenshot 2025-01-10 at 16 04 44

jsapp/js/theme/kobo/Select.module.css
```diff
@@ -0,0 +1,39 @@
...
+  background-color: var(--mantine-color-blue-8);
+}
+
+.section[data-position='right'] {
+  display: flex;
+  width: auto;
+}
+
+.section,
+.section svg {
```
magicznyleszek, line 37: What is that svg? I thought it's the triangle icon, but it's not

jsapp/js/components/common/Select.stories.tsx
```diff
@@ -0,0 +1,80 @@
+import {Select} from './Select';
+import type {Meta, StoryObj} from '@storybook/react';
+
+/**
+ * Mantine [Select](https://mantine.dev/core/select/) component stories.
+ * See detailed uses in [Mantine's Select page](https://mantine.dev/core/select/)
+ */
+const meta: Meta<typeof Select> = {
+  title: 'Mantine/Select',
```
magicznyleszek, line 9: Shouldn't this be just `Common/Select`?

magicznyleszek (changes requested):

jamesrkiger:
Can we make it so the select closes when a user tabs to the next element? On the members table, if I open the select then tab to the member actions button, the select stays open and now I can't close it with 'esc'. I am noticing that this is also how the select behaves in Mantine examples, unfortunately, but I do think it's wrong and would be worth fixing if possible.

jsapp/js/theme/kobo/Select.module.css
pauloamorimbr: The `CloseButton` used for the `clearable` option uses an SVG for it.

jsapp/js/theme/kobo/Select.module.css
pauloamorimbr: Finally, after tons of researching and headbutting in the wall...I got to solve this by setting `ComboboxProps > dropdownPadding` to zero! 🤦🏻  ...learning curve!

pauloamorimbr:
> Can we make it so the select closes when a user tabs to the next element? On the members table, if I open the select then tab to the member actions button, the select stays open and now I can't close it with 'esc'. I am noticing that this is also how the select behaves in Mantine examples, unfortunately, but I do think it's wrong and would be worth fixing if possible.

I tried hard but I just wasn't able to fix this behavior. I came to a point where it would be better to just get Mantine's select code and change it's implementation to add our needed behaviors and have a better control to fix this. 😞 

We may end up doing just that in a future task.

pauloamorimbr:
> Weird issue here:
> 
> image
> 
> To reproduce:
> 
> 1. Open up the select component for an org member (member role specifically)
> 2. Use arrow down to navigate to member option
> 3. Move the mouse to highlight the admin option
> 4. Notice that the text color on 'member' remains white, so it basically disappears.
> 
> Obviously this is an edge case and I'm not sure if it's really a bug or if it's more a question of how our hover and focus style choices are interacting. Let me know what you think.

Just to mention that in geenral some styles were changed (this one you found was really a bug tho) so we aren't really following the current koboSelect styling in favor of following Tessa's indications.

jsapp/js/theme/kobo/Select.module.css
pauloamorimbr: removed in the new styling

jamesrkiger (approved):
Would appreciate it if could deprecate the old KoboSelect story. Other than that, lgtm

PR #5418, opened 2025-01-13
fix(auditLogs): correctly serialize audit logs from deleted users

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Fixes a 500 error from the various audit log endpoints when there are actions by deleted users.

### 📖 Description
Return empty user and username fields in the response if the user was deleted after the log was created. This applies to `/api/v2/audit-logs`, `api/v2/assets/<uid>/history`, and  `api/v2/project-history-logs`.

### 💭 Notes
Small fix in the serializer. Also updates the ProjectHistoryLog serializer to inherit from the AuditLogSerializer so we don't have to duplicate the method fields.

### 👀 Preview steps

Bug template:
1. ℹ️ have a super user account and a project
2. Create a new user (user1) and give them the `Edit Form` permission on the project.
3. Log in as user1 and make an edit to the project.
4. Log out user1 and log back in as the super user
5. Delete user1. You can do this from the admin page if you delete the user from the User list, then from the Trash Bin.
6. Go to:
  a. `api/v2/audit-logs`
  b. `api/v2/project-history-logs`
  c. `api/v2/assets/<uid>/history`
7. 🔴 [on main] All will return a 500 error (`AttributeError: 'NoneType' object has no attribute 'username'`)
8. 🟢 [on PR] The endpoint will return the expected logs. For all user1's actions, the user and username fields will be empty. but the user_uid should still refer to the old user.

### DIFF
--- a/kobo/apps/audit_log/serializers.py
+++ b/kobo/apps/audit_log/serializers.py
@@ -45,7 +45,7 @@ class AuditLogSerializer(serializers.ModelSerializer):
         return audit_log.date_created.strftime('%Y-%m-%dT%H:%M:%SZ')
 
     def get_username(self, audit_log):
-        return audit_log.user.username
+        return getattr(audit_log.user, 'username', None)
 
 
 class AccessLogSerializer(serializers.Serializer):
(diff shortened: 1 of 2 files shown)


noliveleger (approved):
LGTM

PR #5423, opened 2025-01-16
feat(projectHistoryLogs): log deleted submissions

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Create logs when submissions are deleted.

### 📖 Description
Each deletion is logged individually even if it was done in bulk. Unlike other actions, even failed deletions will be logged as long as the process of deletion was started, since some of the submissions may have been deleted before the failure. In this case, we prefer false positives to missing logs.

### 💭 Notes
The more robust way to handle this would probably be to redo the bulk-deletion process entirely such that each deletion is done as a single transaction with deletion from mongo, deleting docs from storage, creating a log. This would obviously take a lot more time and probably would work better as a celery task than an API endpoint.. At this moment we did not think it was worth it to rethink the whole process, and instead settled on just logging every *attempt* to delete, even if that attempt failed. This is in-keeping with our logging strategy so far, which has focused on intent more than effect.

For single deletes, this just adds a listener on the post_delete signal for Instances which attaches the necessary information to the request, similar to how individual updates were handled. 

### 👀 Preview steps
Feature/no-change template:
1. ℹ️ have 2 accounts and a project
2. Give user2 permission to add submissions to the project
3. Add at least 2 submissions as user1 and 2 as user 2
4. Enable anonymous submissions
5. Add at least 2 anonymous submissions 
6. Go to Project > Data and get the _ids of 1 submission each for user1, user2, and anonymous
7. For each submission, in the terminal, run `curl -X DELETE -H 'Authorization: Token <your token>' localhost/api/v2/assets/<asset_uid>/data/<submission _id>/ `
8. Go to `api/v2/assets/<asset_uid>/history`
9. 🟢 There should be a project history log with `action="delete-submission"` and the usual metadata, along with `"submission":{"submitted_by": <user1/user2/AnonymousUser>}`
10. Go back to Project > Data
11. Select one submission each for user1, user2, and anonymous
12. Hit the Delete button and confirm
13. 🟢 Go back to `api/v2/assets/<asset_uid>/history`. There should be three new project history logs (the same ones as were created with a single delete)
14. To test that we still get logs in case of a database timeout, use [mockobo](https://github.com/kobotoolbox/mockobo) to submit a 4000 submissions to your project (may want to do this in groups).
15. Go to Project > Data and select all submissions
16. Hit Delete and confirm
17. Before the request finishes (it should take a few seconds), kill your mongo server locally (`docker compose stop mongo` if you're using kobo-compose)
18. 🟢 Go back to `api/v2/assets/<asset_uid>/history`. The log count should have gone up by 4000. You can spot check a few to make sure they are deletion logs.
  * sometimes it takes a minute for the logs to show up because the front end times out before the backend does. you may need to reload a few times.

### DIFF
--- a/kobo/apps/audit_log/middleware.py
+++ b/kobo/apps/audit_log/middleware.py
@@ -1,7 +1,6 @@
 from rest_framework import status
 
 from kobo.apps.audit_log.models import AuditType, ProjectHistoryLog
-from kpi.utils import log
 
 
 def create_project_history_log_middleware(get_response):
@@ -23,9 +22,8 @@ def create_project_history_log_middleware(get_response):
         elif (
             log_type == AuditType.PROJECT_HISTORY
             and url_name == 'submission-bulk'
-            and request.METHOD == 'DELETE'
+            and request.method == 'DELETE'
         ):
-            log.info('Creating anyway')
             ProjectHistoryLog.create_from_request(request)
         return response
 
--- a/kobo/apps/audit_log/middleware.py
+++ b/kobo/apps/audit_log/middleware.py
@@ -1,6 +1,7 @@
 from rest_framework import status
 
 from kobo.apps.audit_log.models import AuditType, ProjectHistoryLog
+from kpi.utils import log
 
 
 def create_project_history_log_middleware(get_response):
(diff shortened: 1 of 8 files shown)


Guitlle (approved):
Looks good, I saw that it includes a few lines for openrosa Instance/xform delete signal. Can you mention that in the notes?

PR #5424, opened 2025-01-16
feat(organizations): add theming/story for Mantine TextInput TASK-1378

RuthShryock:
### 🗒️ Checklist

1. [X] run linter locally
2. [X] update all related docs (API, README, inline, etc.), if any
3. [X] draft PR with a title `<type>(<scope>)<!>: <title> TASK-1234`
4. [X] tag PR: at least `frontend` or `backend` unless it's global
5. [X] fill in the template below and delete template comments
6. [X] review thyself: read the diff and repro the preview as written
7. [X] open PR & confirm that CI passes
8. [X] request reviewers, if needed

### 📣 Summary
Add theming and storybook for custom TextInput component. 

### 💭 Notes
- Mantine has different [input types](https://v1.mantine.dev/guides/inputs/) that are all based on the same underlying Input and InputWrapper components. The purpose of this PR is to configure TextInput, but most of the theme work has been done in those two base components so we will hopefully not have to repeat it for other input types.
- This pr just includes the basic functionalities for a TextInput to work for in our organization task so more work will be needed in the future. 

### 👀 Preview steps

- Check out storybook and make sure that the TextInput is as similar to our TextBox component in terms of visuals and functionalities.


jsapp/js/components/common/textInput.stories.tsx
```diff
@@ -0,0 +1,114 @@
...
+    },
+    disabled: {
+      description: 'Disables the input',
+      control: 'boolean',
+    },
+    required: {
+      description: 'Marks the input as required',
+      control: 'boolean',
+    },
+    error: {
```
magicznyleszek, line 68: Nitpick: when input is disabled and has errors (case that shouldn't really happen 😅), the right side icon is red, left side icon is not.

jsapp/js/components/common/textInput.tsx
```diff
@@ -0,0 +1,61 @@
+import React, {useEffect, useState} from 'react';
+import {
+  TextInput as MantineTextInput,
+  type TextInputProps,
+} from '@mantine/core';
+import Icon from './icon';
+
+interface CustomTextInputProps extends Omit<TextInputProps, 'onChange'> {
```
magicznyleszek, line 8: I'm not sure if this is possible here, but our previous `TextBox` had an option to set the input type, e.g. "email" or "password".

jsapp/js/components/common/textInput.stories.tsx
```diff
@@ -0,0 +1,114 @@
...
+      description: 'Input value change callback',
+    },
+    size: {
+      description:
+        'Changes the size of the component (similar sizing as Button)',
+      defaultValue: {summary: 'lg'},
+      options: ['sm', 'md', 'lg'],
+      control: {type: 'radio'},
+    },
+    leftSection: {
```
magicznyleszek, line 33: When error message is set, left section icon is not red

jsapp/js/components/common/textInput.tsx
magicznyleszek: Also there was a possibility to set `onBlur` and `onKeyPress` previously - would it be possible with this one?

magicznyleszek: And `disableAutocomplete` :-D And `data-cy`

jsapp/js/components/common/textInput.stories.tsx
magicznyleszek: Also wanted to point out that with error the placeholder/value text is not red… but then I've checked `TextBox` and it's the way we did it. I kinda like the Mantine way more, with making the text red - what do you think about making it so?

jsapp/js/components/common/textInput.tsx
```diff
@@ -0,0 +1,61 @@
...
+      }
+      styles={(theme) => {
+        return {
+          input: {
+            borderColor: disabled
+              ? theme.colors.gray[2]
+              : error
+                ? theme.colors.red[7]
+                : theme.colors.gray[6],
+            color: disabled ? theme.colors.gray[2] : theme.colors.gray[1],
```
magicznyleszek, line 55: I find it confusing that we set the styles both here and in the css file. I would expect all of them to be put in theme file, given we have some dynamic-ish colors? Or everything in css file?

magicznyleszek (changes requested):

jsapp/js/components/common/textInput.stories.tsx
jamesrkiger: Updated error text color

jsapp/js/components/common/textInput.tsx
jamesrkiger: Per our Zulip discussion, I have not added types and we will have to configure NumberInput, TextArea, etc. components in the future. We can figure out disabling autocomplete when we implement the PasswordInput component.

Regarding event props, these should still be accessible (though I think Mantine gives us "onKeyDown" and "onKeyUp", rather than just keypress). I figure we can sort out any issues with these when we actually need them. 

For data-cy, Mantine allows adding data attributes with the [mod prop](https://mantine.dev/styles/data-attributes/#mod-prop). We might have to experiment with this to pass the attribute to the right place, but it seems doable in theory. I don't think we should spend time on it right now, though.

jsapp/js/theme/kobo/Input.module.css
```diff
@@ -0,0 +1,3 @@
+.input:focus-visible:not([data-error]) {
```
jamesrkiger, line 1: I moved all token overrides into the theme file, but this seemed best to leave in css even if it's short.

pauloamorimbr:
In the docs they suggest to use CSS modules due to performance instead of the vars approach, and also I believe it's clearer to set the CSS directly instead of setting the variables. Just my opinion.

I've created a draft PR with some changes fr us to talk about when possible: https://github.com/kobotoolbox/kpi/pull/5444/files

jsapp/js/components/common/textBox.stories.tsx
```diff
@@ -16,7 +16,7 @@ const textBoxTypes: TextBoxType[] = [
 const textBoxSizes: TextBoxSize[] = ['s', 'm', 'l'];
 
 export default {
-  title: 'common/TextBox',
+  title: 'commonDeprecated/TextBox',
```
magicznyleszek, line 19: Nitpick: should we add `@deprecated` tag to `TextBox.tsx`?

jsapp/js/components/common/textInput.tsx
```diff
@@ -0,0 +1,40 @@
...
+// so we don't need to repeat this code for other input types
+const TextInput = forwardRef<HTMLInputElement, TextInputProps>(
+  ({leftIconName, rightIconName, ...others}, ref) => {
+    const leftSection = leftIconName ? (
+      <Icon name={leftIconName} size='s' />
+    ) : null;
+    const rightSection = others.error ? (
+      <Icon name='alert' size='s' />
+    ) : rightIconName ? (
+      <Icon name={leftIconName} size='s' />
```
magicznyleszek, line 26:
```suggestion
      <Icon name={rightIconName} size='s' />
```

magicznyleszek (changes requested):

jsapp/js/components/common/textBox.stories.tsx
jamesrkiger: Marked custom TextBox component as deprecated

jamesrkiger:
@magicznyleszek @pauloamorimbr I have updated this PR after working with Paulo on a better approach:
https://github.com/kobotoolbox/kpi/pull/5424/commits/4d9a78b5b9797c86a3d69e57020ddacda62e5fa9

I removed the wrapped component. The main reason I used a wrapped component was for replacing right-side icons with an error icon when the input has errors. But our [designs](https://www.figma.com/design/UmFnSLNikt6aJTo7dEfERm/%E2%9A%99%EF%B8%8F-Kobo-KPI-Design-System?node-id=410-1896&m=dev) don't actually include that right-side error icon, so I decided to stick with a simpler implementation.

jsapp/js/theme/kobo/InputWrapper.ts
pauloamorimbr, line 1: Seems like this file is not being used anymore! 🤔

jsapp/js/theme/kobo/InputBase.ts
```diff
@@ -0,0 +1,17 @@
...
+      classNames: {
+        label: classes.label,
+      },
+    },
+    classNames: {
+      input: classes.input,
+      section: classes.section,
+    },
+  },
+});
```
pauloamorimbr, line 17: Lint's asking for a newline here! 🤷🏻 

```suggestion
});

```

pauloamorimbr (changes requested):
2 really simple fixes  and it seems ready to go! 💯

PR #5425, opened 2025-01-17
feat(organizations): add theming/story for Mantine Loader and LoadingOverlay TASK-1464

pauloamorimbr:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Theming for Mantine's Loader and LoadingOverlay was added to match Kobo's current LoadingSpinner for general purpose.

### 📖 Description
Both 'regular' and 'big' variant were added with its respective type names to be used with Mantine's `Loader` and `LoadingOverlay` components.
To simplify the implementation I decided to simply use the existing `LoadingSpinner` component instead of extracting it's contents.
Since Mantine's `Loader` doesn't have the `message` parameter, right now I'm fixing it on `false` so it won't appear. We may need to implement a Wrapper in the future to add a message, depending on our future needs.

* `LoadingOverlay` is already in use at https://github.com/kobotoolbox/kpi/pull/5414

### 👀 Preview steps
- Use storybook to check the visual of the component and compare with the existing loader in use in the app.

### DIFF
--- a/jsapp/js/components/common/loader.stories.tsx
+++ b/jsapp/js/components/common/loader.stories.tsx
@@ -1,0 +1,37 @@
+import {Center, Loader} from '@mantine/core';
+import type {Meta, StoryObj} from '@storybook/react';
+
+/**
+ * Mantine [Loader](https://mantine.dev/core/loader/) component stories.
+ */
+const meta: Meta<typeof Loader> = {
+  title: 'Common/Loader',
+  component: Loader,
+  decorators: [
+    (Story) => (
+      <Center w='400' p='md'>
+        <Story />
+      </Center>
+    ),
+  ],
+  argTypes: {
+    size: {
+      description: 'Select size',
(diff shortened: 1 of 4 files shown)


jamesrkiger (approved):
Should we deprecate this?
https://github.com/kobotoolbox/kpi/blob/pamorim/task-1464-mantine-theme-for-loader/jsapp/js/components/common/loadingSpinner.stories.tsx

Other than that, looks good to me.

pauloamorimbr:
> Should we deprecate this? https://github.com/kobotoolbox/kpi/blob/pamorim/task-1464-mantine-theme-for-loader/jsapp/js/components/common/loadingSpinner.stories.tsx
> 
> Other than that, looks good to me.

I don't think so... I'm just using the default mantine component, which does not use the message parameter. Instead of deprecating the current one with a message we should probably make a wrap so Mantine's one would properly replace it.

PR #5426, opened 2025-01-17
refactor(projectDownloads): migrate files to TypeScript

magicznyleszek:
### 💭 Notes
Migrated whole `jsapp/js/components/projectDownloads` directory to TypeScript. Also defined `LangString` (as alias of string) to give more clarity to the code. Also made some tiny linter fixes here and there.

### 👀 Preview steps
Project Downloads should still work :D I.e.

1. Have a project with submissions
2. Go to Project -> Data -> Downloads
3. Make multiple different exports, each with different settings (try to cover all possible inputs, not necessarily all combinations)
4. Delete some exports
5. Download an export


magicznyleszek:
@duvld has no issues with the code changes. He had problem with making exports work on his local environment. We've asked @pauloamorimbr to test exports and they work for him. They also work for me. Thus I'm, merging this :)

PR #5436, opened 2025-01-21
fix(exports): tasks stuck in processing TASK-1243

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Fixes occasional problem where tasks end up showing as 'Processing' forever.

### 💭 Notes
Tasks were failing when the celery task started too quickly and tried to fetch the task before it had actually hit the database. This PR implements 2 guards against this:
1. Call the tasks from within an on_commit block to ensure they only start after the task object has made it to the database
2. Automatically catch ObjectDoesNotExist exceptions and retry the task a hardcoded number of times. `task.retry` takes care of putting some time in between retries

### 👀 Preview steps
This bug tends to only appear in low traffic when the celery queue is empty. Make sure `CELERY_TASK_ALWAYS_EAGER` is false. 

3. Export the project submissions
4. 🔴 [on main] The job will show as 'Processing' indefinitely. In the worker logs, you'll see `kpi.models.import_export_task.SubmissionExportTask.DoesNotExist: SubmissionExportTask matching query does not exist`
5. 🟢 [on PR] The job will only show as 'Processing' for a few seconds and then will complete.

### DIFF
--- a/kobo/apps/audit_log/views.py
+++ b/kobo/apps/audit_log/views.py
@@ -1,3 +1,4 @@
+from django.db import transaction
 from rest_framework import mixins, status, viewsets
 from rest_framework.decorators import action
 from rest_framework.renderers import BrowsableAPIRenderer, JSONRenderer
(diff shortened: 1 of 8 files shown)


noliveleger (approved):
LGTM. I left some tiny comments which are not blockers to merge this PR. 
It would be great to address them, though ;-)

PR #5437, opened 2025-01-22
feat(helpBubble): add Academy link and update text of existing links

magicznyleszek:
### 📣 Summary
Add new link to KoboToolbox Academy in the `HelpBubble` popup. The link is configurable through `/admin` UI (constance). Two existing links in the same popup had their text updated.

### 💭 Notes
Before --> after:
Screenshot 2025-01-23 at 13 29 01

### 👀 Preview steps

Feature/no-change template:
1. ℹ️ have account and be logged in
2. open HelpBubble (down left area of UI, accessible on almost all routes)
3. 🟢 notice that new element was added
4. 🟢 notice that previous 2 elements have changed text


kobo/settings/base.py
```diff
@@ -217,6 +217,12 @@
         env.str('KOBO_SUPPORT_URL', 'https://support.kobotoolbox.org/'),
         'URL for "KoboToolbox Help Center"',
     ),
+    'ACADEMY_URL': (
```
duvld, line 220: Curious why do we need to check the environment for these links?

kpi/tests/api/test_api_environment.py
```diff
@@ -316,7 +317,7 @@ def test_free_tier_override_uses_organization_owner_join_date(
     def test_social_apps(self):
         # GET mutates state, call it first to test num queries later
         self.client.get(self.url, format='json')
-        queries = FuzzyInt(18, 27)
+        queries = FuzzyInt(18, 28)
```
duvld, line 320: Why this change? Not disagreeing with it but I'm just not sure where this is coming from

duvld (approved):

kpi/tests/api/test_api_environment.py
magicznyleszek: I think this checks how many things are there in environment endpoint, and since we've added one, it needs to be increased:)

Guitlle: Yes, a new constance configuration parameter brings an additional DB query

kobo/settings/base.py
Guitlle: I just followed the same code for the other URLs. But yeah, maybe it's not necessary for this one

PR #5438, opened 2025-01-22
chore(OpenRosa): remove some deprecated endpoints of v1 not called anymore by KPI TASK-1356

Guitlle:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
The OpenRosa V1 API contains many endpoints that are deprecated because the functionality is no longer needed as a REST API but rather as code utilities and classes, instead. The deleted endpoints are:

* bulk_delete
* bulk_validation_status
* validation_status
* destroy
* enketo
* enketo_edit
* enketo_view

### 📖 Description
This PR deletes deprecated code, endpoints and tests, from the OpenRosa V1 API 

### 👀 Preview steps
No preview steps are needed, all unit tests are passing.


noliveleger (commented):
LGTM, but we need to add a REMOVALS.md like we did in KoboCAT repo.
https://github.com/kobotoolbox/kobocat/blob/main/REMOVALS.md

Please have a look at [README](https://github.com/kobotoolbox/kobocat?tab=readme-ov-file#deprecation-notices) in KoboCAT to do something similar in `kobo/apps/openrosa/README` which would points to a REMOVALS.md

noliveleger (approved):

PR #5443, opened 2025-01-23
fix(billing): require org id for service/asset usage TASK-1476

jamesrkiger:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Removes option for API calls to retrieve user-based service/asset usage. 

### 💭 Notes
We previously allowed for the possibility that users would not be organization members. If a user wasn't an org member, we used a separate API endpoint to retrieve their usage data. Now, however, from a frontend perspective users are always organization members and we should always use the organization usage endpoints. 

### 👀 Preview steps
1. As a member of an MMO with at least one admin member, create and deploy a project and make a submission.
2. Access the organization usage page as an admin and view the per-project usage table. 
3. On main, this table will be empty. On this PR branch, it will include the project.
4. Access the organization usage page as the owner and view the per-project usage table.
5. On main, the project will be shown as a draft. On this branch it will show up as deployed.

### DIFF
--- a/jsapp/js/account/usage/usage.api.ts
+++ b/jsapp/js/account/usage/usage.api.ts
@@ -44,46 +44,30 @@ export interface UsageResponse {
   };
 }
 
-const USAGE_URL = '/api/v2/service_usage/';
-const ORGANIZATION_USAGE_URL = '/api/v2/organizations/:organization_id/service_usage/';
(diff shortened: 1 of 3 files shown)


pauloamorimbr (approved):
LGTM!

PR #5444, opened 2025-01-23
Suggested changes for text input pr

pauloamorimbr:
Draft PR to serve as a reference for changes


jsapp/js/theme/kobo/Input.module.css
pauloamorimbr, line 1: I believe we could use this file as a generic for all the text inputs.... or even create a single "Theme" file for all the text inputs to share the same default props and classes.

jsapp/js/theme/kobo/Input.module.css
```diff
@@ -1,3 +1,35 @@
...
+}
+
+.section {
+  color: var(--mantine-color-gray-2);
+}
+
+.label {
+  font-size: var(--mantine-font-size-sm);
+  color: var(--mantine-color-gray-1);
+  margin-bottom: 5px;
```
pauloamorimbr, line 34: Ideally for any layout definition we would use the style prop with the sizes from Mantine, e.g: 
In the TextInput.extend... in the Theme file:
`mb="xs"`

But Maybe the sizes from Mantine won't serve us as we would expect.

jsapp/js/theme/kobo/InputWrapper.ts
pauloamorimbr, line 1: Instead of styling the InputWrapper, we can do that with the TextInput, which exposes the .label directly. 
IDK, it feels like we have more control by doing it individually... or maybe we could save on code and do it on the extended as it was done before. 🤔

jsapp/js/components/common/textInput.stories.tsx
```diff
@@ -1,12 +1,26 @@
...
 
-const inputSizes: Array<TextInputProps['size']> = [
-  'sm',
-  'md',
-  'lg',
-];
+const inputSizes: Array<TextInputProps['size']> = ['sm', 'md', 'lg'];
+
+const Render = ({...args}: TextInputProps) => {
+  const [, updateArgs] = useArgs();
```
pauloamorimbr, line 13: Not really important, but this way we can edit the value in the input and update the args.

PR #5451, opened 2025-01-27
feat(organizations): narrow down `My projects` list  of the organization owner TASK-1187

rajpatel24:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Distinguish between projects created by the organization owner and those created by other members by introducing a flag to manage project visibility in the "My Projects" list.

### 📖 Description
To improve the differentiation of projects in the "My Projects" and "My Org's Projects" lists, the following changes have been implemented:

- Introduced a new flag, `is_excluded_from_projects_list`, in the `Asset` model to determine whether a project should appear in the owner's "My Projects" list.
- Added a new filter backend, `ExcludeOrgAssetFilter`, in `AssetViewSet` to exclude assets with the flag enabled.
- Updated relevant logic during invite creation and acceptance to set or update the flag appropriately.
- Added unit test to verify the functionality, including flag behavior and proper filtering in the viewset.

This change ensures that only projects created by the owner or explicitly shared with them are displayed in the "My Projects" list, addressing the issue of incorrect visibility after organizational changes.


kobo/apps/project_ownership/utils.py
```diff
@@ -27,6 +28,14 @@ def create_invite(
     TransferStatus = apps.get_model('project_ownership', 'TransferStatus')
 
     with transaction.atomic():
+        # Set 'is_excluded_from_projects_list' to 'True' for assets owned by the
+        # sender or invitee joining the organization. These assets should be
+        # excluded from the organization owner's 'My Projects' list
+        real_owner = get_real_owner(sender)
+        for asset in assets:
+            asset.is_excluded_from_projects_list = real_owner != sender
+        Asset.objects.bulk_update(assets, ['is_excluded_from_projects_list'])
```
noliveleger, line 37: This should be moved inside the "transfer_project()" method to respect atomicity (only per asset, no bulk_update). 
Moreover, "create_invite" is also used for regular project transfer ownership. 
This should happen only if the invite is a `OrgMembershipAutoInvite` instance.

kpi/views/v2/asset.py
```diff
@@ -367,6 +372,7 @@ class AssetViewSet(
         KpiObjectPermissionsFilter,
         SearchFilter,
         AssetOrderingFilter,
+        ExcludeOrgAssetFilter
```
noliveleger, line 375: We probably want to exclude the org assets before starting to order them? 
Does it change something to reorder this list?

noliveleger (changes requested):

kpi/views/v2/asset.py
rajpatel24: Applying `ExcludeOrgAssetFilter` earlier could reduce the dataset size and improve performance for subsequent operations like ordering and searching. I have reordered the `filter_backends` to exclude org assets upfront

kobo/apps/project_ownership/models/transfer.py
```diff
@@ -337,15 +337,22 @@ def _update_invite_status(self):
...
+        is_excluded = (
+            isinstance(self.invite, OrgMembershipAutoInvite)
+            and get_real_owner(self.invite.sender) != self.invite.sender
+        )
+
+        if self.asset.is_excluded_from_projects_list != is_excluded:
+            self.asset.is_excluded_from_projects_list = is_excluded
             self.asset.save(
                 update_fields=['is_excluded_from_projects_list'],
                 adjust_content=False
```
noliveleger, line 358: you should pass `create_version=False` too. We don't want to create a version for that.

noliveleger (changes requested):

PR #5457, opened 2025-01-28
feat(organizations): Add isEmpty to avatar component and use it for members table TASK-1420

pauloamorimbr:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Ghost variant was added to display a "ghost" state in members table when user is invited but still not active

### 📖 Description
- Added a `isEmpty` prop for our existent `Avatar` component for it to render the empty-dashed circle.
- At first I intended to add theme and use Mantine's Avatar component, but since our avatar has so many specific details the effort wouldn't be worth it, so I just customized the existent component.
- In the future we can migrate the internal structure of our avatar component to use Mantine's layout components.
- Avatar story was changed to add the isEmpty as a config option

### 👀 Preview steps
One can check the implementation result in Avatar's storybook story
For reference, I'm adding a screenshot of the property in use:
image

### DIFF
--- a/jsapp/js/components/common/avatar.tsx
+++ b/jsapp/js/components/common/avatar.tsx
@@ -32,7 +32,7 @@ interface AvatarProps {
   isUsernameVisible?: boolean;
   fullName?: string;
   email?: string;
-  variant?: 'default' | 'ghost';
+  isEmpty?: boolean;
 }
 
 /**
@@ -45,11 +45,9 @@ export default function Avatar(props: AvatarProps) {
     props.fullName !== undefined ||
     props.email !== undefined;
 
-  const isGhost = props.variant === 'ghost';
-
   return (
     <div className={cx(styles.avatar, styles[`avatar-size-${props.size}`])}>
-      {isGhost ? (
+      {props.isEmpty ? (
         <div className={cx(styles.initials, styles.ghost)}>
           <svg width='100%' viewBox='0 0 24 24'>
             <circle cx='12' cy='12' r='11' />
@@ -70,7 +68,7 @@ export default function Avatar(props: AvatarProps) {
             [styles.hasFullName]: props.fullName !== undefined,
           })}
         >
-          {!isGhost && props.fullName !== undefined && (
+          {props.fullName !== undefined && (
             <span className={styles.fullName}>{props.fullName}</span>
           )}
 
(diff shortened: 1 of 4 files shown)


jsapp/js/components/common/avatar.tsx
```diff
@@ -32,49 +32,58 @@ interface AvatarProps {
...
     <div className={cx(styles.avatar, styles[`avatar-size-${props.size}`])}>
-      <div
-        className={styles.initials}
-        style={{backgroundColor: `${stringToHSL(props.username, 80, 40)}`}}
-      >
-        {props.username.charAt(0)}
-      </div>
+      {isGhost ? (
+        <div className={cx(styles.initials, styles.ghost)}>
+          <svg width='100%' viewBox='0 0 24 24'>
```
pauloamorimbr, line 54: Decided to use SVG we could have control over the dash offset and size

jsapp/js/components/common/avatar.tsx
```diff
@@ -32,49 +32,58 @@ interface AvatarProps {
...
-          className={cx(
-            styles.text,
-            {[styles.hasFullName]: props.fullName !== undefined}
-          )}
+          className={cx(styles.text, {
+            [styles.hasFullName]: props.fullName !== undefined,
+          })}
         >
-          {props.fullName !== undefined &&
+          {!isGhost && props.fullName !== undefined && (
```
jamesrkiger, line 73: Is there any reason the display of  fullName here (or email below) needs to be tied to whether or not the avatar circle is showing up as "ghost"?

jsapp/js/components/common/avatar.tsx
```diff
@@ -32,49 +32,58 @@ interface AvatarProps {
   isUsernameVisible?: boolean;
   fullName?: string;
   email?: string;
+  variant?: 'default' | 'ghost';
```
jamesrkiger, line 35: I'm not saying you have to change this, but I would argue that "outlineOnly" or something like that would be more descriptive than "ghost". You could also consider just making it a bool prop rather than a "variant".

jamesrkiger (commented):
Left some comments for your consideration, but not blocking

pauloamorimbr:
Thanks for the inputs, @jamesrkiger. Did some changes to adjust based on them.

jamesrkiger (changes requested):
Sorry, I've got one last issue for you:

If you use storybook and toggle the isEmpty control on a small avatar component, you will notice the component's alignment shifting (by a very small amount, but visible nonetheless). It looks like the size of the circle changes. It's not noticeable when the size is 'm', but the size changes there, too. It's like a 1 pixel difference, but still probably better to just fix it now rather than later.

jamesrkiger (approved):
lgtm

PR #5459, opened 2025-01-29
fix(organizations): adjust self-ownership check logic TASK-1490

jamesrkiger:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Updates `isSelfOwned()` util to check `owner__username` rather than `owner_label` and drops use of util for determining when to display 'me' vs. owner avatar.

### 💭 Notes
The basic issue here was that asset.owner_label will not match an asset owner's username when they are in an MMO, as the owner label will be the name of their organization. However, the isSelfOwned() util was being used both for determining when to display an asset owner avatar **and** for determining whether or not the user has an effective ownership role over the asset. These are separate questions and so I have removed the use of isSelfOwned() for the former purpose, leaving instances of the second as is. 

Bug template:
1. As the owner of an MMO, create a project and access the project settings view
2. On main, see that there is no "Delete Project and Data" button
3. On this branch, see that the button is displayed
4. Share project with another user without assigning permissions for deleting project
5. Login as second user and see that the button is not displayed

### DIFF
--- a/jsapp/js/assetUtils.ts
+++ b/jsapp/js/assetUtils.ts
@@ -671,7 +671,7 @@ export function isSelfOwned(asset: AssetResponse | ProjectViewAsset) {
   return (
     asset &&
     sessionStore.currentAccount &&
-    asset.owner_label === sessionStore.currentAccount.username
+    asset.owner__username === sessionStore.currentAccount.username
   );
 }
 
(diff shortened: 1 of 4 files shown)


magicznyleszek (approved):
Code looks good, bug is fixed

PR #5464, opened 2025-01-30
fix(billing): remove quantity dropdown from frontend for nlp add-ons TASK-1482

RuthShryock:
### 🗒️ Checklist

1. [X] run linter locally
2. [X] update all related docs (API, README, inline, etc.), if any
3. [X] draft PR with a title `<type>(<scope>)<!>: <title> TASK-1234`
4. [X] tag PR: at least `frontend` or `backend` unless it's global
5. [X] fill in the template below and delete template comments
6. [X] review thyself: read the diff and repro the preview as written
7. [X] open PR & confirm that CI passes
8. [X] request reviewers, if needed

### 📣 Summary
Removes quantity dropdown from frontend for nlp add-ons.

### 👀 Preview steps

Bug template:
2. Check out the Account Settings < Add-ons
4. 🔴 [on main] notice that the NLP Package has a quantity dropdown 
5. 🟢 [on PR] notice that the NLP Package no longer has a quantity dropdown

Thoroughly test add-ons and subscription plans to make sure that everything is still working correctly.


jsapp/js/account/addOns/oneTimeAddOnRow.component.tsx
```diff
@@ -186,15 +186,7 @@ export const OneTimeAddOnRow = ({
...
-          ) : (
-            <KoboSelect3
-              size={'fit'}
-              name={t('quantity')}
-              options={quantityOptions}
-              onChange={onChangeQuantity}
-              value={quantity}
-            />
-          )}
+          ) : null}
```
Akuukis, line 189: style nitpick: simpler like this?
```tsx
          {displayName === 'File Storage' && (
            <KoboSelect3
            <KoboSelect3
              size={'fit'}
              size={'fit'}
              name={t('prices')}
              name={t('prices')}
              options={priceOptions}
              options={priceOptions}
              onChange={onChangePrice}
              onChange={onChangePrice}
              value={selectedPrice.id}
              value={selectedPrice.id}
            />
            />
          )}

**Review by Akuukis — CHANGES_REQUESTED** [review]
code LGTM :heavy_check_mark: previewed successfully locally as written :heavy_check_mark: 

The only problem is that this is a bundle of 3 different fixes, please break up and PR separately.

**Review by jamesrkiger — CHANGES_REQUESTED** [review]
This seems to be fine for the UI change of removing the dropdown, but there is a lot of logic in the code for handling addon quantities (as opposed to subscription quantities, which are separate) that needs to be removed. Would you mind going through to remove any code that is no longer needed as a result of this change?

**Review by Akuukis — APPROVED** [review]
LGTM from my POV but James raised a good point :point_up:

**Review by jamesrkiger — CHANGES_REQUESTED** [review]
Sorry, but I guess I should have specified that I was referring to the frontend code rather than the backend code. That's OK. We probably do want to consider removing the related backend code, but that will require some more careful testing. Can you split the backend work off into a separate PR and assign it to Guillermo and I for review? 

For the frontend code, can you remove the interface changes for the time being? I would like us to be able to review/merge the frontend changes without having to rush the backend review. 

Please let me know if that makes sense.

File: jsapp/js/account/addOns/addOnList.module.scss [hunk 2]
```diff
@@ -149,6 +149,10 @@
 
   .oneTime {
     justify-content: center;
+
+    :last-child {
```
jamesrkiger, line 153: Maybe some weird git issue. I think you removed this in a previous PR

PR #5465, opened 2025-01-31
fix(submissions): deletion issue after project transfer TASK-1504

noliveleger:
### 📣 Summary
Fixed a bug where submissions were not being fully deleted after a project transfer, ensuring proper cleanup in both PostgreSQL and MongoDB. This issue was introduced during the development of the current release and never appeared in a published version.

### 📖 Description
A bug prevented submissions from being deleted after a project was transferred. While they were successfully removed from PostgreSQL, they remained in MongoDB, leading to data inconsistencies and preventing proper cleanup. This fix ensures that submissions are deleted from both databases, restoring expected behavior.


rajpatel24 (approved):
LGTM

PR #5435, opened 2025-01-21
feat(submissions)!: add submission unique identifier (`meta/rootUuid`) in data API response TASK-1458

noliveleger:
### 📣 Summary
Added the unique identifier rootUuid to the data API response for submissions

### 📖 Description
The `meta/rootUuid` field, which serves as a unique identifier per project, has been added to the data API response for submissions. The `meta/rootUuid` remains unchanged even when a submission is edited, ensuring it consistently represents the original submission. This enhancement allows API consumers to track and reference submissions more effectively using their unique identifiers.

## Notes
`meta/rootUuid` in edit submissions - which was saved without the namespace - is now saved with it to comply to OpenRosa specs and to behave like `meta/instanceID` and `meta/deprecatedID`

PR #5476, opened 2025-02-01
feat(submissions): frontend support for 'uuid:' prefix in 'meta/rootUuid'

p2edwards:
### 💭 Notes

Frontend compatibility for #5435. 

**🐍** With #5435, a `meta/rootUuid` from the submissions data API now includes the `'uuid:'` prefix.

| field | example | changes |
|-|-:|:-|
|**meta/rootUuid**|`uuid:4e30afbf-d6cf-4fd7-ac59-77c30ece4b6b`| <ol><li>✴️ meta/rootUuid **now always includes a prefix**. <ul><li><b>default prefix is `uuid:`.</b><li> can be a custom prefix like `org.kobotoolbox:`.<li>in this way it's like meta/instanceId. (and unlike _uuid)</ul><li>✴️ **meta/rootUuid is now always present,**<ul><li>...regardless of whether or not this submission has been edited.<li>Previously the frontend could not assume this.</ul></li></ol>|
|meta/instanceId | `uuid:0d299cf7-2b02-460e-aba5-3b295170b69e` | (unchanged) |
|_uuid | `0d299cf7-2b02-460e-aba5-3b295170b69e` | (unchanged) |

🔍 Since the frontend had expected the `uuid:` prefix to be stripped in both `_uuid` and `rootUuid`...

 ||| Need  |   |
|-|:--- |:---------------|:---------------------------------------- |
|1. |🔍 Whenever we compare a `_uuid` and a `meta/rootUuid` for **equivalence**... | Treat 'uuid:example' and 'example' as equivalent | ❌ <s>uuid === sub['meta/rootUuid']</s><br> ✴️ **matchUuid**(uuid, sub['meta/rootUuid'])  |
|2. | 🔍 Whenever we take a `meta/rootUuid` **in the place of** a `_uuid`... | Remove the default prefix (for compatibility with other stores/id's)| ❌ <s>id = sub['meta/rootUuid'] </s><br>✴️ id = **removeDefaultUuidPrefix**(sub['meta/rootUuid']) |
|3.| 🔍 Whenever we **query the API** based on a `meta/rootUuid`... | Include the expected prefix|❌ <s>{ 'meta/rootUuid': id }</s> <br>✴️ { 'meta/rootUuid': **addDefaultUuidPrefix**(id) } |

#### 🕵️ Frontend reviewers!

I'm asking for help to make sure these changes are comprehensive and valid.

🔍 **Comprehensiveness**. I've conducted a search of the frontend for usages of `rootUuid`, and of `uuid` generally. **But I may have missed something,** or there may be some cases I didn't think of. Please double-check.

🐞 **Testing**:  In my testing so far, didn't find any editing issues in this PR or its merge target branch (except unrelated things, already present in `main`).

I think that's because the frontend code relating to meta/rootUuid was pretty defensive already, with fallbacks on many code paths when it couldn't find a matching `meta/rootUuid`. Still,

- Check that Enketo links and the Single Processing view continue to work as expected, with this PR. 
- Look for other problems that I may have missed.

🥏 **Later follow-up (optimizations)**. With `meta/rootUuid` being stored for all submissions and always present in submission API responses, we can apply these optimizations, which **I opted not to include yet**, to keep this PR small. In the future,

  - We can simplify/remove some code that uses _uuid as a fallback if it can't find meta/rootUuid
  - We could probably simplify some queries around fetching related submissions.
 
Let's remember to make a follow-up task to look into this, as my hunch is some views or stores could reduce their queries based on this, and that would be nice both for frontend and backend.

### 👀 Preview steps

1. 🟢 [on PR] Click around, editing submissions and viewing things in data.
    - Hunt for any problems or glitches introduced around editing submissions that weren't already present in [main].
2.  🔴 [on the PR's merge target, task-1458-inject-root-uuid-in-api-response]
    - See if any of the code paths in this PR actually fix something important.


jsapp/js/utils.ts
```diff
@@ -532,3 +532,59 @@ export function getExponentialDelayTime(
...
+/**
+ * Add a default 'uuid:' prefix to the provided identifier,
+ * if there's no prefix already.
+ *
+ *  Examples:
+ *    'kobotoolbox.org:123456789' -> unchanged
+ *    'uuid:123456789' -> unchanged
+ *    '123456789' -> 'uuid:123456789'
+ *
+ * 🐍 Equivalent to add_uuid_prefix on the backend.
```
p2edwards, line 547: xref [🐍 kobo/apps/openrosa/apps/logger/xform_instance_parser.py](https://github.com/kobotoolbox/kpi/pull/5435/files#diff-b1b52e81f116826f3b9bba9460299d530ded922d0b48d96bc6647394e19281b1)

jsapp/js/utils.ts
```diff
@@ -532,3 +532,59 @@ export function getExponentialDelayTime(
...
+/**
+ * Remove the default 'uuid:' prefix from a provided identifier,
+ * while preserving any custom prefixes (e.g. 'kobotoolbox.org:123456789'),
+ * which are allowed by OpenRosa to reduce the chance of ID collisions.
+ *
+ * 🐍 Equivalent to remove_uuid_prefix on the backend.
+ */
+export function removeDefaultUuidPrefix(uuid: string) {
+  return uuid.replace(/^uuid:/, '');
+}
```
p2edwards, line 562: xref [🐍 kobo/apps/openrosa/apps/logger/xform_instance_parser.py](https://github.com/kobotoolbox/kpi/blob/efd05b375cfa9565261d2f238d3d6ebd0af47b50/kobo/apps/openrosa/apps/logger/xform_instance_parser.py#L137-L145)

```py
def remove_uuid_prefix(uuid_: str) -> str:
    """
    Strips the 'uuid:' prefix from the provided identifier if it exists.
    This preserves any custom ID schemes (e.g., 'kobotoolbox.org:123456789')
    while ensuring only the 'uuid:' prefix is removed. This approach
    adheres to the OpenRosa spec, allowing custom prefixes to be stored
    intact in the database to prevent potential ID collisions.
    """
    return re.sub(r'^uuid:', '', uuid_)
```

noliveleger (approved):
LGTM

PR #5478, opened 2025-02-03
fix(SSO): only allow SSO email when registering TASK-1493

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Do not allow SSO users to register with a different email then the one provided by the server.

### 📖 Description
Make the `email` field readonly when creating an account with SSO.

### 💭 Notes
This was originally a setting, but we've decided we want the field to always be read-only. In addition to the UI change, this PR also adds a validator on the email field to make sure that no one can register with a different email via a clever POST request. The error message will require translation.
This should also fix TASK-1493.

### 👀 Preview steps

Bug template:
1. ℹ️ Enable Kobo Google Apps SSO
2. On the login page, click `Create an account` -> `Register with SSO` -> `Log In` 
3. Sign in to your Google account
4. 🔴 [on main] Change the email field in the registration form and click 'Register and Save'
5. :red_circle:  [on main] Go to Account Settings -> Security. Under the email address you'll see "Check your email <new_email>. A verification link has been sent to confirm your ownership. Once confirmed, this address will replace <correct_email>"
6. 🟢 [on PR] The email field is read-only

### DIFF
--- a/kobo/apps/accounts/forms.py
+++ b/kobo/apps/accounts/forms.py
@@ -11,14 +11,13 @@ from allauth.account.utils import (
 from allauth.socialaccount.forms import SignupForm as BaseSocialSignupForm
 from django import forms
 from django.contrib.auth import get_user_model
-from django.conf import settings
 from django.utils.safestring import mark_safe
-from django.utils.translation import gettext, gettext_lazy as t
+from django.utils.translation import gettext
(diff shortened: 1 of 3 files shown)


RuthShryock (approved):
LGTM!

PR #5477, opened 2025-02-03
feat(organizations): restrict duplicate invitations and allow role updates by organization owner or admin TASK-1500

rajpatel24:
### 📣 Summary
Restricts duplicate invitations and allows invitee role updates by the organization owner or admin.

### 📖 Description
This update prevents owners from creating new invitations for users who already have an active invitation. Additionally, it enables organization owners and admins to update the invitee's role, ensuring better control over team management.

> PATCH https://[kpi]/api/v2/organizations/org_12345/invites/7a4f9a3b-9112-43cc-a6ae-bb4a6583b4b2/

- Update an organization invite to accept, decline, cancel, expire, or resend.
- Update the role of the invitee to `admin` or `member`. Only the owner or admin can update the role.

Payload (To change the status):
```
{
    "status": "accepted"
}
```

Payload (To update Role - Only owner or admin can update role):
```
{
    "role": "admin"
}
```

Response:
```
{
    "url": "http://kf.kobo.local/api/v2/organizations/org_12345/invites/7a4f9a3b-9112-43cc-a6ae-bb4a6583b4b2/",
    "invited_by": "http://kf.kobo.local/api/v2/users/gtl_raj/",
    "status": "accepted",
    "invitee_role": "admin",
    "created": "2025-01-06T13:01:40Z",
    "modified": "2025-01-06T13:01:40Z",
    "invitee": "demo14"
}
```

### 👀 Preview steps
1. As an organization owner, create an invitation for userA.
2. Try inviting userA again, it should not allow duplicate invitations.
3. Update the invitee’s role to admin (default is member), only the owner or admin should be able to do this.
4. Accept the invitation as userA and verify the assigned role.

PR #5480, opened 2025-02-03
feat(organizations): adjust members table for invitees TASK-1499

jamesrkiger:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Adjusts members table columns to handle data for both organization members and invited organization members.

### 💭 Notes
I have mostly left the 'actions' column untouched, aside from changes necessary to work with new types. This column will be changed in a separate task.

This PR depends on #5477 and should be merged after it. It also anticipates work that is yet to be done on the members list API (see [here](https://www.notion.so/kobotoolbox/Do-not-reveal-user-information-until-they-accept-invites-1907e515f65480739575da841e8bd6ac)), but the change there is going to be that member details are null for invited members and all data should be retrieved from the invite field. So this should be fine to merge before that later work is done.

### 👀 Preview steps
Feature/no-change template:
1. Use django shell or direct API to create an invitation for a user to join an MMO
2. As an owner or admin of the MMO, navigate to the members table
3. Columns for organization members should not change between this branch and `main`
4. The invitee's row should match what we have in the figma designs for this project

### DIFF
--- a/jsapp/js/account/organization/MemberRoleSelector.tsx
+++ b/jsapp/js/account/organization/MemberRoleSelector.tsx
@@ -1,6 +1,6 @@
 import {Select} from 'jsapp/js/components/common/Select';
 import {usePatchOrganizationMember} from './membersQuery';
-import {MemberInviteStatus, usePatchMemberInvite} from './membersInviteQuery';
+import {usePatchMemberInvite} from './membersInviteQuery';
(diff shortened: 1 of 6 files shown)


pauloamorimbr (changes requested):
The only real change is the missing `await`. Other than that LGTM! 💯 

Took a good look at the Python part and seems good, but I'm not that experienced on that!

PR #5487, opened 2025-02-04
build(frontend): use pinned, active and current node version

Akuukis:

Support only pinned (v20.17), active (v22 until Nov 2025) and current (v24 after May 2025) node versions.


jnm (approved):
Approved. Yes, fine to drop support for Node 16.

PR #5492, opened 2025-02-06
fix(organization)!: allow external users to GET their invitation before joining the organization TASK-1500

noliveleger:
### 📣 Summary
Fixed an issue where external users could not retrieve their invitation details before officially becoming part of the organization. This change ensures they can access their invitation and review relevant information before accepting.

### 📖 Description
Previously, external users were unable to fetch their invitation details via the API before joining an organization. This fix updates the permission logic to allow invited users to perform a GET request on their invitation while maintaining security constraints.

### 💭 Notes
The permission class has been refactored. The validation of the status has been moved in the serializer instead the permission class. The base queryset is more permissive but restricted at the object level (with `get_object()`). 

Breaking change: The API returns a 400 instead of 404 when user tries actions other than `GET` on invitation objects there are allowed to see.


rajpatel24 (approved):
LGTM

PR #5497, opened 2025-02-07
build(frontend): windows-compatible path.join

Akuukis:
### 🗒️ Checklist

1. [x] run linter locally

### 💭 Notes

Found this along the way, let's use `path.join` in a windows compatible way to have backslash in all path segments instead of only the first one.

### 👀 Preview steps

Build frontend from windows command shell (not Git Bash, it uses slashes the unix way)


Akuukis:
looks like webpack or path handles windows slashes internally and this makes no difference.

PR #5501, opened 2025-02-07
docs: fix some formatting

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 💭 Notes
Abstracts the documentation for the PH log endpoints into a parametrized method since they were 90% the same and it was annoying to always have to update the docs in 2 places every time we add a new action or need a format fix.

### DIFF
--- a/kobo/apps/audit_log/views.py
+++ b/kobo/apps/audit_log/views.py
@@ -335,77 +335,7 @@ class AccessLogViewSet(AuditLogViewSet):
     filter_backends = (AccessLogPermissionsFilter,)
     serializer_class = AccessLogSerializer
 
-
-class AllProjectHistoryLogViewSet(AuditLogViewSet):
-    """
-    Project history logs
-
-    Lists all project history logs for all projects. Only available to superusers.
-
-    <pre class="prettyprint">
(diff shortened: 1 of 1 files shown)


kobo/apps/audit_log/views.py
```diff
@@ -591,9 +588,19 @@ class AllProjectHistoryLogViewSet(AuditLogViewSet):
...
+    >      curl -X GET https://[kpi-url]{example_path}?offset=100&limit=50
+
     """
 
+
+class AllProjectHistoryLogViewSet(AuditLogViewSet):
+    __doc__ = generate_ph_view_set_logstring(
+        'List all project history logs for all projects. Only available to superusers.',
+        '/api/v2/project-history/logs',
+        '/api/v2/project-history/logs',
```
rajpatel24, line 600: @rgraber Shouldn't this be `/api/v2/project-history-logs` instead of `/api/v2/project-history/logs`, or am I missing something here?

rgraber: You are correct! Good catch. Fixing.

rajpatel24 (approved):
LGTM

PR #5503, opened 2025-02-08
fix(organization): return HTTP 429 when retrying to resend an invitation too quickly TASK-1513

noliveleger:
### 📣 Summary
Changed the response from HTTP 400 to HTTP 429 (Too Many Requests) when resending an invitation too quickly.

### 📖 Description
Previously, when an admin attempted to resend an invitation too quickly, the system enforced a rate limit but returned HTTP 400 (Bad Request) without guidance on when to retry. This update improves accuracy by returning HTTP 429 (Too Many Requests), which better represents the nature of the restriction.

Additionally, the response now includes the `Retry-After` header, specifying the number of seconds the client must wait before resending the invitation.


rajpatel24 (approved):
LGTM

PR #5505, opened 2025-02-10
feat(organizations): implement 'retry later' message for resend TASK-1532

pauloamorimbr:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
- Headers are now returned by API error handler FailResponse object
- When trying to resend an organization invitation too often now the user gets a proper message

### 📖 Description
We're using the response status 429 and the `retry-later` header information to properly format an error message to the user.

### 👀 Preview steps
1. ℹ️ be part of an organization
2. be admin or owner
3. invite a new member to the organization
4. Try to resend the invite via invitee actions 
5. There should be a proper message if the resend is sent too soon

### DIFF
--- a/jsapp/js/account/organization/InviteeActionsDropdown.tsx
+++ b/jsapp/js/account/organization/InviteeActionsDropdown.tsx
@@ -39,7 +39,7 @@ export default function InviteeActionsDropdown({
         )
         return
       }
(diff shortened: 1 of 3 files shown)


jamesrkiger (changes requested):
When a non-429 error comes in we end up getting two notifications:
image

PR #5516, opened 2025-02-12
fix(frontend root): redirect to log in

pauloamorimbr:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Not logged users are now properly redirected to login screen when accessing parts of the app that requires auth

### 📖 Description
- `RequireOrg` component was locking app into loading state and not moving forward to redirect
- `RequireOrg` was moved into `requireAuth` component so Org is only fetched when user is logged in

### 👀 Preview steps
1. ℹ️ log out of the app
2. try to access `/#/forms`
3. 🔴 [on main] user gets locked in an infinite loading
4. 🟢 [on PR] user is properly redirected to login screen


jamesrkiger (approved):
lgtm

PR #5522, opened 2025-02-13
feat(OrgInviteModal): better handling of custom errors TASK-1093

magicznyleszek:
### 📣 Summary
Use error message provided by the endpoint if possible. Otherwise use generic error. Also avoid displaying multiple error notifications in the modal.

### 💭 Notes
Changed `usePatchMemberInvite` to make the error notification optional. Also fixed "x" button getting auto focused when modal was opened.

### 👀 Preview steps

Reproduction:
1. ℹ️ create an invite for "joe" (not a member of organization)
2. log in as other user (member of organizatio)
3. open invite link
4. notice that the invitation modal opens
5. click "Accept" or "Decline"
6. 🟢 notice that the content of the modal changes into error message
7. 🟢 notice that error notification no longer appear (they appear in `main`)

### DIFF
--- a/jsapp/js/account/organization/invites/OrgInviteModal.tsx
+++ b/jsapp/js/account/organization/invites/OrgInviteModal.tsx
@@ -14,6 +14,7 @@ import { getSimpleMMOLabel } from 'js/account/organization/organization.utils'
 import envStore from 'jsapp/js/envStore'
 import subscriptionStore from 'jsapp/js/account/subscriptionStore'
 import { notify } from 'jsapp/js/utils'
+import { useSession } from 'jsapp/js/stores/useSession'
 // Constants and types
 import { endpoints } from 'jsapp/js/api.endpoints'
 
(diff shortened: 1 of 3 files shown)


pauloamorimbr (approved):
LGTM! 💯 

I just noticed 2 things that I believe are not directly related to this PR:
1. I was able to open a link to an invitation when logged to a different user from the invite. I believe it shouldn't be possible.

2. Because of the automatic error handling from our API, in the case of accepting a removed invitation we see a lot of notifications popping up displaying errors.

Are any of those things we should worry about?

magicznyleszek:
@pauloamorimbr for the multiple notifications, I think it comes from the fact that query is trying multiple times to get invitation (response is 404 each time). Maybe we should `retry: false`? Or add `displayErrorNotification` option similar that I've added to  `usePatchMemberInvite`? Because the modal will display some error eventually, notifications are not needed as much…

magicznyleszek:
@pauloamorimbr re "I was able to open a link to an invitation when logged to a different user from the invite. I believe it shouldn't be possible." - what do you think should happen? If you try to accept or decline, it would display an error, right? So for you to be able to open such link, you would need to have an access to invited user email, or be part of the organization and an admin/owner, and get the invite id from the API response, and then build url, and then open it. Right? 😅

pauloamorimbr:
Taking a second look now, I think we could:
- use your `displayErrorNotification` solution to avoid double notification for the same error anyways
- Set the `enabled` in the query to check for the inviteUrl
- Use the inviteUrl as a state we can change when there's an error, so we can tell the query to "not run" if we detect some issue. Or even pre-process some errors, like if it's the correct logged user, and not even set the inviteUrl, so we avoid having invalid calls.

regarding the invitation opening on the wrong logged user, what I thought - but I don't know if it makes sense - is that some user may be logged in a personal account, and is being invited in a corp account, when it clicks the email it can get messy on where it's trying to accept. But that would be for another PR, since it can be a complex solution.....maybe.

PR #5525, opened 2025-02-14
fix(organization)!: revoke asset permissions before removing a member from the organization TASK-1560

rajpatel24:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Revoke organization asset permissions before removing a member from the organization to ensure proper access control.

### 📖 Description
When a member is deleted from an organization via the /members DELETE API, their associated permissions on organizational assets are now revoked before the deletion takes place. This prevents orphaned permissions and ensures that removed users no longer have access to organization resources.

### 👀 Preview steps
1. Ensure you have a user account and an asset.
2. As an organization owner, send an invitation to that user to join the organization.
3. Once the user joins, confirm that the asset is now owned by the organization.
4. As an organization owner, remove the user from the organization via the `/members` DELETE API.
5. Verify that the removed user no longer has access to the asset.

### DIFF
--- a/kobo/apps/organizations/tests/test_organization_members_api.py
+++ b/kobo/apps/organizations/tests/test_organization_members_api.py
@@ -6,6 +6,8 @@ from kobo.apps.kobo_auth.shortcuts import User
 from kobo.apps.organizations.tests.test_organizations_api import (
     BaseOrganizationAssetApiTestCase
 )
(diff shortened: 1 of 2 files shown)


RuthShryock (approved):
LGTM, tested and works well!

PR #5526, opened 2025-02-14
chore(githubActions): update ubuntu version in github actions 

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 💭 Notes
Developer-only changes. Also specifies which version of python:3.10 we use to prevent surprise updates to a new debian.

### DIFF
--- a/.github/workflows/npm-test.yml
+++ b/.github/workflows/npm-test.yml
@@ -3,7 +3,7 @@ name: npm-test
 on: workflow_call
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
 
     strategy:
 
--- a/.github/workflows/npm-test.yml
+++ b/.github/workflows/npm-test.yml
@@ -3,7 +3,7 @@ name: npm-test
 on: workflow_call
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
 
     strategy:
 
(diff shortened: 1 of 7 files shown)


.github/workflows/npm-test.yml
```diff
@@ -3,7 +3,7 @@ name: npm-test
 on: workflow_call
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
jnm, line 6: Could we pin this (and the other) to match the `Dockerfile`? `python:3.10` is based on Debian 12. I don't see a Debian runner for GitHub Actions, but Debian 12 corresponds to Ubuntu 22.04 LTS per https://askubuntu.com/a/445496.

Ideally the `Dockerfile` would carry an admonition to update GitHub Actions runners whenever the base image is changed.

jnm (changes requested):

.github/workflows/npm-test.yml
rgraber: I think python:3.10 is [based on Debian 11](https://hub.docker.com/layers/library/python/3.10/images/sha256-a4fbd0cbe6e333ed05a2af41d0771aecc90ad804fe8a58023c52265ad2bfea36)

rgraber: Should we be using 3.10-bookworm?

jnm (approved):
Thanks! Could you update the PR title before you merge?

PR #5527, opened 2025-02-17
fix(translations): use trimmed in all blocktrans blocks TASK-1553

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📖 Description
Developer-only changes

### 💭 Notes
This was done with a simple find-and-replace. To test it,  I compiled the messages and ran
`find . -name django.po | xargs awk '{ if (lastline == "msgid \"\"" && $0 ~ "\"\\\\n\"") {print FILENAME, $0} {lastline=$0}}' | wc -l` and checked to make sure the output was empty.

The `awk` command looks for sections of a .po file that look like
```
msgid ""
"\n"
```
, which is what it looks like when a translation string begins with a newline char. The full command lists all po files, checks for that sequence of lines using `awk` for each one, and counts the total. On main, the count was 257. Using this branch, the count is 2. It should be 0, but for some reason `makemessages` had trouble with `zh-Hant` and `zh-Hans` so it didn't actually update those files. That doesn't matter for this PR.

### :eyes: Preview steps

1. In the kpi container run `python ./manage makemessages -a`
2. Locally, cd to the `locale` directory inside kpi
3.  run `find . -name django.po | xargs awk '{ if (lastline == "msgid \"\"" && $0 ~ "\"\\\\n\"") {print FILENAME, $0} {lastline=$0}}' | wc -l`
4. :red_circle: [on main] The result will be > 200
5. :green_circle: [on PR] The result will be 0 (or 2, if django has trouble compiling zh-Hant and zh-Hans. Not sure if that was a local issue).

### DIFF
--- a/kobo/apps/accounts/templates/account/email/email_confirmation_message.txt
+++ b/kobo/apps/accounts/templates/account/email/email_confirmation_message.txt
@@ -1,11 +1,11 @@
 {% load i18n %}
 {% load url_creation %}
-{% blocktrans %}To confirm the new email address associated with your KoboToolbox account, please follow the link below.{% endblocktrans %}
+{% blocktrans trimmed %}To confirm the new email address associated with your KoboToolbox account, please follow the link below.{% endblocktrans %}
 
 {{ activate_url }}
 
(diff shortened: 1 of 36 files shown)


noliveleger:
@rgraber I guess in your preview steps you meant: 

`python ./manage.py makemessages -a`

PR #5528, opened 2025-02-17
fix(frontend): improve paddings on Project Landing Page TASK-1565

magicznyleszek:
### 💭 Notes

As we want to go forward with Mantine, I've used `Stack` to fix the missing paddings.

The route already was using a "system" for paddings in the shape of `form-view__cell--paddings` class name. I've modified it to use the same size of paddings as the `Stack` (I went with `lg`, which is 20px - same as previous paddings).

I also simplified some styles by removing one shots `form-view__cell--small-paddings` and `form-view__cell--paddings-small` :heart:

### 👀 Preview steps

Reproduction:
2. deploy project
3. edit form
4. go to Project Landing Page (`/#/forms/<uid>/landing`)
5. 🟢 notice that there is an alert "If you want to make these changes(…)"
6. 🟢 notice alert has paddings around it (not on bottom), which is not present on `main`
7. 🟢 notice the paddings of all other sections here match the paddings around the alert

### DIFF
--- a/jsapp/js/components/formLanding/formLanding.js
+++ b/jsapp/js/components/formLanding/formLanding.js
@@ -28,6 +28,7 @@ import AnonymousSubmission from 'js/components/anonymousSubmission.component'
 import NewFeatureDialog from 'js/components/newFeatureDialog.component'
 import pageState from 'js/pageState.store'
 import Button from 'js/components/common/button'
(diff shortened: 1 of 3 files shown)


RuthShryock (approved):
LGTM :)

PR #5529, opened 2025-02-17
feat(activityLogs): handle modify-qa-data action in UI

magicznyleszek:
### 💭 Notes
Simply adds action to the list of definitions. The code does the rest :bow:

Also improve how `AUDIT_ACTION_TYPES` order property is being done which makes it easier to order the actions (rather than jumping around and modifying all the numbers).

### 👀 Preview steps

Testing:
1. ℹ️ have account and a project with audio questions and submission(s)
2. go to Project -> Data -> Table and open NLP view for one of the audio question responses
3. create a transcript
4. modify transcript text and SAVE
5. go to Project -> Settings -> Activity
6. 🟢 notice that there is an action that says "<username> edited qualitative analysis data"
7. 🟢 notice that "Filter by" dropdown has an option "edit qualitative analysis data"


jsapp/js/components/activity/activity.constants.ts
```diff
@@ -43,205 +44,249 @@ type AuditActionTypes = {
...
+  AuditActions['delete-service'],
+  AuditActions['add-submission'],
+  AuditActions['modify-submission'],
+  AuditActions['delete-submission'],
+]
+
 export const AUDIT_ACTION_TYPES: AuditActionTypes = {
   'add-media': {
-    order: 10,
+    order: AUDIT_ACTIONS_ORDER.indexOf(AuditActions['add-media']),
```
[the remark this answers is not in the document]
Akuukis: I prefer to have only 1 long list, that's less repetition and less places to forget to update when the time comes. For example:
- don't add AUDIT_ACTIONS_ORDER
- rename `export const AUDIT_ACTION_TYPES` to `const _AUDIT_ACTION_TYPES`, make it an array with key as `id` prop, and sort it,
- add
    ```ts
    export const AUDIT_ACTION_TYPES = _AUDIT_ACTION_TYPES.reduce((dict, action, i) => {...dict, [action.id]: {...action, order: i}, {}) 
    ```
- add `as const` whenever possible and otherwise try to make it as a constant and strict.

jsapp/js/components/activity/activity.constants.ts
```diff
@@ -38,216 +39,169 @@ export enum AuditActions {
...
+    label: t('delete a submission'),
+    message: t('##username## deleted a submission'),
   },
+]
+
+type AuditActionTypes = {
+  [P in AuditActions]: AuditActionDefinition
 }
 
+export const AUDIT_ACTION_TYPES: AuditActionTypes = _AUDIT_ACTION_TYPES.reduce((acc, action, index) => {
```
magicznyleszek, line 198: @Akuukis @pauloamorimbr what do you think about this? :)

Akuukis: Looks good! Although not nice due linter :sweat: 

Why the `  if (action.name) {`?

magicznyleszek: > Looks good! Although not nice due linter 😓
> 
> Why the ` if (action.name) {`?

Yeah linter made this a bit 🤢. I thin the `if` is a leftover from a WIP iteration of this. It is not needed now, good catch! Removed it

Akuukis: @magicznyleszek also see https://biomejs.dev/formatter/#ignore-code, I think this place justifies ignoring formatter.

pauloamorimbr:
I just notice that there's no TASK-### in the description. Is there a task for this? 🤔

magicznyleszek:
> I just noticed that there's no TASK-### in the description. Is there a task for this? 🤔

There is none 🤫 But there is Tino's comment on Zulip. Seems more important :P

Akuukis (approved):
code LGTM, not tested.

PR #5532, opened 2025-02-18
fix(invitations): hide invitation accepted banner for users who are no longer members of organization TASK-1585

magicznyleszek:
### 💭 Notes
I've renamed `showBanner` to `isBannerDismissed` for clarity and improved the `if`s list. The actual fix is verifying `orgQuery.data.is_mmo`.

### 👀 Preview steps

Bug template:
1. ℹ️ have two accounts: "joe" and "bo"; "joe" is org owner, "bo" is not
2. send invite from "joe" to "bo" for joining organization
3. accept invite as "bo" (mind https://www.notion.so/kobotoolbox/Display-invite-accepted-banner-immediately-19d7e515f6548087afedf3644410c72d?pvs=4)
4. 🟢 as "bo" notice that the invite accepted banner shows up (on top of `/#/projects/home` route) (make sure not to dismiss the banner with "x" button)
5. as "joe" then remove "bo" from organization
6. 🔴 [on main] as "bo" notice that the banner is still visible
7. 🟢 [on PR] as "bo" notice that the banner is not visible


jamesrkiger (approved):
lgtm

PR #5540, opened 2025-02-19
fix(organization)!: restrict invite details to the user's current organization TASK-1590

rajpatel24:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Updated the members API endpoint to ensure invite details are only available for the organization a user currently belongs to.

### 📖 Description
Previously, the members API endpoint could display invite details from a user's past organization, even after they joined a new one. This update ensures that:

- Invite details are only shown if the user is still part of the organization.
- If a user is removed from an organization, their invite details for that organization will no longer be visible.
- If the user joins a different organization, invite details from the previous organization will not be carried over.

This change improves data consistency and ensures that invite details are relevant only within the organization where the user is currently a member.

### DIFF
--- a/kobo/apps/organizations/tests/test_organization_members_api.py
+++ b/kobo/apps/organizations/tests/test_organization_members_api.py
@@ -3,6 +3,10 @@ from django.urls import reverse
 from rest_framework import status
 
 from kobo.apps.kobo_auth.shortcuts import User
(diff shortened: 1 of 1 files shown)


noliveleger (approved):
LGTM

PR #5541, opened 2025-02-19
fix(organization): assign invitations to members correctly TASK-1600

noliveleger:
### 📣 Summary
Fixed an issue where invitations were not properly assigned to the correct organization members and ensured that invitations reflect the intended recipient and organization.

### 📖 Description
An issue caused invitations to be incorrectly assigned, leading to inconsistencies in organization membership management. This fix ensures that invitations are correctly linked to the intended recipients, preventing access errors or misdirected invitations.

### 💭 Notes
`invite` property queryset has been optimized to avoid one query per member from the list endpoint.

### 👀 Preview steps

Bug template:
1. Create two MMO
2. From the owner account, invite the owner of the second
4. 🔴 [on main] notice that in the account of the 2nd owner, the organization member table shows the owner as invited
5. 🟢 [on PR] notice that in the account of the 2nd owner, the organization member table shows the owner as the owner and active

### DIFF
--- a/kobo/apps/organizations/serializers.py
+++ b/kobo/apps/organizations/serializers.py
@@ -79,6 +79,7 @@ class OrganizationUserSerializer(serializers.ModelSerializer):
         ]
 
     def get_url(self, obj):
+
         request = self.context.get('request')
         return reverse(
             'organization-members-detail',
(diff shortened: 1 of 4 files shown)


rajpatel24 (approved):
LGTM

PR #5539, opened 2025-02-19
feat(permissions): incomplete permission removal with “Removed shared project”  TASK-1494

Guitlle:
### 🗒️ Checklist

1. [x] run linter locally
2. [x] update all related docs (API, README, inline, etc.), if any
3. [x] draft PR with a title `<type>(<scope>)<!>: <title> TASK-1234`
4. [x] tag PR: at least `frontend` or `backend` unless it's global
5. [x] fill in the template below and delete template comments
6. [x] review thyself: read the diff and repro the preview as written
7. [x] open PR & confirm that CI passes
8. [x] request reviewers, if needed
9. [ ] delete this section before merging

### 📣 Summary
Add a new endpoint to delete all the permissions for a user by deleting any one of the user permissions

### 👀 Preview steps

1. Login as a user (user A)
2. Create or use an existing project to share it with another user (user B)
3. Log in as user B and go to the project page, 
4. Go to the "Form" tab, and in the three dots button, in the top right corner, click on the option "Remove shared project"
5. Log out
6. Log in as user A and check the project sharing settings, the user B should not appear anymore.

PR #5543, opened 2025-02-19
feat(permissions): use `delete-all` permissions endpoint on frontend TASK-1494

RuthShryock:
### 📣 Summary
The "Remove shared project" button and the "Delete" button in the sharing modal were not fully removing users' access, so they have been updated to completely remove all their permissions and ensure they are fully removed from the project.

### 👀 Preview steps
(Same as for backend pr: #5539)

**Test "Remove shared project" button** 

1. Login as a user (user A)
2. Create or use an existing project to share it with another user (user B)
3. Log in as user B and go to the project page, 
4. Go to the "Form" tab, and in the three dots button, in the top right corner, click on the option "Remove shared project"
5. Log out
6. Log in as user A and check the project sharing settings, the user B should not appear anymore.
7. 🔴 [on main] notice that user B still appears with "add submissions" permission
8. 🟢 [on PR] notice that user B does not appear

**Test "Delete" button in the sharing modal** 

9. Repeat steps 1-3 (note: you must assign user B with the `manage-asset` permission in order to access the Sharing Form)
10. Go to the "Summary" tab and click on the "Share Project" button 
11. A modal opens, click on the trash can/"Delete" button for the user B
12. 🔴 [on main] notice that api calls are made to the permission-assignments endpoint
13. 🟢 [on PR] notice that the delete-all endpoint is called
(Deletion behavior should remain the same as on main, just the endpoints change)


magicznyleszek (approved):
@RuthShryock FE part looks good and everything works correctly :)

magicznyleszek (changes requested):
I had problem with the reproduction steps for second case - without `manage-asset` permission, I am not able to open Sharing Form :)

As for the UX, this should be improved:
1. Share a form (`manage-asset` permission) from user A to user B
2. Log in as B
3. Go to `/#/forms/<uid>/settings/sharing` and use trash can button for user B row
4. 🟢 The call to `delete-all` is successful
5. 🔴 The call to fetch permissions returns 404
6. 🔴 User B stays on the route without access to the project

The same goes for sharing modal.

I think the fix would be to redirect user to main route after they delete themselves :)

magicznyleszek (approved):
Everything fixed :)

PR #5544, opened 2025-02-19
feat(massEmail): create new models for mass email TASK-751

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📖 Description
Developer-only changes

### 💭 Notes
This is just meant as set up and will not add any functionality. Most of it comes from the default django app setup command. There will undoubtedly be changes as we go but at least now we have a consistent place to make those changes.
No tests yet because there isn't any logic to test.
Had to add a "noqa" annotation to the long line in the migration because the way the formatter wants to fix it actually changes the meaning slightly (has to do with messing around with quotes) and it causes django to want to make a new migration.

### Preview
Feature/no-change template:
2. Go to admin
4. 🟢 There should be a new Mass emails config object that you can add/edit
5. :green_circle: When adding a config, there will be no available options for the query because none have been implemented yet


rajpatel24 (approved):
LGTM

PR #5545, opened 2025-02-20
fix(invites): display invite accepted banner for users who join another organization TASK-1601

magicznyleszek:
### 💭 Notes
Include organization id in the local storage key for banner dismissal. Also move `orgQuery` outside of `OrgInviteAcceptedBanner` to make it simpler to get organization id 👌

### 👀 Preview steps

Bug template:
1. ℹ️ have three accounts: "joe", "moe" and "bo"; "joe" and "moe" each are org owners, "bo" is not
2. send invite from "joe" to "bo" for joining joe's organization
3. accept invite as "bo"
4. 🟢 as "bo" notice that the invite accepted banner shows up (on top of `/#/projects/home` route)
5. dismiss the banner with "x" button (a localStorage item will be added)
6. as "joe" then remove "bo" from organization
7. send invite from "moe" to "bo" for joining moe's organization
8. accept invite as "bo"
9. 🔴 [on main] as "bo" notice that the banner doesn't appear
10. 🟢 [on PR] as "bo" notice that the banner does appear

### DIFF
--- a/jsapp/js/account/organization/invites/OrgInviteAcceptedBanner.tsx
+++ b/jsapp/js/account/organization/invites/OrgInviteAcceptedBanner.tsx
@@ -4,10 +4,15 @@ import React, { useEffect, useState } from 'react'
 import Alert from 'js/components/common/alert'
 // Stores, hooks and utilities
 import { useOrganizationMemberDetailQuery } from 'jsapp/js/account/organization/membersQuery'
-import { useOrganizationQuery } from '../organizationQuery'
 import { useSafeUsernameStorageKey } from 'jsapp/js/hooks/useSafeUsernameStorageKey'
 // Constants and types
-import { MemberInviteStatus } from 'js/account/organization//membersInviteQuery'
+import { type Organization } from 'js/account/organization/organizationQuery'
+import { MemberInviteStatus } from 'js/account/organization/membersInviteQuery'
+
+interface OrgInviteAcceptedBannerProps {
+  username: string
+  organization: Organization
+}
 
 /**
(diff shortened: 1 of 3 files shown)


jsapp/js/account/organization/invites/OrgInviteAcceptedBanner.tsx
```diff
@@ -4,22 +4,33 @@ import React, { useEffect, useState } from 'react'
...
   const organizationMemberDetailQuery = useOrganizationMemberDetailQuery(props.username, false)
-  const orgQuery = useOrganizationQuery()
   const [isBannerDismissed, setIsBannerDismissed] = useState<boolean>(false)
-  const localStorageKey = useSafeUsernameStorageKey('kpiOrgInviteAcceptedBanner', props.username)
+  /**
+   * For the local storage key we include organization id, because otherwise the banner would not appear when user would
+   * leave one organization and join another.
+   */
+  const localStorageKey = useSafeUsernameStorageKey(
+    `kpiOrgInviteAcceptedBanner-${props.organization.id}`,
```
jamesrkiger, line 31: What would you think about using the invite url instead of the org id here? That way the token would be unique to the specific invitation it is related to. So someone could leave an org and come back however many times and the banner would work like intended.

jamesrkiger (changes requested):
Left one comment.

jsapp/js/account/organization/invites/OrgInviteAcceptedBanner.tsx
magicznyleszek: Good idea, will see what changes are needed here

jamesrkiger (approved):
lgtm

PR #5551, opened 2025-02-21
fix(projectTransfer)!: fix asset visibility issue when user joins another organization

rajpatel24:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Ensures that assets shared with an organization owner remain visible even after the asset owner joins a different organization.

### 📖 Description
Previously, when an external user shared an asset with an organization owner and later joined a different organization, the asset became invisible to the original organization. This fix ensures that the asset remains visible in both organizations, maintaining expected access permissions. A test case has been added to verify this behavior.

### DIFF
--- a/kpi/tests/test_assets.py
+++ b/kpi/tests/test_assets.py
@@ -14,7 +14,7 @@ from rest_framework import serializers, status
 from kobo.apps.kobo_auth.shortcuts import User
 from kobo.apps.openrosa.apps.logger.models import XForm
 from kobo.apps.openrosa.apps.logger.xform_instance_parser import XFormInstanceParser
-from kobo.apps.organizations.models import Organization, OrganizationUser
+from kobo.apps.organizations.models import Organization
 from kobo.apps.organizations.tasks import transfer_member_data_ownership_to_org
 from kobo.apps.organizations.tests.test_organizations_api import (
     BaseOrganizationAssetApiTestCase
@@ -922,6 +922,7 @@ class TestAssetExcludedFromProjectsListFlag(BaseOrganizationAssetApiTestCase):
         self.thirduser = User.objects.create_user(
             username='thirduser', password='thirduser'
         )
+        self.asset_list_url = reverse(self._get_endpoint('asset-list'))
         self.asset_detail_url = lambda uid: reverse(
             self._get_endpoint('asset-detail'),
             kwargs={'uid': uid}
@@ -930,12 +931,13 @@ class TestAssetExcludedFromProjectsListFlag(BaseOrganizationAssetApiTestCase):
             self._get_endpoint('project-ownership-invite-detail'),
             kwargs={'uid': uid}
         )
+        self.org_assets_list_url = lambda org_id: reverse(
+            self._get_endpoint('organizations-assets'),
+            kwargs={'id': org_id}
+        )
 
-    def _add_user_to_organization(self, user):
-        org_user = OrganizationUser.objects.get(user=user)
-        Organization.objects.filter(organization_users=org_user).delete()
-        org_user.organization = self.organization
-        org_user.save()
+    def _add_user_to_organization(self, user, organization):
+        organization.add_user(user)
 
         # Transfer the ownership of the user's assets to the organization
         transfer_member_data_ownership_to_org(user.pk)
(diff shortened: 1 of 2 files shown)


kpi/tests/test_assets.py
```diff
@@ -931,10 +932,10 @@ def setUp(self):
             kwargs={'uid': uid}
         )
 
-    def _add_user_to_organization(self, user):
+    def _add_user_to_organization(self, user, organization):
         org_user = OrganizationUser.objects.get(user=user)
         Organization.objects.filter(organization_users=org_user).delete()
-        org_user.organization = self.organization
+        org_user.organization = organization
         org_user.save()
```
noliveleger, line 939: Could these lines not be replaced by `organization.add_user(user)`? 
Moreover it would ensure that the org is really a MMO before adding the user to it.

kpi/tests/test_assets.py
```diff
@@ -1002,3 +1003,40 @@ def test_asset_is_excluded_from_projects_list_flag(self):
...
         self.assertFalse(owner_asset.is_excluded_from_projects_list)
+
+    def test_asset_visibility_after_transfer(self):
+        """
+        Test to ensure that an asset shared with an organization owner remains
+        visible after the asset owner joins a different organization
+        """
+        # Step 1: Create an asset owned by an external user
+        self._create_asset_by_bob()
+        asset = Asset.objects.get(owner=self.external_user)
```
noliveleger, line 1014: I think it would be better to use the `uid` of the response to ensure we retrieve the correct asset.

```
response = self._create_asset_by_bob()
asset = Asset.objects.get(uid=response.data['uid'])
```

kpi/tests/test_assets.py
```diff
@@ -1002,3 +1003,40 @@ def test_asset_is_excluded_from_projects_list_flag(self):
...
+            any(obj['uid'] == asset.uid for obj in response.data['results'])
+        )
+
+        # Step 4: Invite the external user to another organization (OrgB)
+        another_org = Organization.objects.create(
+            id='org1234', name='Another Organization', mmo_override=True
+        )
+        another_org.add_user(self.thirduser)
+
+        self.client.force_login(self.org_owner)
```
noliveleger, line 1034: Why do we `force_login` again? Isn't `self.org_owner` already logged in (line 1021)?

kpi/tests/test_assets.py
```diff
@@ -1002,3 +1003,40 @@ def test_asset_is_excluded_from_projects_list_flag(self):
...
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertTrue(
+            any(obj['uid'] == asset.uid for obj in response.data['results'])
+        )
+
+        # Step 4: Invite the external user to another organization (OrgB)
+        another_org = Organization.objects.create(
+            id='org1234', name='Another Organization', mmo_override=True
+        )
+        another_org.add_user(self.thirduser)
```
noliveleger, line 1032: I guess, `self.thirduser` is the owner of `another_org`, right? 
Would you mind to change the name of the `another_org` and use a more obvious variable name to represent that `thirduser` is the owner. Like `thiruser_org` ;-)

kpi/tests/test_assets.py
```diff
@@ -1002,3 +1003,40 @@ def test_asset_is_excluded_from_projects_list_flag(self):
...
+
+        self.client.force_login(self.org_owner)
+        self._add_user_to_organization(self.external_user, another_org)
+
+        # Step 5: Verify that the asset is still visible in OrgA and OrgB
```
noliveleger, line 1037: I suppose that, here, we want to test (only) that OrgB owner can see the asset. We already test the visibility for OrgA previously? Please update your comment. Otherwise, something is missing here.
We are still logged as OrgA owner and we only test the response of the "MyProjects" list.
OrgA owner should see the project in their "MyProjects" list, but OrgB owner should see the project in their "My Org Projects" list.

noliveleger (changes requested):

kpi/tests/test_assets.py
rajpatel24: Correct!

kpi/tests/test_assets.py
rajpatel24: Yes, `self.org_owner` is already logged in, so we don't need to force login again. Updated this as well.

noliveleger (approved):

PR #5559, opened 2025-02-27
fix(NLP): missing suggested language in language selector DEV-17

magicznyleszek:
### 💭 Notes
Loop of asynchronous calls was bad. I made Promise.all that fixes the problem. 

### 👀 Preview steps
I can reproduce this on kf.main but I believe this is on production too.

1. Create a form with one audio question
2. Create two submissions (each with one audio file)
3. From the data table, click Open for the first submission.
4. Create a manual transcription in English. Save it.
5. Delete the transcript.
6. Go to the second submission (from the data table or by just clicking the up/down button from the same screen).
7. Create a manual transcription in Spanish. Save it.
8. Go back to the first subscription. Try to create a new transcript in English.

Expected behavior: The top of the language list shows English in the first position.

Instead: English is not displayed at all (and can't be found by searching either). Spanish is displayed in first position.

### DIFF
--- a/jsapp/js/components/languages/languageSelector.tsx
+++ b/jsapp/js/components/languages/languageSelector.tsx
@@ -179,22 +179,18 @@ class LanguageSelector extends React.Component<
   async fetchSuggestedLanguages() {
     this.setState({ suggestedLanguages: undefined });
     if (this.props.suggestedLanguages) {
-      try {
-        const languages = await Promise.all(
-          this.props.suggestedLanguages.map(async (languageCode) => {
-            try {
-              return await languagesStore.getLanguage(languageCode);
-            } catch (error) {
-              console.error(`Language ${languageCode} not found 2`);
-              return null;
-            }
-          })
-        );
-        const suggestedLanguages = languages.filter((language) => language !== null);
-        this.setState({ suggestedLanguages });
-      } catch (error) {
-        console.error('Error fetching suggested languages', error);
-      }
+      const languages = await Promise.all(
+        this.props.suggestedLanguages.map(async (languageCode) => {
+          try {
+            return await languagesStore.getLanguage(languageCode);
+          } catch (error) {
+            console.error(`Language ${languageCode} not found 2`);
+            return null;
+          }
+        })
+      );
+      const suggestedLanguages = languages.filter((language) => language !== null);
+      this.setState({ suggestedLanguages });
     }
   }
 
--- a/jsapp/js/components/languages/languageSelector.tsx
+++ b/jsapp/js/components/languages/languageSelector.tsx
@@ -190,8 +190,8 @@ class LanguageSelector extends React.Component<
             }
           })
         );
-        const validLanguages = languages.filter((language) => language !== null);
-        this.setState({ suggestedLanguages: validLanguages });
+        const suggestedLanguages = languages.filter((language) => language !== null);
+        this.setState({ suggestedLanguages });
       } catch (error) {
         console.error('Error fetching suggested languages', error);
       }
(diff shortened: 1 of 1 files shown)


jsapp/js/components/languages/languageSelector.tsx
```diff
@@ -176,27 +176,25 @@ class LanguageSelector extends React.Component<
...
+            try {
+              return await languagesStore.getLanguage(languageCode);
+            } catch (error) {
+              console.error(`Language ${languageCode} not found 2`);
+              return null;
+            }
+          })
+        );
+        const validLanguages = languages.filter((language) => language !== null);
+        this.setState({ suggestedLanguages: validLanguages });
```
Akuukis, line 194: I don't see how a new synonym helps more than confuses.
```suggestion
        const suggestedLanguages = languages.filter((language) => language !== null);
        this.setState({ suggestedLanguages });
```

jsapp/js/components/languages/languageSelector.tsx
```diff
@@ -176,27 +176,25 @@ class LanguageSelector extends React.Component<
...
-          console.error(`Language ${languageCode} not found 2`);
-        }
-      });
+      try {
+        const languages = await Promise.all(
+          this.props.suggestedLanguages.map(async (languageCode) => {
+            try {
+              return await languagesStore.getLanguage(languageCode);
+            } catch (error) {
+              console.error(`Language ${languageCode} not found 2`);
```
Akuukis, line 188: typo?
```suggestion
              console.error(`Language ${languageCode} not found`);
```

jsapp/js/components/languages/languageSelector.tsx
```diff
@@ -176,27 +176,25 @@ class LanguageSelector extends React.Component<
...
+              return null;
+            }
+          })
+        );
+        const validLanguages = languages.filter((language) => language !== null);
+        this.setState({ suggestedLanguages: validLanguages });
+      } catch (error) {
+        console.error('Error fetching suggested languages', error);
```
Akuukis, line 196: Fetching errors will be caught in the try-catch above. This may error only if `setState` throws, but does `setState` even throws at all? 
```suggestion
        console.error('Failed to set suggested languages', error);
```

Akuukis (approved):
✅ Code LGTM, see style nitpicks inline.

✅ happy path tested previewed:
- [x] reproduced the bug on the main
- [x] reproduced the bug on the commit before PR
- [x] reproduced the expected behavior on PR

:warning: did not stumble upon errors so that's not previewed.

jsapp/js/components/languages/languageSelector.tsx
magicznyleszek: If you search for `} not found ` in the codebase, you will see we have 6 identical errors. John added the numbers so it's clear which one is which when looking into them in Sentry

jsapp/js/components/languages/languageSelector.tsx
magicznyleszek: I think you're right. I am removing the unneeded `try catch`

jsapp/js/components/languages/languageSelector.tsx
Akuukis: Oh ok, we'd rather write something descriptive instead (like the method name calling it) but that's out of scope of this PR.

magicznyleszek: Agree. I made a task for this: https://linear.app/kobotoolbox/issue/DEV-18/rename-numbered-language-x-not-found-1-errors-to-be-more-descriptive

PR #5562, opened 2025-02-27
feat(massEmail): query for users reaching their maximum storage limit TASK-1454

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Create the query for retrieving users who are >90% and >100% of their storage limits.

### 📖 Description
The query returns a list of user uids. It will be used as part of a MassEmailConfig. Emails that use that query will be sent to the list of users generated by it. The usage calculated is the same as what is shown on the Usage page.

### 💭 Notes

Part of this PR is taking the logic we use to determine the limits for one organization and expanding it to be able to calculate the limits for multiple organizations. This means moving some of the python logic to the ORM query, but the results should be exactly the same. In particular, the selection of price metadata over product metadata is done with a Coalesce call rather than fetching both and then choosing which one to use in the python processing code. All unit tests still passed. Also added a few new ones.

Note: This will affect how we get all usage limits, not just storage bytes, but again it should be a pure refactor so nothing user-facing will change. This should make it easier in the future to write the queries for those approaching their NLP and submission limits.

Similarly, this PR extends the usage calculator query for submission bytes to get the usage for multiple users at once.

The actual user queries get all user limits, then all user storage, then compare the two to see who is within the specified range.

### 👀 Preview steps

* Note: after every "purchase," make sure to resync the subscriber

1. Enable stripe on your local setup
2. Have at least 2 organizations/accounts. 
3. Set up org2 with the Professional Plan. Keep org1 on the community plan.
4. Create a project in org2 with an audio question and add a few submissions with attachments.
5. In admin, go to Djstripe -> Product -> Up to 5 GB of file storage
6. Edit the product metadata with ` "storage_bytes_limit": "111000"` and save
7. Log in as the org1 owner and purchase the 5GB storage add on
8. Still as the org1 owner, create and deploy a new project with an audio question.
9. Add a submission with [sample-6s.zip](https://github.com/user-attachments/files/19032797/sample-6s.zip). This file is ~103K so it should put you between 90 and 100 percent of the limit
10. In the shell, import and run `get_users_over_90_percent_of_storage_limit()`
11. 🟢 the result should be a list with just the user_uid of the owner of org1
12. Run `get_users_over_100_percent_of_storage_limit()`
13. 🟢 The result should be an empty list
14. Add another submission with the same audio file. This should push you over the storage limit.
15. In the shell, rerun `get_users_over_100_percent_of_storage_limit()`
16. 🟢 The result should be a list with just the user_uid of the owner of org1
17. As the org1 owner, upgrade to the 10GB storage addon
18. In the shell, rerun `get_users_over_90_percent_of_storage_limit()` and `get_users_over_100_percent_of_storage_limit()`. 
19. 🟢 Both should result in an empty list


noliveleger (changes requested):

jamesrkiger (approved):
LGTM

kpi/utils/usage_calculator.py
```diff
@@ -12,6 +12,16 @@
...
 
 
+def get_storage_usage_by_user_id(user_ids: list[int] = None) -> dict[int, int]:
+    xforms = XForm.objects.exclude(pending_delete=True)
+    if user_ids is not None:
+        xforms = xforms.filter(user_id__in=user_ids)
+    xform_query = xforms.values('user').annotate(
+        bytes_sum=Coalesce(Sum('attachment_storage_bytes'), 0)
+    )
+    return {res['user']: res['bytes_sum'] for res in xform_query}
```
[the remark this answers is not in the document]
noliveleger: As per our discussion, we won't change anything.
See internal discussion https://chat.kobotoolbox.org/#narrow/stream/4-Kobo-Dev/topic/Mass.20emails.20feature/near/564307

kobo/apps/stripe/utils.py
```diff
@@ -34,53 +103,9 @@ def get_organization_plan_limit(
...
-
-    if relevant_limit == 'unlimited':
-        return inf
-
-    if relevant_limit:
-        return int(relevant_limit)
-
-    return inf
+    return get_organization_plan_limits(usage_type, [organization]).get(
+        organization.id, None
```
noliveleger, line 107: I think `None` is redundant in that case. 

```
d = {}
d.get('foo') is None  # True
```

kobo/apps/mass_emails/user_queries.py
```diff
@@ -0,0 +1,47 @@
...
+) -> QuerySet:
+    """
+    Returns all users whose storage usage is between minimum and maximum percent
+    of their plan limit
+
+    :param minimum: float. Minimum usage, eg 0.9 for 90% of the limit. Default 0
+    :param maximum: float. Maximum usage, eg 1 for 100% of the limit. Default inf
+    """
+    minimum = minimum or 0
+    maximum = maximum or inf
```
rajpatel24, line 22: Do we really need to set `minimum = minimum or 0` and `maximum = maximum or inf`, given that the default values for minimum and maximum are already set in the function signature?

rajpatel24 (approved):
LGTM

PR #5564, opened 2025-03-03
fix(openRosa): correct project lookup fallback between OpenRosa and KPI TASK-1645

noliveleger:
### 📣 Summary
Fixed a typo in the project lookup fallback when resolving links between OpenRosa and KPI.

### 📖 Description
A typo in the lookup fallback caused issues when resolving the link between KPI projects (Asset) and their corresponding OpenRosa projects (XForm). This fix corrects the fallback logic, ensuring that the correct KPI project is retrieved from an OpenRosa project when necessary.

### 👀 Preview steps
The preview steps are hard to reproduce since a project is now always deployed from KPI (not support by `v1` anymore).
See the new unit test to simulate preview steps.


rajpatel24 (approved):
LGTM

PR #5569, opened 2025-03-04
feat(massEmail): query for identifying inactive users based on login, form activity, and submissions TASK-1581

rajpatel24:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Detect and return a list of users who have been inactive for a specified period based on login activity, project updates, and submissions.

### 📖 Description
This update introduces a method to identify inactive users who have not:

- Logged into their account.
- Modified their projects or had modifications made by others.
- Received submissions to their projects.

By default, the inactivity period is set to 365 days but can be configured dynamically.

### 👀 Preview steps

1. Run `get_inactive_users(days=1)` in the shell, users inactive for 1+ days should be listed.
2. Modify `inactivity_threshold = now() - timedelta(minutes=2)`
3. Log in & log out, run `get_inactive_users()`, user should NOT be listed. Wait 2 mins, run again, user should be listed now.
4. Log in, create a project, run `get_inactive_users()`, user should NOT be listed. Wait 2 mins, run again, user should be listed now.
5. Change the name of the project, run `get_inactive_users()`, user should NOT be listed. Wait 2 mins, run again, user should be listed now. 
6. Add a submission, run `get_inactive_users()`, user should NOT be listed. Wait 2 mins, run again, user should be listed.


kobo/apps/mass_emails/user_queries.py
```diff
@@ -1,13 +1,57 @@
...
+
+    :return: A queryset of inactive users
+    """
+    # Calculate the user inactivity threshold date
+    days = days or 365
+    inactivity_threshold = now() - timedelta(days=days)
+
+    # Identify users who have not logged in within the given period
+    inactive_users = User.objects.filter(
+        Q(last_login__lt=inactivity_threshold) | Q(last_login__isnull=True)
```
rgraber, line 34: This will catch new users who haven't logged in yet, which we don't want. Might need to check creation date as well.

kobo/apps/mass_emails/tests/test_user_inactive_query.py
```diff
@@ -0,0 +1,139 @@
...
+        user = self._create_user(username, getattr(self, last_login))
+        inactive_users = get_inactive_users()
+        self.assertEqual(expected, user in inactive_users)
+
+    @data(
+        ('old_date', True),
+        ('recent_date', False)
+    )
+    @unpack
+    def test_inactive_users_based_on_form_activity(self, date, expected):
```
rgraber, line 95: A more accurate test would be to create an asset with an old date, then hit the endpoint to modify the asset and make sure the user is no longer listed as inactive, rather than forcing the update/creation date directly.

kobo/apps/mass_emails/tests/test_user_inactive_query.py
```diff
@@ -0,0 +1,139 @@
...
+    def test_inactive_users_after_form_edit_without_redeployment(self):
+        user = self._create_user('active_asset', self.old_date)
+        asset = self._create_asset(user, self.old_date, self.old_date)
+
+        # Ensure user is in the inactive list
+        inactive_users = get_inactive_users()
+        self.assertEqual(True, user in inactive_users)
+
+        # Edit the asset without redeploying
+        Asset.objects.filter(id=asset.id).update(
```
rgraber, line 114: see above: a better test would be to modify the asset without manually forcing the last_modified date.

rgraber (changes requested):
This won't catch users who have deleted a project, which I think we should count as an update.

rajpatel24:
> This won't catch users who have deleted a project, which I think we should count as an update.

@rgraber Indeed, this won't catch projects deleted by users who haven't logged in, such as those deleted via the API or by an admin. I have a question: How exactly can we track deleted projects once they've been removed from the 'trash bin'? For example, if `ProjectA` is deleted, how can we track that `ProjectA` was deleted by `ownerA` on a specific date? Would audit logs be appropriate for retrieving this information, or is there a better approach we should consider?

kobo/apps/mass_emails/tests/test_user_inactive_query.py
```diff
@@ -0,0 +1,163 @@
...
+        asset = self._create_asset(user, self.old_date, self.old_date)
+        inactive_users = get_inactive_users()
+        self.assertTrue(user in inactive_users)
+
+        # Update the asset and ensure the user is no longer inactive
+        self._update_asset(asset, {'name': 'Updated asset'}, user)
+        inactive_users = get_inactive_users()
+        self.assertFalse(user in inactive_users)
+
+    def test_user_becomes_active_after_submission_update(self):
```
rgraber, line 119: I had completely forgotten about modifying submissions. Nice catch!

kobo/apps/mass_emails/user_queries.py
```diff
@@ -1,13 +1,58 @@
...
+from kpi.models import Asset
 from kpi.utils.usage_calculator import get_storage_usage_by_user_id
 
 
+def get_inactive_users(days: int = 365) -> QuerySet:
+    """
+    Retrieve users who have been inactive for a specified number of days.
+
+    A user is considered inactive if:
+    - They have not logged in within the given period (or never logged in).
```
rgraber, line 20: Can we update this comment to note that users who were recently created will not be considered inactive, even if they haven't logged in yet?

rgraber (approved):
One request to change a comment, but it's not a blocker. Nice job!

PR #5572, opened 2025-03-05
feat(massEmails): query for users approaching submission limits TASK-1579

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Query for users who are approaching their maximum submission limit.

### 📖 Description
Create queries for users who are over 90% and 100%, respectively, of their submission limit. These queries will be accessible by mass email jobs.

### 💭 Notes
Similar to https://github.com/kobotoolbox/kpi/pull/5562, this PR is mostly about taking existing rules for determining submission limits/submission counts and applying them to all organizations at once. It's not as simple as the storage limits because submissions are measured per billing period, so we have to calculate the current billing period for every org. 

Similar to the storage usage PR, we refactored the code for getting billing dates for an individual organization to just call the method for getting all billing dates with a filter. There's a unit test to make sure this doesn't increase the number of queries since the logic is a little more complicated. The basic logic is the same: look for an active subscription first, then a canceled subscription, then default to the first of the current month.

Unlike with the storage PR, we didn't also refactor the method to get all submissions within the billing period for an individual organization because it's very hard to do without increasing the number of queries (because of the variable billing period, which is calculated in python and not easily convertible to a SQL method). Instead we just add a new method that gets all submissions from the most recent billing period for all orgs.

This PR also moves some of the billing date calculation logic to stripe/utils.py because it's stripe-specific and not applicable to if Stripe is disabled.

Finally, this PR refactors `get_users_within_range_of_usage_limit` to work with either storage or submission usage types because the basic procedure is the same: get all limits, get all usage, and compare usage to limits.

### 👀 Preview steps

* Note: after every "purchase," make sure to resync the subscriber

1. ℹ️ have at least 2 accounts and at least one project on each account
3. ℹ️ have Stripe configured 
4. "Purchase" the Professional Plan for user1 (yearly or monthly). Leave user2 on the Community plan.
5. Go to `/admin` and update product metadata of the `Community (5 Prices)` with `"submission_limit":10`
6. Submit 9 submissions to one project on each account
7. In the shell, run `users_above_90_percent_submissions`
8. 🟢 user2 should appear in the results, user1 should not
9. Run `users_above_100_percent_submissions`
10. Neither user should appear in the results
11. Add another submission to a project in user2's account (ideally a different project than the one with the 9 subs).
12. In the shell, rerun `users_above_100_percent_submissions`
13. 🟢 user2 should appear in the results
14. "Purchase" a professional plan for user2
15. In the shell, rerun `users_above_100_percent_submissions`
16. 🟢 user2 should not longer appear in the results


rajpatel24 (approved):
LGTM

PR #5574, opened 2025-03-07
feat(massEmail): create/register celery tasks mass emails TASK-750

Guitlle:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Celery task and admin action to send emails for mass email configurations asynchronously by user request on the admin screen.

### 📖 Description
The mass email configs admin action "Send emails" triggers a celery task for each email config that creates or uses an existing mass email job that is associated to a set of mass email records. The records are linked to a user, they start with enqueued status and when they are processed by the celery task, their status is set to sent or failed. This is work in progress, future PRs will enable more functionality.

### 👀 Preview steps
1. (Optional) set the email backend: `EMAIL_BACKEND = 'django.core.mail.backends.filebased.EmailBackend'` to debug
2. Go to the Mass email configs admin page
3. Create a new mass email config instance
4. On the mass email configs admin page enable the checkbox to select the new mass email config, then in the action dropdown select "Send emails" and click on "Go"
5. Check that the emails were sent in the emails folder (kpi_home/emails/)
6. Check the logs at `kobo-docker/log/kpi/celery_kpi_worker.log` and ensure that they mention the emails sent.

### DIFF
--- a/kobo/apps/mass_emails/models.py
+++ b/kobo/apps/mass_emails/models.py
@@ -4,7 +4,6 @@ from django.db import models
 
 from kobo.apps.kobo_auth.shortcuts import User
 from kobo.apps.mass_emails.user_queries import (
-    get_all_users,
     get_inactive_users,
     get_users_over_90_percent_of_storage_limit,
     get_users_over_90_percent_of_submission_limit,
@@ -15,7 +14,6 @@ from kpi.fields import KpiUidField
 from kpi.models.abstract_models import AbstractTimeStampedModel
 
 USER_QUERIES: dict[str, Callable] = {
-    'all_users': get_all_users,
     'users_above_90_percent_storage': get_users_over_90_percent_of_storage_limit,
     'users_above_100_percent_storage': get_users_over_100_percent_of_storage_limit,
     'users_inactive_for_365_days': get_inactive_users,
--- a/kobo/apps/mass_emails/models.py
+++ b/kobo/apps/mass_emails/models.py
@@ -5,6 +5,7 @@ from django.db import models
 from kobo.apps.kobo_auth.shortcuts import User
 from kobo.apps.mass_emails.user_queries import (
     get_inactive_users,
+    get_all_users,
     get_users_over_90_percent_of_storage_limit,
     get_users_over_90_percent_of_submission_limit,
     get_users_over_100_percent_of_storage_limit,
@@ -14,6 +15,7 @@ from kpi.fields import KpiUidField
 from kpi.models.abstract_models import AbstractTimeStampedModel
 
 USER_QUERIES: dict[str, Callable] = {
+    'all_users': get_all_users,
     'users_above_90_percent_storage': get_users_over_90_percent_of_storage_limit,
     'users_above_100_percent_storage': get_users_over_100_percent_of_storage_limit,
     'users_inactive_for_365_days': get_inactive_users,
--- a/kobo/apps/mass_emails/models.py
+++ b/kobo/apps/mass_emails/models.py
@@ -4,8 +4,8 @@ from django.db import models
 
 from kobo.apps.kobo_auth.shortcuts import User
 from kobo.apps.mass_emails.user_queries import (
-    get_inactive_users,
     get_all_users,
+    get_inactive_users,
     get_users_over_90_percent_of_storage_limit,
     get_users_over_90_percent_of_submission_limit,
     get_users_over_100_percent_of_storage_limit,
--- a/kobo/apps/mass_emails/tasks.py
+++ b/kobo/apps/mass_emails/tasks.py
@@ -5,6 +5,7 @@ from constance import config
 from django.conf import settings
 from django.db.models import Count, Q
 from django.utils.timezone import now
+from django.utils.translation import gettext
 
 from kobo.celery import celery_app
 from kpi.utils.log import logging
@@ -111,10 +112,14 @@ def send_emails(email_config_uid: str):
         ]
         for record in records:
             logging.info(f'Processing MassEmailRecord({record})')
+            org_user = record.user.organization.organization_users.get(user=record.user)
+            plan_name = org_user.active_subscription_status
+            if plan_name == '' or plan_name is None:
+                plan_name = gettext('Community Plan')
             data = {
                 'username': record.user.username,
                 'full_name': record.user.first_name + ' ' + record.user.last_name,
-                'plan_name': 'Test plan',
+                'plan_name': plan_name,
             }
             content = render_template(email_config.template, data)
             message = EmailMessage(
--- a/kobo/apps/mass_emails/tasks.py
+++ b/kobo/apps/mass_emails/tasks.py
@@ -127,19 +127,19 @@ class MassEmailSender:
                     )
                 self.cache_limit_value(None, self.total_records)
             else:
-                total_limit = 0
+                day_limit = 0
                 for email_config in self.configs:
-                    if total_limit >= MAX_EMAILS:
+                    if day_limit >= MAX_EMAILS:
                         break
                     config_limit = ceil(
                         email_config.enqueued_records_count
                         / self.total_records
                         * MAX_EMAILS
                     )
-                    if total_limit + config_limit > MAX_EMAILS:
-                        config_limit = MAX_EMAILS - total_limits
+                    if day_limit + config_limit > MAX_EMAILS:
+                        config_limit = MAX_EMAILS - day_limit
                     self.cache_limit_value(email_config, config_limit)
-                    total_limit += config_limit
+                    day_limit += config_limit
                 self.cache_limit_value(None, MAX_EMAILS)
 
     def send_day_emails(self):
@@ -162,9 +162,12 @@ class MassEmailSender:
     def send_email(self, email_config, record):
         logging.info(f'Processing MassEmailRecord({record})')
         org_user = record.user.organization.organization_users.get(user=record.user)
-        plan_name = org_user.active_subscription_status
-        if plan_name == '' or plan_name is None:
-            plan_name = gettext('Community Plan')
+        if settings.STRIPE_ENABLED:
+            plan_name = org_user.active_subscription_status
+            if plan_name == '' or plan_name is None:
+                plan_name = gettext('Community Plan')
+        else:
+            plan_name = gettext('Not available')
         data = {
             'username': record.user.username,
             'full_name': record.user.first_name + ' ' + record.user.last_name,
--- a/kobo/apps/mass_emails/tasks.py
+++ b/kobo/apps/mass_emails/tasks.py
@@ -93,8 +93,11 @@ def send_emails(email_config_uid: str):
     ).get(uid=email_config_uid)
     if email_config is None:
         return
+    
+    if email_config.jobs_count > 1:
+        raise NotImplementedError
 
-    if email_config.jobs_count == 0:
+    if email_config.jobs_count == 0: # Create job if no job exists
         create_email_records(email_config)
 
     from_email = settings.DEFAULT_FROM_EMAIL
--- a/kobo/apps/mass_emails/tasks.py
+++ b/kobo/apps/mass_emails/tasks.py
@@ -1,4 +1,5 @@
 from datetime import timedelta
+from math import ceil
 
 from constance import config
 from django.conf import settings
@@ -10,12 +11,13 @@ from kpi.utils.log import logging
 from kpi.utils.mailer import EmailMessage, Mailer
 from kobo.apps.mass_emails.models import (
     EmailStatus,
+    MassEmailConfig,
     MassEmailRecord,
     MassEmailJob,
     USER_QUERIES
 )
 
-placeholder_values = {
+templates_placeholders = {
     '##username##': 'username',
     '##full_name##': 'full_name',
     '##plan_name##': 'plan_name',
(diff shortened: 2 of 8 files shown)


kobo/apps/mass_emails/tasks.py
```diff
@@ -0,0 +1,88 @@
...
+    for job in email_config.jobs.all():
+        records = job.records.filter(status=EmailStatus.ENQUEUED)[
+            : settings.MAX_MASS_EMAILS_PER_DAY
+        ]
+        for record in records:
+            logging.info(f'Processing MassEmailRecord({record})')
+            data = {
+                'username': record.user.username,
+                'full_name': record.user.first_name + ' ' + record.user.last_name,
+                'plan_name': 'Test plan',
```
rgraber, line 73: Is this being put off to a future PR? If so I think we should just not have plan_name as a variable to start out with rather than risk sending a bunch of emails with 'Test plan' in them

kobo/apps/mass_emails/tasks.py
```diff
@@ -0,0 +1,88 @@
...
+        create_email_records(email_config)
+
+    from_email = settings.DEFAULT_FROM_EMAIL
+    logging.info(
+        f'Processing MassEmailConfig(uid={email_config.uid}, '
+        f'name={email_config.name}, subject={email_config.subject})'
+    )
+    for job in email_config.jobs.all():
+        records = job.records.filter(status=EmailStatus.ENQUEUED)[
+            : settings.MAX_MASS_EMAILS_PER_DAY
```
rgraber, line 66: Are we handling sending the remaining emails in a future PR?

rgraber, line 66: The max mass emails per day needs to be divided among all existing MassEmailConfigs (see the note in the requirements)

rgraber (changes requested):

kobo/apps/mass_emails/models.py
```diff
@@ -5,13 +5,15 @@
...
     get_inactive_users,
+    get_all_users,
     get_users_over_90_percent_of_storage_limit,
     get_users_over_100_percent_of_storage_limit,
 )
 from kpi.fields import KpiUidField
 from kpi.models.abstract_models import AbstractTimeStampedModel
 
 USER_QUERIES: dict[str, Callable] = {
+    'all_users': get_all_users,
```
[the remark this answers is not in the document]
Guitlle: To easily test the email feature

kobo/apps/mass_emails/tasks.py
```diff
@@ -0,0 +1,93 @@
...
+    job = MassEmailJob.objects.create(
+        email_config=email_config,
+    )
+    users = USER_QUERIES[email_config.query]()
+    for user in users:
+        MassEmailRecord.objects.create(
+            user=user,
+            email_job=job,
+            status=EmailStatus.ENQUEUED,
+        )
```
rajpatel24, line 40: @Guitlle Would it be possible to use `bulk_create()` here for improved efficiency? It would reduce the number of queries and ensure that all records are created in a single transaction

Guitlle: yes,

kobo/apps/mass_emails/tasks.py
Guitlle: I think so, this is only for the admin action and some logic, but the automated task will handle this.

kobo/apps/mass_emails/tasks.py
```diff
@@ -0,0 +1,122 @@
...
+    ]
+    MassEmailRecord.objects.bulk_create(records)
+
+
+@celery_app.task
+def send_emails(email_config_uid: str, should_create_job: bool = False):
+    today = date.today().isoformat()
+    cache_key = f'MassEmailConfig_Lock_{today}_{email_config_uid}'
+    if cache.get(cache_key) == 'Locked':
+        raise ExecutionBlockedException
```
rajpatel24, line 57: This doesn't show any error to the admin, it definitely blocks the execution of the same config. However, would it be beneficial to display an error message to the admin? What you think?

kobo/apps/mass_emails/tasks.py
```diff
@@ -0,0 +1,122 @@
...
+    cache.set(cache_key, 'Locked', 60*60*24) # 1 day lock
```
rgraber, line 58: I think we can simplify this a bit without having to use the cache. We can have something like
```
current_running_jobs = my_config.jobs.filter(Q(jobs__records__status=EmailStatus.ENQUEUED))
if current_running_jobs > 0:
    raise NotGonnaHappen('Send is already in progress')
...otherwise, create a new job and go on our way
```
we have a task to cleanup records that have been enqueued for too long, so even if whatever called send_email before is no longer running, we should still get unblocked eventually

kobo/apps/mass_emails/tasks.py
Guitlle: It's no longer an admin action

kobo/apps/mass_emails/tasks.py
```diff
@@ -0,0 +1,122 @@
...
+            email_job=job,
+            status=EmailStatus.ENQUEUED,
+        )
+        for user in users
+    ]
+    MassEmailRecord.objects.bulk_create(records)
+
+
+@celery_app.task
+def send_emails(email_config_uid: str, should_create_job: bool = False):
```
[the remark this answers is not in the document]
Guitlle: No longer needed, it was attempting to do too much in one function

kobo/apps/mass_emails/tasks.py
```diff
@@ -33,20 +67,115 @@ def mark_old_enqueued_mass_email_record_as_failed():
...
+                continue
+            records = MassEmailRecord.objects.filter(
+                status=EmailStatus.ENQUEUED,
+                email_job__email_config=email_config,
+            )[:limit]
+            logging.info(
+                f'Processing {limit} records for MassEmailConfig({email_config})'
+            )
+            for record in records:
+                self.send_email(email_config, record)
```
rajpatel24, line 133: I agree with Becca. It would be better to decrease the limit as each email is sent instead of setting it to zero immediately. This way, if the task fails before finishing, the remaining emails for that config can still be processed in subsequent runs.

kobo/apps/mass_emails/tests/test_celery_tasks.py
```diff
@@ -0,0 +1,89 @@
...
+            'username': 'Test Username',
+            'full_name': 'Test Full Name',
+            'plan_name': 'Test Plan Name',
+            'date_created': 'date_created',
+        }
+
+        rendered = render_template(self.template, data)
+        assert 'Username: Test Username' in rendered
+        assert 'Full name: Test Full Name' in rendered
+        assert 'Plan name: Test Plan Name' in rendered
```
rajpatel24, line 89: Would it be possible to add a test to ensure that the limits decrement as each email is sent, rather than being reset immediately? :-)

rajpatel24 (changes requested):

kobo/apps/mass_emails/tasks.py
```diff
@@ -33,20 +67,115 @@ def mark_old_enqueued_mass_email_record_as_failed():
...
+            self.limits = {}
+            if self.total_records < MAX_EMAILS:
+                for email_config in self.configs:
+                    self.limits[email_config.id] = email_config.enqueued_records_count
+            else:
+                total_limits = 0
+                for email_config in self.configs:
+                    if total_limits >= MAX_EMAILS:
+                        break
+                    config_limit = ceil(
```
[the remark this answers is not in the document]
Guitlle: If we use floor and the number is 0.3, 0.9, it will round to 0, though

kobo/apps/mass_emails/tasks.py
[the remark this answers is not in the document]
Guitlle: Yes, that's better, we should not have it all in one dict but have separate values for each config, that way it is faster

kobo/apps/mass_emails/tests/test_celery_tasks.py
```diff
@@ -0,0 +1,89 @@
...
+    def test_send_emails_limits(self, now_mock, now_mock_B):
+        now_mock.return_value = datetime(2025, 1, 1, 0, 0, 0, 0, pytz.UTC)
+        now_mock_B.return_value = now_mock.return_value
+        send_emails()
+        assert len(mail.outbox) == 10
+        now_mock.return_value = datetime(2025, 1, 2, 0, 0, 0, 0, pytz.UTC)
+        now_mock_B.return_value = now_mock.return_value
+        send_emails()
+        assert len(mail.outbox) == 20
+        now_mock.return_value = datetime(2025, 1, 2, 0, 0, 0, 0, pytz.UTC)
```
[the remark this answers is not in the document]
Guitlle: You're right, it's unnecessary

kobo/apps/mass_emails/tests/test_celery_tasks.py
Guitlle: Yes, I can make that unit test

kobo/apps/mass_emails/tasks.py
```diff
@@ -33,20 +67,130 @@ def mark_old_enqueued_mass_email_record_as_failed():
...
 
-    logging.info(
-        f'Created {len(records)} MassEmailRecord(s) for {email_config.name} '
-        f'with query {email_config.query}'
-    )
+class MassEmailSender:
+
+    def __init__(self):
+        self.today = localdate()
+        self.cache_key_prefix = f'mass_emails_{self.today.isoformat()}_email_limits'
```
rgraber, line 81: What do you think about changing this to `emails_remaining` instead of `email_limits`? By no means mandatory but I think it would be a little clearer.

kobo/apps/mass_emails/tasks.py
rgraber: I do think we want to decrement the counter before sending the email though. It's a little counter-intuitive but it reduces the risk of sending too many emails because the task was canceled or died between sending the email and decrementing the counter

rgraber (changes requested):
One more hopefully small request and a non-blocking thought about naming things

rgraber:
For posterity: to test this, I had to update the `get_inactive_users` query to look for users who were inactive for 2 minutes rather than a year

kobo/apps/mass_emails/tasks.py
```diff
@@ -33,20 +67,131 @@ def mark_old_enqueued_mass_email_record_as_failed():
...
+        self.limits = {}
+        self.total_limit = cache.get(f'{self.cache_key_prefix}_total')
+
+        if self.total_limit is not None:
+            for email_config in self.configs:
+                stored_limit = cache.get(f'{self.cache_key_prefix}_{email_config.id}')
+                self.limits[email_config.id] = stored_limit
+        else:
+            logging.info('Setting up MassEmailConfig limits for the current day')
+            MAX_EMAILS = settings.MAX_MASS_EMAILS_PER_DAY
```
rajpatel24, line 122: The declaration of `MAX_EMAILS = settings.MAX_MASS_EMAILS_PER_DAY` is redundant as it is already defined above

kobo/apps/mass_emails/tasks.py
```diff
@@ -33,20 +67,131 @@ def mark_old_enqueued_mass_email_record_as_failed():
...
+                    if total_limit >= MAX_EMAILS:
+                        break
+                    config_limit = ceil(
+                        email_config.enqueued_records_count
+                        / self.total_records
+                        * MAX_EMAILS
+                    )
+                    if total_limit + config_limit > MAX_EMAILS:
+                        config_limit = MAX_EMAILS - total_limits
+                    self.cache_limit_value(email_config, config_limit)
```
rajpatel24, line 141: Please correct the undefined variable `total_limits` to `total_limit`

rajpatel24 (changes requested):

kobo/apps/mass_emails/tasks.py
rgraber: I think it's supposed to be total_limits, since the idea is just to give this config however many slots we have left.

kobo/apps/mass_emails/tasks.py
```diff
@@ -33,20 +67,131 @@ def mark_old_enqueued_mass_email_record_as_failed():
...
+            )
+            for record in records:
+                self.cache_limit_value(email_config, self.limits[email_config.id] - 1)
+                self.cache_limit_value(None, self.total_limit - 1)
+                self.send_email(email_config, record)
+
+    def send_email(self, email_config, record):
+        logging.info(f'Processing MassEmailRecord({record})')
+        org_user = record.user.organization.organization_users.get(user=record.user)
+        plan_name = org_user.active_subscription_status
```
rajpatel24, line 165: Not sure if you have already planned it, but as Becca suggested on Zulip, we need to add a condition to check whether Stripe is enabled, as we need to be able to send emails to inactive users on non-stripe servers :-)

kobo/apps/mass_emails/tasks.py
Guitlle: I fixed it, changed the name to avoid confusion with existing names

kpi/exceptions.py
```diff
@@ -223,3 +223,7 @@ class XPathNotFoundException(Exception):
 
 class XlsFormatException(Exception):
     pass
+
+
+class ExecutionBlockedException(Exception):
```
rajpatel24, line 228: Do we still need this, since we are not using this exception anywhere now?

kobo/apps/mass_emails/tasks.py
rajpatel24: > The declaration of `MAX_EMAILS = settings.MAX_MASS_EMAILS_PER_DAY` is redundant as it is already defined above

Do we need this? I think it is redundant and should be removed

rgraber (approved):
Nice work

rajpatel24 (approved):
LGTM

PR #5577, opened 2025-03-10
feat(attachments): create query file for removing attachments and a way to test it TASK-570 TASK-572

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 💭 Notes
This has multiple different changes in it, so that it is much easier to test and review :)

The changes:
- Added upcoming endpoints to `api.endpoints.ts`
- Defined `removingAttachmentsEnabled` feature flag 🏴 
- Created `AttachmentActionsDropdown` component
  - This component is meant to be used in all places that display an attachment. It has an option to "Download" or "Delete" attachment.
  - It has a stories file that mostly works - notifications don't appear, and attachment is not being actually deleted, but you can get the general vibe of the component, and test `is_deleted` flag with it (see `submissionData` arg description)
  - Note: it is being hidden behind feature flag, so you need to set it for Storybook too
  - It has a mocks file that have an asset object with single submission with single attachment 👌 
- Uses `AttachmentActionsDropdown` inside Single Submission Modal, so that it can be tested in real scenario (apart from testing in Storybook)
- Improved `SubmissionResponse` interface to make it clearer which uuid we should use
  - Also changed `deployment__data_download_links`, as it turns out not all `ExportTypeName`s can happen in there :)

In the next PRs I am going to:
1. Use `AttachmentActionsDropdown` in the rest of the places (here I only add it in one place)
2. Write code for handling attachments deletion in Submission Modal (here I only left a comment)

### 👀 Preview steps

Testing:
1. ℹ️ Have account and a project with `image` question and a submission
2. Go to Project -> Data -> Table and open Single Submission Modal (eye button)
   - Note: enable `removingAttachmentsEnabled` feature flag (feature flag needs to be present in the url when you're opening the modal, because hook checks it when it runs during component start)
3. 🟢 Notice that "…" button appears next to the image
4. Click "…"
5. 🟢 Notice that "Download" and "Delete" options are present
   - Note: if your user doesn't have `change_submissions` permission, "Delete" will be disabled
6. 🟢 Upon clicking "Download", file will be downloaded
7. Click "Delete" option
8. 🟢 Notice a modal appearing with "Cancel" and "Delete" buttons
9. Click "Delete" button (it uses a mocked call)
   - 33% times an error will happen (error notification, and modal stays open)
   - 66% times a success will happen (success notification, and modal closes)


biome.jsonc
```diff
@@ -9,7 +9,8 @@
       "*-min.js",
       "bootstrap*.js",
       "jquery*.js",
-      "kobo/apps/openrosa/apps/logger/fixtures/*"
+      "kobo/apps/openrosa/apps/logger/fixtures/*",
+      "kobo/apps/organizations/static/*"
```
duvld, line 13: Where did this change come from? It doesn't seem related to the tasks unless I'm mistaken 🤞

jsapp/js/attachments/AttachmentActionsDropdown.tsx
```diff
@@ -0,0 +1,143 @@
...
+ * provided attachment. Delete option would display a safety check modal.
+ */
+export default function AttachmentActionsDropdown(props: AttachmentActionsDropdownProps) {
+  const [isDeleteModalOpen, setIsDeleteModalOpen] = useState<boolean>(false)
+  const [isDeletePending, setIsDeletePending] = useState<boolean>(false)
+  const removeAttachmentMutation = useRemoveAttachment(props.asset.uid, props.submissionData['meta/rootUuid'])
+  const isFeatureEnabled = useFeatureFlag(FeatureFlag.removingAttachmentsEnabled)
+
+  // TODO: remove this when feature is ready. For now we hide the whole thing by not rendering anything.
+  if (!isFeatureEnabled) {
```
duvld, line 36: A bit of confusion, in the [spec](https://www.notion.so/kobotoolbox/Removing-attachments-8e2eb538680e427fa14173ec523c58e2?p=cb1e69c1f6ec4dd7b4586c7a1c45b5e7&pm=s) it's written that
> if feature flag is not enabled, this dropdown will only have “Download” option

but here we don't render the entire thing. It does say that for if the attachment is deleted then this should be the case.

jsapp/js/attachments/AttachmentActionsDropdown.tsx
```diff
@@ -0,0 +1,143 @@
...
+    return null
+  }
+
+  const attachment = props.submissionData._attachments.find((a) => a.id === props.attachmentId)
```
duvld, line 40: Personally I don't like single letter variables, I think we can say `element` or something like that?

jsapp/js/attachments/AttachmentActionsDropdown.tsx
```diff
@@ -0,0 +1,143 @@
...
+  // Safety check, ideally parent component should not render this component if attachment is deleted.
+  if (attachment.is_deleted) {
+    return null
+  }
+
+  async function handleConfirmDelete() {
+    setIsDeletePending(true)
+
+    try {
+      // We use `!` to assert that the attachment exists, because this code would be unreachable without it.
```
duvld, line 54: Curious why the check on line 41 is not sufficient?

jsapp/js/attachments/AttachmentActionsDropdown.tsx
```diff
@@ -0,0 +1,143 @@
...
+        <Stack>
+          <p>
+            {t('Are you sure you want to delete this ##attachment_type##?').replace(
+              '##attachment_type##',
+              attachmentTypeName,
+            )}
+          </p>
+
+          <Group justify='flex-end'>
+            <Button variant='light' size='lg' onClick={() => setIsDeleteModalOpen(false)} disabled={isDeletePending}>
```
duvld, line 125: It looks like [they specified the delete to be invisible](https://www.notion.so/kobotoolbox/Removing-attachments-8e2eb538680e427fa14173ec523c58e2?pvs=4#592a78e7a68344228b216dc028db6ce8) if the user does not have enough permissions (not disabled)

jsapp/js/dataInterface.ts
```diff
@@ -196,24 +198,26 @@ export interface SubmissionResponse {
...
   }
-  _version_: string
+  _version_?: string
   _xform_id_string: string
   deviceid?: string
   end?: string
+  // `meta/rootUuid` is persistent across edits while `_uuid` is not;
+  // use the persistent identifier if present.
+  _id: number
+  _uuid: string
```
duvld, line 220: Not sure how these relate to the PR either, unless I'm missing something here too 🙇‍♂️

jsapp/js/dataInterface.ts
```diff
@@ -632,7 +636,14 @@ export interface AssetResponse extends AssetRequestObject {
   }
   deployment__active: boolean
   deployment__data_download_links?: {
-    [key in ExportTypeName]: string | undefined
+    csv_legacy: string
```
duvld, line 639: Same as above

duvld (commented):
Few questions

biome.jsonc
magicznyleszek: Biome was complaining about these files, making the PR fail. Kalvis said to do that change :)

magicznyleszek: Acutally Kalvis said something otherwise after some time :D https://chat.kobotoolbox.org/#narrow/stream/4-Kobo-Dev/topic/Biome.20issues/near/570825

I will undo this change

jsapp/js/components/submissions/submissionDataTable.tsx
```diff
@@ -235,6 +236,29 @@ class SubmissionDataTable extends React.Component<SubmissionDataTableProps> {
               {filename}
             </a>
           )}
+
+          {type !== null && (
+            <AttachmentActionsDropdown
```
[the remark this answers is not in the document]
magicznyleszek: Fixed this :)

jsapp/js/attachments/AttachmentActionsDropdown.tsx
```diff
@@ -0,0 +1,143 @@
...
+              {t('Cancel')}
+            </Button>
+
+            <Button
+              variant='danger'
```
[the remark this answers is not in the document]
magicznyleszek: [I've asked Tessa about this](https://chat.kobotoolbox.org/#narrow/stream/7-Kobo-UX.2FUI.2FProduct/topic/Mantine.20Styling.20Questions/near/571510). For the Members Table the menu looks a bit different, and I'm guessing we want to be consistent

jsapp/js/attachments/AttachmentActionsDropdown.tsx
magicznyleszek: Right! Fixed this. Now only "Delete" button is hidden behind feature flag

jsapp/js/attachments/AttachmentActionsDropdown.tsx
magicznyleszek: My guess is that it's because this line here is inside a function, so different scope. I thought it wouldn't be needed, but added this after TS complained :)

jsapp/js/attachments/AttachmentActionsDropdown.tsx
magicznyleszek: Fixed :)

jsapp/js/dataInterface.ts
magicznyleszek: After I've added the mock file (content of the file is real server responses), it turned out that not all of `ExportTypeName`s are always present, so I changed the interface to be more precise in that

jsapp/js/attachments/AttachmentActionsDropdown.tsx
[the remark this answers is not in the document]
magicznyleszek: Ok for now I just did the same thing as we did for the Members Table dropdow (the invitations one). I would say it's fine for now, if we are going to change this one, we will have to change the invitations one too :shrug:

duvld (approved):
LGTM

jsapp/js/attachments/AttachmentActionsDropdown.tsx
Akuukis: See https://github.com/kobotoolbox/kpi/pull/5597

jsapp/js/attachments/AttachmentActionsDropdown.tsx
Akuukis: See https://github.com/kobotoolbox/kpi/pull/5597

PR #5582, opened 2025-03-12
fix(abstractTimestampedModel): restore `date_modified` saving behavior

noliveleger:
### 📣 Summary
Restored the intended behavior of automatically saving modification timestamps.

### 💭 Notes
When `date_modified` needs to be updated, either call `.save()` without the `update_fields` parameter or explicitly include `date_modified` in `update_fields`. Otherwise, the modification timestamp will not be updated.

### 👀 Preview steps

1. From the shell, retrieve an object which inherit from `AbstractTimestampedModel`
2. Try to update only one field of the object with `.save(update_fields=['field_name'])`
3. On release branch, `date_modified` has changed
4. On this PR, `date_modified` has not

### DIFF
--- a/kobo/apps/accounts/mfa/models.py
+++ b/kobo/apps/accounts/mfa/models.py
@@ -60,7 +60,12 @@ class MfaMethod(TrenchMFAMethod, AbstractTimeStampedModel):
         if update_fields:
             update_fields += ['date_disabled']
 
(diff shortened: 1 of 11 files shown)


rajpatel24 (approved):
LGTM

jnm (approved):

PR #5591, opened 2025-03-13
feat(attachments): add a simple component for a deleted attachment TASK-568

duvld:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Adds a text component to display when an attachment has been deleted from a submission.

### 👀 Preview steps

Check the storybook or make it display in the data table.

To make it display on the data table:
1. Ensure your form has media questions and submissions
1. Open either `mediaCell.tsx` or `audioCell.tsx` (whichever is your media type)
2. Return `<DeletedAttachment />` instead of what is already there
3. Notice the *Deleted* in place of the media cell


magicznyleszek (approved):
I think this is all fine. I have some ideas on adding some predefined styles (`align-self: center`) to make it centered automatically in some situations, but I think we will see if anything is needed when we use it in all the places.

PR #5593, opened 2025-03-13
ci: run ci steps locally

Akuukis:
### 💭 Notes

Run CI steps locally with a `./scripts/ci_locally.sh` bash script.


noliveleger:
@Akuukis , the script is running as expected (in the container) locally. 
FYI: I made some changes with pytest to make it run faster.

Otherwise, LGTM

scripts/ci_locally.sh
```diff
@@ -0,0 +1,84 @@
...
+
+
+echo -e '\n\n## Job: Darker'
+
+echo -e '\n\n### Step: Set up Python'
+echo 'Disclaimer: CI installs Python 3.10. Please eyeball if your local env has 3.10.'
+python --version
+
+echo -e '\n\n### Step: Install pip dependencies'
+python -m pip install darker[isort] flake8 flake8-quotes isort --quiet
```
jnm, line 44: Is this a fail-fast optimization for CI? If all of `dev_requirements.txt` ends up being installed anyway, maybe we could avoid and just run `darker` later?

scripts/ci_locally.sh
```diff
@@ -0,0 +1,84 @@
...
+
+echo -e '\n\n### Step: Run Darker'
+echo 'Disclaimer: CI runs Darker only on the last commit.. I think?'
```
jnm, line 47: Maybe I'm understanding wrongly, but a discrepancy like this sounds like it'd lead to vastly different output? cc @noliveleger

scripts/ci_locally.sh
```diff
@@ -0,0 +1,84 @@
...
+python --version
+
+echo -e '\n\n### Step: Install pip-tools'
+python -m pip install pip-tools==7.\*
+
+echo -e '\n\n### Step: Update Debian package lists'
+DEBIAN_FRONTEND=noninteractive apt-get -qq -y update
+
+echo -e '\n\n### Step: Install Debian dependencies'
+DEBIAN_FRONTEND=noninteractive apt-get -qq -y install gdal-bin gettext libproj-dev postgresql-client ffmpeg gcc libc-dev build-essential
```
jnm, line 64: Unless the intent is to run this in a vanilla container totally distinct from what'd be used for normal development, these should be omitted. My understanding was this script should serve as a quick way for developers to check the outcome of a merge before pushing to GitHub—and so it should be tailored to work best in the environment where a developer would already be when doing such a merge. That, and achieving the same total format/lint/test coverage, is more important than following the CI bit by bit.

scripts/ci_locally.sh
```diff
@@ -0,0 +1,84 @@
...
+git submodule init && git submodule update --remote && python manage.py compilemessages
+
+echo -e '\n\n### Step: Test back-end code'
+echo 'Disclaimer: CI uses pytest with coverage option, skipping locally.'
+# Speed-up pytest by running tests in parallel
+pytest -q --disable-warnings -n auto || true
+# Run only the last failed tests sequentially.
+# Tests that previously failed due to concurrency should pass now.
+# If they still fail, it indicates genuine test failures.
+pytest -q --disable-warnings --lf -rf
```
jnm, line 79: Ha! Well, one improvement at a time :innocent:

scripts/ci_locally.sh
```diff
@@ -0,0 +1,84 @@
...
+SKIP_TS_CHECK=true npm run build
+
+echo -e '\n\n### Step: Check ESLint, errors only'
+npm run lint:eslint -- --quiet
+
+echo -e '\n\n### Step: Run unit tests and xlform tests'
+npx jest --config ./jsapp/jest/unit.config.ts --ci
+
+echo -e '\n\n### Step: Run component tests with Jest'
+npx jest --config ./jsapp/jest/jest.config.ts --ci
```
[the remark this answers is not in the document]
jnm: @noliveleger I'd be curious about the same for the back-end section: is there one script that both developers (via this do-it-all script) and the CI could call to accomplish all the formatting/linting/testing?

jnm (changes requested):
My points are for discussion; curious to read feedback. Thanks

scripts/ci_locally.sh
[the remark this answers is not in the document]
Akuukis: Sounds good, added for FE. Leaving this unresolved for @noliveleger for BE.

scripts/ci_locally.sh
```diff
@@ -0,0 +1,84 @@
...
+
+
+echo -e '\n\n# Run CI jobs locally'
+echo 'Disclaimer: local environment may differ from Github Actions environment.'
+
+
+echo -e '\n\n### Step: Setup Node'
+echo 'Disclaimer: CI runs matrix on v20.17.0 and 22. Please eyeball if your local version matches.'
+node -v
+echo 'Disclaimer: CI caches node_modules, skipping locally.'
```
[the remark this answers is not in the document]
Akuukis: fixed

scripts/ci_locally.sh
```diff
@@ -0,0 +1,84 @@
...
+set +x
+
+
+echo -e '\n\n# Run CI jobs locally'
+echo 'Disclaimer: local environment may differ from Github Actions environment.'
+
+
+echo -e '\n\n### Step: Setup Node'
+echo 'Disclaimer: CI runs matrix on v20.17.0 and 22. Please eyeball if your local version matches.'
+node -v
```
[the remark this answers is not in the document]
Akuukis: good idea, added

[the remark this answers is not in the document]
Akuukis: @jnm CI fails now because it matrix over both v20 and v22. Shall I add if-else for both in hints?

jnm:
applied #donotmerge; needs discussion with @p2edwards

Akuukis:
Thanks, I figured out now that's `./run.py -cf exec kpi pytest` while all of kobo-install containers are up in the background

scripts/ci_locally.sh
noliveleger: As per discussed, I removed all python dependencies installation. Everything should be already installed in the container of a dev environment.

scripts/ci_locally.sh
noliveleger: Darker compares the last commit with the working tree. 
I changed it to `-r origin/<current_branch>`

scripts/ci_locally.sh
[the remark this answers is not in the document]
noliveleger: As per discussed, to complicated to implement and keep optimization for CI.

scripts/ci_locally.sh
noliveleger: As per discussed, apt dependencies installation were removed.

jnm (approved):
Works for me successfully in my local environment. Thanks

Akuukis:
LGTM & works for me locally (Ubuntu 24.04). Pushed doc update with a copy/paste-able snippet on how to run it: 
```
cd kobo-install
./run.py
./run.py -cf exec kpi ./scripts/ci_locally.sh
```

PR #5594, opened 2025-03-13
ci: format github action alert title

Akuukis:
### 💭 Notes

Format internal alerts. Change the full commit message to only title of it.

### DIFF
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -28,8 +28,6 @@ jobs:
       id: title
       env:
         MESSAGE: ${{ github.event.commits[0].message }}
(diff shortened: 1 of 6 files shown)


jnm (changes requested):
Could you add a temporary commit that:
1. Makes these workflows run on all branches
2. Intentionally breaks something

…so that we can see an example run without having to merge untested stuff to `main` and `release/` branches? If everything's good, we can drop the temporary commit.

Also, as a quick fix if you wouldn't mind, please disambiguate the `name:`s for these to avoid this:
image

Akuukis:
@jnm both done :ok_hand:

jnm (approved):

PR #5596, opened 2025-03-14
refactor(attachment): add delete status to attachments to (pre) support manual deletion TASK-1529

noliveleger:
### 📣 Summary
Introduced a delete status for attachments to prepare for manual deletion support.

### 📖 Description
A new delete status has been added to attachments, allowing the system to track attachments marked for deletion without immediately removing them. This refactor is a preparatory step for supporting manual attachment deletion in the future.


rajpatel24 (approved):
LGTM

PR #5599, opened 2025-03-17
feat(attachments): add download and delete buttons TASK-571

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 💭 Notes

We don't touch the Project > Data > Gallery in MVP, so this would be the only place that we don't give option to delete attachments from.

Changes here:
- `AttachmentActionsDropdown` now displays "Delete" button in red color (Tessa's request)
- `AttachmentActionsDropdown` uses `<a>` for "Download" button for better UX
- `AudioPlayer` has option to render things in "right section"
- `SidebarSubmissionMedia` uses `AttachmentActionsDropdown`
- `SidebarSubmissionMedia` doesn't handle "video" question types anymore (we weren't using this option)
- `MediaCell` (and `TableMediaPreview`) now handles "file" question types
- `MediaCell` now displays filename next to the icon
   - Reasoning: previously `file` was displaying filename there, and it is definitely useful to see it. I decided to change this for the other types as well, as it is better to see some different text next to icon rather than the identical icon for all submissions
- `MediaCell` uses `AttachmentActionsDropdown`
- Created `markAttachmentAsDeleted` utility function that adds `is_deleted` flag to given attachment in a submission data object
- `SubmissionDataTable` now reacts to attachment being deleted, and causes two things:
   1. memoized submission data in `SubmissionModal` is being updated (using `markAttachmentAsDeleted`) 
   2. `DataTable` is force refreshing submissions, so that it renders the deleted attachment (this will be visible once `is_deleted` flag is being handled)
- Fixed `DataTable` listener for `actions.resources.refreshTableSubmissions`, as we broke it recently (probably I did it when migrating to TS)
- Fixed `TableBulkCheckbox` default export (it worked previously, but now Biome was complaining)

### 👀 Preview steps

Case A: testing steps for Submission Modal:
1. ℹ️ Have a project with `audio`, `file`, `image`, and `video` question
2. Make a submission that has responses to all the questions
   - go to Project > Form and use "Open" button in the "Collect data" section
3. Go to Project > Data > Table
4. Open submission ("eye" icon button)
5. 🟢 Notice that each question response has now "…" button on the right side
   - Using "Download" button (link) should work for each of question type
   - Using "Delete" button should do this for each question type:
      1. A 33% chance of action failing (mock) would cause error notification to appear
      2. A 67% chance of action suceeding (also mock) would cause success notification to appear, "…" modal to be closed, and memoized attachment data in Submission Modal getting `is_deleted` flag
6. 🟢 Notice that when attachment `is_deleted` the "…" button is not rendered
   - Note that the attachment would still appear, as the work on displaying `DeletedAttachment` component is still in progress

Case B: testing steps for Table Media Preview modal (`file`, `image`, and `video` questions):
1. Go through steps 1-3 of Case A
2. 🟢 Notice in the table that cell now has a file name and an icon button (previously it was just the file name for `file` and just icon for the other two)
3. Open a modal (icon button in the cell)
4. 🟢 Notice in the modal that we now have "…" button in top right instead of "Download" button
   - Using "Download" button (link) should work
   - Using "Delete" button should do this for each question type:
      1. A 33% chance of action failing (mock) would cause error notification to appear
      2. A 67% chance of action suceeding (also mock) would cause success notification to appear, "…" modal to be closed, and Data Table submissions to be refetched
5. 🟢 Notice that for `file` type the file name and mimetype are being displayed as modal content

Case C: testing steps for NLP View:
1. Go through steps 1-3 of Case A
2. For `audio` question response use "Open" button
4. 🟢 Notice that in the right side the audio player now has "…" button
   - Using "Download" button (link) should work
   - Using "Delete" button should do this:
      1. A 33% chance of action failing (mock) would cause error notification to appear
      2. A 67% chance of action suceeding (also mock) would cause success notification to appear, "…" modal to be closed, and submission data to be refetched


jsapp/js/components/processing/sidebar/sidebarSubmissionMedia.tsx
```diff
@@ -36,19 +36,27 @@ export default function SidebarSubmissionMedia(props: SidebarSubmissionMediaProp
...
+            mediaURL={attachment.download_url}
+            filename={attachment.filename}
+            rightHeaderSection={
+              store.data.submissionData && (
+                <AttachmentActionsDropdown
+                  asset={props.asset}
+                  submissionData={store.data.submissionData}
+                  attachmentId={attachment.id}
+                  onDeleted={() => {
+                    // TODO: this might be done with a bit more elegant UX, as calling the function causes a whole page
```
duvld, line 49: :pray: comment answered most of the questions I had. I agree with the best approach being to make the store use react-query. Maybe we should create a task for this in post mvp?

jsapp/js/attachments/AttachmentActionsDropdown.tsx
```diff
@@ -88,13 +88,23 @@ export default function AttachmentActionsDropdown(props: AttachmentActionsDropdo
         </Menu.Target>
 
         <Menu.Dropdown>
-          <Menu.Item onClick={requestDownloadFile} leftSection={<Icon name='download' />}>
+          {/* TODO: once we get this button to be able to "save as", we can remove `target` */}
```
duvld, line 91: I'm not sure if this UX is actually an improvement. I guess opening the media in a new tab where the user can "Save as" themselves is more in control than forcing a download into the default location. Either case I think "save as" is the definite best UX--do you remember what is stopping us from implementing a download link to save as?

jsapp/js/components/processing/sidebar/sidebarSubmissionMedia.tsx
duvld: BTW, would we encounter strange behaviour if we remove the audio attachment from the NLP view? That's the only way we can get to through the UI IIRC. It would definitely show a blank box after rerender so we should set it to disable the audio in the sidebar (and disable the switch in settings).

We should also explain the consequences of deleting the audio from the NLP screen in the delete modal

jsapp/js/components/submissions/submissionUtils.ts
```diff
@@ -698,3 +699,27 @@ export function getQuestionXPath(surveyRows: SurveyRow[], rowName: string) {
...
+/**
+ * In given submission data, it finds provided attachment, sets its `is_deleted`
+ * flag to `true` and then returns the updated submission data.
+ */
+export function markAttachmentAsDeleted(
+  submissionData: SubmissionResponse,
+  targetAttachmentId: number,
+): SubmissionResponse {
+  const data = clonedeep(submissionData)
+  const targetAttachment = data._attachments.find((a) => a.id === targetAttachmentId)
```
duvld, line 712: `item` instead of `a` pls :)

jsapp/js/components/submissions/tableBulkCheckbox.tsx
```diff
@@ -16,7 +16,7 @@ interface TableBulkCheckboxProps {
   onClearSelection: () => void
 }
 
-export function TableBulkCheckbox(props: TableBulkCheckboxProps) {
+export default function TableBulkCheckbox(props: TableBulkCheckboxProps) {
```
duvld, line 19: Was this a biome/linting change or does this affect something down the line

duvld (commented):
Few questions

jsapp/js/components/submissions/tableBulkCheckbox.tsx
magicznyleszek: It is a biome change :)

jsapp/js/components/processing/sidebar/sidebarSubmissionMedia.tsx
magicznyleszek: I think that deleting audio attachment in the context of NLP should only influence the sidebar player. So instead of player, we should display `DeletedAttachment` in there. So there should be some logic inside `SidebarSubmissionMedia` for it. We have to remember that user can have transcript and translations etc. and deleting audio shouldn't block the access to all that :) I imagine some users would delete attachment immediately after getting transcript

magicznyleszek: As for creating a task, I will ask Jackie where she would like us to put them :)

jsapp/js/attachments/AttachmentActionsDropdown.tsx
magicznyleszek: Blame leads to https://github.com/kobotoolbox/kpi/pull/3418/files#diff-d4c1908e5ae1cb5dc07c0d5d6cd971b8c9a713221c53fd579b57e667f47aafdfR108 but it doesn't answer the question unfortunately.

But I have tested "save as" right now, and turns out it actually works :D I will remove the comment…

jsapp/js/components/submissions/submissionUtils.ts
magicznyleszek: Fixed :bow:

duvld (approved):

PR #5610, opened 2025-03-24
refactor(massEmails): refactor usage queries to make nlp easier

rgraber:
### 🗒️ Checklist

1. [ ] run linter locally

### 👀 Preview steps

Bug template:
2. do this
3. do that

Feature/no-change template:
2. do this
3. do that
5. do that another thing

### DIFF
--- a/kobo/apps/stripe/utils.py
+++ b/kobo/apps/stripe/utils.py
@@ -35,6 +35,11 @@ def get_default_add_on_limits():
 def get_active_subscription_billing_dates_by_org(
     organizations: list[Organization] = None,
 ):
+    """
+    Retrieve the current billing cycle for provided organizations. Does not include
+    the billing cycle for storage addons, as they are not measured monthly. If no
+    list of organizations is provided, retrieves details for all organizations.
+    """
     now = timezone.now().replace(tzinfo=ZoneInfo('UTC'))
     first_of_this_month = datetime(now.year, now.month, 1, tzinfo=ZoneInfo('UTC'))
     first_of_next_month = first_of_this_month + relativedelta(months=1)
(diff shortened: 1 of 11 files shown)


kobo/apps/stripe/utils.py
```diff
@@ -247,59 +178,329 @@ def get_billing_dates_after_canceled_subscription(
...
+
+def get_default_add_on_limits():
+    return {
+        'submission_limit': 0,
+        'asr_seconds_limit': 0,
+        'mt_characters_limit': 0,
+    }
+
+
+def get_organization_limit(
```
jamesrkiger, line 254: I'm wondering why we switched to calling this `get_organization_limit` from `get_organization_plan_limits`. I think the modifier 'plan' (or something equivalent) might be worth including to make sure we're clear that these numbers represent only the subscription-based, recurring portion of an organization's usage limits (as opposed to their effective limits, which would need to include one-time addons for NLP usage ... and someday submissions). I know the current work isn't handling one-time addons, but that seems an important distinction to make at least in name.

rgraber: I was going to include one-time addons in the next pr.

rgraber: But I see your point. I don't know when the next PR will be done so shouldn't leave a confusing name in the code.

rgraber: How do you feel about 'get_organization_subscription_limit' since I'm using subscription to mean plan + addon?

Guitlle (approved):
LGTM

rajpatel24 (approved):
Please fix the failing check before merging. I don't think it's a major change. Everything else looks good to me

PR #5613, opened 2025-03-25
perf(attachments): backfill attachments in chunks to optimize queries TASK-1534

RuthShryock:


kobo/apps/long_running_migrations/jobs/0007_backfill_attachment_model.py
```diff
@@ -1,61 +1,74 @@
...
+                fields=[
+                    'xform_id',
+                    'user_id',
+                    'uid',
+                    'date_created',
+                    'date_modified',
+                    'delete_status',
+                ],
+            )
+            attachment_pk = attachment.pk
```
noliveleger, line 53: Even it is going to be set multiple time nothing, let's move `attachment_pk = attachment.pk` inside the loop , right below `attachment.user_id = instance.xform.user_id`

kobo/apps/long_running_migrations/jobs/0007_backfill_attachment_model.py
```diff
@@ -1,61 +1,74 @@
...
-                'uid',
-                'date_created',
-                'date_modified',
-                'delete_status',
-            ],
-        )
+        .filter(
+            Q(date_created__isnull=True) | Q(date_modified__isnull=True),
+            pk__gt=attachment_pk,
+        )[:CHUNK_SIZE]
```
noliveleger, line 73: You need to `order_by('pk')` to ensure we are always fetching the correct subset of data. Right now it is random, nothing guarantee the subset will contain `CHUNK_SIZE` elements.

kobo/apps/long_running_migrations/jobs/0007_backfill_attachment_model.py
```diff
@@ -1,61 +1,74 @@
...
+                    'date_modified',
+                    'delete_status',
+                ],
+            )
+            attachment_pk = attachment.pk
+        else:
+            stop = True
+
+
+def get_queryset(attachment_pk):
```
noliveleger, line 58: Even it's obvious, let's use some annotation typings ;-)

noliveleger (changes requested):

kobo/apps/long_running_migrations/jobs/0007_backfill_attachment_model.py
```diff
@@ -1,61 +1,77 @@
...
+
+
+def get_queryset(attachment_pk: int) -> QuerySet[Attachment]:
+    return (
+        Attachment.all_objects.select_related('instance__xform')
+        .only(
+            'id',
             'deleted_at',
             'date_created',
-            'instance__date_created',
```
noliveleger, line 23: I think we need this one too. Otherwise, the field will be deferred and Django will have to make an extra query each time to retrieve it.

noliveleger (changes requested):

PR #5617, opened 2025-03-25
fix(longRunningMigration): fix crash in long-running migration when multiple submissions share the same UUID

noliveleger:
### 📣 Summary
Fixed a bug that caused the root UUID long-running migration to crash when duplicate submission UUIDs were encountered.

### 📖 Description
The long-running migration responsible for backfilling the root_uuid field crashed when two or more submissions within the same project shared the same uuid. This fix ensures that the migration can handle such cases safely without interrupting the overall process.

### 👀 Preview steps

Bug template:
2. submit data (at least 2 submissions)
3. update the XML, `uuid` of 2 submissions, set their `root_uuid` to null (directly in the database)
4. reset long running migration 0005 to "created"
5. from the shell run `execute_long_running_migrations()` 
4. 🔴 [on main] long running migration 0005 crashes
5. 🟢 [on PR] long running migration 0005 runs and fields are updated


rajpatel24 (approved):
LGTM

PR #5618, opened 2025-03-26
fix: add activity log implementation into release branch 2.025.02 TASK-1714

pauloamorimbr:
### 🗒️ Checklist

1. [x] run linter locally

### 📖 Description
We need to add the implementation code for the activity logs feature existing in main into the release branch 2.025.02.
Front end and back end code will be added via this PR
The release branch was created at a point in time where only half of the implementation was done and now we have a bug in the release branch because of the missing half and inconsistency between front and backend codes.

### 👀 Preview steps

Bug template:
2. Do action in the project
3. Check the Activity panel inside the project settings
4. 🔴 [on the release branch] note that the filter loading empty and the export button is not present
5. 🟢 [on PR] note that the filter options are loading correctly and the export button is present


noliveleger (approved):
LGTM

PR #5623, opened 2025-03-26
fix(organizations): projects of users with no previous organization were not transferred to a new one TASK-1716

noliveleger:
### 📣 Summary
Fixed an issue where projects belonging to users with no prior organization were not transferred when joining a new organization (when users were added from Django Admin).

### 📖 Description
Users who did not belong to any organizations were not having their projects transferred to the new organization upon joining. This was due to a race condition in the ownership transfer logic, which failed to handle users without any previous org affiliation.

### 👀 Preview steps

Bug template:
2. Create a MMO
3. Log into a regular user account and create project, deploy it
4. Log out
3. From the shell, remove the organization of a user 
4. Add this user to the MMO from Django Admin
4. 🔴 [on release branch] notice that the user is added but their projects still belong to them
5. 🟢 [on PR] notice that their projects are transferred to the new org

### DIFF
--- a/kobo/apps/organizations/admin/organization.py
+++ b/kobo/apps/organizations/admin/organization.py
@@ -1,6 +1,6 @@
 from django.conf import settings
 from django.contrib import admin, messages
-from django.db.models import Count
+from django.db import transaction
 from django.urls import reverse
 from django.utils.safestring import mark_safe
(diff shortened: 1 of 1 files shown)


rajpatel24 (approved):
LGTM

PR #5625, opened 2025-03-26
fix(NLP): correct paths for grouped questions

jnm:
### 📣 Summary
Resolves a problem where transcripts, translations, and qualitative analysis responses were missing from the data table view and single submission modal for grouped questions

Assumptions in the construction of `_supplementalDetails` paths relied on the old dash-delimited `qpath`s and broke for slash-delimited XPaths. See also #5616 for similar backend fix; internal discussion: https://chat.kobotoolbox.org/#narrow/stream/6-Kobo-Support/topic/Qualitative.20Analysis.20Feature/near/578393.

### 👀 Preview steps
1. Create a survey with an audio question inside a group
2. Collect an audio response
3. Add a transcript, translation, and qualitative analysis question + response
4. See the NLP & QA data saved properly and visible in the single processing view
5. 🔴 [on release branch] Notice that none of the NLP or QA data appear in the data table or single submission modal
6. 🟢 [on PR] All data appears correctly

### DIFF
--- a/jsapp/js/components/submissions/submissionUtils.ts
+++ b/jsapp/js/components/submissions/submissionUtils.ts
@@ -14,6 +14,7 @@ import {
   GROUP_TYPES_BEGIN,
   QUESTION_TYPES,
   CHOICE_LISTS,
+  SUPPLEMENTAL_DETAILS_PROP,
 } from 'js/constants';
 import type {AnyRowTypeName} from 'js/constants';
 import type {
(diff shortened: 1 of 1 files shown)


magicznyleszek (approved):
Everything works now. Interestingly we have unit tests for `getSupplementalDetailsContent` and they didn't caught all that :(

jnm:
Added https://linear.app/kobotoolbox/issue/DEV-34/tests-for-getsupplementaldetailscontent-are-not-sufficient for unit test improvement

PR #5635, opened 2025-03-28
refactor(trashBin): standardize status toggling via abstract base method TASK-1525

noliveleger:
### 📣 Summary
Refactored `toggle_<model>_statuses` into an abstract method on the base trash bin class and prepares the codebase for managing attachments in the trash bin in future work.

### 📖 Description
As part of upcoming work to support attachments in the trash bin system, the `toggle_<model>_statuses` method has been moved to the base class as an abstract method. This change standardizes the implementation pattern for all models handled by the trash bin system.


noliveleger:
merged since #5634 has been reviewed  and merged and contains changes related to this PR.

PR #5640, opened 2025-04-01
feat(massEmails): create celery task for generating daily user lists TASK-1690

rajpatel24:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Added a new task to generate and enqueue `MassEmailRecords` while tracking processed configurations using caching.

### 📖 Description
This update introduces a task that processes active MassEmailConfig entries, gathering all configs with `date_created < today` and `live==True`, ensuring emails are only sent to eligible users based on predefined rules. The task:

- Caches processed email configs to avoid duplicate processing.
- If a cache object for today's emails doesn't already exist, it creates a new cache object `<todays_date>_emails` as an empty set, representing emails for which today's user list has already been generated.
- Skips configs with existing enqueued or sent records.
- Filters out users who have received the same email within the configured frequency.
- Handles duplicate entry errors to ensure robustness.

This prevents redundant emails and improves the reliability of the email queuing system.

### ⚠️ Note

Scheduling this Celery task has been intentionally postponed as per project requirements. This ensures it does not interfere with existing email-sending tasks. The implementation focuses solely on generating and enqueuing email records, with scheduling to be handled in a future update.

### 👀 Preview steps

1. Set `inactivity_threshold = now() - timedelta(minutes=1)` for the inactive user query.
2. Create a config with `date_created < today`, `live=True`, `frequency=-1` and existing records with status `ENQUEUED` or `SENT`. Run `generate_mass_email_user_lists()` in the shell and verify that no new records are created.
3. Create a config with `date_created < today`, `live=True` and existing records with status `ENQUEUED`. Run the function and verify that no new records are created.
4. Create a config with no records. Run the function and verify that new records are created.


rgraber (changes requested):
It looks like this PR no longer deals with frequency = -1 ?

kobo/apps/mass_emails/tasks.py
```diff
@@ -210,3 +211,79 @@ def send_email(self, email_config, record):
...
+    frequency = 1: Daily emails
+    frequency > 1: Recurring emails
+    """
+    users = USER_QUERIES.get(email_config.query, lambda: [])()
+    today_midnight = now().replace(hour=0, minute=0, second=0, microsecond=0)
+    cutoff_date = today_midnight - timedelta(days=email_config.frequency-1)
+
+    # For daily emails, cutoff date is today midnight
+    if email_config.frequency == 1:
+        cutoff_date = today_midnight
```
rgraber, line 230: i don't think we need the extra calculation. if the frequency is 1, the timedelta will be 0

rajpatel24: Ah, I missed updating the condition for `frequency=1`. You're right, we don't need the extra calculation since the timedelta will be 0

rajpatel24:
> It looks like this PR no longer deals with frequency = -1 ?

The PR still handles `frequency = -1`. We're adding the config to the cache when `frequency = -1` and there are records with status `enqueued` or `sent`. If no such records exist, new ones will be created. I have updated the tests to verify this behavior more thoroughly

kobo/apps/mass_emails/tasks.py
```diff
@@ -210,3 +211,75 @@ def send_email(self, email_config, record):
...
+        if email_config.id in processed_configs:
+            continue
+
+        enqueued_records = MassEmailRecord.objects.filter(
+            email_job__email_config=email_config,
+            status=EmailStatus.ENQUEUED
+        )
+
+        # Skip processing for one time emails that have been sent or enqueued
+        if email_config.frequency == -1 and enqueued_records.filter(
```
rgraber, line 265: This still misses sent records because `enqueued_records` has already been filtered down to things with `status=enqueued`, so it will never find any with `status=sent`. 
Verified this locally.

kobo/apps/mass_emails/tests/test_celery_tasks.py
```diff
@@ -124,3 +133,208 @@ def test_send_is_throttled(self):
...
+            date_modified=now() - timedelta(days=days_ago)
+        )
+
+    def test_one_time_email_handling(self):
+        """
+        Test one-time email configs (frequency=-1) behave correctly
+        """
+        # Case 1: Existing records should prevent new emails from being generated
+        one_time_config = self._create_email_config('ExistingRecordsTest')
+        self._create_email_record(self.user1, one_time_config, EmailStatus.ENQUEUED)
```
rgraber, line 189: if you remove this line the test will fail

kobo/apps/mass_emails/tasks.py
rajpatel24: Ah, got it! Thanks for pointing that out

Guitlle (changes requested):

kobo/apps/mass_emails/tasks.py
```diff
@@ -210,3 +211,77 @@ def send_email(self, email_config, record):
...
+    """
+    Generates daily user lists for MassEmailConfigs, skipping already processed
+    configs and users
+    """
+    # ToDo: Scheduling the Celery task for this implementation is postponed for
+    #       future development, as outlined in the project requirements. It has
+    #       been intentionally left out to avoid interference with the existing
+    #       email sending tasks.
+
+    today = now().date()
```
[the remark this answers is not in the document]
rajpatel24: Shouldn't it be `now().date()` for consistency, especially for generating the cache key? And since the `TIME_ZONE` is set to UTC, wouldn't both `now().date()` and `localdate()` return the same UTC date in this case?

rgraber (changes requested):
Per @noliveleger 's request, let's use `from django.utils import timezone` instead of `from django.utils.timezone import now`
Also request for a log statement
Other than that LGTM

rajpatel24:
@rgraber I agree with your point about timezone-aware dates. Using `timezone.now().date()` will help ensure consistency across the application. Let's await @Guitlle  input on this

kobo/apps/mass_emails/tasks.py
```diff
@@ -210,3 +212,84 @@ def send_email(self, email_config, record):
...
+    Generates daily user lists for MassEmailConfigs, skipping already processed
+    configs and users
+    """
+    # ToDo: Scheduling the Celery task for this implementation is postponed for
+    #       future development, as outlined in the project requirements. It has
+    #       been intentionally left out to avoid interference with the existing
+    #       email sending tasks.
+
+    today = timezone.now().date()
+    cache_key = f'{today}_emails'
```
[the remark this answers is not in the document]
rajpatel24: I agree with you. Updated the key to `mass_emails_{today}_emails`

rgraber (approved):
LGTM

sksondra (commented):
Good from QA.

PR #5641, opened 2025-04-01
fix(massEmails): throttle mass email send

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Limit the rate of emails sent per second to avoid hitting SES limits.

### 👀 Preview steps

1. ℹ️ have at least 2 users
2. In your settings, set `MASS_EMAIL_THROTTLE_PER_SECOND` to 1
3. Update the `get_inactive_users` query to use a timedelta of 0
4. Create a MassEmailConfig that uses the inactive users query
5. Clear the redis cache and reschedule the the mass-email cron as necessary so you don't have to wait for the send. `redis-cli KEYS "*mass_emails*" | xargs redis-cli DEL` will clear the relevant keys  when run in the redis container
6. Open the kpi_worker logs
7. Add the new config to the daily send
8. 🟢 Monitor the sends in the kpi worker logs. There should be about 5 seconds between `Processing MassEmailRecord(id)` logs

### DIFF
--- a/kobo/apps/mass_emails/tasks.py
+++ b/kobo/apps/mass_emails/tasks.py
@@ -176,7 +176,6 @@ class MassEmailSender:
                 self.send_email(email_config, record)
                 emails_sent += 1
 
-
     def send_email(self, email_config, record):
         logging.info(f'Processing MassEmailRecord({record})')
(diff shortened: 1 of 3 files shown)


Guitlle (approved):
Looks good to me

PR #5643, opened 2025-04-02
feat(massEmails): Test admin action for sending emails TASK-1732

Guitlle:
### 🗒️ Checklist

1. [ ] run linter locally

### 📣 Summary
Adds an action to the mass email configuration admin to test configurations by sending the first 20 enqueued emails.

### 👀 Preview steps
1. Create a new mass email configuration,  use the inactive users query
2. Modify temporarily the inactive users default parameter, to 2 days instead of 365
3. Restart celery containers to re-load the code 
4. Select the created configuration, in the admin list
5. Execute action
6. Check on sent emails

### DIFF
--- a/kobo/apps/mass_emails/admin.py
+++ b/kobo/apps/mass_emails/admin.py
@@ -1,3 +1,4 @@
+from django.conf import settings
 from django.contrib import admin, messages
 
 from .models import EmailStatus, MassEmailConfig, MassEmailRecord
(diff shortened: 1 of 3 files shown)


rgraber (changes requested):
Updating the cache could cause concurrency problems if more than one user is testing at the same time.
I think a safer option would be just to query for the users, select the first 20, and immediately call Mailer.sendEmail() rather than going through the regular process. I don't think we even need MassEmailRecords. This way we can be 100% sure testing will never interfere with the daily send, even if for some reason both are enabled.

PR #4403, opened 2023-04-06
Adds multiple features to the formbuilder, including `select_x_from_file` and others

dorey:
## Description
Adds the following new features to the formbuilder
1. Option to use an auto-completing dropdown for a `select_one` question
    - Listed under "Appearance (advanced)" in the question settings
2. Adds `select_one_from_file` question type to the formbuilder
    - The file must be a `.csv` and attached as a media file (requires attached file on form)
    - The file must already be attached in order to select the question type in the formbuilder
3. Adds `select_multiple_from_file` question type to the formbuilder
    - The file must be a `.csv` and attached as a media file (requires attached file on form)
    - The file must already be attached in order to select the question type in the formbuilder
4. Adds meta question `start-geopoint` to the formbuilder sidebar
    - Listed with the other meta questions in the sidebar
5. Adds ability to set appearance values on groups
    - Listed under "Appearance (advanced)" in the group settings

## Related issues
Fixes #2712, #2919, #470, probably some others

PR #5646, opened 2025-04-03
fix(formBuilder): Question Matrix settings crash DEV-39

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Opening settings for Question Matrix is possible again.

### 💭 Notes

I suspect the thing that broke it was #4403 and it was being broken for the last year. The investigation into what should appear in the settings lead me to believe that the `appearance` for `kobomatrix` is always `'field-list'`, so without much refactor I fixed the problem.

The things I've changed in the code:
- The list of `appearance` types now has a list for `kobomatrix`
- In the `afterRender` callback of `appearance` I fixed how we select current option, as there was a separate bug in there. In old code appearance type was always a `string` (type name), and in new code it can be also `string[]` (with first element being the type name), so we can't do a simple comparison anymore.

### 👀 Preview steps

Bug template:
1. Open Form Builder
2. Add new row "Question Matrix"
3. Open "Settings" (top right "cog" icon button)
4. 🔴 [on main] notice that settings don't show up and there is error in console
5. 🟢 [on PR] notice that settings do open correctly

### DIFF
--- a/jsapp/xlform/src/view.rowDetail.coffee
+++ b/jsapp/xlform/src/view.rowDetail.coffee
@@ -456,6 +456,9 @@ module.exports = do ->
 
   viewRowDetail.DetailViewMixins.appearance =
     getTypes: () ->
+      fieldListType = ['field-list', 'Show all questions in this group on the same screen']
+      groupTypes = ['select', fieldListType, ['other', 'Advanced']]
+
       types =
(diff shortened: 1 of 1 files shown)


pauloamorimbr (approved):
LGTM! Tested and works nicely! ✅

PR #5649, opened 2025-04-04
fix(stripe): annotate methods requiring stripe functionality

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Fixes a bug that was preventing startup when stripe was disabled.

### 💭 Notes
Adds a `stripe_required` annotation to make it clearer which methods require stripe to be enabled. Also removes an unused method, sets billing dates to the first of the month when stripe is disabled, and moves all stripe-disabled functionality to get_organization_effective_limits, since that's the main method that other apps interact with.

### 👀 Preview steps
Build KPI with stripe disabled.

### DIFF
--- a/kobo/apps/mass_emails/tests/test_usage_limit_user_queries.py
+++ b/kobo/apps/mass_emails/tests/test_usage_limit_user_queries.py
@@ -1,10 +1,13 @@
 from math import inf
+from typing import get_args
 from unittest.mock import patch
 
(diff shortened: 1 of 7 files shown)


Akuukis (approved):
~~LGTM~~ Tested, fixes the error I ran into recently. Started a fresh kobo-install instance, logged in, and no more errors.

PR #5650, opened 2025-04-04
ci: pytest with stripe disabled

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 💭 Notes
Adds a new GitHub action for running tests with Stripe disabled. In addition to turning off the boolean setting STRIPE_ENABLED, this settings file removes stripe apps from INSTALLED_APPS so the models/fields do not end up in the database. This should expose any instances where code was relying on these models or fields without an enclosing `if STRIPE_ENABLED` block.

Also uses `pytest.mark.skipif` to toggle tests that specifically require stripe to be enabled/disabled. This is better than using @override_settings because that only flips the STRIPE_ENABLED boolean and does not affect which models are in the database, so it's possible for bugs of the type mentioned before to slip through. To make this work, we needed to take some stripe-specific code out of `test_data.json`. It was only used by a few test suites however, so it was just a matter of moving the key generation to a mixin.

The workflow file was just straight up copy-pasted with the command altered to use the non-stripe environment. Just changing DJANGO_SETTINGS_MODULE in the job env is not enough because it overrides it with what is in pyproject.toml.


.github/workflows/pytest-stripe-disabled.yml
```diff
@@ -0,0 +1,68 @@
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: pytest
```
noliveleger, line 3: If we are to use this action, let's give it a different name from regular pytest.

kobo/apps/trackers/tests/test_utils.py
```diff
@@ -39,6 +38,7 @@ def setUp(self):
         self.asset.deploy(backend='mock', active=True)
 
     def _create_product(self, product_metadata):
+        from djstripe.models import Price, Product
```
noliveleger, line 41: Can't we use your decorator (`@requires_stripe`) here? 

If the answer is yes: 
I would like to see something like

```
Price = kwargs['product_model']
my_var = Price.objects.filter(...)
```
instead of calling `kwargs['product_model'].objects` directly 🙏

kobo/apps/trackers/tests/test_utils.py
```diff
@@ -54,6 +54,7 @@ def _make_payment(
         customer,
         payment_status='succeeded',
     ):
+        from djstripe.models import Charge, PaymentIntent
```
noliveleger, line 57: same question: Can't we use your decorator (@requires_stripe) here?

kobo/settings/testing_no_stripe.py
```diff
@@ -0,0 +1,4 @@
+from .testing import *  # noqa
+STRIPE_ENABLED = False
+INSTALLED_APPS = tuple(filter(lambda x: x not in ['djstripe', 'kobo.apps.stripe'],
+                              INSTALLED_APPS))  # noqa
```
noliveleger, line 4: Does not seem to be formatted with Black? 
It gives me 

```python 
INSTALLED_APPS = tuple(
    filter(lambda x: x not in ['djstripe', 'kobo.apps.stripe'], INSTALLED_APPS)
)
```

Why is the 2nd `noqa` needed?

kpi/tests/api/test_api_environment.py
```diff
@@ -100,15 +104,7 @@ def setUp(self):
...
-            'stripe_public_key': (
-                str(
-                    APIKey.objects.get(
-                        type='publishable', livemode=settings.STRIPE_LIVE_MODE
-                    ).secret
-                )
-                if settings.STRIPE_ENABLED
-                else None
-            ),
+            'stripe_public_key': None,
```
noliveleger, line 107: Let's leave a comment here that `stripe_public_key` can be overridden below when Stripe is enabled.

noliveleger (changes requested):
Running the unit tests for Python is already super long with CI. It takes about 40 minutes to accomplish. 
Do you think we could skip tests inside `openrosa` folder. Stripe should be only KPI, shouldn't it?

noliveleger:
> > Running the unit tests for Python is already super long with CI. It takes about 40 minutes to accomplish. Do you think we could skip tests inside `openrosa` folder. Stripe should be only KPI, shouldn't it?
> 
> I don't think this takes any longer because it runs in parallel. 

🤦  - Right! 

> Stripe stuff should only be KPI yes, but the recent bugs have mostly happened because we didn't realize something was secretly calling methods with stripe usage in the codeflow, so I think we should test all tests that aren't specifically marked as expecting stripe or expecting not-stripe.

You have a point.

noliveleger (approved):
LGTM

PR #5652, opened 2025-04-07
refactor(submissionUtils): migrate code to TypeScript TASK-1760

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 💭 Notes

Changes here:
- migrated `jsapp/js/components/submissions/submissionUtils.mocks.js` and `jsapp/js/components/submissions/submissionUtils.tests.js` to TypeScript
   - had to rewrite mock objects to match the TS interfaces
   - small improvements in few types
- `DisplayGroup` (from `jsapp/js/components/submissions/submissionUtils.tsx`) no longer has redundant `addChild()` function (did this to simplify writing tests and mocks)

### 👀 Preview steps

Testing:
1. run `npm test`? :shrug:
2. read code :shrug:


jsapp/js/components/submissions/submissionUtils.tsx
```diff
@@ -84,10 +84,6 @@ export class DisplayGroup {
       this.xpathNodes = xpathNodes
     }
   }
-
-  addChild(child: DisplayResponse | DisplayGroup) {
```
duvld, line 87: I think removing this function and instead using `children.push` directly makes it technically less safe (because there's no restriction on what can be pushed now)? I'm sure co-pilot would've caught it that is really true, seeing as `parentGroup` is typed.

jsapp/js/dataInterface.ts
```diff
@@ -173,6 +177,13 @@ export interface SubmissionSupplementalDetails {
   }
 }
 
+/**
+ * This is a completely empty object.
+ *
+ * We can't use `{}`, as it means "any non-nullish value". We are using `Record<string, never>` as the closes thing.
```
duvld, line 183: I don't understand why `{} | undefined` is not sufficient given the comment on line 246 below

duvld (commented):
Tiny questions

jsapp/js/dataInterface.ts
magicznyleszek: If I use `{}` then in the mocks the objects with missing propertes are not being caught :(

jsapp/js/components/submissions/submissionUtils.tsx
magicznyleszek: If I add `parentGroup.children.push('a')` somewhere in the function, TS will display an error :)

magicznyleszek: I think it's safe enough to use

duvld:
LGTM

jsapp/js/components/submissions/submissionUtils.tests.ts
```diff
@@ -52,144 +48,69 @@ chai.use(chaiExclude)
...
     const fileName = submissionWithAttachmentsWithUnicode.A_picture
-    const test = getMediaAttachment(submissionWithAttachmentsWithUnicode, fileName)
+    const test = getMediaAttachment(
+      submissionWithAttachmentsWithUnicode,
+      'kobo/attachments/45748fd461814880bd9545c8c8827d78/4cfa16e8-f29b-41a9-984c-2bf7fe05064b/Un_ete_au_Quebec_Canada-19_41_32.jpg',
+      'A_picture',
+    )
     const target = submissionWithAttachmentsWithUnicode._attachments[0]
-    expect(test).to.deep.equal(target)
+    chai.expect(test).to.deep.equal(target)
```
Akuukis, line 113: `fileName` is unused now?

jsapp/js/components/submissions/submissionUtils.mocks.ts
```diff
@@ -1,65 +1,129 @@
...
     name: 'blue',
     label: ['Blue'],
     list_name: 'fav_col_list',
+    $autovalue: '',
+    $kuid: '',
   },
 ]
 
-export const simpleSurveySubmission = {
+export const simpleSurveyAsset: AssetResponse = {
```
Akuukis, line 77: What do you say to use `as const satisfies Type` for mocks? I think that's more convenient. The `as const` will better reflect the nature of the mock being specific, and `satisfies Type` will assert type-safety whether we update types.

The differences:
```ts
interface Test {
  a: number
  b?: string
}

const v = { a: 1 } satisfies Test
// const v: {
//     a: number;
// }
const v2 = { a: 1 } as const satisfies Test
// const v2: {
//     readonly a: 1;
// }
const v3: Test = { a: 1 }
// const v3: Test
```

Akuukis (changes requested):

PR #5655, opened 2025-04-08
feat(longRunningMigrations): add system check to validate completion of critical long-running migrations

noliveleger:
### 📣 Summary
Added a Django system check to ensure important long-running migrations have completed

### 📖 Description
This PR introduces a Django system check that validates whether specific long-running migrations—considered critical to data integrity or system functionality—have been completed. If any of these migrations are still running or incomplete, the system check will raise an error to prevent the application from starting.

This mechanism not only protects the system from running in an inconsistent state, but also serves as a useful signal for sysadmins, giving them immediate visibility into whether the database is fully up-to-date. Instead of silently failing or relying on logs, this check provides a clear and explicit message during startup, making it easier to identify and act on pending background data migrations.

### 👀 Preview steps

Bug template:
1. Go to admin > Long running migrations 
2. Update 0007... to "in_progress" (or created, does not matter)
3. 🔴 [on main] The application starts as usual
5. 🟢 [on PR] The application does not start, Django system check returns an error

### DIFF
--- a/kobo/apps/long_running_migrations/app.py
+++ b/kobo/apps/long_running_migrations/app.py
@@ -1,6 +1,45 @@
 from django.apps import AppConfig
+from django.core.checks import Error, register, Tags
+from django.db import connection
+
+from .constants import MUST_COMPLETE_LONG_RUNNING_MIGRATIONS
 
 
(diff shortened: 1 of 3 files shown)


sksondra (commented):
Good from QA.

PR #5659, opened 2025-04-09
feat(massEmails): Celery task for sending emails TASK-1692

Guitlle:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Adds a new task to sync with the recurring emails feature. For now this task is not enabled but it will replace the old task in the near future.

### 👀 Preview steps
For now unit tests cover the functionality, though the task can be executed manually. These steps may be followed:

1. Set the django emails backend setting to file backend or console backend in order to be able to debug
2. Set up one or more mass email configs
3. You can use the inactive user query, and change temporarily the inactive days parameter so that the users email records list is not empty
4. Enter to the manage.py shell and import the tasks generate_mass_email_user_lists and _send_emails 
5. Execute generate_mass_email_user_lists manually and then _send_emails 
6. Review that the emails were sent

### DIFF
--- a/kobo/apps/mass_emails/tasks.py
+++ b/kobo/apps/mass_emails/tasks.py
@@ -35,6 +35,7 @@ templates_placeholders = {
 
 PROCESSED_EMAILS_CACHE_KEY = 'mass_emails_{today}_emails'
 
+
 def enqueue_mass_email_records(email_config):
     """
     Creates a email job and enqueues email records for users based on query
--- a/kobo/apps/mass_emails/tasks.py
+++ b/kobo/apps/mass_emails/tasks.py
@@ -99,7 +99,7 @@ class MassEmailSender:
                 'jobs__records',
                 filter=Q(jobs__records__status=EmailStatus.ENQUEUED),
             )
-        )
+        ).filter(enqueued_records_count__gt=0)
         logging.info(f'Found {self.total_records} enqueued records')
         self.get_day_limits()
 
--- a/kobo/apps/mass_emails/tasks.py
+++ b/kobo/apps/mass_emails/tasks.py
@@ -228,6 +228,29 @@ def send_emails():
     sender.send_day_emails()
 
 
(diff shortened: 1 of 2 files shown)


kobo/apps/mass_emails/tasks.py
```diff
@@ -228,6 +228,29 @@ def send_emails():
...
+@celery_app.task(time_limit=3300)  # 55 minutes
+def _send_emails():
+    """Send the emails for the current day. It schedules the emails if they have not
+    been scheduled yet.
+
+    NOTE: This function will replace the function called send_emails in the near future.
+    """
+    today = timezone.now().date()
+    sender = MassEmailSender()
+    cache_key = f'mass_emails_{today}_emails'
```
rgraber, line 240: Can we use a constant for this template so we can keep it consistent between this and `generate_mass_email_user_lists`?

rgraber (changes requested):
One small request

kobo/apps/mass_emails/tasks.py
```diff
@@ -228,6 +230,29 @@ def send_emails():
...
+    been scheduled yet.
+
+    NOTE: This function will replace the function called send_emails in the near future.
+    """
+    today = timezone.now().date()
+    sender = MassEmailSender()
+    cache_key = PROCESSED_EMAILS_CACHE_KEY.format(today=today)
+    cached_data = cache.get(cache_key, [])
+    processed_configs = set(cached_data)
+    config_ids = {email_config.id for email_config in sender.configs}
```
rajpatel24, line 245: It looks like `sender.configs` returns all configs without considering the status of their records. I think we want to only fetch the configs that have enqueued records.

We could modify it like so:
```
active_configs = sender.configs.filter(enqueued_records_count__gt=0)
config_ids = {email_config.id for email_config in active_configs}
```

rajpatel24 (changes requested):

rgraber (approved):
LGTM

PR #5663, opened 2025-04-10
refactor(longRunningMigrations): update MongoDB documents alongside PostgreSQL in attachment backfill TASK-1773

noliveleger:
### 📣 Summary
Refactored long-running migration `0007` to update MongoDB documents in the same loop as PostgreSQL.

### 📖 Description
This PR refactors the long-running migration `0007`, which backfills the Attachment model, to simultaneously update corresponding MongoDB documents. Specifically, the migration now injects the `uid` field directly into each attachment in MongoDB while processing the matching records in PostgreSQL.

It prepares the codebase for upcoming work on attachment handling.

### 👀 Preview steps

1. Use `release/2.024.10` first (with no migrations from `2.024.14` or `main`)
2. Add a couple of submissions with attachments
3. 🔴 [release/2.024.10] notice that data endpoint (`/api/v2/assets/<asset_uid>/data.json`)  does not expose `uid` in attachments.
4. Checkout this branch 
5. 🟢 [on PR] notice that data endpoint does not expose `uid` in attachments.


rajpatel24 (approved):
LGTM

PR #5664, opened 2025-04-11
fix: prevent 500 error when submission `root_uuid` is null in data API

noliveleger:
### 📣 Summary
Fixed a server error triggered when accessing the data API and `root_uuid` was missing on a submission instance.

### 📖 Description
The data API endpoint returned a 500 Internal Server Error when attempting to fetch a submission where `instance.root_uuid` was null.


noliveleger:
Tests are passing locally. I'm not waiting for CI to finish before merging. PRs is needed to create new image.

PR #5676, opened 2025-04-16
refactor(storybook): migrate button tests DEV-97

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 💭 Notes

Changes here:
- Renamed `button.stories.tsx` to `ButtonNew.stories.tsx` to match the component and make it clear it's distinct from old button
- Migrated two tests from `.spec.tsx` file into stories
- Reorganized stories of `Button`, now it has "Docs", "Default", "Preview" and two tests

### 👀 Preview steps

Testing:
1. ℹ️ run storybook
2. go to Design System → Button → Test Click
3. open "Interactions" tab
4. 🟢 notice that test PASS

Second case:
1. do same thing for Design System → Button → Test Click Disabled


jsapp/js/components/common/ButtonNew.stories.tsx
```diff
@@ -105,49 +105,17 @@ const meta: Meta<typeof Button> = {
 }
```
Akuukis, line 105: use `satisfies Meta<typeof Button>`

Akuukis (approved):
See the pushed commit. LGTM and tested, except:
- [ ] one inline comment  
- [ ] jest fails with removing the last `*.spec.ts` file

PR #5681, opened 2025-04-16
fix(languageSelector): force refresh languages list after clearing the searchbox DEV-183

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 👀 Preview steps

Steps:

1. Have a project with `audio` question and a submission
2. Go to Project → Data → Table
3. Use "Open" button for the `audio` question response to open NLP View
4. Click "Begin" to start creating transcript
5. Type "swed" in the search box
6. 🟢 Notice the results are only one item: "Swedish"
7. Use "x" button in the searchbox to clear results
8. 🟢 Notice the searchbox is cleared
9. 🔴 Notice the list is not refreshed on kf.beta
10. 🟢 Notice the list is properly refreshed on PR branch

### DIFF
--- a/jsapp/js/components/languages/languageSelector.tsx
+++ b/jsapp/js/components/languages/languageSelector.tsx
@@ -189,12 +189,12 @@ class LanguageSelector extends React.Component<LanguageSelectorProps, LanguageSe
   }
 
   clearSearchPhrase() {
-    this.setSearchPhrase('')
+    this.setSearchPhrase('', true)
   }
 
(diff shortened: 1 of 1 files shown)


sksondra (commented):
Noticed an issue while testing - there's an extra "x" button that is causing the same issue [here.](https://linear.app/kobotoolbox/issue/DEV-360/extra-x-button-causing-language-list-not-to-reset)

PR #5685, opened 2025-04-16
feat(billing): add xform pending_transfer flag (TASK-1638)

jamesrkiger:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Adds a new `pending_transfer` field to the xform model and updates the `suspend_submissions()` method on  `OpenRosaDeploymentBackend` to update this field instead of updating the `submissions_suspended` flag on the user model. 

### 💭 Notes
1. This PR includes the work for [TASK-1642](https://www.notion.so/kobotoolbox/Add-new-field-to-XForm-1a77e515f6548010925ce5115d551866?pvs=4). I thought it would be better to just do one PR so there could be some test coverage when adding the field and relevant migration.

2. There is, as far as I can tell, no existing test coverage for the actual operation of the `suspend_submissions()` method, neither on its own nor in the context of project transfers. My impression is that this would be a fairly complicated test, so I didn't want to spend time on it if it is out of scope. If you think we should add coverage for this in the context of this PR, please let me know.

### 👀 Preview steps
I don't think there is a very practical way to test this change in action, short of creating a very large project that will require significant transfer time and attempting to make submissions in that time period. I updated a test [here](https://www.notion.so/kobotoolbox/Add-new-field-to-XForm-1a77e515f6548010925ce5115d551866?pvs=4) to ensure that the `pending_transfer` flag itself does indeed block submissions.


noliveleger (changes requested):

noliveleger (approved):
LGTM

sksondra (commented):
Good from QA.

> I don't think there is a very practical way to test this change in action, short of creating a very large project that will require significant transfer time and attempting to make submissions in that time period.

Since this is a bit tricky to test itself, I went through the related PRs/tasks and made sure everything looks good. I'll keep this PR in mind when I test other billing/transfer related tasks.

PR #5687, opened 2025-04-17
fix(attachmentStorage): fix UserProfile storage counter to exclude soft-deleted XForms TASK-1777

rajpatel24:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Updates the management command to ensure that deleted projects (forms in the trash) are no longer counted toward the user profile's attachment storage usage.

### 📖 Description
Previously, when a user soft-deleted a project, the media files associated with that form continued to be included in the UserProfile's `attachment_storage_bytes` counter, which is used to track total storage usage per account. This led to inconsistencies between the backend storage counter and what users saw in the `/service_usage/` page, which already excluded deleted forms.

This fix updates the `update_attachment_storage_bytes` management command to exclude soft-deleted forms (`pending_delete=True`) from the total storage calculation at the user profile level. Individual XForm storage totals are still updated, even for trashed forms, so they remain accurate if restored.


kobo/apps/trash_bin/models/project.py
```diff
@@ -83,4 +86,11 @@ def toggle_statuses(
                 )
                 assert updated >= kc_updated
 
+                # Update attachment storage counters
+                attachments = Attachment.objects.filter(
+                    xform__kpi_asset_uid__in=object_identifiers
+                )
+                bulk_update_attachment_storage_counters(
+                    attachments, subtract=not active
+                )
```
noliveleger, line 95: Beware, I think the calculation always takes place whatever the value of `toggle_delete`. 
When `toggle_delete` is False, projects are only archived. The storage should not be affected. If the boolean `toggle_delete` is not obvious enough, do not hesitate to rename it and add explanation in the docstring how it should be used. i.e. 

`active`| `toggle_delete` | behaviour
--- | --- | --- 
False | False | Archive project
False | True | Prepare project for soft-deletion (`move_to_trash` needs to be called right after)
True | False | Unarchive project
True | True | Prepare project to be restored from trash (`put_back` needs to be called right after)

kobo/apps/trash_bin/tests/test_utils.py
```diff
@@ -331,3 +341,174 @@ def test_delete_project(self):
...
+                    'filename': f'{username}/audio_conversion_test_image.jpg',
+                    'mimetype': 'image/jpeg',
+                },
+            ],
+            '_submitted_by': username,
+        }
+        asset.deployment.mock_submissions([submission])
+
+        # Fetch the storage data before moving the project to trash
+        xform_before = XForm.all_objects.filter(kpi_asset_uid=asset.uid).aggregate(
```
noliveleger, line 386: Use already calculated fields:
- `UserProfile.attachment_storage_bytes`
- `XForm.attachment_storage_bytes`

You need to compare the values with these fields

kobo/apps/trash_bin/tests/test_utils.py
```diff
@@ -331,3 +341,174 @@ def test_delete_project(self):
...
+            total_bytes=Sum('attachment_storage_bytes')
+        )['total_bytes']
+        user_profile_before = UserProfile.objects.filter(user_id=user.pk).aggregate(
+            total_bytes=Sum('attachment_storage_bytes')
+        )['total_bytes']
+
+        # Move project to trash
+        ProjectTrash.toggle_statuses(
+            [asset.uid], active=False, toggle_delete=True
+        )
```
noliveleger, line 396: You need to add a test when projects are only archived

kobo/apps/trash_bin/tests/test_utils.py
```diff
@@ -331,3 +341,174 @@ def test_delete_project(self):
...
+        # Counters should be restored to original values
+        self.assertEqual(
+            self.xform.attachment_storage_bytes, original_xform_bytes
+        )
+        self.assertEqual(
+            self.user_profile.attachment_storage_bytes, original_user_bytes
+        )
+        self.assertIsNone(self.attachment.delete_status)
+
+    def test_deleting_submission_does_not_decrease_counters_twice(self):
```
noliveleger, line 489: 🍾

noliveleger (changes requested):

kobo/apps/trash_bin/tests/test_utils.py
noliveleger: Please split this test: one for deletion, one for archiving (even if they use almost the same code).

noliveleger (changes requested):

noliveleger (approved):
LGTM

sksondra (commented):
Good from QA.

PR #5688, opened 2025-04-17
fix(migrations): fix null search_fields

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Fixes 500 errors when updating organization names.

This migration should be applied with `--fake` on large instances.

### 💭 Notes
Use LazyDefaultJSONBField instead of just a regular JSON field so python will automatically convert null `search_field`s to empty dictionaries when initializing the model.


noliveleger (approved):
LGTM

PR #5689, opened 2025-04-17
fix(api): prevent 500 errors when accessing v1 API endpoints

noliveleger:
### 📣 Summary
Fixed server errors caused by unexpected permission mutations in views using `DjangoModelPermissions`.

### 📖 Description
This PR resolves 500 Internal Server Errors occurring when accessing certain endpoints of the v1 API. The issue was caused by an unintended mutation of the `perms_map` attribute in classes inheriting from DjangoModelPermissions.

Since perms_map is shared across instances, modifying it dynamically introduced side effects across views and requests. This fix prevents the mutation and ensures that permission handling remains isolated and reliable across the API.

### 👀 Preview steps

Bug template:
2. preview the project
3. access `http://kc/api/v1/forms.json`
4. 🔴 [on release branch] receive a 500 (because of AssertionError)
5. 🟢 [on PR] everything is ok

### DIFF
--- a/kobo/apps/openrosa/apps/api/permissions.py
+++ b/kobo/apps/openrosa/apps/api/permissions.py
@@ -1,4 +1,5 @@
-# coding: utf-8
+from copy import deepcopy
+
 from django.http import Http404
 from rest_framework.permissions import (
     SAFE_METHODS,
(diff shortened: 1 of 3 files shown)


rgraber (approved):
LGTM

PR #5690, opened 2025-04-17
refactor(migrations): optimize heavy migrations to reduce deployment downtime for release `2.025.02` 

noliveleger:
### 📣 Summary
Refactored Django migrations that were slow to apply on large datasets.

### 📖 Description
This PR refactors slow Django migrations originally introduced in release `2.025.02`. These migrations caused significant delays during deployment in environments with large datasets.

The goal is to reduce downtime by restructuring expensive operations and deferring them when possible. This includes moving certain logic to long-running background migrations, and avoiding operations that would trigger full table rewrites during deployment.

### Notes 
- The long-running job `0005_backfill_logger_instance_root_uuid` could enter a race condition with migration `logger.0041`.
 A check has been added to ensure it does not start until `logger.0041` has been fully applied.
- An `error` field has also been added to long-running migration jobs to provide feedback in case of failures. This information is visible from the Admin UI
- Default value (`False`)  of `Asset.is_excluded_from_projects_list`, removed in `kpi.0063` is now backfilled with long-running migration job `0007`
- Django migration `organization.0007` has been moved to long-running migration job `0008`

- The branch has been tested on staging servers (`kvin` and `master`) respectively on `2.024.36h`  and `2.025.02c`

### 👀 Preview steps

Pretty hard to give instructions to test without a (very) large dataset but at least, migrations can be tested. 

1. Create a version on last version of release `2.024.36` and add some data. 
2. Upgrade to this branch and apply migrations (`./scripts/migrate.sh`)
3. Repeat 1 and 2 with `SKIP_HEAVY_MIGRATIONS=True` as a environment variable
   3a. follow instructions to apply indexes manually in PostgreSQL
4. When it's done,  force long-running migrations to run with

```python
from kobo.apps.long_running_migrations.tasks import execute_long_running_migrations
execute_long_running_migrations()
# or it could be executed in Celery but make sure your celery 
# containers are up-to-date and are running the correct version
# execute_long_running_migrations.delay() 
```

5. Repeat 1 to 4 but starting from version `2.025.02d`


jnm (approved):
Approved per our call :+1:

PR #5692, opened 2025-04-20
refactor(submissions): optimize `clean_duplicated_submissions` command with batching and project scoping

noliveleger:
### 📣 Summary
Refactored the `clean_duplicated_submissions` management command to process instances in smaller batches and grouped instances by project to ensure UUID uniqueness is enforced per project, not globally.

### 📖 Description
The `clean_duplicated_submissions` command has been refactored to address performance and correctness issues on large datasets. Previously, it attempted to load and process all submissions at once, which could cause out-of-memory (OOM) errors. It also enforced submission uuid uniqueness globally, instead of per project, which is incorrect.

### DIFF
--- a/kobo/apps/openrosa/apps/logger/management/commands/clean_duplicated_submissions.py
+++ b/kobo/apps/openrosa/apps/logger/management/commands/clean_duplicated_submissions.py
@@ -1,3 +1,4 @@
+import time
 from collections import defaultdict
 
 from django.conf import settings
@@ -5,6 +6,7 @@ from django.core.management.base import BaseCommand
 from django.db import transaction
 from django.db.models import F
 from django.db.models.aggregates import Count
+from more_itertools import chunked
 
 from kobo.apps.openrosa.apps.logger.models import (
     DailyXFormSubmissionCounter,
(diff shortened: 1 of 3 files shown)


kobo/apps/openrosa/apps/logger/management/commands/clean_duplicated_submissions.py
```diff
@@ -36,60 +35,65 @@ def add_arguments(self, parser):
...
-        query = (
-            query.values_list('uuid', flat=True)
+        queryset = self._filter_queryset(
+            Instance.objects.values_list('uuid', flat=True)
             .annotate(count_uuid=Count('uuid'))
             .filter(count_uuid__gt=1)
-            .distinct()
+            .distinct(),
+            **options,
         )
```
rajpatel24, line 60: If we want to enforce UUID uniqueness per project and not globally, isn't something wrong here?

This code retrieves UUIDs that are duplicated globally. Then, we loop over each `xform_id` after detecting global duplicates, which seems too late, as we have already included UUIDs that are duplicated across different projects.

Maybe it should be something like this:
```
queryset = self._filter_queryset(
    Instance.objects.values('uuid', 'xform_id')
    .annotate(count=Count('id'))
    .filter(count__gt=1),
    **options,
)
```

Please correct me if I misunderstood anything :-)

rajpatel24 (changes requested):

rajpatel24 (approved):
LGTM

PR #5694, opened 2025-04-21
fix(attachments): fix import order

duvld:
### 💭 Notes

Run biome import sort

### 👀 Preview steps

Bug template:
4. 🔴 [on main] biome complains
5. 🟢 [on PR] biome doesn't complain


sksondra (commented):
Good from QA.

PR #5695, opened 2025-04-21
feature(attachments): make bulk delete attachments account for partial permissions TASK-559

duvld:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Displays the correct number of attachments/submissions in the UI when bulk deleting based on current user permissions.

### 👀 Preview steps

Feature/no-change template:
1.  Have a form with submissions and `removingAttachmentsEnabled` flag enabled
2. Give partial delete permissions to another user somehow. The way I did it was:
    1. Have 2 accounts
    2. Submit data logged in on both accounts (so there are submissions from multiple accounts)
    3. Give one user permission to view submissions from both account
    4. Give the same user in step above permission to only delete from one account, let's call this the Test User.
6. As the Test User, select all submissions and click "Delete media files only"
8. 🟢 notice that the number of attachments counted in the modal correctly matches those that come from the submission(s) Test User can delete
9. 🟢notice that after a successful "delete" the notify displays the correct number of submissions that Test User can delete

### DIFF
--- a/jsapp/js/attachments/BulkDeleteMediaFiles.tsx
+++ b/jsapp/js/attachments/BulkDeleteMediaFiles.tsx
@@ -68,36 +68,46 @@ export default function BulkDeleteMediaFiles(props: BulkDeleteMediaFilesProps) {
 
       <Modal opened={opened} onClose={handleCloseModal} title={t('Delete media files')} size={'md'}>
         <FocusTrap.InitialFocus />
-        <Stack>
-          <Checkbox
-            label={
-              <Text>
-                {t('You are about to permanently remove the following media files from the selected submissions:')}
-                <br />
-                {getMediaCount(filteredSubmissions)}
-              </Text>
-            }
-            onClick={() => setWarningAcknowledged(!warningAcknowledged)}
-            checked={warningAcknowledged}
-          />
-          <Alert iconName='warning' type='error'>
-            {t('Careful - it is not possible to recover deleted media files')}
-          </Alert>
-          <Group justify='flex-end'>
-            <Button variant='light' size='lg' onClick={close} disabled={isDeletePending}>
-              {t('Cancel')}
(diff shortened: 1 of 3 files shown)


magicznyleszek (changes requested):

jsapp/js/attachments/BulkDeleteMediaFiles.tsx
```diff
@@ -61,36 +68,46 @@ export default function BulkDeleteMediaFiles(props: BulkDeleteMediaFilesProps) {
...
-              onClick={handleConfirmDelete}
-            >
-              {t('Delete')}
-            </Button>
-          </Group>
-        </Stack>
+        {filteredSubmissions.length === 0 && (
+          // This can only happen if the user has permissions to view (but not delete) certain submissions
+          <Stack>
+            <Text>{t('You do not have permissions to delete the selected submissions.')}</Text>
```
magicznyleszek, line 75:
```suggestion
            <Text>{t('You do not have sufficient permissions to delete attachments from the selected submissions.')}</Text>
```
What do you think about this? We are deleting attachments here, right? :)

magicznyleszek: And would you also make a screenshot of that modal and ask Tessa what does she thing about it? I am wondering for example if we should use `<Alert` here with yellow color for this message. But I would rather merge this PR and have some additional change requested by Tessa in other PR :)

magicznyleszek (changes requested):

jsapp/js/attachments/BulkDeleteMediaFiles.tsx
duvld: Let's merge it and let Tessa give a more comprehensive review of the UI once it's done

PR #5700, opened 2025-04-23
fix(PyMongo): enforce maxTimeMS using low-level command to ensure query timeout

noliveleger:
### 📣 Summary
Enforced maxTimeMS on MongoDB queries by using the low-level `command()` API.

### 📖 Description
PyMongo does not always forward the `maxTimeMS` parameter to MongoDB in certain high-level query operations, which can result in long-running queries blocking resources indefinitely.

To address this, this PR enforces the use of maxTimeMS by leveraging MongoDB’s low-level `command()` interface, ensuring that the timeout is always respected regardless of the type of operation or query abstraction used.

### Notes
Unfortunately, MockMongo does not support `command()` yet and raise `NotImplementedError`.

### DIFF
--- a/kobo/apps/openrosa/apps/logger/management/commands/clean_duplicated_submissions.py
+++ b/kobo/apps/openrosa/apps/logger/management/commands/clean_duplicated_submissions.py
@@ -139,7 +139,7 @@ class Command(BaseCommand):
 
             instance_queryset.delete()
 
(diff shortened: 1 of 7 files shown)


jnm (changes requested):
This looks good, but are the `locale` changes intentional?

jnm (approved):

PR #5701, opened 2025-04-23
perf(tasks): reduce the number of stuck tasks restarted to ease DB load

noliveleger:
### 📣 Summary
Limited the number of stuck tasks restarted at once to reduce pressure on the database.

### 📖 Description
This update improves the performance of the task restarter by reducing the number of stuck or incomplete tasks restarted in a single run. Restarting too many tasks at once could result in a burst of activity and high load on the database, potentially slowing down the system.

The new behavior introduces a more gradual, controlled approach to restarting tasks, ensuring smoother system operation while still addressing unfinished work in a timely manner.

### DIFF
--- a/kobo/apps/project_ownership/tasks.py
+++ b/kobo/apps/project_ownership/tasks.py
@@ -266,11 +266,15 @@ def task_restarter():
     )
 
     # Restart tasks that were stopped unintentionally.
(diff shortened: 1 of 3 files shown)


jnm (approved):
:+1:

PR #5703, opened 2025-04-24
fix(billing): filter for plans when querying active subscription details DEV-206

jamesrkiger:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Fixes method for querying an organization's relevant subscription to ensure that only subscription plans are queried.

### 💭 Notes
The original error we ran into was that organizations with an MMO-enabled subscription **and** an SSO addon (with `'product_type': 'addon'` in product metadata) are not being recognized as MMOs in our code. This is because the `active_subscription_billing_details` Organization method does not distinguish between 'addon' and 'plan' subscriptions. However, because the billing code as of release `2.025.02` uses `active_subscription_billing_details` for determining billing dates, I have chosen to leave that method untouched and instead run a similar, but fixed, query directly in the `is_mmo` method.

### 👀 Preview steps
I leave it to the reviewer to determine whether an actual test is worth the effort. Ping me for help with step 2 if you decide to go for it.
1. In a stripe-enabled (and synced) environment with a new user/org, navigate to http://kf.kobo.local/#/account/plan?type=custom_teams and sign up for the Teams Unlimited plan. 
2. Add a recurring addon for SSO (this has to be done on the Stripe platform, as it is not a public-facing option)
3. Sync subscriptions with Stripe
4. Navigate to http://kf.kobo.local/#/account/
5. 🔴 [on `release/2.025.02`] Notice that the UI does not reflect MMO status (no org name in header, no "organization" section in sidenav)
6. 🟢 [on PR] notice that MMO status is reflected

### DIFF
--- a/kobo/apps/organizations/models.py
+++ b/kobo/apps/organizations/models.py
@@ -83,7 +83,7 @@ class Organization(AbstractOrganization):
             Organization.objects.prefetch_related('djstripe_customers')
             .filter(
                 djstripe_customers__subscriptions__status__in=ACTIVE_STRIPE_STATUSES,
-                djstripe_customers__subscriptions__items__price__product__metadata__product_type='plan',
+                djstripe_customers__subscriptions__items__price__product__metadata__product_type='plan',  # noqa
                 djstripe_customers__subscriber=self.id,
             )
             .order_by('-djstripe_customers__subscriptions__start_date')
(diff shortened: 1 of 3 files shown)


rgraber (changes requested):
One request for a comment

rgraber (approved):
LGTM

PR #5706, opened 2025-04-24
feat(projectHistoryLogs): add project owner to related viewsets

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Add project owner to project history logs for adding/removing media, exporting data, importing fields from other projects, and modifying REST services.

### 💭 Notes
Though the endpoints are very different, it makes sense to do this in one PR because they all go through the `_related_request_base` method and require the exact same change in the viewset.

### 👀 Preview steps

1. ℹ️ have account and at least 2 projects
2. Submit some data to one of the projects
3. Export data from the project
4. Go to Settings > Media. Add and remove an image from the project.
5. Make sure the second project has data sharing enabled.
6. Go to Settings > Connect Project
7. Import fields from the second project, change the imported fields, then remove the import.
8. Go to Settings > REST Services
9. Add a new REST service, change its name, then remove it
10. Go to `api/v2/assets/*uid*/history`
11. 🟢 There should be 9 new project history logs (add/delete-media, dis/connect-project, modify-imported-fields, register/delete-service, modify-service, and export). Each should have the project_owner in the metadata.

### DIFF
--- a/kobo/apps/audit_log/models.py
+++ b/kobo/apps/audit_log/models.py
@@ -937,12 +937,14 @@ class ProjectHistoryLog(AuditLog):
             # request failed, don't try to log
             return
         object_id = source_data.pop('object_id')
+        owner = source_data.pop('asset.owner.username')
 
         metadata = {
             'asset_uid': asset_uid,
             'log_subtype': PROJECT_HISTORY_LOG_PROJECT_SUBTYPE,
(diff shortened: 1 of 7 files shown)


Guitlle (approved):
It works

sksondra (commented):
Good from QA.

PR #5708, opened 2025-04-24
fix(mongoHelper): use correct MongoDB collection for query operations

noliveleger:
### 📣 Summary
Fixed an issue where queries were executed on the wrong MongoDB collection.

### 📖 Description
A bug caused certain operations to target the wrong MongoDB collection, leading to incorrect results or missing data during queries or updates. This fix ensures that all MongoDB interactions use the correct collection name.

### 💭 Notes
The data is not permanently lost. Only the MongoDB copy of data from PostgreSQL was missing. It can be fully restored using the following command:

```bash
./manage.py sync_mongo --remongo <username> <xform_id_string>
```

This will resync the expected data into MongoDB from the authoritative PostgreSQL source.

### 👀 Preview steps

Bug template:
2. submit data
3. 🔴 [on 2.025.02f] go to table view, data is not present
4. 🟢 [on PR] data is present
5. 🔴 [on 2.025.02f] transfer a project to another user, project is transferred but data may still belong to previous owner
6. 🟢 [on PR] data does belong to new owner
7. 🔴 [on 2.025.02f] delete data from table view. Reload, data is still there 
8. 🟢 [on PR] delete data. data is gone. From the shell: use `settings.MONGO_DB.instances.find_one({'_id': <submission_id>}) to validate the document is not in MongoDB anymore

### DIFF
--- a/kpi/utils/mongo_helper.py
+++ b/kpi/utils/mongo_helper.py
@@ -74,6 +74,7 @@ class MongoHelper:
     USERFORM_ID = '_userform_id'
     SUBMISSION_UUID = '_uuid'
     DEFAULT_BATCHSIZE = 1000
+    COLLECTION = 'instances'
 
     @classmethod
     def decode(cls, key):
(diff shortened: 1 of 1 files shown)


Guitlle (approved):

rajpatel24 (approved):
LGTM

PR #5720, opened 2025-04-29
test(projectHistory): FormActivity stories DEV-100

magicznyleszek:
### 💭 Notes

Interesting things here:
- Added `storybook-addon-remix-react-router` that allows using router (useful for mocking parameters in route path)
- Created `/api/v2/assets/<uid>/history/` endpoint mock
- Created `/api/v2/assets/<uid>/history/actions/` endpoint mock
- Cleaned up some history related types (updated to match current BE)
- Introduced `getTextContentOnly` utility for getting text content from HTML (useful for targeting canvas elements inside stories)
- Added stories for `FormActivity` component that includes a test (filtering by activity type)

### 👀 Preview steps

Testing:
1. ℹ️ run storybook
2. go to Features → FormActivity → Test Filtering By Activity Type story
3. open "Interactions" tab
4. 🟢 notice that test PASS

### DIFF
--- a/jsapp/js/endpoints/assetHistory.mock.ts
+++ b/jsapp/js/endpoints/assetHistory.mock.ts
@@ -1,7 +1,7 @@
 import { http, HttpResponse, type PathParams } from 'msw'
 import { endpoints } from '#/api.endpoints'
+import { type ActivityLogsItem, AuditActions, AuditSubTypes } from '#/components/activity/activity.constants'
 import type { PaginatedResponse } from '#/dataInterface'
-import { type ActivityLogsItem, AuditActions, AuditSubTypes } from './activity.constants'
 
 export const mockAssetUid = 'a1234567890bcdEFGhijkl'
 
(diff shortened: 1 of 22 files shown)


jsapp/js/endpoints/assetHistory.mock.ts
```diff
@@ -0,0 +1,630 @@
...
+    } else if (searchParams.get('limit') === '10' && searchParams.get('offset') === '10') {
+      return HttpResponse.json(assetHistoryResponsePage2nd)
+    } else {
+      return HttpResponse.json(assetHistoryResponsePage1st)
+    }
+  },
+)
+export default assetHistoryMock
+
+const assetHistoryResponsePage1st: PaginatedResponse<ActivityLogsItem> = {
```
pauloamorimbr, line 28: For simplicity, WDYT about using page1, page2, page3 instead of ordinals? 😁 

Like: `assetHistoryResponsePage1`

...and even, since we're already inside of the assetHistory mock we could as well just call it `responsePage1`, or even `page1`. WDYT?

pauloamorimbr (approved):
LGTM! 👏🏻  💯 
well done!

Left 2 little suggestions.

jsapp/js/endpoints/assetHistory.mock.ts
magicznyleszek: I think there is no harm in having long name, but I did change 1st to 1. Also updated other mock file accordingly.

PR #5723, opened 2025-04-30
ci(storybook): install playwright and run tests

magicznyleszek:
### 🗒️ Checklist

1. [ ] run linter locally

### 👀 Preview steps

Bug template:
2. do this
3. do that

Feature/no-change template:
2. do this
3. do that
5. do that another thing


pauloamorimbr (approved):
LGTM! 👍🏻 

I'm not experienced with playright, but I took a deep look, researched and as far as I could check, it's all right, CI working, I didn't catch any errors!

.github/workflows/npm-test.yml
```diff
@@ -76,3 +76,26 @@ jobs:
...
+        if: steps.playwright-cache.outputs.cache-hit != 'true'
+      - run: npx playwright install-deps
+        if: steps.playwright-cache.outputs.cache-hit != 'true'
+
+      # Run the storybook test suite
+      - name: Serve Storybook and run tests
+        run: |
+          npx concurrently -k -s first -n "SB,TEST" -c "magenta,blue" \
+            "npx http-server storybook-static --port 6006 --silent" \
+            "npx wait-on tcp:6006 && npx test-storybook"
```
Akuukis, line 101: I see in CI that only chromium is used, let's enable firefox and webkit as well, see https://storybook.js.org/docs/writing-tests/test-runner#cli-options

.github/workflows/npm-test.yml
```diff
@@ -76,3 +76,26 @@ jobs:
...
+          path: |
+            ~/.cache/ms-playwright
+          key: ${{ runner.os }}-playwright-${{ env.PLAYWRIGHT_VERSION }}
+      - run: npx playwright install --with-deps
+        if: steps.playwright-cache.outputs.cache-hit != 'true'
+      - run: npx playwright install-deps
+        if: steps.playwright-cache.outputs.cache-hit != 'true'
+
+      # Run the storybook test suite
+      - name: Serve Storybook and run tests
```
Akuukis, line 97: add this to `ci_locally.sh` as well

package.json
```diff
@@ -216,11 +218,12 @@
     "postinstall": "patch-package && npm run copy-fonts",
     "build": "npm run hint build && npm run build:app && npm run build:storybook",
     "build:app": "webpack --config webpack/prod.config.js",
-    "build:storybook": "webpack --config webpack/prod.config.js",
+    "build:storybook": "sb build",
     "watch": "npm run hint watch && webpack-dev-server --config webpack/dev.server.js",
     "watch:storybook:": "storybook dev -p 6006 --no-open",
     "ci": "npm run lint && SKIP_TS_CHECK=true npm run build && npm run test",
     "test": "npm run test:unit",
```
Akuukis, line 225:
```suggestion
    "test": "npm run test:unit && npm run test:storybook",
```

Akuukis (changes requested):

.github/workflows/npm-test.yml
magicznyleszek: I see that `ci_locally.sh` is running `npm run ci`, which is running `npm run test`, which now has storybooks tests in it :)

.github/workflows/npm-test.yml
magicznyleszek: I've added more browsers 👍

magicznyleszek: I've also reorganized scripts a bit, as I've noticed that running tets on my local machine worked only if I was already running storybook (`npm run storybook`) on separate terminal tab.

Now we have two scripts:
1. `test:storybook-ci` - assumes storybook is built, and assumes playwright is installed (both of these are already done in `.github/workflows/npm-test.yml`)
2. `test:storybook-locally` - builds storybook first (without it there is a risk that tests will be run on storybook built on different commit - I had that error), then installs playwright (without caching)

For running the tests, I've added `--testTimeout=30000 --maxWorkers=2` as the tests were failing sometimes on my machine.

magicznyleszek: I also had to add some hack for webkit browser (linked the bug in comment)

PR #5724, opened 2025-04-30
fix(projectOwnershipTransfer): make transfer more resilient by catching exceptions

noliveleger:
### 📣 Summary
Improved the robustness of the project ownership transfer process by catching and handling exceptions.

### 📖 Description
This update enhances the resilience of the project ownership transfer process by catching exceptions that may occur during the transfer of individual resources (e.g., assets, attachments, submissions).

Instead of failing the entire transfer due to a single error, the system now logs the exception and proceeds with the remaining steps. This ensures that most of the transfer completes successfully, reducing manual recovery efforts.

### 💭 Notes
Improvements: 
- Does not try to transfer member's project to org if user does not any
- Detects if Asset object has a valid XForm counterpart, flags the transfer as failed.
- Detects if the recipient has already a project with the same id_string (could be possible with old project)  - in `logger_xform` table, flags the transfer as failed

### 👀 Preview steps

Bug template:

**First case:**
1. ℹ️ have an account and two projects with attachments
2. Add several submissions to both
3. From the shell, load the first related asset, find its xForm (`xf = a.deployment.xform`). Delete it. If it does not work from the shell, do it directly in the PostgreSQL with psql
4. Add the user to an MMO
5. 🔴 [on release branch] notice the transfer fails abruptly, the second project is not transferred
6. 🟢 [on PR] notice that the first transfer fails gracefully, the second project is transferred

**Second case:**
1. ℹ️ have an account and one project with attachments
2. Add several submissions
3. From the shell, load the first attachment of the first submission, delete the related file from the storage
4. Transfer the project to another user
5. 🔴 [on release branch] notice the transfer fails abruptly, none of the other attachments are transferred
6. 🟢 [on PR] notice the transfer keeps going, error is logged and all other attachments are transferred

### DIFF
--- a/kobo/apps/long_running_migrations/jobs/0003_transfer_members_data_ownership_to_org.py
+++ b/kobo/apps/long_running_migrations/jobs/0003_transfer_members_data_ownership_to_org.py
@@ -20,4 +20,6 @@ def run():
                 owner_id=OuterRef('pk')
             )
         ).exclude(pk=organization.owner_user_object.pk):
+            if not Asset.objects.filter(owner_id=user.pk).exists():
+                continue
             transfer_member_data_ownership_to_org(user.pk)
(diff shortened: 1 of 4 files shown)


RuthShryock (approved):
LGTM

PR #5726, opened 2025-04-30
fix(submissions): Fix media question parsing bug causing silent data loss with appearance attribute

rajpatel24:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Fixes a bug where media questions (e.g., images) could be silently ignored if their XML attributes were not in a specific order.

### 📖 Description
The previous logic assumed that the `ref` attribute would always appear immediately after the `mediatype` attribute in media upload questions. This assumption caused issues when the attribute order varied. For example, if `appearance` appeared between `mediatype` and `ref`, the parser would fail to detect the ref attribute, resulting in valid media questions being missed and their data silently dropped. This fix makes the parsing logic more flexible by detecting both attributes within the same node regardless of order, preventing silent data loss.

### 👀 Preview steps
1. Add two image questions:
- Leave the first as-is.
- Set the second’s appearance to signature.
2. Deploy the form.
3. Open in Enketo, attach an image to the first, draw a signature for the second.
4. Submit, then check the data table in KPI. Both images should be present.
5. 🔴 [on `release/2.025.02`] notice that the first question's image is present, but the signature is missing
6. 🟢 [on PR] both images should be present.

### DIFF
--- a/kobo/apps/openrosa/apps/logger/xform_instance_parser.py
+++ b/kobo/apps/openrosa/apps/logger/xform_instance_parser.py
@@ -390,35 +390,16 @@ def get_xform_media_question_xpaths(
 ) -> list:
     parser = XFormInstanceParser(xform.xml, xform.data_dictionary(use_cache=True))
     all_attributes = _get_all_attributes(parser.get_root_node())
-    media_field_xpaths = []
-    # This code expects that the attributes from Enketo Express are **always**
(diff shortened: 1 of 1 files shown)


noliveleger (changes requested):

noliveleger:
Closed in favor of #5729

PR #5727, opened 2025-04-30
perf(projectOwnershipTransfer): optimize database queries to retrieve project names faster in Admin UI

noliveleger:
### 📣 Summary
Reduced the number of database queries and response time for listing transferred projects in Admin UI

### 📖 Description
This PR optimizes those queries, significantly reducing database load and improving UI responsiveness — especially in organizations with a large number of projects.

### 👀 Preview steps
Hard to test without a (very) large dataset
Use debug toolbar to see that the number of SQL queries decrease ( `DEBUG_TOOLBAR` env variable in `dev.py`)

snippet to create invites based on an existing one: 

```python
From the shell
i = Invite.objects.last()
transfers = list(i.transfers.all())
for j in range(20000):
     i.pk = None
     i.uid = None
     i.save()
     for t in transfers:
         t.pk = None
         t.uid = None
         t.invite = i
         t.save()
```

Bug template:
1. Open http://kf/admin/project_ownership/invite
2. Look at debug tool bar menu
4. 🔴 [on release branch] notice the number of queries is high
5. 🟢 [on PR] notice the number of queries is low

Locally with 20000, it went down from 200~ish to ~10. With production, the page load went from 78s to 1s.

### DIFF
--- a/kobo/apps/project_ownership/admin.py
+++ b/kobo/apps/project_ownership/admin.py
@@ -1,13 +1,45 @@
+from collections import defaultdict
+
 from django.contrib import admin
 from django.utils.html import linebreaks
 from django.utils.safestring import mark_safe
 
-from .models import Invite
+from .models import Invite, Transfer
+from .models.invite import InviteType
 from .models.transfer import TransferStatusTypeChoices
+
+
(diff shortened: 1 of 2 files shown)


Guitlle (approved):
LGTM

PR #5728, opened 2025-04-30
perf(mongoSync): speed up mongo sync 

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Improve performance of `sync_mongo` job.

### 💭 Notes
One major inefficiency in the sync_mongo job was that every ParsedInstance needed to dict-ify itself in isolation, which requires information that can only be obtained by parsing the XForm. The ParsedInstances had no way of sharing the information from the XForm, even though it was the same every time, so the job ended up parsing the XForm again for every single ParsedInstance, which is very slow for large XForms.

This PR allows creating the dict version of a submission to be done outside the ParsedInstance.save() code, then have the ParsedInstance use the pre-made dict to create the mongo entry. This means that the code for creating the dict version can share information not available inside the ParsedInstance itself. Specifically, we can determine the repeating groups ahead of time, which is the only reason we actually need to parse the XForm.

Eventually, this could use a larger refactor to make it easier to debug. Right now, pi.save() calls a very long chain of methods that goes through Instance, XFormInstanceParser, and DataDictionary, with necessary data cached on different objects and sometimes inaccessible to the calling object. It's conceivable we could get rid of ParsedInstance, XFormInstanceParser, and DataDictionary altogether and just get the information we need from XForm and Instance objects. ParsedInstance and XFormInstanceParser are basically layers on top of Instance, and DataDictionary is basically a layer on top of XForm.

### 👀 Preview steps

On both the release branch and the PR branch:
1. ℹ️ have account and a large project with at least one repeating group (see internal [example](https://chat.kobotoolbox.org/#narrow/stream/4-Kobo-Dev/topic/Inefficency.20of.20ParsedInstance.2Esave.28.29/near/602833))
2. Create about ~10 submissions (should be enough to see the time difference).
3. Note: if you're using the xls provided, the easiest way to make 10 submissions without having to go through a massive form is to select 'PAP type: MST'. This form has been modified from the one originally taken from prod. It may also be faster to just create one or two submissions and then use the `api/v2/assets/uid/data/id/duplicate` endpoint to copy them as what's in the submission is not terribly important
4. Run the Django IPython shell
5. `from django.core.management import call_command`
6. `time call_command('sync_mongo', '-a', '-r', <username>, <project uid>)`
7. :green_circle: the process should be much faster on the PR branch than on main
8. In the mongo container, run `mongoexport --db=formhub --collection=instances --query='{"_xform_id_string": "<uid>"}' --out=/tmp/instances.json` and copy the results locally
9. Once you have copied both .jsons locally, compare them using your favorite diff tool to make sure they are the same

### DIFF
--- a/kobo/apps/openrosa/apps/logger/models/xform.py
+++ b/kobo/apps/openrosa/apps/logger/models/xform.py
@@ -17,8 +17,6 @@ from taggit.managers import TaggableManager
 
 from kobo.apps.kobo_auth.shortcuts import User
 from kobo.apps.openrosa.apps.logger.exceptions import XLSFormError
-from kobo.apps.openrosa.apps.logger.xform_instance_parser import get_abbreviated_xpath, XFormInstanceParser
-from kobo.apps.openrosa.apps.viewer.models import ParsedInstance
 from kobo.apps.openrosa.koboform.pyxform_utils import convert_csv_to_xls
 from kobo.apps.openrosa.libs.constants import (
     CAN_ADD_SUBMISSIONS,
@@ -359,6 +357,3 @@ class XForm(AbstractTimeStampedModel):
                     return convert_csv_to_xls(ff.read())
                 else:
                     return BytesIO(ff.read())
-
-
-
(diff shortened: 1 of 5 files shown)


jnm (approved):
- [x]  Code looks good :eyes: 
- [x] Tested with sample form and 11 submissions:
    - [x] No differences to JSON :heavy_check_mark: 
    - [x] Faster by a factor of 4 :tada:

PR #5730, opened 2025-05-01
fix(trashBin): prevent restarter from restarting tasks still in grace period

noliveleger:
### 📣 Summary
Updated the trash bin restarter to skip tasks that are still within their grace period.

### 📖 Description
The trash bin task restarter was previously restarting tasks even if they were still within their grace period, leading to premature retries. This fix ensures that tasks are only considered for restart if their grace period has expired.


rajpatel24 (approved):
LGTM

PR #5734, opened 2025-05-02
test(storybook): enable a11y tests DEV-95

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 👀 Preview steps

Bug template:
2. do this
3. do that

Feature/no-change template:
2. do this
3. do that
5. do that another thing


pauloamorimbr (approved):
LGTM! 👍🏻 
AFAICT it's all good!

.storybook/test-runner.ts
```diff
@@ -0,0 +1,34 @@
...
+    await injectAxe(page)
+  },
+  async postVisit(page, context) {
+    // Get entire context of a story, including parameters, args, argTypes, etc.
+    const storyContext = await getStoryContext(page, context)
+
+    // Do not test a11y for stories that disable a11y
+    if (storyContext.parameters?.a11y?.test === 'todo') {
+      return
+    }
```
Akuukis, line 16: Docs use `disable` prop, why check `test`?
```
    // Do not run a11y tests on disabled stories.
    if (storyContext.parameters?.a11y?.disable) {
      return;
    }
```

Akuukis (changes requested):
- [ ] CI [job](https://github.com/kobotoolbox/kpi/actions/runs/14836077035/job/41647564191?pr=5734) has a lot of "No accessibility violations detected!" warnings, perhaps it doesn't contain any errors and the TODO should be removed?
- [ ] [docs say](https://storybook.js.org/docs/writing-tests/accessibility-testing#a11y-config-with-the-test-runner) that "todo" will "Run accessibility tests; violations return [..] a summary count in CLI/CI" but I don't see it in the CI output. Imho that's related to my inline comment.

.storybook/test-runner.ts
magicznyleszek: I was following this page: https://storybook.js.org/docs/writing-tests/accessibility-testing#configure-accessibility-tests-with-the-test-addon which suggests adding `parameters: { a11y: { test: 'todo' } },`. But the ignored (todo) stories were actually not being ignored when I run a11y tests. Turned out that parts of the docs are referring to storybook v9 (which is in beta now), and parts to v8.

I got some help from storybook devs: https://github.com/storybookjs/storybook/discussions/31368 (let's keep in mind that they are responsive!).

I thought it would be ok to leave those `test: 'todo'` instead of using some `manual: true` or other way. "todo" feels more meaningful :)

magicznyleszek: I hid the "No accessibility violations detected!" with `verbose: false`. The errors would still appear correctly.

Akuukis: lets try to upgrade :crossed_fingers:

Akuukis (changes requested):
code LGTM, clicked around storybook and all works. :tada: Other problems:
- [CI job "Playwright: get install version" fails](https://github.com/kobotoolbox/kpi/actions/runs/15062178258/job/42339218623?pr=5734#step:14:9), and the next step caches it wrong
-  [CI job that tests stories](https://github.com/kobotoolbox/kpi/actions/runs/15062178258/job/42339218623?pr=5734#step:19:612) outputs `error: no kpi-root-path meta tag set. defaulting to ""`, can you look into it?

magicznyleszek:
@Akuukis I've found different way to get playwright version that now works correctly. I also added some code to avoid that kpi-root-path error from happening within storybook context

magicznyleszek:
This was helpful: https://github.com/microsoft/playwright/issues/7249

Akuukis (changes requested):
- caching still doesn't work right, version variable is empty, see logs.
- why take version from package.json instead of lockfile?

Akuukis (approved):
LGTM :tada: I run the CI again and saw a cache hit :ok:

PR #5736, opened 2025-05-05
fix(permissions): sharing modal correctly reflects asset type in permissions labels DEV-255

duvld:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Fixes a bug where the sharing modals will always display "form" or "project" regardless of asset type. Now the modal correctly references the asset being shared.

### 👀 Preview steps

1. Create a project, and a few library items:
  - a single question
  - a question block
  - a template
  - a collection 
3. Open the sharing modal for each of the above
4. 🔴 [on main] notice that the permissions displayed is not correctly reflected based on the type of asset being shared
5. 🟢 [on PR] notice that the permissions displayed change according to the asset type being shared


jsapp/js/components/permissions/userAssetPermsEditor.component.tsx
```diff
@@ -386,12 +392,20 @@ export default class UserAssetPermsEditor extends React.Component<
     // much code to make it perfect
     const disabledPropName = (checkboxName + CHECKBOX_DISABLED_SUFFIX) as keyof UserAssetPermsEditorState
     const isDisabled = Boolean(this.state[disabledPropName])
+
+    let checkboxLabel = ''
+    if (this.props.assetType !== AssetTypeName.survey) {
+      checkboxLabel = getFriendlyPermLabelWithSuffix(this.props.assetType, undefined, checkboxName)
+    } else {
+      checkboxLabel = CHECKBOX_LABELS[checkboxName]
```
magicznyleszek, line 400: ~~I wonder if we shouldn't mimic the other place, and use `checkboxLabel = getFriendlyPermName(CHECKBOX_PERM_PAIRS[checkboxName])`. This way~~ we could make another helper function to avoid duplicating that if…

So it would then become something like

```ts
const checkboxLabel = getContextualPermLabel(
  this.props.assetType,
  // this gets permission codename
  permConfig.getPermissionByCodename(CHECKBOX_PERM_PAIRS[checkboxName])
)
```

And in `userPermissionRow.component.tsx` it would become:
```ts
const friendlyPermName = getContextualPermLabel(
  this.props.assetType,
  perm
)
```

magicznyleszek: The new function should operate on `AssetTypeName` and `PermissionCodename`.

To get permissionwhen you only have permission codename (or url) (like in `userAssetPermsEditor.component.tsx`), you need to use `permConfig.getPermissionByCodename`

magicznyleszek: And then, that `getContextualPermLabel` (not sure about name) function would be something like:

```ts
/**
 * Returns human friendly label of a permission, taking asset type into account
 */
export function getContextualPermLabel(
  assetType: AssetTypeName,
  permCodename: PermissionCodename
) {
  if (assetType !== AssetTypeName.survey) {
    return getFriendlyPermLabelWithSuffix(this.props.assetType, permCodename)
  } else {
    return getFriendlyPermName(permCodename)
  }
}
```

jsapp/js/components/permissions/utils.ts
```diff
@@ -460,6 +461,66 @@ export function getPermLabel(perm: PermissionResponse) {
...
 
+/**
+ * Returns the permission label with the correct suffix depending on the given asset type.
+ *
+ * Example: if we are sharing a library collection, the permissions will all say "[View, Edit, Manage] collection"
+ */
+export function getFriendlyPermLabelWithSuffix(
+  assetType: AssetTypeName | undefined,
+  perm?: PermissionResponse,
+  checkboxNameParam?: CheckboxNameAll,
```
magicznyleszek, line 472: With the proposition in comments above, I think we could only use `assetType` and `perm` here, right? Because all the places that will use this function would pass `PermissionResponse`.

jsapp/js/components/permissions/utils.ts
```diff
@@ -460,6 +461,66 @@ export function getPermLabel(perm: PermissionResponse) {
...
+    checkboxNameAll = checkboxNameParam
+  }
+
+  if (checkboxNameAll === 'formView' || checkboxNameAll === 'formEdit' || checkboxNameAll === 'formManage') {
+    switch (assetType) {
+      case AssetTypeName.block:
+        if (checkboxNameAll === 'formView') checkboxNameSuffix = t('View block')
+        if (checkboxNameAll === 'formEdit') checkboxNameSuffix = t('Edit block')
+        if (checkboxNameAll === 'formManage') checkboxNameSuffix = t('Manage block')
+        break
```
magicznyleszek, line 495: If you would use `return` in each line, you don't need `break`:

```ts
        if (checkboxNameAll === 'formManage') return t('Manage block')
```

jsapp/js/components/permissions/utils.ts
```diff
@@ -460,6 +461,66 @@ export function getPermLabel(perm: PermissionResponse) {
...
+      case AssetTypeName.collection:
+        if (checkboxNameAll === 'formView') checkboxNameSuffix = t('View collection')
+        if (checkboxNameAll === 'formEdit') checkboxNameSuffix = t('Edit collection')
+        if (checkboxNameAll === 'formManage') checkboxNameSuffix = t('Manage collection')
```
magicznyleszek, line 499: I like that we are returning whole strings (for translators) :heart:

jsapp/js/components/permissions/utils.ts
magicznyleszek: And then we don't need `checkboxNameSuffix` variable, and we can rename `checkboxNameAll` (confusing) to `checkboxName`

jsapp/js/components/permissions/utils.ts
```diff
@@ -460,6 +461,66 @@ export function getPermLabel(perm: PermissionResponse) {
...
+      case AssetTypeName.question:
+        if (checkboxNameAll === 'formView') checkboxNameSuffix = t('View question')
+        if (checkboxNameAll === 'formEdit') checkboxNameSuffix = t('Edit question')
+        if (checkboxNameAll === 'formManage') checkboxNameSuffix = t('Manage question')
+        break
+      case AssetTypeName.survey:
+        if (checkboxNameAll === 'formManage') {
+          checkboxNameSuffix = 'project'
+        } else {
+          checkboxNameSuffix = 'form'
```
magicznyleszek, line 515: And I am bit confused here… For `collection` etc. we are returning "View collection", but for `survey` we are returning "project" instead of "View project" (and "form" instead of "View form"?

magicznyleszek (changes requested):

jsapp/js/components/permissions/utils.ts
duvld: I like doing it this way because it removes the confusing `...Suffix` variable, but adding return to each line still gets typescript to complain about fallthrough. In order to not have a `break` be necessary the code would look like this:
```typescript
        if (checkboxName === 'formView') return t('View block')
        else if (checkboxName === 'formEdit') return t('Edit block')
        else return t('Manage block')
```

which I think is worse than a `break` at the end, let me know what you think :)

duvld: The compromise I made: https://github.com/kobotoolbox/kpi/pull/5736/commits/64268fa3de1f10cc0658bbd77063426c8fbc6ca1

jsapp/js/components/permissions/userAssetPermsEditor.component.tsx
duvld: I think we're both confused here :P I agree we should have a helper function that simplifies the logic, but I think your example doesn't exactly work. `getFriendlyPermName` takes `PermissionResponse` as a parameter, not `PermissionCodename`. 

`permConfig.getPermissionByCodename(CHECKBOX_PERM_PAIRS[checkboxName])` returns a `PermissionDefinition`, which is slightly different from `PermissionResponse`, so it cannot be used as a param here either.

Here's how it looks like these are derived from each other:

`PermissionResponse` can be used to get `PermissionCodename` via `getCheckboxNameByPermisson()`
`PermissionCodename` can be used to get `PermissionDefinition` via `permConfig.getPermissionByCodename(CHECKBOX_PERM_PAIRS[checkboxName])`

Both components can use `PermissionCodename` to bridge this gap, but then we have an issue with the new `getContextualPermLabel()`, which needs `PermissionResponse` to use `getFriendlyPermName()`.

TL:DR, the two components use different logic to display the checkboxes, which (at least I have found) to be

duvld: I also may be overlooking a simple link that ties them all together :bow:

jsapp/js/components/permissions/utils.ts
duvld: Commented on the propositions above, but they still yielded simpler paramters :)

magicznyleszek (changes requested):
When testing this I discovered a new bug (also happens on kf.beta) - https://linear.app/kobotoolbox/issue/DEV-266/unable-to-uncheck-view-checkbox-when-sharing-library-item - so If you're looking for more work after this, please fix it :)

jsapp/js/components/permissions/utils.ts
```diff
@@ -460,6 +461,42 @@ export function getPermLabel(perm: PermissionResponse) {
   return '???'
 }
 
+/**
+ * Returns the permission label with the correct suffix depending on the given asset type.
```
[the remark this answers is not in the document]
duvld: I think handling surveys in this function would be prettier for this function, but a lot of code depends on the default label being the survey suffix so I think leaving a comment is better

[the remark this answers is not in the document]
duvld: I still compensated for this by adding a default return being `CHECKBOX_LABELS[checkboxName]` which would return the unchanged survey suffix

jsapp/js/components/permissions/utils.ts
```diff
@@ -460,6 +461,42 @@ export function getPermLabel(perm: PermissionResponse) {
...
+export function getContextualPermLabel(
+  assetType: AssetTypeName | undefined,
+  checkboxName: CheckboxNameAll | undefined,
+) {
+  if (checkboxName === 'formView' || checkboxName === 'formEdit' || checkboxName === 'formManage') {
+    switch (assetType) {
+      case AssetTypeName.block:
+        if (checkboxName === 'formView') return t('View block')
+        else if (checkboxName === 'formEdit') return t('Edit block')
+        else return t('Manage block')
```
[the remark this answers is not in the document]
duvld: Did it for all the different types, it makes more constants but make this function a lot prettier, let me know if that looks like what you intended :)

jsapp/js/components/permissions/sharingForm.component.tsx
```diff
@@ -273,6 +275,7 @@ export default class SharingForm extends React.Component<SharingFormProps, Shari
 
               <UserAssetPermsEditor
                 assetUid={this.props.assetUid}
+                assetType={assetType}
```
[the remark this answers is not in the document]
duvld: oops! done now

jsapp/js/components/permissions/permConstants.ts
```diff
@@ -247,6 +247,34 @@ export const CHECKBOX_LABELS: { [key in CheckboxNameAll]: string } = {
 }
 Object.freeze(CHECKBOX_LABELS)
 
+export const CHECKBOX_LABELS_BLOCK: { [key in 'formView' | 'formEdit' | 'formManage']: string } = {
```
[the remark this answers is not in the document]
duvld: I was wondering if we should have a slightly redundant definition like that, but I like this solution

PR #5737, opened 2025-05-05
fix(permissions): hide `discover_asset` in the UI DEV-257

duvld:
### 📣 Summary

Fixes bug that displayed ??? in the collection sharing modal when giving a user Manage project permissions.

### 💭 Notes
Sharing modal was confused and displayed ??? for the `discover_asset` permission for collections. There are enough indicators for if a collection is public so we don't display anything for this permission.

### 👀 Preview steps

Bug template:
1. ℹ️ Create a collection
2. Share collection to another user
3. Give that user "Manage project" permissions
5. 🔴 [on main] notice that in the permissions row there is a ??? permission among the other assigned ones
6. 🟢 [on PR] notice that ??? is gone


duvld:
https://linear.app/kobotoolbox/issue/DEV-256/sharing-modal-for-a-collection-shows-next-to-a-users-permissions

PR #5740, opened 2025-05-06
fix(massEmails): fix 500 from usage queries

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Fixes a 500 error when trying to add usage-query-dependent emails to the daily send.

### 💭 Notes
The usage queries were returning a list of uids rather than a list of users. We didn't catch this earlier because we usually tested with the inactive user query, which was easier to manipulate and returned a list of users. The unit tests also focused on the base method, `get_users_within_range_of_usage_limit`, where most of the logic actually happened.

### 👀 Preview steps

Bug template:
1. ℹ️ have 2 user accounts each with a project with an audio question
2. ℹ️  make sure stripe is enabled. Both users should be on the community plan
3.  ℹ️ enable NLP
4. You may want to configure your environment to use `EMAIL_BACKEND: 'django.core.mail.backends.filebased.EmailBackend'` to more easily see which emails are sent
5. set MASS_EMAILS_CONDENSE_SEND to True either by configuring your build environment or manually updating kobo/settings/base.py locally
6. In django admin, update the Community product metadata with `"submission_limit": "10", "nlp_character_limit": "10", "storage_bytes_limit": "1030700"`
7. In django admin, create a MassEmailConfig for each query *except* users_inactive_for_365_days
8. For user1/project1, create 9 submissions using [sample-6s-14_44_13.zip](https://github.com/user-attachments/files/20071865/sample-6s-14_44_13.zip) as the audio sample (unzipped)
9. For one of these submissions, add a 9-letter word as a transcription (eg "pineapple") and get an automated translation into whatever language
10. For user2/project2, create 10 submissions using the same sample
11. For one of these submissions, add a 9-letter word as a transcription (eg "manticores") and get an automated translation into whatever language
12. In django admin, add each MassEmailConfig to the daily send (go one by one rather than in bulk)
13. :red_circle: [on main] Each attempt will result in a 500
14. :green_circle: [on PR] Each attempt should succeed
15. :green_circle: [on PR] at the next 15 minute interval, user1 should receive all the emails for >90% usage. user2 should receive all emails for the >100% usage. Each user should receive 3 emails.

### DIFF
--- a/kobo/apps/mass_emails/user_queries.py
+++ b/kobo/apps/mass_emails/user_queries.py
@@ -139,41 +139,33 @@ def get_users_within_range_of_usage_limit(
     return User.objects.filter(id__in=user_ids)
 
 
-def get_users_over_90_percent_of_storage_limit():
-    results = get_users_within_range_of_usage_limit(
+def get_users_over_90_percent_of_storage_limit() -> QuerySet:
+    return get_users_within_range_of_usage_limit(
         usage_types=['storage'], minimum=0.9, maximum=1
     )
-    return [user.extra_details.uid for user in results]
 
 
-def get_users_over_100_percent_of_storage_limit():
-    results = get_users_within_range_of_usage_limit(usage_types=['storage'], minimum=1)
-    return [user.extra_details.uid for user in results]
+def get_users_over_100_percent_of_storage_limit() -> QuerySet:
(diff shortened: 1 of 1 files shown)


Guitlle (approved):

sksondra (commented):
Good from QA.

PR #5741, opened 2025-05-06
fix(projectTransfer): reset legacy project identifier to KPI identifier on transfer conflict DEV-253

noliveleger:
### 📣 Summary
Resolved identifier collisions when transferring projects between users with legacy id_string values.

### 📖 Description
When transferring a project, a conflict may occur if both the sender and the recipient have a project with the same legacy `id_string`, due to the uniqueness constraint of `id_string` per user.

This PR detects such collisions and resolves them by resetting the `id_string` of the project to match its unique KPI identifier (`Asset.uid`) before the transfer proceeds. This allows the receiving user to retain their version without disruption, while the legacy project adopts a safe, unique identifier.

⚠️ Note: Historical data will continue to use the original legacy id_string as the root node in their XML submissions.

### 👀 Preview steps

Bug template:
1. Modify [this line](https://github.com/kobotoolbox/kpi/blob/c7641726228cac5869c01ba14d353bd667836320/kobo/apps/openrosa/apps/logger/models/xform.py#L313) to assign a random string, e.g. `self.id_string = 'foo'`
2. Log in as userA 
3. Create a project (ProjectA) and deploy it
4. Log in as userB
5. Create a project (ProjectB) and deploy it, add submissions (with pictures - not required but better for testing). Keep note of the EnketoID
6. Validate in DB you have two XForm with same `id_string` (but different users)
7. Undo the modification applied in `#1` 
8. Transfer ProjectB to userA
9. 🔴 [on release branch] transfer should fail
10. 🟢 [on PR] 
   - **DO NOT FORGET** to restart your Celery worker containers
   - transfer should succeed
   - userA should be the owner of ProjectB 
   - userA should see the data and attachments
   - The Enketo ID should remain the same as before the transfer

11. Repeat step 2 to 8 to make sure regular projects still transfer flawlessly

### DIFF
--- a/kobo/apps/project_ownership/models/transfer.py
+++ b/kobo/apps/project_ownership/models/transfer.py
@@ -143,7 +143,7 @@ class Transfer(AbstractTimeStampedModel):
                         try:
                             xform = deployment.xform
                         except (InvalidXFormException, MissingXFormException):
-                            id_string = deployment.backend_response.get('id_string')
+                            id_string = deployment.xform_id_string
                             self.status = (
                                 TransferStatusChoices.FAILED,
                                 f'Project `{self.asset.uid}` could not be transferred '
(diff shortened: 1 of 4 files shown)


rajpatel24 (approved):
LGTM

PR #5742, opened 2025-05-07
docs(api): add README for API documentation setup and usage

noliveleger:
### 📣 Summary
Added a README file explaining how the API documentation system works.

### 📖 Description
This PR introduces a README.md file dedicated to the API documentation system. It explains how the documentation is generated, how to preview it locally, and how developers can contribute to documenting new or existing endpoints.

The README is primarily focused on the KPI v2 API, which is now being progressively documented.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -38,7 +38,7 @@ requests and responses.
 
 ---
 
-## @extend_schema
+## @extend\_schema
 
 The `@extend_schema` decorator is used to enrich individual view methods or custom
 actions with additional metadata for documentation purposes. It allows you to manually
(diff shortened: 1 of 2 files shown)


docs/README.md
```diff
@@ -0,0 +1,302 @@
...
+These serializers are useful for defining lightweight, one-off response schemas or for
+custom endpoints with specific structure requirements.
+
+Example:
+
+```python
+CategoryListInlineSerializer = inline_serializer(
+    name='CategoryListInlineSerializer',
+    fields={
+        'url': serializers.URLField(),
```
ZoeGodin, line 146: It might be better to remove ```serializers.URLField()``` since it doesn’t actually work in this context and could mislead users into thinking it does.

PR #5747, opened 2025-05-09
refactor(trashBin): unify status toggling inside `move_to_trash()` and `put_back()` for all trash models TASK-1826

rajpatel24:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Simplified the trash process for attachments, projects, and users by consolidating the status toggle into a single step.

### 📖 Description
Previously, moving items (attachments, projects, or user accounts) to the trash required calling both `toggle_statuses()` and `move_to_trash()` separately. This update refactors `move_to_trash()` to internally handle the status change, unifying the logic across all trash models.

### 👀 Preview steps
This is a refactoring-only change, so functionality should remain unchanged. To verify:
1. Move a Project, Attachment, and User to the trash.
2. Restore each of them from the trash.
3. Confirm that behavior matches what's currently on the main branch.
No change in user experience is expected.


kobo/apps/trash_bin/utils/trash.py
```diff
@@ -221,7 +231,13 @@ def put_back(
         - `'attachment_uid'`
     """
 
-    trash_model, fk_field_name, related_model, *others = _get_settings(trash_type)
+    trash_model, fk_field_name, related_model, unique_identifier, *others = (
```
noliveleger, line 234: I cannot comment on line 213, but let's use typing annotations to specify the return the value.
Also, what about using typing aliases to specify the type of `objects_list` too?

noliveleger (approved):
LGTM

PR #5748, opened 2025-05-09
refactor(billing): standardize usage naming TASK-1827

jamesrkiger:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Switches to standardized choices for referring to usage types in backend code.

### 📖 Description
For the usage limits project we will be adding a model that includes a field for storing a usage type. Usage types are very stable. There are only four choices and any future changes will only entail the addition of further choices. As such, it makes the most sense to use a character choice field for this purpose. (Note: I had considered using an Integer choice field, but because we use the string values so frequently and the table we will be adding the choice field on will not be particularly big, I decided against the added complexity).

However, in the code we currently have two almost, but not quite identical ways of writing usage types, reflected in our `USAGE_LIMIT_MAP`

```python
USAGE_LIMIT_MAP = {
    'characters': 'mt_characters',
    'seconds': 'asr_seconds',
    'storage': 'storage_bytes',
    'submission': 'submission',
}
```

Aside from saving a few keystrokes in the backend code when typing a usage type (an advantage erased by having to use this mapping across the codebase), there doesn’t seem to be any reason for this divergence. 

This PR creates a django TextChoices enum using the values from the above mapping and replaces all instances of keys from the above mapping accordingly.

### 👀 Preview steps
The affected strings are only used in internal code. The API isn't impacted. So existing unit test coverage should be sufficient here.

### DIFF
--- a/kobo/apps/stripe/tests/test_one_time_addons_api.py
+++ b/kobo/apps/stripe/tests/test_one_time_addons_api.py
@@ -6,7 +6,6 @@ from rest_framework import status
 
 from kobo.apps.kobo_auth.shortcuts import User
 from kobo.apps.organizations.models import Organization
-from kobo.apps.stripe.constants import USAGE_LIMIT_MAP
 from kobo.apps.stripe.models import PlanAddOn
 from kobo.apps.stripe.tests.utils import _create_one_time_addon_product, _create_payment
 from kpi.tests.kpi_test_case import BaseTestCase
@@ -124,10 +123,10 @@ class OneTimeAddOnAPITestCase(BaseTestCase):
         assert response_get_list.status_code == status.HTTP_200_OK
         assert response_get_list.data['results'] == []
 
-    @data('characters', 'seconds')
+    @data('mt_characters', 'asr_seconds')
     def test_get_user_totals(self, usage_type):
         limit = 2000
-        usage_limit_key = f'{USAGE_LIMIT_MAP[usage_type]}_limit'
+        usage_limit_key = f'{usage_type}_limit'
         self._create_product(
             metadata={
                 'product_type': 'addon_onetime',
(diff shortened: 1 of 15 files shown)


kobo/apps/stripe/tests/test_one_time_addons_api.py
rgraber, line 1: Is there a reason you didn't want to use UsageType.CHARACTERS, etc. in this file?

kobo/apps/stripe/utils.py
```diff
@@ -381,30 +381,29 @@ def get_organizations_subscription_limits(
         if row['product_type'] == 'plan':
             row_limits = {
                 f'{usage_type}_limit': row[f'{usage_type}_limit']
-                for usage_type in get_args(UsageType)
+                for usage_type, _ in UsageType.choices
             }
         elif row['product_type'] == 'addon':
-            row_limits['addon_storage_limit'] = row['storage_limit']
+            row_limits['addon_storage_limit'] = row['storage_bytes_limit']
```
rgraber, line 387:
```suggestion
            row_limits['addon_storage_limit'] = row[f'{UsageType.STORAGE_BYTES}_limit']
```

kpi/tests/test_usage_calculator.py
rgraber, line 1: Same comment as before, can we use the UsageType constants here?

kpi/utils/usage_calculator.py
```diff
@@ -88,8 +89,8 @@ def get_nlp_usage_in_date_range_by_user_id(date_ranges_by_user) -> dict[int, NLP
     results = {}
     for row in nlp_tracking:
         results[row['user_id']] = {
-            'seconds': row['asr_seconds_current_period'],
-            'characters': row['mt_characters_current_period'],
+            UsageType.ASR_SECONDS: row['asr_seconds_current_period'],
```
rgraber, line 92:
```suggestion
            UsageType.ASR_SECONDS: row[f'{UsageType.ASR_SECONDS}_current_period'],
```

rgraber (changes requested):
Some other places to use the constants

PR #5752, opened 2025-05-09
fix(storage): Use correct database when no options are passed to update_attachment_storage_bytes command

noliveleger:
### 📣 Summary
Ensured the `update_attachment_storage_bytes` management command uses the appropriate database when no options are provided.

### 👀 Preview steps

Bug template:
1. Use the management command without options
4. 🔴 [on release] Management command crashes
5. 🟢 [on PR] Management command runs


rajpatel24 (approved):
LGTM

PR #5753, opened 2025-05-12
fix(openRosa): list anonymous projects manageable by user

noliveleger:
### 📣 Summary
Fixed logic to return projects that allow anonymous submissions and are manageable by the user.

### 📖 Description
Previously, the OpenRosa API used the project owner's account to list all projects that allowed anonymous submissions. After account unification into organizations, this resulted in a list that was too large to handle, causing 502/504 errors in KoboCollect and making devices difficult to maintain since they required manual reconfiguration.

With this PR:
- Only projects that allow anonymous submissions and that the user has manage_asset permission on are returned.
- The behavior remains transparent for users — devices do not need to be reconfigured.

### Preview steps

1. ℹ️ have account (`user_a`) and a few projects
2. Activate Allow submissions for anonymous for all of them
3. Add this account to a MMO
4. Go to http://kf.kobo.local/user_a/formList
4. 🔴 [on release branch] Response is empty
6. 🟢 [on this PR] Projects should be listed


Guitlle (approved):
LGTM

PR #5755, opened 2025-05-13
fix:(reports): Fix infinite reloading failed report DEV-265

pauloamorimbr:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
This PR fixes a bug where under a certain condition loading a report containing a Group By configured fails and the app tries to reload it infinitely.

### 👀 Preview steps
Bug template:

1. Have at least 1 select one type question in the form
2. Have a number input in the form
3. Generate multiple submissions
4. Generate report
5. Setup a Group By for the report (in report settings)
6. Report will display normally
7. Open the data and bulk edit the submissions to insert an invalid value into the number field
8. Try to load the report
9. Error should happen

### DIFF
--- a/jsapp/js/components/reports/reports.tsx
+++ b/jsapp/js/components/reports/reports.tsx
@@ -180,7 +180,7 @@ export default class Reports extends React.Component<ReportsProps, ReportsState>
                 ),
               )
 
-              // reset default report groupBy if it fails and notify user
+              // reset default report groupBy if it fails
               reportStyles.default.groupDataBy = undefined
               this.setState({ reportStyles: reportStyles }, () => {
                 // Retry loading report data without groupBy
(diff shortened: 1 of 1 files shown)


jsapp/js/components/reports/reports.tsx
```diff
@@ -173,16 +173,19 @@ export default class Reports extends React.Component<ReportsProps, ReportsState>
...
-              this.setState({ reportStyles: reportStyles })
               notify.error(
                 t('Could not load grouped results via "##". Will attempt to load the ungrouped report.').replace(
                   '##',
                   groupBy,
                 ),
               )
-              this.loadReportData()
+
+              // reset default report groupBy if it fails and notify user
```
magicznyleszek, line 183: A nitpick, but this comment says "…and notify user", but `notify.error(` is being called above already. So either edit the comment or move or split into two :)

magicznyleszek (approved):

PR #5757, opened 2025-05-13
docs(api): Cleaned urls and objects of current endpoints TASK-1724

ZoeGodin:
### 📣 Summary
Cleaned URLs and object builders of current endpoints

### 📖 Description
Created an url_type_builder to better organize the code and changed the way the object extensions were built.

### 💭 Notes
Needs to be merges after #5749


noliveleger (approved):
LGTM

PR #5759, opened 2025-05-13
docs(api):  payload format validation - TASK-1724

ZoeGodin:
### 📣 Summary
Added payload validation so user can see only right format is JSON.

### DIFF
--- a/kpi/views/v2/asset_snapshot.py
+++ b/kpi/views/v2/asset_snapshot.py
@@ -66,7 +66,7 @@ from kpi.views.v2.open_rosa import OpenRosaViewSetMixin  # noqa
     # description for post
     create=extend_schema(
         description=read_md('kpi', 'asset_snapshots/create.md'),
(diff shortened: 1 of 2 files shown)


kpi/views/v2/asset_snapshot.py
```diff
@@ -66,7 +66,7 @@
     # description for post
     create=extend_schema(
         description=read_md('kpi', 'asset_snapshots/create.md'),
-        request=AssetSnapshotCreateRequestInlineSerializer,
+        request={'application/json': AssetSnapshotCreateRequestInlineSerializer},
```
noliveleger, line 69: Is this because the response could JSON or XML (but the request only in JSON?

ZoeGodin: This was because we can only send json in the create and the schema somehow detected that we could both send json and xml

noliveleger (approved):
LGTM

PR #5765, opened 2025-05-14
feat(removingAttachments): use API instead of mocks TASK-1685

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Remove mock code that was operating single attachment deletion button and bulk attachment deletion modal. Use ready APIs instead.

### 💭 Notes

Changes here:
- Cleaned attachments related up endpoints (removed unused, updated existing ones)
- Fixed bug in `fetchData` (from `api.ts`) - it was giving false failures when API call was successfull but without content (basically any `204` would appear as failed)
- For `BulkDeleteMediaFiles` modal:
  - Use `meta/rootUuid`s (with necessary utility being used)
  - Use `onChange` for checkbox to avoid console error
  - Fixed a bug where you check checkbox and close modal with "Cancel" (checkbox wasn't being reset to unchecked)

### 👀 Preview steps

Testing:
1. ℹ️ have a project with `image` question and multiple submissions
3. find a cell with image response and click on image icon button
4. 🟢 notice image renders properly in the media modal
5. use "…" icon button to open dropdown and select "Delete"
6. confirm "Delete"
7. 🟢 notice that media modal closes and success notification appears
8. open dev tools on "Network" tab
9. open media modal again (image icon button)
10. 🟢 notice infinite spinner instead of image; in dev tools notice `404` response for getting the image

The reason we have infinite spinner and `404` is becaused FE code expects `is_deleted` flag to be present on deleted attachments, and this code is not yet done on BE :ok:


duvld:
I don't remember where we landed on the "Open" button for audio attachments (that brings you to the single processing view). Did we want to leave it even though the corresponding audio attachment is deleted? If so, we should make it more obvious that the audio is deleted: the play button is still present before or after a hard refresh (in the table view)

duvld:
Otherwise LGTM, all the requests are sending the right payload

magicznyleszek:
All the issues found are connected to `is_deleted` flag not being implemented yet

sksondra (commented):
Good from QA.

PR #5768, opened 2025-05-14
fix(auth): improve validation on user profile updates

noliveleger:
### 📣 Summary
Improved validation logic for user profile updates.


kpi/tests/api/test_api_current_user.py
```diff
@@ -147,3 +147,15 @@ def test_validated_password_becomes_true_on_password_change(self):
     def test_cannot_update_organization_fields_with_mmo(self):
         self.user.organization.mmo_override = True
         self.user.organization.save()
+
+    def test_cannot_change_statuses(self):
```
jnm, line 151: verified that this fails without the serializer change :heavy_check_mark:

jnm: CI tests pass; only failure is unrelated, known billing date issue

jnm (approved):

PR #5772, opened 2025-05-15
fix(longRunningMigration): do not block fresh installs with system checks DEV-272

noliveleger:
### 📣 Summary
Prevented system checks from raising errors on fresh installs when long-running migrations haven't been executed.

### 📖 Description
System checks for long-running migrations were incorrectly raising errors even on fresh installs, where no data exists and migrations are not required. This caused unnecessary blocking during setup or deployment of a new environment.

### 👀 Preview steps

1. Create a fresh install with kobo-install
2. 🔴 [on main] notice that the KPI container will crash on its first attempt to connect to the database
3. Stop containers
4. Delete kobo-docker, kobo-env folders
5. In KPI folder, checkout this branch
6. run kobo-install setup again (`--setup`)
7. starts containers if they are not started yet
7. 🟢 [on PR] notice that the KPI container starts and kobo-install displays the message "Postgres is up and running"

### DIFF
--- a/kobo/apps/long_running_migrations/app.py
+++ b/kobo/apps/long_running_migrations/app.py
@@ -1,6 +1,7 @@
 from django.apps import AppConfig
-from django.core.checks import Error, register, Tags
+from django.core.checks import Error, Tags, register
 from django.db import connection
+from django.db.utils import OperationalError, ProgrammingError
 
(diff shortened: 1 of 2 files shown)


duvld:
LGTM

duvld (approved):

PR #5778, opened 2025-05-19
fix(billing): direct storage addon holders to Stripe for upgrade DEV-275

jamesrkiger:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Quick fix for bug that made upgrading storage addons overly complicated. Always direct users with storage addons to Stripe account management instead of attempting to purchase new addon directly. 

### 💭 Notes
1. This is a workaround until we get a better system in place. Ideally, users with an existing storage addon should be able to select a new addon, click the button, and only be directed to Stripe to confirm change. But currently attempting to purchase a new addon results in the user being prompted by Stripe to confirm their email address, after which they are directed to Stripe account management. So we are just cutting out that email confirmation to make it a less annoying.

2. I renamed the oneTimeAddOnRow component because we use it for both recurring and one time addons.

3. I did a little refactoring to make the jsx template easier to read.

Bug template:
1. As a new user in a Stripe-enabled environment, purchase a 5gb storage addon and sync stripe
2. Navigate to addons page.
3. On `release/2.025.10`, observe that the button on the storage addon row says "Buy now" unless you use the select to pick your current storage plan. If you click on "Buy now" you are directed to Stripe, which asks you to confirm your email
4. On this branch, observe that the button always says "Manage" and clicking on it takes you to your Stripe account management page
5. Purchase an NLP addon
6. Notice that the behavior of the NLP addons row on the addons page is unchanged between this branch and `release/2.025.10`.

### DIFF
--- a/jsapp/js/account/addOns/addOnList.component.tsx
+++ b/jsapp/js/account/addOns/addOnList.component.tsx
@@ -1,6 +1,6 @@
 import React, { useContext, useEffect, useState } from 'react'
 
-import { OneTimeAddOnRow } from '#/account/addOns/oneTimeAddOnRow.component'
+import { AddOnProductRow } from '#/account/addOns/addOnProductRow.component'
 import type { Organization } from '#/account/organization/organizationQuery'
 import type { OneTimeAddOn, Price, Product, SubscriptionInfo } from '#/account/stripe.types'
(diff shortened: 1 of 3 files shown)


pauloamorimbr (approved):
LGTM! 💯

PR #5785, opened 2025-05-20
fix(permissions): make sharing form request permissions upon opening DEV-128

duvld:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Fixes issue where wharing modal did not update permissions until a hard refresh.

### 💭 Notes

Originally thought of as a caching issue, but that's not the case. The modal is just not requesting permissions on open.

One liner but that raises some questions. It feels like something we would have made a decision to not do in the past since it's a simple get request that would make sense to be there. I tried looking in the commit history to see if this line was ever removed but didn't find anything 🤷

A better solution would use a store, or react query, but until then this modal should be updating correctly

### 👀 Preview steps

Bug template:
1. ℹ️ have multiple accounts and an asset
2. Open the sharing modal anywhere
3. Grant some permissions to a user (can really be any permission change, but these steps follow adding a user)
4. Close the modal
5. Open the modal again
6. 🔴 [on main] notice that the permissions are not updated and the new user is not in the modal
7. 🟢 [on PR] notice that the permissions are updated and the new user is in the modal on after closing and opening the sharing modal again


jsapp/js/components/permissions/sharingForm.component.tsx
```diff
@@ -81,6 +81,8 @@ export default class SharingForm extends React.Component<SharingFormProps, Shari
     )
     if (this.props.assetUid) {
       actions.resources.loadAsset({ id: this.props.assetUid })
+      // Update permissions on modal open
+      actions.permissions.getAssetPermissions(this.props.assetUid)
```
magicznyleszek, line 85: AFAIK the thing we get from `actions.permissions.getAssetPermissions` (which is using `/api/v2/assets/<uid>/permission-assignments/` endpoint) is the same thing we get from `actions.resources.loadAsset` (which is using `/api/v2/assets/<uid>/`) inside `permissions` prop.

magicznyleszek: Does it mean that this bug happens because of the `assetCache` from `actions.js`? I.e. we are getting cached asset after permissions were changed

magicznyleszek: Maybe a bit better fix would be `actions.resources.loadAsset({ id: this.props.assetUid }, true)`?

magicznyleszek (approved):

sksondra (commented):
Good from QA.

Minor note: the "Team Members" section on the project summary page does not change unless the modal is opened a second time, or the page is refreshed. The modal itself works as intended.
image

PR #5788, opened 2025-05-21
feat(attachments): Update mongo documents to expose attachment deleted flag TASK-581

rajpatel24:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Added `is_deleted` flag to attachments in Mongo to reflect deletion status after attachment trash or restore actions.

### 📖 Description
This update ensures that the _attachments array stored in Mongo instance documents accurately reflects the deletion status of each attachment.

This change ensures that:
- Each attachment in the `_attachments` array now includes an `is_deleted` flag.
- The flag is automatically updated when attachments are moved to trash or restored.
- This is implemented efficiently using Mongo `bulk_write()` operations.

### 👀 Preview steps
1. Log into an account that has a project with at least one submission containing one or more attachments.
2. n Mongo, query the relevant instance document and confirm that each attachment in the `_attachments` array has `is_deleted: false`
3. Use the attachment deletion API to delete one or more attachments from the submission
4. Check the same Mongo instance document again. The corresponding attachment(s) in `_attachments` should now have `is_deleted: true`.
5. From the Django admin, restore the deleted attachment.
6. Re-check the Mongo document. The `is_deleted` flag for that attachment should now be set back to `false`.


kobo/apps/openrosa/apps/viewer/models/parsed_instance.py
```diff
@@ -443,20 +445,47 @@ def get_notes(self):
...
+            xform_instances.bulk_write(attachments_to_update)
+
 
-def _get_attachments_from_instance(instance):
+def _get_attachments_from_instance(instance_id):
     attachments = []
-    for a in instance.attachments.all():
+    for a in Attachment.all_objects.filter(
+        instance_id=instance_id
+    ).exclude(delete_status=AttachmentDeleteStatus.SOFT_DELETED):
```
rgraber, line 475: Do we want to exclude AttachmentDeleteStatus.DELETED as well?

rajpatel24: Since we’re using `Attachment.all_objects` instead of `instance.attachments.all()`, the default manager no longer automatically filters by `delete_status`. We definitely want to exclude attachments with the `AttachmentDeleteStatus.SOFT_DELETED` status, as they are soft deleted. Previously, this was handled automatically by the default manager, but now that we're using `Attachment.all_objects()`, we need to explicitly exclude them

sksondra (commented):
Good from QA.

PR #5789, opened 2025-05-21
fix(storybook): make dates consistent in UniversalTable tests DEV-277

magicznyleszek:
### 💭 Notes
Uses `faker.setDefaultRefDate()` to ensure we have identical dates in example test data.


jnm (commented):
Just a test for future QA workflow :)

PR #5793, opened 2025-05-22
feat(attachments): Add tests for `move_to_trash` and `put_back` support for `AttachmentTrash` TASK-1786

rajpatel24:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Added unit tests for attachment trash functionality, including `move_to_trash`, `put_back`, and media file deletion behavior.

### 📖 Description
Added unit tests to verify the behavior of `move_to_trash` and `put_back` for attachment objects. These tests confirm that:
- Attachments are properly marked for deletion and storage counters updated
- Media files are deleted via the `empty_attachment` task while retaining the DB record
- Attachments can be restored and associated storage values are recalculated
- Task restart logic works correctly based on status and time thresholds

### DIFF
--- a/kobo/apps/trash_bin/tests/test_utils.py
+++ b/kobo/apps/trash_bin/tests/test_utils.py
@@ -683,7 +683,10 @@ class AttachmentTrashTestCase(TestCase):
         ).exists()
 
     @data(
-        # Format: (task status, within grace period, is stuck, expected restart count)
+        # Format: (task status, is time frozen, is stuck, expected restart count)
+        # `is_time_frozen=True` simulates a past time (grace period expired)
+        # `is_time_frozen=False` uses current time (still within grace period)
+
         # Trivial case: Task is pending and still within the grace period
         (TrashStatus.PENDING, False, False, 0),
         # Task is pending but the grace period has expired — it should be restarted
@@ -703,8 +706,11 @@ class AttachmentTrashTestCase(TestCase):
         A freshly created task should not be restarted if it is in grace period.
         """
         if is_time_frozen:
+            # Freeze time to simulate that the grace period has passed.
+            # This allows us to test behavior after grace period expiration.
             frozen_time = '2024-12-10'
         else:
+            # Use the current time, meaning the task is still within its grace period.
             frozen_time = str(timezone.now())
 
         with freeze_time(frozen_time):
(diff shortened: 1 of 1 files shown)


kobo/apps/trash_bin/tests/test_utils.py
```diff
@@ -592,3 +594,215 @@ def test_storage_does_not_change_on_archive_unarchive(self):
...
+            metadata={
+                'attachment_uid': self.attachment.uid,
+                'attachment_name': self.attachment.media_file_basename,
+                'instance__root_uuid': self.attachment.instance.root_uuid,
+            },
+        ).exists()
+
+    @data(
+        # Format: (task status, within grace period, is stuck, expected restart count)
+        # Trivial case: Task is pending and still within the grace period
```
rgraber, line 687: This is confusing. If the task is within the grace period, why is "within grace period" False?

kobo/apps/trash_bin/tests/test_utils.py
```diff
@@ -592,3 +594,215 @@ def test_storage_does_not_change_on_archive_unarchive(self):
...
+        # Task failed and is considered stuck — still no restart (handled differently)
+        (TrashStatus.FAILED, True, True, 0),
+    )
+    @unpack
+    def test_task_restarter(self, status, is_time_frozen, is_stuck, restart_count):
+        """
+        A freshly created task should not be restarted if it is in grace period.
+        """
+        if is_time_frozen:
+            frozen_time = '2024-12-10'
```
rgraber, line 706: A comment here about how freezing time determines whether something is or is not in the grace period would be helpful

kobo/apps/trash_bin/tests/test_utils.py
```diff
@@ -592,3 +594,215 @@ def test_storage_does_not_change_on_archive_unarchive(self):
...
+        with freeze_time(frozen_time):
+            self._move_attachment_to_trash(self.attachment, self.user)
+
+        if status != TrashStatus.PENDING:
+            # Only tasks that are not pending can be considered as stuck
+            if is_stuck:
+                # Fake the task was started but is now stuck (its last run is older than
+                # the expected threshold).
+                last_run = timezone.now() - timedelta(
+                    seconds=settings.CELERY_LONG_RUNNING_TASK_TIME_LIMIT + 60 * 5 + 10
```
rgraber, line 719: Why are we adding 5 minutes + 10 seconds to the time limit?

rgraber (changes requested):
Could use some clarifying comments

kobo/apps/trash_bin/tests/test_utils.py
rajpatel24: you're absolutely right that the comment was misleading. I’ve updated it to clarify that `is_time_frozen` controls whether we're simulating a time after the grace period.

kobo/apps/trash_bin/tests/test_utils.py
rajpatel24: We're simulating a stuck task, which means it hasn't updated in a long time. We calculate that "long time" as beyond the Celery time limit, plus a grace buffer of 5 minutes and a small margin (10s) to account for clock drift or scheduling delays.

PR #5795, opened 2025-05-23
feat(formLanding): General UI improvements of the deployed form history view DEV-38

duvld:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Some UI improvements to help modernize the form version history view.
- Updating the "deployed" badge to match our other badges 
- Left-align the date in the "Last modified" column instead of center right aligned
- Clone icons should sit under the clone header and not next to the date when the form was last modified.

### 💭 Notes

Added some wrappers to help facilitate CSS changes.

### 👀 Preview steps
Feature/no-change template:
2. Have multiple deployed versions of the project
3. Open the form history view
4. 🟢 notice that view matches (reasonably) to this mock:
image

### DIFF
--- a/jsapp/js/components/formLanding/formLanding.js
+++ b/jsapp/js/components/formLanding/formLanding.js
@@ -30,6 +30,7 @@ import { ROUTES } from '#/router/routerConstants'
 import sessionStore from '#/stores/session'
 import { ANON_USERNAME, buildUserUrl } from '#/users/utils'
 import { formatTime, notify } from '#/utils'
+import AssetStatusBadge from '#/components/common/assetStatusBadge'
 
 const DVCOUNT_LIMIT_MINIMUM = 20
 const ANON_CAN_ADD_PERM_URL = permConfig.getPermissionByCodename(PERMISSIONS_CODENAMES.add_submissions).url
@@ -233,9 +234,9 @@ class FormLanding extends React.Component {
                 return (
                   <bem.FormView__group m='items' key={n} className={n >= this.state.DVCOUNT_LIMIT ? 'hidden' : ''}>
                     <bem.FormView__label m='version'>
-                      {`v${dvcount - n}`}
+                      <span>{`v${dvcount - n}`}</span>
                       {item.uid === this.state.deployed_version_id && this.state.deployment__active && (
-                        <bem.FormView__cell m='deployed'>{t('Deployed')}</bem.FormView__cell>
+                        <AssetStatusBadge deploymentStatus={this.state.deployment_status} />
                       )}
                     </bem.FormView__label>
                     <bem.FormView__label m='date'>{formatTime(item.date_deployed)}</bem.FormView__label>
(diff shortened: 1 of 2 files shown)


jsapp/scss/components/_kobo.form-view.scss
```diff
@@ -443,9 +443,18 @@ $side-tabs-width-mobile: 70px;
       &--version {
         width: 39%;
         text-align: initial;
+
+        > span {
+          margin-right: 12px;
+          vertical-align: middle;
```
magicznyleszek, line 449: Maybe it would be nicer to use flexbox with gap here?

jsapp/js/components/formLanding/formLanding.js
```diff
@@ -233,23 +234,26 @@ class FormLanding extends React.Component {
                 return (
                   <bem.FormView__group m='items' key={n} className={n >= this.state.DVCOUNT_LIMIT ? 'hidden' : ''}>
                     <bem.FormView__label m='version'>
-                      {`v${dvcount - n}`}
+                      <span>{`v${dvcount - n}`}</span>
                       {item.uid === this.state.deployed_version_id && this.state.deployment__active && (
-                        <bem.FormView__cell m='deployed'>{t('Deployed')}</bem.FormView__cell>
```
magicznyleszek, line 238: I'm guessing after this change we can remove `.form-view__cell--deployed` styles?

jsapp/js/components/formLanding/formLanding.js
```diff
@@ -233,23 +234,26 @@ class FormLanding extends React.Component {
...
-                        size='m'
-                        onClick={() => {
-                          this.saveCloneAs(item.uid)
-                        }}
-                        startIcon='duplicate'
-                        tooltip={t('Clone this version as a new project')}
-                        tooltipPosition='right'
-                      />
+                      <bem.FormView__label>
+                        <Button
```
magicznyleszek, line 245: We already touch the code here, so let's swich to our new button `jsapp/js/components/common/ButtonNew.tsx` :)

duvld: I'm actually a bit confused here. On `ButtonNew` it says to not use this for an icon with no text, and use ActionIcon instead. `ActionIcon` doesn't have a tooltip so just using that would be a regression. I think this:
```typescript
<Tooltip label={t('Clone this version as a new project')}>
  <ActionIcon
    size='md'
    variant='transparent'
    onClick={...}
>
  <Icon name='duplicate' size='l' />
```

Is so complicated for a "button and a tooltip", which `ButtonNew` would solve (but warns to not use for this case). Should I extend our `ActionIcon.tsx` to include tooltip?

magicznyleszek: Hmm… I thought this would be simpler to do… 😓 If you have time for adding tooltip to `ActionIcon`, let's do it. But no pressure, given the situation, it's fine to leave old button too.

sksondra (commented):
Good from QA.

PR #5799, opened 2025-05-26
chore: install playwright deps within ci_locally.sh

Akuukis:
### 💭 Notes

Playwright requires system dependencies that will be installed now.

### 👀 Preview steps

Run the `ci_locally.sh` script.


noliveleger (approved):
LGTM

PR #5801, opened 2025-05-28
fix(projects): "fields" selector uses same default display list as my projects route DEV-292

duvld:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Fixes issue where the "fields" selector would show "Countries" checked but the columns don't display it. Now the "fields" selector leaves "Countries" unchecked by default (to match the columns).

### 💭 Notes

See [notes ](https://linear.app/kobotoolbox/issue/DEV-292/countries-field-checked-as-visible-but-not-showing-in-project-view#comment-6d5f8d32) on linear

### 👀 Preview steps

Bug template:
1. ℹ️ have an account and multiple projects
2. navigate to the homepage (my projects)
3. reset the "fields" selector and refresh
4. 🟢 [on main] notice that the "countries" column is not present
5. click on "fields" selector
6. 🔴 [on main] notice that "countries" is checked
7. 🟢 [on PR] notice "countries" is not checked


jsapp/js/projects/projectViews/projectsFieldsSelector.tsx
```diff
@@ -8,7 +8,7 @@ import KoboModalContent from '#/components/modals/koboModalContent'
 import KoboModalFooter from '#/components/modals/koboModalFooter'
 import KoboModalHeader from '#/components/modals/koboModalHeader'
 import type { ProjectFieldName } from './constants'
-import { DEFAULT_VISIBLE_FIELDS, PROJECT_FIELDS } from './constants'
+import { HOME_DEFAULT_VISIBLE_FIELDS, PROJECT_FIELDS } from './constants'
```
magicznyleszek: Unfortunately this is a bit more complex. We have three different "project views": My Projects, My Org Projects, and custom ones. All three use the same component `UniversalProjectsRoute`. It has a prop `defaultVisibleFields` that should be passed on into `ProjectsFieldsSelector` instead of using one from `./constants`.

The fix you did fixed the problem for "My Projects", but broke it for custom views.

To have custom views, you need to set them up through admin (easiest is to create one with `*` for countries). To have org orojects views, you need to be in organization (but AFAIK, "My Org Projects" uses same defaults as "My Projects")

magicznyleszek (changes requested):

PR #5811, opened 2025-05-30
fix(massEmails): make get_plan_name ordering more deterministic TASK-1836

jamesrkiger:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Ensure `get_plan_name()` util function returns plans before addons.

### 💭 Notes
Previously non-deterministic ordering was causing a unit test to fail occasionally.

### 👀 Preview steps
On `main` the `test_get_plan_name` unit test fails about 1/10 times. So if you run `pytest  --reuse-db kobo/apps/stripe/tests/test_stripe_utils.py` 20 times you should see the improvement on this branch.


sksondra (commented):
Good from QA.

PR #5817, opened 2025-06-02
feat(billing): create useTrackingPeriod hook DEV-494

pauloamorimbr:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
This PR:
- Adds the `useTrackingPeriod` hook
- Replaces current uses of the `trackingPeriod`, previously from the `useUsage` hook the the new hook

### 👀 Preview steps
Feature/no-change template:
2. Have stripe setup
3. Go to account settings
4. Go to Usage panel
5. 🟢 notice that the tracking period of the messages (year or month) is properly displayed
6. Go to `Per Project Total` tab
5. 🟢 notice that the tracking period of the messages (year or month) is properly displayed


jsapp/js/account/usage/useTrackingPeriod.tsx
```diff
@@ -0,0 +1,16 @@
+import { useEffect, useState } from 'react'
+import { getSubscriptionInterval } from '../stripe.api'
+import type { RecurringInterval } from '../stripe.types'
+
+export const useTrackingPeriod = (): RecurringInterval => {
+  const [trackingPeriod, setTrackingPeriod] = useState<RecurringInterval>('month')
+
+  useEffect(() => {
+    const fetchTrackingPeriod = async () => {
+      setTrackingPeriod(await getSubscriptionInterval())
```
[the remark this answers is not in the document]
Akuukis: here are helpers from my other project to generalize this:
```ts
/**
 * Like `useEffect` that accepts an async function and has a fallback catch clause.
 */
export const useAsyncEffect = (afn: () => Promise<void>, deps?: DependencyList): void => {
  useEffect(() => {
    afn().catch((err) => console.error(`Async Effect "${afn.name}" failed:`, err));
  }, deps);
};
export const useAsyncEffectOnce = (afn: () => Promise<void>): void => useAsyncEffect(afn, []);
```

then 

```ts
useAsyncEffectOnce(async () => {
  setTrackingPeriod(await getSubscriptionInterval())
})

File: jsapp/js/account/usage/useTrackingPeriod.tsx [hunk 2]
```diff
@@ -0,0 +1,16 @@
+import { useEffect, useState } from 'react'
+import { getSubscriptionInterval } from '../stripe.api'
```
jamesrkiger: I think we should move the logic (and/or code) from getSubscriptionInterval directly into this hook and remove it from the util file. This is the only place we use that method, so I don't see much value in having it as a separate util. Also, it depends on SubscriptionStore and I'd like to make this hook's dependency on SubscriptionStore more explicit.

jsapp/js/account/usage/useTrackingPeriod.tsx
```diff
@@ -0,0 +1,16 @@
...
+import type { RecurringInterval } from '../stripe.types'
+
+export const useTrackingPeriod = (): RecurringInterval => {
```
jamesrkiger: What would you think about renaming this useBillingPeriod (and renaming instance variables accordingly)? It could be argued that it isn't always a billing period (on non-Stripe instances tracking is technically more accurate). But I think I'd rather be more descriptive, especially given that most places we actually use this will be billing related.

jsapp/js/account/usage/useTrackingPeriod.tsx
pauloamorimbr: 💯 
I wasn't sure if I should "disconnect" it from the stripe utils, since it's still kind of a stripe information. But I agree it makes much more sense. I'm gonna change this.

jsapp/js/account/usage/useTrackingPeriod.tsx
pauloamorimbr: So...I've took another path.
Since the tracking period is part of the subscription, and detaching only it is impossible, I've actually implemented an `useSubscriptionQuery` hook to gather subscription data via react query and make the parts of it available for use. Among the data available is the `billingPeriod`, which replaced the tracking period.

See what you think about it.

jamesrkiger: @pauloamorimbr In and of itself, I like this. But I don't like splitting the responsibility for handling subscription data between the subscription store and this new hook.

Case in point: Take a look at the limit notification banner that is supposed to appear on the projects list page if you are over a usage limit. With this new code, the banner doesn't show up when you load the app (to save time, you can reproduce by just hardcoding in a big number of submissions in the useUsage hook), because the subscription store isn't initialized even though the subscription endpoint has been called for the new subscription data hook. If you navigate to the account section of the app, then back to the projects list, the banner will show up because the subscription store gets initialized in the account section. It's not a hard problem to fix, but I think we're inviting such bugs if we have redundant data management like this. 

At the same time, you can also see duplicate calls to the subscription endpoint (one for the new hook and one for the store) on the same page if you navigate by URL or refresh the addons page, for example. Not the end of the world, of course, but as a matter of principle I think that's something we should try to avoid.

The [original idea](https://linear.app/kobotoolbox/issue/DEV-494/make-usetrackingperiod-hook) for this task was to use data from the subscription store in a small hook for retrieving the billing period. @Akuukis Is right that the ideal would be to get rid of subscription store entirely. And that is our mid-term goal. But we decided on a more piecemeal approach for this project because overhauling the guts of our frontend handling seemed like too much scope creep for a fairly small frontend project like this one.

I would suggest that we either bite the bullet and remove subscription store entirely or stick with the original plan and use subscription store data for a small useBillingPeriod hook. Let me know what you think.

I'd also appreciate any input from @Akuukis on this.

pauloamorimbr: I just decided to create the `useBillingPeriod` hook to get data from the subscriptionStore.
I'm handling the async function as a data source and I'm using react query to get call the function in the hook as needed, avoiding async call inside of a `useEffect` or helpers to do so.

I've left the subscriptionQuery hook in the code (FOR NOW), just for us to talk about it, and then, since we're more likely not to use it now, I'll move it to some branch/PR related to react query migration, since it's work done and its likely to only need the said QA.

Akuukis: @jamesrkiger feel free to tap into maintenance budget to do this as part of the project.

pauloamorimbr: @Akuukis Talking to @jamesrkiger we've decided to keep this small as intended and do the proper migration of the subscriptionStore in a [future task](https://linear.app/kobotoolbox/issue/DEV-586/migrate-subscriptionstore-to-reactquery).
I've added the code for the useSubscriptionQuery hook to a branch in KPI and linked it at the task so we can have a head start in the future.

@jamesrkiger I've cleared this PR and I believe it's ready for review now. Thank you!

jamesrkiger (approved):
lgtm

sksondra (commented):
Good from QA.

PR #5818, opened 2025-06-03
fix(projectHistoryLogs): fix 500 when exporting as anonymous user DEV-376

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Fix an error when downloading public data while not logged in.

### 👀 Preview steps

Bug template:
1. ℹ️ have an account and a project with some submissions
2. Allow anonymous access to project submissions
3. In a private browser tab, go to the public URL for project data
4. Go to Downloads and try to export as CSV
5. 🔴 [on main] Error
6. 🟢 [on PR] The export should complete successfully

### DIFF
--- a/kobo/apps/audit_log/tests/test_project_history_logs.py
+++ b/kobo/apps/audit_log/tests/test_project_history_logs.py
@@ -939,8 +939,8 @@ class TestProjectHistoryLogs(BaseAuditLogTestCase):
             'xls_types_as_text': False,
             'include_media_url': True,
         }
-        self.asset.assign_perm(perm=PERM_VIEW_SUBMISSIONS, user_obj=AnonymousUser())
         if anonymous_user:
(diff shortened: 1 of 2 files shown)


kobo/apps/audit_log/tests/test_project_history_logs.py
```diff
@@ -938,6 +939,9 @@ def test_export_creates_log(self):
             'xls_types_as_text': False,
             'include_media_url': True,
         }
+        self.asset.assign_perm(perm=PERM_VIEW_SUBMISSIONS, user_obj=AnonymousUser())
```
noliveleger: I'm just wondering if this line should be moved inside the  `if` statement below.
Let's say, at a later time, we change something which makes this test fail if the project does not allow view submissions to anonymous user).

rgraber: That's a good point

PR #5830, opened 2025-06-06
fix(massEmails): no error on empty test user list DEV-562

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Do not attempt to send emails when the test email configuration is empty.

### 📖 Description
Fixes an error wherein if the `MASS_EMAIL_TEST_EMAILS` configuration in Constance was empty, any email using the `test_users` query would throw errors, which could kill the whole process and prevent other emails from sending.

### 👀 Preview steps
Remember to restart the kpi_worker after switching branches

1. In Constance, reset `MASS_EMAIL_TEST_EMAILS` to default
2. Create a MassEmailConfig that uses the `test_users` query and set it to live
3.  🔴 [on main] At one minute after the next 15 m boundary the send process will fail
4. 🟢 [on PR] No errors in kpi_worker


Guitlle (approved):

sksondra (commented):
Good from QA.

PR #5835, opened 2025-06-11
perf(attachments): Improve bulk update attachments storage counters DEV-351

rajpatel24:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Improved performance of storage counter updates when bulk trashing or restoring attachments and projects.

### 📖 Description
- Refactors the logic for updating attachment storage counters on user profiles and xforms during bulk trash or restore operations.
- With this PR, when a XForm is trashed or restored, only the `UserProfile` storage counter will be updated. The XForm’s storage value remains unchanged because trashed forms are excluded from queries and considered deleted along with their attachments.

### DIFF
--- a/kobo/apps/openrosa/apps/logger/utils/attachment.py
+++ b/kobo/apps/openrosa/apps/logger/utils/attachment.py
@@ -1,10 +1,11 @@
-from django.db import transaction
 from django.db.models import F, OuterRef, Sum, Subquery
 from django.db.models.functions import Coalesce
 from django.db.models.query import QuerySet
 
 from kobo.apps.openrosa.apps.logger.models import Attachment, XForm
+from kobo.apps.openrosa.apps.logger.models.attachment import AttachmentDeleteStatus
 from kobo.apps.openrosa.apps.main.models import UserProfile
+from kpi.deployment_backends.kc_access.utils import kc_transaction_atomic
 
 
 def bulk_update_attachment_storage_counters(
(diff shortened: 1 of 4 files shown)


kobo/apps/openrosa/apps/logger/utils/attachment.py
```diff
@@ -1,26 +1,38 @@
...
+    attachments = Attachment.all_objects.filter(
+        uid__in=attachment_identifiers,
+        delete_status=target_delete_status
+    )
+    if not attachments.exists():
+        return
+
+    attachment_ids = attachments.values_list('pk', flat=True).distinct()
+    user_ids = attachments.values_list('user_id', flat=True).distinct()
+    xform_ids = attachments.values_list('xform_id', flat=True).distinct()
```
noliveleger: I think `distinct` is not required here. IMO, it could slow down the query if we are updating a large dataset. Morever I don't think it's a problem to call `pk__in=[1,1,1,2,2,2]` vs `pk__in=[1,2]` because PostgreSQL will probably do a great job to optimize the array itself vs forcing the `distinct()`.

At least, `pk` in `attachment_ids` shouldn't be duplicated, right?

kobo/apps/openrosa/apps/logger/utils/attachment.py
```diff
@@ -38,7 +50,7 @@ def bulk_update_attachment_storage_counters(
         .values('total_size')
     )
 
-    with transaction.atomic():
+    with kc_transaction_atomic():
```
noliveleger: You can probably use `conditional_kc_transaction_atomic` to avoid nested transactions, in that case.

kobo/apps/trash_bin/tests/test_utils.py
```diff
@@ -501,7 +501,7 @@ def test_storage_updates_on_project_trash_and_restore(self):
         )
         xform.refresh_from_db()
         user_profile.refresh_from_db()
-        self.assertEqual(xform.attachment_storage_bytes, 0)
+        self.assertGreater(xform.attachment_storage_bytes, 0)
```
noliveleger: Since we know the size of the attachments, can we hardcode the value here and use `assertEqual` instead. 
`assertGreater` makes me feel a little bit scared.

kobo/apps/openrosa/apps/logger/utils/attachment.py
rajpatel24: You're right, `distinct()` on `pk` isn't needed, and for `user_id` and `xform_id`, it's not required either since PostgreSQL handles duplicates efficiently.

noliveleger (approved):
LGTM

PR #5838, opened 2025-06-11
fix(attachments): hide the "Delete media files only" button from any permission except `change_submissions` or higher

duvld:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Those with `delete_submissions` permission for submissions cannot see the "Delete media files only" button in the bulk options

### 💭 Notes

Missed a spot when writing the guards for displaying the button. Clicking it would trigger another set of guards in the actual component so the user couldn't delete attachments this way anyway.

### 👀 Preview steps

1. ℹ️ have 3 accounts and a project with media attachment submissions
2. allow a `user_a` to have `delete_submissions` permissions
3. allow a `user_b` to have `change_submissions` permissions
4. 🔴 [on main] notice that both `user_a` and `user_b` can see "Delete media files only" button in the bulk select options'
6. 🟢 [on PR] notice `user_a` cannot see  "Delete media files only" button, but `user_b` still can

### DIFF
--- a/jsapp/js/components/submissions/tableBulkOptions.tsx
+++ b/jsapp/js/components/submissions/tableBulkOptions.tsx
@@ -267,7 +267,7 @@ class TableBulkOptions extends React.Component<TableBulkOptionsProps> {
           )}
 
         {Object.keys(this.props.selectedRows).length > 0 &&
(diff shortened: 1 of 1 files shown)


magicznyleszek (approved):

sksondra (commented):
Good from QA.

PR #5841, opened 2025-06-12
feat(massEmails): preview query results DEV-478

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Allow users to download a list of expected recipients of a mass email based on the query used.

### 📖 Description
Creates ExportJobs for MassEmailConfigs that will have details about the expected recipients for the emails. Once requested, the user will receive an email when the job is done and be able to download the list from Django admin.

### 💭 Notes
Uses the existing django-import-export-celery framework even though it's more meant for exporting lists of models rather than this kind of byproduct of a model. However, the framework is already there and allows multiple formats as well as handling the email portion, and the tweaks necessary to make it work for a list of users rather than a list of configs are pretty slight (just doing some work in the `after_export` hook).
If there are no recipients, the process will just return an empty csv (only headers). This is not ideal, but there is not an easy way to tell the user that no recipients were found using just the library. The job status would be the most obvious place to put this kind of information, but is part of the ExportJob object which we don't have access to in any of the export methods. We could run the query first before creating the export job and use the usual admin error alerts if the query returns empty, but the query could take a long time and this would defeat the purpose of doing all the work in the background.

### 👀 Preview steps
Run mailhog or file-backed email for easier testing.

1. ℹ️ have a superuser account and at least one other
2. In Django admin, update MASS_EMAIL_TEST_EMAILS in Constance with at least 2 user emails
3. In Django admin, create a MassEmailConfig that uses the 'test_users' query. Frequency/live doesn't matter here.
4. On the list of MassEmailConfigs, select your new configuration.
5. 🟢 There should be an "Export with celery" action
6. Trigger the export
7. You will be taken to a prefilled ExportJob creation form. Save it.
8. 🟢 Your superuser account should receive an email with the subject  " Django: Export job completed" and a link to an Export Job
9. 🟢 You should be able to download the list of recipients from the ExportJob (it will be the 'Exported file' field).
10. 🟢 The list should contain the email, username, and uid of all recipients

### DIFF
--- a/kobo/apps/mass_emails/models.py
+++ b/kobo/apps/mass_emails/models.py
@@ -113,8 +113,11 @@ class MassEmailConfigExpectedRecipientsResource(resources.ModelResource):
                         user['extra_details__uid'],
                     ]
                 )
-        dataset._data = reformatted
+        # empty the old dataset so we can set the new headers without an
+        # "InvalidDimensions" error
+        dataset.wipe()
         dataset.headers = ['MassEmailConfig name', 'username', 'email', 'uid']
+        dataset._data = reformatted
 
 
 class MassEmailJob(AbstractTimeStampedModel):


sksondra (commented):
Good from QA.

PR #5843, opened 2025-06-13
feat(massEmails): download a CSV of recipients from the last 30 days DEV-587

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Allow superusers to download a report of recipients of all emails for the past 30 days.

### 💭 Notes
Adds a new resource for exporting a report of all users who received a given email or emails. It only looks back 30 days to avoid massively long queries. In an ideal world this would be configurable but it proved very difficult to hook into the import-export-celery process to insert this kind of configuration, and I thought it would be worth it to have the same process for looking at both expected and actual recipients.
Also adds the MassEmailConfig uid to the expected recipients export for consistency.

### 👀 Preview steps
Use MailHog or file-based emails for easiest testing

1. ℹ️ have a super user account
2. Ensure there is at least one email in `MASS_EMAIL_TEST_EMAILS`
3. Create a new MassEmailConfig using the test_users query and set it to live (frequency doesn't matter, though it might be easier to see results if you set it to daily)
4. Wait for the email to send at least once
5. In the list view, select the mass email config you just created and select the `Export with celery` action
6. In the ExportJob form, select "Last 30 days recipients resource" and save
7. 🟢 You should receive an email with a link to the ExportJob, which will in turn have a link to download a csv with information about all the recipients of the email

### DIFF
--- a/kobo/apps/mass_emails/models.py
+++ b/kobo/apps/mass_emails/models.py
@@ -1,7 +1,9 @@
 from collections.abc import Callable
+from datetime import timedelta
 from enum import Enum
 
 from django.db import models
+from django.utils import timezone
 from import_export import fields, resources
 
 from kobo.apps.kobo_auth.shortcuts import User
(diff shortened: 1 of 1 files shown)


sksondra (commented):
Good from QA.

PR #5844, opened 2025-06-13
fix(attachments): fix attachment deletion for edited submissions TASK-1841

rajpatel24:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Fixes an issue where attachments could not be deleted from edited submissions due to incorrect handling of submission UUIDs.

### 📖 Description
This fix updates the attachment deletion logic to correctly filter using `root_uuid` for edited submissions. Previously, the API returned a 400 error when edited submissions were included, treating their root UUIDs as invalid. The updated query ensures attachments from both original and edited submissions can be deleted successfully.

### DIFF
--- a/kpi/serializers/v2/attachment_delete.py
+++ b/kpi/serializers/v2/attachment_delete.py
@@ -1,3 +1,4 @@
+from django.db.models import Q
 from django.utils.translation import gettext as t
 from rest_framework import serializers
 
@@ -54,9 +55,12 @@ class AttachmentDeleteSerializer(serializers.Serializer):
             attachment_uids = [self.validated_data['attachment_uid']]
             field = 'attachment_uid'
         else:
+            uuids = self.validated_data['submission_root_uuids']
             attachments = Attachment.objects.filter(
-                xform_id=deployment.xform_id,
-                instance__uuid__in=self.validated_data['submission_root_uuids'],
+                Q(xform_id=deployment.xform_id) & (
+                    Q(instance__root_uuid__in=uuids) |
+                    (Q(instance__uuid__in=uuids) & Q(instance__root_uuid__isnull=True))
+                )
             )
             attachment_uids = list(attachments.values_list('uid', flat=True))
             field = 'submission_root_uuids'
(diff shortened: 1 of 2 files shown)


kpi/serializers/v2/attachment_delete.py
```diff
@@ -54,9 +55,12 @@ def save(self, request, asset):
...
             field = 'attachment_uid'
         else:
+            uuids = self.validated_data['submission_root_uuids']
             attachments = Attachment.objects.filter(
-                xform_id=deployment.xform_id,
-                instance__uuid__in=self.validated_data['submission_root_uuids'],
+                Q(xform_id=deployment.xform_id) & (
+                    Q(instance__root_uuid__in=uuids) |
+                    (Q(instance__uuid__in=uuids) & Q(instance__root_uuid__isnull=True))
+                )
```
noliveleger: This could be simplified with  : 

```python 
Q(instance__root_uuid__in= uuids)
| Q(instance__uuid__in= uuids, instance__root_uuid__isnull=True),
xform_id=deployment.xform_id
```

Let's add a TODO to remove the condition when `0005` long-running migration is complete (in few months) (same for the NLP bugfix already merged please.

kpi/tests/api/v2/test_api_attachments_delete_viewset.py
```diff
@@ -300,6 +300,53 @@ def test_bulk_delete_attachments_success(self):
...
+        """
+        Test bulk deletion of attachments for a submission that has been edited
+        """
+        self.client.login(username='someuser', password='someuser')
+        edited_submission = {
+            '__version__': self.asset.latest_deployed_version.uid,
+            'meta': {
+                'instanceID': 'uuid:test_edited_uuid',
+                'rootUuid': 'uuid:test_uuid1',
+                'deprecatedID': 'uuid:test_uuid1',
```
noliveleger, line 313: We need a test for old submissions also, where `root_uuid` is null.
It can be faked with `Instance.objects.get(root_uuid=test_uuid1).update(root_uuid=None)`
You could use `ddt` to avoid rewriting another test but just use a condition to reset the root_uuid?

kpi/serializers/v2/attachment_delete.py
noliveleger: @rajpatel24 , please use the simplified version for the filter ;-)

sksondra (commented):
Good from QA.

PR #5848, opened 2025-06-17
refactor(billing): split stripe utils into several files DEV-604

jamesrkiger:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Refactors stripe utils to reduce file size and risk of circular dependencies.

### 📖 Description
Splits stripe utils into the following four files:
`billing_dates.py`
`import_management.py`
`misc.py`
`subscription_limits.py`

### 👀 Preview steps
No behavior should change. Risks should be covered by unit tests.

### DIFF
--- a/kobo/apps/stripe/tests/test_stripe_utils.py
+++ b/kobo/apps/stripe/tests/test_stripe_utils.py
@@ -24,8 +24,6 @@ from kobo.apps.stripe.utils.billing_dates import (
     get_current_billing_period_dates_by_org,
     get_current_billing_period_dates_for_active_plans,
 )
-from kobo.apps.stripe.utils.view_utils import (
-)
 from kobo.apps.stripe.utils.subscription_limits import (
     determine_limit,
     get_default_plan_name,


kobo/apps/stripe/utils/misc.py
```diff
@@ -0,0 +1,72 @@
...
+    default_plan = (
+        Product.objects.filter(metadata__default_free_plan='true')
+        .values('name')
+        .first()
+    )
+    if default_plan is not None:
+        return default_plan['name']
+
+
+def generate_return_url(product_metadata):
```
rgraber: This and the method below are only called in views.py, maybe they should either move to that file or be their own view_utils.py file? I'm fundamentally opposed to having a 'misc' file

kobo/apps/stripe/utils/misc.py
```diff
@@ -0,0 +1,72 @@
...
+        total_price = total_price / price.transform_quantity['divide_by']
+        if price.transform_quantity['round'] == 'up':
+            total_price = ceil(total_price)
+        else:
+            total_price = floor(total_price)
+    return total_price * price.unit_amount
+
+
+@requires_stripe
+def get_plan_name(org_user: OrganizationUser, **kwargs) -> str | None:
```
rgraber: I think we can reasonably include the plan name stuff in `subscription_limits.py`

rgraber (changes requested):
I would really love to not have a misc.py file

PR #5849, opened 2025-06-17
fix(massEmails): rename export action DEV-605

rgraber:
### 🗒️ Checklist

1. [ ] run linter locally

### 📣 Summary
Rename the export action from 'Export with celery' to 'Export recipient list' for clarity

### 💭 Notes
Wraps the method in another one so we can change the short description in this instance without updating it for anyone else who may want to use the method.

### 👀 Preview steps
The action in the admin for MassEmailConfig should show 'Export recipient list' instead of 'Export with celery.' It should work the same.


Guitlle (approved):

sksondra (commented):
Good from QA.

PR #5856, opened 2025-06-18
fix(organizations): expect import column to be named organization_id DEV-343

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Allow re-importing of an exported user list.

### 📖 Description
Fixes a bug wherein if a user exported a list of an organization users and updated the organizations for some of the users, importing the new file would fail.

### 💭 Notes
Use "organization_id" as the id column for both import and export. Note when modifying the file you still have to update the organization name as well, otherwise the import will fail. This is probably more of a feature than a bug since it ensures whoever is editing the file has the correct organization in mind.

### 👀 Preview steps
1. ℹ️ have an account and at least one additional MMO
2. In Django admin, export the list of OrganizationUsers
3. Update the organization and organization_id columns on one of the rows to the name and uid of new MMO
4. Reimport the updated file
5. 🔴 [on main] Error "Organization xyz does not exist."
6. 🟢 [on PR] The user is moved to the new organization

### DIFF
--- a/kobo/apps/organizations/admin/organization_user.py
+++ b/kobo/apps/organizations/admin/organization_user.py
@@ -149,7 +149,7 @@ class OrgUserResource(resources.ModelResource):
 
     def before_import_row(self, row, **kwargs):
 
-        if not (organization := self._get_organization(row.get('organization'))):
+        if not (organization := self._get_organization(row.get('organization_id'))):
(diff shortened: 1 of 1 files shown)


noliveleger (approved):
LGTM

PR #5862, opened 2025-06-25
fix(attachments): include older xpaths in data api

rgraber:
### 🗒️ Checklist

1. [ ] run linter locally

### 📣 Summary

TODO

### 📖 Description

TODO

TODO

### 💭 Notes

TODO

### 👀 Preview steps

2. do this
3. do that
6. do that another thing


kpi/models/asset.py
```diff
@@ -663,11 +663,31 @@ def get_advanced_submission_schema(self, url=None, content=False):
             content, self.advanced_features, url=url
         )
 
-    @cache_for_request
-    def get_attachment_xpaths(self, deployed: bool = True) -> Optional[list]:
+    def get_all_attachment_xpaths(self):
```
noliveleger, line 665: Let's keep the decorator `@cache_for_request` here. 

`versions = self.asset_versions.all()` always fetches the data from the DB. Without the decorator, this query would be executed once per submission on the current page, resulting in N queries for N submissions.

kpi/models/asset.py
```diff
@@ -663,11 +663,31 @@ def get_advanced_submission_schema(self, url=None, content=False):
...
+                insert_xpath = True
+
+            xpaths.update(
+                self.get_attachment_xpaths_from_version(
+                    version, insert_xpath=insert_xpath
+                )
+            )
+        return list(xpaths)
+
+    def get_attachment_xpaths(self, deployed: bool = True):
```
noliveleger: Any reason why you did not pass the submission `__version__` to get its specific xpath?  

```
[
{
  "_id": 49532,
  "group1/image1: "image_2.png",
  "__version__": "vWb5E8biHYnH4MoptGTYs",
},
{
  _id: 49531,
  "image1": "image_2.png",
  "__version__": "vWiCvkE8biHYnH4MosEy4i",
}
```

rgraber: For this method I left the original signature so it won't break any current usage. get_attachment_xpaths_from_version is the one where we pass specific versions

kpi/models/asset.py
rgraber: Good call. I'll add a unit test for that since it's something that could be easily overlooked in the future

RuthShryock (commented):
Re-tested and works great :)

kpi/models/asset.py
noliveleger: Fair enough but my point is, since we are in the scope of the submission when we call `attachment_xpaths = self.asset.get_all_attachment_xpaths()`, we do know the version of the form that submission was using when it was created, we could directly send that version to `attachment_xpaths = get_attachment_xpaths_from_version(version=submission['__version__'])` and get an even more accurate list.  

but while I'm writing this comment, I'm thinking that your solution is probably better bullet proof with edits.  🤷‍♂️  Does the version change on edits? TBH, I don't remember.

rgraber: i don't think the version changes on edits but let me check

rgraber: The version does not change so I think we may need to get all versions. we could restructure the code so that we get all these ahead of time though and pass the results to each call

kpi/models/asset.py
```diff
@@ -664,10 +664,24 @@ def get_advanced_submission_schema(self, url=None, content=False):
         )
 
     @cache_for_request
-    def get_attachment_xpaths(self, deployed: bool = True) -> Optional[list]:
-        version = (
-            self.latest_deployed_version if deployed else self.latest_version
-        )
+    def get_all_attachment_xpaths(self):
+        versions = self.asset_versions.all()
```
noliveleger: Just in case, we only need the deployed versions. OpenRosa does not know nothing about not deployed versions.

rgraber: There was a test specifically for making sure we hit non_deployed versions too, was that outdated?

noliveleger: I don't know for that test, but in this particular case, a submission version cannot match a non-deployed version.

rgraber: Could it have been for the qpath-to-xpath work? that's the other place the method was called

rgraber: Oh wait that was called with deployed=True. :shrug: i'll take it out

noliveleger: I remember that I wanted it to be future-proof and told myself "Maybe we may need to have xpaths with none-
deployed versions". That's why the parameter `deployed` was there. But IIRC, we never used `get_attachment_xpaths(deployed=False)` anywhere.

PR #5863, opened 2025-06-25
fix(projectOwnership): prevent failed transfers caused by race conditions DEV-629

noliveleger:
### 📣 Summary
Ensures ownership transfer tasks don’t start before the database state is fully committed.

### 📖 Description
This fix addresses a race condition where Celery tasks for project ownership transfer are triggered too early, before the database has fully persisted the new state after a save() operation.

As a result, the task may incorrectly detect the new owner or fail altogether.

The logic now ensures that Celery only runs after the database state is consistent.

### Preview steps
See #5857

🔴 [on the release branch] In some circumstances, the projects are not correctly transferred because the recipient is either null or the previous org owner.  (You may need to repeat this step many times to get the race condition)
🟢 [on PR] The transfer should succeed every time

### DIFF
--- a/kobo/apps/organizations/admin/organization_user.py
+++ b/kobo/apps/organizations/admin/organization_user.py
@@ -3,6 +3,7 @@ from collections import defaultdict
 from django import forms
 from django.conf import settings
 from django.contrib import admin, messages
+from django.db import transaction
 from django.db.models import Count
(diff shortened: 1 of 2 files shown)


rajpatel24 (approved):
LGTM

PR #5860, opened 2025-06-20
feat(billing): ensure accurate usage limit reporting DEV-444

jamesrkiger:
### 🗒️ Checklist

1. [x] run linter locally
2. [x] update developer docs (API, README, inline, etc.), if any
3. [x] for user-facing doc changes create a Zulip thread at #Kobo support docs, if any
4. [x] draft PR with a title `<type>(<scope>)<!>: <title> DEV-1234`
5. [x] assign yourself, tag PR: at least `Front end` and/or `Back end` or `workflow`
6. [x] fill in the template below and delete template comments
7. [x] review thyself: read the diff and repro the preview as written
8. [x] open PR & confirm that CI passes & request reviewers, if needed
9. [ ] delete this section before merging

### 📣 Summary
Adds a `check_limit_exceeded()` util to update ExceededLimitCounters after submissions are submitted or nlp counters are updated.

### 💭 Notes
Function uses cache to ensure it only runs at most once for a given limit type in the settings.ENDPOINT_CACHE_DURATION window.

However, when it does run it adds a significant number of queries to an xform submission request. 

PR includes addition of fakeredis package to dev dependencies for testing cache expiration with freezegun.

### 👀 Preview steps
Focus should be on unit tests for the time being, but if you'd like to see it in action the easiest way is through NLP features

1. With a user on a community plan (with Stripe enabled and synched), create a project and submit an audio file
2. Manually copy in a long text over 6,000 characters as a transcript 
3. Translate it into another language
4. Check django shell to see that ExceededLimitCounter has been added

Can also be tested with submissions by manually setting the default free community plan product "submission_limit" in django admin to a small number.

PR #5866, opened 2025-06-26
feat(billing): block submissions for users over usage limit DEV-485

jamesrkiger:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Blocks submissions with a 402 payment required error if asset owner has exceeded their submission or storage limit.

### 📖 Description
Determines whether asset owner is over limit by calling `ServiceUsageCalculator.get_balances()` using the calculator's cache. So although it does add more queries to the submission API, these queries will only be run intermittently for any given asset owner.

### 💭 Notes

### 👀 Preview steps
1. Create a Stripe-enabled instance and sync stripe prices and products. For testing convenience set env var ENDPOINT_CACHE_DURATION to a small number like 5.
2. On a Stripe-enabled and synced instance, use django-admin to edit the metadata "submission_limit" for the default free community plan product (stripe id: "prod_EqgISHHelN2gXC")  to a small number.
3. Create a new project and make as many submissions as you are allowed under your new submission usage limit
4. Make sure you wait until endpoint cache duration passes, then make a new submission
5. Notice that Enketo displays an error message including the 402 error code.

### DIFF
--- a/kobo/apps/openrosa/libs/utils/logger_tools.py
+++ b/kobo/apps/openrosa/libs/utils/logger_tools.py
@@ -313,6 +313,10 @@ def create_instance(
                 )
 
             if settings.STRIPE_ENABLED:
+                from kobo.apps.stripe.utils.limit_enforcement import (
+                    check_exceeded_limit,
+                )
+
                 check_exceeded_limit(xform.user, UsageType.SUBMISSION)
                 check_exceeded_limit(xform.user, UsageType.STORAGE_BYTES)
 


kobo/apps/openrosa/apps/api/tests/viewsets/test_xform_submission_api.py
```diff
@@ -68,10 +70,59 @@ def test_query_counts(self):
...
             if settings.STRIPE_ENABLED:
-                expected_queries = FuzzyInt(69, 73)
+                expected_queries = FuzzyInt(79, 83)
             with self.assertNumQueries(expected_queries):
                 self.view(request)
 
+    @pytest.mark.skipif(
+        not settings.STRIPE_ENABLED, reason='Requires stripe functionality'
+    )
+    def test_over_limit_submission_rejection(self):
```
noliveleger: I would like to see the same test with an authenticated user, just in case!

kobo/apps/openrosa/apps/logger/exceptions.py
```diff
@@ -87,3 +87,7 @@ class TemporarilyUnavailableError(Exception):
 
 class XLSFormError(Exception):
     pass
+
+
+class ExceededUsageLimitError(Exception):
+    pass
```
noliveleger: Please keep the sort of the classes in that file, i.e.: A->Z.

kobo/apps/openrosa/libs/utils/logger_tools.py
```diff
@@ -202,6 +209,13 @@ def create_instance(
     xml_hash = Instance.get_hash(xml)
     xform = get_xform_from_submission(xml, username, uuid)
     check_submission_permissions(request, xform)
+    if settings.STRIPE_ENABLED and check_usage_limits:
+        calculator = ServiceUsageCalculator(xform.user)
+        balances = calculator.get_usage_balances()
+        for usage_type in [UsageType.STORAGE_BYTES, UsageType.SUBMISSION]:
+            balance = balances[usage_type]
+            if balance and balance['exceeded']:
+                raise ExceededUsageLimitError()
```
noliveleger, line 215: At which moment a balance does not exist? 

If it always exist, could it be simplified with something like

``` 
if any([
    balances[UsageType.STORAGE_BYTES]['exceeded'], 
    balances[UsageType. SUBMISSION]['exceeded'],
]):
    raise ExceededUsageLimitError()
```

kobo/apps/openrosa/libs/utils/logger_tools.py
```diff
@@ -153,6 +155,7 @@ def create_instance(
     uuid: str = None,
     date_created_override: datetime = None,
     request: Optional['rest_framework.request.Request'] = None,
+    check_usage_limits: bool = True
```
noliveleger: I thought we would need `check_usage_limits=True` at some points in the tests. It seems it is not the case.
So what about only use `if ... and not settings.TESTING`  in `create_instance` directly? 

Do you think it could be useful for future purposes?

noliveleger (changes requested):

kobo/apps/openrosa/libs/utils/logger_tools.py
jamesrkiger: Balances are `None` when a user has unlimited usage for a given usage type.

kobo/apps/openrosa/libs/utils/logger_tools.py
jamesrkiger: I can imagine it having uses in the future. Personally, I think it makes sense to keep it configurable. There's no in principle reason the checks should always be skipped in testing.

noliveleger: fair enough

PR #5867, opened 2025-06-26
docs(api): adding API endpoints for asset-data in KPI documentation DEV-527

ZoeGodin:
### 📣 Summary
Added API documentation for `/api/v2/docs/asset_data` endpoint.

### 📖 Description
This PR introduces the endpoint documentation that will be used for the auto-generating API documentation of the asset_attachments endpoint. With these changes, users will be able to understand, test and use the `asset_data` endpoint more easily.

### 💭 Notes
Due to current limitations in drf-spectacular support for OAS 3.1, it's not possible to include structured request or response bodies for certain endpoints. This applies to all functions flagged as "vague."
As a workaround, example payloads have been included directly in the description fields of the affected endpoints. Full support is pending a future update of the tooling.

### 👀 Preview steps
To see the changes, visit `http://kf.kobo.local/api/v2/docs/`. The page of the Kobo API will now be visible and generated by swagger. To see the changes added to `asset_data` endpoint, scroll down to the category of the same name or make a research in the search bar. Documentation for the endpoint will be provided with a HTTP code, a body and/or a response when necessary.

### DIFF
--- a/kpi/docs/api/v2/data/retrieve.md
+++ b/kpi/docs/api/v2/data/retrieve.md
@@ -16,3 +16,13 @@ or
 curl -X GET https://kf.kobotoolbox.org/api/v2/assets/{uid}/data/{id}/?format=xml
 curl -X GET https://kf.kobotoolbox.org/api/v2/assets/{uid}/data/{id}/?format=json
 ```
+
+
+### ⚠️ Note: DRF-Spectacular Limitation
+
+Due to limitations in **DRF-Spectacular**, the `ACCEPT` headers do not sync properly with the request. As a result, all responses will default to `application/json`, regardless of the specified format.
+
+This means that while alternative formats (like XML) are technically supported and will work via command-line tools (e.g., `curl`), **they will not work** when trying out the endpoint directly from the documentation page.
+
+We’ve still included the header to show supported formats, but keep in mind:
+**Only `application/json` will be used in the docs UI.**
--- a/kpi/docs/api/v2/data/list.md
+++ b/kpi/docs/api/v2/data/list.md
@@ -42,3 +42,15 @@ Question types are mapped to GeoJSON geometry types as follows:
 * `geopoint` to `Point`;
 * `geotrace` to `LineString`;
 * `geoshape` to `Polygon`.
+
+
+
+### ⚠️ Note: DRF-Spectacular Limitation
+
+Due to limitations in **DRF-Spectacular**, the `ACCEPT` headers do not sync properly with the request. As a result, all responses will default to `application/json`, regardless of the specified format.
+
+This means that while alternative formats (like XML) are technically supported and will work via command-line tools (e.g., `curl`), **they will not work** when trying out the endpoint directly from the documentation page.
+
+We’ve still included the header to show supported formats, but keep in mind:
+**Only `application/json` will be used in the docs UI.**
+


kpi/docs/api/v2/data/bulk_delete.md
```diff
@@ -0,0 +1,27 @@
...
+
+> **Payload**
+>
+>        {
+>           "submissions_ids": [
+>               1,
+>               2
+>           ]
+>        }
+* Where: "submissions_ids" (required) is a list of submission root id on the data
```
rajpatel24: How about using backticks instead of double-quotes? It would make the field names stand out more clearly.
"submissions_ids" -> `submissions_ids`
"number_of_submissions" -> `number_of_submissions`

kpi/docs/api/v2/data/bulk_delete.md
```diff
@@ -0,0 +1,27 @@
...
+to delete
+
+
+> **Response**
+>
+>        {
+>           "detail": "{number_of_submissions} submissions have been deleted"
+>        }
```
rajpatel24: We could avoid using `<` to stay consistent with the design. What do you think?

kpi/docs/api/v2/data/bulk_update.md
```diff
@@ -0,0 +1,15 @@
+## Bulk updating of submissions
+
+Where `<field_to_update_n>` is a string and should be an existing XML field value of the submissions.
```
rajpatel24, line 3: This is a bit confusing. You might have forgotten to include some code before 'Where <field_to_update_n> is a string...'. Can we update this? Possibly with an example?

kpi/docs/api/v2/data/validation_status_update.md
```diff
@@ -0,0 +1,6 @@
+## Update the validation of a submission
+
+where `<validation_status.uid>` is a string and can be one of these values:
+* `validation_status_approved`
+* `validation_status_not_approved`
+* `validation_status_on_hold`
```
rajpatel24: I think the wording can be improved slightly for clarity. How about changing it to something like this?

```
The validation status of the submission can be updated. The `validation_status.uid` should be a `string`, and it must be one of the following values:

- `validation_status_approved`
- `validation_status_not_approved`
- `validation_status_on_hold`
```

kpi/docs/api/v2/data/validation_statuses_delete.md
```diff
@@ -0,0 +1,30 @@
...
+  * `validation_status_approved`
+  * `validation_status_not_approved`
+  * `validation_status_on_hold`
+
+
+> **Response**
+>
+>        {
+>           "detail": "{number_of_submissions} submissions have been updated"
+>        }
```
rajpatel24: We could avoid using `<` to stay consistent with the design. What do you think?

kpi/docs/api/v2/data/validation_statuses_update.md
```diff
@@ -0,0 +1,15 @@
+## Bulk update submissions status
+
+* Where: "submissions_ids" (required) is a list of submission root id on the data
+to delete
+* Where: "validation_status.uid" (required)  is a string and can be one of these values:
+  * `validation_status_approved`
+  * `validation_status_not_approved`
+  * `validation_status_on_hold`
```
rajpatel24, line 8: Have you tried running this API with this payload? It gives me a 400 error every time.

I think the API expects the data to be wrapped inside a `payload` key at the root level. This is because the `DataBulkActionsValidator` serializer is defined with:

```
payload = WritableJSONField()
```

So, instead of sending the data like this:
```
{
  "submission_ids": [0],
  "validation_status.uid": "string"
}
```

It needs to be structured like this:
```
{
  "payload": {
    "submission_ids": [0],
    "validation_status.uid": "string"
  }
}
```

Could you take a look and try updating the payload and the description accordingly?

kpi/docs/api/v2/data/validation_statuses_delete.md
```diff
@@ -0,0 +1,30 @@
...
+>             2
+>           ],
+>           "validation_status.uid": <validation_status>
+>        }
+* Where: "submissions_ids" (required) is a list of submission root id on the data
+to delete
+* Where: "validation_status.uid" (required)  is a string and can be one of these values:
+  * `validation_status_approved`
+  * `validation_status_not_approved`
+  * `validation_status_on_hold`
```
rajpatel24: Does DELETE work with these strings?

I think the payload should look like this:

```
{
  "payload": {
    "submission_ids": [0],
    "validation_status.uid": "no_status"
  }
}
```

Please check my comment in `validation_statuses_update.md` :-)

rajpatel24 (changes requested):

kpi/docs/api/v2/data/bulk_delete.md
ZoeGodin: Changed the double-quotes into backticks

kpi/docs/api/v2/data/bulk_delete.md
ZoeGodin: Changed them so we're using markdown code text instead.

kpi/docs/api/v2/data/validation_status_update.md
ZoeGodin: Added more clarity to the patch documentation.

kpi/docs/api/v2/data/validation_statuses_delete.md
ZoeGodin: Replaced them with markdown example type for consistence.

kpi/docs/api/v2/data/bulk_update.md
ZoeGodin: There isn't supposed to be an example as the example is the payload. This is meant to add context to the fields user will see when trying or looking at the api on the documentation page.

kpi/docs/api/v2/data/validation_statuses_delete.md
ZoeGodin: The payload indeed takes the `payload` argument- changed the example and the validation status string documentation :)

kpi/docs/api/v2/data/validation_statuses_update.md
ZoeGodin: Indeed, the payload key is necessary. Changed it so the  payload key appears in the example

kpi/schema_extensions/v2/data/serializers.py
```diff
@@ -0,0 +1,93 @@
...
+    },
+)
+
+DataBulkUpdate = inline_serializer_class(
+    name='DataBulkUpdate',
+    fields={
+        'submission_ids': DataBulkDeleteField(),
+        'data': DataBulkUpdateField(),
+    },
+)
```
rajpatel24, line 26: The payload seems to be incorrect for this endpoint (PATCH `/data/bulk/`). It expects the data to be wrapped inside a `payload` key, like this:

```
{
  "payload": {
    "submission_ids": [
        "0",
        "1"
    ],
    "data": {
        "field_to_update": "string"
    }
  }
}
```

Would you mind having a look and updating it?  :-)

kpi/docs/api/v2/data/retrieve.md
```diff
@@ -0,0 +1,18 @@
...
+
+```shell
+curl -X GET https://kf.kobotoolbox.org/api/v2/assets/{uid}/data/{id}.xml
+curl -X GET https://kf.kobotoolbox.org/api/v2/assets/{uid}/data/{id}.json
+```
+
+or
+
+```shell
+curl -X GET https://kf.kobotoolbox.org/api/v2/assets/{uid}/data/{id}/?format=xml
```
rajpatel24, line 16: For this endpoint, selecting `xml` as the response format results in a `406 Not Acceptable` error. This happens because the `Accept` header in the generated curl remains as `application/json`, even when `?format=xml` is specified in the URL.

I’m not sure if there's a way to make Swagger UI automatically sync the `Accept` header with the selected format, do you know if that's possible?

If not, maybe we can just add a note in the description to clarify that users should manually set `Accept: application/xml` when requesting XML responses.

Please do this for the `list` endpoint as well.

kpi/schema_extensions/v2/data/serializers.py
ZoeGodin: Done, payload key has been added :)

kpi/docs/api/v2/data/retrieve.md
ZoeGodin: After much searching, there is no way to *automatically* sync them up- the only thing I have found that could have had the potential to work was the openApiParameter variable but it wasn't what we wanted. Instead, Ive added a note in the list and retrieve documentation.

ZoeGodin: After much searching, there is no way to automatically sync them up- the only thing I have found that could have had the potential to work was the openApiParameter variable but it wasn't what we wanted. Instead, Ive added a note in the list and retrieve documentation (to be confirmed).
results

rajpatel24 (approved):
LGTM

PR #5870, opened 2025-06-27
docs(api): refactoring of response util

ZoeGodin:
### 📣 Summary
Refactored response util to accept response description and custom error media_type if needed.


kpi/utils/schema_extensions/response.py
```diff
@@ -209,20 +223,22 @@ def open_api_generic_response(
...
     raise_not_found: bool = True,
+    error_media_type: Optional[str] = 'application/json',
     **kwargs,
 ) -> OpenApiGenericResponse:
     success_key = http_code
     if media_type:
         success_key = (http_code, media_type)
 
-    response = {success_key: OpenApiResponse(response=given_serializer)}
+    response = {success_key: OpenApiResponse(response=given_serializer, **kwargs)}
```
noliveleger: 🤔 , should not it be sanitarized? `**kwargs` seems to be related to `open_api_generic_response` and `open_api_error_responses`. What if, we pass something the `OpenApiResponse` does not expect?

noliveleger (approved):
LGTM

kpi/utils/schema_extensions/response.py
ZoeGodin: The reason why the kwargs is here is because we will sometime need the description for the response. It is the only place where it is needed but I can put a specific field for it.

Its most useful when there is no response but we still want to describe something that does happen, example:

``` 
 responses=open_api_200_ok_response(
            description='Will return a content type with the type of the attachment as well as the attachment itself.', 
            require_auth=False,
            raise_access_forbidden=False,
            validate_payload=False,
        ),
```

noliveleger: Let's only pass description then. To make sure we don't pass something to  `OpenApiResponse` it does not expect.

```
response = {success_key: OpenApiResponse(response=given_serializer, description=kwargs.get('description'))}
```

ZoeGodin: The line has been added and the error_media_type have been moved close to the media_type parameter for more clarity

noliveleger (approved):
LGTM

PR #5872, opened 2025-06-30
feat(billing): block asr/mt actions when usage limit exceeded DEV-489

jamesrkiger:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Blocks ASR/MT actions with a 402 payment required error if asset owner has exceeded the relevant ASR/MT usage limit.

### 📖 Description
Determines whether asset owner is over limit by calling `ServiceUsageCalculator.get_balances()` using the calculator's cache. So although it does add more queries to the ASRT/MT actions, these queries will only be run intermittently for any given asset owner.

### 💭 Notes

### 👀 Preview steps
1. Create a Stripe-enabled instance and sync stripe prices and products. For testing convenience set env var ENDPOINT_CACHE_DURATION to a small number like 10.
2. Use django-admin to edit the metadata "mt_characters_limit" for the default free community plan product (stripe id: "prod_EqgISHHelN2gXC")  to a small number.
3. In`/kpi/jsapp/js/account/usage/useUsage.hook.ts` on line 46, change `translationChars` to 0 (to bypass client-side restriction).
4. Create a new project with an audio question and make a submission with an audio file.
5. Write a manual transcript (e.g. "this is a test")
6. Request an automated translation.
7. Repeat until you have used up your usage balance.
8. Notice that you get an error message including the 402 error code.

### DIFF
--- a/kobo/apps/openrosa/libs/utils/logger_tools.py
+++ b/kobo/apps/openrosa/libs/utils/logger_tools.py
@@ -302,7 +302,7 @@ def create_instance(
                 from kobo.apps.stripe.utils.limit_enforcement import (
                     check_exceeded_limit,
                 )
-                
+
                 check_exceeded_limit(xform.user, UsageType.SUBMISSION)
                 check_exceeded_limit(xform.user, UsageType.STORAGE_BYTES)
 


kobo/apps/openrosa/apps/logger/tests/test_simple_submission.py
```diff
@@ -1,7 +1,10 @@
 import uuid
 
+import pytest
+from django.conf import settings
 from django.test import RequestFactory, TestCase
 from pyxform import SurveyElementBuilder
+from unittest.mock import patch
```
noliveleger: Wrong block, use `./format-python.sh <base branch>` to let this handle it for you. (be sure to have committed all your changes before running this command, in case you want to revert something).

kobo/apps/openrosa/libs/utils/logger_tools.py
```diff
@@ -88,6 +89,9 @@
 from kpi.utils.mongo_helper import MongoHelper
 from kpi.utils.object_permission import get_database_user
 
+if settings.STRIPE_ENABLED:
+    from kobo.apps.stripe.utils.limit_enforcement import check_exceeded_limit
+
```
noliveleger: I would like to avoid this kind of statement.
Could it possible to modify the decorator to support an optional parameter with would only return none or false. e.g. : `@requires_stripe(raise_error=False)`

We would avoid the double `if settings.STRIPE_ENABLED` here.

kobo/apps/trackers/tests/test_trackers.py
```diff
@@ -1,6 +1,11 @@
 from datetime import datetime
 
+import pytest
+from django.conf import settings
+from unittest.mock import patch
```
noliveleger: 🙈

kobo/apps/trackers/utils.py
```diff
@@ -69,6 +70,14 @@ def update_nlp_counter(
         **kwargs,
     )
 
+    if not deduct:
```
noliveleger: I know it does not come from this PR, but does this variable name sound weird only to me?

kobo/apps/openrosa/libs/utils/logger_tools.py
jamesrkiger: @noliveleger This change is actually from the base PR. I have [removed the double conditional](https://github.com/kobotoolbox/kpi/pull/5860/commits/290247ca628f0d19f3887aa069e1ba5c5079c8ec) (without adjusting requires_stripe).

Turns out I just needed to adjust the patch routes in the unit tests.

kobo/apps/trackers/utils.py
jamesrkiger: It's not just you. I don't like it either. I contemplated changing it, but to be honest I think I would be more inclined to just remove it entirely and just reference the setting var directly. `settings.STRIPE_ENABLED` isn't that verbose, imo. Either that or I change `deduct` to `stripe_enabled`. Do you have a preference?

noliveleger: The latter 💯

sksondra (commented):
Good from QA.

PR #5873, opened 2025-06-30
feat(billing): Replace useUsage with useServiceUsageQuery DEV-493

pauloamorimbr:
### 🗒️ Checklist

1. [x] run linter locally

### 📖 Description
This PR replaces the current `useUsage` by a react query implementation in `useServiceUsage` hook.
All uses of the current `useUsage` and `usageContext` were removed and replaced by the new hook.

### 👀 Preview steps
2. ℹ️ have stripe setup and an organization
3. Go to the account page
4. 🟢 All information regarding the usage should still be displayed properly
5. ℹ️ Be the organization owner
6. Go over limits (or simulate it by changing values in `useServiceUsage`)
7. Over limit banners should be displayed properly.


pauloamorimbr:
> Thanks, Paolo. This looks good to me for the most part. There's only one small change in behavior that I think it might be good to address: Because of RQ's caching, loading the usage page no longer necessarily fetches new data. Can we adjust the behavior so that on the usage page, specifically, we always refetch the data?

Done it.

For ensuring the updated data itself, I just lowered the stale time, since there aren't many calls spread around the app.

Removing the stale time made a multiple API call to pop up due to the limits banner also using the usage hook indirectly.
After some tracking down, I found a bug on `useSession` that was causing orgQuery to reload, causing a domino effect. 
But due to the components hierarchy the double call was still happening, so lowering the stale time to 1 second did the trick.

jamesrkiger (approved):
lgtm

PR #5877, opened 2025-07-01
ci: semi-automatic continuous bi-weekly releases DEV-299

Akuukis:
### 💭 Notes

Introduce a semi-automatic continuous bi-weekly release cycle. This PR scaffolds the plan, with automatic branch creation, several Zulip notifications and plenty of TODOs. Implementing those TODOs are out of scope of this PR, as they were done manually so far anyways, and will be implemented one-by-one by followup PRs.

### DIFF
--- a/.github/workflows/release-3-tag.yml
+++ b/.github/workflows/release-3-tag.yml
@@ -28,8 +28,8 @@ jobs:
       id: version
       run: |
         set -xe
-        BRANCH=git rev-parse --abbrev-ref HEAD
-        VERSION=`echo $BRANCH | sed -e 's/release\///'`
+        TAG=$GITHUB_REF
+        VERSION=$(echo "$TAG" | sed 's/v//')
         echo "version=$VERSION" >> "$GITHUB_OUTPUT"
 
   deploy-to-master:
@@ -53,8 +53,6 @@ jobs:
         ## see https://chat.kobotoolbox.org/#narrow/stream/4-Kobo-Dev/topic/2.2E024.2E33.20Release/near/472818
         ## Note: script does two things - deploys and updates docker-compose - see if can seperate that for easier reruns
         ## Note: this server has lots of data, run migrations or smtg with a 1 minute timeout to flag if some migrations run too long
-        VERSION=$(echo "v2.025.26" | sed 's/v//')
-        echo "$GITHUB_REF -> $VERSION"
         # pipenv shell
         # fab update_kpi:$VERSION
         # fab dev_deploy:master,kpi,$GITHUB_REF
--- a/.github/workflows/release-4-announce.yml
+++ b/.github/workflows/release-4-announce.yml
@@ -15,9 +15,7 @@ jobs:
     - name: Formats the version number
       id: version
       run: |
-        set -xe
-        BRANCH=git rev-parse --abbrev-ref HEAD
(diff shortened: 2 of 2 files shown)


.github/workflows/linear.yml
```diff
@@ -0,0 +1,27 @@
...
+      description:
+        required: true
+        type: string
+    secrets:
+      LINEAR_API_KEY_GITHUB_ACTIONS_BOT:
+        required: true
+
+jobs:
+  linear:
+    runs-on: ubuntu-22.04
```
noliveleger: can't it be on `24.04` like others?

.github/workflows/release-1-branch.yml
```diff
@@ -0,0 +1,79 @@
+name: 'CD: branch'
+
+on:
+  schedule:
+    # It's good to deploy early in week:
+    # - In the best case, will QA, tag and deploy on the same Tuesday or Wednesday.
+    # - In an ok case, will QA, tag and deploy on the next Monday-Wednesday.
+    # - In the worst case, the last chance is to deploy on after-next Monday.
+    - cron: '17 5 * * 2'
```
noliveleger, line 9: If I read this correctly, it's every Tueday at 5:17 UTC. 

I'm curious why '17 , not 15 or 30?

.github/workflows/release-2-stabilize.yml
```diff
@@ -0,0 +1,131 @@
...
+  changelog:
+    runs-on: ubuntu-24.04
+    steps:
+    - name: post to Zulip Linear updated draft of changelog
+      run: |
+        echo 'TODO, for now generate changelog manually'
+        echo 'TODO, for now send to Zulip manually'
+
+  deploy-to-beta:
+    runs-on: ubuntu-22.04
```
noliveleger: same question? Why do we mix both versions of ubuntu? any problem to keep `24.04` everywhere?

.github/workflows/release-3-tag.yml
```diff
@@ -0,0 +1,118 @@
...
+    - uses: actions/checkout@v4
+
+    - name: deploy to master
+      run: |
+        echo 'TODO, for now deploy manually on kf.master.kbtdev'
+        ## see https://github.com/kobotoolbox/kobo-deployments
+        ## see https://chat.kobotoolbox.org/#narrow/stream/4-Kobo-Dev/topic/2.2E024.2E33.20Release/near/472818
+        ## Note: script does two things - deploys and updates docker-compose - see if can seperate that for easier reruns
+        ## Note: this server has lots of data, run migrations or smtg with a 1 minute timeout to flag if some migrations run too long
+        VERSION=$(echo "v2.025.26" | sed 's/v//')
```
noliveleger: Even if it runs manually, we could already detect the correct version?

.github/workflows/release-1-branch.yml
Akuukis: Github recommends so. It's a random non-round number, because most of world run their jobs on rounds numbers and the github actions spikes and queues some of them for soon after, even leading to timeouts/cancels. This should be more reliable.

.github/workflows/release-3-tag.yml
Akuukis: oops, a debug leftover

PR #5882, opened 2025-07-02
feat(billing): reset usage/limit data on subscription change DEV-486

jamesrkiger:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Clears usage calculator cache and updates or removes counters as necessary whenever a subscription is saved.

### 💭 Notes

### DIFF
--- a/kobo/apps/stripe/__init__.py
+++ b/kobo/apps/stripe/__init__.py
@@ -1,0 +1,13 @@
+# coding: utf-8
+from django.apps import AppConfig
+
+
+class StripeAppConfig(AppConfig):
+    name = 'kobo.apps.stripe'
+    verbose_name = 'Stripe'
+
+    def ready(self):
+        # Makes sure all signal handlers are connected
+        from kobo.apps.stripe import signals
+
+        super().ready()
--- a/kobo/apps/stripe/models.py
+++ b/kobo/apps/stripe/models.py
@@ -5,8 +5,6 @@ from django.core.exceptions import ObjectDoesNotExist
 from django.db import models
 from django.db.models import IntegerField, Q, Sum
 from django.db.models.functions import Cast, Coalesce
-from django.db.models.signals import post_save
-from django.dispatch import receiver
 from djstripe.enums import PaymentIntentStatus
 from djstripe.models import Charge, Price, Subscription
 
@@ -306,11 +304,6 @@ class PlanAddOn(models.Model):
         return self.product.metadata.get('valid_tags', '').split(',')
 
 
-@receiver(post_save, sender=Charge)
-def make_add_on_for_charge(sender, instance, created, **kwargs):
-    PlanAddOn.create_or_update_one_time_add_on(instance)
-
-
 class ExceededLimitCounter(models.Model):
     user = models.ForeignKey(
         User,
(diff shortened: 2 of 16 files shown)


kobo/apps/stripe/__init__.py
```diff
@@ -0,0 +1,13 @@
+# coding: utf-8
```
noliveleger: `# coding: utf-8` can be safely removed.

Actually I'm removing them one by one from actual code. It's not needed in Python3 anymore. (@bufke would say: "I told you! ;-)")

kobo/apps/stripe/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+
+class StripeAppConfig(AppConfig):
+    name = 'kobo.apps.stripe'
+    verbose_name = 'Stripe'
+
+    def ready(self):
+        # Makes sure all signal handlers are connected
+        from kobo.apps.stripe import signals  # noqa
+
+        super().ready()
```
noliveleger: I want to make all Django apps consistant and move this logic in the proper file "apps.py" at some point.

Would you mind to do the same (and update `base.py` accordingly) for this app?

kobo/apps/stripe/models.py
```diff
@@ -306,11 +304,6 @@ def valid_tags(self) -> List:
         return self.product.metadata.get('valid_tags', '').split(',')
 
 
-@receiver(post_save, sender=Charge)
-def make_add_on_for_charge(sender, instance, created, **kwargs):
-    PlanAddOn.create_or_update_one_time_add_on(instance)
-
-
```
noliveleger, line 313: ❤️

kobo/apps/stripe/signals.py
```diff
@@ -0,0 +1,36 @@
...
+
+    user_id = subscription_metadata.get('kpi_owner_user_id', '')
+    if not user_id:
+        return
+
+    user = User.objects.filter(id=user_id).first()
+    if not user:
+        return
+
+    ServiceUsageCalculator(user)._clear_cache()
```
noliveleger: if we need to call `_clear_cache()` outside the scope of the class. Let's make it a "public" method, let's remove the '_'. Could be in another (in-between) PR since we need to refactor this at several place.

kobo/apps/stripe/tests/test_stripe_utils.py
```diff
@@ -694,3 +699,30 @@ def test_check_exceeded_limit_cache_restriction(self):
             ) as patched:
                 check_exceeded_limit(self.someuser, UsageType.SUBMISSION)
                 patched.assert_called_once()
+
+    def test_update_or_remove_limit_counter(self):
+        mock_balances = {
+            UsageType.SUBMISSION: {'exceeded': True},
+        }
+        counter = baker.make(
```
noliveleger, line 707: Does it really make fast to use `baker.make()` rather than `ExceededLimitCounter.objects.create()`?

jamesrkiger: I can't say I know too much about the performance comparisons. I doubt we're going to be saving much time here either way. I tend to use model bakery by default when it's available, since it's nice not have to think about every field that might not be optional in a model.

noliveleger: I think the opposite because it can silently hide a problem.

noliveleger (approved):
LGTM

PR #5894, opened 2025-07-07
ci: update job filters

Akuukis:
### 💭 Notes

Update job filters and be more strict about running workflows.


.github/filters.yml
```diff
@@ -3,6 +3,11 @@
 # If any file changed matches any of the filters in a list, the output for that
 # list is set to the string 'true'.
 
+all:
+  - '.github/filters.yml'          # ci
```
noliveleger, line 7: is it ok to include itself?

noliveleger: is it ok to include itself?

PR #5881, opened 2025-07-02
ci: upgrade to ubuntu-24.04 for all jobs DEV-647

Akuukis:
### 🗒️ Checklist

1. [ ] run linter locally
2. [ ] update developer docs (API, README, inline, etc.), if any
3. [ ] for user-facing doc changes create a Zulip thread at #Kobo support docs, if any
4. [ ] draft PR with a title `<type>(<scope>)<!>: <title> DEV-1234`
5. [ ] assign yourself, tag PR: at least `Front end` and/or `Back end` or `workflow`
6. [ ] fill in the template below and delete template comments
7. [ ] review thyself: read the diff and repro the preview as written
8. [ ] open PR & confirm that CI passes & request reviewers, if needed
9. [ ] delete this section before merging

### 💭 Notes

Upgrade from ubuntu-22.04 to ubuntu-24.04 for all jobs.

### 👀 Preview steps

See that CI passes.

PR #5895, opened 2025-07-07
ci: share playwright cache across branches

Akuukis:
### 💭 Notes

Currently playwright is cached per-branch. Let's share the cache globally, that will save a minute per PR.

Note: wait few weeks after #5881 so that all branches align to ubuntu 24.04. Github's `runner.os` gives "Linux" without a version, so sadly can't nicely cache per-OS.


magicznyleszek (approved):
Looks good to my inexperienced eyes :)

PR #5898, opened 2025-07-08
ci(releases): use ISO week number DEV-299

Akuukis:
### 💭 Notes

Use ISO week number. Also, don't notify when not creating a branch and add a manual trigger to re-run this after a merge.

### DIFF
--- a/.github/workflows/release-1-branch.yml
+++ b/.github/workflows/release-1-branch.yml
@@ -1,6 +1,7 @@
 name: 'CD: branch'
 
 on:
+  workflow_dispatch:
   schedule:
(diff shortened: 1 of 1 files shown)


.github/workflows/release-1-branch.yml
```diff
@@ -23,7 +24,7 @@ jobs:
       id: week
       run: |
         set -xe
-        if (( $(date +%U) % 2 == 0 )); then
+        if (( $(date +%V) % 2 == 0 )); then
```
noliveleger, line 27: Good catch!

noliveleger (approved):
LGTM

PR #5902, opened 2025-07-08
docs(api): adding API endpoints for versions in KPI documentation DEV-536

ZoeGodin:
### 📣 Summary
Added API documentation for `/api/v2/docs/versions` endpoint.

### 📖 Description
This PR introduces the endpoint documentation that will be used for the auto-generating API documentation of the `versions` endpoint. With these changes, users will be able to understand, test and use the `versions` endpoint more easily.

### 👀 Preview steps
To see the changes, visit `http://kf.kobo.local/api/v2/docs/`. The page of the Kobo API will now be visible and generated by swagger. To see the changes added to `versions` endpoint, scroll down to the category of the same name or make a research in the search bar. Documentation for the endpoint will be provided with a HTTP code, a body and/or a response when necessary.

### DIFF
--- a/kpi/views/v2/asset_version.py
+++ b/kpi/views/v2/asset_version.py
@@ -1,5 +1,5 @@
 # coding: utf-8
-from drf_spectacular.utils import extend_schema
+from drf_spectacular.utils import extend_schema, extend_schema_view
 from rest_framework import viewsets
 from rest_framework_extensions.mixins import NestedViewSetMixin
 
@@ -13,7 +13,19 @@ from kpi.utils.viewset_mixins import AssetNestedObjectViewsetMixin
 
 
 @extend_schema(
-    tags=['versions'],
+    tags=['Versions'],
+)
+@extend_schema_view(
+    create=extend_schema(),
+    destroy=extend_schema(),
+    list=extend_schema(),
+    update=extend_schema(
+        exclude=True,
+    ),
+    retrieve=extend_schema(),
+    partial_update=extend_schema(
+        exclude=True,
+    ),
 )
 class AssetVersionViewSet(AssetNestedObjectViewsetMixin,
                           NestedViewSetMixin, viewsets.ModelViewSet):


kpi/schema_extensions/v2/versions/extensions.py
```diff
@@ -0,0 +1,87 @@
...
+from kpi.utils.schema_extensions.url_builder import build_url_type
+
+
+class ContentFieldExtension(OpenApiSerializerFieldExtension):
+    target_class = 'kpi.schema_extensions.v2.versions.fields.ContentField'
+
+    def map_serializer_field(self, auto_schema, direction):
+        return build_object_type(
+            properties={
+                'schema': build_basic_type(OpenApiTypes.INT),
```
rajpatel24: I think this is a string, not an integer. The API response returns `"1"` (in quotes), so it might be worth double-checking the expected type.

kpi/schema_extensions/imports.py
```diff
@@ -6,4 +6,5 @@
 import kpi.schema_extensions.v2.data.extensions
 import kpi.schema_extensions.v2.export_settings.extensions
 import kpi.schema_extensions.v2.openrosa.extensions
-import kpi.schema_extensions.v2.paired_data.extensions
+import kpi.schema_extensions.v2.versions.extensions
+import kpi.schema_extensions.v2.paired_data.extensions
```
rajpatel24: Let’s stick to alphabetical order

rajpatel24 (changes requested):

kpi/schema_extensions/v2/versions/extensions.py
ZoeGodin: Indeed, it is a string- changed it so it reflects the correct type.

kpi/schema_extensions/imports.py
ZoeGodin: Reordered the imports

rajpatel24 (approved):
LGTM

PR #5903, opened 2025-07-08
fix(table): Improve the table error message DEV-692

duvld:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Stops the table view from spitting out raw html when an error occurs. Replaces it with a more user friendly error

### 💭 Notes
Utilizes `handleApiFail` to display a helpful toast with the specific error "plucked" out. Changes the error handling JSX code in the table without getting too complicated and editing other files.

### 👀 Preview steps

2. submit some data
3. go to `data.py` and under the `def list ...` line add `1/0` to throw an error
4. navigate to the data tab
5. 🔴 [on release] notice that the error message is raw html
6. 🟢 [on PR] notice that the error message is more user friendly

### DIFF
--- a/jsapp/js/components/submissions/table.tsx
+++ b/jsapp/js/components/submissions/table.tsx
@@ -1321,7 +1321,7 @@ export class DataTable extends React.Component<DataTableProps, DataTableState> {
 
   render() {
     if (this.state.error && typeof this.state.error === 'string') {
-      const supportMessage = `Please try again later, or [contact the support team](##SUPPORT_URL##) if this happens repeatedly.`
+      const supportMessage = t('Please try again later, or [contact the support team](##SUPPORT_URL##) if this happens repeatedly.')
       return (
         <bem.FormView m='ui-panel'>
           <CenteredMessage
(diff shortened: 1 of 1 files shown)


jsapp/js/components/submissions/table.tsx
```diff
@@ -1331,7 +1323,24 @@ export class DataTable extends React.Component<DataTableProps, DataTableState> {
...
+          <CenteredMessage
+            message={
+              <div>
+                <h2>{t('Oops! Something went wrong on our end.')}</h2>
+                <div>
+                  Please try again later, or{' '}
+                  <a href={envStore.data.support_url} target='_blank'>
+                    contact the support team
+                  </a>{' '}
+                  if this happens repeatedly.
```
pauloamorimbr: We should add a translation for this as well. There are examples using markdown here with translations [here](https://github.com/kobotoolbox/kpi/blob/0cab89d48c01785d7ca9b2827c6c770315c9fcaf/jsapp/js/components/usageLimits/overLimitBanner.component.tsx#L178) so the link can be inside of the translation as well all in 1 single phrase.

pauloamorimbr (changes requested):

PR #5904, opened 2025-07-08
fix(dataTable): prevent 500 error when opening data table view DEV-695

noliveleger:
### 📣 Summary
Prevent a 500 error when opening data table view when the project contains versions created before the NLP feature.

### 📖 Description
This PR resolves a 500 Internal Server Error that could occur when opening the data table view for projects with old versions.
The issue was caused by a TypeError when the system attempted to access a `$xpath` property that doesn’t exist on versions created/deployed before the introduction of the NLP feature.

### 👀 Preview steps

1. Follow steps from #5862 up to step 4
2. From the shell (or from the DB directly with pgAdmin, psql or dbeaver) remove the `$xpath` property from `version_content` field for every question in the first version of your project.
3. Go to DATA > Table
4. 🔴 [on current release] Receive a 500 (look at the Python console, it prints `TypeError: 'NoneType' object is not iterable`
5. 🟢 [on PR] The data appears as expected

### Related issues
kpi#5862

### DIFF
--- a/kpi/models/asset.py
+++ b/kpi/models/asset.py
@@ -666,20 +666,13 @@ class Asset(
         # return deployed versions first
         versions = self.asset_versions.filter(deployed=True).order_by('-date_modified')
         xpaths = set()
-        for i, version in enumerate(versions):
-            # insert the xpaths if this is the latest deployed version
-            insert_xpath = i == 0
+        for version in versions:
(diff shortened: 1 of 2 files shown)


RuthShryock (approved):
Tested and lgtm

PR #5909, opened 2025-07-09
feat(attachments): Add Celery task to process users exceeding storage limit DEV-239

rajpatel24:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Adds a new celery task that identifies users who have exceeded their storage limits for a configurable number of days and dispatches a cleanup task for each. This prepares the foundation for automated enforcement of storage quotas via attachment deletion.

### 📖 Description
This PR introduces a Celery task (`dispatch_attachment_cleanup_for_users_exceeding_limits`) that:

- Runs periodically via Celery Beat (every 30 mins, currently commented out in `CELERY_BEAT_SCHEDULE`)
- Filters users from the `ExceededLimitCounter` model who have exceeded their storage quota for more than `STORAGE_OVERAGE_ATTACHMENT_DELETION_GRACE_PERIOD` days (default: 90)
- For each qualifying user, dispatches a cleanup task (`auto_delete_excess_attachments(user_id)`), currently a placeholder.

Additional Constance settings added:

- `AUTO_DELETE_ATTACHMENTS`: Global toggle for enabling/disabling the auto deletion (default: False)
- `STORAGE_OVERAGE_ATTACHMENT_DELETION_GRACE_PERIOD`: Grace period before triggering cleanup for over-quota users

Notes for reviewers:

- Task is not yet active: The beat schedule is commented out in settings and will be enabled once the attachment deletion logic is implemented in `auto_delete_excess_attachments()`
- Unit tests will be added along with the implementation of `auto_delete_excess_attachments`, when the full deletion pipeline is built. 
- The `ExceededLimitCounter` model is currently not updated automatically. Integration with periodic quota tracking will be handled later in the billing project.

### DIFF
--- a/kobo/apps/trash_bin/tasks/attachment.py
+++ b/kobo/apps/trash_bin/tasks/attachment.py
@@ -2,11 +2,15 @@ import logging
 
 from celery.exceptions import SoftTimeLimitExceeded, TimeLimitExceeded
 from celery.signals import task_failure, task_retry
+from constance import config
 from django.conf import settings
+if settings.STRIPE_ENABLED:
+    from kobo.apps.stripe.models import ExceededLimitCounter
 
 from kobo.celery import celery_app
 from ..exceptions import TrashTaskInProgressError
 from ..models.attachment import AttachmentTrash
+from ...organizations.constants import UsageType
 from ..utils import process_deletion, trash_bin_task_failure, trash_bin_task_retry
 from ..utils.attachment import delete_attachment
 
@@ -53,3 +57,28 @@ def empty_attachment_failure(sender=None, **kwargs):
 @task_retry.connect(sender=empty_attachment)
 def empty_attachment_retry(sender=None, **kwargs):
     trash_bin_task_retry(AttachmentTrash, **kwargs)
+
+
+# @celery_app.task
+def schedule_auto_attachment_cleanup_for_users():
+    """
+    Identifies users exceeding storage limits beyond the grace period and
+    schedules a cleanup task for each.
+
+    Runs only if AUTO_DELETE_ATTACHMENTS and Stripe billing is enabled.
+    """
+    if not config.AUTO_DELETE_ATTACHMENTS or not settings.STRIPE_ENABLED:
+        return
+
+    exceeded_counters = ExceededLimitCounter.objects.filter(
+        limit_type=UsageType.STORAGE_BYTES,
+        days__gte=config.STORAGE_OVERAGE_ATTACHMENT_DELETION_GRACE_PERIOD
+    ).select_related('user')
+
+    for counter in exceeded_counters:
+        auto_delete_excess_attachments(counter.user.pk)
+
+
+def auto_delete_excess_attachments(user_id: int):
+    # ToDo: Implement the logic to auto-delete excess attachments for the user.
+    pass
--- a/kobo/settings/base.py
+++ b/kobo/settings/base.py
@@ -458,6 +458,16 @@ CONSTANCE_CONFIG = {
         'them and before automatically hard-deleting them by the system',
         'positive_int',
     ),
+    'AUTO_DELETE_ATTACHMENTS': (
+        False,
+        'Enable automatic deletion of attachments for users who have exceeded '
+        'their storage limits.'
+    ),
+    'STORAGE_OVERAGE_ATTACHMENT_DELETION_GRACE_PERIOD': (
+        90,
+        'Number of days to keep attachments after the user has exceeded their '
+        'storage limits.'
+    ),
     # Toggle for ZXCVBN
     'ENABLE_PASSWORD_ENTROPY_METER': (
         True,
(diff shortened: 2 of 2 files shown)


kobo/apps/trash_bin/tasks/attachment.py
```diff
@@ -2,11 +2,15 @@
 
 from celery.exceptions import SoftTimeLimitExceeded, TimeLimitExceeded
 from celery.signals import task_failure, task_retry
+from constance import config
 from django.conf import settings
+if settings.STRIPE_ENABLED:
+    from kobo.apps.stripe.models import ExceededLimitCounter
```
noliveleger: This import should not be in this block. 
BTW, don't we have a decorator for methods which require stripe to avoid this kind of conditional import?

kobo/apps/trash_bin/tasks/attachment.py
```diff
@@ -2,11 +2,15 @@
...
 
 from kobo.celery import celery_app
 from ..exceptions import TrashTaskInProgressError
 from ..models.attachment import AttachmentTrash
+from ...organizations.constants import UsageType
```
noliveleger: three-dots import should not be in the middle of two-dots imports ;-)

kobo/apps/trash_bin/tasks/attachment.py
```diff
@@ -53,3 +57,28 @@ def empty_attachment_failure(sender=None, **kwargs):
...
+    if not config.AUTO_DELETE_ATTACHMENTS or not settings.STRIPE_ENABLED:
+        return
+
+    exceeded_counters = ExceededLimitCounter.objects.filter(
+        limit_type=UsageType.STORAGE_BYTES,
+        days__gte=config.STORAGE_OVERAGE_ATTACHMENT_DELETION_GRACE_PERIOD
+    ).select_related('user')
+
+    for counter in exceeded_counters:
+        auto_delete_excess_attachments(counter.user.pk)
```
noliveleger: I think it could be possible that it takes more than 30 minutes to delete some attachments. Thus, we should make `auto_delete_excess_attachments` also a task and delegate it to Celery too.

kobo/apps/trash_bin/tasks/attachment.py
```diff
@@ -53,3 +57,28 @@ def empty_attachment_failure(sender=None, **kwargs):
...
+
+
+def auto_delete_excess_attachments(user_id: int):
+    # ToDo: Implement the logic to auto-delete excess attachments for the user.
+    pass
```
noliveleger, line 88: Make sure the implementation take into account, we should not run it if it is also already running (maybe a lock with Redis?)

kobo/settings/base.py
```diff
@@ -757,7 +767,9 @@
     'Trash bin': (
         'ACCOUNT_TRASH_GRACE_PERIOD',
         'ATTACHMENT_TRASH_GRACE_PERIOD',
+        'AUTO_DELETE_ATTACHMENTS',
```
noliveleger: Let's leave all the `_GRACE_PERIOD` together. Move `AUTO_DELETE_ATTACHMENTS`

kobo/settings/base.py
```diff
@@ -458,6 +458,16 @@
         'them and before automatically hard-deleting them by the system',
         'positive_int',
     ),
+    'AUTO_DELETE_ATTACHMENTS': (
+        False,
+        'Enable automatic deletion of attachments for users who have exceeded '
+        'their storage limits.'
+    ),
+    'STORAGE_OVERAGE_ATTACHMENT_DELETION_GRACE_PERIOD': (
```
noliveleger: I think the name should be `LIMIT_ATTACHMENT_REMOVAL_GRACE_PERIOD`

kobo/apps/trash_bin/tasks/attachment.py
noliveleger: what about using `counter.user_id` and not using `select_related('user')` if we don't need anything else to `User` model

noliveleger (changes requested):

kobo/apps/trash_bin/tasks/attachment.py
rajpatel24: Yes, we have decorator `@requires_stripe`. I have updated the code to use that.

kobo/apps/trash_bin/tasks/attachment.py
rajpatel24: Implemented in [kpi#5938](https://github.com/kobotoolbox/kpi/pull/5938)

kobo/apps/trash_bin/tasks/attachment.py
rajpatel24: Implemented in [kpi#5938](https://github.com/kobotoolbox/kpi/pull/5938)

kobo/apps/trash_bin/tasks/attachment.py
rajpatel24: You're right! We don't need any extra details. Updated the code accordingly.

PR #5918, opened 2025-07-14
fix(table): show hard coded error message for `500`'s

duvld:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Fixes a regression introduced in #5903 where the error message intended for `500`'s was displaying for empty forms, undeployed forms, and/or deployed forms with no data. Reverts the behaviour to show the response text in these cases

### 👀 Preview steps

1. ℹ️ have an account and 2 projects
2. leave both projects empty
3. deploy one project
4. check the data table for both projects
6. 🔴 [on main] notice that the "Oops!" message exists for both these projects
7. 🟢 [on PR] notice that the error messages are correct and relevant to the issue (not deployed, or no submissions)
8. add `1/0` to line 444 on `data.py`
9. 🟢 notice that the data table still does not show an a mess of html
10. remove the `1/0` in `data.py`
11. submit some data in a form, or navigate to a form with data
9. 🟢 notice that the data table displays as expected

### DIFF
--- a/jsapp/js/components/submissions/table.tsx
+++ b/jsapp/js/components/submissions/table.tsx
@@ -113,6 +113,8 @@ interface DataTableState {
   pageSize: number
   currentPage: number
   error: string | boolean
+  errorNumber: number | null
+  errorStatus: string | null
   showLabels: boolean
   translationIndex: number
   showGroupName: boolean
@@ -161,6 +163,8 @@ export class DataTable extends React.Component<DataTableProps, DataTableState> {
       pageSize: 30,
       currentPage: 0,
       error: false,
+      errorNumber: null,
+      errorStatus: null,
       showLabels: true,
       translationIndex: 0,
       showGroupName: true,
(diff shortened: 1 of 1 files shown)


jsapp/js/components/submissions/table.tsx
```diff
@@ -1320,7 +1347,7 @@ export class DataTable extends React.Component<DataTableProps, DataTableState> {
   }
 
   render() {
-    if (this.state.error && typeof this.state.error === 'string') {
+    if (this.state.errorNumber && this.state.errorNumber === 500) {
```
pauloamorimbr: I believe you can simplify this. State will always exist, and if errorNumber is undefined or null...or anything other than 500 it won't go inside the if.

```suggestion
    if (this.state.errorNumber === 500) {
```

jsapp/js/components/submissions/table.tsx
```diff
@@ -1344,6 +1371,14 @@ export class DataTable extends React.Component<DataTableProps, DataTableState> {
       )
     }
 
+    if (this.state.error && typeof this.state.error === 'string') {
```
pauloamorimbr: Same thing here... if the state.error is undefined or null the type won't be string anyways.

```suggestion
    if (typeof this.state.error === 'string') {
```

jsapp/js/components/submissions/table.tsx
```diff
@@ -326,12 +330,35 @@ export class DataTable extends React.Component<DataTableProps, DataTableState> {
...
-    if (error?.responseText) {
-      handleApiFail(error)
+    if (error?.status) {
+      this.setState({ errorNumber: error.status })
+
+      if (error.status === 500 && error?.responseText) {
+        handleApiFail(error)
+      }
+
+      // 500's from the backend gives us an html responseText, so we show a hard coded string during render instead
```
pauloamorimbr: This comment seems wrong. Did you mean 400? 🤔

jsapp/js/components/submissions/table.tsx
```diff
@@ -326,12 +330,35 @@ export class DataTable extends React.Component<DataTableProps, DataTableState> {
   }
 
   onGetSubmissionsFailed(error: FailResponse) {
```
pauloamorimbr, line 335: I think some comments here would help understanding the logic for someone who doesn't know what BE is returning in the different situations.

jsapp/js/components/submissions/table.tsx
duvld: No it's 500's that give the html response, I'm leaving this comment explaining the following lines where I exclude 500's from plucking out a "displayedErr"

duvld: I gave a more detailed comment for the method itself, maybe that makes it clearer 😅

PR #5920, opened 2025-07-15
fix(auditLogs): catch TypeErrors from copying request data DEV-711

rgraber:
### 📣 Summary
Fix an error that was preventing submissions with large uploads.

### 💭 Notes
Fixes an issue wherein submissions with large uploaded files attached were failing. This was due to an issue in the audit log code, where we copied request.data to the internal WSGI request for later processing by the middleware. The bug emerged because Django uses `TemporaryUploadedFile`s when uploading files larger than the configured max size, and those cannot be copied using deepcopy.
This PR just catches that error, logs it, and moves on. It's not the most foolproof of solutions if we ever need to log those requests but for now we don't and this is blocking a fairly urgent fix (DEV-708). It also means we don't have to go into the request.data() dictionary to figure out what is in there before trying to copy it. Since request.data() could be fairly large it will save some time.

### 👀 Preview steps

1. ℹ️ have an account and a project with an upload question
2. Update FILE_UPLOAD_MAX_MEMORY_SIZE to 100
3. Add a submission including a file
4. 🔴 [on release] The submission fails with `TypeError: cannot pickle '_io.BufferedRandom' object`
5. 🟢 [on PR] The submission goes through

### DIFF
--- a/kobo/apps/audit_log/base_views.py
+++ b/kobo/apps/audit_log/base_views.py
@@ -1,6 +1,7 @@
 from rest_framework import mixins, viewsets
 from rest_framework.views import APIView
 
+from kpi.utils.log import logging
 from kpi.utils.viewset_mixins import AssetNestedObjectViewsetMixin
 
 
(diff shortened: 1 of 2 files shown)


rajpatel24 (approved):
LGTM

PR #5922, opened 2025-07-16
feat(user): include uid in /me endpoint DEV-234

rgraber:
### 🗒️ Checklist

1. [ ] run linter locally

### 📣 Summary
Add user uid to the current user endpoint.

### 💭 Notes
Adds the uid from extra_details into the /me response. We can't nest it under the existing extra_details field because that is a writeable JSON used for updating extra_details.data. In an ideal world we would have the extra_details field return a dictionary with both the uid and the data JSON and only make the data part writeable, but  that would require a new version of the API since it would break existing contracts. 

### 👀 Preview steps

1. ℹ️ have an account
2. Go to `/me`
3. 🟢 Notice your uid is returned in the response under `extra_details__uid`

### DIFF
--- a/kobo/apps/long_running_migrations/app.py
+++ b/kobo/apps/long_running_migrations/app.py
@@ -99,4 +99,4 @@ def check_must_complete_long_running_migrations(app_configs, **kwargs):
     ]
 
 
-#register(check_must_complete_long_running_migrations, Tags.database)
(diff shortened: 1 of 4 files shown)


Guitlle (approved):
LGTM

sksondra (commented):
Good from QA.

PR #5933, opened 2025-07-17
feat(billing): recurring task to update limit counters DEV-487

jamesrkiger:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Adds a recurring task to check usage and update or remove exceeded limit counters that are more than one day old and not modified in the past day to ensure that counters remain up-to-date.

### 👀 Preview steps
I have added a unit test for this task. There are currently no user-facing impacts or UI effects of usage counters, so they are best tested in python.

### DIFF
--- a/kobo/apps/stripe/tasks.py
+++ b/kobo/apps/stripe/tasks.py
@@ -21,6 +21,10 @@ def update_exceeded_limit_counters():
 
 @celery_app.task
 def update_counter(counter_id):
-    counter = ExceededLimitCounter.objects.filter(id=counter_id).first()
+    counter = (
+        ExceededLimitCounter.objects.filter(id=counter_id)
+        .select_related('user')
+        .first()
+    )
     if counter:
         update_or_remove_limit_counter(counter)
--- a/kobo/settings/base.py
+++ b/kobo/settings/base.py
@@ -1322,6 +1322,8 @@ CELERY_BEAT_SCHEDULE = {
 }
 
 if STRIPE_ENABLED:
+    # Schedule to run once per celery timeout
+    # with a five minute buffer
     CELERY_BEAT_SCHEDULE['update-exceeded-limit-counters'] = {
         'task': 'kobo.apps.stripe.tasks.update_exceeded_limit_counters',
         'schedule': timedelta(seconds=CELERY_TASK_TIME_LIMIT + (60 * 5)),


kobo/apps/stripe/tests/test_celery_tasks.py
```diff
@@ -0,0 +1,52 @@
...
+        cls.first_organization = baker.make(Organization, id='123456abcdef')
+        cls.someuser = User.objects.get(username='someuser')
+        cls.first_organization.add_user(cls.someuser, is_admin=True)
+
+    def test_clear_usage_cache_and_counters_on_save(self):
+        """
+        Ensure that task runs updates for counters more than one day old
+        and not modified in past day
+        """
+        test_counter = baker.make(
```
jamesrkiger, line 29: I wanted to use create() to make Olivier happy, but unfortunately in this case it would have been overly complicated because of the auto_add date fields so I decided to stick with bakery since it bypasses those.

kobo/settings/base.py
```diff
@@ -1321,6 +1321,12 @@ def dj_stripe_request_callback_method():
     }
 }
 
+if STRIPE_ENABLED:
+    CELERY_BEAT_SCHEDULE['update-exceeded-limit-counters'] = {
+        'task': 'kobo.apps.stripe.tasks.update_exceeded_limit_counters',
+        'schedule': timedelta(seconds=CELERY_TASK_TIME_LIMIT + (60 * 5)),
```
noliveleger, line 1329: @jamesrkiger would you mind to explain when you want to run it explicitly, please?

kobo/apps/stripe/tests/test_celery_tasks.py
noliveleger: 😢

kobo/apps/stripe/tasks.py
```diff
@@ -0,0 +1,26 @@
...
+    )
+
+    for counter in qs:
+        # Pass in id because celery can't take the django object as an arg
+        update_counter.delay(counter.id)
+
+
+@celery_app.task
+def update_counter(counter_id):
+    counter = ExceededLimitCounter.objects.filter(id=counter_id).first()
```
noliveleger: We should use `select_related('user')` since `update_or_remove_limit_counter` is always calling the related `User` object.

kobo/apps/stripe/tasks.py
```diff
@@ -0,0 +1,26 @@
...
+    if counter:
+        update_or_remove_limit_counter(counter)
```
noliveleger, line 30: I guess you use `if counter:` in case a previous could have deleted the counter, but I'm wondering if it's enough. 🤔

jamesrkiger: Based on our discussion last week, I'm leaving his as is. In the rare event that a counter is deleted elsewhere while one of these checks is in progress, it is fine if we run an update() that doesn't do anything.  I'm still leaving this check if statement in here, though, because we do need the user from the counter before we run the update.

noliveleger (approved):
LGTM

PR #5937, opened 2025-07-17
feat(usageLimits): Add limit banners to new pages DEV-443

pauloamorimbr:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
This PR adds the service limits banner to the following views:
- Project summary
- Project form
- Project data
- Project settings (’general’ sub-section)

### 💭 Notes

To add the limits banner it took a different strategy for each view, since all around the app the code is still missing some patterns.

**Form Summary:** It was needed to add a container for the banner and separate the panels

**Form Landing:** Just adding the banner component worked nicely

**Form Data:** This is the most sensible one. Since there are a lot of calculations already in place for the table and containers, I decided to add the banner and rework the minimum possible the CSS for the tabele to keep behaving as it should. There's also a small CSS 'hack' to remove margins from the banner component. Not ideal, but the other option would be a massive rework of the component, which is not fit for this PR.

**Form Settings:** Some CSS needed to be moved so the banner would fit properly in the expected margins.

⚠️ There are some CSS and text changes when looking at the Figma design, but changing the banner colors could also mean changing the global CSS variables being used, so I decided to not change those. As for the texts, I believe it's a matter for another task, if needed, taking in account some specific definitions we would need to get from Product.

### 👀 Preview steps
2. Navigate through the project views
3. Mock service usage data if needed
4. 🟢 Banner should be displayed properly, when needed, without breaking any view

### DIFF
--- a/jsapp/js/components/modalForms/projectSettings.module.scss
+++ b/jsapp/js/components/modalForms/projectSettings.module.scss
@@ -1,5 +1,5 @@
-@use "scss/colors";
-@use "scss/breakpoints";
+@use 'scss/colors';
+@use 'scss/breakpoints';
 
 .uploadInstructions {
   margin-bottom: 20px;
--- a/jsapp/js/components/permissions/sharingForm.scss
+++ b/jsapp/js/components/permissions/sharingForm.scss
@@ -1,7 +1,7 @@
-@use "scss/colors";
-@use "scss/_variables";
-@use "scss/libs/_mdl";
-@use "scss/mixins";
+@use 'scss/colors';
+@use 'scss/_variables';
+@use 'scss/libs/_mdl';
+@use 'scss/mixins';
 
 $s-flexed-row-button-width: 100px;
 $s-flexed-row-select-width: 200px;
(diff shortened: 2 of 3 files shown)


jamesrkiger:
Can you take a look and see how difficult it would be to get this banner on the form sharing page (subview on the settings page)? It would be nice if there was a banner there indicating a user is over their limit to supplement the tooltip. If it's going to be too much of a pain, don't worry about it (it's not in the designs), but I think it would be nice addition if it's easy enough to do.

jamesrkiger:
I'm not sure why this broke, but it works on main so I assume it's something in this branch:
image

jamesrkiger:
> > Can you take a look and see how difficult it would be to get this banner on the form sharing page (subview on the settings page)? It would be nice if there was a banner there indicating a user is over their limit to supplement the tooltip. If it's going to be too much of a pain, don't worry about it (it's not in the designs), but I think it would be nice addition if it's easy enough to do.
> 
> Done! Not the most straightforward thing, but wasn't too hard to do! 🙌🏻

Thank you. Can you adjust it so it only shows up on the sharing page and not in this modal? (The modal doesn't have ownership transfer, so it doesn't need the warning).

image

jamesrkiger (approved):
LGTM

PR #5938, opened 2025-07-18
feat(attachments): Add celery task to auto delete excess attachments DEV-240

rajpatel24:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Adds auto_delete_excess_attachments task that soft-deletes a user’s oldest attachments until their storage usage falls below the configured limit. This completes the attachment cleanup workflow for users exceeding their storage quotas.

### 📖 Description
This PR adds the `auto_delete_excess_attachments(user_id)` Celery task, which:
- Checks if the user is exceeding their storage quota
- If so, iterates over the user's oldest attachments (`date_created` order)
- Soft-deletes attachments using the existing `move_to_trash()` utility
- Stops once the user is back within the allowed quota

- Uses `ServiceUsageCalculator(disable_cache=True)` to get accurate, uncached usage data
- Retries automatically on transient failures using `autoretry_for` with exponential backoff


kobo/apps/trash_bin/tasks/attachment.py
```diff
@@ -79,6 +89,62 @@ def schedule_auto_attachment_cleanup_for_users():
         auto_delete_excess_attachments(counter.user.pk)
```
noliveleger: Like I said in the other PR, you probably want to call `auto_delete_excess_attachments.delay()` . 
I think we should also need to have some lock mechanism to avoid running the same task twice (e.g. with `cache.lock`)

kobo/apps/trash_bin/tasks/attachment.py
```diff
@@ -79,6 +89,62 @@ def schedule_auto_attachment_cleanup_for_users():
...
+    queue='kpi_low_priority_queue',
+    soft_time_limit=settings.CELERY_LONG_RUNNING_TASK_SOFT_TIME_LIMIT,
+    time_limit=settings.CELERY_LONG_RUNNING_TASK_TIME_LIMIT,
+)
 def auto_delete_excess_attachments(user_id: int):
-    # ToDo: Implement the logic to auto-delete excess attachments for the user.
-    pass
+    user = User.objects.get(pk=user_id)
+    usage_balance = ServiceUsageCalculator(
+        user, disable_cache=True
```
noliveleger: Why do we want to disable the cache? Especially if the task is running every 30 minutes.

kobo/apps/trash_bin/tasks/attachment.py
```diff
@@ -79,6 +89,62 @@ def schedule_auto_attachment_cleanup_for_users():
...
+        TrashTaskInProgressError,
+        SoftTimeLimitExceeded,
+    ),
+    retry_backoff=60,
+    retry_backoff_max=600,
+    max_retries=5,
+    retry_jitter=False,
+    queue='kpi_low_priority_queue',
+    soft_time_limit=settings.CELERY_LONG_RUNNING_TASK_SOFT_TIME_LIMIT,
+    time_limit=settings.CELERY_LONG_RUNNING_TASK_TIME_LIMIT,
```
noliveleger: Do you think it's necessary. Moving files to trash should be fast enough to run under 30 minutes (the default)?

rajpatel24: ~~I’ve kept `soft_time_limit=600` and `time_limit=900` to provide a safety net in case the task unexpectedly hangs~~

kobo/apps/trash_bin/tests/storage_cleanup/test_attachment_cleanup.py
```diff
@@ -0,0 +1,217 @@
...
+from django.test import TestCase
+
+if settings.STRIPE_ENABLED:
+    from kobo.apps.stripe.models import ExceededLimitCounter
+
+from kobo.apps.kobo_auth.shortcuts import User
+from kobo.apps.openrosa.apps.logger.models import Attachment
+from kobo.apps.organizations.constants import UsageType
+from kobo.apps.trash_bin.tasks.attachment import auto_delete_excess_attachments, \
+    schedule_auto_attachment_cleanup_for_users
```
noliveleger: Black Formatting 🙏

kobo/apps/trash_bin/tasks/attachment.py
```diff
@@ -80,9 +89,56 @@ def schedule_auto_attachment_cleanup_for_users(**stripe_models):
...
-        auto_delete_excess_attachments(counter.user_id)
+        auto_delete_excess_attachments.delay(counter.user_id)
 
 
+@celery_app.task(queue='kpi_low_priority_queue')
 def auto_delete_excess_attachments(user_id: int):
-    # ToDo: Implement the logic to auto-delete excess attachments for the user.
-    pass
+    cache_key = f'auto_delete_excess_attachments_lock_for_user_{user_id}'
+    with cache.lock(cache_key, timeout=300, blocking_timeout=0) as lock_acquired:
```
noliveleger: Let's use Celery values from settings. We don't want the lock to be released before the Celery task is done.

kobo/apps/trash_bin/tasks/attachment.py
```diff
@@ -80,9 +89,56 @@ def schedule_auto_attachment_cleanup_for_users(**stripe_models):
...
+        return
+
+    exceeded_bytes = balance_info['balance_value'] * -1
+    logging.info(
+        f'User `{user_id}` has exceeded storage limits by {exceeded_bytes} bytes.'
+    )
+
+    attachments_to_trash = []
+    trashed_bytes = 0
+    queryset = Attachment.objects.filter(user=user).order_by('date_created').only(
```
noliveleger: I think Django is smart enough to recognize `user=user` as `user_id=user.pk` but I think it would be more cautious to use it that way. `User` comes from KPI database and `Attachment` is running on Kobocat database.

kobo/apps/trash_bin/tests/storage_cleanup/test_attachment_cleanup.py
```diff
@@ -0,0 +1,217 @@
...
+from unittest.mock import patch
+
+from constance import config
+from constance.test import override_config
+from django.conf import settings
+from django.core.cache import cache
+from django.test import TestCase
+
+if settings.STRIPE_ENABLED:
+    from kobo.apps.stripe.models import ExceededLimitCounter
```
noliveleger: I thought we could use the decoration and avoid the conditional import?

kobo/apps/trash_bin/tests/storage_cleanup/test_attachment_cleanup.py
```diff
@@ -0,0 +1,217 @@
...
+        }
+        with patch(
+            'kobo.apps.subsequences.api_view.ServiceUsageCalculator.get_usage_balances',  # noqa
+            return_value=mock_balances,
+        ):
+            auto_delete_excess_attachments(self.owner.pk)
+            self.attachment.refresh_from_db()
+            self.assertFalse(Attachment.objects.filter(pk=self.attachment.pk).exists())
+
+    def test_auto_delete_stops_when_user_is_within_storage_limit(self):
```
noliveleger: I think this test name is confusing. I think this test is kind cover by @jamesrkiger's work. I would rather see that the user has _n_ attachments and they are above the limit but after deleting _x_ attachments, they go below their limit and the loop inside `auto_delete_excess_attachments` stop. Then, assert that only the new attachments remain.

kobo/apps/trash_bin/tests/storage_cleanup/test_attachment_cleanup.py
```diff
@@ -0,0 +1,217 @@
...
+            user=self.owner,
+            limit_type=UsageType.STORAGE_BYTES,
+            days=config.LIMIT_ATTACHMENT_REMOVAL_GRACE_PERIOD + 1
+        )
+
+        with patch(
+            'kobo.apps.trash_bin.tasks.attachment.auto_delete_excess_attachments.delay'
+        ) as mock_task:
+            schedule_auto_attachment_cleanup_for_users()
+            mock_task.assert_called_once_with(self.owner.pk)
```
noliveleger: We have only 1 user. Would be great to assert the `schedule_auto_attachment_cleanup_for_users ` does not run for the other user.

kobo/apps/trash_bin/tests/storage_cleanup/test_attachment_cleanup.py
```diff
@@ -0,0 +1,217 @@
...
+        ) as mock_usage:
+            mock_usage.return_value = {}
+
+            auto_delete_excess_attachments(self.owner.pk)
+            self.assertTrue(Attachment.objects.filter(pk=self.attachment.pk).exists())
+
+    @pytest.mark.skipif(
+        not settings.STRIPE_ENABLED, reason='Requires stripe functionality'
+    )
+    @override_config(AUTO_DELETE_ATTACHMENTS=True)
```
noliveleger, line 196: What about adding a test for `AUTO_DELETE_ATTACHMENTS=False`  where nothing is called? 
It could be useful.

kobo/apps/trash_bin/tests/storage_cleanup/test_attachment_cleanup.py
```diff
@@ -0,0 +1,242 @@
...
+            return_value=mock_balances,
+        ):
+            auto_delete_excess_attachments(self.owner.pk)
+
+        # Confirm enough attachments were deleted to get under the limit
+        remaining_attachments = Attachment.objects.filter(user=self.owner)
+        remaining_size = sum(att.media_file_size for att in remaining_attachments)
+
+        self.assertLessEqual(remaining_size, limit_bytes)
+        self.assertEqual(remaining_attachments.count(), 1)
```
noliveleger, line 125: We should assert that remaining one is the newest one to be sure the deletion is applied in ASC order.

kobo/apps/trash_bin/tests/storage_cleanup/test_attachment_cleanup.py
```diff
@@ -0,0 +1,242 @@
...
+    @override_config(AUTO_DELETE_ATTACHMENTS=True)
+    def test_schedule_cleanup_task_only_for_users_exceeding_grace_period(
+        self, **stripe_models
+    ):
+        """
+        Test that only users over their limit for more than the grace period are
+        scheduled for attachment cleanup
+        """
+        anotheruser = User.objects.create(username='anotheruser', password='password')
+        ExceededLimitCounter = stripe_models['exceeded_limit_counter_model']
```
noliveleger, line 158: ❤️

kobo/apps/trash_bin/tests/storage_cleanup/test_attachment_cleanup.py
```diff
@@ -0,0 +1,242 @@
...
+    )
+    @override_config(AUTO_DELETE_ATTACHMENTS=True)
+    def test_auto_delete_excess_attachments_skips_if_lock_held(self):
+        """
+        Test that the task does not run if a cache lock is already held
+        """
+        lock_key = f'auto_delete_excess_attachments_lock_for_user_{self.owner.pk}'
+
+        # Manually acquire the lock to simulate another task running
+        lock = cache.lock(lock_key, timeout=30)
```
noliveleger, line 204: You could use a try/finally to always release the lock at the end of the test to avoid waiting for 30 seconds

PR #5939, opened 2025-07-18
docs(api): adding API endpoints for me in KPI documentation DEV-758

ZoeGodin:
### 📣 Summary
Added API documentation for `/api/v2/docs/me` endpoint.

### 📖 Description
This PR introduces the endpoint documentation that will be used for the auto-generating API documentation of the `me` endpoint. With these changes, users will be able to understand, test and use the `me` endpoint more easily.

### 👀 Preview steps
To see the changes, visit `http://kf.kobo.local/api/v2/docs/`. The page of the Kobo API will now be visible and generated by swagger. To see the changes added to `me` endpoint, scroll down to the category of the same name or make a research in the search bar. Documentation for the endpoint will be provided with a HTTP code, a body and/or a response when necessary.


kpi/schema_extensions/v2/me/extensions.py
```diff
@@ -0,0 +1,79 @@
...
+                'twitter': build_basic_type(OpenApiTypes.STR),
+                'linkedin': build_basic_type(OpenApiTypes.STR),
+                'instagram': build_basic_type(OpenApiTypes.STR),
+                'organization': build_basic_type(OpenApiTypes.STR),
+                'last_ui_language': build_basic_type(OpenApiTypes.STR),
+                'organization_type': build_basic_type(OpenApiTypes.STR),
+                'organization_website': build_basic_type(OpenApiTypes.STR),
+                'project_views_settings': build_object_type(
+                    properties={
+                        'kobo_my_projects': build_object_type(
```
rgraber: I think this was just the name of someone's specific project view, we can just say this is a dict

rgraber (changes requested):
Why are some fields an example and some fields a type? Also the username looks funny 
Screenshot from 2025-07-22 11-53-51

kpi/schema_extensions/v2/me/extensions.py
ZoeGodin: Done!

rgraber:
> > Why are some fields an example and some fields a type? Also the username looks funny Screenshot from 2025-07-22 11-53-51
> 
> This is because I use the serializer itself instead of an inline_class_serializer so some fields will have examples and others no since they were extended. I can create an inline_class_serializer to uniformize everything if it would look better.

I think we should make it uniform. Otherwise it's confusing

PR #5941, opened 2025-07-21
docs(api): adding API endpoints for invites in KPI documentation DEV-541

ZoeGodin:
### 📣 Summary
Added API documentation for `/api/v2/docs/invites` endpoint.

### 📖 Description
This PR introduces the endpoint documentation that will be used for the auto-generating API documentation of the `invites` endpoint. With these changes, users will be able to understand, test and use the `invites` endpoint more easily.

### 👀 Preview steps
To see the changes, visit `http://kf.kobo.local/api/v2/docs/`. The page of the Kobo API will now be visible and generated by swagger. To see the changes added to `invites` endpoint, scroll down to the category of the same name or make a research in the search bar. Documentation for the endpoint will be provided with a HTTP code, a body and/or a response when necessary.

### DIFF
--- a/kobo/apps/organizations/views.py
+++ b/kobo/apps/organizations/views.py
@@ -4,7 +4,6 @@ from django.db.models.expressions import Exists
 from django.utils.http import http_date
 from drf_spectacular.openapi import AutoSchema
 from drf_spectacular.utils import OpenApiParameter, extend_schema, extend_schema_view
-from drf_spectacular.utils import OpenApiParameter, extend_schema, extend_schema_view
 from rest_framework import status, viewsets
 from rest_framework.decorators import action
 from rest_framework.renderers import JSONRenderer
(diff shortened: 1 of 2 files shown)


kobo/apps/organizations/views.py
```diff
@@ -545,163 +608,112 @@ def perform_destroy(self, instance):
 
 
 @extend_schema(
-    tags=['invites'],
+    tags=['Project Ownership Transfer'],
```
rajpatel24: Shouldn't it be `Organization Invites` or perhaps `organization-invites`, since the endpoints are for handling organization invitations?

rajpatel24 (changes requested):

kobo/apps/organizations/views.py
ZoeGodin: Like Olivier commented, the tag should be "Project Ownership Transfer". See comment left previously.
https://github.com/kobotoolbox/kpi/pull/5941#issuecomment-3097086792

noliveleger: @ZoeGodin this one is a different endpoint. @rajpatel24 is right. it should be "Organization Invites" . It's used to Invite people to join an org. The other one from #5941, is to accept (or decline) a project ownership transfer.

ZoeGodin: I changed the tag so it is more representative of the organization it is part from.

PR #5945, opened 2025-07-21
docs(api): adding API endpoints for members in KPI documentation DEV-542

ZoeGodin:
### 📣 Summary
Added API documentation for `/api/v2/docs/members` endpoint.

### 📖 Description
This PR introduces the endpoint documentation that will be used for the auto-generating API documentation of the members endpoint. With these changes, users will be able to understand, test and use the `members` endpoint more easily.

### 👀 Preview steps
To see the changes, visit `http://kf.kobo.local/api/v2/docs/`. The page of the Kobo API will now be visible and generated by swagger. To see the changes added to `members` endpoint, scroll down to the category of the same name or make a research in the search bar. Documentation for the endpoint will be provided with a HTTP code, a body and/or a response when necessary.


kobo/apps/organizations/views.py
```diff
@@ -265,13 +272,81 @@ def asset_usage(self, request, pk=None, *args, **kwargs):
...
+    ),
+    list=extend_schema(
+        description='list',
+        responses=open_api_200_ok_response(
+            MemberListResponse,
+            require_auth=False,
+            raise_access_forbidden=False,
+            validate_payload=False,
+        ),
+    ),
```
rajpatel24: Why is `user__username` being asked as a required field? I think it's worth double-checking this endpoint, as I don't believe we need `user__username` as a required field for the `/api/v2/organizations/{organization_id}/members/` list API.

kobo/apps/organizations/views.py
```diff
@@ -265,13 +272,81 @@ def asset_usage(self, request, pk=None, *args, **kwargs):
 
 
 @extend_schema(
-    tags=['members'],
+    tags=['Members'],
+    parameters=[
+        OpenApiParameter(
+            name='user__username',
```
rajpatel24: How about just keeping `username` instead of `user__username` for consistency? What do you think?

kpi/docs/api/v2/members/update.md
```diff
@@ -0,0 +1,7 @@
+## Update Member Role
+
+Updates the role of a member within the organization to `admin` or
+`member`.
+
+- **admin**: Grants the member admin privileges within the organization
+- **member**: Revokes admin privileges, setting the member as a regular user
```
rajpatel24: Have you tried running the `destroy` and `update` APIs? They always give me a 403 error. Is it the same on your end?

ZoeGodin: This is a bug probably introduced with a change from another branch but there will be a fix on another pr.

kobo/apps/organizations/views.py
ZoeGodin: It was indeed an error and was replaced by organization_id

kobo/apps/organizations/views.py
ZoeGodin: Unfortunately I cannot since I cannot because the parameter needs to be the same that will be applied for the path. The reason there are OpenApiParameter is because the schema wasn't automatically recognizing the type so it needs to be specified in the schema- for this reason the name must be the exact same.

image

noliveleger: > How about just keeping `username` instead of `user__username` for consistency? What do you think?

@rajpatel24 , We would need to change the lookup_field in the viewset. I agree it should be changed. 

https://github.com/kobotoolbox/kpi/blob/cea4d0524ca3276e91610c41c290e5e0f8de7174/kobo/apps/organizations/views.py#L441 (Don't use blame on it 🙈 )

kobo/apps/organizations/views.py
```diff
@@ -265,13 +272,81 @@ def asset_usage(self, request, pk=None, *args, **kwargs):
 
 
 @extend_schema(
-    tags=['members'],
+    tags=['Members'],
+    parameters=[
+        OpenApiParameter(
+            name='organization_id',
+            type=int,
```
rajpatel24: Shouldn't it be a string instead of an integer?

noliveleger: It is a string ;-)

ZoeGodin: Changed for a string :smile:

rajpatel24 (approved):
LGTM

PR #5973, opened 2025-07-23
ci(releases): use Github App for cascading merges DEV-561

Akuukis:
### 💭 Notes

See Github docs [here](https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-when-your-workflow-runs/triggering-a-workflow#triggering-a-workflow-from-a-workflow).


.github/workflows/release-2-stabilize.yml
```diff
@@ -98,6 +98,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
+          token: ${{ secrets.CI_RELEASE_PAT }}
```
magicznyleszek: Do we want to comment somehow why we are using custom token?

PR #5977, opened 2025-07-23
Inject Django CSRF token in SwaggerUI template

ZoeGodin:
### 📣 Summary
Inject Django CSRF token in SwaggerUI template

### 📖 Description
Inject Django CSRF token in SwaggerUI template so user may still have the ability to use the swagger POST, PATCH and DELETE try out function.


RuthShryock (approved):
Works great and lgtm :)



(One side note that I found while testing) 
There is missing documentation for 201 returned by POST requests to the `/api/v2/assets` endpoint. I'm not sure if you want to add that here or in another pr?
Image

RuthShryock (approved):
LGTM 👌🏻

PR #5978, opened 2025-07-23
docs(api): removing openRosa endpoints from schema DEV-782

ZoeGodin:
### 📣 Summary
Removed the OpenRosa endpoints from appearing the schema.

### 📖 Description
Removed the OpenRosa endpoints from appearing in the `/api/v2/docs/` schema since they will be moved to their own docs later on.

### 💭 Notes

The endpoints have been 'removed' by simply excluding them from the current schema. Unfortunately I cannot remove them with the pre-processing hooks as we still want them in the schema for later on when we will be documenting their own docs (probably using a post-processing hook to separate the schema).


RuthShryock (approved):
LGTM

PR #5993, opened 2025-07-25
feat(permissions): remove guardian dependency DEV-229 DEV-230

Guitlle:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
This PR removes all code related to the guardian permissions library to merge the permissions system in KoboCat/OpenRosa and KPI and is a step towards refactoring the KPI permission system. 

### 👀 Preview steps
1. You need to have 4 users, A, B, C and D 
2. Create and deploy an asset from user A, share it with user B, and share it partially to user C (to see and edit it's own submissions only)
3. Ensure users A, B and C have their expected permissions and that they perform actions they should not be able to do (user C, should not be able to view and edit submissions from other users, and so on)
4. Ensure anonymous behavior is the correct, enable and disable anonymous submissions and test accordingly
5. Check that the permission changes create the correct project history logs

This should also pass through QA in a staging server to ensure it won't disrupt production, as this is a very drastic change in the codebase


kobo/conftest.py
```diff
@@ -0,0 +1,15 @@
...
+
+@fixture(autouse=True)
+def anonymous_user():
+
+    if connection.connection is None:
+        return
+
+    anon = get_anonymous_user()
+
+    return anon
```
noliveleger: Let's a comment somewhere here to explain why we are doing this. 
Was it too difficult to do it with a fixture?

kpi/deployment_backends/openrosa_backend.py
```diff
@@ -105,24 +105,10 @@ def attachment_storage_bytes(self):
...
-            and list(users_with_perms)[0].id == self.asset.owner_id
-        ):
-            return
-
-        with kc_transaction_atomic():
-            for user, perms in users_with_perms.items():
-                if user.id == self.asset.owner_id:
-                    continue
-                assign_applicable_kc_permissions(self.asset, user, perms)
+        pass
```
noliveleger: why don't we remove this entirely?

kpi/mixins/object_permission.py
```diff
@@ -442,22 +413,21 @@ def get_all_implied_perms(cls, for_instance=None):
 
     @transaction.atomic
     @kc_transaction_atomic
```
noliveleger, line 414: We probably don't need `@kc_transaction_atomic` anymore

kpi/mixins/object_permission.py
```diff
@@ -532,9 +496,12 @@ def assign_perm(self, user_obj, perm, deny=False, defer_recalc=False,
             codename=codename,
             deny=deny,
         )
-        # Assign any applicable KC permissions
-        if not deny and not skip_kc:
-            assign_applicable_kc_permissions(self, user_obj, codename)
+
+        # Handle KoboCat xform flags for the anonymous user
+        if not deny and is_anonymous:
+            if xform_id := _get_xform_id_for_asset(self):
```
noliveleger: Let's move this condition in `set_kc_anonymous_permissions_xform_flags(self, codename, xform_id)` and do nothing is asset has no deployment. 
Otherwise, rename `_get_xform_id_for_asset` to `get_xform_id_for_asset` since it is called outside the `utils.py` file.

kpi/mixins/object_permission.py
```diff
@@ -680,24 +647,22 @@ def has_perms(self, user_obj: User, perms: list, all_: bool = True) -> bool:
 
     @transaction.atomic
     @kc_transaction_atomic
```
noliveleger, line 647: We probably don't need `@kc_transaction_atomic` anymore here too.

kpi/mixins/object_permission.py
```diff
@@ -755,6 +716,14 @@ def remove_perm(self, user_obj, perm, defer_recalc=False, skip_kc=False):
         # Recalculate all descendants
         self.recalculate_descendants_perms()
 
+        is_anonymous = is_user_anonymous(user_obj)
+        # Handle KoboCat xform flags for the anonymous user
+        if is_anonymous:
+            if xform_id := _get_xform_id_for_asset(self):
+                set_kc_anonymous_permissions_xform_flags(
+                    self, codename, xform_id, remove=True
```
noliveleger: Same comment about ` _get_xform_id_for_asset`

kpi/tests/conftest.py
noliveleger, line 1: We can probably remove it entirely?

noliveleger (changes requested):

kobo/conftest.py
Guitlle: It was fine, but It broke some tests that don't have a DB. That's why this if was needed, to avoid interacting with the DB when it's not available

kpi/deployment_backends/openrosa_backend.py
Guitlle: Sure, I think this abstract method is obsolete

kpi/mixins/object_permission.py
Guitlle: Ok, that's better

Guitlle: Ok, that's better

kpi/tests/conftest.py
Guitlle: Yes

kpi/mixins/object_permission.py
```diff
@@ -441,23 +409,21 @@ def get_all_implied_perms(cls, for_instance=None):
         }
 
     @transaction.atomic
-    @kc_transaction_atomic
```
noliveleger, line 444: @Guitlle Sorry I think I misled you on this one. 
I forgot about `set_kc_anonymous_permissions_xform_flags` in this method. In that case, we need to keep `@kc_transaction_atomic`

kpi/mixins/object_permission.py
```diff
@@ -679,25 +641,22 @@ def has_perms(self, user_obj: User, perms: list, all_: bool = True) -> bool:
         return fn(perm in perms for perm in self.get_perms(user_obj))
 
     @transaction.atomic
-    @kc_transaction_atomic
```
noliveleger, line 682: Same comment as above 

> I forgot about `set_kc_anonymous_permissions_xform_flags` in this method. In that case, we need to keep `@kc_transaction_atomic`

noliveleger (approved):
Please revert the `@kc_transaction_atomic` change (my bad). Otherwise, LGTM. 
You can merge when it's done.

PR #6034, opened 2025-07-30
fix(projectOwnership): restart pending transfers DEV-799

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Ensure ownership transfers do not get stuck in a "pending" state. 

### 💭 Notes
It happens that a k8s pod sometimes restarts in the middle of processing a TransferStatus. When this happens, other TransferStatuses under the same transfer may get stuck in a pending state, which means they will not get picked up by the task_restarter, which only looks for tasks that were in progress. This PR updates the task restarter to look for TransferStatuses that got stuck in pending as well. It also looks for entire Transfers that may have gotten stuck in pending, though this is rarer.  

### 👀 Preview steps

You may want to update the crontab schedule for project-ownership-task-restarter to run a little more frequently for easier testing. It may also be easier to test if you set PROJECT_OWNERSHIP_RESUME_THRESHOLD in Constance to something very small (like 2 minutes) so the task restarter will not think it is too soon.

1. ℹ️ Have an MMO with at least one asset and two users
2. Update kobo.apps.project_ownership.tasks.async_task with `time.sleep(200)` at the beginning of the task. This will make it easier to simulate a pod restarting mid-task
3. In Django admin, transfer the MMO from the owner to a different user by updating the OrganizationOwner
4. Quickly, kill the kpi_worker container
5. In the django shell, notice that several TransferStatus objects associated with the transfer of ownership still have `status='pending'`
6. Remove the `time.sleep(200)` call and restart the worker 
7. Wait for the project-ownership-task-restarter job to run
8. 🔴 [on main] notice that all TransferStatuses that were pending are still pending
9. 🟢 [on PR] notice that some of the TransferStatuses have been moved to in progress or success. Eventually they should all be completed, depending on how many assets needed to be transferred and how many submissions/attachments/media files there were.

### DIFF
--- a/kobo/apps/project_ownership/tasks.py
+++ b/kobo/apps/project_ownership/tasks.py
@@ -304,8 +304,10 @@ def task_restarter():
             tasks_by_transfer[transfer_id].append(transfer_status.status_type)
 
     full_transfer_pks = list(full_transfer_pks)
-    transfers_to_restart = Transfer.objects.filter(
-        pk__in=full_transfer_pks[: settings.MAX_RESTARTED_TRANSFERS]
+    transfers_to_restart = (
+        Transfer.objects.defer('asset__content')
+        .filter(pk__in=full_transfer_pks[:settings.MAX_RESTARTED_TRANSFERS])
+        .select_related('invite', 'invite__sender', 'invite__recipient', 'asset')
     )
 
     # transform {'transfer_pk': ['attachments', 'media_files']...}
@@ -314,9 +316,9 @@ def task_restarter():
         [(pk, status_type) for status_type in tasks_by_transfer[pk]]
         for pk in tasks_by_transfer
     ]
-    all_tasks_to_restart = list(itertools.chain(*all_tasks_to_restart))[
-        : settings.MAX_RESTARTED_TASKS
-    ]
+    all_tasks_to_restart = list(
+        itertools.chain(*all_tasks_to_restart)
+    )[:settings.MAX_RESTARTED_TASKS]
 
     for transfer_id, status_type in all_tasks_to_restart:
         async_task.delay(transfer_id, status_type)
--- a/kobo/apps/project_ownership/tests/test_tasks.py
+++ b/kobo/apps/project_ownership/tests/test_tasks.py
@@ -151,3 +151,13 @@ class ProjectOwnershipTasksTestCase(TestCase):
                 task_restarter()
         assert len(patched_task.call_args_list) == 2
         assert len(patched_transfer.call_args_list) == 2
+
+    def test_query_count(self):
+        assets = [baker.make(Asset, owner=self.someuser, uid=f'a{i}') for i in range(6)]
+        with freeze_time(timezone.now() - timedelta(minutes=8)):
+            # create multiple pending transfers
+            for i, asset in enumerate(assets):
+                self.create_standard_transfer(assets[i])
+        with self.assertNumQueries(1348):
+            with patch('kobo.apps.project_ownership.models.transfer.async_task.delay'):
+                task_restarter()


kobo/apps/project_ownership/tasks.py
```diff
@@ -264,16 +264,42 @@ def task_restarter():
...
     )
+    restarted_transfers = []
+    for transfer_status in TransferStatus.objects.filter(
+        date_modified__lte=resume_threshold,
+        status=TransferStatusChoices.PENDING,
+        status_type=TransferStatusTypeChoices.GLOBAL,
+    ):
+        # restart any transfers that were left in pending but not picked up
+        transfer_status.transfer.transfer_project()
+        restarted_transfers.append(transfer_status.transfer)
```
noliveleger: For any reason, we have _n_ (where _n_ is a high number) projects to restart in this loop, it could take a while. The other loop could never make it. 

Any possibility to avoid to restart transfers sequentially?

rgraber: Some options:
1. Put a hard limit on this loop the same way we do with MAX_RESTARTED_TASKS
2. Spawn new tasks that call transfer_project rather than doing it directly in the loop
3. Have a separate task altogether for calling transfer project and be very careful we don't accidentally pick up the same tasks twice
4. Do this loop second 

1 and 4 are the easiest, followed by 2 and then 3.

kobo/apps/project_ownership/tasks.py
```diff
@@ -264,16 +264,42 @@ def task_restarter():
     stuck_threshold = timezone.now() - timedelta(
         minutes=config.PROJECT_OWNERSHIP_STUCK_THRESHOLD
     )
+    restarted_transfers = []
+    for transfer_status in TransferStatus.objects.filter(
+        date_modified__lte=resume_threshold,
+        status=TransferStatusChoices.PENDING,
+        status_type=TransferStatusTypeChoices.GLOBAL,
+    )[: settings.MAX_RESTARTED_TRANSFERS]:
```
noliveleger: lint: remove extra space in `[:  settings`

kobo/apps/project_ownership/tasks.py
```diff
@@ -264,16 +264,42 @@ def task_restarter():
...
+        # restart any transfers that were left in pending but not picked up
+        transfer_status.transfer.transfer_project()
+        restarted_transfers.append(transfer_status.transfer)
```
noliveleger: Instead of holding in memory the whole object, I would use only the PK of the transfer and I would change the query below to 

```python
...
# filter out any we might already restart via the transfer
Q(transfer_id__in=restarted_transfer_ids)
```

kobo/apps/project_ownership/tasks.py
```diff
@@ -264,16 +264,42 @@ def task_restarter():
     stuck_threshold = timezone.now() - timedelta(
         minutes=config.PROJECT_OWNERSHIP_STUCK_THRESHOLD
     )
+    restarted_transfers = []
+    for transfer_status in TransferStatus.objects.filter(
```
noliveleger: I thought we said we would start the "pending" after the stuck tasks? 
Is still the first thing to happen because of `MAX_RESTARTED_TRANSFERS` limit?

kobo/apps/project_ownership/tests/test_tasks.py
```diff
@@ -0,0 +1,158 @@
...
+    @unpack
+    def test_task_restarter_ignores_tasks_too_old_or_too_new(
+        self, in_progress, too_old, too_recent
+    ):
+        if too_recent:
+            time_delta = 2
+        elif too_old:
+            time_delta = 15
+        else:
+            time_delta = 0
```
noliveleger: Does it ever enter this condition? 

It seems a task cannot be too old and too recent at the same time, why don't we simplify this with 

```python 
def test_task_restarter_ignores_tasks_too_old_or_too_new(
        self, in_progress, time_delta
    ):
```
with a comment for each `@data`  occurence?

kobo/apps/project_ownership/tasks.py
rgraber: I didn't realize we wanted to both limit the transfers and do them second. I can move it around.

kobo/apps/project_ownership/tasks.py
```diff
@@ -258,23 +260,65 @@ def task_restarter():
...
+
+    # transform {'transfer_pk': ['attachments', 'media_files']...}
+    # to [(pk, 'attachments'), (pk, 'media_files')]
+    all_tasks_to_restart = [
+        [(pk, status_type) for status_type in tasks_by_transfer[pk]]
+        for pk in tasks_by_transfer
+    ]
+    all_tasks_to_restart = list(itertools.chain(*all_tasks_to_restart))[
+        : settings.MAX_RESTARTED_TASKS
+    ]
```
noliveleger: 🙈  

Probably better this way  

```python
all_tasks_to_restart = list(
    itertools.chain(*all_tasks_to_restart)
)[:settings.MAX_RESTARTED_TASKS]
```

kobo/apps/project_ownership/tasks.py
```diff
@@ -258,23 +260,65 @@ def task_restarter():
...
+        else:
+            # if we know we'll need to transfer the whole project
+            # don't bother adding individual task restarts
+            if transfer_id in full_transfer_pks:
+                continue
+            tasks_by_transfer[transfer_id].append(transfer_status.status_type)
+
+    full_transfer_pks = list(full_transfer_pks)
+    transfers_to_restart = Transfer.objects.filter(
+        pk__in=full_transfer_pks[: settings.MAX_RESTARTED_TRANSFERS]
```
noliveleger: I don't know why Darker is producing this space but it should not (according to https://peps.python.org/pep-0008/#whitespace-in-expressions-and-statements)

No space after the colon

kobo/apps/project_ownership/tasks.py
```diff
@@ -258,23 +260,65 @@ def task_restarter():
...
+        [(pk, status_type) for status_type in tasks_by_transfer[pk]]
+        for pk in tasks_by_transfer
+    ]
+    all_tasks_to_restart = list(itertools.chain(*all_tasks_to_restart))[
+        : settings.MAX_RESTARTED_TASKS
+    ]
+
+    for transfer_id, status_type in all_tasks_to_restart:
+        async_task.delay(transfer_id, status_type)
+    for transfer in transfers_to_restart:
```
noliveleger, line 325: What about loading related tables to avoid making more queries to DB and adding more latency. 

```
transfers_to_restart.select_related('invite', 'invite__sender', 'invite__recipient', 'asset')
```

Ideally, I would also defer 'asset.content' if it's not needed. 

```
transfers_to_restart.defer('asset__content').select_related('invite', 'invite__sender', 'invite__recipient', 'asset')
```

Probably it needs a `.distinct()` though because of the JOINs.

noliveleger (changes requested):

kobo/apps/project_ownership/tasks.py
rgraber: I'm a little confused. Why would the JOINs necessitate a distinct?

rgraber: It seems to work without the distinct FWIW

noliveleger: > I'm a little confused. Why would the JOINs necessitate a distinct?

Nevermind, there are all One-to-One relationships 🤦 .

kobo/apps/project_ownership/tests/test_tasks.py
```diff
@@ -151,3 +151,13 @@ def test_task_restarter_limits_number_of_restarts(self):
...
         assert len(patched_task.call_args_list) == 2
         assert len(patched_transfer.call_args_list) == 2
+
+    def test_query_count(self):
+        assets = [baker.make(Asset, owner=self.someuser, uid=f'a{i}') for i in range(6)]
+        with freeze_time(timezone.now() - timedelta(minutes=8)):
+            # create multiple pending transfers
+            for i, asset in enumerate(assets):
+                self.create_standard_transfer(assets[i])
+        with self.assertNumQueries(1348):
```
noliveleger, line 161: Only 1348 :-P?

noliveleger (approved):
LGTM

noliveleger:
@rgraber I merged it myself to prepare the new release image.

PR #6036, opened 2025-07-30
refactor(formbuilder): remove implicit returns for `model.row.coffee` DEV-818

duvld:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Fixes DS102 from decaffeinate process

### 👀 Preview steps

1. Copy all of `model.row.coffee` to https://decaffeinate-project.org/repl/
2. 🔴 [on main] notice that there is a `DS102` comment generated
3. 🟢[on PR] notice that there is no `DS102` comment generated
4. Run `npm run test`
5. 🟢[on PR] notice all tests pass
6. ℹ️ have an account and a project
7. 🟢[on PR] check that all formbuilder functionality works

### DIFF
--- a/jsapp/xlform/src/model.row.coffee
+++ b/jsapp/xlform/src/model.row.coffee
@@ -28,32 +28,36 @@ module.exports = do ->
     ensureKuid: ->
       if '$kuid' not of @attributes
         @set '$kuid', txtid()
+      return
 
     initialize: ->
       @ensureKuid()
       @convertAttributesToRowDetails()
+      return
 
     isError: -> false
     convertAttributesToRowDetails: ->
       for key, val of @attributes
         unless val instanceof $rowDetail.RowDetail
           @set key, new $rowDetail.RowDetail({key: key, value: val}, {_parent: @}), {silent: true}
+      return
     attributesArray: ()->
       arr = ([k, v] for k, v of @attributes)
       arr.sort (a,b)-> if a[1]._order < b[1]._order then -1 else 1
-      arr
+      return arr
 
     isGroup: ->
-      @constructor.kls is "Group"
+      return @constructor.kls is "Group"
 
     isInGroup: ->
-      @_parent?._parent?.constructor.kls is "Group"
+      return @_parent?._parent?.constructor.kls is "Group"
 
     detach: (opts)->
       if @_parent
         @_parent.remove @, opts
         @_parent = null
       ` `
+      return
 
     selectableRows: () ->
       questions = []
@@ -68,11 +72,13 @@ module.exports = do ->
         limit = limit || question is @
         if !limit && question.getValue('type') not in non_selectable
           questions.push question
+          return
       , includeGroups:true
-      questions
+      return questions
 
     export_relevant_values: (survey_arr, additionalSheets)->
       survey_arr.push @toJSON2()
+      return
 
     toJSON2: ->
       outObj = {}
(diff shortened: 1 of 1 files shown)


jsapp/xlform/src/model.row.coffee
```diff
@@ -28,32 +28,36 @@ module.exports = do ->
...
     isInGroup: ->
-      @_parent?._parent?.constructor.kls is "Group"
+      return @_parent?._parent?.constructor.kls is "Group"
 
     detach: (opts)->
       if @_parent
         @_parent.remove @, opts
         @_parent = null
       ` `
+      return
```
magicznyleszek, line 59: What are those two backticks with a space? Shouldn't we return them? But if they don't do anything, maybe let's remove them.

jsapp/xlform/src/model.row.coffee
```diff
@@ -137,10 +147,12 @@ module.exports = do ->
...
           export_relevant_values: (surv, addl)->
             surv.push _.extend({}, _end_json)
+            return
           toJSON: ->
-            _.extend({}, _end_json)
+            return _.extend({}, _end_json)
 
-        cb(obj)  if ctxt.includeGroupEnds
+        if ctxt.includeGroupEnds
+          return cb(obj)
```
magicznyleszek, line 154: That coffee code looked very confusing 👍

jsapp/xlform/src/model.row.coffee
```diff
@@ -235,15 +250,17 @@ module.exports = do ->
...
   class ScoreChoiceList extends Array
 
   class ScoreRow extends SimpleRow
     initialize: ->
       @set('type', 'score__row')
       @simpleEnsureKuid()
+      return
     export_relevant_values: (surv, sheets)->
-      surv.push(@attributes)
+      return surv.push(@attributes)
```
magicznyleszek: In all the other `export_relevant_values` we don't return anything. And AFAIK we only call this function in one place and it is not being assigned to anything

jsapp/xlform/src/model.row.coffee
```diff
@@ -273,6 +290,7 @@ module.exports = do ->
         ctx.warnings.push "Score choices list not set"
         @_scoreChoices = @getSurvey().choices.add(name: txtid())
       ``
+      return
```
magicznyleszek, line 291: Same question here - shouldn't we remove the backtics? Maybe two backticks were being used as implicit return?

jsapp/xlform/src/model.row.coffee
```diff
@@ -283,6 +301,7 @@ module.exports = do ->
       output['kobo--score-choices'] = @getList().get('name')
       survey_arr.push(output)
       ``
+      return
```
magicznyleszek, line 301: Same question here - shouldn't we remove the backtics? Maybe two backticks were being used as implicit return?

magicznyleszek (changes requested):

jsapp/xlform/src/model.row.coffee
duvld: Ah, good catch, done to everything

jsapp/xlform/src/model.row.coffee
duvld: This was part of the ones where I figured returning something *can't hurt*, as like you said the other ones don't return anything. But thinking about it again it may be more confusing if we ever touch this code in the future and some of them return something and others don't, so I removed it :)

magicznyleszek:
@duvld approved, the CI fails for unrelated reason - please merge :)

PR #6050, opened 2025-08-05
fix(projectOwnership): remove failing test DEV-799

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 💭 Notes
Removes a test that was failing. It's only a query count test, not a functional one, and the failure was an undercount so it should be safe to remove.

### 👀 Preview steps
None, only a test.


noliveleger (approved):
LGTM, but I did not run the tests locally. We should wait for the all the tests to pass before merging.

PR #6053, opened 2025-08-05
chore(permissions): clean up obsolete permissions code DEV-824

Guitlle:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Cleaned up all appearances of *_xform permissions in the codebase. Changed tests and constants in order to avoid referencing any Openrosa permission

### 👀 Preview steps
This is covered in unit tests, though QA should test functionalities about permissions in order to ensure nothing is broken.


kobo/apps/openrosa/apps/api/permissions.py
```diff
@@ -31,13 +28,26 @@ class ViewDjangoObjectPermissions(DjangoObjectPermissions):
     }
 
 
+class ViewAssetPermissions(DjangoObjectPermissions):
```
noliveleger: I would rather use `AssetObjectPermission`  or something similar. `ViewAssetPermissions` is confusing because it sounds like `view_asset` permission.

kobo/apps/openrosa/apps/logger/migrations/0015_add_delete_data_permission.py
```diff
@@ -127,15 +81,15 @@ def forwards_func(apps, schema_editor):
     sys.stderr.flush()
     create_new_perms(apps)
     grant_model_level_perms(apps)
-    # Deprecated openrosa permissions:
+    # Deprecated openrosa guardian based permissions:
     #   grant_object_level_perms(apps)
```
noliveleger: I think this could be removed

kobo/apps/openrosa/apps/logger/migrations/0015_add_delete_data_permission.py
```diff
@@ -127,15 +81,15 @@ def forwards_func(apps, schema_editor):
...
 
 
 def reverse_func(apps, schema_editor):
     # In testing, removal took only a small fraction of the time that it took
     # to create the assignments
     remove_model_level_perms(apps)
-    # Deprecated openrosa permissions:
+    # Deprecated openrosa guardian based permissions:
     #   remove_object_level_perms(apps)
```
noliveleger: I think this could be removed

kpi/utils/object_permission.py
```diff
@@ -138,6 +138,7 @@ def get_objects_for_user(
             new_app_label, codename = perm.split('.', 1)
             if app_label != new_app_label:
                 # Transitioning away from generic-relation ObjectPermission
+                breakpoint()
```
noliveleger: 🙈

noliveleger (commented):
Just had a look before final release.

kpi/utils/object_permission.py
Guitlle: 👀

kobo/apps/openrosa/apps/api/permissions.py
```diff
@@ -299,40 +274,37 @@ def __init__(self, *args, **kwargs):
...
     def has_permission(self, request, view):
-
         if request.user and request.user.is_superuser:
             return True
 
         return super().has_permission(request, view)
 
     def has_object_permission(self, request, view, obj):
-        view.model = XForm
+        view.model = obj.instance.xform.asset._meta.model
```
noliveleger: Is it because we cannot import Asset directly? If yes, let's add a comment to explain why.

EDIT: 

Actually, I think we should avoid this kind of call because it creates 3 queries to DB, only to get the model class. 

Maybe, it's an import problem, it would be better to do something like
```
from django.apps import apps
...
            Asset = apps.get_model('kpi', 'Asset')  # noqa
            view.model = Asset
```

kobo/apps/openrosa/apps/api/permissions.py
```diff
@@ -299,40 +274,37 @@ def __init__(self, *args, **kwargs):
...
 
         if request.user and request.user.is_superuser:
             return True
 
-        return super().has_object_permission(
-            request, view, obj.instance.xform)
+        return super().has_object_permission(request, view, obj.instance.xform.asset)
```
noliveleger: Tiny optimization, it could be `obj.xform.asset` since we now denormalize the data and we can access the XForm directly without going through the Instance table.

noliveleger (changes requested):
Just tiny changes.

kobo/apps/openrosa/apps/api/permissions.py
Guitlle: Ok

kobo/apps/openrosa/apps/api/permissions.py
Guitlle: thanks, ok

kobo/apps/openrosa/apps/api/permissions.py
Guitlle: I think we don't really  need this , actually

Issue #5975, opened 2025-07-23
bulk-submission-form doesn't accept non-image attachments


nicopace:
When using the bulk submission form, it's possible to upload ZIP files containing XML submissions and their corresponding attachments. However, the current implementation only accepts JPEG files as attachments, excluding all others.

The relevant code is:

[https://github.com/kobotoolbox/kpi/blob/d6464023bdaf1365e07d5e273877351c6ba7f2eb/kobo/apps/openrosa/apps/logger/import\_tools.py#L86-L89](https://github.com/kobotoolbox/kpi/blob/d6464023bdaf1365e07d5e273877351c6ba7f2eb/kobo/apps/openrosa/apps/logger/import_tools.py#L86-L89)

This behavior is incorrect. The expected logic should be:

* Exclude the XML file itself,
* Include **any** attachment referenced in the XML submission, regardless of its file type or extension.

This limitation prevents the use of valid non-JPEG attachments (e.g., PDFs, audio files), which are supported in other parts of the platform.

Thanks!

tiritea:
FYI I've confirmed this behavior also, specifically audio file attachments cannot but uploaded via the bulk-upload interface, but images can. Presumably, looking at the code above, its likely nothing but images can be uploaded (!). This is especially problematic for audio files, because it makes it impossible to get `background-audio` files re-submitted, because these are **not** exposed via the Enketo edit interface (whereas it is still possible to patch in other attachments after-the-fact via editing the submission in Kobo).

There is also no indication in the bulk-upload interface that certain files have not in fact been uploaded. That could cause a user to inadvertently delete and loose their original submission data, thinking it had been safely uploaded into Kobo when it fact some attachments were not.

tiritea:
FYI some further background here: https://community.kobotoolbox.org/t/xml-submission-via-bulk-upload-works-but-audio-file-is-not-attached-500-error/70891

noliveleger:
Hello @nicopace , we accept contributions. But you need to comply to our Python coding rules though ;-) 
Follow the template checkboxes and then you can submit your PR. Please make sur to add unit tests.

nicopace:
Thanks for the encouragement @noliveleger .

I did my best to do a comprenhensive reporting so people more involved in coding could carry on with the task.
I regretably lack the time to dive into the implementation details of KoboToolbox, as most of my time I spend it in user facing tech support for Indigenous communities, so I can't dive into this issue, as simple as it may seem.

[...]

PR #6055, opened 2025-08-06
fix(import_tools): accept non-image attachments

p2edwards:
Draft for #5975

It's a start, but it's still missing something.

On this branch, attachments submitted via zip:

- ✅ do appear in the Enketo preview
- ❌ but won't load in the data table
- ❌ nor `api/v2/assets/{uid}/data/{id}/attachments/{att_id}/

TODO

### 📖 Description

TODO

TODO

### 💭 Notes

TODO

### 👀 Preview steps

2. do this
3. do that
6. do that another thing

-->


kobo/apps/openrosa/apps/logger/xform_fs.py
```diff
@@ -1,35 +1,61 @@
...
 
+    @property
+    def mentioned_in_xml(self):
+        if not hasattr(self, '_mentioned_in_xml'):
+            parser = etree.XMLParser()
+            root = etree.fromstring(self.xml, parser=parser)
+            namespaces = root.nsmap
+            self._mentioned_in_xml = set(
+                root.xpath(
+                    '//*/text()',
```
p2edwards: Enketo seems to consistently use `type="file"`:

```xml
<sound_file type="file">wave.wav</sound_file>
<image_file type="file">thanks.png</image_file>
<any_file type="file">test.pdf</any_file>
```

I used the xpath `//*/[@type="file"]/text()` accordingly, to get a list of possible attachment names to filter by.

But that was too strict for the previous test, `test_importing_b1_and_b2`.

```xml
<?xml version='1.0' ?>
<tutorial id="tutorial">
	<name>A</name>
	<age>1</age>
	<picture>1330640038667.jpg</picture>
	<has_children>0</has_children>
	<gps />
	<web_browsers>ie</web_browsers>
	<meta>
		<instanceID>uuid:435f173c688e482486a48661700467gh</instanceID>
	</meta>
</tutorial>
```

I wonder how often that's seen in the wild? I wanted to err on the side of compatibility. So I switched to `//*/text()`, which is still stricter than some ideas but, say, doesn't consult the form definition to try to be exact about which fields are meant to be files.

Alternatives: 

- Be stricter than this, but report errors in `/bulk-submission` response instead of omitting files silently.

jnm: IMO we should match the behavior when handling single submissions and accept *all* attached files regardless of the XML submission content. In the future, we could be more discerning for all submission handling, but I want to minimize the discrepancies between this bulk code and the standard single-submission logic. (The rest of the PR does help achieve that.)

Additionally, we have to watch out for things like background audio and audit logs, which don't appear the same way in the submission XML as regular audio/video/file questions. Here's an example (probably not the best) of us having to deal with that: https://github.com/kobotoolbox/kobocat/pull/905.

@noliveleger I'll give you the determination here.

kobo/apps/openrosa/apps/logger/tests/bulk_submission/test_bulk_submission_attachments.py
```diff
@@ -0,0 +1,87 @@
...
+
+        basenames_count = 0
+        for instance in queryset.all():
+
+            # Check the database
+            expected = [
+                ('test.pdf', 'application/pdf'),
+                ('thanks.png', 'image/png'),
+                ('wave.wav', 'audio/x-wav'),
+            ]
```
jnm, line 62: Related to https://github.com/kobotoolbox/kpi/pull/6055/files#r2263019446, maybe we should include an audit CSV and background audio recording. Will leave the decision to @noliveleger as to whether it's worth delaying this fix for that.

I'll "request changes" here, but it's really because I want the `xform_fs` discussion to be resolved before merge.

jnm (changes requested):

kobo/apps/openrosa/apps/logger/xform_fs.py
```diff
@@ -44,15 +41,16 @@ def xml(self):
 
     @classmethod
     def is_valid_instance(cls, filepath):
-        if not filepath.endswith(".xml"):
+        if not filepath.endswith('.xml'):
             return False
         with open(filepath, 'r') as ff:
             fxml = ff.read().strip()
             if fxml.startswith('<?xml'):
+                # <?xml version="1.0" ?>
```
p2edwards, line 49: We accept single submissions without `<?xml version="1.0" ?>`. Should we remove the check here to match?

noliveleger: The question is: should we accept single submission without the declaration. 

See https://www.w3.org/TR/xml11/#sec-prolog-dtd

Let's keep it.

kobo/apps/openrosa/apps/logger/xform_fs.py
jnm: Resolved by removing `mentioned_in_xml` logic :ok_hand:

kobo/apps/openrosa/apps/logger/tests/bulk_submission/test_bulk_submission_attachments.py
jnm: Since including files is no longer contingent on XML parsing, I'm fine not having additional tests at this time

jnm (approved):

PR #6069, opened 2025-08-07
fix(auditLogs): use fast pagination

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Speed up the audit log endpoints.

### 💭 Notes
Use FastPagination on the audit log endpoints to avoid having to do a select distinct on a large number of fields on a large table.

### 👀 Preview steps
Run the server with --print-sql.

1. Go to `/api/v2/audit-logs/?q=log_type:asset-management&format=json`
2. 🔴 [on main] Notice there is a query in the logs for `SELECT COUNT(*) from (SELECT DISTINCT <lots of fields> from audit_log_auditlog...)`
3. 🟢 [on PR] Notice the query is now `SELECT COUNT(*) from (SELECT DISTINCT id FROM audit_log_auditlog...)`

### DIFF
--- a/kobo/apps/audit_log/views.py
+++ b/kobo/apps/audit_log/views.py
@@ -11,7 +11,7 @@ from kpi.models.import_export_task import (
     ImportExportTask,
     ProjectHistoryLogExportTask,
 )
-from kpi.paginators import FastPagination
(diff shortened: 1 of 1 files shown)


noliveleger (approved):
LGTM

PR #6085, opened 2025-08-11
feat(docs): change status field type to enum DEV-869

ZoeGodin:
### 📣 Summary
Changed status field type from string to enum to better represent the api.

### DIFF
--- a/kobo/apps/project_ownership/schema_extensions/v2/project_ownership/invites/extensions.py
+++ b/kobo/apps/project_ownership/schema_extensions/v2/project_ownership/invites/extensions.py
@@ -3,8 +3,10 @@ from drf_spectacular.plumbing import (
     build_array_type,
     build_basic_type,
     build_object_type,
+    build_choice_field,
 )
 from drf_spectacular.types import OpenApiTypes
+from rest_framework import serializers
 
 from kpi.schema_extensions.v2.generic.schema import (
     ASSET_URL_SCHEMA,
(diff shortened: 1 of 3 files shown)


kobo/apps/project_ownership/schema_extensions/v2/project_ownership/invites/extensions.py
```diff
@@ -15,6 +17,28 @@
...
+      ('accepted', 'Accepted'),
+      ('cancelled', 'Cancelled'),
+      ('declined', 'Declined'),
+      ('expired', 'Expired'),
+      ('pending', 'Pending'),
+      ('Resent', 'Resent'),
+    ]
+
+
+class EnumTest(OpenApiSerializerFieldExtension):
```
noliveleger: 👀  
I guess we could use a better class name.

kobo/apps/project_ownership/schema_extensions/v2/project_ownership/invites/extensions.py
```diff
@@ -15,6 +17,28 @@
...
+    target_class = 'kobo.apps.project_ownership.schema_extensions.v2.project_ownership.invites.fields.StatusEnumField'  # noqa
+
+    def map_serializer_field(self, auto_schema, direction):
+        return build_choice_field(
+            field=serializers.ChoiceField(
+                choices=MemberInviteStatus.CHOICES
```
noliveleger: What about using a constant directly in that class instead of declaring another class only for that purpose?

kobo/apps/project_ownership/schema_extensions/v2/project_ownership/invites/extensions.py
ZoeGodin: Indeed we could :see_no_evil:

kobo/apps/project_ownership/schema_extensions/v2/project_ownership/invites/extensions.py
ZoeGodin: Indeed, it would save up some space. Changed.

noliveleger (approved):
LGTM

PR #6087, opened 2025-08-11
 fix(docs): fix bad url building DEV-878 

ZoeGodin:
### 📣 Summary
Correct the URLs that are not being constructed correctly with the url_builder util.

### 👀 Preview steps
Go to `http://kf.kobo.local/api/v2/docs/`, when an endpoint has an URL, make sure it is properly constructed.

Bad example:
`/api/v2/user-detailurl/{username: bob}/`
Good example:
`/api/v2/users/bob/`


noliveleger (approved):
LGTM

PR #6110, opened 2025-08-13
fix(docs): fix `/me/` field with right values DEV-889

ZoeGodin:
### 📣 Summary

Update API documentation to match `git_rev` correct value in `/me/` endpoint.

### DIFF
--- a/kpi/schema_extensions/v2/me/extensions.py
+++ b/kpi/schema_extensions/v2/me/extensions.py
@@ -47,7 +47,7 @@ class GitRevFieldExtension(OpenApiSerializerFieldExtension):
                 'short': GENERIC_STRING_SCHEMA,
                 'long': GENERIC_STRING_SCHEMA,
                 'branch': GENERIC_STRING_SCHEMA,
-                'tag': build_basic_type(OpenApiTypes.BOOL),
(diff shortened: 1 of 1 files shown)


kpi/schema_extensions/v2/me/extensions.py
```diff
@@ -38,6 +38,20 @@ def map_serializer_field(self, auto_schema, direction):
...
+class GitRevFieldExtension(OpenApiSerializerFieldExtension):
+    target_class = 'kpi.schema_extensions.v2.me.fields.GitRevField'
+
+    def map_serializer_field(self, auto_schema, direction):
+        return build_object_type(
+            properties={
+                'short': GENERIC_STRING_SCHEMA,
+                'long': GENERIC_STRING_SCHEMA,
+                'branch': GENERIC_STRING_SCHEMA,
+                'tag': build_basic_type(OpenApiTypes.BOOL),
```
rajpatel24: Shouldn't `tag` be a string in some cases, rather than just a boolean?
All four can be either string or boolean false depending on Git command

ZoeGodin: All right, I will depict them all as strings then.

rajpatel24 (approved):

PR #6114, opened 2025-08-13
ci(biome): fix filter syntax DEV-862

Akuukis:
### 👀 Preview steps

1. :red_circle: see that #6092 didn't trigger Biome job
2. :green_circle: see that this PR triggers Biome job, although both change the same `.js` file


noliveleger (approved):
LGTM

PR #6122, opened 2025-08-14
fix(docs): fix `member` response and payload fields DEV-894

ZoeGodin:
### 📣 Summary
Fixed members response and payload fields in `/api/v2/organizations/{organization_id}/members/`

### DIFF
--- a/kpi/schema_extensions/v2/members/extensions.py
+++ b/kpi/schema_extensions/v2/members/extensions.py
@@ -1,12 +1,15 @@
 from drf_spectacular.extensions import OpenApiSerializerFieldExtension
-from drf_spectacular.plumbing import build_basic_type, build_object_type
+from drf_spectacular.plumbing import build_basic_type, build_object_type, \
+    build_choice_field
 from drf_spectacular.types import OpenApiTypes
 
 from kpi.schema_extensions.v2.generic.schema import (
     GENERIC_STRING_SCHEMA,
     USER_URL_SCHEMA,
 )
+from rest_framework import serializers
 from kpi.utils.schema_extensions.url_builder import build_url_type
+from .schema import CHOICES
 
 
 class InviteFieldExtension(OpenApiSerializerFieldExtension):
@@ -22,11 +25,16 @@ class InviteFieldExtension(OpenApiSerializerFieldExtension):
                 ),
                 'invited_by': USER_URL_SCHEMA,
                 'status': GENERIC_STRING_SCHEMA,
-                'invitee_role': GENERIC_STRING_SCHEMA,
-                'created': build_basic_type(OpenApiTypes.DATETIME),
+                'invitee_role': build_choice_field(
+                    field=serializers.ChoiceField(
+                        choices=CHOICES
+                    )
+                ),
+                'date_created': build_basic_type(OpenApiTypes.DATETIME),
                 'modified': build_basic_type(OpenApiTypes.DATETIME),
                 'invitee': GENERIC_STRING_SCHEMA,
-            }
+            },
+            required=['date_created', 'invitee'],
         )
 
 
@@ -38,6 +46,17 @@ class MemberUrlFieldExtension(OpenApiSerializerFieldExtension):
             'api_v2:organization-members-detail',
             organization_id='orgR6zUBwMHop2mgGygtFd6c',
             user__username='bob',
+        )
+
+
+class RoleChoiceFieldExtension(OpenApiSerializerFieldExtension):
+    target_class = 'kpi.schema_extensions.v2.members.fields.RoleChoiceField'
+
+    def map_serializer_field(self, auto_schema, direction):
+        return build_choice_field(
+            field=serializers.ChoiceField(
+                choices=CHOICES
+            )
         )
 
 
--- a/kpi/schema_extensions/v2/members/extensions.py
+++ b/kpi/schema_extensions/v2/members/extensions.py
@@ -3,6 +3,7 @@ from drf_spectacular.plumbing import build_basic_type, build_object_type, \
     build_choice_field
 from drf_spectacular.types import OpenApiTypes
 
+from kpi.models.import_export_task import ImportExportStatusChoices
 from kpi.schema_extensions.v2.generic.schema import (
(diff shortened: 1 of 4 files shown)


kpi/schema_extensions/v2/members/extensions.py
```diff
@@ -1,12 +1,16 @@
 from drf_spectacular.extensions import OpenApiSerializerFieldExtension
-from drf_spectacular.plumbing import build_basic_type, build_object_type
+from drf_spectacular.plumbing import build_basic_type, build_object_type, \
```
noliveleger: Please use  black format

kpi/schema_extensions/v2/members/extensions.py
```diff
@@ -1,12 +1,16 @@
...
 from drf_spectacular.types import OpenApiTypes
 
+from kpi.models.import_export_task import ImportExportStatusChoices
 from kpi.schema_extensions.v2.generic.schema import (
     GENERIC_STRING_SCHEMA,
     USER_URL_SCHEMA,
 )
+from rest_framework import serializers
 from kpi.utils.schema_extensions.url_builder import build_url_type
+from .schema import CHOICES, PAYLOAD_CHOICES
```
noliveleger: `CHOICES` and `PAYLOAD_CHOICES` are too generic. Please rename them to something more obvious.

kpi/schema_extensions/v2/members/schema.py
```diff
@@ -0,0 +1,10 @@
+CHOICES = [
+    ('admin', 'Admin'),
+    ('member', 'Member'),
```
noliveleger: Can we use already existing constants? `ORG_ADMIN_ROLE` and `ORG_MEMBER_ROLE`. BTW, I don't think the back end returns capitalized words . AFAIK, The API only returns the value of the constants.

kpi/schema_extensions/v2/members/schema.py
```diff
@@ -0,0 +1,10 @@
...
+]
+
+PAYLOAD_CHOICES = [
+    ('admin', 'Admin'),
+    ('member', 'Member'),
+    ('owner', 'Owner'),
+]
```
noliveleger: Same comment here ;-)

noliveleger (changes requested):

kpi/schema_extensions/v2/members/extensions.py
ZoeGodin: Formated

kpi/schema_extensions/v2/members/schema.py
ZoeGodin: Used the constants in the tuple.

kpi/schema_extensions/v2/members/schema.py
ZoeGodin: Same as previous comment

kpi/schema_extensions/v2/members/extensions.py
ZoeGodin: Names have been changed to better reflect the fields

kpi/schema_extensions/v2/members/extensions.py
```diff
@@ -21,12 +28,30 @@ def map_serializer_field(self, auto_schema, direction):
                     guid='f3ba00b2-372b-4283-9d57-adbe7d5b1bf1',
                 ),
                 'invited_by': USER_URL_SCHEMA,
-                'status': GENERIC_STRING_SCHEMA,
-                'invitee_role': GENERIC_STRING_SCHEMA,
+                'status': build_choice_field(
+                    field=serializers.ChoiceField(
+                        choices=ImportExportStatusChoices
```
noliveleger: Why don't you use `ROLE_CHOICES_PAYLOAD_ENUM_SCHEMA`  here?  
By the way, `ROLE_CHOICES_PAYLOAD_ENUM_SCHEMA`  and `ROLE_CHOICES_ENUM_SCHEMA` are not really schemas but enums. I would remove the `_SCHEMA` from their names for consistency.  

If `ImportExportStatusChoices` can be used everywhere `ROLE_CHOICES_PAYLOAD_ENUM_SCHEMA` is used, I would get rid of `ROLE_CHOICES_PAYLOAD_ENUM_SCHEMA` but ⚠️ maybe it's not possible because the app could not be ready before DRF-spectucular tries to load all its extensions.

ZoeGodin: If you are talking abut the status field it is because the enums are supposed to describe a role for an invitee, not the status of an invite. This is the reason we use the new status enum. I am not sure why we would get rid of `ROLE_CHOICES_PAYLOAD_ENUM_SCHEMA` as it describe the payload possibilities but I did change their names, so maybe you could add another comment here clarifying?

noliveleger: @ZoeGodin my comments make totally no sense :-P. Forget what I said. Sorry for the confusion.

noliveleger (approved):
LGTM

kpi/schema_extensions/v2/members/extensions.py
```diff
@@ -6,15 +6,14 @@
...
-from kpi.models.import_export_task import ImportExportStatusChoices
 from kpi.schema_extensions.v2.generic.schema import (
     GENERIC_STRING_SCHEMA,
     USER_URL_SCHEMA,
 )
 from rest_framework import serializers
 from kpi.utils.schema_extensions.url_builder import build_url_type
 from .schema import ROLE_CHOICES_ENUM, ROLE_CHOICES_PAYLOAD_ENUM
-
+from kobo.apps.project_ownership.schema_extensions.v2.project_ownership.invites.extensions import StatusEnumFieldExtension  # noqa
```
noliveleger: Please review the imports.

ZoeGodin: Used linter as I wasn't sure what else you meant by 'review the imports'.

noliveleger: Exactly what I meant.
Always use the linter and the formatter. 

As a reminder, that's why the checklist if present in the PR template ;-)

PR #6129, opened 2025-08-18
feat(auditLogs): optimize search filter for audit log endpoints

rgraber:
### 🗒️ Checklist

1. [ ] run linter locally

### 📣 Summary
Performance enhancement for audit log, access log, and project history log endpoints.

### 💭 Notes
The SearchFilter automatically runs `distinct` on all queries in order to handle the possibility that searching on an n-to-many field has returned multiple results for the same object. This distinct is very slow, so add a property to the view that tells the SearchFilter to skip it. This property should only be set on models that do not have any kind of n-to-many fields.

### 👀 Preview steps
This is best previewed by running the server with `--print-sql` or turning on the debug toolbar 

1. ℹ️ have an account
2. Navigate to `api/v2/audit-logs/?q=user__username:<username>
3. 🟢 [on PR] Notice the audit log query does not use `distinct` for the search or for the count

### DIFF
--- a/kobo/apps/audit_log/views.py
+++ b/kobo/apps/audit_log/views.py
@@ -165,7 +165,7 @@ class AuditLogViewSet(mixins.ListModelMixin, viewsets.GenericViewSet):
     filter_backends = (SearchFilter,)
     # audit logs have no n-to-many fields, so don't bother running "distinct" on
     # search results
-    requires_distinct = False
+    skip_distinct = True
 
     search_default_field_lookups = [
         'app_label__icontains',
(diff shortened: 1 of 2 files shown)


kobo/apps/audit_log/views.py
```diff
@@ -163,6 +163,9 @@ class AuditLogViewSet(mixins.ListModelMixin, viewsets.GenericViewSet):
         AuditLog.objects.select_related('user').all().order_by('-date_created')
     )
     filter_backends = (SearchFilter,)
+    # audit logs have no n-to-many fields, so don't bother running "distinct" on
+    # search results
+    skip_distinct = True
```
noliveleger, line 168: I would prefer a more obvious name. Unfortunately, I don't have a suggestion.

I asked AI for suggestions, guess what: it has suggested `skip_distinct` 🤦 

Well! Nvm.

PR #6131, opened 2025-08-18
feat(massEmails): get name from extra details DEV-903

rgraber:
### 🗒️ Checklist

1. [ ] run linter locally

### 📣 Summary
Use the user-entered name in mass email templates.

### 💭 Notes
When a user enters their name, it actually ends up in user.extra_details.data['name'] rather than in user.first_name + user.last_name. The latter is only inputted in Django admin. Use the user-inputted name for the full name when sending emails. 'Name' is always set in extra details via standardize_json_field so we shouldn't get any key errors, even if the user hasn't entered a name.

### 👀 Preview steps
This is most easily tested with `MASS_EMAILS_CONDENSE_SEND` set to true.

1. ℹ️ have an account
2. Add a full name in Account Settings
3. Add the user's email to Constance > MASS_EMAIL_TEST_EMAILS
4. Create a new mass email config using the `test_users` query that uses `##full_name##` somewhere in the template and set it to live
6. 🟢 At the next email send, the name entered in the account settings should be rendered in the email

### DIFF
--- a/kobo/apps/mass_emails/tasks.py
+++ b/kobo/apps/mass_emails/tasks.py
@@ -274,7 +274,7 @@ class MassEmailSender:
         plan_name = self.get_plan_name(org_user)
         data = {
             'username': record.user.username,
-            'full_name': record.user.first_name + ' ' + record.user.last_name,
+            'full_name': record.user.extra_details.data['name'],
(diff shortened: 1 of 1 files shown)


tinok (commented):
QA passed.

PR #6132, opened 2025-08-18
fix(synchronousExports): alleviate lock contention DEV-905

jnm:
### 📣 Summary

Synchronous export requests now return a 503 immediately when an identical request is already in progress, as opposed to hanging, which could contribute to service outages.

### 📖 Description

Within the time that a synchronous request is allowed to complete (typically 120 seconds), requests for in-progress, identical exports return a 503 immediately. When an export has failed, additional requests for that identical export will return a 500 without retrying until the synchronous export cache maximum age has been reached (typically 300 seconds).

1. Never block more than one uWSGI worker per synchronous export
2. Other attempts to create a synchronous export identical to one in progress will return a 503 to the client
3. The 503 will contain a `Retry-After` HTTP header directing the client how long to wait before retrying
4. Prevent failed synchronous exports from being reattempted within the window of `SYNCHRONOUS_EXPORT_CACHE_MAX_AGE` (Constance setting)
5. Attempt to provide useful, human-readable messages to the client in addition to the `Retry-After` header

### 👀 Preview steps

:warning: Google Chrome refuses to send simultaneous identical HTTP requests from different tabs. If you want to test the behavior in a browser, use Firefox, or append a unique query string to the synchronous export URL for each request.

1. Check out the base branch for this PR (currently `release/2.025.29`)
2. Create a project with some data
3. Add an artificial delay in the export process using this patch:
```diff
diff --git a/kpi/models/import_export_task.py b/kpi/models/import_export_task.py
index f6855c8e1..d2288295b 100644
--- a/kpi/models/import_export_task.py
+++ b/kpi/models/import_export_task.py
@@ -933,6 +933,7 @@ class SubmissionExportTaskBase(ImportExportTask):
                 'are valid export types'
             )
 
+        import time; time.sleep(15)
         export, submission_stream = self.get_export_object()
         filename = self._build_export_filename(export, export_type)
         absolute_filepath = self.get_absolute_filepath(filename)
```
4. Start `pg_activity` locally (syntax hint: `PGPASSWORD=<password> pg_activity --host <host> --port <port> --user <user>`)
5. Run this quick-and-dirty load generator:
```bash
while true; do
    running_curl_count=$(pgrep curl | wc -l)
    if [ "$running_curl_count" -lt 50 ]; then
        (
            curl -v -L \
                --user <kpi username>:<kpi password> \
                'http://<kpi host>/api/v2/assets/<asset uid>/export-settings/<export settings uid>/data.csv' \
            2>&1 | grep -E 'HTTP|[Rr]etry|_uuid'
        ) &
    fi
done
```
6. See a bunch of locks pile up in `pg_activity` :lock: :exclamation:
7. Stash the patch, check out the branch for this PR, and reapply the patch
8. Important: use the Django shell to delete the `SubmissionSynchronousExport` object; otherwise, the export result will be cached and returned immediately
9. Rerun the load generator
10. Make sure locks do not pile up

### DIFF
--- a/kpi/models/import_export_task.py
+++ b/kpi/models/import_export_task.py
@@ -21,7 +21,6 @@ from django.db import models, transaction
 from django.db.models import CharField, F, Value
 from django.db.models.functions import Concat
 from django.db.models.query import QuerySet
-from django.urls import reverse
 from django.utils import timezone
 from django.utils.translation import gettext as t
 from formpack.constants import KOBO_LOCK_SHEET
@@ -38,6 +37,7 @@ from openpyxl.utils.exceptions import InvalidFileException
 from private_storage.fields import PrivateFileField
 from pyxform.xls2json_backends import xls_to_dict, xlsx_to_dict
 from rest_framework import exceptions
+from rest_framework.reverse import reverse
 from werkzeug.http import parse_options_header
 
 from kobo.apps.reports.report_data import build_formpack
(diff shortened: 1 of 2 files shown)


kpi/models/import_export_task.py
```diff
@@ -24,6 +23,13 @@
...
+from openpyxl.utils.exceptions import InvalidFileException
+from private_storage.fields import PrivateFileField
+from pyxform.xls2json_backends import xls_to_dict, xlsx_to_dict
+from rest_framework import exceptions
+from werkzeug.http import parse_options_header
+
+import formpack
 from formpack.constants import KOBO_LOCK_SHEET
 from formpack.schema.fields import (
     IdCopyField,
```
noliveleger: I don't know why darker wants it in the third block but the imports were perfectly fine before.

jnm: Thanks for fixing :+1:

PR #6133, opened 2025-08-18
fix(attachment): avoid blocking migration

noliveleger:
### 📣 Summary
Use `SKIP_HEAVY_MIGRATIONS` flag to alter `logger_attachment` table to drop `NULL` on specific fields


Guitlle (approved):
LGTM

PR #6135, opened 2025-08-19
fix(nlp): Permission error in qualitative analysis UI DEV-162

Guitlle:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Fix an error in the transcription screen UI that expected the permission manage_asset to allow translation. It was replaced by the change_submissions

### 👀 Preview steps
1. You need 2 users to test, have NLP enabled and working
2. Create a project, or use an existing one, with audio fields for transcription and translation available
3.  Share the project with the secondary user with partial permissions to add and change submissions but no asset management permissions
4. Log in with the secondary user and attempt to make a translation, it should be allowed


kpi/tests/api/v2/test_api_assets.py
```diff
@@ -869,6 +869,34 @@ def test_project_views_for_anotheruser_can_view_reports(self):
...
+                    'languages': ['en']
+                },
+                'translation': {
+                    'languages': []
+                },
+            },
+        }
+
+        response = self.client.patch(asset_detail_url, data, format='json')
+        assert response.status_code == 403
```
noliveleger: let's use `status.HTTP_<code>` instead of the integer representation. (i.e.:  `== status.HTTP_403_FORBIDDEN`). 
Please do the same for subsequent assertions.

Guitlle: Ok

kpi/permissions.py
```diff
@@ -154,6 +162,9 @@ def has_object_permission(self, request, view, obj):
         ) or (
             method == 'GET'
             and user_has_project_view_asset_perm(obj, user, PERM_VIEW_ASSET)
+        ) or (
+            method == 'PATCH'
+            and user_can_modify_advanced_features(obj, user, request)
```
noliveleger: why not grouping all `PATCH` conditions altogether?

Guitlle: Yeah, I think that's better

PR #6140, opened 2025-08-19
fix(docs): use only JSONRenderer as the default renderer DEV-904

ZoeGodin:
### 📣 Summary
Use only JSONRenderer as the default renderer and remove BrowsableAPIRenderer (exept in openrosa endpoints)

### 📖 Description
This PR cleans the `api/v2` renderers from all `browsableApiRenderers` and assure a level of cleanliness to the renderers as well as making sure we are using the JSONRenderer as our default.


kobo/apps/openrosa/apps/api/viewsets/xform_submission_api.py
```diff
@@ -125,9 +125,10 @@ class XFormSubmissionApi(
     filter_backends = (filters.AnonDjangoObjectPermissionFilter,)
     model = Instance
     permission_classes = (permissions.AllowAny,)
-    renderer_classes = (TemplateXMLRenderer,
-                        JSONRenderer,
-                        BrowsableAPIRenderer)
```
noliveleger, line 130: Please leave the BrowsableAPIRenderer for OpenRosa for now. 
We will remove it entirely when `v1` is gone.

kpi/views/v2/asset_usage.py
```diff
@@ -35,7 +35,9 @@ class AssetUsageViewSet(ListModelMixin, viewsets.GenericViewSet):
     - docs/api/v2/asset_usage/list.md
     """
 
-    renderer_classes = (renderers.JSONRenderer,)
+    renderer_classes = [
+        renderers.JSONRenderer,
+    ]
```
noliveleger: You can keep it as a tuple and it could stay on one line.

noliveleger (changes requested):
I would like to see `renderer_classes` to be removed entirely in the viewset if it only needs `JSONRenderer` and the default in the settings should be only `JSONRenderer`. 
Let's override `renderer_classes` only when we need other renderers. 
Like I said, do not touch `OpenRosa` viewsets.

kpi/views/v2/asset_usage.py
ZoeGodin: I guess this is not really necessary if I am to follow your direction on this comment, no?
[comment](https://github.com/kobotoolbox/kpi/pull/6140#pullrequestreview-3134031806)

noliveleger: Totally ;-)

kobo/apps/openrosa/apps/api/viewsets/xform_submission_api.py
ZoeGodin: rewrote it in the code

noliveleger (approved):
LGTM

PR #6142, opened 2025-08-20
feat(dataCollectors): set up initial models and admin DEV-786

rgraber:
### 🗒️ Checklist

1. [ ] run linter locally

### 📣 Summary
Add new data collector models to Django admin.

### 📖 Description
Adds DataCollector and DataCollectorGroup models to Django admin. A DataCollectorGroup may have several DataCollectors and be associated with several Assets. Every DataCollector has a token, which may be rotated using Django admin.

### 💭 Notes
The admin for DataCollectors and DataCollectorGroups is extremely rudimentary because it turns out making a nice interface for one-to-many relationships is hard. For this PR, the user can add assets to a DataCollectorGroup by selecting from a giant checklist of all available assets. In the future, this will be enhanced with an autocomplete that allows them to search for assets before adding them. Also in the future, users will be able to add DataCollectors on the DataCollectorGroup page rather than having to go add the group to the DataCollector objects. These efforts are part of DEV-920 in Linear.

TODO

### 👀 Preview steps

2. do this
3. do that
6. do that another thing


kobo/apps/kobo_auth/models.py
```diff
@@ -87,3 +90,37 @@ def sync_to_openrosa_db(self):
...
+            self.token = self.generate_key()
+        super().save()
+
+    def rotate_token(self):
+        self.token = None
+        self.save()
+
+    @classmethod
+    def generate_key(cls):
+        return secrets.token_hex(20)
```
noliveleger: Let's use `secrets.token_urlsafe()` instead.

kobo/apps/kobo_auth/models.py
```diff
@@ -87,3 +90,37 @@ def sync_to_openrosa_db(self):
...
+        settings.AUTH_USER_MODEL, null=True, on_delete=models.SET_NULL
+    )
+    uid = KpiUidField(uid_prefix='dcg', primary_key=True)
+    name = fields.CharField(max_length=200)
+
+    def __str__(self):
+        return '{} ({})'.format(self.name, self.uid)
+
+
+class DataCollector(models.Model):
```
noliveleger: Make it inherit from AbstractTimestampModel. 
It always useful for debugging purpose (to have the creation, modification dates).

kobo/apps/kobo_auth/models.py
```diff
@@ -87,3 +90,37 @@ def sync_to_openrosa_db(self):
...
+    uid = KpiUidField(uid_prefix='dc', primary_key=True)
+    name = fields.CharField(null=True, blank=True, max_length=200)
+    token = fields.CharField(max_length=40)
```
noliveleger: I would leave a little more room enough if our token is smaller. In the future, if we want to increase the token length,  we won't have to create another migration.  
What about 64 characters?  (e.g. hashlib.sha256()) 

Also, I think, the field should be unique (we don't want two tokens to be the same) and indexed (the authentication will use it 100% of the time in the `WHERE` clause).

noliveleger (changes requested):
Please move data collector stuff to its own django app.

kobo/apps/kobo_auth/test_models.py
noliveleger, line 1: Would you mind to move it to its own folder `tests/` please. 
I can imagine we are going to add more tests and files. I'd rather see them grouped inside the same folder.

noliveleger (changes requested):
Please move data collector stuff to its own django app.

kobo/apps/data_collectors/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+
+class DataCollectorsConfig(AppConfig):
+    default_auto_field = 'django.db.models.BigAutoField'
+    name = 'kobo.apps.data_collectors'
+    verbose_name = 'Data collectors'
+
+    def ready(self):
+        from kobo.apps.data_collectors import signals  # noqa
+
+        super().ready()
```
noliveleger, line 12: Would you mind to move it to `apps`?  FYI, there is a task in linear to make it consistent and move all others in their respective `apps.py`.
Since this one is new code, let's do this directly. (but if you don't want to mess with that, fair enough, it will be done while tackling the other task I just spoke about).

rgraber: Moving this around is causing errors so I'll let it be done with the next task

PR #6152, opened 2025-08-22
fix(instanceHistory): Persist instance history when instance is deleted DEV-752

Guitlle:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
The InstanceHistory objects are not deleted when the referenced instance is deleted anymore. They remain for an configurable amount of days after which they are collected and purged 

### 👀 Preview steps
This is covered in an unit test, it doesn't need preview


rgraber (approved):
LGTM

PR #6165, opened 2025-08-26
feat(instanceHistory): Celery task to clean instance history records DEV-753

Guitlle:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
InstanceHistory objects are kept for an amount of days given by the constance parameter SUBMISSION_HISTORY_GRACE_PERIOD. After this period has passed, the records that have no associated instance (xform_instance = NULL) are deleted by a celery task that runs weekly.

### 👀 Preview steps
This is covered with unit tests. However, to test you could:

1. Have a project with submissions
2. Delete some submissions 
3. Configure the constance parameter SUBMISSION_HISTORY_GRACE_PERIOD to 0 days. 
4. Get a shell (`./manage.py shell_plus`, via docker exec bash)  and inspect the InstanceHistory objects
```
InstanceHistory.objects.filter(xform_instance__isnull=True).count()
```
5. Manually call the task, from the shell:
```
from kobo.apps.openrosa.apps.logger.tasks import delete_expired_instance_history_records

delete_expired_instance_history_records()
```
6. Inspect again the count `InstanceHistory.objects.filter(xform_instance__isnull=True).count()`

### DIFF
--- a/kobo/apps/openrosa/apps/logger/tasks.py
+++ b/kobo/apps/openrosa/apps/logger/tasks.py
@@ -217,11 +217,13 @@ def delete_expired_instance_history_records(chunk_size=10000, max_records=100000
         xform_instance=None,
     ).values_list('id', flat=True)[:max_records]
     n_objs = len(history_ids)
-    logging.warning(f'Found {n_objs} expired InstanceHistory objects. Cleaning up ...')
+    logging.info(f'Found {n_objs} expired InstanceHistory objects. Cleaning up ...')
     for page_start in range(0, n_objs, chunk_size):
         batch_ids = history_ids[page_start:page_start + chunk_size]
         logging.warning(
             f'Deleting batch of {len(batch_ids)} InstanceHistory objects...'
         )
         queryset = InstanceHistory.objects.filter(pk__in=batch_ids)
+        # We use _raw_delete to avoid going through the ORM delete process.
+        # This model is very simple and can be deleted directly without issues
         queryset._raw_delete(queryset.db)
--- a/kobo/apps/openrosa/apps/logger/tasks.py
+++ b/kobo/apps/openrosa/apps/logger/tasks.py
@@ -219,7 +219,7 @@ def delete_expired_instance_history_records(chunk_size=1e4, max_records=1e6):
     n_objs = len(history_ids)
     logging.warning(f'Found {n_objs} expired InstanceHistory objects. Cleaning up ...')
     for page_start in range(0, n_objs, chunk_size):
-        batch_ids = history_ids[page_start : page_start + chunk_size]
+        batch_ids = history_ids[page_start:page_start + chunk_size]
         logging.warning(
             f'Deleting batch of {len(batch_ids)} InstanceHistory objects...'
         )
--- a/kobo/apps/openrosa/apps/logger/tasks.py
+++ b/kobo/apps/openrosa/apps/logger/tasks.py
@@ -208,7 +208,7 @@ def sync_storage_counters(**kwargs):
 
 
 @celery_app.task
-def delete_expired_instance_history_records(chunk_size=1e4, max_records=1e6):
+def delete_expired_instance_history_records(chunk_size=10000, max_records=1000000):
     threshold_time = timezone.now() - timedelta(
         days=config.SUBMISSION_HISTORY_GRACE_PERIOD
     )
--- a/kobo/apps/openrosa/apps/logger/tasks.py
+++ b/kobo/apps/openrosa/apps/logger/tasks.py
@@ -8,6 +8,7 @@ from datetime import timedelta
 from io import StringIO
 
 from celery import shared_task
+from constance import config
 from dateutil import relativedelta
 from django.conf import settings
 from django.core.management import call_command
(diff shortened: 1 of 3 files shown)


kobo/apps/openrosa/apps/logger/tasks.py
```diff
@@ -203,3 +205,23 @@ def sync_storage_counters(**kwargs):
         sync=True,
         no_lock=no_lock,
     )
+
+
+@celery_app.task
+def delete_expired_instance_history_records(chunk_size=1e4, max_records=1e6):
```
rajpatel24: The default args here use floats (`1e4`, `1e6`). That leads to a `TypeError` when passed into queryset slicing. Could we switch these to integer literals?

Guitlle: Ok

kobo/apps/openrosa/apps/logger/tasks.py
```diff
@@ -203,3 +205,23 @@ def sync_storage_counters(**kwargs):
...
+def delete_expired_instance_history_records(chunk_size=10000, max_records=1000000):
+    threshold_time = timezone.now() - timedelta(
+        days=config.SUBMISSION_HISTORY_GRACE_PERIOD
+    )
+    history_ids = InstanceHistory.objects.filter(
+        date_created__lte=threshold_time,
+        xform_instance=None,
+    ).values_list('id', flat=True)[:max_records]
+    n_objs = len(history_ids)
+    logging.warning(f'Found {n_objs} expired InstanceHistory objects. Cleaning up ...')
```
rajpatel24: Shouldn't this be `logging.info` instead of `logging.warning`, since this isn't actually a warning? What do you think?

kobo/settings/base.py
```diff
@@ -1294,6 +1300,11 @@ def dj_stripe_request_callback_method():
     'delete-daily-xform-submissions-counter': {
         'task': 'kobo.apps.openrosa.apps.logger.tasks.delete_daily_counters',
         'schedule': crontab(hour=0, minute=0),
+        'options': {'queue': 'kobocat_queue'},
+    },
+    'delete-expired-instance-history-records': {
+        'task': 'kobo.apps.openrosa.apps.logger.tasks.delete_expired_instance_history_records',  # noqa
+        'schedule': crontab(day_of_week=1, hour=10, minute=0),
```
rajpatel24: This will run the task once a week. Was a weekly schedule chosen due to the expected low volume? If the data volume grows, running it weekly could lead to a large backlog, potentially resulting in long-running deletions.

kobo/apps/openrosa/apps/logger/tasks.py
```diff
@@ -203,3 +205,23 @@ def sync_storage_counters(**kwargs):
...
+    for page_start in range(0, n_objs, chunk_size):
+        batch_ids = history_ids[page_start:page_start + chunk_size]
+        logging.warning(
+            f'Deleting batch of {len(batch_ids)} InstanceHistory objects...'
+        )
+        queryset = InstanceHistory.objects.filter(pk__in=batch_ids)
+        queryset._raw_delete(queryset.db)
```
rajpatel24, line 229: We’re using `_raw_delete`, which is an internal Django API. It works and is faster, but it also skips signals and isn’t part of the public API. If the intention here is purely performance, could we add a short comment explaining why we prefer it over `queryset.delete()`?

kobo/apps/openrosa/apps/logger/tests/test_tasks.py
```diff
@@ -0,0 +1,66 @@
...
+            )
+
+        chunk_size = 5
+        max_records = 20
+        days_threshold = config.SUBMISSION_HISTORY_GRACE_PERIOD
+        now_mock = patch(
+            'django.utils.timezone.now',
+            return_value=d0 + timedelta(days=days_threshold - 1),
+        )
+        now_mock.start()
```
rajpatel24: Would it be better to use `with patch(...)` instead of `start()` and `stop()` to prevent dangling mocks in case the test fails?

rajpatel24 (changes requested):

kobo/apps/openrosa/apps/logger/tasks.py
Guitlle: Ok, I think it make sense

kobo/apps/openrosa/apps/logger/tasks.py
Guitlle: Ok, yeah, it's a faster delete because it uses a very simple query and this model is not complex and doesn't need any special considerations for the deletion of the records

kobo/apps/openrosa/apps/logger/tests/test_tasks.py
Guitlle: Ok, sure, we can change it to the with statements

kobo/settings/base.py
Guitlle: I didn't agreed with other people on this period of time. I think it works ok because there are not that many deleted submissions that often. Do you propose a better period of time?

rajpatel24 (approved):
LGTM

PR #6188, opened 2025-08-29
fix(exports): use empty object instead of null for empty payload

duvld:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Fixes an issue where exporting project history activity results in a `403`, citing a `CSRF token` error.

### 💭 Notes

This updates a fix to DEV-833 where we are using the wrong parameters for `fetchPost`. Since this call doesn't need a payload that fix used `null` which caused this bug

### 👀 Preview steps

2. ℹ️  have some email service set up
3. navigate to a project's settings > activity
4. ensure there is something to export (edit the form and save it)
5. click "export all data"
6. 🔴 [on main] notice that there is a error toast message and no export is created
7. 🟢 [on PR] notice that the export is created as expected


rgraber (approved):
Works for me but definitely wait for Leszek

magicznyleszek (approved):

PR #6199, opened 2025-09-02
fix(frontend): roboto font loading DEV-887

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 💭 Notes

Firefox has a feature that protects users from being targeted using their available system fonts. This feature causes our Roboto font to not be properly loaded when using Private browsing. I'm updating the way in which we load Roboto to avoid the problem.

### 👀 Preview steps

1. Use Private browsing in Firefox
2. Open any KPI route (don't even need to be logged in)
3. 🔴 [on main] notice that the font used is the default sans-serif browser font
4. 🟢 [on PR] notice that the font used is Roboto

ℹ️ The easiest way is to use Dev Tools, and in "Inspector" tab selecting "Fonts" sub-tab on the right. "All Fonts on Page" should list only `k-icons` and few different `Roboto` (all comming from `/static/compiled/…`).


jsapp/js/main.js
```diff
@@ -10,6 +10,16 @@
...
 import 'jquery-ui/ui/widgets/sortable'
 import '@mantine/core/styles.css'
+// We import all the weights and styles we actually use here to avoid unnecessary weight (pun intended).
+import '@fontsource/roboto/400.css'
+import '@fontsource/roboto/400-italic.css'
+import '@fontsource/roboto/500.css'
+// Roboto doesn't have semibold (600) version. We do use 600 font weight a lot in our styles, but they all are rendered
+// as weight 700. To have true 600, we would need to switch to variable font Roboto Flex - but this would require more
+// refactor, and AFAIK would also not play nicely with Mantine which uses "Roboto" as default font and doesn't seem
+// ready for usng a variable font.
```
pauloamorimbr: typo

```suggestion
// ready for using a variable font.
```

jsapp/js/main.js
```diff
@@ -10,6 +10,16 @@
  */
 import 'jquery-ui/ui/widgets/sortable'
 import '@mantine/core/styles.css'
+// We import all the weights and styles we actually use here to avoid unnecessary weight (pun intended).
+import '@fontsource/roboto/400.css'
+import '@fontsource/roboto/400-italic.css'
+import '@fontsource/roboto/500.css'
+// Roboto doesn't have semibold (600) version. We do use 600 font weight a lot in our styles, but they all are rendered
```
pauloamorimbr, line 17: This is really weird. Google does provide the roboto 600, I've converted it to woff and woff2 to use in other projects already. IDK why fontsource doesn't provide it!

pauloamorimbr (approved):
I tried some different ways and couldn't reproduce the firefox issue on my computer. Maybe it's a version thing..

Anyways, code wise seems good. Nothing breaks, everything works and feels like an improvement! GJ!

left a typo nitpick comment only

sksondra (commented):
Good from QA.

PR #6203, opened 2025-09-03
feat(dataCollectors): create authentication class for data collectors DEV-787

rgraber:
### 🗒️ Checklist

1. [ ] run linter locally

### 💭 Notes
Developer-only changes. Sets up the authentication class we will use to authenticate DataCollectors, who have authentication tokens but are not actual Django users. 
Since we need some kind of user to verify permissions, we create a new "fake" user class, DataCollectorUser, that inherits from AnonymousUser. This type of user has a list of assets it is allowed to view/submit to.
When we authenticate, we grab the token from the url and use it to look up a DataCollector object. If there is one, we create a new DataCollectorUser and give it a list of assets corresponding to the assets belonging to the DataCollector's group. This DataCollecotrUser's only permissions are PERM_ADD_SUBMISSION on any asset in its whitelist.

### 👀 Preview steps
None, dev-only.

### DIFF
--- a/kobo/apps/data_collectors/authentication.py
+++ b/kobo/apps/data_collectors/authentication.py
@@ -1,0 +1,46 @@
+from django.contrib.auth.models import AnonymousUser
+from rest_framework.authentication import BaseAuthentication
+
+from kobo.apps.data_collectors.models import DataCollector
+from kobo.apps.openrosa.apps.logger.models import XForm
+from kpi.constants import PERM_ADD_SUBMISSIONS
+
+
+class DataCollectorUser(AnonymousUser):
+    @property
+    def is_authenticated(self):
+        # Always return True. This is a way to tell if
+        # the user has been authenticated in permissions
+        return True
+
+    assets = []
+    name = None
+
+    def has_perm(self, perm, obj=...):
+        if perm != PERM_ADD_SUBMISSIONS:
+            return False
+        if not isinstance(obj, XForm):
+            return False
+        return obj.kpi_asset_uid in self.assets
+
+
+class DataCollectorTokenAuthentication(BaseAuthentication):
+    def authenticate(self, request):
+        context = request.parser_context
+        kwargs = context.get('kwargs', {})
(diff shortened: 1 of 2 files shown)


kobo/apps/data_collectors/authentication.py
```diff
@@ -0,0 +1,47 @@
...
+        try:
+            collector = DataCollector.objects.get(token=key)
+            server_user = DataCollectorUser()
+            group = collector.group
+            if group:
+                server_user.assets = list(group.assets.values_list('uid', flat=True))
+                server_user.name = collector.name
+            return server_user, key
+        except DataCollector.DoesNotExist:
+            raise DataCollector.AuthenticationFailed('Invalid token.')
```
rajpatel24: Is `AuthenticationFailed` actually defined on the `DataCollector` class? If not, this might raise an `AttributeError`. Should it instead be imported from somewhere like `rest_framework.exceptions`?

kobo/apps/data_collectors/authentication.py
```diff
@@ -0,0 +1,47 @@
+from django.contrib.auth.models import AnonymousUser
+from rest_framework.authentication import BaseAuthentication
+
+from kobo.apps.data_collectors.models import DataCollector
+from kobo.apps.openrosa.apps.logger.models import XForm
+from kpi.constants import PERM_ADD_SUBMISSIONS
+
+
+class DataCollectorUser(AnonymousUser):
```
rajpatel24, line 10: `DataCollectorUser` subclasses `AnonymousUser` and overrides `is_authenticated` to return `True`, but it still inherits `is_anonymous = True` by default. This could potentially lead to confusion in permission checks or any logic that relies on `is_anonymous`. Would it make sense to override `is_anonymous` to return `False` here?

rajpatel24 (changes requested):

kobo/apps/data_collectors/authentication.py
rgraber: Good catch

PR #6207, opened 2025-09-04
fix(audioRecording): background audio file being soft deleted on creation DEV-976

pauloamorimbr:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
This PR adds the pattern for the background audio file generated by Enketo to the fix to avoid it being soft-deleted on creation

### 👀 Preview steps
1. ℹ️ have an account
2. ℹ️ have a project with background audio and another media, like audio question
3. ⚠️ The current main of Enketo needs to be used to test this
4. Post a submission
5. 🔴 [on main] the generated background audio file gets deleted (only the name of the file appears in the data-table and it's marked as deleted on database)
6. 🟢 [on PR] background audio file appears normally

### DIFF
--- a/kobo/apps/openrosa/libs/utils/logger_tools.py
+++ b/kobo/apps/openrosa/libs/utils/logger_tools.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import contextlib
 import hashlib
+import logging
 import os
 import re
 import sys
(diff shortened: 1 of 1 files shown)


sksondra (commented):
Good from QA.

PR #6229, opened 2025-09-09
Revert "feat(assetVersions): remove reversion version foreign key DEV-838"

jnm:
`kpi/migrations/0070_alter_assetversion_reversion_version.py` attempts to rename the column `_reversion_version` to `_reversion_version_id` in the table `kpi_assetversion`. Unfortunately, this can fail for existing databases because the column was already called `_reversion_version_id` in the first place, owing to Django's column-naming conventions for `model.OneToOneField`.

I would suggest redoing this PR in a way that changes `_reversion_version = models.OneToOneField(…)` to `_reversion_version = models.IntegerField(…, db_column='_reversion_version_id)`. That would have the added bonus of avoiding the need to rename a bunch of Python `_reversion_version` references to `_reversion_version_id`.


jnm:
Nevermind. Because `_reversion_version` (the `OneToOneField`) could also be accessed as `_reversion_version_id`, reverting this would just create more of a mess. I'll see if I can combine the field change and rename in a single migration to solve the issues with existing databases

rgraber:
I'm unable to revert 0069 after running this

rgraber (approved):
LGTM

jnm:
Thanks! I edited the preview steps just now to account for the problem you observed. Here's the relevant snippet from our internal discussion:
> After step (7), a migration called kpi.0069_alter_assetversion_reversion_version has already been logged by Django (in the django_reversions table), and so [the original] step (9) does not actually run the new work I'd done :facepalm:

PR #6234, opened 2025-09-11
feat(billing): switch asset usage endpoint to limit offset pagination DEV-987

jamesrkiger:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Removes custom asset usage page-size pagination, thereby using KPI's default limit-offset pagination, and adjusts drf_spectacular code for the endpoint to improve API docs/frontend helpers.

### 💭 Notes
I am marking this as a draft for the time being, but it is ready for review. We will merge it as part of the corresponding frontend changes (See [DEV-1013](https://linear.app/kobotoolbox/issue/DEV-1013/migrate-asset-usage-table-to-universaltable-component)).

### 👀 Preview steps
No need for testing on this. Testing will come with frontend PR.


kpi/serializers/v2/service_usage.py
```diff
@@ -1,19 +1,29 @@
...
 from kobo.apps.organizations.utils import get_billing_dates
 from kpi.deployment_backends.openrosa_backend import OpenRosaDeploymentBackend
-from kpi.models.asset import Asset
+from kpi.models.asset import Asset, AssetDeploymentStatus
+from kpi.schema_extensions.v2.organizations.fields import UrlField
+from kpi.utils.schema_extensions.fields import HyperlinkedIdentityFieldWithSchemaField
 from kpi.utils.usage_calculator import ServiceUsageCalculator
 
 
+class NlpUsageSerializer(serializers.Serializer):
```
jamesrkiger: This is the main difference from how I've seen the drf_spectacular code used in other parts of the app. I thought it would be preferable to use the actual AssetUsageSerializer in `views.py`, rather than creating an inline serializer. To me, it makes the code more readable. But I can also see that there might be some confusion of concerns, since the NlpUsageSerializer here is being passed into `@extend_schema_field` rather than used directly in the AssetUsageSerializer. Hoping to get your input on this, @noliveleger.

noliveleger: @jamesrkiger , LGTM but since this serializer is only for extensions, I would like to move it to `schema_extensions/v2/organizations/serializers.py` even if it's not an inline serializer class.

PR #6238, opened 2025-09-11
fix(sso): do not allow sso users to change email from account settings DEV-1005

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Do not allow users who use SSO login to change their emails via the Account Settings.

### 💭 Notes
Update the EmailAddress serializer to not create new EmailAddresses if the request user has SSO enabled. We can check this by seeing if the user has any linked SocialAccounts.

### 👀 Preview steps
Enable SSO locally 

2. Enable SSO on the account
3. Go to Account Settings > Security
4. Open the network tab
5. Try to update your email
6. 🔴 [on release branch] The request will succeed. You will get a message "Check your email <new_email>..."
7. 🟢 [on PR] The request will return 400. There will be no error in the UI but that is something frontend is working on

### DIFF
--- a/kobo/apps/accounts/serializers.py
+++ b/kobo/apps/accounts/serializers.py
@@ -30,7 +30,7 @@ class EmailAddressSerializer(serializers.ModelSerializer):
         """
         user = self.context['request'].user
         organization = user.organization
-        if user.socialaccount_set.first():
+        if user.socialaccount_set.exists():
             raise serializers.ValidationError(
                 {'email': t('This action is not allowed.')}
             )
--- a/kobo/apps/accounts/serializers.py
+++ b/kobo/apps/accounts/serializers.py
@@ -26,10 +26,14 @@ class EmailAddressSerializer(serializers.ModelSerializer):
     def validate(self, attrs):
         """
         Validates that only owners or admins of the organization can update
-        their email
+        their email and only if they don't have SSO
         """
         user = self.context['request'].user
         organization = user.organization
+        if user.socialaccount_set.first():
+            raise serializers.ValidationError(
+                {'email': t('This action is not allowed.')}
+            )
         if organization.is_owner(user) or organization.is_admin(user):
             return attrs
         raise serializers.ValidationError(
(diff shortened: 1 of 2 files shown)


kobo/apps/accounts/tests/test_email.py
```diff
@@ -171,3 +173,22 @@ def test_that_non_mmo_user_can_update_email(self):
...
+        else:
+            user = self.non_mmo_user
+        baker.make('socialaccount.SocialAccount', user=user)
+        self.client.force_login(user)
+        data = {'email': 'nonmmo@example.com'}
+        res = self.client.post(self.url_list, data, format='json')
+        self.assertEqual(res.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(
+            user.emailaddress_set.filter(email=data['email']).count(), 0
+        )
```
jnm, line 194: This is fine in and of itself, but it made me wonder about where we store email addresses that are received from SSO providers. I was going to propose a change like
```python
self.assertEqual(
    list(user.emailaddress_set.values_list('email', flat=True)),
    ['address@beforeattempt.fun']
)
```
but currently `emailaddress_set` is empty—maybe that's just due to the way the test data is created, though, and not indicative of real-world scenarios:
```
193  ->         self.assertEqual(
194                 user.emailaddress_set.filter(email=data['email']).count(), 0
195             )
(Pdb) user.emailaddress_set.exists()
False
```

kobo/apps/accounts/serializers.py
```diff
@@ -26,10 +26,14 @@ def create(self, validated_data):
     def validate(self, attrs):
         """
         Validates that only owners or admins of the organization can update
-        their email
+        their email and only if they don't have SSO
         """
         user = self.context['request'].user
         organization = user.organization
+        if user.socialaccount_set.first():
```
jnm: I would say `exists()` is idiomatic, unless there's a need to retrieve the first object and check that it's falsy. But I only mention that because there may be a future scenario where avoiding fetching the entire object makes a difference (this is not such a scenario, haha)

kobo/apps/accounts/tests/test_email.py
rgraber: Using baker means we don't trigger any signals or anything that might fire to create EmailAddresses so it's expected that the email addresses will be empty in this case. All the machinery to create EmailAddresses is within django_allauth

PR #6243, opened 2025-09-13
feat(api): add `/user-reports/` endpoint for superusers

RuthTurk:
### 🗒️ Checklist

1. [ ] run linter locally

### 📣 Summary

TODO

### 📖 Description

TODO

TODO

### 💭 Notes

TODO

### 👀 Preview steps

2. do this
3. do that
6. do that another thing


kobo/apps/user_reports/migrations/0001_initial.py
```diff
@@ -0,0 +1,419 @@
...
+        Run the SQL query below in PostgreSQL directly:
+
+        {DROP_MV_SQL}
+
+        """
+    )
+
+
+class Migration(migrations.Migration):
+    atomic = False
```
noliveleger, line 444: Why do we need this line?

kobo/apps/user_reports/migrations/0001_initial.py
```diff
@@ -0,0 +1,419 @@
...
+
+    dependencies = [
+        ('trackers', '0005_remove_year_and_month'),
+        ('mfa', '0004_alter_mfamethod_date_created_and_more'),
+        ('kpi', '0069_alter_assetversion_reversion_version'),
+    ]
+
+    if settings.SKIP_HEAVY_MIGRATIONS or not settings.STRIPE_ENABLED:
```
noliveleger: Why do we need this condition: `or not settings.STRIPE_ENABLED`?

kpi/models/user_reports.py
```diff
@@ -0,0 +1,78 @@
...
+class BillingAndUsageSnapshot(models.Model):
+    """
+    A snapshot table for storing precomputed organization billing and usage data.
+
+    Why this table exists:
+    1. Maintaining billing period calculations directly inside the materialized view
+       would make it too complex and hard to manage.
+    2. Usage data such as total submissions, current period submissions, and storage
+       resides in the `kobocat` db, while the materialized view lives in the `kpi`
+       db. Joining across databases for 1.7M+ users would be inefficient.
```
noliveleger: > Joining across databases for 1.7M+ users would be inefficient.

AFAIK, it's not possible natively anyway. We need an extension to do that.

kpi/paginators.py
```diff
@@ -151,6 +151,14 @@ def get_count(self, queryset):
...
 
 
+class LimitStartPagination(LimitOffsetPagination):
+    """
+    A pagination class that supports '?limit=<count>&start=<offset>' query parameters.
+    """
+
+    offset_query_param = 'start'
+
+
```
noliveleger: is it really needed? we can tell the product team to use `offset` instead of `start` to avoid having two almost identical class.

kpi/serializers/v2/user_reports.py
```diff
@@ -0,0 +1,122 @@
+from typing import Any, Dict
```
noliveleger: `Dict` is deprecated. Please use built-in `dict` instead.

kpi/tasks.py
```diff
@@ -1,22 +1,30 @@
 import time
+import uuid
+from typing import List
```
noliveleger: `List` is deprecated. Please use `list`

kpi/utils/usage_calculator.py
```diff
@@ -278,3 +278,78 @@ def _get_cache_hash(self):
             return f'user-{self.user.id}'
         else:
             return f'organization-{self.organization.id}'
+
+
+class BillingAndUsageCalculator:
```
noliveleger: Please organize the file so that classes and methods are listed alphabetically (e.g. BillingAndUsageCalculator >  ServiceUsageCalculator, calculate_usage_batch  > get_effective_user_id,  just touch the code you added).

kpi/utils/usage_calculator.py
```diff
@@ -278,3 +278,78 @@ def _get_cache_hash(self):
...
+    def get_effective_user_id(self, organization: Organization) -> int:
+        return organization.owner.organization_user.user.pk
```
noliveleger: This could be simplified with `organization.owner_user_object.pk` and use the cache if it is called more than once.

kpi/tasks.py
```diff
@@ -140,3 +148,137 @@ def remove_old_versions():
...
+            .filter(~Q(last_snapshot_run_id=run_id))
+            .values_list('pk', flat=True)[:DELETE_BATCH]
+        )
+        if not stale_ids:
+            break
+        BillingAndUsageSnapshot.objects.filter(pk__in=stale_ids).delete()
+
+    # Refresh materialized view
+    with connection.cursor() as cursor:
+        cursor.execute('REFRESH MATERIALIZED VIEW CONCURRENTLY user_reports_mv;')
```
noliveleger: it seems that it does not work with `CONCURRENTLY` 

> ERROR:  cannot refresh materialized view "public.user_reports_mv" concurrently
> HINT:  Create a unique index with no WHERE clause on one or more columns of the materialized view.

but running without concurrenty will lock the view for ~ a 1 minute on production.

kobo/apps/user_reports/models.py
```diff
@@ -39,6 +47,29 @@ def __str__(self):
         return f'BillingAndUsageSnapshot(org={self.organization_id})'
 
 
+class BillingAndUsageSnapshotRun(models.Model):
```
noliveleger: Please use the abstract timestamp model you created a while ago. (please do the same for each model).

No need to have `started_at`, `last_heartbeat_at` and `expires_at`. 
All the logic could be done with `date_created` and `date_modified` (i.e., update `date_modified` instead `last_heartbeat_at`, set `date_created` instead of `started_at` and update your logic to calculate your expiry as per our discussion).

kobo/apps/user_reports/tasks.py
```diff
@@ -31,8 +42,43 @@ def keyset_chunk_org_ids(chunk_size):
...
+    with transaction.atomic():
+        qs = BillingAndUsageSnapshotRun.objects.select_for_update(
+            skip_locked=True
+        ).filter(
+            status=BillingAndUsageSnapshotStatus.RUNNING
+        )
+
+        # Only consider non-expired runs
+        qs = qs.filter(
+            Q(expires_at__isnull=True) | Q(expires_at__gt=now)
```
noliveleger: As per our discussion, let's use `date_created__lte=now - timedelta(hours=RUN_EXPIRY_HOURS)` instead.

kobo/apps/user_reports/tasks.py
noliveleger, line 1: Please move helpers, utilities to their own files. Leave `tasks.py` as simple as possible.  
Try to split "refresh_user_report_snapshots" in several pieces. It's pretty long.

kobo/apps/user_reports/tasks.py
```diff
@@ -31,8 +42,43 @@ def keyset_chunk_org_ids(chunk_size):
...
+            started_at=now,
+            last_heartbeat_at=now,
+            last_processed_org_id=None,
+            expires_at=expires_at,
+        )
+        return run, True
+
+
+@celery_app.task(soft_time_limit=60 * 60 * 1, time_limit=60 * 60 * 2)
+def refresh_user_report_snapshots(batch_size: int = BATCH_SIZE):
```
noliveleger: Increase the soft_limit and hard limit to something really bigger (we know that the task will probably run for 2 hours to complete its job). 

```
@celery_app.task(
    queue='kpi_low_priority_queue',
    soft_time_limit=settings.CELERY_LONG_RUNNING_TASK_SOFT_TIME_LIMIT,
    time_limit=settings.CELERY_LONG_RUNNING_TASK_TIME_LIMIT,
)
```

kobo/apps/user_reports/tasks.py
```diff
@@ -49,103 +95,125 @@ def refresh_user_report_snapshots(batch_size: int = 500):
...
-                    ),
-                    billing_period_start=data.get('billing_period_start'),
-                    billing_period_end=data.get('billing_period_end'),
-                    snapshot_created_at=now,
-                    last_snapshot_run_id=run_id,
+    # Claim or create a run
+    run, created = claim_or_create_run()
+    run_id = run.run_id
+    last_processed_org_id = run.last_processed_org_id
+    try:
```
noliveleger: I think the workflow is not really what I had in mind. 
First thing, sorry to have misled you but let a "run"  run through all organizations whatever it takes 2 hours or 3 days. We won't run a new batch from scratch if the last one was still in progress. 

We know that the celery scheduler will trigger the task every 30 minutes.

The idea is to do something like that: 

```mermaid
flowchart TD
  A["Task fires every 30 min"] --> B{"Acquire Redis lock (non-blocking, TTL >= hard_limit)"}
  B -- no --> C["Exit early (another worker is running)"]
  B -- yes --> D{"Any run with status == IN_PROGRESS?"}
  D -- yes --> E["Load existing run (cursor = last_processed_org_id)"]
  D -- no --> F["Create new run (status=IN_PROGRESS, cursor=0)"]
  E --> L
  F --> L

  %% Loop section
  L{"More chunks to process?"}
  L -- no --> H{"All work done (before hard limit?)"}
  L -- yes --> G["Process next chunk"]
  G --> P["Save last_processed_org_id (update cursor)"]
  P --> T{"Exceeded hard limit?"}
  T -- yes --> J["Exit; lock TTL will expire (resume next run)"]
  T -- no --> L

  %% Completion section
  H -- yes --> I["Mark run DONE, release lock, exit"]
  H -- no --> J
```

kobo/apps/user_reports/tasks.py
```diff
@@ -49,103 +95,125 @@ def refresh_user_report_snapshots(batch_size: int = 500):
...
+        for id_chunk in keyset_chunk_org_ids_after(last_processed_org_id, batch_size):
+            if not id_chunk:
+                continue
+
+            orgs = list(Organization.objects.filter(id__in=id_chunk).order_by('id'))
```
noliveleger: I would like to avoid to cast orgs to a list (even if `calculate_usage_batch` signature expected a list). We should change it to support an iterator (like Queryset, list or generator). 

Moreover `keyset_chunk_org_ids_after` seems a little bit complicated. Please have a look at what [we already do for long running migrations](https://github.com/kobotoolbox/kpi/blob/1e1434cbb78c0299d041006b6ae08978f9f4c273/kobo/apps/long_running_migrations/jobs/0009_backfill_attachment_model.py#L70-L90)

kobo/apps/user_reports/tasks.py
```diff
@@ -49,103 +95,125 @@ def refresh_user_report_snapshots(batch_size: int = 500):
...
+
+            # Compute billing and usage in batch
+            billing_map = get_current_billing_period_dates_by_org(orgs)
+            usage_map = calc.calculate_usage_batch(orgs, billing_map)
+
+            # Fetch existing snapshot rows for this chunk
+            existing_qs = BillingAndUsageSnapshot.objects.filter(
+                organization_id__in=[str(i) for i in id_chunk]
+            )
```
noliveleger: If `orgs` is a queryset, we can avoid to create a list comprehension because Django is smart enough to use the correct field for us with `BillingAndUsageSnapshot.objects.filter(organization_id__in=orgs)`

kobo/apps/user_reports/tasks.py
```diff
@@ -49,103 +95,125 @@ def refresh_user_report_snapshots(batch_size: int = 500):
...
-                    ),
-                    current_period_submissions=data.get(
-                        'current_period_submissions', 0
-                    ),
-                    billing_period_start=data.get('billing_period_start'),
-                    billing_period_end=data.get('billing_period_end'),
-                    snapshot_created_at=now,
-                    last_snapshot_run_id=run_id,
+    # Claim or create a run
+    run, created = claim_or_create_run()
```
noliveleger: Look at the diagram below. I would prefer to avoid to use "select_from_update" . Since, the redis lock is atomic so no concurrent worker should write to the DB at the same time .
But we can enforce an integrity check at the db level...
```
class BillingAndUsageSnapshotRun(AbstractTimeStampedModel):
    ...
    status = models.CharField(max_length=20)
    last_processed_org_id = models.IntegerField(null=True, blank=True)
    singleton = models.BooleanField(default=True, editable=False)  # always True

    class Meta:
        constraints = [
            models.UniqueConstraint(
                fields=('singleton',),
                condition=Q(status=BillingAndUsageSnapshotStatus.IN_PROGRESS),
                name='uniq_run_in_progress',
            ),
        ]
```

then, we can use get_or_create 

```
obj, created = BillingAndUsageSnapshotRun.objects.get_or_create(
    status=BillingAndUsageSnapshotStatus.IN_PROGRESS, 
)

last_processed_org_id = '' if created else obj.last_processed_org_id
```

kobo/apps/user_reports/models.py
noliveleger: Update: expiry is not required anymore. See [other comment](https://github.com/kobotoolbox/kpi/pull/6243/files#r2404062805)

kobo/apps/user_reports/models.py
```diff
@@ -0,0 +1,110 @@
...
+    storage_bytes_total = models.BigIntegerField(default=0)
+    submission_counts_all_time = models.BigIntegerField(default=0)
+    current_period_submissions = models.BigIntegerField(default=0)
+    billing_period_start = models.DateTimeField(null=True, blank=True)
+    billing_period_end = models.DateTimeField(null=True, blank=True)
+    snapshot_created_at = models.DateTimeField(default=timezone.now)
+    last_snapshot_run_id = models.UUIDField(null=True, blank=True, db_index=True)
+
+    class Meta:
+        managed = False
```
noliveleger: Why don't we let Django manage this model?

kobo/apps/user_reports/models.py
```diff
@@ -0,0 +1,110 @@
...
+    )
+    started_at = models.DateTimeField(default=timezone.now)
+    last_heartbeat_at = models.DateTimeField(default=timezone.now)
+    last_processed_org_id = models.CharField(null=True, blank=True)
+    expires_at = models.DateTimeField(null=True, blank=True)
+    details = models.JSONField(null=True, blank=True)
+
+    class Meta:
+        db_table = 'billing_and_usage_snapshot_run'
+        managed = False
```
noliveleger: Why don't we let Django manage this model?

kobo/apps/user_reports/migrations/0001_initial.py
noliveleger, line 1: I would create a first migration with managed models and 2nd migration with materialized view

noliveleger (changes requested):

kobo/apps/user_reports/tasks.py
```diff
@@ -0,0 +1,102 @@
...
+           - Delete stale snapshot rows (not updated in this run).
+           - Refresh the `user_reports_userreportsmv` materialized view concurrently.
+           - Mark the run as 'completed'.
+    """
+    calc = BillingAndUsageCalculator()
+    cache_key = 'billing_and_usage_snapshot:run_lock'
+    lock_timeout = settings.CELERY_LONG_RUNNING_TASK_TIME_LIMIT + 60
+    lock = cache.lock(cache_key, timeout=lock_timeout)
+    if not lock.acquire(blocking=False, blocking_timeout=0):
+        logging.error('Task is already running')
```
noliveleger: This was for debugging purpose. I'm wondering if we should keep this. It will probably flood Sentry logs for nothing. Maybe we can change it for `logging.info(' Nothing to do, task is already running!)`? 
What do you think?

kobo/apps/user_reports/tests/test_user_reports.py
```diff
@@ -0,0 +1,285 @@
...
+        mock_limits = {
+            self.someuser.organization.id: {f'{UsageType.SUBMISSION}_limit': 1}
+        }
+        with patch(
+            'kobo.apps.user_reports.seralizers.get_organizations_effective_limits',
+            return_value=mock_limits,
+        ):
+            with connection.cursor() as cursor:
+                cursor.execute('REFRESH MATERIALIZED VIEW user_reports_userreportsmv;')
+
```
noliveleger: Can we use an utility function to avoid repeating thse two lines several times?

kobo/apps/user_reports/utils/snapshot_refresh_helpers.py
```diff
@@ -0,0 +1,103 @@
...
+        )
+        if not stale_ids:
+            break
+        BillingAndUsageSnapshot.objects.filter(pk__in=stale_ids).delete()
+
+    # Refresh materialized view
+    with connection.cursor() as cursor:
+        cursor.execute(
+            'REFRESH MATERIALIZED VIEW CONCURRENTLY user_reports_userreportsmv;'
+        )
```
noliveleger: We can probably use the same helper here and in the tests.

kobo/apps/user_reports/migrations/0002_create_user_reports_mv.py
```diff
@@ -0,0 +1,377 @@
...
+                            djstripe_taxrate AS tax
+                            ON sub_tax_rate.taxrate_id = tax.djstripe_id
+                        WHERE
+                            sub_tax_rate.subscription_id::text = sub.id::text
+                    )
+                )
+            ) FILTER (WHERE sub.id IS NOT NULL),
+            '[]'::jsonb
+        )::text AS subscriptions
+    FROM auth_user au
```
noliveleger: We probably want to remove anonymous user from the list (e.g. auth_user.id != settings.ANONYMOUS_ID)

PR #6250, opened 2025-09-15
fix(sso): update primary email when enabling SSO DEV-1005

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Update a user's email if they enable SSO and do not allow them to change it as long as SSO is enabled.

### 📖 Description
When a user connects their account to an SSO provider with an email address, we should update their email address to match the email used by the SSO. Once this is set, the user should not be allowed to change it.

### 💭 Notes
This PR makes the assumption that if a user has a SocialAccount and an EmailAddress, the email address probably came from the SSO and so should not be changed. There is an edge case where this wouldn't be true:
1. ACCOUNT_EMAIL_REQUIRED = False
2. The SSO provider does not use email
3. The user separately created and verified their email

However, this should be extremely rare, and it's very hard to determine if an EmailAddress came from an SSO, so we're choosing to keep the assumption in.

Besides preventing users from updating their emails if they have SSO enabled, this PR also takes care of setting the correct email address for a user if they enable SSO on a pre-existing account. It's somewhat surprising allauth doesn't do this automatically, though it seems like something they considered: https://github.com/pennersr/django-allauth/blob/52e90f442ee4f5c2f15f609d130f4b61397b6f3c/allauth/socialaccount/models.py#L313. In case we get multiple email addresses from the SSO provider, which is possible, we use allauth's `cleanup_email_addresses` to select the correct primary one and set all the other ones to not-primary. Then we manually send the email_confirmed signal to indicate that we have confirmed this new primary email, since by the time the `social_account_added` signal is called, we know the email has been confirmed. The existing handler for the signal takes care of cleaning up the rest of the non-primary emails.

We do this instead of removing all the non-primary addresses just in case we ever change the logic around what to do when we have a new verified email and decide we want to allow >1 email per user.

### 👀 Preview steps
Enable SSO locally.

1. ℹ️ have an account and a project. Use a different email for the account than you will use for SSO.
5. In Account Settings > Security, enable Single Sign-On
6. Connect your account with Kobo Google Apps
7. Return to Account Settings > Security
8. 🔴 [on release] The email address has not changed
9. 🟢 [on PR] The email address has changed to the SSO email
10. Try to change the email
11. 🔴 [on release] You are able to update the email
12. 🟢 [on PR] updating the email results in a 400 (this will not yet be reflected in the UI, but if you reload the page you should see that the email hasn't changed)

### DIFF
--- a/kobo/apps/accounts/signals.py
+++ b/kobo/apps/accounts/signals.py
@@ -1,0 +1,29 @@
+from allauth.account.models import EmailAddress
+from allauth.account.utils import cleanup_email_addresses
+from allauth.socialaccount.models import SocialLogin, SocialAccount
+from allauth.socialaccount.signals import social_account_added, social_account_updated, pre_social_login
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+
+@receiver(social_account_added)
+def update_email(*args, **kwargs):
+    sociallogin = kwargs.get('sociallogin')
+    request = kwargs.get('request')
+    social_email_addresses = sociallogin.email_addresses
+    social_user = sociallogin.user
+    for social_email in social_email_addresses:
+        # the auto-created EmailAddresses don't have the user already attached (?!)
+        social_email.user = social_user
+    existing_email_addresses = list(EmailAddress.objects.filter(user=social_user))
+    # put the social email addresses first so they get marked as primary
+    all_email_addresses = [*social_email_addresses, *existing_email_addresses]
+    emails, primary = cleanup_email_addresses(request, all_email_addresses)
+    primary.set_as_primary()
+    EmailAddress.objects.bulk_create(
+        emails,
+        update_conflicts=True,
+        unique_fields=['email','user_id'],
+        update_fields=['primary']
+    )
+
--- a/kobo/apps/accounts/signals.py
+++ b/kobo/apps/accounts/signals.py
@@ -21,6 +21,11 @@ def update_email(*args, **kwargs):
     # put the social email addresses first so they get marked as primary
     all_email_addresses = [*social_email_addresses, *existing_email_addresses]
     emails, primary = cleanup_email_addresses(request, all_email_addresses)
+
+    # cleanup_email_addresses doesn't actually call set_as_primary on the primary
+    # email so do that now
+    primary.set_as_primary()
+
     # update existing emails to reflect that they are no longer primary
     # and add any new emails from the SocialLogin
     EmailAddress.objects.bulk_create(
--- a/kobo/apps/accounts/signals.py
+++ b/kobo/apps/accounts/signals.py
@@ -1,8 +1,7 @@
 from allauth.account.models import EmailAddress
+from allauth.account.signals import email_confirmed
 from allauth.account.utils import cleanup_email_addresses
-from allauth.socialaccount.models import SocialLogin, SocialAccount
-from allauth.socialaccount.signals import social_account_added, social_account_updated, pre_social_login
-from django.db.models.signals import post_save
+from allauth.socialaccount.signals import social_account_added
 from django.dispatch import receiver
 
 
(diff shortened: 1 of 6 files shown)


kobo/apps/accounts/tests/test_email.py
```diff
@@ -171,3 +174,25 @@ def test_that_non_mmo_user_can_update_email(self):
...
+        # in real life connecting the social account would have made an EmailAddress
+        email_address = baker.make('account.emailaddress', user=user)
+        self.client.force_login(user)
+        data = {'email': 'new@example.com'}
+        res = self.client.post(self.url_list, data, format='json')
+        self.assertEqual(res.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(user.emailaddress_set.count(), 1)
+        user_email = EmailAddress.objects.get(user=user)
+        self.assertEqual(user_email.email, email_address.email)
+        self.assertEqual(user.emailaddress_set.count(), 1)
```
jnm, line 201: At the end of this test, ideally I'd like to
```
user.refresh_from_db()
self.assertEqual(user.email, email_address.email)
```
That currently fails, though:
```
AssertionError: '' != 'OqMiOVCCPQ@example.com'
```
I assume that's because Model Bakery, even when being used to create single objects, does not fire the appropriate signals (?)

If we delegate the responsibility for checking that to `test_social_login_connect_updates_primary_email()`, let's leave a comment here to that effect

kobo/apps/accounts/tests/test_email.py
```diff
@@ -171,3 +174,25 @@ def test_that_non_mmo_user_can_update_email(self):
...
+        baker.make('socialaccount.SocialAccount', user=user)
+        # in real life connecting the social account would have made an EmailAddress
+        email_address = baker.make('account.emailaddress', user=user)
+        self.client.force_login(user)
+        data = {'email': 'new@example.com'}
+        res = self.client.post(self.url_list, data, format='json')
+        self.assertEqual(res.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(user.emailaddress_set.count(), 1)
+        user_email = EmailAddress.objects.get(user=user)
+        self.assertEqual(user_email.email, email_address.email)
```
jnm, line 200: We know (at this moment) that Model Bakery will never pick `new@example.com` randomly, but I'd like to assert that the email address is *not* equal to the one in the POST data as a matter of course and to safeguard against any weird future changes to fixtures etc.

kobo/apps/accounts/tests/test_signals.py
```diff
@@ -0,0 +1,34 @@
...
+            primary=True
+        )
+        social_login = SocialLogin(email_addresses=[new_email], user=self.user)
+        update_email(request=request, sociallogin=social_login)
+        # we should have gotten rid of any old EmailAddresses
+        assert EmailAddress.objects.count() == 1
+        address = EmailAddress.objects.first()
+        assert address.email == 'someuser_sso@example.com'
+        assert address.verified
+        assert address.primary
```
jnm: I'd like to verify that the `User.email` attribute is being updated. I stuck an `assert address.user.email == address.email` and it passed; is this the appropriate place for that?

kobo/apps/accounts/signals.py
```diff
@@ -0,0 +1,44 @@
+from allauth.account.models import EmailAddress
+from allauth.account.signals import email_confirmed
+from allauth.account.utils import cleanup_email_addresses
+from allauth.socialaccount.signals import social_account_added
+from django.dispatch import receiver
+
+
+@receiver(social_account_added)
+def update_email(*args, **kwargs):
```
jnm, line 9: 🫣 Should we report any of these django-allauth behaviors as bugs upstream?

rgraber: I was debating that. They seem to already have some idea of the overall problem that connecting to an existing account doesn't update emails correctly (https://github.com/pennersr/django-allauth/blob/52e90f442ee4f5c2f15f609d130f4b61397b6f3c/allauth/socialaccount/models.py#L312), which is the main problem and the reason we have to do all this.

jnm: Gotaha. Es lo que hay :shrug:

PR #6257, opened 2025-09-16
fix(api): restore default renderers for API `v1` DEV-1006 

noliveleger:
### 📣 Summary
Reinstate API v1’s renderer settings so browsers receive HTML by default instead of XML.

### 📖 Description
This change restores the default renderer configuration for API v1 after the v2 removal of BrowsableAPIRenderer inadvertently affected global content negotiation. Because many browsers send broad Accept headers that include XML, API v1 endpoints were returning XML even when the response payload was JSON. The fix scopes renderer settings per API version: v1 once again includes its original renderer set (with JSON prioritized and Browsable API where intended), while v2 keeps the leaner renderer list. 

### 👀 Preview steps

1. Go to https://kc.domain.tld/api/v1/
2. Navigate to endpoints (users, data, forms etc...)
3. 🔴 [on main] receive the error "error on line 1 at column 1:"
4. 🟢 [on PR] you can see the HTML documentation properly

### DIFF
--- a/kobo/apps/openrosa/apps/api/viewsets/data_viewset.py
+++ b/kobo/apps/openrosa/apps/api/viewsets/data_viewset.py
@@ -8,10 +8,10 @@ from rest_framework import status
 from rest_framework.decorators import action
 from rest_framework.exceptions import ParseError
 from rest_framework.response import Response
-from rest_framework.settings import api_settings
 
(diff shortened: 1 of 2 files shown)


Guitlle (approved):
it works

PR #6265, opened 2025-09-17
fix(account): disallow editing email if linked to social account DEV-1011

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 👀 Preview steps

1. ℹ️ Register an account
2. Go to Account Settings → Security
3. 🟢 Notice that you are able to change email

1. ℹ️ Set up SSO (there is a guide on Notion)
2. ℹ️ Register an account through SSO
3. Go to Account Settings → Security
4. 🔴 [on main] Notice that you are able to change email and it doesn't seem to work correctly
5. 🟢 [on PR] notice that email changing is disabled

### DIFF
--- a/jsapp/js/account/security/email/emailSection.component.tsx
+++ b/jsapp/js/account/security/email/emailSection.component.tsx
@@ -112,6 +112,8 @@ export default function EmailSection() {
   const unverifiedEmail = email.emails.find((userEmail) => !userEmail.verified && !userEmail.primary)
   const isReady = session.isInitialLoadComplete && 'email' in currentAccount
   const userCanChangeEmail = orgQuery.data?.is_mmo ? orgQuery.data.request_user_role !== 'member' : true
+  const isSSO =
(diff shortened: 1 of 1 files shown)


jsapp/js/account/security/email/emailSection.component.tsx
pauloamorimbr, line 97: I don't know how critical is this change, but since it's pretty easy to remove the "disabled" of a button, I'd add some other layers of security, like making an early return here

```suggestion
    if(isSSO) return
    setEmail({
```

pauloamorimbr, line 104: ...same here.

```suggestion
    if(isSSO) return
    const emailPattern = /[^@]+@[^@]+\.[^@]+/
```

pauloamorimbr (approved):
LGTM! 🌟 

I've left 2 comments regarding increased safety in case of this email change limitation is a more "critical" thing. But I believe it's already being blocked in the backend, so I'm not sure it would be needed.

Feel free to dismiss the comments!

PR #6267, opened 2025-09-18
feat(usageLimits): Check all users storage usage to get their ExceededLimitCounter's DEV-916

Guitlle:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
A new long running process is added to backfill exceeded limits counters for storage usage type.

### 💭 Notes
The long running job goes through users in the platform that are organization owners and have not got a flag in their ExtraDetails data with the key 'done_storage_limits_check'. Users with pending ExceededLimitCounter object for the storage usage type are put through the function `kobo.apps.stripe.utils.limit_enforcement.check_exceeded_limit` in order to generate the counter entry if necessary. This may be a very long running process, depending on the amount of users registered in the system.

### 👀 Preview steps
1. Ensure stripe is enabled in your dev environment
2. Restart your kpi container in order to make it run the migrations, this will ensure you get the long running process created
3. Execute the long running migration from the shell. Run it twice to ensure that it does not process the same users twice.
```
In [1]: lrm = LongRunningMigration.objects.filter(name__contains='0011').first()

In [2]: lrm.status
Out[2]: 'created'

In [3]: lrm.execute()
Checking exceeded limits for user_a...
Checking exceeded limits for user_b_b...
Checking exceeded limits for user_b_f...
...
Done. Created 8 counters.

In [4]: lrm.status
Out[4]: LongRunningMigrationStatus.COMPLETED

In [5]: lrm.status = 'created'; lrm.save()

In [6]: lrm.execute()
Done. Created 0 counters.

```

### DIFF
--- a/kobo/apps/long_running_migrations/jobs/0011_backfill_exceeded_limit_counters.py
+++ b/kobo/apps/long_running_migrations/jobs/0011_backfill_exceeded_limit_counters.py
@@ -14,15 +14,18 @@ CHUNK_SIZE = 1000
 
 
 def process_user(user_id, username):
+    print(f'Checking exceeded limits for {username}...')
     counter = ExceededLimitCounter.objects.filter(
         user_id=user_id,
         limit_type=UsageType.STORAGE_BYTES,
     ).first()
+    user = User.objects.get(pk=user_id)
     if counter is None:
-        print(f'Checking exceeded limits for {username}...')
-        user = User.objects.get(pk=user_id)
         counter = check_exceeded_limit(user, UsageType.STORAGE_BYTES)
-        return 1 if counter else 0
+
+    user.extra_details.data['done_storage_limits_check'] = True
+    user.extra_details.save()
+    return 1 if counter is None else 0
 
 
 def get_queryset(from_user_pk: int) -> QuerySet:
@@ -30,8 +33,10 @@ def get_queryset(from_user_pk: int) -> QuerySet:
         User.objects.order_by('pk')
         .filter(
             pk__gt=from_user_pk,
-            # Filter organization admins
-            organizations_organizationuser__is_admin=True,
+            # Filter organization owners only
+            organizations_organizationuser__organizationowner__id__isnull=False,
+            # Filter out users that already went through the exceeded limits check
+            extra_details__data__done_storage_limits_check__isnull=True,
         )
         .values('id', 'username')[:CHUNK_SIZE]
     )
@@ -41,7 +46,7 @@ def get_queryset(from_user_pk: int) -> QuerySet:
 
 def run():
     """
-    Checks exceeded storage limits on all users
+    Checks exceeded storage limits on all users for the STORAGE_BYTES usage type
     """
     if not settings.STRIPE_ENABLED:
         print('Nothing to do because Stripe is disabled.')


kobo/apps/long_running_migrations/jobs/0011_check_storage_exceeded_limits.py
```diff
@@ -0,0 +1,49 @@
...
+
+def run():
+    """
+    Checks exceeded storage limits on all users
+    """
+    if not settings.STRIPE_ENABLED:
+        print(f'Stripe is disabled')
+        return
+    created_counters = 0
+    users = User.objects.all()
```
noliveleger: On big databases like our production servers, I'm afraid that it will be killed because of OOM.
Please look at other long running migrations. Please update them by batch (e.g.: take a look at 0009_backfill_attachment_model). 
Do not use "from more_itertools import chunked", it is really bad on memory.
I'm totally convinced that the process won't run in only one run since it will take more than the Celery limit to process all the users. We need to find a way to avoid to exclude users already processed when the task is restarted.

noliveleger: Looking at the query, I don't think we want to go through all the users list.
Aren't limit only for organization owners?  (standalone account and MMO). We probably want to narrow down the list.

kobo/apps/long_running_migrations/jobs/0011_check_storage_exceeded_limits.py
```diff
@@ -0,0 +1,49 @@
...
+        print(f'Using {executor._max_workers} concurrent workers')
+        futures = [
+            executor.submit(process_user, user.pk, user.username) for user in users
+        ]
+
+    total_instances = 0
+    for future in futures:
+        result = future.result()
+        if type(result) is int:
+            created_counters += result
```
noliveleger: Like I said earlier, we are not in the context of the big EC2 instance, the process will run in a pod which can be killed at any moment if it consumes too much memory and/or CPU. Many other Celery tasks can run in the same pod, so make sure to keep the process as lightweight as possible. IMO, [it's better to run slower and reach the finish line than to go faster and never succeed](https://blogs.transparent.com/french/la-fontaines-fable-le-lievre-et-la-tortue-the-hare-and-the-tortoise/) :-P

kobo/apps/stripe/utils/limit_enforcement.py
```diff
@@ -18,6 +18,9 @@ def check_exceeded_limit(user, usage_type: UsageType, **kwargs):
     the ENDPOINT_CACHE_DURATION
     """
     org = user.organization
+    if org is None:
+        return
```
noliveleger: How can it be possible since [`user.organization`](https://github.com/kobotoolbox/kpi/blob/0fb73a084eaee7a5797594bb85de7932d14cf23e/kobo/apps/kobo_auth/models.py#L64-L66) will create an org if it does not exist? The only case I see, would be for anonymous user but it should not be in the users list anyway.

kobo/apps/stripe/utils/limit_enforcement.py
```diff
@@ -46,6 +49,8 @@ def check_exceeded_limit(user, usage_type: UsageType, **kwargs):
             counter.days += delta.days
             counter.save()
 
+        return counter
```
noliveleger: The next line will be never be reached if we return here.

noliveleger (changes requested):

kobo/apps/long_running_migrations/jobs/0011_check_storage_exceeded_limits.py
noliveleger, line 1: I would keep the same convention for the name of files, i.e. : "0011_back_fill_exceeded_limit_counters"

kobo/apps/long_running_migrations/jobs/0011_backfill_exceeded_limit_counters.py
```diff
@@ -0,0 +1,71 @@
...
+        return 1 if counter else 0
+
+
+def get_queryset(from_user_pk: int) -> QuerySet:
+    users = (
+        User.objects.order_by('pk')
+        .filter(
+            pk__gt=from_user_pk,
+            # Filter organization admins
+            organizations_organizationuser__is_admin=True,
```
noliveleger: Don't we want `owner` instead?

kobo/apps/long_running_migrations/jobs/0011_backfill_exceeded_limit_counters.py
```diff
@@ -0,0 +1,71 @@
...
+    if counter is None:
+        print(f'Checking exceeded limits for {username}...')
+        user = User.objects.get(pk=user_id)
+        counter = check_exceeded_limit(user, UsageType.STORAGE_BYTES)
+        return 1 if counter else 0
+
+
+def get_queryset(from_user_pk: int) -> QuerySet:
+    users = (
+        User.objects.order_by('pk')
```
noliveleger, line 33: Maybe we should use a flag on ExtraDetail to avoid starting from scratch if the task is killed. 
Otherwise, we are always restarting from 0. 
We can reset or remove the flag at the end when they are all processed.

PR #6268, opened 2025-09-18
feat(projectOwnership): store all transfer errors 

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Retain all errors from transferring a project, displaying up to 100 in Django admin.

### 📖 Description
When transferring a project, there may be any number of errors that occur when handling individual files. Store individual errors for individual files instead of just the broad 'Some attachments could not be transferred' for easier debugging. These individual errors will be displayed in Django admin.

Adds a trivial migration (new table). Table may become very large if there are lots of transfers with lots of errors.

### 💭 Notes
Creates a new model for storing all errors from a transfer since if we are transferring a project with a lot of attachments and they all fail we could end up with thousands of errors. We do want to track them all, but not in one long string field, so we add the TransferStatusError object. If there are more than a hundred errors, we don't want to overwhelm the display in Django admin, so we just show the first 100 and it will be up to a developer to look for the rest.

Additional updates:

Make the task to move media files more like the task to move attachments by recording individual errors.

Updates the process to move files to reraise errors if requested instead of just logging and swallowing them. This allows us to have more specific error logs rather than just "Failed to move attachment."

Updates the save method on attachments to only update the user_id if `update_fields=['user_id']` rather than going through all the recalculation of hash and size. This prevents us from having duplicate errors from trying to move and then save a busted file.

Record an error if we're trying to update the status of a TransferStatus from 'success' to something else. This should never happen but we have reason to believe there is a race condition that is making it happen.

### 👀 Preview steps
Make sure to restart the worker container after switching branches!

1. ℹ️ have an MMO, an account not in the MMO, and a project with attachments
2. Add several submissions with attachments to the project
3. In a kpi shell, manually delete the uploaded files using `default_kobocat_storage.delete(<file>)`. 
4. In Django admin, add the account not in the MMO to the MMO. This should start the transfer process.
5. Go to the Invite in Django admin
6. 🟢 [on PR] You should see an error for each attachment and media file that failed to transfer
7. Repeat steps 1-6 using a project with media files (if you try to do both at once odds are one will fail and stop the other from running at all). You may need to use `default_storage` instead of `default_kobocat_storage` to delete.

### DIFF
--- a/kobo/apps/project_ownership/admin.py
+++ b/kobo/apps/project_ownership/admin.py
@@ -97,6 +97,7 @@ class InviteAdmin(admin.ModelAdmin):
         )
 
     def get_transfers(self, obj):
+        date_format = '%Y-%m-%d %H:%M:%S'
         html = '<ul>'
         for transfer in obj.transfers.all():
             html += f'<li>{transfer.asset.name} #{transfer.asset.uid}</li>'
@@ -105,7 +106,7 @@ class InviteAdmin(admin.ModelAdmin):
                 status_type=TransferStatusTypeChoices.GLOBAL
             ):
                 errors = [
-                    f'{error.date_created} - {error.error}'
+                    f'[{error.date_created.strftime(date_format)}] - {error.error}'
                     for error in status.errors.filter(error__isnull=False)
                 ]
                 if status.error:


kpi/fields/file.py
```diff
@@ -10,7 +10,7 @@
 
 class ExtendedFieldFile(FieldFile):
 
-    def move(self, target_folder: str) -> bool:
+    def move(self, target_folder: str, reraise_errors=False) -> bool:
```
noliveleger: even if it's obvious, let's keep the typing annotation for `reraise_errors` too 🙏 .

kobo/apps/project_ownership/utils.py
```diff
@@ -126,10 +129,11 @@ def move_attachments(transfer: 'project_ownership.Transfer'):
                     )
                 else:
```
noliveleger, line 130: If I understand the code correctly, if `reraise_errors` is True, we never enter the `else` condition and stop the loop directly... Meaning we won't move other attachments even it could be done flawlessly.
It would be better to keep going the loop AND change the status when all the attachments are processed. What do you think?

kobo/apps/project_ownership/utils.py
```diff
@@ -169,36 +181,64 @@ def move_media_files(transfer: 'project_ownership.Transfer'):
...
+                            if not kc_obj.data_file.move(
+                                kc_target_folder, reraise_errors=True
+                            ):
+                                success = False
+                                TransferStatus._add_error(
+                                    transfer_status,
+                                    f'Kobocat file {kc_obj.data_file.name}'
+                                    f' (#{media_file.pk})'
+                                    f' could not be moved to {kc_target_folder}',
+                                )
```
noliveleger, line 227: Same comment here. This will never happen right?

noliveleger (changes requested):

kobo/apps/project_ownership/utils.py
rgraber: If reraise errors is True and we throw an error we should end up in the Exception handler and we'll continue the loop. The else clause is for if `move` returns False but doesn't raise an error, which would be weird, but is theoretically possible

kobo/apps/project_ownership/utils.py
rgraber: Theoretically `move` could still return False even if it doesn't raise an error. This way we won't miss things if the implementation of `move` changes

kobo/apps/project_ownership/utils.py
noliveleger: Oh I got fooled by the indentation. Nevermind about the loop. You're right! 🤦 

For the `else` clause, it can enter it only iif it's not S3. I missed that. With S3, the "reraise" will supersede the return. 

Thank you for the explanation.

noliveleger (approved):
CI has not finished yet but tests are passing locally.

PR #6269, opened 2025-09-18
fix(assetUsage): migrate asset usage table

duvld:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Use the new table component to display asset usage data. Fixes old pagination issues

### 💭 Notes
Modeled after how access logs uses the universal table. Universal table does not support sorting AFAIK, so there is a UX regression. The sorting didn't' really work before though, as the table wasn't displaying accurate information

### 👀 Preview steps

1. ℹ️ have an account and many projects
2. submit some data
3. navigate to account/usage > per project total tab
4. 🔴 [on main] notice that pagination doesn't work
5. 🟢 [on PR] notice that pagination does work
6. 🟢 [on PR] notice that table displays the correct data with no regressions (aside from the ability to sort)
7. ℹ️ have a MMO with different member's permissions
8. 🟢 notice that the table has no regressions related to MMOs


magicznyleszek:
@duvld you forgot to add task id in PR title :)

jsapp/js/account/usage/assetUsage.api.ts
```diff
@@ -47,8 +42,11 @@ export async function getOrgAssetUsage(
     params.set('ordering', orderingPrefix + fieldDefinition.apiOrderingName)
   }
 
-  return fetchGet<AssetUsage>(`${apiUrl}?${params}`, {
-    includeHeaders: true,
-    errorMessageDisplay: t('There was an error fetching asset usage data.'),
-  })
+  return {
+    status: 200 as const,
```
jamesrkiger, line 46: What's the story here? Pre-defining the response status seems strange.

jamesrkiger (changes requested):
Left a couple comments, but it's looking good so far.

jsapp/js/account/usage/usageProjectBreakdown.tsx
```diff
@@ -1,103 +1,43 @@
...
-    }
-  }
-
-  const isActiveBack = currentPage > 1
-  const isActiveForward = currentPage < Math.ceil(Number.parseInt(projectData.count) / USAGE_ASSETS_PER_PAGE)
+  const queryResult = useQuery({
+    queryKey: [QueryKeys.assetUsage, pagination.limit, pagination.offset, orgQuery.data, orgQuery.data?.id, order],
+    queryFn: () => getOrgAssetUsage(pagination.limit, pagination.offset, orgQuery.data ? orgQuery.data.id : '', order),
+    placeholderData: keepPreviousData,
+  })
```
Akuukis: good to migrate to custom react-queries :ok_hand: But let's migrate straight Orval's react-query! 
See `useOrganizationsServiceUsageRetrieve`, or `useOrganizationsServiceUsageSummary` that just got merged into `main`. Afterwards the whole of `assetUsage.api.ts` could be deleted.

P.S. #6307 also uses Orval's react-query response in UniversalTable

jsapp/js/account/usage/assetUsage.api.ts
Akuukis: see my other comment

jsapp/js/account/usage/usageProjectBreakdown.tsx
```diff
@@ -1,101 +1,68 @@
...
+        notify(t('There was an error getting the list.'), 'error') // TODO: update message in backend (DEV-1218).
+        return false
+      },
+    },
+  })
 
   const usageName: ProjectFieldDefinition = {
     name: 'name',
-    label: t('##count## Projects').replace('##count##', projectData.count),
+    label: getUsageNameLabel(),
```
Akuukis: why bother a separate function for a oneliner that's used once? Also, here's the correct way to check whether response succeeded or errored.

```suggestion
    label: queryResult.data.status === 200 ? t('##count## Projects').replace('##count##', queryResult.data.data.count.toString()) : t('Projects')
```

jsapp/js/account/usage/usageProjectBreakdown.tsx
```diff
@@ -1,101 +1,68 @@
...
-    }
-    const endRange = Math.min(currentPage * USAGE_ASSETS_PER_PAGE, totalProjects)
-    return `${startRange}-${endRange} of ${totalProjects}`
-  }
-
-  const handleClick = async (event: React.MouseEvent<HTMLButtonElement>, buttonType: ButtonType): Promise<void> => {
-    event.preventDefault()
+  function getQueryParams() {
+    // HACK FIX: a bit of a roundabout way to incorporate what the backend expects without diving too deep into changing
+    // existing types
```
Akuukis: This comment is not helpful as it answers nothing actually. It logically begs the questions "why don't you dive deep into fixing existing types then" and "which backend expectations" and "which existing types", all of which are not answered.

How about "TODO: align <SortableProjectColumnHeader /> props with backend pagination params to simplify away this helper."?

jsapp/js/account/usage/usageProjectBreakdown.tsx
```diff
@@ -1,101 +1,68 @@
...
+      placeholderData: keepPreviousData,
+      // We might want to improve this in future, for now let's not retry
+      retry: false,
+      // The `refetchOnWindowFocus` option is `true` by default, I'm setting it
+      // here so we don't forget about it.
+      refetchOnWindowFocus: true,
+      throwOnError: () => {
+        notify(t('There was an error getting the list.'), 'error') // TODO: update message in backend (DEV-1218).
+        return false
+      },
```
Akuukis, line 54: are you sure you need any of these options here? I see they are verbatim copied from the `useOrganizationsMembersList`. By default use defaults which is no options.

jsapp/js/account/usage/usageProjectBreakdown.tsx
```diff
@@ -112,33 +79,91 @@ const ProjectBreakdown = () => {
     setOrder(newOrder)
   }
```
Akuukis, line 78: why bother an identity function, just use `setOrder` directly instead of `updateOrder`?

jsapp/js/projects/projectsTable/sortableProjectColumnHeader.tsx
```diff
@@ -17,6 +17,11 @@ interface SortableProjectColumnHeaderProps {
   order: ProjectsTableOrder
   onChangeOrderRequested: (order: ProjectsTableOrder) => void
   onHideFieldRequested?: (fieldName: ProjectFieldName) => void
+  /**
+   * Note: There is only one use case so far where we don't want this to be resizeable, so we set the default to show
+   * the resizer as to not change any other code. See DEV-1255.
+   */
+  notResizeable?: boolean
```
Akuukis: 1. positive naming is better than negative naming, please consider names like `fixedWidth` instead. AFAIK all flags are false by default, so lack of `fixedWidth` should be self-explanatory.
2. this comment is not answering anything. Which usecase? Why? Won't there be another usecase in future? I looked into it, consider renaming it to something like "For compatibility with react-table set `fixedWidth` because X otherwise Y".

Akuukis (changes requested):
see the failing CI, code review inline, not tested.

jsapp/js/account/usage/usageProjectBreakdown.tsx
```diff
@@ -112,33 +79,91 @@ const ProjectBreakdown = () => {
...
+      size: 100,
+      cellFormatter: (data: CustomAssetUsage) =>
+        convertSecondsToMinutes(data.nlp_usage_current_period.total_nlp_asr_seconds).toLocaleString(),
+    },
+    {
+      key: 'translation_characters',
+      label: t('Translation characters'),
+      size: 100,
+      cellFormatter: (data: CustomAssetUsage) =>
+        convertSecondsToMinutes(data.nlp_usage_current_period.total_nlp_mt_characters).toLocaleString(),
```
jamesrkiger: Copy paste error? No need to convert character count.

jamesrkiger (changes requested):
Left one comment that needs a quick fix

jsapp/js/account/usage/usageProjectBreakdown.tsx
duvld: This still needs a check for undefined on `queryResult.data`, changed it to this:
```typescript
label: (queryResult.data && queryResult.data.status === 200) ? t('##count## Projects').replace('##count##', queryResult.data.data.count.toString()) : t('Projects'),
```

jsapp/js/account/usage/usageProjectBreakdown.tsx
duvld: Yes that sounds a lot better, thanks

jsapp/js/account/usage/usageProjectBreakdown.tsx
duvld: Removed other options except for the error throwing

jsapp/js/account/usage/usageProjectBreakdown.tsx
duvld: I don't know, I didn't write this code. I don't want to change old code if I'm editing an existing file so I usually just leave the logic as is. Perhaps it was originally supposed to hold more logic here, but Ruth never got to it. Do you think we should simplify this?

jsapp/js/projects/projectsTable/sortableProjectColumnHeader.tsx
duvld: That's clearer--thank you

PR #6273, opened 2025-09-22
fix(projectOwnershipTransfer): ensure new owners have full data visibility regardless of prior partial permissions DEV-1022

noliveleger:
### 📣 Summary
New owners now automatically gain full access to all submission data on their assets, overriding any previously assigned partial permissions.

### 📖 Description
When a user becomes the owner of an asset, the system now guarantees full visibility and edit rights over all submission data for that asset. Any earlier partial permissions are cleared for the new owner, and full submission permissions are granted alongside existing owner privileges. 

### 👀 Preview steps

2. Assign partial permission to one user (UserB) (e.g. : view only their submissions)
3. Add some submissions with the owner account
4. Transfer the project to UserB
5. 🔴 [on main] notice that UserB is the owner but cannot see the data
6. 🟢 [on PR] notice that UserB is the owner but can see the data

Repeat the steps with the PR branch to ensure the result is the same.


RuthTurk (approved):
LGTM

PR #6275, opened 2025-09-23
fix(attachmentTrash): fix cleanup of orphaned `AttachmentTrash` and `PeriodicTasks` DEV-971

rajpatel24:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Fixed an issue where deleted submissions left behind orphaned attachment trash records and their associated periodic tasks.

### 📖 Description
Previously, if a submission was deleted after its attachments were moved to the trash, the Admin still displayed those attachments in the trash bin. Opening them caused errors, and their cleanup periodic tasks were never removed. This update ensures that attachment trash entries and their periodic tasks are correctly removed when the related submission or attachment is deleted. A new management command also cleans up any leftover broken records from the past.

### 👀 Preview steps

1. ℹ️ have an account with a project and a few submissions that include attachments.
2. delete an attachment from a submission.
3. check the attachment trash bin in Admin. Trash entry and periodic task should be created.
4. now delete the submission.
5. 🔴 [on main] the attachment trash object is still visible in Admin, and opening it raises an error.
7. 🟢 [on PR] the trash object and its associated periodic task are correctly removed.


kobo/apps/openrosa/apps/logger/signals.py
```diff
@@ -53,6 +54,20 @@ def pre_delete_attachment(instance, **kwargs):
...
+        with transaction.atomic():
+            att_trash = AttachmentTrash.objects.get(attachment_id=attachment.pk)
+            if att_trash:
+                periodic_task = att_trash.periodic_task
+                att_trash.delete()
+
+                if periodic_task:
+                    periodic_task.delete()
+    except Exception as e:
+        logging.error('Failed to delete attachment trash: ' + str(e), exc_info=True)
```
noliveleger: I don't want to log error each time we try to delete an attachment (like when we delete a submission, related attachments are deleted in CASCADE). It's possible that an attachment is deleted manually without being in the trash before.

kobo/apps/openrosa/apps/logger/signals.py
```diff
@@ -53,6 +54,20 @@ def pre_delete_attachment(instance, **kwargs):
     if only_update_counters or not (media_file_name := str(attachment.media_file)):
         return
 
+    # Clean-up AttachmentTrash and related PeriodicTask
+    try:
+        AttachmentTrash = apps.get_model('trash_bin', 'AttachmentTrash')
+        with transaction.atomic():
+            att_trash = AttachmentTrash.objects.get(attachment_id=attachment.pk)
```
noliveleger:
```python
att_trash = AttachmentTrash.objects.get(attachment_id=attachment.pk)
```

This is going to fail each time it does not exist (see other comment on submission deletion and CASCADE). 

IMO, it would be better to have something like 

```python 
AttachmentTrash = apps.get_model('trash_bin', 'AttachmentTrash')
with transaction.atomic():
    try
        att_trash = AttachmentTrash.objects.get(attachment_id=attachment.pk)
    except AttachmentTrash.DoesNotExist:
        pass
    else:
        periodic_task = att_trash.periodic_task
        att_trash.delete()
        if periodic_task:
            periodic_task.delete()
```

rajpatel24: Ah, how did I miss that!! Updated it now. Thanks :)

kobo/apps/openrosa/apps/logger/signals.py
```diff
@@ -55,18 +55,17 @@ def pre_delete_attachment(instance, **kwargs):
         return
 
     # Clean-up AttachmentTrash and related PeriodicTask
-    try:
-        AttachmentTrash = apps.get_model('trash_bin', 'AttachmentTrash')
-        with transaction.atomic():
+    AttachmentTrash = apps.get_model('trash_bin', 'AttachmentTrash')
+    with transaction.atomic():
```
noliveleger: Let's move the transaction.atomic() to wrap only write queries (i.e.: 65 above "periodic_task = att_trash.periodic_task)

kobo/apps/trash_bin/tests/test_utils.py
```diff
@@ -839,6 +839,21 @@ def test_management_command_cleans_up_orphaned_attachment_trash(self):
...
+        """
+        # Ensure no trash exists
+        self.assertFalse(
+            AttachmentTrash.objects.filter(attachment_id=self.attachment.pk).exists()
+        )
+
+        self.attachment.delete()
+
+        # Attachment must be gone
+        self.assertFalse(Attachment.objects.filter(pk=self.attachment.pk).exists())
```
noliveleger: We should use `all_objects` here to be sure it's not hidden by any manager.

noliveleger (changes requested):

kobo/apps/trash_bin/management/commands/cleanup_orphan_attachment_trash.py
```diff
@@ -0,0 +1,52 @@
...
+        att_trash_qs = AttachmentTrash.objects.select_related(
+            'periodic_task'
+        ).order_by('pk')
+
+        deleted_count = 0
+        for trash in att_trash_qs.iterator():
+            try:
+                attachment_exists = Attachment.all_objects.filter(
+                    pk=trash.attachment_id
+                ).exists()
```
noliveleger: I think I would do it in two steps to filter down directly the real orphans. 
What about getting all the attachment ids, compare them with `Attachment.all_objects.values_list('pk').filter(delete_status='pending')` , get only the missing ones then narrow down `att_trash_qs` based those missing attachment ids? 

`Attachment.all_objects.values_list('pk').filter(delete_status='pending')` must be quite slow on production though. You can test it and I'm not sure if the filter is the real one, please adapt it.

noliveleger:
@rajpatel24 , LGTM except schemas are not up-to-date. You can merge it as soon as the CI passes.

noliveleger (approved):
@rajpatel24 , LGTM except schemas are not up-to-date. You can merge it as soon as the CI passes.

PR #6284, opened 2025-09-24
fix(api): type /organizations/:id/service_usage response DEV-1062

Guitlle:
### 🗒️ Checklist

1. [x] run linter locally

### 💭 Notes
This PR changes the props for the service usage response balances data, making them nullable and required. 
The exceeded prop was changed to bool.
The service usage balance data was unified to avoid duplication.

### 👀 Preview steps
Check the files generated by Orval: 
* `jsapp/js/api/models/serviceUsageBalances.ts`
* `jsapp/js/api/models/serviceUsageResponse.ts`
* `jsapp/js/api/models/serviceUsageBalanceData.ts`
* `jsapp/js/api/models/nullableServiceUsageBalanceData.ts` 
* `jsapp/js/api/models/organizationServiceUsageResponse.ts`


jsapp/js/api/models/organizationServiceUsageResponseBalances.ts
```diff
@@ -8,8 +8,12 @@
...
-  mt_characters?: ServiceUsageBalanceData
+  /** @nullable */
+  submission: ServiceUsageBalanceData
+  /** @nullable */
+  storage_bytes: ServiceUsageBalanceData
+  /** @nullable */
+  asr_seconds: ServiceUsageBalanceData
+  /** @nullable */
+  mt_characters: ServiceUsageBalanceData
 }
```
Akuukis: Properties are required now, so far so good, but that `@nullable` comment doesn't matter much. This should be generated like this instead:
```ts
export type OrganizationServiceUsageResponseBalances = {
  submission: ServiceUsageBalanceData | null
  storage_bytes: ServiceUsageBalanceData | null
  asr_seconds: ServiceUsageBalanceData | null
  mt_characters: ServiceUsageBalanceData | null
}
```

jsapp/js/api/models/serviceUsageResponseBalances.ts
```diff
@@ -8,8 +8,12 @@
 import type { ServiceUsageBalanceData } from './serviceUsageBalanceData'
 
 export type ServiceUsageResponseBalances = {
```
Akuukis: `ServiceUsageResponseBalances` seems identical to `OrganizationServiceUsageResponseBalances`, why not have one interface instead and reuse it in both places?

Guitlle: Ok

jsapp/js/api/models/organizationServiceUsageResponseBalances.ts
Guitlle: Done, though Orval won't generate inline `| null` definitions, it needs an intermediate type

Akuukis (changes requested):
code LGTM and tested :tada: 

Let's get the title and description in order now:
- let's have a shorter title, how about `fix(api): type /organizations/:id/service_usage response DEV-1062`?
- let's change Summary section for Notes section in description, I believe that users don't care to know.
- please review Notes section, there are more changes now.
- please review the Preview steps, for example the mentioned files do not exist anymore.

PR #6262, opened 2025-09-17
feat(billing): enable LLM usage tracking DEV-728

jamesrkiger:
### 🗒️ Checklist

1. [x] run linter locally
2. [x] update developer docs (API, README, inline, etc.), if any
3. [x] for user-facing doc changes create a Zulip thread at #Kobo support docs, if any
4. [x] draft PR with a title `<type>(<scope>)<!>: <title> DEV-1234`
5. [x] assign yourself, tag PR: at least `Front end` and/or `Back end` or `workflow`
6. [x] fill in the template below and delete template comments
7. [x] review thyself: read the diff and repro the preview as written
8. [x] open PR & confirm that CI passes & request reviewers, if needed
9. [ ] delete this section before merging

### 📣 Summary
Enables tracking of automated QAn (LLM) usage via `NLPUsageCounter` in the `trackers` app, adds `llm_requests` to service usage serializer and updates API/docs accordingly.

### 💭 Notes
This is preliminary work for enabling the tracking of usage and limits for llm requests on the backend. As far as possible, I have avoided changing code specifically related to Stripe, which I will adjust in a subsequent task. I wanted to break up this work across at least a couple of PRs to prevent a massive, unreadable diff. So there are a few temporary code changes aimed at making these changes work without changing stripe code paths.

**frontend changes** In `jsapp/js/account/usage/useOrganizationsServiceUsageSummary.ts`, there was a type error resulting from the fact that the service usage `balances` field can now have an llm_requests key that we don't account for on the frontend. I adjusted this helper to just ignore that field for now and left a todo comment to adjust it later when we are ready to support this usage type on the frontend.

### 👀 Preview steps
Since the automated qualitative analysis feature has not been developed yet, there isn't much to test here beyond what is covered in the unit tests. However, you can confirm the changes by accessing the app and looking at the ServiceUsage response in browser. You will see new entries for `llm_requests` in `total_nlp_usage` and `balances` ,

PR #6291, opened 2025-09-25
feat(billing): handle limit aggregation for llm requests DEV-1031

jamesrkiger:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Enables aggregation of automated QA billing limits and the inclusion of these limits in service usage data returned by the API.

### 💭 Notes
We still don't actually have an automated QA feature yet, so this code doesn't do much on its own. But when the feature is built and creates usage counters (see #6262), these will be compared against "llm_requests_limit" metadata in our Stripe plans and included in the calculation of limit balances.

### 👀 Preview steps
1. On a Stripe-enabled instance, create a new user on a community plan.
2. In django-admin add "llm_requests_limit" to the community plan product metadata
3. Create a project and, using the django shell, add an NLP usage counter for LLM requests on the new project.
4. Check the API response at `api/v2/organizations/{org_id}/service_usage/` and see that the usage is correctly compared to the new plan limit number in the `balances` field.


magicznyleszek:
@jamesrkiger the title shouldn't have square brackets :)

PR #6292, opened 2025-09-25
fix(api):  type OrganizationResponse DEV-1063

Guitlle:
### 🗒️ Checklist

1. [x] run linter locally

### 💭 Notes
Create OrganizationResponse serializer to avoid properties mismatch in Orval output models

### 👀 Preview steps
See the generated file at `jsapp/js/api/models/organizationResponse.ts` and check that the `generate_api.sh` script does not throw any warnings.

### DIFF
--- a/jsapp/js/api/models/patchedOrganizationPatchPayload.ts
+++ b/jsapp/js/api/models/patchedOrganizationPatchPayload.ts
@@ -5,9 +5,12 @@
  * Powerful and intuitive data collection tools to make an impact
  * OpenAPI spec version: 2.0.0 (api_v2)
  */
+import type { OrganizationTypeEnum } from './organizationTypeEnum'
 
 export interface PatchedOrganizationPatchPayload {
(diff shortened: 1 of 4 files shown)


jsapp/js/api/models/organizationResponse.ts
```diff
@@ -0,0 +1,28 @@
...
+  readonly created: string
+  readonly modified: string
+  readonly is_owner: boolean
+  readonly is_mmo: boolean
+  readonly request_user_role: MemberRoleEnum
+  readonly members: string
+  readonly assets: string
+  readonly service_usage: string
+  readonly asset_usage: string
+}
```
Akuukis, line 28: this solves website and organization_type props to be required, but all the props are still inconsistently readonly or not.

Guitlle: Since this is the response, it should all be read only

PR #6296, opened 2025-09-29
refactor(frontend): add typed recordKeys and friends DEV-1077

Akuukis:
### 💭 Notes

Add and use utilities for Records. Javascript objects may hold any string/number/symbol properties, and so `Object.keys` types by default are understandably protective. Let's call some objects records, where properties are known to have only properties we expect it to have.

*Mostly* type-only changes, except for few places where swapping either of three for another would simplify the code.

P.S. See `ANALYSIS_QUESTION_TYPES`, this refactor revealed a bug that I solved along the way.

### DIFF
--- a/jsapp/js/utils.ts
+++ b/jsapp/js/utils.ts
@@ -26,21 +26,19 @@ type KeyValue<T extends object, K extends keyof T = keyof T> = [K, T[K]]
  *
  * P.S. Prefer mapped types (`Record<K,V>`) over index signatures (`{[k:K]: V}`) where possible in your types.
  */
-export const recordEntries = <T extends object>(o: T) => Object.entries(o) as KeyValue<T>[];
+export const recordEntries = <T extends object>(o: T) => Object.entries(o) as KeyValue<T>[]
 /**
  * A strongly-typed version of `Object.keys()`, always use this instead.
  *
  * P.S. Prefer mapped types (`Record<K,V>`) over index signatures (`{[k:K]: V}`) where possible in your types.
  */
-export const recordKeys = <T extends object>(o: T) => Object.keys(o) as (keyof T)[];
+export const recordKeys = <T extends object>(o: T) => Object.keys(o) as (keyof T)[]
 /**
  * A strongly-typed version of `Object.values()`, always use this instead.
  *
  * P.S. Prefer mapped types (`Record<K,V>`) over index signatures (`{[k:K]: V}`) where possible in your types.
  */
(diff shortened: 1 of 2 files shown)


jsapp/js/utils.ts
```diff
@@ -16,6 +16,27 @@ import { toast } from 'react-hot-toast'
...
+/**
+ * A typed `Object.entries`, assuming the object is a simple Record<T, U>.
+ */
+export const recordEntries = <T extends object>(o: T) => Object.entries(o) as KeyValue<T>[]
+/**
+ * A typed `Object.keys`, assuming the object is a simple Record<T, U>.
+ *
+ * P.S. Prefer `Record<K,V>` over `{[k:K]: V}`, note that `recordKeys<Record<string, string>(..): string[]`,
+ * but `recordKeys<{[k: string]: string}>(..): (string | number)[]`.
+ */
```
magicznyleszek, line 34:
```suggestion
/**
 * A typed `Object.keys`, assuming the object is a simple Record<T, U>.
 *
 * P.S. Prefer `Record<K, V>` over `{ [k: K]: V } `, note that first one gives a list of strings, and second list of
 * strings or numbers - so `recordKeys < Record<string, string>(..): string[]`,
 * but `recordKeys<{ [k: string]: string }>(..): (string | number)[]`.
 */
```
What do you think of adding some simple language explanation before that code examples?

jsapp/js/assetUtils.ts
```diff
@@ -669,14 +669,14 @@ export function getAssetProcessingRows(assetUid: string) {
...
-    Object.keys(foundAsset.advanced_submission_schema.properties).forEach((propertyName) => {
+    recordKeys(foundAsset.advanced_submission_schema.properties).forEach((propertyName) => {
       if (foundAsset.advanced_submission_schema?.properties !== undefined) {
         const propertyObj = foundAsset.advanced_submission_schema.properties[propertyName]
         // NOTE: we assume that the properties will hold only a special string
         // "submission" property and one object property for each
         // processing-enabled row.
         if (propertyObj.type === 'object') {
-          rows.push(propertyName)
+          rows.push(propertyName as string)
```
magicznyleszek: I'm trying to understand why is `propertyName` a `string | number` after using `recordKeys` and not just a `string`…

Akuukis: pushed a fix, see [fa026f2](https://github.com/kobotoolbox/kpi/pull/6296/commits/fa026f29d4da21f9a5ef04f24ed5cfbfd7a11fba)

PR #6305, opened 2025-09-29
refactor(reactQuery): useOrganizationsServiceUsageSummary

Akuukis:
### 🗒️ Checklist

1. [ ] run linter locally

### 💭 Notes

Use auto-generated react-query hooks and types by Orval, as well remove the hand-written one.

Furthermore, add `useOrganizationsServiceUsageSummary` hook to handle the "summary" object creation from the API's raw response.

### 👀 Preview steps

1. ℹ️ login and click around
2. 🟢 nothing crashes
3. 🟢 service_usage endpoint is called only once*
   - 🟢 except on entering `http://kf.kobo.local/#/account/organization/usage`, it always fetches it again 
4. Have a project with some submissions, share it for anonymous views


jsapp/js/account/usage/usage.component.tsx
```diff
@@ -54,17 +57,20 @@ export default function Usage() {
...
-    if (!usageData) return ''
-    const startDate = formatDate(usageData.currentPeriodStart)
-    const endDate = formatDate(usageData.currentPeriodEnd)
+    if (usageData?.status !== 200) return ''
+    const startDate = formatDate(usageData.data.currentPeriodStart)
+    const endDate = formatDate(usageData.data.currentPeriodEnd)
     return t('##start_date## to ##end_date##').replace('##start_date##', startDate).replace('##end_date##', endDate)
-  }, [usageData?.currentPeriodStart, usageData?.currentPeriodEnd])
+  }, [
+    (usageData?.data as OrganizationsServiceUsageSummary)?.currentPeriodStart,
```
jamesrkiger: `usageData.data` seems a little redundant. Maybe we should rename `usageData` as  `usageSummary` or `usage`? The same issue comes up in other files below, in case you think it's worth changing.

Akuukis: It comes from the fact that our custom `fetchDataRaw` returns another `{data, status, headers}` object instead of a simple `data` in order to make `headers` available to do things like at the start of `transformOrganizationsService` function. It's the case for all query responses, although headers are used quite rarely.

I'd like to investigate how to simplify this setup, but that's surely out of scope.

jamesrkiger (approved):
lgtm

PR #6315, opened 2025-10-01
fix(api): make MemberListResponse.invite nullable DEV-1093

Guitlle:
### 🗒️ Checklist

1. [x] run linter locally

### 💭 Notes
Make the property `MemberListResponse.invite` optional/nullable

### 👀 Preview steps
Check the generated Orval files: 
- `jsapp/js/api/models/memberListResponse.ts`
- `jsapp/js/api/models/memberListResponseInvite.ts`


jsapp/js/api/models/memberListResponse.ts
```diff
@@ -9,14 +9,15 @@ import type { MemberListResponseInvite } from './memberListResponseInvite'
...
   user: string
   user__username: string
   user__email: string
   user__extra_details__name: string
-  role: MemberRoleEnum
   user__has_mfa_enabled: boolean
   date_joined: string
   user__is_active: boolean
+  /** @nullable */
   invite: MemberListResponseInvite
```
Akuukis, line 22: API returns `{}` atm, it's not the same as `null`. Notably in javascript, `{}` is truthy but `null` is falsey.

Perhaps api should return `null` instead, would that make more sense?

Guitlle: Ok, I did that. I'm waiting for tests to pass and confirm it's all ok

Akuukis (approved):
LGTM & tested

Consider using `fix(api)` instead of `fix(openApi)` to indicate that the actual api response was changed.

PR #6316, opened 2025-10-02
ci(releases): run version even if a test fails DEV-1109

Akuukis:
### 💭 Notes

Always run version, so that failure notification has access to variables.

### 👀 Preview steps

See https://github.com/kobotoolbox/kpi/actions/runs/18171862423


magicznyleszek (approved):
LGTM

PR #6322, opened 2025-10-02
ci(releases): don't trigger on branch creation

Akuukis:
### 🗒️ Checklist

1. [x] run linter locally

### DIFF
--- a/.github/workflows/release-2-stabilize.yml
+++ b/.github/workflows/release-2-stabilize.yml
@@ -2,7 +2,6 @@ name: 'CD: stabilize'
 
 
 on:
-  create:
   push:
     branches: [ 'release/**' ]
 
--- a/.github/workflows/release-2-stabilize.yml
+++ b/.github/workflows/release-2-stabilize.yml
@@ -10,7 +10,6 @@ jobs:
 
 
   changes:
-    if: ${{ startsWith(github.ref, 'refs/heads/release/') }}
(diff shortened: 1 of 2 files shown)


.github/workflows/release-2-stabilize.yml
```diff
@@ -2,7 +2,6 @@ name: 'CD: stabilize'
 
 
 on:
-  create:
```
jnm, line 5: There's also the conditional on the `changes` job:

        if: ${{ startsWith(github.ref, 'refs/heads/release/') }}

The dependency chain of `needs` makes me think that all jobs would have to pass through that—so basically I'm curious:
1. why stuff was running on non-release branches in the first place
2. if we need this `if` in addition to the the `push` pattern `branches: [ 'release/**' ]`

Akuukis: hmm good point. Actually this is a bigger mystery than that, because in the problematic workflow the changes were properly skipped, but `needs.changes.result == 'success'` condition on the `version` job didn't work as expected :thinking:

Akuukis: After checking the GHA history, I think we don't need either `create` nor abovementioned check, because `push` event seems like a superset of `create` event.

Akuukis:
Nothing changed, everything works again today. I will blame it on a hiccup on Github's side that messed up their triggers.

PR #6330, opened 2025-10-03
fix(permissions): prevent granting permissions to inactive users DEV-1122

noliveleger:
### 📣 Summary
Stop assigning any permissions to users whose accounts are inactive.

### 📖 Description
This change enforces an `is_active` check across all permission grant paths (single and bulk). When a user is inactive, permission assignments are rejected and the bad request is returned.

### 👀 Preview steps

2. make another account (UserB) inactive
3. 🔴 [on release branch] notice you can share the your project with UserB
4. 🟢 [on PR] notice that you cannot share it. An error is returned


Guitlle (approved):
Looks good 👍🏾

PR #6331, opened 2025-10-03
chores(openAPI): update HTML template title and text body

noliveleger:
### 📣 Summary
Replace default DRF title and update the top level text


jnm (approved):
Added the change to prevent displaying the seemingly-arbitrary list of endpoints when the root is loaded

PR #6334, opened 2025-10-06
fix(accounts): block self-deletion when the account still owns data DEV-1124

noliveleger:
### 📣 Summary
Prevent users from deleting their own account if it still contains projects, collections, blocks, questions, templates, or other owned resources.

### 📖 Description
This change adds an “emptiness” check to the self-deletion flow. A user can only delete their account if it no longer owns any resources (e.g., projects, collections, blocks, questions, templates) and has no remaining ownership ties. If the account is not empty, the delete request is rejected.

### 👀 Preview steps

2. Make sure the Admin > Constance > Config > `ALLOW_SELF_ACCOUNT_DELETION` is checked
3. Try to delete your account
4. 🔴 [on main] the request is successful
5. 🟢 [on PR] the request is rejected with a 400

### DIFF
--- a/kpi/views/current_user.py
+++ b/kpi/views/current_user.py
@@ -9,6 +9,7 @@ from rest_framework.response import Response
 
 from kobo.apps.kobo_auth.shortcuts import User
 from kobo.apps.trash_bin.utils import move_to_trash
+from kpi.constants import ASSET_TYPE_EMPTY
 from kpi.schema_extensions.v2.me.serializers import (
     CurrentUserDeleteRequest,
     MeListResponse,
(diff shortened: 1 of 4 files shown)


kpi/views/current_user.py
```diff
@@ -84,6 +85,11 @@ def destroy(self, request, *args, **kwargs):
         if confirm != user.extra_details.uid:
             raise serializers.ValidationError({'detail': t('Invalid confirmation')})
 
+        if user.assets.exclude(asset_type=ASSET_TYPE_EMPTY).count():
```
rgraber: nit: it's more idiomatic (and I believe faster) to use `exists()` rather than `count`

rgraber (changes requested):
One comment and also there seem to be some changes to static/openapi that i'm pretty sure are merge artifacts

kpi/views/current_user.py
noliveleger: Totally right!

PR #6336, opened 2025-10-07
fix(assets): fix infinite load when previewing form as anonymous user DEV-1023

rajpatel24:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Fixes an issue where publicly shared forms failed to load when viewed by an anonymous user. Anonymous form previews now open correctly without infinite loading.

### 📖 Description
Previously, when a project owner shared a form publicly ("Anyone can view this form") and an anonymous user tried to preview it, the page would load indefinitely.
This happened because the backend `asset_snapshots` API returned a `401 Authentication credentials were not provided` error when anonymous users attempted to create a snapshot for form preview.

This PR updates the permission logic to allow anonymous snapshot creation for publicly shared forms while still enforcing proper object-level permissions and password protection.
As a result, public (anonymous) form previews now work as expected, with no regressions for authenticated users or private assets.

### 👀 Preview steps

1. ℹ️ Have an account with a deployed project and at least one submission.
2. Go to Sharing Permissions: either from the “Summary” page modal or the Settings -> Sharing section.
3. Check the box to make the form and/or submissions available publicly.
4. Copy the "Shareable link" link provided.
5. Log out
6. Paste the link into your browser (you should now be an anonymous user).
7. Under the Form tab, click the preview icon.
8. 🔴 [on release] the form preview loads infinitely.
9. 🟢 [on PR] the form preview opens correctly for the anonymous user.

### DIFF
--- a/kpi/permissions.py
+++ b/kpi/permissions.py
@@ -285,8 +285,9 @@ class AssetSnapshotPermission(AssetPermission):
     def has_permission(self, request, view):
         self.validate_password(request)
 
-        # Allow POST requests to pass through to the serializer for
-        # object-level checks
+        # Allow anonymous users to send POST requests to preview public forms.
+        # Object-level access control is handled by the serializer, ensuring
+        # that only public forms can be accessed or previewed.
         if request.method == 'POST' and view.action == 'create':
             return True
 
--- a/kpi/tests/api/v2/test_api_asset_snapshots.py
+++ b/kpi/tests/api/v2/test_api_asset_snapshots.py
@@ -2,16 +2,17 @@
 import re
 
 from django.urls import reverse
-from django.contrib.auth.models import AnonymousUser
 from rest_framework import status
 from rest_framework.exceptions import ParseError
 
 from kobo.apps.form_disclaimer.models import FormDisclaimer
 from kobo.apps.kobo_auth.shortcuts import User
 from kobo.apps.languages.models.language import Language
+from kpi.constants import PERM_VIEW_ASSET
 from kpi.models.asset import AssetSnapshot
 from kpi.tests.kpi_test_case import KpiTestCase
 from kpi.urls.router_api_v2 import URL_NAMESPACE as ROUTER_URL_NAMESPACE
+from kpi.utils.object_permission import get_anonymous_user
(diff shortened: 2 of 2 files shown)


kpi/tests/api/v2/test_api_asset_snapshots.py
```diff
@@ -2,6 +2,7 @@
 import re
 
 from django.urls import reverse
+from django.contrib.auth.models import AnonymousUser
```
noliveleger: It should come before "from django.urls import reverse" A -> Z

kpi/tests/api/v2/test_api_asset_snapshots.py
```diff
@@ -239,6 +240,23 @@ def test_xml_renderer_with_invalid_asset(self):
...
         self.assertIn('_fail', str(context.exception))
 
+    def test_anonymous_can_create_snapshot_when_asset_shared_public(self):
+        self.client.login(username='someuser', password='someuser')
+        asset = self.create_asset('Public asset', self.form_source, format='json')
+        asset_url = reverse(self._get_endpoint('asset-detail'), args=(asset.uid,))
+
+        # Assign 'view' permission on the asset to AnonymousUser
+        anon_user = AnonymousUser()
+        self.add_perm(asset, anon_user, 'view_')
```
noliveleger: I was not aware about this method, we should get rid of it entirely.
Let's use `asset.assign_perm(get_anonymous_user(), PERM_VIEW_ASSET)` instead (for new code).

kpi/permissions.py
```diff
@@ -282,6 +282,16 @@ def __init__(self, *args, **kwargs):
                     'model_name': model_name,
                 }
 
+    def has_permission(self, request, view):
+        self.validate_password(request)
+
+        # Allow POST requests to pass through to the serializer for
+        # object-level checks
```
noliveleger: I think we should make the description more obvious. 
What about: 

> Allow anonymous users to send POST requests in order to preview public forms.
> Object-level access control is handled by the serializer, which ensures that only public forms can be accessed or previewed.

noliveleger (changes requested):

PR #6338, opened 2025-10-07
test(storybook): DeleteAccountBanner DEV-1117

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 💭 Notes

Changes:
- Moved `meMock` and `environmentMock` to be globally defined for all stories
- `DeleteAccountBanner.stories.tsx` now has two more stories: User Has Assets, and User Owns MMO
- `DeleteAccountBanner.tsx` no longer needs test code to be present
- msw mocks now rely on overrides to test particular cases
- Bumped `msw` and `msw-storybook-addon` versions while debugging this, hoping it would solve an issue I had with msw returning 404s for existing mocks. The problem solved by itself :shrug: but this doesn't break anything so I left it here. 

The problem was caused by mocks not returning proper customized responses for non-default cases. I redid the way we handle those and was able to do it in such a way that we no longer need to adjust original components.

**Unfortunately** the solution above will not work for "Docs" view (the one with all stories for given component), as MSW loads all these stories in same iframe/window. Meaning that the last defined MSW mock will take priority. 

So if one story returns a different response for `/assets` endpoint, it will work when viewed separately (and AFAIK in Chromatic, as it doesn't care about "Docs" view), but in "Docs" view, all `/assets` calls will return identical response (the last defined mock).

This is a known limitation as I learned and I haven't found a nice solution. The one that would work would require us having parameters being passed to components (and then further to queries and all places that make the API call) so that finally it will end up added as search query. The only way for MSW mock to return different response (regardless of being rendered alone or among other stories) is to have the URL being different. I am not sure if we are ok and willing to make such a change in the code to make the "Docs" view work.

The Docs view issue will be fixed separately - https://linear.app/kobotoolbox/issue/DEV-1139/teststorybook-fix-msw-in-docs-view

### 👀 Preview steps

1. ℹ️ run storybook (`npm run storybook`)
2. Open http://localhost:6006/
3. Go to `DeleteAccountBanner` stories
5. 🟢 Notice that "User Has Assets" story displays "You need to delete or transfer ownership of all projects…" message
6. 🟢 Notice that "User Has No Assets" story displays "Delete your account and all your account data." message
7. 🟢 Notice that "User Owns MMO" story displays "You need to transfer ownership of your organization…" message

### DIFF
--- a/jsapp/js/account/DeleteAccountBanner.stories.tsx
+++ b/jsapp/js/account/DeleteAccountBanner.stories.tsx
@@ -2,7 +2,6 @@ import type { Meta, StoryObj } from '@storybook/react-webpack5'
 import type { DecoratorFunction } from '@storybook/types'
 import { withRouter } from 'storybook-addon-remix-react-router'
 import assetsMock from '#/endpoints/assets.mocks'
-import { meMockResponse } from '#/endpoints/me.mocks'
 import organizationMock from '#/endpoints/organization.mocks'
 import { queryClientDecorator } from '#/query/queryClient.mocks'
 import RequireAuth from '#/router/requireAuth'
@@ -17,7 +16,7 @@ const meta: Meta<typeof DeleteAccountBanner> = {
   parameters: {
     msw: {
       handlers: {
-        org: organizationMock({ id: meMockResponse.organization!.uid }),
+        organization: organizationMock(),
         assets: assetsMock(),
       },
     },
@@ -47,7 +46,7 @@ export const UserOwnsMMO: Story = {
   parameters: {
     msw: {
       handlers: {
-        org: organizationMock({ id: meMockResponse.organization!.uid, is_mmo: true }),
+        organization: organizationMock({ is_mmo: true }),
       },
     },
   },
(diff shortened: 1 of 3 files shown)


jsapp/js/account/DeleteAccountBanner.stories.tsx
```diff
@@ -17,7 +16,11 @@ const meta: Meta<typeof DeleteAccountBanner> = {
   argTypes: {},
   parameters: {
     msw: {
-      handlers: [assetsMock, meMock, environmentMock, organizationMock(meMockResponse.organization!.uid)],
+      handlers: {
+        me: meMockResponse,
```
magicznyleszek: @Akuukis why are we adding `meMockResponse`? It's just a `const`, right? Plus we already have `meMock` being added in `preview.tsx` to all the stories

Akuukis: right, that's redundant :ok_hand:

Akuukis (approved):
LGTM & tested. Please improve writing:
- type is `test`, as there's no user-facing changes fixes.
- update notes according to latest changes
- update notes to be clear about fixing "Docs" page being out of scope.

PR #6341, opened 2025-10-08
perf(submissions): consolidate storage and submission counter updates for project and user profile DEV-1140

noliveleger:
### 📣 Summary
Merge updates of storage and submission counters for both projects and user profiles into a single operation each

### 📖 Description
This change streamlines how submission-related counters are updated. Previously, storage and submission counters for projects and user profiles were updated separately, resulting in four distinct database operations. The new implementation consolidates these into two unified updates, one for the project and one for the user profile, executed within a single transaction.

### DIFF
--- a/kobo/apps/openrosa/libs/utils/logger_tools.py
+++ b/kobo/apps/openrosa/libs/utils/logger_tools.py
@@ -793,15 +793,11 @@ def save_attachments(
     media_files: Generator[File],
 ) -> tuple[int, bool]:
     """
-    Return a tuple of two lists.
-    - The former is new attachments
-    - The latter is the replaced/soft-deleted attachments
-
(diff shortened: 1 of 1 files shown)


kobo/apps/openrosa/libs/utils/logger_tools.py
```diff
@@ -787,7 +791,7 @@ def safe_create_instance(
 def save_attachments(
     instance: Instance,
     media_files: Generator[File],
-) -> tuple[list[Attachment], list[Attachment]]:
+) -> tuple[int, bool]:
     """
     Return a tuple of two lists.
     - The former is new attachments
```
rgraber: looks like this comment is outdated

PR #6342, opened 2025-10-08
feat(api): Add filtering and ordering support for /`user-reports` endpoint DEV-287

rajpatel24:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Adds robust filtering and ordering capabilities to the /user-reports API endpoint, enabling users and admins to query large user datasets efficiently by date, usage metrics, and subscription attributes, all backed by the optimized materialized view for scalable performance on millions of records.

### 📖 Description
This PR introduces the filtering and ordering layer for the `/user-reports` endpoint, built on top of the existing materialized view (`user_reports_mv`) that aggregates user-level billing and usage data.

Also, added targeted indexes on high-cardinality numeric and timestamp columns to support range queries without full table scans.

### DIFF
--- a/kobo/apps/user_reports/migrations/0003_add_user_reports_mv_indexes.py
+++ b/kobo/apps/user_reports/migrations/0003_add_user_reports_mv_indexes.py
@@ -1,0 +1,118 @@
+from django.db import migrations
+
+CREATE_PG_TRGM = 'CREATE EXTENSION IF NOT EXISTS pg_trgm;'
+DROP_PG_TRGM = 'DROP EXTENSION IF EXISTS pg_trgm;'
+
+# Btree functional indexes for fast case-insensitive prefix (istartswith) searches
+CREATE_IDX_EMAIL_LOWER = """
+CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_user_reports_email_lower_textpat
+ON user_reports_mv (lower(email) text_pattern_ops);
+"""
+DROP_IDX_EMAIL_LOWER = """
+DROP INDEX IF EXISTS idx_user_reports_email_lower_textpat;
+"""
+
+CREATE_IDX_USERNAME_LOWER = """
+CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_user_reports_username_lower_textpat
+ON user_reports_mv (lower(username) text_pattern_ops);
+"""
+DROP_IDX_USERNAME_LOWER = """
+DROP INDEX IF EXISTS idx_user_reports_username_lower_textpat;
+"""
+
+# GIN index for JSONB subscriptions column
+CREATE_IDX_SUBSCRIPTIONS_GIN = """
+CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_user_reports_subscriptions_gin
+ON user_reports_mv USING gin (subscriptions);
+"""
+DROP_IDX_SUBSCRIPTIONS_GIN = """
+DROP INDEX IF EXISTS idx_user_reports_subscriptions_gin;
+"""
+
+# Numeric / ordering indexes
+CREATE_IDX_SUBMISSIONS_ALL_TIME = """
+CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_user_reports_submission_all_time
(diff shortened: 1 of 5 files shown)


kobo/apps/user_reports/migrations/0003_add_user_reports_mv_indexes.py
```diff
@@ -0,0 +1,123 @@
+from django.conf import settings
+from django.db import migrations
+
+CREATE_PG_TRGM = 'CREATE EXTENSION IF NOT EXISTS pg_trgm;'
+DROP_PG_TRGM = 'DROP EXTENSION IF EXISTS pg_trgm;'
```
noliveleger: What's the goal of this extension?

kobo/apps/user_reports/tests/test_user_reports.py
```diff
@@ -298,3 +298,120 @@ def _get_someuser_data(self):
         )
         self.assertIsNotNone(someuser_data)
         return someuser_data
+
+
+class UserReportsFilterAndOrderingTestCase(BaseTestCase):
```
noliveleger, line 308: This class contains several tests which should run only in stripe environment. 
You can use "@pytest.mark.skipif"  and "requires_stripe" decorator.

kpi/docs/api/v2/user_reports/list.md
```diff
@@ -70,6 +67,25 @@ Supports exact matching and range operators (e.g., gte, lte)
...
+
+`?q=subscriptions__0__id__isnull:False`
+
+No subscriptions
+
+`?q=subscriptions__0__id__isnull:True`
+
+Subscription id
+
+`?q=subscriptions__icontains:sub_1QwPItAR39rDI89stzLJ040p`
```
noliveleger: I think you can use this ` q=subscriptions[]__id:sub_1QwPItAR39rDI89stzLJ040p` instead. The former will search all occurrences of `sub_1QwPItAR39rDI89stzLJ040p` in the JSON field. The latter is more specific and will search only for every `id` property  inside the list which matches `sub_1QwPItAR39rDI89stzLJ040p`.

kobo/apps/user_reports/migrations/0003_add_user_reports_mv_indexes.py
rajpatel24: It enables PostgreSQL’s trigram indexing, which speeds up partial and case-insensitive text searches (ex: `username__icontains=raj`) on large datasets. This helps with fast filtering on fields like username and email.

I've made some updates to the `0003_*` migration file, but I’d like your opinion on how we should approach this. Creating these indexes can take a significant amount of time, and I’m not sure which fields we should index.
Maybe we can revisit this after we deploy the MV and related components, and then decide which indexes to keep and which ones to remove. What do you think?

kobo/apps/user_reports/tests/test_user_reports.py
```diff
@@ -420,3 +405,34 @@ def test_balances_nested_json_filter(self):
...
+            items__price__livemode=False,
+            items__price__billing_scheme=BillingScheme.per_unit,
+            livemode=False,
+            metadata={'organization_id': str(self.organization.id)},
+        )
+        refresh_user_reports_materialized_view(concurrently=False)
+
+        # Filter by subscription ID
+        res = self._get_results(
+            {'q': f'subscriptions__0__id__icontains:{self.subscription.id}'}
```
noliveleger: I think this could be simplified with `subscriptions[]__id:{self.subscription.id}`

kobo/apps/user_reports/tests/test_user_reports.py
```diff
@@ -420,3 +405,34 @@ def test_balances_nested_json_filter(self):
...
+        )
+        self.assertTrue(any(r['username'] == 'someuser' for r in res['results']))
+
+        # Filter by subscription status
+        res = self._get_results(
+            {'q': f'subscriptions__0__status__icontains:{self.subscription.status}'}
```
noliveleger: Same here with `subscriptions[]__status:{self.subscription. status}`

noliveleger (approved):
LGTM

PR #6355, opened 2025-10-10
feat(dataCollectors): display collection client URLs in Django Admin DEV-920

noliveleger:
### 📣 Summary
Show ready-to-use links for Enketo Express and KoboCollect on Data Collector pages in Django Admin


sksondra (commented):
Good from QA.

PR #6357, opened 2025-10-10
fix(dataCollectors): remove assets from groups if owners lose permission DEV-1150

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Ensure data collectors can no longer access projects if the group owner has lost permission.

### 📖 Description
Data collector group owners can only add an asset to their group if they have the manage-asset permission. If that permission is removed, the asset should no longer be assigned to the group and all the associated data collector enketo links should be removed.

### Notes
The big problem that needed to be solved with this PR was the situation where, as part of recalculating permissions, we delete all old ones before possibly adding them back. To make matters even more complicated, we do this every time we save an asset.

The approach this PR takes is to refactor the recalculating of permissions to only delete the old permissions that we know we want to delete, rather than deleting and re-adding the ones we care about, except in the case where `return_instead_of_creating` is passed to the recalculation method. In that case, it's the responsibility of the caller to handle whatever problems may arise from removing the manage_asset permission. This approach has a few benefits:
1. no new db calls. this actually reduces the number of db calls because we aren't re-creating the same permissions that already exist
2. no data collector-specific code outside of the data collector app
3. we can enforce the removal of assets at the model level, no matter if the permission was removed permissions from the DCG owner via the expected endpoint(s) or a different path

Other approaches considered:
1. Updating the bulk delete endpoint to call `post_remove_perm` for each permission removed and using that as the signal instead of ObjectPermission.post_delete. Rejected because it would add at least one database call to the endpoint (to get all existing permissions so we knew which ones we were removing) and because it's endpoint-specific and not future-proof against other code that deletes permissions.
2. Updating `_recalculate_inherited_perms` to call `post_remove_perm` for each permission that is *really* removed instead of removed and added back. Rejected because it would similarly add a database call
3. Updating the custom save() method on ObjectPermission to fire a custom signal when the object was *really* being removed and not just temporarily. Rejected because there are cases where `_recalculate_inherited_perms` will delete a permission and not add it back, and we would have no way of knowing if the first deletion is temporary or not.

### 👀 Preview steps

1. ℹ️ have at least 2 accounts and a project
5. Grant manage_asset permissions to user 2
6. In Django admin, create a data collector group with user 2 as the owner and assign the project to it
7. Add a data collector to the group 
8. Copy the URL link from the data collector's admin page (you either need to Save and Continue Editing or just reopen it after you save)
9. Verify the URL opens the project
10. Revoke user 2's manage_asset permissions
11. Refresh the data collector URL (removing the x from the URL to make sure it isn't cached)
12. :red_circle: [on main] survey still opens
13. 🟢 [on PR] error
14. Repeat 2-10 but remove user 2 entirely from the project instead of just changing permissions

### DIFF
--- a/hub/admin/extend_user.py
+++ b/hub/admin/extend_user.py
@@ -121,10 +121,10 @@ class InactiveUsersAsOfFilter(SimpleListFilter):
 
     def lookups(self, request, model_admin):
         return (
-            ('90', 'Inactive ≥ 90 days'),
-            ('180', 'Inactive ≥ 180 days'),
-            ('365', 'Inactive ≥ 365 days'),
-            ('730', 'Inactive ≥ 730 days'),
+            ('90', 'Inactive > 90 days'),
+            ('180', 'Inactive > 180 days'),
+            ('365', 'Inactive > 365 days'),
+            ('730', 'Inactive > 730 days'),
         )
 
     def queryset(self, request, queryset):
@@ -197,7 +197,7 @@ class ExtendedUserAdmin(AdvancedSearchMixin, UserAdmin):
             {'fields': ('deployed_forms_count', 'monthly_submission_count')},
         ),
     )
-    actions = ['remove', 'delete']
+    actions = ['remove', 'delete', 'mark_inactive']
 
     class Media:
         css = {'all': ('admin/css/inline_as_fieldset.css',)}
@@ -229,6 +229,18 @@ class ExtendedUserAdmin(AdvancedSearchMixin, UserAdmin):
         users = list(queryset.values('pk', 'username'))
         self._remove_or_delete(
             request, users=users, grace_period=0, retain_placeholder=False
+        )
+
+    @admin.action(description='Mark selected users inactive')
+    def mark_inactive(self, request, queryset, **kwargs):
+        """
+        Mark selected users as inactive
+        """
+        updated_count = queryset.update(is_active=False)
+        self.message_user(
+            request,
+            f'{updated_count} user(s) marked inactive.',
+            level=messages.SUCCESS,
         )
 
     def deployed_forms_count(self, obj):
--- a/kobo/apps/hook/views/v2/hook.py
+++ b/kobo/apps/hook/views/v2/hook.py
@@ -181,7 +181,7 @@ class HookViewSet(
         serializer.save(asset=self.asset)
 
     @action(detail=True, methods=['PATCH'])
-    def retry(self, request, uid=None, *args, **kwargs):
+    def retry(self, request, uid_hook, *args, **kwargs):
         hook = self.get_object()
         response = {'detail': t('Task successfully scheduled')}
         status_code = status.HTTP_200_OK
--- a/kobo/apps/hook/views/v2/hook_log.py
+++ b/kobo/apps/hook/views/v2/hook_log.py
@@ -106,7 +106,7 @@ class HookLogViewSet(AssetNestedObjectViewsetMixin,
         return queryset
 
     @action(detail=True, methods=['PATCH'])
-    def retry(self, request, uid=None, *args, **kwargs):
+    def retry(self, request, uid_log, *args, **kwargs):
         """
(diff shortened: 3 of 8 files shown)


kobo/apps/subsequences/tests/test_submission_extras_api_post.py
```diff
@@ -555,7 +555,7 @@ def test_google_transcript_post(self, m1, m2):
             'submission': submission_id,
             'q1': {GOOGLETS: {'status': 'requested', 'languageCode': ''}}
         }
-        with self.assertNumQueries(FuzzyInt(49, 65)):
+        with self.assertNumQueries(FuzzyInt(45, 65)):
```
rgraber, line 558: We've reduced the number of queries in _recalculate_inherited_perms!

kobo/apps/data_collectors/tests/test_models.py
```diff
@@ -98,3 +101,53 @@ def test_change_group_and_token_removes_and_sets_correct_links(
...
+        asset.data_collector_group = another_group
+        asset.save()
+        asset.deploy(backend='mock')
+        data_collector = DataCollector.objects.create(name='DC', group=another_group)
+        self.client.force_login(user=someuser)
+        with patch.object(
+            MockDeploymentBackend, 'remove_data_collector_enketo_links'
+        ) as patched_remove_links:
+            url = reverse('api_v2:asset-permission-assignment-bulk-actions',
+                          kwargs={'uid_asset': asset.uid})
```
noliveleger: black format 🙏

kpi/mixins/object_permission.py
```diff
@@ -320,12 +310,27 @@ def _recalculate_inherited_perms(
...
-                    objects_to_return.append(new_permission)
-                else:
-                    new_permission.save()
+                objects_to_return.append(new_permission)
         if return_instead_of_creating:
+            if not stale_already_deleted:
+                self.permissions.filter(inherited=True).delete()
             return objects_to_return
+        else:
+            existing_objs = self.permissions.filter(inherited=True)
```
noliveleger, line 319: We may need to add `deny=False` just in case. 
In a collection, the child could inherit from the parent but denied at the child level. 

TBH, I don't remember what's the expected behaviour with collections. 

For example:

- Create a collection 
- Assign `manage_asset` to UserA
- Create a block within the collection
- UserA receives inherited permissions on the block from its parent
- Remove the permission at the block level
- Reassign the permission at the block level 

🤔  Maybe not! 
I made some tests and the results with `assign_perm` and `remove_perm` are exactly the same. (still buggy, user receives twice `view_asset` - one inherited, one not - but it's out of scope of this PR)

TL;DR - Do not change anything!

kpi/mixins/object_permission.py
```diff
@@ -320,12 +310,27 @@ def _recalculate_inherited_perms(
...
+                if {'permission_id': perm.permission_id, 'user_id': perm.user_id}
+                not in existing
+            ]
+            for perm in objects_to_return:
+                existing_objs = existing_objs.exclude(
+                    permission_id=perm.permission_id, user_id=perm.user_id
+                )
+            existing_objs.delete()
+            for perm in to_add:
+                perm.save()
```
noliveleger: Could we use `bulk_create` (with ignore conflicts, update_fields) to save more queries? Or it's not possible because `.save()` calls a signal needed to calculate descendant and `bulk_create` does not?

noliveleger (changes requested):

kpi/mixins/object_permission.py
rgraber: Testing locally it seems that using bulk_create is fine. At least no unit tests fail. The request to calculate descendants come from objects that have the ObjectPermissionMixin rather than the ObjectPermission models.
ETA we don't need to use update_fields though since we shouldn't be updating anything

noliveleger (approved):
LGTM. Always fun to have less queries to assign perms ;-)

PR #6364, opened 2025-10-13
test(backend): mock invalid non-XLS URL import

rgraber:
### 💭 Notes
Use a non-xls-string instead of hitting Google so we're not subject to Google's whims when trying to test what to do with non-xls input. Updated both v1 and v2 tests to avoid any confusion when eventually v1 is removed (ie to make it obvious that they are redundant and do the same thing).

### DIFF
--- a/kpi/tests/api/v2/test_api_imports.py
+++ b/kpi/tests/api/v2/test_api_imports.py
@@ -906,6 +906,7 @@ class AssetImportTaskTest(BaseTestCase):
         }
         self._post_import_task_and_compare_created_asset_to_source(task_data,
                                                                    self.asset)
+
     @responses.activate
     def test_import_non_xls_url(self):
         """


kpi/tests/api/v2/test_api_imports.py
```diff
@@ -907,12 +907,13 @@ def test_import_asset_xls(self):
...
                                                                    self.asset)
 
-    def test_import_non_xls_url(self):
+    def test_import_non_xls(self):
         """
         Make sure the import fails with a meaningful error
         """
+        encoded_xls = base64.b64encode(b'I am not xls')
         task_data = {
-            'url': 'https://www.google.com/',
```
jnm, line 915: I'm sorry for doing this evil thing, but could we keep test coverage of invalid URL imports? Maybe by mocking the request? See comments on #1727, where there were problems with handling non-XLS URLs 🧓

kpi/tests/api/v1/test_api_imports.py
```diff
@@ -218,12 +218,13 @@ def test_import_asset_xls(self):
...
 
-    def test_import_non_xls_url(self):
+    def test_import_non_xls(self):
         """
         Make sure the import fails with a meaningful error
         """
+        encoded_xls = base64.b64encode(b'I am not xls')
         task_data = {
-            'url': 'https://www.google.com/',
+            'base64Encoded': 'base64:{}'.format(to_str(encoded_xls)),
```
jnm: I'm requesting some changes but only on the v2 tests, but please use your judgment about whether it'd be better (less confusing in the future) to have both v1 and v2 tests identical, or just avoid more busy work by leaving this v1 test as-is.

I'm actually surprised at the code duplication in the viewset; I thought identical v1 endpoints imported their logic from v2 🤔  Anyway, I know @noliveleger is excited to remove all of KPI v1 beginning next year.

Image

PR #6323, opened 2025-10-02
refactor(api)!: show `uid` instead of `parent_lookup` DEV-1049

RuthTurk:
### 📣 Summary
Update the drf-spectacular api documentation to use a consistent `uid_<object_name>` naming convention.

### 👀 Preview steps

1. ℹ️ checkout `api/v2/docs`
2. 🟢 [on PR] make sure that all `parent_lookup_` have been changed to `uid_` and that all nested uids are correct. 
Example:
`/api/v2/assets/{parent_lookup_asset}/hooks/{parent_lookup_hook}/logs/{uid}/`
Should be
`/api/v2/assets/{uid_asset}/hooks/{uid_hook}/logs/{uid_log}/`
3.  Test the endpoints to make sure they are still working

PR #6366, opened 2025-10-13
fix(frontend): align import logic with api change from `uid` to `uid_import` DEV-1156

RuthTurk:
### 🗒️ Checklist

1. [X] run linter locally
2. [X] update developer docs (API, README, inline, etc.), if any
3. [X] for user-facing doc changes create a Zulip thread at `#Support Docs Updates`, if any
4. [X] draft PR with a title `<type>(<scope>)<!>: <title> DEV-1234`
5. [X] assign yourself, tag PR: at least `Front end` and/or `Back end` or `workflow`
6. [X] fill in the template below and delete template comments
7. [X] review thyself: read the diff and repro the preview as written
8. [X] open PR & confirm that CI passes & request reviewers, if needed

### 💭 Notes
The bug was a result of #6323 

### 👀 Preview steps

1. ℹ️ have an account
2. upload XLSForm
3. click import button
4. 🔴 [on release] receive an error message: TypeError
undefined is not an object (evaluating 'response.messages.error')
5. Check network tab and notice that a request was made to `http://kf.kobo.local/api/v2/imports/undefined/` returning a 404 
5. 🟢 [on PR] form is imported successfully

### DIFF
--- a/jsapp/js/dropzone.utils.tsx
+++ b/jsapp/js/dropzone.utils.tsx
@@ -115,7 +115,7 @@ function onImportSingleXLSFormFile(name: string, base64Encoded: string | ArrayBu
         // We have to dig deep for that single asset uid :)
         const firstCreated = importData.messages.created[0]
         if (firstCreated?.uid) {
-          router!.navigate(ROUTES.FORM.replace(':uid_import', firstCreated.uid))
+          router!.navigate(ROUTES.FORM.replace(':uid', firstCreated.uid))
         }
       }
     },


jsapp/js/dropzone.utils.tsx
```diff
@@ -115,7 +115,7 @@ function onImportSingleXLSFormFile(name: string, base64Encoded: string | ArrayBu
         // We have to dig deep for that single asset uid :)
         const firstCreated = importData.messages.created[0]
         if (firstCreated?.uid) {
-          router!.navigate(ROUTES.FORM.replace(':uid', firstCreated.uid))
+          router!.navigate(ROUTES.FORM.replace(':uid_import', firstCreated.uid))
```
Akuukis: should update `ROUTES.FORM` as well that's still `'/forms/:uid'`?

RuthTurk: Thanks for flagging this, I tested further and realized that it was actually not a necessary change and removed it. Please test again just to verify 🙏🏻

PR #6384, opened 2025-10-17
chore(openAPI): update orval with main DEV-1173

RuthTurk:
### 💭 Notes
Neglected to run orval with #6383, so updating it here to fix the failure on main


Akuukis (approved):
LGTM, ran `npm run build:orval` locally and produced no diff ✔️

PR #6385, opened 2025-10-19
fix(submissions): reject submissions with invalid xml character DEV-1129

rajpatel24:
### 📣 Summary
Reject form submissions containing invalid or invisible XML characters, to prevent data corruption when responses are later viewed or edited.

### 📖 Description
Some clients can submit XML content containing characters outside the range permitted by the XML standard (invisible control characters, unpaired surrogates, etc.). These submissions could previously be stored as-is, only to later fail when opened or edited in Enketo, risking data loss or broken records.

This adds server-side validation (validate_xml_chars) that rejects such submissions upfront with a clear 400 error message, instead of letting them silently corrupt data down the line. Only characters permitted by the XML standard are accepted.

This validation is kept as a safety net for any client (not just Enketo) that might still submit disallowed characters, now that Enketo itself has been fixed to strip them before submission.

### 👀 Preview steps

Manual testing through the browser/Enketo won't reliably reproduce this: Enketo now strips invalid characters client-side before submission, so the payloads this PR guards against never reach the server via normal form-filling. Please validate via the unit tests instead.

If you want to trigger this manually anyway, curl directly against the submission endpoint bypasses Enketo's client-side stripping and will reproduce the rejection


noliveleger:
Draft until DEV-1128 is done!

kobo/apps/openrosa/libs/utils/logger_tools.py
```diff
@@ -360,6 +362,26 @@ def dict2xform(submission: dict, xform_id_string: str) -> str:
     return xml_head + dict2xml(submission) + xml_tail
 
 
+def validate_xml_chars(xml: str) -> None:
+    """
+    Reject a submission whose XML contains characters outside the ranges
+    allowed by the XML spec (https://www.w3.org/TR/REC-xml/#charsets)
+    """
+    invalid_xml_char_re = re.compile(
```
rgraber, line 370: Why are we mixing `\x` `\u` and `\U` representations?

rgraber (commented):
One question

kobo/apps/openrosa/libs/utils/logger_tools.py
rajpatel24: Ah, good catch! There's no functional difference. I just wasn't consistent with the escape notation. I'll standardize on `\u` for values ≤ `0xFFFF` and keep `\U` for values above `0xFFFF`

rgraber (approved):
LGTM

PR #6386, opened 2025-10-20
ci(openapi): fix drf-spectacular job DEV-1177

Akuukis:
### 💭 Notes

#6287 never worked (see [logs](https://github.com/kobotoolbox/kpi/actions/runs/17979532426/job/51141577133#step:9:31)) for two reasons:
- `/dev/tty` doesn't exist in GHAs -> add a workaround
- run script through `kobo-install` like locally makes no sense on GHA -> revert to run script directly

Also
- add bash flags to fail on errors within piping
- simplify the job one-value-matrix
- run job on manual schema changes

### 👀 Preview steps

1. see the GHA job
2. 🟢 notice that schema is generated normally successfully
3. 🟢 notice that diff check fails, because `main` contains outdated schema file.

### DIFF
--- a/.github/workflows/openapi.yml
+++ b/.github/workflows/openapi.yml
@@ -44,7 +44,7 @@ jobs:
       - name: Set up Python 3.10
         uses: actions/setup-python@v6
         with:
-          python-version: 3.10
+          python-version: '3.10'
 
       - name: Install pip-tools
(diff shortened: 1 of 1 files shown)


RuthTurk (approved):
LGTM

PR #6388, opened 2025-10-20
fix(data): dynamically add empty validation status dict DEV-1152

rgraber:
### 🗒️ Checklist

1. [ ] run linter locally

### 📣 Summary
Fix error when loading data tables for submissions created before validation_statuses were added.

### 💭 Notes
Set `_validation_status` to `{}` if it's None on the Instance (which may happen if instances are bulk updated or created) when saving to mongo, and, on the other side, set `_validation_status` to `{}` when returning mongo results if it would otherwise be empty.

### 👀 Preview steps

2. Add submissions to the project
3. In a mongo shell, run `db.instances.updateMany({"_userform_id": "<username>_<asset_uid>"}, {$unset: {"_validation_status":""}})` . This will mimic old instances where `_validation_status` isn't present
4. Go to the data table for the project
5. 🔴 [on main] Error
6. 🟢 [on PR] Data table loads successfully

### DIFF
--- a/kobo/apps/openrosa/apps/viewer/models/parsed_instance.py
+++ b/kobo/apps/openrosa/apps/viewer/models/parsed_instance.py
@@ -11,7 +11,7 @@ from pymongo import UpdateOne
 from pymongo.errors import PyMongoError
 
 from kobo.apps.hook.utils.services import call_services
-from kobo.apps.openrosa.apps.logger.models import Instance, Note, XForm, Attachment
+from kobo.apps.openrosa.apps.logger.models import Attachment, Instance, Note, XForm
 from kobo.apps.openrosa.apps.logger.models.attachment import AttachmentDeleteStatus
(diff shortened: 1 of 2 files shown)


kobo/apps/openrosa/apps/viewer/models/parsed_instance.py
```diff
@@ -304,7 +304,7 @@ def to_dict_for_mongo(self):
                 MONGO_STRFTIME),
             TAGS: list(self.instance.tags.names()),
             NOTES: self.get_notes(),
-            VALIDATION_STATUS: self.instance.get_validation_status(),
+            VALIDATION_STATUS: self.instance.get_validation_status() or {},
```
noliveleger: I would move the logic of the `or {}` inside `get_validation_status()`. This way it would be consistent all the time when `self.instance.get_validation_status()` is called.

noliveleger (approved):
LGTM

PR #6390, opened 2025-10-20
refactor(mantine): migrate QA tags response form to mantine DEV-1168

duvld:
### 🗒️ Checklist

1. [x] run linter locally

### 💭 Notes
TLDR: Needed to create custom theme for mantine's `TagInput` as it has some hacky styling that is different than our defaults.

Mantine's `TagInput` (and I suspect every other tag input library in some way) has a "input" that's actually a div, which wraps around an actual `<input>` element inside. This way it can display the tags in the same wrapper, before where the user would actually type the tags in. This means the default styles for mantine `<input>`'s can't apply here. The div wrapper "input" needs `focus-within` rather than `focus-visible`, as well as other default styles that we want to apply to its wrapper div instead. It's probably easier to understand by reading the code changes

### 👀 Preview steps
1. ℹ️ have an account and a project with NLP submissions
2. open NLP view
3. add a tags analysis question
5. 🟢 [on PR] notice that there is no functional differences between main

### DIFF
--- a/jsapp/js/components/processing/analysis/responseForms/integerResponseForm.component.tsx
+++ b/jsapp/js/components/processing/analysis/responseForms/integerResponseForm.component.tsx
@@ -1,6 +1,5 @@
+import { NumberInput } from '@mantine/core'
 import React, { useContext, useState } from 'react'
-
-import TextBox from '#/components/common/textBox'
 import AnalysisQuestionsContext from '#/components/processing/analysis/analysisQuestions.context'
 import { AUTO_SAVE_TYPING_DELAY } from '#/components/processing/analysis/constants'
 import {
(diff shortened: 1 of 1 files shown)


jsapp/js/components/processing/analysis/responseForms/tagsResponseForm.component.tsx
```diff
@@ -60,24 +60,7 @@ export default function TagsResponseForm(props: TagsResponseFormProps) {
...
 
       <section className={commonStyles.content}>
-        <TagsInput
-          value={response}
-          onChange={onTagsChange}
-          // Adds a listener to changes on the internal text field before
-          // text is added as a tag
-          inputProps={{
-            onChange: () => {
-              analysisQuestions?.dispatch({ type: 'hasUnsavedWork' })
```
magicznyleszek, line 70: I wanted to comment that this is now missing, but I tested current code on kf.beta, and I think this was completely unneeded, as there is no way to leave Analysis tab with some tag being lost.

magicznyleszek (approved):

PR #6396, opened 2025-10-21
feat(formBuilder): silently support background-geopoint type DEV-1118

magicznyleszek:
### 📣 Summary

Forms with `background-geopoint` no longer display errors when opened through Form Builder. Saving such form will not result in lost row.

### 💭 Notes

I've added `background-geopoint` type to Form Builder's known types. And there is an override to render an empty template instead of default row. We are prepared to handle some other new types if there is such a need.

There is also a fallback label to be displayed in Single Submission modal for `background-geopoint` rows. AFAIK it never has a `label`, so we are displaying something more useful than empty.

### 👀 Preview steps

Testing

1. ℹ️ Import project with `background-geopoint` question - [BackgroundGeo.xlsx](https://github.com/user-attachments/files/23051715/BackgroundGeo.xlsx)
2. Open form in Form Builder
3. 🟢 Notice that there is no error
4. Add single question
5. Save
6. Go to Project → Form (landing page)
7. Use `…` button ("More actions") in top right and from dropdown choose "Download XLS"
8. 🟢 Notice that downloaded xls file contains `background-geopoint` question


jsapp/xlform/src/view.row.coffee
```diff
@@ -57,6 +57,13 @@ module.exports = do ->
     getRawType: ->
       return @model.get('type').get('typeId')
 
+    # All row types are supported by UI by default. If some type has `supportedByUI` override in `model.configs.coffee`
+    # we respect that.
+    isSupportedByUI: ->
+      if @model.get('type').get('rowType').supportedByUI is false
```
duvld, line 63: Is ` if @model.get('type').get('rowType').supportedByUI is false` checking existence? Why not something like
```
if @model.get('type').get('rowType').supportedByUI
  return false
return true
```
or even
```
return @model.get('type').get('rowType').supportedByUI
```

jsapp/xlform/src/view.row.coffee
```diff
@@ -106,6 +113,11 @@ module.exports = do ->
       return @
 
     _renderRow: ->
+      # For unsupported types we display alternative empty template
+      if not @isSupportedByUI()
+        @$el.html($viewTemplates.$$render('row.unsupportedRowView', @surveyView))
```
duvld, line 118: I don't see this in when I open the formbuilder
image

Is this supposed to be among the other question rows?

magicznyleszek: @duvld it is supposed to be not displayed (it has `display: none`). I thought about not rendering anything, but since I am not familiar with backbone, I thought about having something there. Plus it's easier to debug if a row is there :shrug:

jsapp/xlform/src/view.row.coffee
magicznyleszek: @duvld so the `.supportedByUI` should result in `return false` only if it exists, and only if it is `false`, otherwise it should result in `return true`. So if there is no `.supportedByUI` it should `return true`.

PR #6401, opened 2025-10-22
fix(billing): always send users to stripe portal for plan changes DEV-1004

jamesrkiger:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Changes frontend billing code to always send users changing their plan to Stripe for confirmation instead of handling downgrades in-app.

### 💭 Notes
See [DEV-1004](https://linear.app/kobotoolbox/issue/DEV-1004/account-cant-change-plan-from-storage-to-professional) for background. Previously, when a user was changing their plan, the frontend would determine whether this action was an "upgrade" or a "downgrade" based on the relative prices of the user's current plan and the plan they wanted to switch to. If it was a downgrade, we would display a confirmation modal and only **schedule** an upgrade at the end of their billing cycle if the user confirmed. If it was an upgrade, the frontend would send the user to Stripe for confirmation and any further payment, With this PR, the frontend will now always send users to Stripe, whether for upgrades or downgrades. 

### 👀 Preview steps
This is a very simple change, but I encourage playing around with it locally (beyond the steps described below) to make sure there are no strange scenarios I didn't consider.

1. On a Stripe enabled instance, create a new user.
2. Sign the user up for a 20gb **annual** Storage plan. Sync subscriptions
3. After syncing subscriptions, load the plans page and click "upgrade" to switch to the **monthly** professional plan. (Note: this was the original scenario that started us investigating the issue)
4. 🔴 [on main] You will see a confirmation modal (don't click confirm).
5. 🟢 [on PR] You will be taken to Stripe for confirmation (and it will arrange an immediate plan change). Confirm the change
6. After syncing again, go back to the plans page and try upgrading to the teams plan. This should work the same on main and on this PR.

### DIFF
--- a/kobo/apps/openrosa/apps/logger/models/instance.py
+++ b/kobo/apps/openrosa/apps/logger/models/instance.py
@@ -345,7 +345,7 @@ class Instance(AbstractTimeStampedModel):
         # For example:
         # if not self.validation_status:
         #    self.validation_status = self.asset.settings.get("validation_statuses")[0]
-        return self.validation_status
+        return self.validation_status or {}
 
 
 if Instance.XML_HASH_LENGTH / 2 != sha256().digest_size:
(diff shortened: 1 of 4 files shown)


RuthTurk (approved):
LGTM

PR #6402, opened 2025-10-22
feat(mfa): replace trench based API endpoints, views and forms with allauth implementation DEV-749

Guitlle:
### 🗒️ Checklist

1. [x] run linter locally

### Summary 
Migrate MFA features from Trench to Django Allauth for better integration, maintainability.

### Description
This update replaces all Trench-based multi-factor authentication (MFA) endpoints, views, and forms with a new implementation powered by Django Allauth. The change ensures tighter integration with the existing authentication system, simplifies configuration, and improves long-term maintainability. 

### 💭 Notes
The MFA endpoints were changed to use the allauth MFA feature and removing trench. This covers these endpoints that are used in the frontend:

- /api/v2/auth/str:method/activate/	
- /api/v2/auth/str:method/activate/confirm/	
- /api/v2/auth/str:method/codes/regenerate/
- /api/v2/auth/str:method/deactivate/
- /api/v2/auth/mfa/user-active-methods/

These are now based on the Authenticator model, which can handle TOTP codes and recovery codes.

### 👀 Preview steps
These endpoints are covered in unit tests, so no need to test. However we could test the following sequence to validate it's working ok:

1. Log into your dev env
2. Go to account settings
3. Go to security
4. Enable MFA and you can use pyotp to generate codes, or you can use an auth app in  your phone
5. Scan the QR code or copy the manual key, use pyotp:
```
>>> import pyotp
>>> key = '*****'
>>> totp = pyotp.TOTP(key)
>>> totp.now()

123456
```
6. Enter the generated code and it should now show you the recovery codes. Download the file, and click the blue button
7. You should see that there are two options now, reconfigure MFA and regenerate codes 
8. Test each one of these, it should allow you to follow the steps with no error messages in the KPI container logs or in the frontend app. Use pyotp to generate new codes, as shown above.


scripts/migrate.sh
```diff
@@ -2,6 +2,7 @@
 set -e
 
 python manage.py runscript fix_migrations_for_kobocat_django_app
```
noliveleger: As per our discussion, migrations must be fixed when upgrading KoboCAT migrations from `2.025.43` or `main` branch.

Guitlle: Ok, thanks for the advice

PR #6403, opened 2025-10-22
fix(exports): use API v2 media URLs consistently DEV-1180

noliveleger:
### 📣 Summary
Ensure exported files always reference media URLs from API v2, even when the request context is missing.

### 📖 Description
This fix addresses an issue where media URLs in exports could fall back to the default API v1 paths when the request object was unavailable. The export process now explicitly resolves URLs using the API v2 context, guaranteeing consistent link generation across all environments. This ensures that media references remain correct and accessible, regardless of how or where the export is triggered.

### 👀 Preview steps
1. ℹ️ have an account and a project with media files
2. Submit data
3. Transfer the project to someone else
4. Export data
5. 🔴 [on release branch] notice that media URLs refer to KoboCAT and links don't work
6. 🟢 [on PR] notice  that media URLs refer to KPI and links still work


kpi/models/import_export_task.py
```diff
@@ -1039,9 +1039,6 @@ def get_export_object(
             query=query,
         )
 
-        if source.has_advanced_features:
-            submission_stream = stream_with_extras(submission_stream, source)
-
```
noliveleger, line 1044: Removed because we are already calling it in `get_submissions` already.

kpi/utils/urls.py
```diff
@@ -51,3 +52,29 @@ def is_request_for_html(request: HttpRequest):
...
+    viewname: str,
+    args=None,
+    kwargs=None,
+    request=None,
+    **extra,
+) -> str:
+
+    if 'url_namespace' in extra:
+        url_namespace = extra.pop('url_namespace', None)
+        *_, viewname = viewname.split(':')
```
jnm, line 67: huh, who knew ✨
thanks for teaching me a new trick

PR #6405, opened 2025-10-22
fix(openrosa): prevent 500 error when editing or previewing Dynamic Data Attachment projects DEV-1193 

noliveleger:
### 📣 Summary
Enforce API v2 namespace for OpenRosa URLs in AssetSnapshot to fix broken edit and preview actions.

### 📖 Description
This fix resolves a 500 error that occurred when editing or previewing Dynamic Data Attachment projects. The issue was caused by DRF’s URL resolution logic, which reused the same API version across multiple `reverse()` calls in a shared context. By explicitly enforcing the v2 namespace for OpenRosa URLs in AssetSnapshot, the correct routes are now generated regardless of the active request version, restoring proper behavior for edit and preview actions.

### 👀 Preview steps

1. Create a project (project A)
2. Add a question
3. Deploy it
4. Collect a few submissions (might not be necessary)
5. Enable data sharing in "Connect Projects"
6. Create another project (project B)
7. Add a question
8. Deploy it (might not be necessary)
9. "Import other project data" in "Connect Projects" from project A (import all questions)
10. Try to preview form B
11. 🔴 [on release branch] Failure: "Request to http://kf.kobo.local/asset_snapshots/sXyoXSiepW2EvEMALwMQPP/manifest.xml failed."
12. 🟢 [on PR] You can preview the form and edit data.

### DIFF
--- a/kpi/constants.py
+++ b/kpi/constants.py
@@ -150,4 +150,8 @@ PROJECT_HISTORY_LOG_METADATA_FIELD_OLD = 'old'
 PROJECT_HISTORY_LOG_METADATA_FIELD_ADDED = 'added'
 PROJECT_HISTORY_LOG_METADATA_FIELD_REMOVED = 'removed'
 
-API_NAMESPACES = {'v1': None, 'v2': 'api_v2', 'default': 'api_v2'}
+API_NAMESPACES = {
+    'v1': None,
+    'v2': 'api_v2',
+    'default': 'api_v2',
+}


kpi/models/paired_data.py
```diff
@@ -153,9 +152,13 @@ def get_download_url(self, request):
...
         """
-        return reverse('paired-data-external',
-                       args=(self.asset.uid, self.paired_data_uid, 'xml'),
-                       request=request)
+
+        return versioned_reverse(
+            'paired-data-external',
+            args=(self.asset.uid, self.paired_data_uid, 'xml'),
+            request=request,
+            url_namespace=API_NAMESPACES['default'],
```
Guitlle, line 160: You don't need to specify the default namespace, as it is used by default. I would say we don't need this line here and in the other versioned_reverse calls

Guitlle (approved):
LGTM except for redundant default parameter in versioned_reverse calls

kpi/models/paired_data.py
noliveleger: Did you try without the `url_namespace`?  

The goal of adding it is to be sure that the namespace is forced when resolving the URL. If `request` contains a different versioning scheme (which is the case when we call OpenRosa endpoints nested in AssetSnapShot viewset). This raises an 500 error because it tries to resolve `namespace-different-from-api_v2:paired-data-external`.  

Without `url_namespace` it tries to resolve `openrosa:paired-data-external` (which does not exist), it returns 

```
kpi-1  | Traceback (most recent call last):
kpi-1  |   File "/opt/venv/lib/python3.10/site-packages/django/urls/base.py", line 71, in reverse
kpi-1  |     extra, resolver = resolver.namespace_dict[ns]
kpi-1  | KeyError: 'openrosa'
... 
kpi-1  | django.urls.exceptions.NoReverseMatch: Reverse for 'paired-data-external' not found. 'paired-data-external' is not a valid view function or pattern name.
```

PR #6407, opened 2025-10-23
fix(attachments): force `Content-Disposition` to `attachment` DEV-1200

noliveleger:
### 📣 Summary
Always send files with a `Content-Disposition: attachment` header to ensure proper download behavior.

### 📖 Description
This change enforces the `Content-Disposition` header to be set to attachment for all downloadable responses. Some files were previously displayed inline by browsers instead of being downloaded. By forcing this header, all files are now consistently offered as downloads.

### 👀 Preview steps

1. ℹ️ have an account and a project with attachments
2. Collect data
3. Go to the table view
4. Retrieve the link from the download CTA in the media preview modal
5. Paste the link in another browser window
4. 🔴 [on release branch] notice the `Content-Disposition` header for that file is `inline` and image can be viewed directly in the browser
5. 🟢 [on PR] notice the `Content-Disposition` header for that file is `attachment` and image is directly downloaded

### DIFF
--- a/kobo/apps/openrosa/apps/api/viewsets/attachment_viewset.py
+++ b/kobo/apps/openrosa/apps/api/viewsets/attachment_viewset.py
@@ -160,8 +160,10 @@ class AttachmentViewSet(OpenRosaReadOnlyModelViewSet):
             and self.object.media_file is not None
         ):
             data = self.object.media_file.read()
-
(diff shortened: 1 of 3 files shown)


jnm (approved):
Tested locally 🆗

PR #6409, opened 2025-10-27
fix(assets): handle anonymous user access on asset snapshot list DEV-1105

rajpatel24:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Fixes an issue where anonymous users received a 500 error when accessing the asset snapshots list endpoint. Anonymous requests now safely return an empty response instead of causing a server error.

### 📖 Description
Previously, the `/api/v2/asset_snapshots/` endpoint raised a server error when accessed anonymously, because the code attempted to access `organization.is_admin_only()` even when no organization was associated with the user.
This PR updates the filtering logic to handle anonymous users gracefully and adds a unit test to ensure no 500 error occurs in such cases.

### 👀 Preview steps
1. 🔴 [on release] Hit the `/api/v2/asset_snapshots/` endpoint anonymously. It returns a 500 error.
2. 🟢 [on PR] Hit the same endpoint again. It now returns a 200 response (with an empty list, as expected).
3. Test with an authenticated user, ensure their asset snapshots are still listed correctly.

### DIFF
--- a/kpi/filters.py
+++ b/kpi/filters.py
@@ -423,7 +423,7 @@ class RelatedAssetPermissionsFilter(KpiObjectPermissionsFilter):
 
         user = get_database_user(request.user)
         organization = user.organization
-        if organization.is_admin_only(user):
+        if organization and organization.is_admin_only(user):
             # Admins do not receive explicit permission assignments,
             # but they have the same access to assets as the organization owner.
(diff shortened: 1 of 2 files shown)


RuthTurk (approved):
LGTM

PR #6422, opened 2025-10-28
feat(subsequences): migrate old SubmissionExtras to SubmissionSupplemental DEV-934

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 💭 Notes
Fill out the method for converting old SubmissionExtra content dicts to the new format expected by SubmissionSupplemental for translations and transcripts.
This code makes numerous assumptions to fill in information that is not present in the old structure but required in the new:

1. If old[xpath]['transcript']['value'] == old[xpath]['googlets']['value']and the language codes are the same, we assume the most recent transcript was automatically generated
2. If, for any revision in old[xpath]['transcript']['revisions'], revision['value'] is the same as old[xpath]['googlets']['value'] and the language codes match, we assume that revision was automatically generated. If multiple match, we assume they were all automatically generated. This should be pretty rare but it's possible
3. 1-2 also apply to transcriptions
4. old[xpath]['transcript']['dateModified'] will be assumed to be the creation date of the most recent revision (ie whatever is in old[xpath]['transcript']['value']). The same goes for translations
5. All uuids are newly generated
6. All old transcriptions/translations have status=complete with a _dateAccepted of now() (whenever the code is running)
7. To determine the dependency of any old translation, whether automated or manual:

* If we know the source language, look for the most recent transcript in that language that was created before the translation. 
* If there is none, take the most recent transcription in that language
* If there are no transcriptions in the source language, take the most recent transcript
* If we don't know the source language, take the most recent transcript that was created before the translation
* If there is none, take the most recent transcript

8. We can ignore any badly formatted revisions/transcripts/translations
9. Most recent revisions will be first in the version array

### 👀 Preview steps
On main:
1. ℹ️ have an account and a project with an audio question
2. Enable NLP
3. Add a submission to the project
4. Generate an automatic transcript in English
5. Generate an automatic translation in Spanish
6. Manually edit the English transcript and save
7. Manually edit the Spanish translation and save

On PR branch:
8. Clone the asset
9. PATCH the asset with 
```
{
  "_version": "20250820",
  "_actionConfigs": {
    "<question xpath>": {
      "manual_transcription": [{"language": "en"}],
      "automated_google_transcription": [{"language": "en"}],
      "manual_translation": [{"language": "es"}],
      "automated_google_translation": [{"language": "es"}],
      "qual": [{"val": 2, "uuid": "1f30bec7-4843-43c7-98bc-13114af240d6"}],
    }
  }
}
```
11. In a python shell, run 
```
from kobo.apps.subsequences.schemas import validate_submission_supplement
from kobo.apps.subsequences.utils.versioning import migrate_submission_supplementals
se = SubmissionExtras.objects.get(submission_uuid=<submission uuid>)
validate_submission_supplement(<cloned_asset>, migrate_submission_supplementals(se.content))
```
11. :green_circle: The validation should pass

### DIFF
--- a/kobo/apps/subsequences/README.md
+++ b/kobo/apps/subsequences/README.md
@@ -141,7 +141,12 @@ POST to `/api/v2/assets/{uid_asset}/advanced-features/` with:
   ]
 }
 ```
-See [API documentation](/api/v2/docs/?q=/api/v2/assets/{uid_asset}/advanced-features/) for more details.
+To learn more, visit `/api/v2/docs/` on your KoboToolbox server.
+
+Or use the public documentation endpoints:
+
+- Global Server: [API documentation](https://kf.kobotoolbox.org/api/v2/docs/?q=/api/v2/assets/{uid_asset}/advanced-features/)
+- EU Server: [API documentation](https://eu.kobotoolbox.org/api/v2/docs/?q=/api/v2/assets/{uid_asset}/advanced-features/)
 
 ---
 
@@ -188,7 +193,12 @@ PATCH /api/v2/assets/<asset_uid>/data/<submission_root_uuid>/supplement/
   }
 }
 ```
-See [API documentation](/api/v2/docs/?q=/api/v2/assets/{uid_asset}/data/{id}/supplement/) for more details.
+To learn more, visit `/api/v2/docs/` on your KoboToolbox server.
+
+Or use the public documentation endpoints:
+
+- Global Server: [API documentation](https://kf.kobotoolbox.org/api/v2/docs/?q=/api/v2/assets/{uid_asset}/data/{id}/supplement/)
+- EU Server: [API documentation](https://eu.kobotoolbox.org/api/v2/docs/?q=/api/v2/assets/{uid_asset}/data/{id}/supplement/)
 
 ---
 


kobo/apps/subsequences/utils/versioning.py
noliveleger, line 1: Let's be consistent with our convention and please move "protected" utilities at the end of the file.

kobo/apps/subsequences/schemas.py
```diff
@@ -29,7 +29,7 @@
 
 
 def validate_submission_supplement(asset: 'kpi.models.Asset', supplement: dict):
-    jsonschema.validate(get_submission_supplement_schema(asset), supplement)
+    jsonschema.validate(supplement, get_submission_supplement_schema(asset))
```
noliveleger, line 40: 💪

kobo/apps/subsequences/utils/versioning.py
```diff
@@ -39,6 +173,86 @@ def migrate_advanced_features(advanced_features: dict) -> dict | None:
     return migrated_advanced_features
 
 
+def migrate_submission_supplementals(supplemental_data: dict) -> dict | None:
+    if supplemental_data.get('_version', None) == SCHEMA_VERSIONS[0]:
```
noliveleger: Nit: `.get()` already defaults to `None`, so you can drop the second argument.

kobo/apps/subsequences/tests/test_versioning.py
```diff
@@ -0,0 +1,287 @@
...
+            'kobo.apps.subsequences.utils.versioning.uuid.uuid4',
+            side_effect=['uuid1', 'uuid2', 'uuid3', 'uuid4'],
+        ):
+            with freeze_time(now):
+                migrated = migrate_submission_supplementals(old_version)
+
+        new_version = {
+            '_version': '20250820',
+            'Audio_question': {
+                'automatic_transcription': {
```
noliveleger: That's not the name of the action. It should be `automatic_google_transcription`. We are only using Google for NLP at the moment. The logic works great, but it should match the current action `ID`s. 

Moreover, we need to rename every `automated_*` and `Automated*`  to  their `(A|a)utomatic` counterparts.

kobo/apps/subsequences/tests/test_versioning.py
```diff
@@ -0,0 +1,287 @@
...
+                            '_dateCreated': a_year_and_a_day_ago,
+                            '_dateAccepted': now.isoformat(),
+                            '_uuid': 'uuid2',
+                            'language': 'en',
+                            'value': 'This is audio that I am trying to transcribe.',
+                            'status': 'complete',
+                        }
+                    ]
+                },
+                'automatic_translation': {
```
noliveleger: Same comment as above.

noliveleger:
Nit: `9. PATCH the asset with` is bad formatted, trailing comas are missing inside the `"<question xpath>"` dictionary.

jnm:
> So everything nested inside the question_name dictionary is ignored/by_passed.

Right :facepalm: what we are doing now is NOT putting the action IDs inside `properties`. They're just hanging out at the top level of the object for each question, not doing anything:

```json
{
  "additionalProperties": false,
  "properties": {
    "_version": {
      "const": "20250820"
    },
    "audio": {
      "additionalProperties": false,
      "properties": {},
      "type": "object",
      "automated_google_transcription": {
       …
```

Unfortunately, correcting this causes other problems because the `$defs` paths are wrong.

kobo/apps/subsequences/README.md
```diff
@@ -117,6 +117,31 @@ POST to `/api/v2/assets/{uid_asset}/advanced-features/` with:
 }
 ```
 
+**Example: Enable Qual action**
```
noliveleger: Let's change it to: Qualitative Analysis Text question

kobo/apps/subsequences/README.md
```diff
@@ -117,6 +117,31 @@ POST to `/api/v2/assets/{uid_asset}/advanced-features/` with:
...
+    },
+    {
+      "type": "qualInteger",
+      "uuid": "1a2c8eb0-e2ec-4b3c-942a-c1a5410c081a",
+      "labels": { "_default": "How many characters appear in the story?" }
+    }
+  ]
+}
+```
+
```
noliveleger, line 150: Let's add something like , See  `[API documentation](/api/v2/docs/?q=/api/v2/assets/{uid_asset}/advanced-features/)` for more details. 

* I did escape the markdown in purpose to show you what I was think about, but use markdown directly.

kobo/apps/subsequences/README.md
```diff
@@ -149,6 +174,20 @@ PATCH /api/v2/assets/<asset_uid>/data/<submission_root_uuid>/supplement/
 }
 ```
 
+#### Example: Qual action
```
noliveleger: Same comment for the name of the action. Let's use the entire name.

kobo/apps/subsequences/README.md
```diff
@@ -285,6 +324,31 @@ It describes the configuration stored on a `QuestionAdvancedFeature` when an act
 }
 ```
 
+**Example: enabling Qual**
```
noliveleger: Change the name 🙏

kobo/apps/subsequences/README.md
```diff
@@ -149,6 +174,20 @@ PATCH /api/v2/assets/<asset_uid>/data/<submission_root_uuid>/supplement/
...
+  "text_question": {
+    "qual": {
+      "uuid": "q_uuid",
+      "value": "sentiment_pos"
+    }
+  }
+}
+```
+
 ---
```
noliveleger, line 203: Like advanced_feautres, Add a note to look at the API docs with the correct link.

noliveleger (changes requested):

noliveleger (approved):
LGTM

PR #6439, opened 2025-11-03
refactor(reactQuery): standard error flow for Orval DEV-948

Akuukis:
### 💭 Notes

Let's define expected behavior for API error handling on frontend as follows (see `onErrorDefaultHandler`):
 * backend owns error content: keep error messages updated in backend, don't ignore/override them on frontend.
 * frontend owns error display: ideally nicely inline, by default with a toast using this default handler.
 
With that in mind, this PR introduces a default error handler `onErrorDefaultHandler` for the react query and refactors all react-query instances to use it properly. Dealing with frontend error message overrides will be done seperately as part of DEV-1218.

Out of scope of this PR are non-react-query API calls and the backend side for them, as both will be a long way to go.

This PR likely displays a few error toasts more than before, but I call it `refactor` instead of `fix` because AFAIK the error handling so far was undefined behavior and thus no expected behavior was changed. Let me know if you disagree.

### 👀 Preview steps

1. ℹ️ have two pending invites to your org
2. open member list in two windows side by side (switching tabs will make react-query to refetch)
3. remove the invite in one tab
4. change the invite's role in the other tab
6. 🟢 [on PR] notice a nice error toast (for ServerError)
7. invite the user again
8. in devtools set network to offline
4. change the invite's role again
6. 🟢 [on PR] notice a nice error toast (for TypeError), actually two for both mutation and invalidation

### DIFF
--- a/jsapp/js/api/onErrorDefaultHandler.ts
+++ b/jsapp/js/api/onErrorDefaultHandler.ts
@@ -1,0 +1,79 @@
+import { type Query, QueryClient } from '@tanstack/react-query'
+import { notify } from '#/utils'
+import { ServerError } from './ServerError'
+
+/**
+ * On error Orval's fetch mutator throws either:
+ * - TypeError
+ * - NotAllowedError (DOMException)
+ * - AbortError (DOMException)
+ * - {@link ServerError}
+ */
+export type OrvalFetchError = TypeError | DOMException | ServerError
+
+/**
+ * By default, every Orval request handles error with this routine, and the default can be overriden on per-hook basis.
+ *
+ * Error handling policy:
+ * - backend owns error content: keep error messages updated in backend, don't ignore/override them on frontend.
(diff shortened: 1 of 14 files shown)


jsapp/js/api/onErrorDefaultHandler.ts
```diff
@@ -0,0 +1,79 @@
...
+ *     onError: () => notify(t('some custom message'), 'error'),  // BAD, don't override.
+ *     onError: () => null,                                       // BEST, but handle the error nicely inline instead!
+ *     // no override                                             // GOOD, leave it to the default handler.
+ *     onError: (error: OrvalFetchError, variables, context) => { // GOOD, use callback for more than default handler.
+ *       // Do more stuff.
+ *       onErrorDefaultHandler(error, variables, context)
+ *     },
+ *   },
+ * ```
+ */
```
magicznyleszek, line 48: I really like the comments you're adding in this PR; BAD vs GOOD examples are nice

PR #6440, opened 2025-11-03
fix(data): dynamically add empty validation status dict DEV-1159

rgraber:
### 📣 Summary
Fix error when loading data tables for submissions created before validation_statuses were added.

### 💭 Notes
Backport of https://github.com/kobotoolbox/kpi/pull/6388
Set `_validation_status` to `{}` if it's None on the Instance (which may happen if instances are bulk updated or created) when saving to mongo, and, on the other side, set `_validation_status` to `{}` when returning mongo results if it would otherwise be empty.

### 👀 Preview steps

2. Add submissions to the project
3. In a mongo shell, run `db.instances.updateMany({"_userform_id": "<username>_<asset_uid>"}, {$unset: {"_validation_status":""}})` . This will mimic old instances where `_validation_status` isn't present
4. Go to the data table for the project
5. 🔴 [on main] Error
6. 🟢 [on PR] Data table loads successfully

### DIFF
--- a/kobo/apps/openrosa/apps/logger/models/instance.py
+++ b/kobo/apps/openrosa/apps/logger/models/instance.py
@@ -368,7 +368,7 @@ class Instance(AbstractTimeStampedModel):
         # For example:
         # if not self.validation_status:
         #    self.validation_status = self.asset.settings.get("validation_statuses")[0]
-        return self.validation_status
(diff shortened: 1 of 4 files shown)


noliveleger (approved):
LGTM

PR #6443, opened 2025-11-04
feat(mfa): migrate trench data DEV-854

Guitlle:
### 🗒️ Checklist

1. [x] run linter locally

### 💭 Notes
The trench MFA data is copied to the allauth Authenticator and the MfaMethodsWrapper models in order to remove the trench dependency. In order to avoid disabling MFA for users it migrates the user MFA data from trench when the user logs in. It also includes a long running migration to migrate users MFA data in the background. 

### 👀 Preview steps
1. ℹ️ have an account with MFA enabled from the previous version with trench
2. Log in
3. 🟢 It should show you the same MFA screen after logging in and the codes and backup codes should work as well
4. For the user the migration is seamless
5. Wait or execute manually the long running migration, 
5.1. Restart all the containers so that the celery tasks are updated and the long running migration is registered
5.2. Enter into a django shell and get the task, it should be named `0013_migrate_mfa_data`
```
>>> lrm = LongRunningMigration.objects.last()
>>> lrm.name

'0013_migrate_mfa_data'

>>> lmr.execute()

Migrating MFA data for user abc...
Done. Migrated 1 users with Trench MFA data.
```
7. Check the mfa_authenticator table to see the migrated data, ensure that the secret is the same, and that the backup codes are populated as well for the migrated data


kobo/apps/stripe/tests/test_mfa_login.py
```diff
@@ -212,12 +180,9 @@ def test_mfa_login_with_canceled_subscription_but_previously_set(self):
             'login': 'someuser',
             'password': 'someuser',
         }
-        response = self.client.post(
-            reverse('kobo_login'), data=data, follow=True
-        )
+        response = self.client.post(reverse('kobo_login'), data=data, follow=True)
         self._assert_mfa_login(response)
 
-    @override_config(MFA_ENABLED=True)
```
noliveleger, line 193: Why did you remove this? 
We want to be sure it's enabled whatever is in the config, right?

That's the same comment for all deletions below ;-)

noliveleger (changes requested):

kobo/apps/stripe/tests/test_mfa_login.py
Guitlle: The allauth's MFA_ENABLED setting is not writable, it is True if the allauth.mfa app is installed. That's why the test `test_mfa_globally_disabled_as_user_with_paid_subscription` simulates having that app uninstalled.

kobo/apps/accounts/mfa/tests/test_api.py
```diff
@@ -40,7 +39,6 @@ def test_user_methods_with_date(self):
         for field in expected_fields:
             self.assertTrue(field in results[0])
 
-    @override_config(MFA_ENABLED=True)
```
noliveleger, line 43: I don't know if it's a glich from GitHub but we should not remove `@override_config(MFA_ENABLED=True|False)` anywhere.

Guitlle: Ok, I'm going to check again they are all there

PR #6444, opened 2025-11-05
feat(formBuilder): add default max-pixels value when adding new photo question DEV-966

magicznyleszek:
### 📣 Summary

When creating `image` question in Form Builder, set `max-pixels` to default value (`1024`) instead of leaving it blank. Also applies default values for `range` question type (the change is universal and will work with any future default values we set).

### 💭 Notes

Changes here:
- Added bunch of comments while trying to understand the code
- All predefined `defaultValue`s of parameters will be applied when question is being created (currently we have those for `image` and `range`)
- **NO** allowing to remove `max-pixels` from `image` (will be done as part of separate linear issue)

### 👀 Preview steps

1. ℹ️ Have an account
2. Create a new form in Form Builder
3. Add `image` question
4. **Do not open the question Settings**
5. Save form
6. Go to Project → Form page (`/landing`)
7. Use "Download XLS" from top right "…" ("More actions") dropdown
8. 🔴 [on main] notice that downloaded file has no `max-pixels` value for the question in the "parameters" column
9. 🟢 [on PR] notice that downloaded file has `max-pixels` for the quesion

### DIFF
--- a/jsapp/xlform/src/view.rowSelector.coffee
+++ b/jsapp/xlform/src/view.rowSelector.coffee
@@ -219,7 +219,8 @@ module.exports = do ->
       typeConfig = $configs.questionParams[rowType]
       initialParameters = {}
       for configName, configValue of typeConfig
-        if (configValue.defaultValue)
+        # We need to allow such values as `0` or `false` here, thus a more lengthy check
+        if (configValue.defaultValue isnt null and configValue.defaultValue isnt undefined)
           initialParameters[configName] = configValue.defaultValue
       if (Object.keys(initialParameters).length > 0)
         rowDetails.parameters = writeParameters(initialParameters)


jsapp/xlform/src/view.rowSelector.coffee
```diff
@@ -200,8 +214,21 @@ module.exports = do ->
         survey = @options.survey
         options.at = 0
 
+      # For some questions we start off with some parameters having default values.
+      # The code was added mostly for `image` to have some initial `max-pixels`.
+      typeConfig = $configs.questionParams[rowType]
+      initialParameters = {}
+      for configName, configValue of typeConfig
+        if (configValue.defaultValue)
+          initialParameters[configName] = configValue.defaultValue
```
p2edwards:
```suggestion
        if (configValue.defaultValue?)
          initialParameters[configName] = configValue.defaultValue
```


I think this should be 

```coffee
if (configValue.defaultValue?)
```

(equivalent to `configValue.defaultValue !== null && configValue.defaultValue !== undefined`)

Otherwise it'd (theoretically) skip applying the default `start=0` for the range slider, or if something defaults to `false`. 

Oddly, the range slider start=0 still works. That odd ParamOption onchange event happens immediately afterwards, notices the missing `start` parameter, and sets it to 0 and calls writeParameters again:

```coffee
onChange: (evt) ->
  if @paramType is $configs.paramTypes.number
    val = evt.currentTarget.value
    # make sure that params without removed values keep using default one
    # TODO: is this the behaviour we want to have?
    if val is '' and typeof @paramDefault isnt 'undefined'
      val = "#{@paramDefault}"
    ...
```


image

magicznyleszek: I went with a bit more explicit line, as it produces a JS code more close to what I would wrote myself :)

PR #6445, opened 2025-11-05
feat(analysis): introduce optional clear button DEV-1194

duvld:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Modern implementations of radio components don't allow de-selection. After updating the select one question type in the qualitative analysis section to mantine, we lost this functionality. To ensure the ability to de-select select one questions, they will appear with a "clear" button next to the "delete" button

### Notes
Prerequisite for DEV-1169

### 👀 Preview steps
1. ℹ️ have an account and a project with NLP questions and submissions
2. go to qualitative analysis
3. add a select-one question type
4. 🔴 [on main] notice that there is no clear button
5. 🟢 [on PR] notice the clear button exists and works

### DIFF
--- a/jsapp/js/components/processing/analysis/responseForms/responseWrapper.component.tsx
+++ b/jsapp/js/components/processing/analysis/responseForms/responseWrapper.component.tsx
@@ -111,7 +111,7 @@ export default function ResponseWrapper(props: ResponseWrapperProps) {
               </ButtonNew>
 
               <ButtonNew size='md' onClick={deleteQuestion} variant='danger'>
-                {t('Delete account')}
+                {t('Delete')}
               </ButtonNew>
             </Group>
           </Stack>
--- a/jsapp/js/api/models/hook.ts
+++ b/jsapp/js/api/models/hook.ts
@@ -29,8 +29,8 @@ export interface Hook {
   /** @nullable */
   payload_template?: string | null
   readonly pending_count: number
-  settings: HookSettings
-  subset_fields: string[]
+  settings?: HookSettings
+  subset_fields?: string[]
   readonly success_count: number
   readonly uid: string
   readonly url: string
--- a/jsapp/js/components/processing/analysis/editors/analysisQuestionEditor.component.tsx
+++ b/jsapp/js/components/processing/analysis/editors/analysisQuestionEditor.component.tsx
@@ -201,7 +201,8 @@ export default function AnalysisQuestionEditor(props: AnalysisQuestionEditorProp
 
       {qaDefinition.additionalFieldNames && (
         // Hard coded left padding to account for the 32px icon size + 8px gap
-        <Stack pl={'40px'}>
+        // 0px gap because the children still did not get a mantine refactor so we must respect existing styles
+        <Stack pl={'40px'} gap={'0px'}>
           {question.type === 'qual_auto_keyword_count' && (
             <KeywordSearchFieldsEditor
               questionUuid={question.uuid}
--- a/kobo/apps/hook/serializers/v2/hook.py
+++ b/kobo/apps/hook/serializers/v2/hook.py
@@ -23,11 +23,12 @@ from kpi.utils.schema_extensions.fields import (
 
 class HookSerializer(serializers.ModelSerializer):
 
-    payload_template = serializers.CharField(required=False, allow_blank=True,
-                                             allow_null=True)
+    payload_template = serializers.CharField(
+        required=False, allow_blank=True, allow_null=True
+    )
 
-    settings = JSONFieldWithSchemaField(SettingsField)
-    subset_fields = ListFieldWithSchemaField(SubsetFieldsField)
+    settings = JSONFieldWithSchemaField(SettingsField, required=False)
+    subset_fields = ListFieldWithSchemaField(SubsetFieldsField, required=False)
 
     class Meta:
         model = Hook
--- a/kobo/apps/hook/tests/test_api_hook.py
+++ b/kobo/apps/hook/tests/test_api_hook.py
@@ -63,6 +63,9 @@ class ApiHookTestCase(HookTestCase):
     def test_create_hook(self):
         self._create_hook()
 
+    def test_create_minimal_hook(self):
+        self._create_hook(minimal=True)
+
     def test_editor_access(self):
         hook = self._create_hook()
 
--- a/kobo/apps/hook/utils/tests/mixins.py
+++ b/kobo/apps/hook/utils/tests/mixins.py
@@ -28,17 +28,22 @@ class HookTestCaseMixin:
         self._add_submissions()
 
         url = reverse('hook-list', args=(self.asset.uid,))
+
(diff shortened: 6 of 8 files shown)


jsapp/js/components/processing/analysis/responseForms/responseWrapper.component.tsx
```diff
@@ -22,6 +22,8 @@ import type { AnalysisQuestionInternal } from '../constants'
 
 interface ResponseWrapperProps {
   uuid: string
+  /** Adds a clear button with the given logic */
+  clearable?: Function
```
Akuukis: don't type as `Function` because you always can better be more specific, e.g. `() => unknown`

Akuukis: `clearable` sounds like a boolean. Better name it as `onClear` to sound more like a function and also to align with Mantine naming.

jsapp/js/components/processing/analysis/responseForms/responseWrapper.component.tsx
```diff
@@ -97,6 +99,12 @@ export default function ResponseWrapper(props: ResponseWrapperProps) {
     }
   }
 
+  function handleClear() {
+    if (typeof props.clearable === 'function') {
+      props.clearable()
+    }
```
Akuukis: Why runtime-check for a function, it's already typed as function. `props?.clearable()` would be much cleaner

jsapp/js/components/processing/analysis/responseForms/responseWrapper.component.tsx
```diff
@@ -97,6 +99,12 @@ export default function ResponseWrapper(props: ResponseWrapperProps) {
...
+  }
```
Akuukis: consider to add `event.stopPropagation()`

jsapp/js/components/processing/analysis/responseForms/responseWrapper.component.tsx
```diff
@@ -156,6 +163,10 @@ export default function ResponseWrapper(props: ResponseWrapperProps) {
           onClick={open}
           disabled={!hasManagePermissionsToCurrentAsset() || analysisQuestions.state.isPending}
         />
+
+        {props.clearable && (
+          <ActionIcon variant='light-gray' color='gray' size='sm' iconName='close' onClick={handleClear} />
+        )}
```
Akuukis: don't specify `color`, `variant` should imply a `color` already.

jsapp/js/theme/kobo/ActionIcon.ts
```diff
@@ -18,6 +18,11 @@ export const ActionIconThemeKobo = ActionIcon.extend({
           '--ai-bg': theme.colors.blue[9],
           '--ai-hover': theme.colors.blue[8],
         }),
+        ...(props.variant === 'light-gray' && {
+          '--ai-color': theme.colors.gray[2],
+          '--ai-bg': theme.colors.gray[7],
+          '--ai-hover': theme.colors.gray[6],
+        }),
```
Akuukis, line 25: why grey? isn't that easy to confuse with disabled?

jsapp/js/components/processing/analysis/responseForms/responseWrapper.component.tsx
Akuukis: why a different `variant` than other icons?

Akuukis (changes requested):

jsapp/js/theme/kobo/ActionIcon.ts
duvld: I thought about asking product, but I think it would be easier to merge something then have their feedback later. Are you suggesting we should just the same blue as the edit? I figured we should differentiate somewhat, but I do agree grey is not the best choice. I'm also not sure about the icon either 🤷‍♂️ 

I'll change it to the same blue as edit, and place it next to the edit icon for now

jsapp/js/components/processing/analysis/responseForms/responseWrapper.component.tsx
duvld: Made the change. Any reason we say `unknown` instead of `void`? If we want to get specific, wouldn't `void` imply that the responses shouldn't return anything and make the parameter clearer?

jsapp/js/components/processing/analysis/responseForms/responseWrapper.component.tsx
duvld: I was looking at other places where a function needed to be passed to a wrapper, and we do it like this. I figured it's safer than checking for existence. Though it may be a holdover from converting `es6` to typescript, as now we have a type specified in the prop already, checking for existence should be good enough. Made the change

jsapp/js/components/processing/analysis/responseForms/responseWrapper.component.tsx
duvld: Not sure why I'd need this? No other buttons have this on their listeners, and the parent element does not have any event listeners to propagate to

jsapp/js/components/processing/analysis/responseForms/responseWrapper.component.tsx
duvld: same as above

jsapp/js/components/processing/analysis/responseForms/responseWrapper.component.tsx
Akuukis: Actually you're right, let's use `void`. Note that you *can* return a value where return value is typed `void`, it's just to indicate that it won't matter.

jsapp/js/components/processing/analysis/responseForms/responseWrapper.component.tsx
```diff
@@ -149,6 +156,8 @@ export default function ResponseWrapper(props: ResponseWrapperProps) {
           }
         />
 
+        {props.onClear && <ActionIcon variant='light' size='sm' iconName='close' onClick={handleClear} />}
```
Akuukis: same as but simpler:
```suggestion
        {props.onClear && <ActionIcon variant='light' size='sm' iconName='close' onClick={props?.onClear} />}
```

duvld: The beginning of this line negates the need for the `?` I think, made the changes

Akuukis: Ok! But why didn't you also delete the now unused `handleClear`?

Akuukis (changes requested):
replied last comment

jsapp/js/components/processing/analysis/responseForms/responseWrapper.component.tsx
duvld: Good catch, removed

PR #6448, opened 2025-11-05
test(subsequences): port old unit tests DEV-1231

rgraber:
### 💭 Notes
Port selected tests over from `test_submission_extras_api_post.py` in the old app. Which tests to port were chosen based on whether or not they applied in the new format and if they already had a reasonable equivalent, which many did.
Also cleans up some old tests that were failing or had misleading comments.

### DIFF
--- a/kobo/apps/subsequences/tests/test_submission_stream.py
+++ b/kobo/apps/subsequences/tests/test_submission_stream.py
@@ -6,7 +6,7 @@ from django.test import TestCase
 
 from kobo.apps.openrosa.apps.logger.exceptions import ConflictingSubmissionUUIDError
 from kobo.apps.subsequences.constants import SUPPLEMENT_KEY
-from kobo.apps.subsequences.models import SubmissionSupplement, QuestionAdvancedFeature
+from kobo.apps.subsequences.models import QuestionAdvancedFeature, SubmissionSupplement
 from kobo.apps.subsequences.utils.supplement_data import stream_with_supplements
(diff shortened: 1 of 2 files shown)


kobo/apps/subsequences/tests/test_submission_stream.py
```diff
@@ -0,0 +1,189 @@
+import uuid
+from unittest.mock import patch
+
+from django.contrib.auth import get_user_model
+from django.test import TestCase
+
+from kobo.apps.openrosa.apps.logger.exceptions import ConflictingSubmissionUUIDError
+from kobo.apps.subsequences.constants import SUPPLEMENT_KEY
+from kobo.apps.subsequences.models import SubmissionExtras, QuestionAdvancedFeature
```
noliveleger: `SubmissionExtras` is deprecated, let's use `SubmissionSupplement` instead

kobo/apps/subsequences/tests/test_submission_stream.py
```diff
@@ -0,0 +1,189 @@
...
+        owner = get_user_model().objects.create(username='nlp_owner')
+
+        self.asset = Asset.objects.create(
+            owner=owner,
+            content={
+                'schema': '1',
+                'survey': [
+                    {
+                        'type': 'text',
+                        '$kuid': 'rc9ak31',
```
noliveleger: let's keep the survey to its strict minimum. `$kuid` and `$autoname` are not needed.

PR #6449, opened 2025-11-05
fix(openrosa): sync anonymous submission permission during deployment

noliveleger:
### 📣 Summary
Ensure anonymous submission permission is correctly propagated to OpenRosa so Enketo doesn’t ask for login when anonymous collection is enabled.

### 📖 Description
The issue occurred because the anonymous submission permission was not being copied to OpenRosa during form deployment. As a result, even though the KPI interface showed that anonymous submissions were allowed, the corresponding permission was missing in OpenRosa, causing Enketo to prompt for authentication.

### 👀 Preview steps

1. ℹ️ Configure a project in KPI with anonymous submissions enabled.
2. Deploy the form.
3. 🔴 [on release branch] open the Enketo link and notice it still asks for authentication.
4. 🟢 [on PR] open the same Enketo link and notice it opens directly without login.


Guitlle (approved):
LGTM, and it works as expected

PR #6451, opened 2025-11-07
refactor(connectedProjects): migrate dataAttachments files to TypeScript DEV-1240

magicznyleszek:
### 💭 Notes

Migrated `jsapp/js/components/dataAttachments` code to TS with some `dataInerfaces` and `actions` typings. Didn't want to touch too much there, given we have Orval now.

### 👀 Preview steps

1. ℹ️ Have an account and two deployed projects: A and B
2. In project A go to Settings → Connected Projects
3. Toggle on "Data sharing"
4. In project B go to Settings → Connected Projects
5. Toggle on "Data sharing"
6. In section "Import other project data" select project A and click "Import"
7. In modal choose "Select all" and click "Accept"
8. 🟢 Notice that everything still works

### DIFF
--- a/jsapp/js/components/dataAttachments/dataAttachmentColumnsForm.tsx
+++ b/jsapp/js/components/dataAttachments/dataAttachmentColumnsForm.tsx
@@ -221,7 +221,7 @@ class DataAttachmentColumnsForm extends React.Component<
         <MultiCheckbox
           type='frame'
           items={this.state.columnsToDisplay}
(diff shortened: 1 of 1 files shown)


Akuukis (approved):
LGTM, only behavior change noticed are
- added explicit fallback values in many places
- more `.bind(this)`

Not tested.

Question: why use `||` instead of more strict `??` in many places?

duvld:
Tested, LGTM

PR #6452, opened 2025-11-07
fix(accounts): retain uid for placeholder users DEV-1243

rgraber:
### 📣 Summary
Retain a user's uid when they are set to inactive (ie trashed with a placeholder).

### 👀 Preview steps

2. Generate PH logs as the user (edit the project, add submissions, etc.)
3. Get the user uid
4. Move the user to the trash (remove rather than delete)
5. Empty the trash
6. In the python shell re-fetch the user from the db
7. 🔴 [on release] notice the uid has changed
8. 🟢 [on PR] notice the uid is the same
9. Go to `/api/v2/project-history-logs`
10. 🔴 [on release] user url and username are still present in the logs
8. 🟢 [on PR] user and username are null

### DIFF
--- a/kobo/apps/trash_bin/utils/account.py
+++ b/kobo/apps/trash_bin/utils/account.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from django.conf import settings
 from django.contrib.auth import get_user_model
-from django.db import models, transaction
+from django.db import transaction
 from django.db.models.signals import post_delete
 from django.utils import timezone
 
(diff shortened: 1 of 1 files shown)


kobo/apps/trash_bin/utils/account.py
```diff
@@ -130,22 +129,21 @@ def _replace_user_with_placeholder(
...
+            if retain_audit_logs:
+                # prevent the delete() call from touching the audit logs
+                audit_log_user_field.on_delete = models.DO_NOTHING
             # …and cause a FK violation!
             user.delete()
             # then resolve the violation by creating the placeholder with the
             # same PK as the original user
             placeholder_user.save()
+            placeholder_user.extra_details.uid = uid
+            placeholder_user.extra_details.save()
```
noliveleger: `_replace_user_with_placeholder()` is called [here] (https://github.com/kobotoolbox/kpi/blob/4805001a1fe3cb94269cc12845a8d9b6e62cc4d2/kobo/apps/trash_bin/utils/account.py#L89), where we also save "extra_details".  Could we refactor this to avoid double ExtraUserDetail `save()`?

PR #6462, opened 2025-11-11
refactor(analysis): migrate select one form to mantine DEV-1169

duvld:
### 🗒️ Checklist

1. [x] run linter locally

### 💭 Notes
The "best" solution would probably be to refactor all of `radio.tsx` to mantine, but that creates a bigger problem than this scope. Our `radio.tsx` (not using mantine) has the ability to de-select options, if we convert it to use mantine `Radio`, then there will be a regression to this feature. To offset this for the qualitative analysis project we introduced a clear button--this has not been done everywhere a `radio.tsx` is being used. Further investigation into this is needed, but out of scope of this PR

### 👀 Preview steps
1. ℹ️ have an account and a project with a submission that can open NLP
2. go to qualitative analysis and create a select one question
4. 🔴 [on main] notice that select options are de-selectable
6. 🟢 [on PR] notice that select options are not de-selectable and the clear button works as intended
8. 🟢 notice that all other functionality is preserved


jsapp/js/components/processing/analysis/responseForms/selectOneResponseForm.component.tsx
```diff
@@ -72,14 +72,16 @@ export default function SelectOneResponseForm(props: SelectOneResponseFormProps)
...
+      <Stack gap={'xs'}>
+        {getOptions().map((option) => (
+          <Radio
+            value={option.value}
+            label={option.label}
+            onChange={(newResponse) => onRadioChange(newResponse.currentTarget.value)}
+            checked={response === option.value}
+          />
+        ))}
+      </Stack>
```
Akuukis: please wrap in `Radio.Group` for better semantics, see [Mantine](https://mantine.dev/core/radio/#controlled-radiogroup) and [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/radiogroup_role)

Akuukis (changes requested):

jsapp/js/components/processing/analysis/responseForms/selectOneResponseForm.component.tsx
```diff
@@ -72,14 +72,16 @@ export default function SelectOneResponseForm(props: SelectOneResponseFormProps)
...
-        options={getOptions()}
-        name={question.labels._default}
-        onChange={onRadioChange}
-        selected={response}
-        isClearable
-        isDisabled={!props.canEdit}
-      />
+      <Stack gap={'xs'}>
+        {getOptions().map((option) => (
+          <Radio
```
magicznyleszek: I noticed that Radio doesn't have pointer cursor

jsapp/js/components/processing/analysis/responseForms/selectOneResponseForm.component.tsx
```diff
@@ -72,14 +72,16 @@ export default function SelectOneResponseForm(props: SelectOneResponseFormProps)
 
   return (
     <ResponseWrapper uuid={props.uuid} onClear={() => setResponse('')}>
```
magicznyleszek, line 74: I know this is in another PR, but the clear button placement was a bit confusing to me. There are two question definition related buttons (edit and delete) and between them is a question response related button (clear). Maybe @tesster7 should help us here?

I would expect to have tooltips there at least to make it less confusing :)

The UX is probably better if you're not project manager, because you only have clear.

duvld: I think we can go ahead with the placement for now, and create a ticket in the future when Tess finishes her thoughts/designs :)

duvld: Also this PR is not related to the clear button, it's a bit confusing but it's branched off of the clear button PR here: https://github.com/kobotoolbox/kpi/pull/6445

Akuukis (approved):
code LGTM, not tested

PR #6464, opened 2025-11-11
feat(formBuilder): unset image max-pixels param by clearing the text field DEV-1067

p2edwards:
### 📣 Summary

You can now remove the max-pixels parameter from an image question, by clearing the contents of the max-pixels field  in the image question settings.

### 📖 Description

Adds a new parameter type `{ paramType: 'maxPixels' }` to the Formbuilder, hardcoded for this specific use case. It displays a 'No limit' placeholder if the field is currently blank, and it shows some text hints underneath the field to make note of the default value and explain how to remove the limit.

> max-pixels paramType widget

It introduces three new translation strings:

```
No limit
Default: ##
Leave empty for no limit.
```

### 👀 Preview steps

1. ℹ️ Have an account
2. Create a new form in Form Builder
3. Add `image` question "image 1"
    -  **Do NOT open question settings for "image 1"**.
    - *We'll check this later in the XLS form download - expecting the default max-pixels to be set.*
4. Add another `image` question "image 2"
    - **DO** open settings for "image 2"
    - 🔴 [on main] notice that the 2nd image `max-pixels` has a default set, and is rendered using a "number" widget
    - 🟢 [on PR] notice that the 2nd image `max-pixels` has the default set, and is rendered with the new "maxPixels" widget. Notice that the hint text displays the default, too.
5. **Erase** the max-pixels value for "image 2" so that it is clear. (on PR, the placeholder will read "No limit") 
    - *We'll check this later in the XLS form download - expecting no max-pixels to be set.*
6. Add another `image` question "image 3"
    - **DO** open settings for "image 3"
    - Change the value of max-pixels for "image 3" from the default, to a number of your choosing (example: `640`)
7. 💾 Save the form.
8. Go to Project → Form page (`/landing`)
9. ⬇️ Use "Download XLS" from top right (...) menu
10. 🔴 [on main] notice that the downloaded file has:
    - :+1: `max-pixels=2048` (default) for image 1, 
    - 🔴 **`max-pixels=2048` (default) for image 2**,  (cell should be blank `   `). 
    - :+1: `max-pixels=640` (custom number) for image 3
11. 🟢 [on PR] notice that the downloaded file has: 
    - :+1: `max-pixels=2048` (default) for image 1, 
    - 🟢 **`               ` (blank) for image 2**
    - :+1: `max-pixels=640` (custom number) for image 3
      image
12. Open the form in the formbuilder again, 💾 "Save" and ⬇️ download XLS again.
    - 🟢 [on PR] ensure that the downloaded file **still** has the correct values set for max-pixels (including the **blank**, not the default, for image 2)

### DIFF
--- a/jsapp/xlform/src/view.params.coffee
+++ b/jsapp/xlform/src/view.params.coffee
@@ -54,9 +54,11 @@ module.exports = do ->
     }
 
     initialize: (@paramName, @paramType, @paramDefault, @paramValue='', @onParamChange) ->
-      if @paramValue is '' and typeof @paramDefault isnt 'undefined'
-        # TODO: verify that this doesn't interfere with max-pixels
-        # make sure that params without values use default one
+      # if a parameter is unset, initialize it to the parameter's default, if there is one
+      if  @paramValue is '' and
+          typeof @paramDefault isnt 'undefined' and
+          # exception: for max-pixels, a blank value should be preserved as unset
+          @paramType isnt 'maxPixels'
         @onParamChange(@paramName, @paramDefault)
       return
 


jsapp/xlform/src/model.configs.coffee
```diff
@@ -215,8 +216,8 @@ module.exports = do ->
     }
     image: {
       'max-pixels': {
-        type: configs.paramTypes.number
-        defaultValue: 1024
+        type: configs.paramTypes.maxPixels
+        defaultValue: 2048
```
p2edwards, line 220: Default changed (per discussion)

jsapp/xlform/src/view.params.coffee
```diff
@@ -73,6 +75,11 @@ module.exports = do ->
           val = "#{@paramDefault}"
       else if @paramType is $configs.paramTypes.boolean
         val = evt.currentTarget.checked
+      else if @paramType is $configs.paramTypes.maxPixels
+        if evt.currentTarget.value is ' ' then evt.currentTarget.value = ''
+        val = +evt.currentTarget.value
+        if !val then val = undefined   # +'', +' ' or +'0' unsets max-pixels
```
p2edwards, line 83: Together with `maxPixelsParam(…)` in `view.params.templates.coffee`, this implements a widget that accepts digits as input, treats blank input as an undefined value, and also treats a `' '` input as undefined and trims the text field to display the placeholder again.

jsapp/xlform/src/view.params.templates.coffee
```diff
@@ -30,7 +32,57 @@ module.exports = do ->
...
+          value='#{_.escape currentValue}'
+          inputmode='numeric'
+          pattern='\\d{0,5}'
+          placeholder='#{_.escape t('No limit')}'
+          list=#{suggest_list_id}
+        /><span style='font-weight:normal'> px</span>
+        <p #{style}>#{hints[0]}<br>#{hints[1]}</p>
+        #{suggestions_list}
+      </label>
+      """
```
p2edwards, line 82: I wanted to make it … 

- decently usable
- …reminiscent of KoboCollect's max-pixels pop-up
  - KoboCollect max-pixels selector
- … while allowing completely **custom number** or **blank** input to clear it,
- … with minimal code,
  - … especially not having to maintain any widget state (in Backbone or React)

So here we go: 

> Pseudo-combobox

The `datalist` gives us a sort of free typeahead combo box.
- The no-limit suggestion is `<option value=" "` instead of `value=""` (only because `value=""` isn't allowed.)
- When "No limit" is clicked, we replace `' '` with `''` to reveal the "No limit" placeholder again.


Compared to the number type:

- The **number** type shows the default value in the placeholder text.
- The **max-pixels** type shows "No limit" in the placeholder text.

For both types, what you see in the placeholder text is the result of a blank input if you save the form.

The default **2048px** is shown on a hint line, since it can't be in the placeholder anymore. The rest of the text and suggestions are meant to make leaving the field empty and filling it with numbers feel like real options.

p2edwards: I used some inline styles here,

- partly to keep this PR small and self-contained, 
- and partly to save time.

Let me know if you think they should be a proper integrated class instead.

jsapp/xlform/src/model.configs.coffee
magicznyleszek: After changing this you need to update the preview steps in PR description (step 11) as it uses 1024 still.

jsapp/xlform/src/view.templates.coffee
```diff
@@ -17,9 +17,10 @@ module.exports = do ->
     rowSelector: rowSelector_templates
     surveyApp: surveyApp_templates
 
-  templates['AcceptedFilesView.input'] = accepted_files_templates.acceptedFilesInput
-  templates['ParamsView.numberParam'] = params_templates.numberParam
-  templates['ParamsView.booleanParam'] = params_templates.booleanParam
+  templates['AcceptedFilesView.input']      = accepted_files_templates.acceptedFilesInput
+  templates['ParamsView.numberParam']       = params_templates.numberParam
+  templates['ParamsView.booleanParam']      = params_templates.booleanParam
+  templates['ParamsView.maxPixelsParam']    = params_templates.maxPixelsParam
```
magicznyleszek, line 23: I know you like to indent things visually and I'm not going to complain about this here, but as soon as we turn this into JS (and then TS) all that will be unified and you will be forced to no longer do this :P

PR #6471, opened 2025-11-13
ci(toolVersions): set kubectl and helm version from github environment variable 

ScottMillard:
### 💭 Notes
Helm 4 was released yesterday, which breaks `latest` for helm.  This updates the nonprod deployment pipelines to utilize a github actions environment variable so that all pipelines use the correct version.


noliveleger (approved):
LGTM.
@ScottMillard I think you should backport this PR to the current release though.

PR #6472, opened 2025-11-14
fix(data): update `/api/v2/data/` limits DEV-1278

RuthTurk:
### 📣 Summary
Change maximum number of results that `/api/v2/data/` can return to 100 by default and 1000 maximum.

### 👀 Preview steps

2. add over 100 submissions to the project
3. Checkout the endpoint: `/api/v2/assets/{uid_asset}/data/`
4. 🔴 [on main] the response body contains a results array with all the items
5. 🟢 [on PR] the response body contains a results array with exactly 100 items
6. Test that you can successfully request up to 1000 records, but no more
7. Test that the offset parameter works correctly with the new limits

### DIFF
--- a/kpi/views/v2/data.py
+++ b/kpi/views/v2/data.py
@@ -667,12 +667,14 @@ class DataViewSet(
 
         # Remove `format` from filters. No need to use it
         filters.pop('format', None)
-        # Do not allow requests to retrieve more than `SUBMISSION_LIST_LIMIT`
+        # Do not allow requests to retrieve more than `max_limit`
         # submissions at one time only if a limit is explicitly defined.
         if 'limit' in filters:
             try:
                 filters['limit'] = positive_int(
-                    filters['limit'], strict=True, cutoff=settings.SUBMISSION_LIST_LIMIT
+                    filters['limit'],
+                    strict=True,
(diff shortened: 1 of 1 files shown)


noliveleger (changes requested):
You need to change the behaviour in the data viewset.
If no limit is specified, it will always take the max limit while we want to return the default limit. The max limit should only cap a specified limit.

kpi/views/v2/data.py
```diff
@@ -668,16 +668,19 @@ def _filter_mongo_query(self, request):
...
+                filters['limit'] = positive_int(
+                    filters['limit'], strict=True, cutoff=settings.SUBMISSION_LIST_LIMIT
+                )
+            except ValueError:
+                raise serializers.ValidationError(
+                    {'limit': t('A positive integer is required')}
+                )
+        else:
+            # If no limit is specified, use the default limit (100)
+            filters['limit'] = 100
```
noliveleger: I would use `self.pagination_class.default_limit` instead. To avoid redundancy. 
We could even use `self.pagination_class.max_limit` instead of `settings.SUBMISSION_LIST_LIMIT`. What do you think?

noliveleger (changes requested):

PR #6486, opened 2025-11-19
docs(hsts): note that production overrides Django config INFRA-297

ScottMillard:
### 💭 Notes
The HSTS headers that django can return are sometimes removed/overridden by other infrastructure.  This adds a comment around that fact to help anyone in the future that may be troubleshooting why the headers don't match what is expected.


jnm:
Failing CI due to `kobo/settings/base.py:48:89: E501 line too long (132 > 88 characters) [flake8]`

PR #6487, opened 2025-11-19
perf(assets): prefetch related XForm objects to reduce database queries DEV-880

noliveleger:
### 📣 Summary
Improve performance of asset queries by prefetching related XForm objects.

### 📖 Description
This change optimizes asset retrieval by prefetching related XForm objects, reducing the number of database queries executed when listing or accessing assets. Previously, each access to related XForm data triggered additional queries, which impacted performance on endpoints returning multiple assets. With prefetching, related data is loaded efficiently in a single query, improving response times and lowering database load without altering API behavior or response content.

### Notes
In AssetViewSet, for the list action, we preload several related objects to avoid issuing queries for each individual asset. However, we were previously using the full list of asset IDs to load those relations, which meant we were unnecessarily pulling into memory objects that were not part of the current page.
For example: if the current page displays assets a, b, c but the full result set contains a, b, c, d, e, f, we were still loading all related objects for a, b, c, d, e, f on every page.
This PR now loads only the objects related to a, b, c.

blocked by #6488

### 👀 Preview steps

1. Create multiple projects
2. Deploy them
3. Activate DEBUG_TOOLBAR (return True in `show_toolbar` in dev.py)
4. 🔴 [on release branch] note the number of queries
5. 🟢 [on PR] The number of queries should have decreased by an amount equal to the number of deployed projects.

### DIFF
--- a/kobo/apps/accounts/adapter.py
+++ b/kobo/apps/accounts/adapter.py
@@ -2,14 +2,8 @@ from allauth.account.adapter import DefaultAccountAdapter
 from allauth.account.forms import SignupForm
 from constance import config
 from django.conf import settings
-from django.contrib.auth import REDIRECT_FIELD_NAME
 from django.db import transaction
 from django.utils import timezone
-
-from .mfa.forms import MfaTokenForm
-from .mfa.permissions import mfa_allowed_for_user
-from .mfa.views import MfaTokenView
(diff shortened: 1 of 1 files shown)


RuthTurk (approved):
LGTM

PR #6489, opened 2025-11-20
perf(auditLogs): remove `count` from API response

noliveleger:
### 📣 Summary
Improve audit log performance by removing the expensive count calculation from the response.

### 📖 Description
This change removes the `count` field from the audit logs API response because calculating the total number of documents is a major bottleneck on large datasets. The count query requires scanning the entire collection, significantly slowing down the endpoint. Clients still receive the paginated results as before, just without the heavy count operation.

### 👀 Preview steps

1. Log in as a superuser
2. Try to create a lot of activity (e.g. use mockobo to create 1000+ submissions)
3. Go to `/api/v2/audit-logs/` (or use `curl`) 
4. 🟢 [on PR] notice that `count` is not there anymore and page loads faster. Unfortunately, with less than 1 000 000 records, it's barely noticeable.

### DIFF
--- a/kobo/apps/audit_log/tests/api/v2/test_api_audit_log.py
+++ b/kobo/apps/audit_log/tests/api/v2/test_api_audit_log.py
@@ -71,7 +71,7 @@ class ProjectHistoryLogTestCaseMixin:
             date_created=now,
         )
         response = self.client.get(self.url)
-        self.assertEqual(response.data['count'], 1)
+        self.assertEqual(len(response.data['results']), 1)
         ph_log = response.data['results'][0]
         self.assertListEqual(
             sorted(list(ph_log.keys())),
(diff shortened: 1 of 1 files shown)


jsapp/js/api/models/paginatedAccessLogResponseList.ts
```diff
@@ -12,7 +12,6 @@ The endpoints are grouped by area of intended use. Each category contains relate
 import type { AccessLogResponse } from './accessLogResponse'
 
 export interface PaginatedAccessLogResponseList {
-  count: number
```
jnm, line 15: The Orval stuff is a new world to me; was this change indeed generated automatically? Wondering because I can't tell (yet) if this response is used for both "all" access logs (superuser endpoint) and per-user (normal user) access logs

noliveleger: It was! 

DRF Spectacular generates a paginated response schema based on the viewset’s paginator, but the component name is derived from the item serializer name, not the paginator. Because both viewsets used the same serializer name, Spectacular produced a single shared component (`PaginatedAccessLogResponseList`) and reused it, even though one endpoint uses a pagination schema without count. Orval then generated identical response types since it depends on this shared (and incorrect) component definition.

( 🤖 )

PR #6498, opened 2025-11-24
build(docker): build node app in a separate docker build stage DEV-1343

p2edwards:

Optimize the docker image build. Compile the node app in a separate stage, then copy the production assets to the final production image.

### 💭 Notes

- 📣📖👷 **This is expected to be a drop-in change for production and dev.** 
  - It should have no effect on development process except to make kpi image builds faster.
  - It should have no immediate effect on production builds. Soon the images can be smaller with an absence of `node_modules` in the final image (but I am leaving it in for now to support transition of developer scripts).
- 💭 **Files changed**
  - Port `copy_fonts.py` to `.sh` so it can be run w/o Python
  - Move `npm install` and the webpack prod build in Docker 'build stage'
  - Rearrange some `Dockerfile` sections to optimize layer-based caching.
- 💭 **Intended effects**
  - KPI docker image rebuilds will be faster in most development scenarios.
    - Full `--no-cache` rebuilds should be faster than before, since they allow the three "slow parts" to happen simultaneously (apt updates, pip dependencies, and now node stuff).
    - Docker image layers for node modules / webpack build steps will be reused (take 0.0-5.0s) if their package dependencies or build inputs haven't changed.
  - ~~Removal of `node_modules` — no longer copied to the final Docker image~~
    - **Edit**: I've added it back temporarily. (bottom of Dockerfile) In case there are some more changes coming soon, I don't want to ask people to update their scripts again. When we're ready to reduce the docker image size by ~600 MB, we can comment / delete these lines.

### 🏁 Special testing

Tested:

- [x] ✅ **kobo-compose**
- [x] ✅ **kobo-install**, **npm running on "host"**, debug mode on
  - [x] webpack production build (static build)
  - [x] webpack dev-server (`npm run watch`)
- [x] ✅ **kobo-install**, **npm running in container**, debug mode on
  - [x] webpack production build (static build) in container
  - [x] webpack dev-server (HMR / `npm run watch` / localhost:3000) in container
    - ⚠️ If you're running `watch` this way, you may need to (eventually) update your script:
      ```diff
         ./run.py -cf run --rm --publish 3000:3000 kpi \
      -      npm run watch
      +      ./docker/webpack-dev-server.sh
      ```
      🟡 `npm run watch` relies on `npm_modules` already being present upon container creation.
      🟢 The new `./docker/webpack-dev-server.sh` will install them if missing.
     - Also note that if **debug mode** is on, the changes should pick up right away without needing to restart the Django web server.
- [x] ✅ **kobo-install**, **production-like build**
  - [x] 🏁 Tested by this method.
      1. Start kobo-install with the answers "workstation: yes", "advanced: no", which pulls the image from docker hub. Let it run.
      2. Take it down with `kobo-install/run.py -cf down && kobo-install/run.py -cb down`
      3. Edit `kobo-docker/docker-compose.frontend.override.yml` like this:
         ```diff
           kpi:
         +   build: ../kpi
           …
           worker:
         +   build: ../kpi
           …
         ```
         and likewise for the rest: 'worker worker_kobocat worker_low_priority worker_long_running_tasks beat'
      4. Rebuild: 
         ```sh
         ./run.py -cf build kpi && \
             ./run.py -cf build \
               worker worker_kobocat worker_low_priority \
               worker_long_running_tasks beat
         ```
      5. Re-up: `./run.py -cb up -d && ./run.py -cf -d`, `./run.py -cf logs -f`
      6. Test in the browser at kf.kobo.local.
- [x] ✅ **"du" deployment** - deploys successfully.

### DIFF
--- a/.github/workflows/nonprod-releases.yml
+++ b/.github/workflows/nonprod-releases.yml
@@ -121,4 +121,4 @@ jobs:
     secrets: inherit
     with:
       topic: "deployments to ${{ github.ref_name }}"
-      content: ":warning: ~~@_*devs*~~ @**Phil Edwards**, your [run #${{github.run_number}}](${{github.server_url}}/${{github.repository}}/actions/runs/${{github.run_id}}) failed on `${{ github.ref_name }}` at [\"$COMMIT_TITLE\"](${{ github.event.compare }})"
+      content: ":warning: @*devs*, [run #${{github.run_number}}](${{github.server_url}}/${{github.repository}}/actions/runs/${{github.run_id}}) failed on `${{ github.ref_name }}` at [\"$COMMIT_TITLE\"](${{ github.event.compare }})"


Dockerfile
```diff
@@ -1,55 +1,152 @@
...
+# that would affect the outcome of 'npm clean-install'.
+COPY --chown=node:node --parents \
+    patches/                  \
+    scripts/copy_fonts.sh     \
+    scripts/generate_icons.js \
+    scripts/hints.js          \
+    .browserslistrc    \
+    package.json       \
+    package-lock.json  \
+    .
```
jnm, line 51: Possible concern about maintenance burden; will it be obvious that someone needs to add something here when a failure occurs?

Dockerfile
```diff
@@ -1,55 +1,152 @@
...
+# and clean the cache for space.
+USER node
+RUN npm clean-install \
+    && npm cache clean --force
+
+# Results in /srv/src/kpi/:
+#   All the sources copied above, plus the generated:
+#   + jsapp/fonts/
+#   + msw-mocks/
+#   + node_modules/
```
jnm, line 63: Idea that the first `COPY` below in the `webpack-build-prod` stage could be simplified to only these?

Dockerfile
```diff
@@ -1,55 +1,152 @@
...
+    .
+
+# Copy the rest of the webpack build inputs from the
+# build context.
+COPY --chown=node:node --parents \
+    jsapp/img/      \
+    jsapp/js/       \
+    jsapp/scss/     \
+    jsapp/xlform/   \
+    jsapp/postcss.config.cjs \
```
jnm: Phil says maybe this could be simplified to `jsapp/` if the `.dockerignore` is applying to the build context

Dockerfile
```diff
@@ -73,75 +170,82 @@ RUN apt-get -qq update && \
...
+RUN mkdir -p "${TMP_DIR}/.npm" && \
     npm config set cache "${TMP_DIR}/.npm" --global && \
-    # to pin an exact npm version, see 'git blame' here
     npm install -g check-dependencies@1 && \
-    rm -rf "${KPI_SRC_DIR}/jsapp/fonts" && \
-    rm -rf "${KPI_SRC_DIR}/jsapp/compiled" && \
-    npm install --quiet && \
     npm cache clean --force
 
-ENV PATH=$PATH:${KPI_NODE_PATH}/.bin
```
jnm, line 118: Phil: should be restored

Dockerfile
```diff
@@ -1,55 +1,152 @@
...
+ENV TMP_DIR=/srv/tmp \
+    VIRTUAL_ENV=/opt/venv
 RUN python -m venv "$VIRTUAL_ENV"
 ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 COPY ./dependencies/pip/requirements.txt "${TMP_DIR}/pip_dependencies.txt"
 RUN uv pip sync "${TMP_DIR}/pip_dependencies.txt" 1>/dev/null
 
+###########################################
+#                                         #
+# 🐍 Python / kpi container (build & run) #
```
jnm: The Python emphasis here confused my small brain about this really being the "final" / "whole" KPI image that contains all Python and JS for the application to run in production.

jnm: "KPI production image"? / `kpi-app`?

Dockerfile
```diff
@@ -73,75 +170,82 @@ RUN apt-get -qq update && \
...
+####################################
+RUN adduser --disabled-password --gecos '' "$UWSGI_USER"
+
+#################################################
+# Set up Node for kobo-docker lifecycle scripts #
+#################################################
+RUN mkdir -p "${TMP_DIR}/.npm" && \
     npm config set cache "${TMP_DIR}/.npm" --global && \
-    # to pin an exact npm version, see 'git blame' here
     npm install -g check-dependencies@1 && \
```
jnm: Used by conditional `npm install` stuff in `entrypoint.sh`; we must test to see if this still works. Ideally, the conditional dependency stuff (including Python) would be removed in favor of layer caching facilitated by this PR, but that might not cover all existing developer use cases

Dockerfile
```diff
@@ -160,6 +259,18 @@ RUN chown -R "${UWSGI_USER}:${UWSGI_GROUP}" ${KPI_SRC_DIR}/emails/ && \
...
     chown -R root:root "${TMP_DIR}/.npm"
 
+# ################################
+# # TMP NOTE, 2026/01/29         #
+# # If you needed to, you could  #
+# # copy node_modules here, too. #
+# ################################
+# COPY --from=npm-install --parents \
+#     /srv/src/kpi/./node_modules/  \
+#     .
```
p2edwards: (note to self: test this comment if you're going to tempt anyone with it)

jnm: We decided to do this by default for now and avoid asking people to change their behavior yet, with the expectation that we'll clean up more grossness in the future

docker/webpack-dev-server.sh
```diff
@@ -0,0 +1,35 @@
...
+# If you've been doing this:
+# 
+#       ./run.py -cf run --rm --publish 3000:3000 kpi \
+#                                    npm run watch && \
+#                            ./run.py -cf restart kpi
+#
+# Or this:
+#
+#       ./run.py -cf run --rm --publish 3000:3000 kpi npm run watch ; \
+#                            docker restart kobofe-kpi-1"
```
jnm: While these commands may appear in [Notion](https://www.notion.so/kobotoolbox/Local-KoboToolbox-w-kobo-install-1137e515f65480438e6af95b838e7e85?source=copy_link#1197e515f6548052b056f1bb6e3464e7), we discussed how they were nonsensical or relied on hidden knowledge like the correct moment to press CTRL+C once before (unmentioned) starting `npm run watch` again.

You mentioned you would ask people for their actual commands and test them against your changes.

PR #6507, opened 2025-11-26
fix(exports): block access to old exports when permissions change DEV-1357

rgraber:
### 📣 Summary
Ensure anonymous users cannot access old export files if permission to view submissions is revoked.

### 💭 Notes
Nest anonymous project exports under the asset uid so we can selectively deny access if the public loses permission to view submissions.

### 👀 Preview steps

2. Allow public access to view submissions
3. Create an export with 
```
curl -X POST http://kf.kobo.local/api/v2/assets/<asset_uid>/exports/ --json '{"fields_from_all_versions":true,"fields":[],"group_sep":"/","hierarchy_in_labels":false,"lang":"_default","multiple_select":"both","type":"xls","query":{},"xls_types_as_text":false,"include_media_url":true}'
```
4. Log out any logged in users
5. Navigate to `/api/v2/assets/<asset_uid>/exports`
6. 🔴 [on release] `"result"` field looks like `http://kf.kobo.local:8080/private-media/AnonymousUser/exports/<filename>.xlsx` 
7. 🟢 [on PR] `"result"` field looks like `http://kf.kobo.local:8080/private-media/AnonymousUser/exports/<asset_uid>/<filename>.xlsx` 
8. Save the url from the result field
9. Log in again and disable public access to view submissions
10. Log out
11. Navigate to the saved result url
12. 🔴 [on release] File downloads
13. 🟢 [on PR] permission error

### DIFF
--- a/kpi/tests/api/v2/test_api_exports.py
+++ b/kpi/tests/api/v2/test_api_exports.py
@@ -147,6 +147,10 @@ class AssetExportTaskTestV2(MockDataExportsBase, BaseTestCase):
         download_response = self.client.get(download_url)
         assert download_response.status_code == status.HTTP_200_OK
 
+        self.asset.remove_perm(anon, PERM_VIEW_SUBMISSIONS)
+        download_response = self.client.get(download_url)
+        assert download_response.status_code == status.HTTP_403_FORBIDDEN
+
     def test_export_task_list_anotheruser(self):
         for _type in ['csv', 'xls', 'spss_labels']:
             self._create_export_task(_type=_type)


kpi/utils/private_storage.py
```diff
@@ -40,6 +44,13 @@ def superuser_or_username_matches_prefix(private_file):
     if private_file.relative_name.startswith(
         '{}/'.format(user.username)
     ):
+        if user.pk == -1:
+            filename_regex = r'AnonymousUser/exports/(a[^/]*)/.*'
+            match = re.search(filename_regex, private_file.relative_name)
+            if match:
+                uid = match.groups()[0]
+                a = Asset.objects.get(uid=uid)
```
noliveleger: I don't think we need to all fields for that. Let's load only needed fields (especially avoid `content`) .

kpi/utils/private_storage.py
```diff
@@ -40,6 +44,13 @@ def superuser_or_username_matches_prefix(private_file):
     if private_file.relative_name.startswith(
         '{}/'.format(user.username)
     ):
+        if user.pk == -1:
```
noliveleger: Let's use the utility function for that `is_user_anonymous()`

kpi/models/import_export_task.py
```diff
@@ -497,6 +497,8 @@ def export_upload_to(self, filename):
     more information, see
     https://docs.djangoproject.com/en/1.8/topics/migrations/#serializing-values
     """
+    if isinstance(self, SubmissionExportTask) and self.user.pk == -1:
+        return posixpath.join(self.user.username, 'exports', self.asset.uid, filename)
```
noliveleger: We want to move all exports under the asset uid, not only for the anonymous user.

kpi/utils/private_storage.py
rgraber: It's a bit tricky to determine exactly which fields we need, maybe just defer `content`?

kpi/models/import_export_task.py
```diff
@@ -74,6 +73,7 @@
 )
 from kpi.utils.strings import to_str
 from kpi.zip_importer import HttpContentParse
+from pyxform.xls2json_backends import xls_to_dict, xlsx_to_dict
```
noliveleger: Restored previous imports. 

`formpack` and `pyxform`  do belong to the second block

kpi/models/import_export_task.py
```diff
@@ -497,6 +497,8 @@ def export_upload_to(self, filename):
     more information, see
     https://docs.djangoproject.com/en/1.8/topics/migrations/#serializing-values
     """
+    if isinstance(self, SubmissionExportTask):
+        return posixpath.join(self.user.username, 'exports', self.asset.uid, filename)
```
noliveleger: Moved everything related to assets in the their own folder.

kpi/utils/private_storage.py
noliveleger: with `shell_plus --print-sql` you can see which fields are using when called `.has_perm()`

PR #6509, opened 2025-11-27
feat(exports): add celery task to delete old anonymous exports DEV-1358

rajpatel24:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Automatically removes old anonymous exports to prevent outdated files from being accessible and to keep storage usage under control.

### 📖 Description
Anonymous exports are now cleaned up automatically after they expire. A new recurring background task deletes export files and their corresponding records once they are older than the value defined in `ANON_EXPORTS_CLEANUP_AGE` (configured in Constance, default: 30 minutes).

To ensure system stability, the cleanup runs in small batches of 50 exports per execution and uses a distributed lock to prevent multiple workers from deleting the same exports simultaneously. This keeps storage usage under control and prevents outdated anonymous export files from remaining accessible.

Authenticated exports and other project data are not affected.

### DIFF
--- a/kobo/settings/base.py
+++ b/kobo/settings/base.py
@@ -477,7 +477,7 @@ CONSTANCE_CONFIG = {
     ),
     'ANONYMOUS_EXPORTS_GRACE_PERIOD': (
         30,
-        'Number of minutes after which anonymous export tasks are cleaned up.'
+        'Number of minutes after which anonymous export tasks are cleaned up.',
     ),
     'LIMIT_ATTACHMENT_REMOVAL_GRACE_PERIOD': (
         90,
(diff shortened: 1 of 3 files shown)


kobo/settings/base.py
```diff
@@ -475,6 +475,10 @@
         'Enable automatic deletion of attachments for users who have exceeded '
         'their storage limits.'
     ),
+    'ANON_EXPORTS_CLEANUP_AGE': (
```
noliveleger: Let's be consistent with other similar variables. e.g.: `*_GRACE_PERIOD`. 
Also I would write `ANONYMOUS_EXPORTS_...` .

PR #6510, opened 2025-11-28
feat(usage): add LLM requests usage card under feature flag DEV-732

duvld:
### 🗒️ Checklist

1. [x] run linter locally
3. [] for user-facing doc changes create a Zulip thread at `#Support Docs Updates`, if any

### 💭 Notes
Updates types to get ready for LLM requests stats from backend. Currently these aren't implemented with any limits/addons, so these are just placeholders for now. We also don't have any LLM in place, so the only way to change the usage is manually changing the numbers, which isn't super useful for now

On the code changes side, I've simplified some of the styling logic with mantine so we no longer rely on JS to check if there is an add on to make the relevant cards vertically responsive 

### 👀 Preview steps
1. ℹ️ have an account
2. got to addons page
3. purchase an addon such that it will show up in the usage cards
4. go to usage page
6. 🟢 [on PR] notice that the cards are responsive horizontally and vertically
8. 🟢 notice that the cards a functionally the same as before (no regressions)

### DIFF
--- a/jsapp/js/components/processing/analysis/responseForms/selectOneResponseForm.component.tsx
+++ b/jsapp/js/components/processing/analysis/responseForms/selectOneResponseForm.component.tsx
@@ -1,5 +1,5 @@
+import { Radio, Stack } from '@mantine/core'
 import React, { useContext, useState } from 'react'
-import Radio from '#/components/common/radio'
 import type { RadioOption } from '#/components/common/radio'
 import AnalysisQuestionsContext from '#/components/processing/analysis/analysisQuestions.context'
 import {
@@ -71,15 +71,19 @@ export default function SelectOneResponseForm(props: SelectOneResponseFormProps)
   }
 
   return (
-    <ResponseWrapper uuid={props.uuid} onClear={() => onRadioChange('')}>
-      <Radio
-        options={getOptions()}
-        name={question.labels._default}
-        onChange={onRadioChange}
-        selected={response}
-        isClearable
-        isDisabled={!props.canEdit}
-      />
+    <ResponseWrapper uuid={props.uuid} onClear={() => setResponse('')}>
+      <Radio.Group>
+        <Stack gap={'xs'}>
+          {getOptions().map((option) => (
+            <Radio
+              value={option.value}
+              label={option.label}
+              onChange={(newResponse) => onRadioChange(newResponse.currentTarget.value)}
+              checked={response === option.value}
+            />
+          ))}
+        </Stack>
+      </Radio.Group>
     </ResponseWrapper>
   )
 }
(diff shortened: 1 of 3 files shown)


jsapp/js/featureFlags.ts
```diff
@@ -5,7 +5,8 @@ import { recordValues } from './utils'
  * For our sanity, use camel case and match key with value.
  */
 export enum FeatureFlag {
-  exampleFeatureEnabled = 'exampleFeatureEnabled', //Comment out when we have active FFs
+  //exampleFeatureEnabled = 'exampleFeatureEnabled', //Comment out when we have active FFs
+  llmUsageCardEnabled = 'llmUsageCardEnabled', //Comment out when we have active FFs
```
jamesrkiger: We'll want to use this for other parts of the auto QA feature, so let's make the name more general: `autoQAEnabled`

jsapp/js/account/usage/usage.component.tsx
```diff
@@ -175,73 +166,82 @@ export default function Usage() {
         )}
       </header>
       {limits.stripeEnabled && <YourPlan />}
-      <div className={styles.row}>
-        <div className={cx(styles.row, styles.subrow)}>
-          <div className={styles.box}>
+      <Group align='stretch'>
+        {/* Hard coded width from the figma designs */}
+        <Paper withBorder p={'md'} w={268}>
```
jamesrkiger: Is there any reason we can't move more of this repeated code into the UsageContainer component? The component could just return this <Paper> element, couldn't it? You could probably push the title text in there, too.

jsapp/js/account/stripe.api.ts
```diff
@@ -14,6 +14,7 @@ const DEFAULT_LIMITS: AccountLimit = Object.freeze({
   asr_seconds_limit: Limits.unlimited,
   mt_characters_limit: Limits.unlimited,
   storage_bytes_limit: Limits.unlimited,
+  llp_requests_limit: Limits.unlimited,
```
jamesrkiger: Typo: should be llm

jsapp/js/account/usage/usage.component.tsx
duvld: It opens a bit of a rabbit hole, we can really do a bigger refactor here. Each of these cards can have their info enumerated and iterated through then mapped, and the styling can be removed in favour for a mantine refactor, so I drew the line somewhere.

That said, I don't think moving the `Paper` wrapper into the `UsageContainer` component is a slippery slope, so I made those changes. I had to import the `usage.module.scss` to the `UsageContainer` component, as I think time would be better spent refactoring to mantine than to mess with the current styles

PR #6513, opened 2025-12-02
fix(exports): remove exports based on the last modification time DEV-1373

rajpatel24:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

For automatic export deletion via the Celery task, consider using the last modification time (e.g., `date_created` + `processing_time_seconds`) to determine when to remove them.


kpi/tasks.py
```diff
@@ -85,33 +85,39 @@ def cleanup_anonymous_exports(**kwargs):
     than `EXPORT_CLEANUP_GRACE_PERIOD`, excluding those that are still processing
     """
     BATCH_SIZE = 200
```
noliveleger, line 89: Make a `BATCH_SIZE`  a common setting to be reuse in similar task.

kpi/tasks.py
```diff
@@ -85,33 +85,39 @@ def cleanup_anonymous_exports(**kwargs):
...
         SubmissionExportTask.objects.filter(
-            user=get_anonymous_user(),
-            date_created__lt=cutoff_time,
+            user=get_anonymous_user()
         )
         .exclude(status=ImportExportStatusChoices.PROCESSING)
-        .order_by('date_created')
-        .values_list('pk', flat=True)[:BATCH_SIZE]
+        .only('pk', 'uid', 'date_created', 'data')
+        .order_by('date_created')[:BATCH_SIZE]
```
noliveleger: As per our discussion, keep the `values_list` and use an annotation to filter the ids at the database level

noliveleger (changes requested):

kpi/tasks.py
rajpatel24: @noliveleger I think this isn’t straightforward because `processing_time_seconds` is stored inside the `data` JSON field, and we can’t use it directly inside `timedelta()` as it doesn’t accept `F()` expressions. That means something like: `timedelta(seconds=F('data__processing_time_seconds'))` will always fail.  

To build this value dynamically in the database, we may need to use `ExpressionWrapper` together with a `DurationField` so the whole calculation is performed at the SQL level rather than in Python. But I’m not completely sure what the correct combination of expressions should look like

noliveleger (approved):
LGTM

PR #6515, opened 2025-12-02
feat(exports): remove synchronous exports older than `EXPORT_CLEANUP_GRACE_PERIOD` DEV-1372

rajpatel24:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Automatically removes old synchronous exports to prevent outdated files from being accessible and to keep storage usage under control.

### 📖 Description
Synchronous exports are now cleaned up automatically after they expire. A new recurring background task `cleanup_synchronous_exports` deletes export files and their corresponding records once they are older than the value defined in `EXPORT_CLEANUP_GRACE_PERIOD` (configured in Constance, default: 30 minutes).

To ensure system stability, the cleanup runs in small batches of 200 exports per execution and uses a distributed lock to prevent multiple workers from deleting the same exports simultaneously. This keeps storage usage under control and prevents outdated export files from remaining accessible.

### DIFF
--- a/kobo/apps/mass_emails/models.py
+++ b/kobo/apps/mass_emails/models.py
@@ -8,6 +8,7 @@ from import_export import fields, resources
 
 from kobo.apps.kobo_auth.shortcuts import User
 from kobo.apps.mass_emails.user_queries import (
+    get_active_users,
     get_all_test_users,
     get_inactive_users,
     get_users_over_80_percent_of_nlp_limits,
@@ -28,6 +29,7 @@ USER_QUERIES: dict[str, Callable] = {
     'users_above_90_percent_storage': get_users_over_90_percent_of_storage_limit,
     'users_above_100_percent_storage': get_users_over_100_percent_of_storage_limit,
     'users_inactive_for_365_days': get_inactive_users,
+    'users_active_within_365_days': get_active_users,
     'users_above_80_percent_submissions': get_users_over_80_percent_of_submission_limit,
     'users_above_90_percent_submissions': get_users_over_90_percent_of_submission_limit,
     'users_above_100_percent_submissions': get_users_over_100_percent_of_submission_limit,  # noqa
--- a/kobo/apps/mass_emails/tests/test_user_activity_queries.py
+++ b/kobo/apps/mass_emails/tests/test_user_activity_queries.py
@@ -7,32 +7,33 @@ from django.utils.timezone import now
 from rest_framework import status
 
 from kobo.apps.kobo_auth.shortcuts import User
-from kobo.apps.mass_emails.user_queries import get_inactive_users
+from kobo.apps.mass_emails.user_queries import get_active_users, get_inactive_users
 from kobo.apps.openrosa.apps.logger.models import Instance
 from kpi.models import Asset
 from kpi.tests.base_test_case import BaseTestCase
+from kpi.utils.object_permission import get_anonymous_user
 
 
 @ddt
-class InactiveUserTest(BaseTestCase):
+class UserActivityQueryTests(BaseTestCase):
     """
-    Tests for identifying inactive users based on login, asset modifications,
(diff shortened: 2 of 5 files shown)


kobo/settings/base.py
```diff
@@ -1456,6 +1456,12 @@ def dj_stripe_request_callback_method():
         'options': {'queue': 'kpi_low_priority_queue'},
     },
     # Schedule every 15 minutes
```
noliveleger: The comment is not accurate ;-) every 5

kpi/tests/test_cleanup_synchronous_exports.py
```diff
@@ -0,0 +1,162 @@
...
+
+class SynchronousExportCleanupTestCase(TransactionTestCase):
+    def setUp(self):
+        self.user = User.objects.create(username='testuser')
+
+    @override_config(EXPORT_CLEANUP_GRACE_PERIOD=30)
+    def test_synchronous_exports_older_than_30_minutes_are_deleted(self):
+        # Export older than 30 min - should be deleted
+        old_export = self._create_synchronous_export(
+            minutes_old=31, asset=self._create_asset('Old Asset')
```
noliveleger: You should use `EXPORT_CLEANUP_GRACE_PERIOD` + 1 and `EXPORT_CLEANUP_GRACE_PERIOD` - 1, to make it less hard coded

kpi/tasks.py
```diff
@@ -134,6 +135,63 @@ def cleanup_anonymous_exports(**kwargs):
     logging.info(f'Cleaned up {deleted_count} old anonymous exports.')
 
 
+@celery_app.task
+def cleanup_synchronous_exports(**kwargs):
+    """
+    Task to clean up old synchronous exports that are older than
+    `EXPORT_CLEANUP_GRACE_PERIOD`, excluding those that are still processing
+    """
+    BATCH_SIZE = 200
```
noliveleger: This could be a common setting

kpi/tasks.py
```diff
@@ -134,6 +135,63 @@ def cleanup_anonymous_exports(**kwargs):
...
+
+    cutoff_time = timezone.now() - timedelta(minutes=export_cleanup_grace_period)
+    old_sync_export_ids = (
+        SubmissionSynchronousExport.objects.filter(
+            date_created__lt=cutoff_time,
+        )
+        .exclude(status=ImportExportStatusChoices.PROCESSING)
+        .order_by('date_created')
+        .values_list('pk', flat=True)[:BATCH_SIZE]
+    )
```
noliveleger: Use the same logic for date_created + processing_time.

It would be great to avoid duplicating the code. Could we have an helper to call which would process the deletion whatever the model is? 

⚠️  There is a special case to handle for anonymous exports.

rajpatel24: Added the common helper function

kpi/tasks.py
```diff
@@ -86,52 +82,28 @@ def cleanup_anonymous_exports(**kwargs):
...
+        config.EXPORT_CLEANUP_GRACE_PERIOD * 60
+        < config.SYNCHRONOUS_EXPORT_CACHE_MAX_AGE
+    ):
+        logging.warning(
+            'Synchronous export cleanup skipped because '
+            'EXPORT_CLEANUP_GRACE_PERIOD is less than '
+            'SYNCHRONOUS_EXPORT_CACHE_MAX_AGE.'
+        )
+        return
+    delete_expired_exports(SubmissionSynchronousExport)
```
noliveleger, line 106: I'd rather see the validation in Constance. We don't want the exports to stay on the storage if the setting is wrong

noliveleger: Too complicated to handle in Constance... Let's just add a warning in Constance.
To be handled by another PR.

PR #6520, opened 2025-12-03
feat(exports): remove audit log exports older than `EXPORT_CLEANUP_GRACE_PERIOD` DEV-1369

rajpatel24:
### 📣 Summary
Automatically removes old audit log exports to prevent outdated files from being accessible and to keep storage usage under control.

### 📖 Description
Audit log exports are now cleaned up automatically after they expire. New recurring background tasks `cleanup_access_log_exports` and `cleanup_project_history_log_exports` deletes export files and their corresponding records once they are older than the value defined in `EXPORT_CLEANUP_GRACE_PERIOD` (configured in Constance, default: 30 minutes).

To ensure system stability, the cleanup runs in small batches of 200 exports per execution and uses a distributed lock to prevent multiple workers from deleting the same exports simultaneously. This keeps storage usage under control and prevents outdated export files from remaining accessible.


kpi/tasks.py
```diff
@@ -177,6 +179,24 @@ def cleanup_project_view_exports(**kwargs):
...
+
+@celery_app.task
+def cleanup_project_history_log_exports(**kwargs):
+    """
+    Task to clean up export tasks created by project history logs that are older
+    than `EXPORT_CLEANUP_GRACE_PERIOD`, excluding those that are still processing
+    """
+    cleanup_exports_helper(ProjectHistoryLogExportTask)
+
+
```
noliveleger: Let's move the tasks in their own django app. `audit-logs`

kpi/tests/test_cleanup_audit_log_exports.py
```diff
@@ -0,0 +1,161 @@
...
+            model.objects.filter(pk=old_export.pk).exists()
+        )
+        self.assertTrue(
+            model.objects.filter(pk=recent_export.pk).exists()
+        )
+
+    @data(
+        (AccessLogExportTask, cleanup_access_log_exports),
+        (ProjectHistoryLogExportTask, cleanup_project_history_log_exports),
+    )
```
noliveleger: That's a really good idea. 🧞 .  You could do the same with `AnonymousExport` and `SynchronousExport` and use only one test class.

rajpatel24: done

noliveleger (approved):
LGTM

PR #6521, opened 2025-12-03
feat(gallery): paginated modal for images DEV-1212

magicznyleszek:
### 📣 Summary

Project → Data → Gallery images now open images in a modal with prev/next pagination.

### 💭 Notes

A separate issue of "Load more" loading empty pages is fixed separately - see #6522.

Changes here:
- Added modal handling actions and updated the reducer
- Added Mantine modal for viewing the image
- Switching to Mantine components for some things
- Loading next page of submissions after reaching last image in currently loaded submissions
- Using total submissions count in modal's `x of y` counter
- Changed page size to 20 to ease out the load

### 👀 Preview steps

Note that there is a bug in pagination (https://linear.app/kobotoolbox/issue/DEV-1364/gallery-pagination-works-in-confusing-matter) that will be fixed by #6522

1. ℹ️ have an account and a project with `image` question
2. make 21+ submissions with images
3. go to Project → Data → Gallery
4. 🔴 [on main] notice clicking image opens raw image
5. 🟢 [on PR] notice clicking image opens modal with next/prev buttons
6. click last image (20th)
7. click next button
8. 🟢 notice next button changes into spinner
9. after next page of submissions loads, 21st image is being displayed in modal

### DIFF
--- a/jsapp/js/components/formGallery/formGallery.component.tsx
+++ b/jsapp/js/components/formGallery/formGallery.component.tsx
@@ -3,13 +3,13 @@ import './formGallery.component.scss'
 import { Box, Center, Flex, Image, Modal } from '@mantine/core'
 import React, { useEffect, useMemo, useReducer } from 'react'
 import ReactSelect from 'react-select'
+import { fetchGet, fetchGetUrl } from '#/api'
 import { getFlatQuestionsList } from '#/assetUtils'
 import DeletedAttachment from '#/attachments/deletedAttachment.component'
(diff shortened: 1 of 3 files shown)


duvld (commented):
Code and functionality looks good. Only question I have is should we disable the left arrow on the first image? 
image

It takes us to 20 out of 21, not the bottom of the image list (since we didn't fetch other rows yet). I think the expected function here would bring the user to 21.

PR #6535, opened 2025-12-08
fix(subsequences): generate accurate OpenAPI schemas DEV-1431

RuthTurk:
### Summary 
This update improves the API documentation for subsequences so it more accurately reflects the actual data returned by the API.


kobo/apps/subsequences/models.py
```diff
@@ -82,7 +82,7 @@ def revise_data(asset: 'kpi.Asset', submission: dict, incoming_data: dict) -> di
                     raise InvalidAction from e
 
                 action = feature.to_action()
-                action.check_limits(asset.owner)
+                # action.check_limits(asset.owner)
```
Guitlle: why is this?

PR #6542, opened 2025-12-10
feat(pagination): make pagination class consistent and use the same parameters DEV-1393

Guitlle:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
This PR introduces the DefaultPagination class, which serves as the base pagination class for all paginators in the KPI and KoBo apps. Other pagination classes now inherit from it.
`start` becomes the primary pagination parameter, while `offset` remains supported as an alias for backward compatibility.

It also adds support for the `page` and `page_size` query parameters, which are internally converted into limit and offset values.

### Notes 

`/api/v2/hook/{uid_hook}/logs/` with now use `?start=` and `?limit=` instead of `?page`

### 👀 Preview steps
1. Have an account and a project with many submissions 
2. Go to the project data view and browse the submissions
3. Go to the audit logs and click on the next page button
4. Ensure there are no error messages related to pagination responses in the frontend or in the backend logs

### DIFF
--- a/kpi/paginators.py
+++ b/kpi/paginators.py
@@ -41,7 +41,7 @@ class DefaultPagination(LimitOffsetPagination):
                     'type': ['string', 'null'],
                     'nullable': True,
                     'format': 'uri',
-                    'example': 'https://kf.kobotoolbox.org/accounts/?{offset_param}=400&{limit_param}=100'.format(
+                    'example': 'https://kf.kobotoolbox.org/accounts/?{offset_param}=400&{limit_param}=100'.format(  # noqa E501
                         offset_param=self.offset_query_param,
                         limit_param=self.limit_query_param,
                     ),
@@ -50,7 +50,7 @@ class DefaultPagination(LimitOffsetPagination):
                     'type': ['string', 'null'],
                     'nullable': True,
                     'format': 'uri',
-                    'example': 'https://kf.kobotoolbox.org/accounts/?{offset_param}=200&{limit_param}=100'.format(
+                    'example': 'https://kf.kobotoolbox.org/accounts/?{offset_param}=200&{limit_param}=100'.format(  # noqa E501
                         offset_param=self.offset_query_param,
                         limit_param=self.limit_query_param,
                     ),
@@ -121,7 +121,7 @@ class DefaultPagination(LimitOffsetPagination):
         if self.count == 0 or self.offset > self.count:
             return []
 
-        return list(queryset[self.offset : (self.offset + self.limit)])
+        return list(queryset[self.offset:(self.offset + self.limit)])
 
     def get_schema_operation_parameters(self, view):
         schema = [
--- a/kpi/tests/test_pagination.py
+++ b/kpi/tests/test_pagination.py
@@ -74,4 +74,4 @@ class TestNoCountPagination(TestCase):
         response = self.view(request)
 
         assert 'count' not in response.data
-        assert data[default_limit : (default_limit * 2)] == response.data['results']
+        assert data[default_limit:(default_limit * 2)] == response.data['results']


kpi/paginators.py
```diff
@@ -7,24 +7,103 @@
...
+    well as `page` number.
+
     Adds 'root' to the wrapping response object.
     """
+
     root = SerializerMethodField('get_parent_url', read_only=True)
 
+    limit_query_param = 'limit'
+    default_limit = settings.REST_FRAMEWORK['PAGE_SIZE']
+    max_limit = 1000  # Reasonable maximum limit to avoid sending full querysets
```
noliveleger: I like the idea but let's confirm with the product team before adding this new restriction to all endpoints. 
You may review the logic of `get_limit()` if we need to keep `max_limit` to `None` by default.

kpi/paginators.py
```diff
@@ -7,24 +7,103 @@
...
+                strict=True,
+                cutoff=self.max_limit,
+            )
+        except (ValueError, TypeError):
+            return None
+
+    @classmethod
+    def custom_class(cls, **kwargs):
+        class_name = kwargs.pop('class_name', 'CustomPagination')
+        return type(class_name, (cls,), kwargs)
```
noliveleger: Not used anymore (except the test), is it really worth it to keep it?

kpi/paginators.py
```diff
@@ -121,22 +200,22 @@ def get_paginated_response_schema(self, schema):
...
 
-class DataPagination(LimitOffsetPagination):
+class DataPagination(DefaultPagination):
     """
-    Pagination class for submissions.
+    Pagination for the data viewset
     """
+
     default_limit = 100
-    offset_query_param = 'start'
```
noliveleger, line 130: This should be moved to the parent class. We want `offset_query_param=start` by default. We don't want to change the behaviour of the data endpoint.

kpi/paginators.py
```diff
@@ -7,24 +7,103 @@
...
+        try:
+            return _positive_int(limit, strict=True, cutoff=self.max_limit)
+        except (ValueError, TypeError):
+            return self.default_limit
+
+    def get_offset(self, request):
+        page_number = self.get_page_number(request)
+        offset = (
+            request.query_params.get('offset')
+            or request.query_params.get('start')
```
noliveleger: `start` should be the first. It has priority on `offset`. 

I cannot comment on documentation but update the documentation accordingly. 
Make `offset` a deprecated alias of `start` in every endpoint which uses `offset`.  

data endpoint should not change.

kpi/paginators.py
```diff
@@ -190,8 +271,9 @@ def get_next_link(self):
         return replace_query_param(url, self.offset_query_param, offset)
 
 
-class TinyPaginated(PageNumberPagination):
+class TinyPagination(PageNumberPagination):
```
noliveleger: You can remove this class and use `DefaultPagination` directly. 
Make the HookLogViewset use `DefaultPagination`.
Update the documentation to say `?page=` is a deprecated parameter.
Let's create a task for the front-end to use `start` and `limit`

noliveleger (changes requested):

kpi/tests/test_pagination.py
```diff
@@ -0,0 +1,83 @@
...
+class TestDefaultPagination(TestCase):
+    def setUp(self):
+        self.request_factory = APIRequestFactory()
+
+    def test_custom_pagination_attributes(self):
+        CustomClass = DefaultPagination.custom_class(
+            page_size=123, custom_parameter='abc'
+        )
+        assert CustomClass.custom_parameter == 'abc'
+        assert CustomClass.page_size == 123
```
noliveleger: I don't think we need to keep `def custom_class`, so we can remove this test too.

kpi/tests/test_pagination.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def test_with_start_offset(self):
+        self.view = generics.ListAPIView.as_view(
+            serializer_class=PassThroughSerializer,
+            queryset=range(1, 101),
+            pagination_class=DefaultPagination.custom_class(default_limit=5),
```
noliveleger: We could override REST_FRAMEWORK['PAGE_SIZE'] to achieve the same result without using `custom_class` .

kpi/paginators.py
Guitlle: There's a collision in the parameters. The HookLogFilter uses the start/end fields to filter by date_modified... I think we may want to change the filter, given that start is now officially the pagination field for all paginators in the API. What do you think?

noliveleger: I wrote something about that. I may have forgotten to submit.
I do agree with you. I was thinking about start_date and end_date !
Or we could use our query parser instead , e.g: ?q=date_modified__lte:…

noliveleger (approved):
LGTM

PR #6547, opened 2025-12-12
feat(subsequences): add OpenAPI schema for advanded features DEV-1230

Guitlle:
### 📣 Summary
Add OpenAPI schema for the `/api/v2/assets/{uid_asset}/advanced-features/` endpoint. 

### 📖 Description
The API schema output files and the generated Orval types have been updated with the schema details for the action parameters in the `QuestionAdvancedFeature` model.

### DIFF
--- a/kpi/schema_extensions/v2/subsequences/examples.py
+++ b/kpi/schema_extensions/v2/subsequences/examples.py
@@ -35,7 +35,6 @@ def get_advanced_features_create_examples() -> list[OpenApiExample]:
                         'uuid': 'aaaaaaaa-bbbb-cccc-dddd-eeeeffffffff',
                         'labels': {
                             '_default': 'Qualitative analysis result',
-                            'es': 'Resultado de análisis cualitativo',
                         },
                     },
                 ],
@@ -61,31 +60,24 @@ def get_advanced_features_create_examples() -> list[OpenApiExample]:
                         'uuid': 'aaaaaaaa-bbbb-cccc-dddd-eeeeffffffff',
                         'labels': {
                             '_default': 'Urgency Level',
-                            'es': 'Nivel de Urgencia',
                         },
                         'choices': [
                             {
                                 'uuid': 'qqqqqqqq-bbbb-cccc-dddd-eeeeffffffff',
                                 'labels': {
                                     '_default': 'High',
-                                    'fr': 'Élevé',
-                                    'es': 'Alto',
                                 },
                             },
                             {
                                 'uuid': 'hhhhhhhh-bbbb-cccc-dddd-eeeeffffffff',
                                 'labels': {
                                     '_default': 'Medium',
-                                    'fr': 'Moyen',
-                                    'es': 'Medio',
                                 },
                             },
                             {
                                 'uuid': 'gggggggg-bbbb-cccc-dddd-eeeeffffffff',
                                 'labels': {
                                     '_default': 'Low',
-                                    'fr': 'Bas',
-                                    'es': 'Bajo',
                                 },
                             },
                         ],
@@ -107,15 +99,15 @@ def get_advanced_features_create_examples() -> list[OpenApiExample]:
                         'choices': [
                             {
                                 'uuid': 'xxxxxxxx-bbbb-cccc-dddd-eeeeffffffff',
-                                'labels': {'_default': 'Shelter', 'ar': 'مأوى'},
+                                'labels': {'_default': 'Shelter'},
                             },
                             {
                                 'uuid': 'zzzzzzzz-bbbb-cccc-dddd-eeeeffffffff',
-                                'labels': {'_default': 'Food', 'ar': 'طعام'},
+                                'labels': {'_default': 'Food'},
                             },
                             {
                                 'uuid': 'yyyyyyyy-bbbb-cccc-dddd-eeeeffffffff',
-                                'labels': {'_default': 'Medical', 'ar': 'طبي'},
+                                'labels': {'_default': 'Medical'},
                             },
                         ],
                     },
@@ -161,7 +153,6 @@ def get_advanced_features_list_examples() -> list[OpenApiExample]:
                         'uuid': 'aaaaaaaa-bbbb-cccc-dddd-eeeeffffffff',
                         'labels': {
                             '_default': 'Qualitative analysis result',
-                            'es': 'Resultado de análisis cualitativo',
                         },
                     },
                 ],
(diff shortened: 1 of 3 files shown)


kobo/apps/subsequences/views.py
```diff
@@ -23,13 +28,14 @@
     open_api_201_created_response,
 )
 from kpi.utils.viewset_mixins import AssetNestedObjectViewsetMixin
+from kpi.versioning import APIV2Versioning
```
noliveleger, line 30: Let's specify the custom lookup_field, i.e.:

`lookup_field = 'uid_advanced_feature'`

kobo/apps/subsequences/views.py
```diff
@@ -66,13 +74,14 @@
         ),
         parameters=[
             OpenApiParameter(
-                name='uid',
+                name='id',
```
noliveleger: Let's use the name custom lookup_field `uid_advanced_feature` instead.

kobo/apps/subsequences/views.py
```diff
@@ -66,13 +74,14 @@
...
                 type=str,
                 location=OpenApiParameter.PATH,
                 required=True,
                 description='UID of the action',
```
noliveleger: Maybe `action` is not really accurate in this context?

kobo/apps/subsequences/views.py
```diff
@@ -84,7 +93,7 @@
         ),
         parameters=[
             OpenApiParameter(
-                name='uid',
+                name='id',
```
noliveleger: Please update the `name`

kpi/schema_extensions/v2/subsequences/extensions.py
```diff
@@ -1,11 +1,111 @@
...
+                'uuid': definitions['qualUuid'],
+                'type': definitions['qualQuestionType'],
+                'labels': definitions['qualLabels'],
+                'choices': {
+                    'type': 'array',
+                    'items': definitions['qualChoice'],
+                },
+                'options': {'type': 'object'},
+            },
+            'required': ['uuid', 'type', 'labels'],
```
noliveleger: We should enforce that `choices` should be present only when `type` is `qualSelectMultiple` or `qualSectionOne`.

kobo/apps/subsequences/views.py
```diff
@@ -46,6 +52,7 @@
             require_auth=False,
             raise_access_forbidden=False,
         ),
+        examples=get_advanced_features_create_examples(),
```
noliveleger, line 54: The examples of the response seems wrong. I assume they should match the list.

kobo/apps/subsequences/views.py
```diff
@@ -55,6 +62,7 @@
             raise_access_forbidden=False,
             validate_payload=False,
         ),
+        examples=get_advanced_features_list_examples(),
```
noliveleger, line 64: Some examples are missing (manual transcription, automatic google translation). 
If you did it in purpose, let's add a description with your example and make the example more generic. 

See Delete NLP action for supplement details "

Image

Moreover, the examples are showing a double nested list. 

```json
[  
  [
    {
      "action": "manual_translation",
      "question_xpath": "q1",
      "params": [
        {
          "language": "es"
        }
      ],
      "uid": "qa123456789AbCdEfGhIjklm"
    }
  ]
]
```

kobo/apps/subsequences/views.py
noliveleger, line 47: Please update the markdowns. They could be simplified thanks to the examples you have added to this PR. (e.g. no need to explain what `params` should look like anymore in the markdown, the examples do the trick).

kobo/apps/subsequences/views.py
```diff
@@ -66,13 +74,14 @@
...
+                name='id',
                 type=str,
                 location=OpenApiParameter.PATH,
                 required=True,
                 description='UID of the action',
             ),
         ],
+        examples=get_advanced_features_update_examples(),
     ),
     retrieve=extend_schema(
```
noliveleger, line 86: The example is ~~wrong~~ , is misleading. 

Image

Let's reuse examples of the list. Consider to use the description of the example like describe previously for  `Delete NLP action for supplement details`

kobo/apps/subsequences/views.py
```diff
@@ -55,6 +62,7 @@
             raise_access_forbidden=False,
             validate_payload=False,
         ),
+        examples=get_advanced_features_list_examples(),
     ),
     partial_update=extend_schema(
         description=read_md('subsequences', 'subsequences/update.md'),
```
noliveleger, line 67: Same comment as for "retrieve" . 

Moreover, I'm not sure to understand the labels of the examples. Many are missing too.

kpi/schema_extensions/v2/subsequences/extensions.py
```diff
@@ -1,11 +1,111 @@
...
+            schema=cls._get_qual_params_schema(),
+            type=ResolvedComponent.SCHEMA,
+            object=dict,
+        )
+        registry.register_on_missing(qual_component)
+
+        return [
+            nlp_component.ref,
+            qual_component.ref,
+        ]
```
noliveleger: I created a mixin in #6535 for that purpose. Let's refactor the code when both branches are merged

kpi/schema_extensions/v2/subsequences/extensions.py
```diff
@@ -1,11 +1,111 @@
...
+        action_refs = self._get_action_refs(auto_schema.registry)
+        return build_array_type(schema={'anyOf': action_refs})
+
+    @classmethod
+    def _get_action_refs(cls, registry):
+        nlp_component = ResolvedComponent(
+            name='NLPActionParams',
+            schema=cls._get_nlp_params_schema(),
+            type=ResolvedComponent.SCHEMA,
+            object=dict,
```
noliveleger: I don't think it's expect `dict` here.  If you look at drf-spectular code, it's used to detect collision based on `object.__class__`  , so it returns ` type` all the time for `dict`. 

Not a big deal, though! 

What do you think?
(I personaly used the instance of the `OpenApiSerializerExtension` class).

kpi/schema_extensions/v2/subsequences/extensions.py
Guitlle: Great, that mixin is easier to use

kobo/apps/subsequences/views.py
Guitlle: Ok

kpi/schema_extensions/v2/subsequences/extensions.py
Guitlle: I'm reusing the mixin, so this should go away, right?

rajpatel24 (approved):
LGTM

noliveleger (approved):
LGTM

PR #6550, opened 2025-12-12
feat(subsequences): allow hiding of QA questions DEV-935

rgraber:
### 📣 Summary
Allow hiding of QA questions from editing view.

### 💭 Notes
Automatically set `{"options": {"deleted": True}` if a user attempts to delete a QA question via a PATCH request.
As part of the `update_params` method of the `qual` action, we replace the entire `options` dict within each question with whatever comes in the request. This is because the only thing I know of in the `options` dict is the `hidden` param, which we want to be able to change on the front end.  If it turns out we need the `options` dict for more, we may have to write a more sophisticated way of merging old and new options.

In the future we may want to replace the options dict with just a boolean for `hidden,` but this way minimizes the additional work the frontend will have to do to work with the new API.

### 👀 Preview steps

2. Enable QA with `curl -X POST -H 'Authorization: Token <your token>' http://kf.kobo.local/api/v2/assets/<asset_uid>/advanced-features/ --json '{"question_xpath": <xpath>, "action": <qual>, "params": [{"type": "qualInteger", "label": {"_default": "How many?"}, "uuid":"11111111-1111-1111-1111-1111111111"}]}'`
3. Note the uid of the action that comes back in the response
4. Try to "remove" the question with `curl -X PATCH -H 'Authorization: Token <your token>' http://kf.kobo.local/api/v2/assets/<asset_uid>/advanced-features/<feature_uid>/ --json '{"params": []}'`
5. 🟢 [on PR] notice that the "How many?" question is still in the response, with `{"options": {"hidden": True}}`

### DIFF
--- a/kobo/apps/subsequences/actions/qual.py
+++ b/kobo/apps/subsequences/actions/qual.py
@@ -13,6 +13,9 @@ class QualAction(BaseAction):
         allow_multiple=True, automatic=False, action_data_key='uuid'
     )
     KNOWN_PARAM_KEYS = ['uuid', 'labels', 'options', 'choices', 'type']
+    # Confusing: "deleted" actually means "hidden", we don't delete QA questions
+    # TODO: make "hidden" its own field and remove this option
+    DELETED_OPTION = 'deleted'
 
     # JSON Schema definitions
 
@@ -375,11 +378,11 @@ class QualAction(BaseAction):
                 for choice in choices:
                     new_choices_by_uuid[choice['uuid']] = choice
         for existing_question in self.params:
-            # old question is not in list of new questions, hide it and put it
-            # at the end
+            # old question is not in list of new questions, hide ("delete") it and
+            # put it at the end
             uuid = existing_question['uuid']
             if not (new_question := new_questions_by_uuid.get(uuid)):
-                existing_question.setdefault('options', {})['hidden'] = True
+                existing_question.setdefault('options', {})[self.DELETED_OPTION] = True
                 incoming_params.append(existing_question)
                 continue
             new_type = new_question['type']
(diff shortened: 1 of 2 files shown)


kobo/apps/subsequences/actions/qual.py
```diff
@@ -356,3 +360,42 @@ def transform_data_for_output(
...
+        for new_question in incoming_params:
+            if choices := new_question.get('choices'):
+                for choice in choices:
+                    new_choices_by_uuid[choice['uuid']] = choice
+        for existing_question in self.params:
+            # old question is not in list of new questions, hide it and put it
+            # at the end
+            uuid = existing_question['uuid']
+            if not (new_question := new_questions_by_uuid.get(uuid)):
+                existing_question.setdefault('options', {})['hidden'] = True
```
noliveleger: Any reason you chose `hidden` and did not keep `deleted` as discussed previously? (I saw the notes, but the front-end will have to change `deleted` to `hidden` anyway). 

[#Kobo Dev > Subsequences refactoring @ 💬](https://chat.kobotoolbox.org/#narrow/channel/4-Kobo-Dev/topic/Subsequences.20refactoring/near/760453)

Out of scope: 
>  I would leave it to Becca's decision. 

I just realized that you were the person who asked the question.

rgraber: Why will the frontend have to change `deleted` to `hidden` if `hidden` is what is currently used?

rgraber: Ah, I see. I messed up when I posted the original message in Slack. The existing feature uses `{options: {hidden: True}}`, not `{options: {deleted:True}}` so i just kept the original version

rgraber: Nope, the original was "deleted." I must have gotten confused.

noliveleger (approved):
LGTM

PR #6553, opened 2025-12-15
fix(billing): ovehaul stripe billing config handling DEV-1342

jamesrkiger:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Overhauls handling of Stripe billing portal configurations for more control over product options in the Stripe portal.

### 💭 Notes
The end goal here was to enable us to have some more nuanced rules for when to apply a customer's requested plan change (see related task [DEV-1225](https://linear.app/kobotoolbox/issue/DEV-1225/allow-upgrades-from-annual-lower-tier-plan-to-monthly-higher-tier-plan)). We also want to restrict which products on the Stripe portal are listed as upgrade/downgrade options for a given user based on their current subscription.

Our previous code relied on us using one default portal configuration and then creating new configurations with different product offerings (sometimes on the fly in the code). This resulted in us creating large numbers of portal configurations that are difficult to manage. With this PR, we will operate with a small, predetermined set of configs on the portal. The billing portal view code checks a user's current subscription and requested price (if any) to determine which config should be used.

Unfortunately, portal configurations are not handled by djstripe. Ideally I think we would want to add a db model for these and keep them synced with Stripe. But that would be more time intensive. For now, we will select configs based on hardcoded slugs matching their metadata. 

### 👀 Preview steps
1. On a Stripe-enabled and synced instance, create a new user and sign up for an **annual** 50gb storage addon.
2. After syncing again, navigate to plans page and attempt to upgrade to a **monthly** professional plan 
3. You should be taken to the Stripe billing portal and asked to confirm the change. If you click on "additional details" you should see that this change will be applied immediately and you will receive credit for the remaining time on your annual plan. For testing purposes, don't actually confirm the purchase.
4. On addons page, click "manage" button.
5. You should be taken to the the billing portal where you can change your billing details or change your plan. Click "upgrade plan".
6.  The only available plan options should be storage addons.
7. On a new, separate user account, sign up for an annual professional plan and sync Stripe subscriptions.
8. Attempt to downgrade to a monthly professional plan. 
9. You should be taken to the Stripe billing portal and asked to confirm the change. If you click on "additional details" you should see that this change will be applied at the end of your current billing period (a year from now).

### DIFF
--- a/jsapp/js/api/models/customerPortal.ts
+++ b/jsapp/js/api/models/customerPortal.ts
@@ -13,6 +13,4 @@ The endpoints are grouped by area of intended use. Each category contains relate
 export interface CustomerPortal {
   organization_id: string
   price_id?: string
-  /** @minimum 1 */
-  quantity?: number
 }
--- a/static/openapi/schema_v2.json
+++ b/static/openapi/schema_v2.json
@@ -14246,11 +14246,6 @@
                     },
                     "price_id": {
                         "type": "string"
-                    },
-                    "quantity": {
-                        "type": "integer",
-                        "minimum": 1,
-                        "default": 1
                     }
                 },
                 "required": [
--- a/static/openapi/schema_v2.yaml
+++ b/static/openapi/schema_v2.yaml
@@ -10295,10 +10295,6 @@ components:
           type: string
         price_id:
           type: string
-        quantity:
-          type: integer
-          minimum: 1
-          default: 1
       required:
       - organization_id
     CustomerPortalPostResponse:


kobo/apps/stripe/serializers.py
```diff
@@ -109,7 +109,6 @@ class CustomerPortalSerializer(serializers.Serializer):
         required=False,
         allow_empty=True,
     )
-    quantity = serializers.IntegerField(required=False, default=1, min_value=1)
```
jamesrkiger, line 112: This is not directly related to the other work here, but we don't support varying product "quantities" anymore so I decided to remove it while I was in here.

PR #6559, opened 2025-12-16
refactor(subsequences): refactor data table for consistency

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Updates the /data endpoint to return the answers to QA questions in a different format.

### 💭 Notes
Updates the data endpoint to return a dict of question_id: answer instead of a list that we then have to search by uuid. Keeps it consistent with other subsequence actions and makes it a little quicker to process. Requires a new version of formpack.
We don't include supplement stuff in the docs for the /data endpoint so there's no OpenAPI docs to update.

### 👀 Preview steps
Regression only

2. Add every type of QA question. You can either follow the directions here: (internal [link](https://linear.app/kobotoolbox/issue/DEV-1301/update-formpack-to-accept-new-shape-of-data)) to do this with curl, or switch back to `main` and do it via the UI, then switch back to the PR branch and hit `/api/v2/assets/<asset_uid>/advanced-features` to migrate them.
3. Navigate to the data table
4. 🟢 All QA answers should be present 
5. Export the data
7. 🟢 All QA answers should be in the export


rajpatel24 (approved):
LGTM
Please update to the latest commit from main once you've merged the Formpack PR :)

PR #6560, opened 2025-12-16
fix(assets): make asset list resilient for deployed projects without an XForm

noliveleger:
### 📣 Summary
Prevent errors when listing assets for deployed projects that do not have an associated XForm.

### 📖 Description
This fix improves the robustness of the asset listing by safely handling deployed projects that lack an associated XForm. Previously, such cases could trigger errors or incomplete responses during list operations. The update adds proper guards and fallbacks so these assets are still listed correctly without breaking the endpoint. 

### 👀 Preview steps

2. Deploy it
3. from the shell, delete the related XForm
4. 🔴 [on release branch] notice that the project list returns a 500
5. 🟢 [on PR] notice that the project list loads

### DIFF
--- a/kpi/tests/api/v2/test_api_assets.py
+++ b/kpi/tests/api/v2/test_api_assets.py
@@ -444,6 +444,7 @@ class AssetListApiTests(BaseAssetTestCase):
         assert response.status_code == status.HTTP_200_OK
         assert len(response.data['results']) == 1
         assert response.data['results'][0]['uid'] == asset.uid
+        assert response.data['results'][0]['date_deployed'] is None
 
 
 class AssetProjectViewListApiTests(BaseAssetTestCase):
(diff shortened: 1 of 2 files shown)


kpi/tests/api/v2/test_api_assets.py
```diff
@@ -433,6 +433,18 @@ def test_query_counts(self):
...
+        asset = Asset.objects.get(pk=1)
+        asset.save()
+        asset.deploy(backend='mock', active=True)
+
+        # Simulate desynchronized asset by deleting the related xform
+        asset.deployment.xform.delete()
+        response = self.client.get(self.list_url)
+        assert response.status_code == status.HTTP_200_OK
+        assert len(response.data['results']) == 1
+        assert response.data['results'][0]['uid'] == asset.uid
```
Guitlle, line 446: Should we test if the deployment data or date was deleted too?

noliveleger: the deployment data is not exposed in the serializer but we can definitively test `date_deployed`

PR #6563, opened 2025-12-17
fix(assets): fix an error when searching assets with numeric-only queries DEV-1487

rajpatel24:
### 📣 Summary
Fix a server error that occurred when users searched for assets using numeric-only search terms (e.g. 111).


noliveleger (approved):
LGTM.
This PR is missing a unit test. To be done in another PR.

PR #6569, opened 2025-12-19
feat(messages): add user details to in-app messages DEV-1496

jamesrkiger:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Replace selected 'template tags' with user detail on in-app messages.

### 📖 Description
Enables the use of a few template tag-style strings in in-app messages that will be replaced with user data on the frontend. 

### 👀 Preview steps
1. In django admin, create the following in-app message:
username:  ##username## 
user uid: ##user_uid##
full name: ##user_full_name##
2. Open the app as a logged in user and open message
3. See that the user info has been correctly interpolated
4. In django admin, go to extra details for user and delete the "name" field.
5. Go back to app as that user and see that the "full name" has been replaced by "KoboToolBox user"

### DIFF
--- a/kobo/apps/help/admin.py
+++ b/kobo/apps/help/admin.py
@@ -19,7 +19,8 @@ class InAppMessageAdmin(MarkdownxModelAdminBase):
     user_details_info = (
         'The following template variables can be used to insert user info '
         'in the snippet and message fields on the frontend: '
-        '##username##, ##user_uid##, and ##user_full_name## (defaults to `KoboToolbox user`).'
+        '##username##, ##user_uid##, and ##user_full_name## '
+        '(defaults to `KoboToolbox user`).'
     )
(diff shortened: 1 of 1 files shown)


jsapp/js/components/support/helpBubble.tsx
```diff
@@ -25,11 +25,17 @@ interface HelpBubbleState {
   isOpen: boolean
 }
 
-class HelpBubble extends React.Component<{}, HelpBubbleState> {
+interface HelpBubbleProps {
```
duvld, line 28: My only small nitpick is that maybe these props should be optional? Logically it makes sense why these will never be null/undefined, but in terms of using this component, it can imply we expect to use user data for in-app messages the majority of the time, when in reality it probably will be rare cases.

duvld (approved):
Small question, code looks good and functionality was tested

PR #6567, opened 2025-12-17
feat(qual): add new automatic QA action DEV-1452

rgraber:
### 🗒️ Checklist

1. [x] run linter locally
2. [x] update developer docs (API, README, inline, etc.), if any
3. [x] for user-facing doc changes create a Zulip thread at `#Support Docs Updates`, if any
4. [x] draft PR with a title `<type>(<scope>)<!>: <title> DEV-1234`
5. [x] assign yourself, tag PR: at least `Front end` and/or `Back end` or `workflow`
6. [x] fill in the template below and delete template comments
7. [x] review thyself: read the diff and repro the preview as written
8. [x] open PR & confirm that CI passes & request reviewers, if needed
9. [ ] delete this section before merging

### 📣 Summary
Update advanced features API to allow requesting LLM answers to QA questions.

### 📖 Description
Add a new "automated_chained_qual" action to the advanced features API endpoints. It is currently only a stub and will return canned answers rather than actually hitting an LLM.

### 💭 Notes
Things this PR covers: Adding the new action with params_schema, data_schema, external_data_schema
Things this PR does NOT cover: Concurrency, verification/acceptance, actually making and recording LLM requests, updating the API docs

A few important things of note:
1. There is no concept of deleting an automatic_chained_qual response. To clear a response the user would simply submit a manual_qual action with empty answers.
2. There is  no 'in_progress' status because LLM requests will be done syncronously. If we need to add it to handle concurrency that will be a separate PR.
3. Similarly there is no 'accepted' boolean because that is part of another milestone and we show un-accepted QA answers in the data table anyway now
4. Not updating the API docs because we don't want anyone to actually make requests for this action until we hook it up to an LLM

### 👀 Preview steps

1. ℹ️ have an account and a project with an audio question and a submission
2. Enable manual transcripts with
```
curl -H 'Authorization: Token <your_token>' http://kf.kobo.local/api/v2/assets/<asset_uid>/advanced-features/ 
--json '{"question_xpath":"<xpath>", "action":"manual_transcription", "params": [{"language":"en"}]}'
```
3. Enable automatic QA with 
```
curl -H 'Authorization: Token <your_token>' http://kf.kobo.local/api/v2/assets/<asset_uid>/advanced-features/ 
--json '{"question_xpath":"<xpath>", "action":"automatic_chained_qual", "params": [{"type":"qualText", "uuid":"ce609a1b-73c9-46af-9ec5-54d4df981d2b", "labels": {"_default": "This is a text question"}}]}'
``` 
[...]

PR #6583, opened 2026-01-05
fix(qual): prevent overriding answers with failures DEV-1452

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Prevent LLM failures from overriding existing QA answers.

### 💭 Notes
Update `transform_data_for_output` to check the status, if present, of a version before using it as a candidate for the data response. If the status is failed, do not use the version as a candidate.
We don't need to do this with automatic transcripts or translations because we only use candidates that have been accepted, and a user can't accept a failed version.

### 👀 Preview steps

1. ℹ️ have an account and a project with:
- an audio question
- a submission
- a transcript
- a QA question (any type except notes or tags)
- an automatic QA answer for the submission
If you do not have this handy, instructions on how to do it are here: https://github.com/kobotoolbox/kpi/pull/6567
2. manually update `AutomaticChainedQualAction.run_external_process` to return `{'status':'failed', 'error':'true'}`
3. request a new automatic answer with
```
curl -X PATCH -H 'Authorization: Token <your_token>' http://kf.kobo.local/api/v2/assets/<asset_uid>/data/<submission_uuid>/supplement/ --json '{"_version": "20250820",  "<question_xpath>": {"automatic_chained_qual": {"uuid":"<qa_question_uuid>"}}}'
```
4. Navigate to the data table
5. 🔴 [on refactor-subsequences-2025] QA column for the submission is now empty
6. 🟢 [on PR] QA column for the submission retains the most recent response

### DIFF
--- a/kobo/apps/subsequences/tests/test_automatic_chained_qual.py
+++ b/kobo/apps/subsequences/tests/test_automatic_chained_qual.py
@@ -273,4 +273,4 @@ class TestAutomaticChainedQual(BaseTestCase):
         text_item = output.get(('qual', 'uuid-qual-text'))
         # take the initial note because the most recent request to overwrite failed
         assert text_item['value'] == 'Initial note'
-        assert not text_item.get('error')
+        assert 'error' not in text_item


kobo/apps/subsequences/tests/test_automatic_chained_qual.py
```diff
@@ -237,3 +239,37 @@ def test_update_supplement_api(self):
...
+            'uuid-qual-text': {
+                '_versions': [
+                    {
+                        '_data': {
+                            'uuid': 'uuid-qual-text',
+                            'status': 'failed',
+                            'error': 'Something went wrong',
+                        },
+                        '_dateCreated': today.isoformat(),
+                        '_uuid': 'v1',
```
noliveleger: nitpick: should `v1` come before `v2`? I know if it's only random string but I think it would make it more obvious. What do you think?

kobo/apps/subsequences/tests/test_automatic_chained_qual.py
```diff
@@ -237,3 +239,37 @@ def test_update_supplement_api(self):
...
+                '_dateModified': today.isoformat(),
+            }
+        }
+
+        output = self.action.transform_data_for_output(action_data)
+        assert len(output.keys()) == 1
+
+        text_item = output.get(('qual', 'uuid-qual-text'))
+        # take the initial note because the most recent request to overwrite failed
+        assert text_item['value'] == 'Initial note'
```
noliveleger, line 275: Would we like to assert that `error` is not present or it is overkilled?

PR #6585, opened 2026-01-06
chore(kpi): remove KPI `v1` endpoints DEV-1266

rajpatel24:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Removed KPI `v1` endpoints. (see full list [here](https://github.com/kobotoolbox/kpi/blob/main/DEPRECATION.md#kpi-endpoints) )

### 👀 Preview steps
Ensure all related views and URLs are removed, and tests are updated to use v2 endpoints where applicable.

### DIFF
--- a/kobo/apps/accounts/tests/test_api_tos.py
+++ b/kobo/apps/accounts/tests/test_api_tos.py
@@ -9,7 +9,7 @@ from kpi.tests.base_test_case import BaseTestCase
 
 class TOSTestCase(BaseTestCase):
     def setUp(self) -> None:
-        self.url = reverse(self._get_endpoint('tos'))
+        self.url = reverse('tos')
         self.user = baker.make(
             settings.AUTH_USER_MODEL, username='spongebob', email='me@sponge.bob'
         )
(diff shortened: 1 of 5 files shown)


kobo/apps/hook/tests/test_api_hook.py
```diff
@@ -22,24 +22,29 @@
     PERM_VIEW_SUBMISSIONS,
     SUBMISSION_FORMAT_TYPE_JSON,
 )
+from kpi.urls.router_api_v2 import URL_NAMESPACE as ROUTER_URL_NAMESPACE
```
Guitlle: This is repeated in lots of classes, can we put this as the default namespace in the base test class?

Guitlle (changes requested):

kobo/apps/hook/tests/test_api_hook.py
noliveleger: It was because of `v1`. Since `v1` is being removed I totally agree with you.

rajpatel24: I think we can’t add it to the `BaseTestCase` because it’s used in many places, and not every endpoint relies on `v2` (for example, `/me` or the login/logout endpoints).

rajpatel24: Thanks! Updated the tests accordingly.

PR #6594, opened 2026-01-07
chore(api): remove metadata endpoints DEV-1259

Guitlle:
### 📣 Summary
Removed the metadata v1 endpoints, the obsolete viewset class and the test cases for the viewset.

### 👀 Preview steps
Review the removed code. No preview steps

### DIFF
--- a/kobo/apps/openrosa/apps/api/tests/viewsets/test_abstract_viewset.py
+++ b/kobo/apps/openrosa/apps/api/tests/viewsets/test_abstract_viewset.py
@@ -8,8 +8,6 @@ from django.core.files.base import ContentFile
 from django.test import TestCase
 from django.utils import timezone
 from django_digest.test import DigestAuth
-from rest_framework import status
-from rest_framework.reverse import reverse
 from rest_framework.test import APIRequestFactory
 
 from kobo.apps.kobo_auth.shortcuts import User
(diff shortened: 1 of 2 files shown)


kobo/apps/openrosa/apps/api/viewsets/metadata_viewset.py
```diff
@@ -1,194 +0,0 @@
...
-    content_negotiation_class = MediaFileContentNegotiation
-    filter_backends = (filters.MetaDataFilter,)
-    queryset = MetaData.objects.all()
-    permission_classes = (MetaDataObjectPermissions,)
-    renderer_classes = (
-        renderers.JSONRenderer,
-        renderers.BrowsableAPIRenderer,
-        MediaFileRenderer,
-    )
-    serializer_class = MetaDataSerializer
```
rajpatel24, line 171: Can we also remove `MetaDataFilter` and `MetaDataObjectPermissions` if they are not being used anywhere else now? Also, I think we should remove `MetaDataSerializer` as well. What do you think?

Guitlle: `MetadataFilter` and `MetadataObjectPermissions` are now removed. The serializer is being used by the xform serializer, still.

PR #6595, opened 2026-01-08
fix(dev)!: urgent fix for migrations script error DEV-1536

Guitlle:
### 📣 Summary
This is a backport of #6485. It fixes the error in the `fix_migrations_for_kpi.py` script in a fresh install. This was caused by the script trying to query the django_migrations table before it exists.

### 👀 Preview steps
1. Start a new installation using kobo-install and the branch release/2.025.47
2. 🔴 [on release/2.025.47] the `fix_migrations_for_kpi.py` script fails because the django migrations table does not exist yet
5. 🟢 [on PR] the migrations proceeds as expected

### DIFF
--- a/scripts/fix_migrations_for_kpi.py
+++ b/scripts/fix_migrations_for_kpi.py
@@ -1,3 +1,4 @@
+from django.contrib.auth.management import DEFAULT_DB_ALIAS
 from django.db import connection
 
 from .fix_migrations_for_kobocat import are_migrations_already_applied
@@ -5,8 +6,8 @@ from .fix_migrations_for_kobocat import are_migrations_already_applied
 
 def run():
 
-    if not are_migrations_already_applied('default'):
-        print('Migrations not yet applied. Skip!')
+    if not are_migrations_already_applied(DEFAULT_DB_ALIAS):
+        print('Skipping KPI migration fixes...')
         return
 
     if should_fix_internal_mfa_app_label():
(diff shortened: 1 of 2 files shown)


scripts/fix_migrations_for_kpi.py
```diff
@@ -1,11 +1,12 @@
 from django.db import connection
 
-from .fix_migrations_for_kobocat import are_migration_already_applied
+from .fix_migrations_for_kobocat import are_migrations_already_applied
 
 
 def run():
 
-    if not are_migration_already_applied():
+    if not are_migrations_already_applied('default'):
```
noliveleger: Let's use the setting . 
Replace `'default'` with `settings.DEFAULT_DB_ALIAS`

scripts/fix_migrations_for_kpi.py
```diff
@@ -1,11 +1,12 @@
...
+        print('Migrations not yet applied. Skip!')
```
noliveleger: I think I would use a more obvious message. 
Something like 

> Fresh install detected (no applied migrations on database: xxxx). Skipping.

You can move the print inside the function

Guitlle: ok

scripts/fix_migrations_for_kpi.py
Guitlle: ok

noliveleger (approved):
LGTM

PR #6580, opened 2026-01-05
feat(qual): request automatic LLM processing of QA questions DEV-1453

rgraber:
### 🗒️ Checklist

1. [x] run linter locally
2. [x] update developer docs (API, README, inline, etc.), if any
3. [x] for user-facing doc changes create a Zulip thread at `#Support Docs Updates`, if any
4. [x] draft PR with a title `<type>(<scope>)<!>: <title> DEV-1234`
5. [x] assign yourself, tag PR: at least `Front end` and/or `Back end` or `workflow`
6. [x] fill in the template below and delete template comments
7. [x] review thyself: read the diff and repro the preview as written
8. [ ] open PR & confirm that CI passes & request reviewers, if needed
9. [ ] delete this section before merging

### 📣 Summary
Use Claude LLM to respond to QA questions.

### 📖 Description
Update the `/supplement` endpoint to actually send questions to an LLM for processing (previously returned canned answers).

### 👷 Description for instance maintainers
Assumes AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, and AWS_BEDROCK_REGION_NAME  are available the environment or in Django settings and authenticate a user who has access to AWS Bedrock.

### 💭 Notes
This PR covers creating a prompt to send to the LLM, synchronously sending the prompt and waiting for the response,
then processing the response into a suitable `value`. 
Not covered: Using a backup LLM, OpenAPI docs, UI for tweaking prompts, configurable model temperature and max tokens

All methods for tweaking how the actual text of the prompt is determined are in a separate `prompts.py`. This should give future developers a centralized location to tweak the prompt generation without affecting anything else. All those methods are given doc strings to make it easier to understand what they do and how to override them. 

The model temperature and max tokens can also be adjusted from this file (which is why they are hardcoded instead of taken from settings so all changes can be in one file).

If the model returns something we can't figure out how to parse, it gets logged as a 'failed' version rather than raising an error on the grounds that, from a technical perspective, nothing actually went wrong but the LLM we used didn't understand what the user wanted.

Eventually we'll connect to AWS in a better way but for now this is the way I know how.

### 👀 Preview steps
1. ℹ️ have an account and a project with an audio question and a submission
2. Enable manual transcripts with
```
curl -H 'Authorization: Token <your_token>' http://kf.kobo.local/api/v2/assets/<asset_uid>/advanced-features/ 
--json '{"question_xpath":"<xpath>", "action":"manual_transcription", "params": [{"language":"en"}]}'
```
[...]

PR #6599, opened 2026-01-09
feat(qual): add PH logs for LLM requests DEV-964

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Record project history logs when users request LLM answers to QA questions.

### 📖 Description
Logs for LLM requests will have `modify-automatic-qa-data` as the action and include the model, number of input tokens, and number of output tokens.

### 💭 Notes
Includes temporary accommodations for
1. The case where users include the 'uuid:' prefix in the root uuid in the url
2. The fact that we still use AWS credentials in Django 

Eventually both those things will be fixed and some of the new code can be removed

### 👀 Preview steps

1. Repeat the preview steps 1-5 here: https://github.com/kobotoolbox/kpi/pull/6580
2. Navigate to `/api/v2/assets/<asset_uid>/history`
3. :green_circle: There should be an `update_qa` log with all the new questions that were added
4. :green_circle: There should be a `modify_qa_data` log (for when the transcription was created)
5. Request an LLM answer to the integer question with 
```
curl -X PATCH -H 'Authorization: Token <your_token>' http://kf.kobo.local/api/v2/assets/<asset_uid>/data/<submission_uuid>/supplement/ --json '{"_version": "20250820",  "<question_xpath>": {"automatic_bedrock_qual": {"uuid":"1a2c8eb0-e2ec-4b3c-942a-c1a5410c081a"}}}'
```
7. :green_circle: There should be a `modify-automatic-qa-data` log with an `llm` section in the metadata. The `model` should be `claude-3-5-sonnet-20240620`. The `input_tokens` and `output_tokens` may vary but should be around 91 and 5, respectively.

### DIFF
--- a/kobo/apps/subsequences/actions/automatic_bedrock_qual.py
+++ b/kobo/apps/subsequences/actions/automatic_bedrock_qual.py
@@ -204,6 +204,9 @@ class AutomaticBedrockQual(RequiresTranscriptionMixin, BaseQualAction):
             return response_body['content'][0]['text']
         except (JSONDecodeError, IndexError, KeyError) as e:
             # the response isn't in the form we expected
+            get_current_request().llm_response = {
+                'error': 'Unable to extract answer from LLM response object',
+            }
             raise InvalidResponseFromLLMException(
                 'Unable to extract answer from LLM response object'
             ) from e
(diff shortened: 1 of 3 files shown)


kobo/apps/subsequences/actions/automatic_bedrock_qual.py
Guitlle, line 210: We should also log bad responses. In this case, if the response is not ok, it won't set request.llm_response. Though maybe this could be left for another ticket, what do you think?

rgraber: Good call

PR #6607, opened 2026-01-14
fix(subsequences): throw error when deleting a null value in supplement API DEV-1550

Guitlle:
### 📣 Summary
This PR adds validation to prevent setting value: null on non-existent transcriptions and translations in the submission supplement API.

### 👀 Preview steps
1. ℹ️ have a project with an audio question, in this case the xpath is `audio_recording`
2. Enable manual transcriptions
```
curl -H "Authorization: Token $APIKEY" "http://kf.kobo.local/api/v2/assets/$ASSETID/advanced-features/" --json '{"question_xpath":"audio_recording", "action":"manual_transcription", "params": [{"language":"en"}]}'
```
3. Enter a new submission, it should not have any transcription yet
4. Attempt to set the transcription value to null
```
curl -X PATCH -H "Authorization: Token $APIKEY" "http://kf.kobo.local/api/v2/assets/$ASSETID/data/$SUBID/supplement/" --json '{"_version": "20250820",  "audio_recording": {"manual_transcription": {"language":"en", "value":null}} }'
```
6. 🔴 [on feature-automatic-qa] it would allow you to do this with no error
7. 🟢 [on PR] it should not allow you to do this, and return a Invalid payload error.

### DIFF
--- a/kpi/models/import_export_task.py
+++ b/kpi/models/import_export_task.py
@@ -1061,9 +1061,7 @@ class SubmissionExportTaskBase(ImportExportTask):
         )
 
         if source.has_advanced_features:
-            pack.extend_survey({
-                'additional_fields': get_analysis_form_json(source)
-            })
+            pack.extend_survey(get_analysis_form_json(source))
 
         # Wrap the submission stream in a generator that records the most
         # recent timestamp


kobo/apps/subsequences/actions/base.py
```diff
@@ -431,6 +431,12 @@ def revise_data(
         else:
             dependency_supplemental_data = action_data.pop(self.DEPENDENCY_FIELD, None)
             # Deletion is triggered by passing `{value: null}`.
+            # Check if user is trying to delete null data
+            if (
+                not current_version.get(self.VERSION_DATA_FIELD, {}).get('value')
+                and action_data.get('value') is None
+            ):
+                raise ValueError
```
noliveleger: Let's create a custom `ValueError` exception here. 
`ValueError` is too generic and could appear somewhere else. 

It would be easier to target it [here](https://github.com/kobotoolbox/kpi/pull/6607/changes#diff-e44cbd6664672da75889a1500254b451333c3c43609c30cc8d07b1b8b40c3baeR569) when errors are caught to return a `serializers.ValidatorError`


Moreover, this code seems to cover only manual actions. This should also happen with automated actions. 
I think this validation should take place way before in the code.

kobo/apps/subsequences/tests/api/v2/test_api.py
```diff
@@ -779,6 +779,57 @@ def test_cannot_request_translation_without_transcription(self):
             assert response.status_code == status.HTTP_400_BAD_REQUEST
             assert 'Cannot translate without transcription' in str(response.data)
 
+    def test_cannot_delete_non_existent_transcription(self):
```
noliveleger, line 825: It's not only manual transcription, but also manual translation and automatic transcription and translation.

kobo/apps/subsequences/tests/api/v2/test_api.py
```diff
@@ -779,6 +779,57 @@ def test_cannot_request_translation_without_transcription(self):
...
+        """
+        Users should not be able to delete a transcription from a submission
+        that doesn't have an existing transcription to begin with.
+        Setting `value: null` on a non-existent transcription should return 400.
+        """
```
noliveleger, line 830: I used your the test you created in the #6611 to give an example 


```python
    def test_cannot_delete_non_existent_transcription(self):

        QuestionAdvancedFeature.objects.create(
            asset=self.asset,
            question_xpath='q1',
            action='automatic_google_transcription',
            params=[{'language': 'en'}],
        )

        # Create a mock submission supplement with no 'value' field
        failed_transcription_data = {
            'q1': {
                'automatic_google_transcription': {
                    '_dateCreated': '2026-01-11T01:29:00.908261Z',
                    '_dateModified': '2026-01-11T01:29:00.908261Z',
                    '_versions': [
                        {
                            '_data': {
                                'language': 'en',
                                'status': 'failed',
                                'error': 'Any error',
                            },
                            '_dateCreated': '2026-01-11T01:29:00.908261Z',
                            '_uuid': '08668365-c922-48ea-9f0e-26935ca2755e',
                        }
                    ],
                }
            },
            '_version': '20250820',
        }

        SubmissionSupplement.objects.create(
            asset=self.asset,
            submission_uuid=self.submission_uuid,
            content=failed_transcription_data,
        )

        # Attempt to "delete" a non-existent transcription
        payload = {
            '_version': '20250820',
            'q1': {
                'automatic_google_transcription': {
                    'language': 'en',
                    'value': None,
                },
            },
        }

        response = self.client.patch(
            self.supplement_details_url, data=payload, format='json'
        )
        # Should fail because there's nothing to delete
```

This test will fail, it returns a 200 although it should also return a 400.

kobo/apps/subsequences/actions/base.py
Guitlle: Ok, let me know what you think of `DeletionTargetNotFound`

kobo/apps/subsequences/tests/api/v2/test_api.py
Guitlle: Ok, will add more test cases later, though my latest commits cover for any automatic or manual action

kobo/apps/subsequences/actions/base.py
noliveleger: or `SubsequenceDeletionError`  (to be consistent with `SubsequenceTimeoutError`)

kobo/apps/subsequences/actions/base.py
```diff
@@ -422,6 +423,13 @@ def revise_data(
                 # Stop here to avoid processing data and creating redundant revisions.
                 return None
 
+            # Check if user is trying to delete null data
+            if (
+                not current_version.get(self.VERSION_DATA_FIELD, {}).get('value')
+                and action_data.get('value') is None
+                and service_response.get('status') == 'deleted'
+            ):
+                raise DeletionTargetNotFound
```
noliveleger: would not be possible to get the code out the condition and validate the wrong deletion first? 

To accept a deletion, the conditions must be:
- 'value' must be present in `action_data`
- `action_data['value']` is None
- current_version must have a value

, no matter it's an automatic or manual action.

I do think we can validate this [here](https://github.com/kobotoolbox/kpi/blob/1493c28eaf57af87f25b4218f1a9ca581b76f292/kobo/apps/subsequences/actions/base.py#L411)

Guitlle: Ok, thanks for the hint

noliveleger (approved):
LGTM

PR #6610, opened 2026-01-14
fix(organizations): remove duplicate organizations DEV-1491

rajpatel24:
### 📣 Summary
A user should only be a member of a single organization. This PR removes any duplicate organizations associated with a user.

### 📖 Description
Certain accounts were inadvertently associated with multiple organizations due to historical system behavior. This led to confusion around ownership, permissions, and subscription management.

This PR introduces a long-running migration that ensures each user remains part of exactly one organization going forward. When multiple organizations are found for a user, the system keeps the most appropriate one based on the following rules:

- Prefer multi-member organizations (MMO)
- If none are MMO, prefer the one with an active subscription
- Otherwise, keep the newest organization (based on creation date)
- Where safe, unused organizations are removed. 

In addition, this migration addresses two consistency issues that were observed during cleanup:

1. Any projects that were still owned by the user (instead of the organization) are now transferred to the owner of the kept organization, aligning with the expected ownership model.

2. Organizations that are effectively multi-member but were not marked as MMO will now have their `mmo_override` flag updated for consistency to avoid future ambiguity in membership and billing flows.

After this migration completes, each user will belong to exactly one organization, shared organizations remain intact, and project ownership and MMO status are consistent with current system behavior.

### DIFF
--- a/kobo/apps/long_running_migrations/jobs/0014_fix_duplicate_organizations.py
+++ b/kobo/apps/long_running_migrations/jobs/0014_fix_duplicate_organizations.py
@@ -99,7 +99,7 @@ def run():
 
             # Transfer asset if the user is not the owner of the kept org
             if keep.owner_user_object.pk != uid:
-                transfer_member_data_ownership_to_org.delay(uid)
+                transfer_member_data_ownership_to_org(uid)
 
             for org in to_remove:
                 with transaction.atomic():


kobo/apps/long_running_migrations/jobs/0014_fix_duplicate_organizations.py
```diff
@@ -0,0 +1,113 @@
...
+                if actives:
+                    keep = get_newest_org(actives)
+                else:
+                    keep = get_newest_org(orgs)
+
+            to_remove = [org for org in orgs if org.id != keep.id]
+
+            for org in to_remove:
+                with transaction.atomic():
+                    revoke_org_asset_perms(org, [uid])
```
noliveleger, line 106: Need to be validated, but we want to transfer those projects to the owner of the `keep` org (if they are different from current user).

rajpatel24: Updated the PR!

kobo/apps/long_running_migrations/jobs/0014_fix_duplicate_organizations.py
```diff
@@ -0,0 +1,122 @@
...
```
noliveleger, line 106: I think, before entering the loop (i.e.: `for org in to_remove:`, we should transfer the projects to `keep` if user is not the owner of `keep` - not after). Then if, inside the loop, the `org` has no other users, it's safe to delete it. Otherwise, don't touch it, just revoke access to org_to_be_removed.

PR #6611, opened 2026-01-14
fix(subsequences): fix 500 error when value is missing in supplement data DEV-1547

Guitlle:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Fix to gracefully handle missing value fields in failed transcription/translation entries instead of crashing with a 500 error when loading the data API endpoint

### 👀 Preview steps
This can't be tested by conventional means, as you would need to provoke a failed automatic transcription in order to reproduce the error. However, it's covered via a unit test. Please review the code, it's a simple fix with corresponding unit test


kobo/apps/subsequences/actions/mixins.py
```diff
@@ -139,7 +139,7 @@ def transform_data_for_output(
         return {
             self.col_type: {
                 'languageCode': version_data['_data']['language'],
-                'value': version_data['_data']['value'],
+                'value': version_data['_data'].get('value'),
```
noliveleger, line 142: We are also missing `'regionCode': version_data['_data'].get('locale'),`

kpi/models/import_export_task.py
```diff
@@ -1061,7 +1061,9 @@ def get_export_object(
         )
 
         if source.has_advanced_features:
-            pack.extend_survey(get_analysis_form_json(source))
+            pack.extend_survey({
+                'additional_fields': get_analysis_form_json(source)
+            })
```
noliveleger: Please do not merge the PR until we have a valid explanation for this?

noliveleger (changes requested):

kpi/models/import_export_task.py
Guitlle: I'm reverting it because we fixed this on the base branch

PR #6614, opened 2026-01-15
fix(subsequences): fix OpenAPI schemas with dynamic properties DEV-1573

noliveleger:
### 💭 Notes
Some data supplement responses include dynamic keys (for example, per-question entries) that are modeled using `additionalProperties` in the OpenAPI schema.

When generating TypeScript clients, Orval does not correctly handle these schemas and produces invalid or incomplete typings, leading to type conflicts in the generated models.

Additionally, `patternProperties` is not supported in OpenAPI 3.0.0 and is not handled by the version of DRF-spectacular currently in use. As a result, the schema logic must be expressed using `additionalProperties`, even though it
requires duplicating part of the validation logic.

This PR adjusts the OpenAPI schema generation to work around these limitations so that DRF-spectacular can generate a valid schema and downstream tooling (such as Orval) can produce stable TypeScript types.

There are no changes to the API behavior; this update only affects OpenAPI schema generation and client-side tooling.

### DIFF
--- a/dependencies/pip/dev_requirements.txt
+++ b/dependencies/pip/dev_requirements.txt
@@ -277,7 +277,7 @@ flake8-quotes==3.4.0
     # via -r dependencies/pip/dev_requirements.in
 flower==2.0.1
     # via -r dependencies/pip/requirements.in
-formpack @ git+https://github.com/kobotoolbox/formpack.git@f6d4792760291f8d376de7100ce2ffe39edd4366#egg=formpack
+formpack @ git+https://github.com/kobotoolbox/formpack.git@6f0c7601625f6bf962008b518eab934ca0fdbd57#egg=formpack
     # via -r dependencies/pip/requirements.in
 freezegun==1.5.1
     # via -r dependencies/pip/dev_requirements.in
--- a/dependencies/pip/requirements.txt
+++ b/dependencies/pip/requirements.txt
@@ -225,7 +225,7 @@ fido2==2.0.0
     # via django-allauth
 flower==2.0.1
     # via -r dependencies/pip/requirements.in
-formpack @ git+https://github.com/kobotoolbox/formpack.git@f6d4792760291f8d376de7100ce2ffe39edd4366#egg=formpack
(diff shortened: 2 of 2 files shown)


scripts/orval-fix-referenced-additional-properties.js
```diff
@@ -0,0 +1,81 @@
+/* eslint-disable no-console */
+/**
+ * Orval has a bug that fails to generate imports for $ref in additionalProperties.
+ * See https://github.com/orval-labs/orval/issues/1077.
+ * This is a workaround. Remove it once the underlying bug is fixed.
```
Akuukis:
```suggestion
 * Orval doesn't have a feature yet to support `unevaluatedProperties`.
 * See https://github.com/orval-labs/orval/issues/2156.
 * This is a workaround. Remove it once the underlying bug is fixed.
```

scripts/orval-fix-referenced-additional-properties.js
```diff
@@ -0,0 +1,81 @@
...
+  const { header, rest } = splitHeader(source)
+  if (!header) return `${importLine}\n${source}`
+
+  // Ensure one blank line after the Orval header for readability
+  return `${header}\n${importLine}\n${rest}`
+}
+
+for (const file of FILES) {
+  const filePath = path.join(MODELS_DIR, file)
+  if (!fs.existsSync(filePath)) continue
```
Akuukis, line 62: if a file doesn't exist anymore then we'd rather error and fix the hardcoded FILES above.
```suggestion
  if (!fs.existsSync(filePath)) throw new Error(`File ${file} doesn't exist`)
```

scripts/orval-fix-referenced-additional-properties.js
```diff
@@ -0,0 +1,81 @@
...
+
+  const source = fs.readFileSync(filePath, 'utf8')
+
+  const detected = detectInterfaceWithIndex(source)
+  if (!detected) continue
```
Akuukis: if a file doesn't have interface then we'd rather error and fix the hardcoded FILES above.
```suggestion
  if (!fs.existsSync(filePath)) throw new Error(`File ${file} doesn't have an interface within`)
```

Akuukis (approved):
LGTM 🎉 pre-approving, see suggestions inline

PR #6620, opened 2026-01-19
feat(subsequences): add `locale` field to NLP actions documentation DEV-1546

rajpatel24:
### 📣 Summary
Updated the API documentation to explicitly include the `locale` parameter for NLP actions.

### DIFF
--- a/jsapp/js/api/models/_dataResponseAttachmentsItem.ts
+++ b/jsapp/js/api/models/_dataResponseAttachmentsItem.ts
@@ -10,9 +10,15 @@ The endpoints are grouped by area of intended use. Each category contains relate
  * OpenAPI spec version: 2.0.0 (api_v2)
  */
 
-export type PatchedDataSupplementPayloadOneOfManualTranscription = {
-  language: string
-  locale?: string
-  /** @nullable */
-  value: string | null
+export type _DataResponseAttachmentsItem = {
+  download_url: string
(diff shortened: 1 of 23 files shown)


kobo/apps/subsequences/README.md
```diff
@@ -175,7 +175,7 @@ PATCH /api/v2/assets/<asset_uid>/data/<submission_root_uuid>/supplement/
 {
   "_version": "20250820",
   "audio_question": {
-    "manual_transcription": { "language": "en", "value": "My transcript" }
+    "manual_transcription": { "language": "en", "locale": "en-US", "value": "My transcript" }
```
noliveleger, line 190: No need to add `"locale" : "en-US". Please make another example for manual transcription with locale. 

It has be to obvious, it's optional. (You can leave a note in the example about that).

kpi/schema_extensions/v2/data/examples.py
noliveleger, line 1: Let's create an specific example for that. Do not change every example!

kobo/apps/subsequences/README.md
rajpatel24: Added another example.

kpi/schema_extensions/v2/data/examples.py
rajpatel24: Added specific example.

PR #6622, opened 2026-01-19
fix(openAPI): improve schema for assets list response DEV-1580

magicznyleszek:
### 💭 Notes

I've mimicked some other PRs and did some reverse search to find the place to fix this. Not 100% sure it's the proper way. I did add the `parameters=` for asset list API in `kpi/views/v2/asset.py` and then ran `./scripts/generate_api.sh` from withing container, picked out the changes to commit, and then ran `npm run build:orval`.

I need this for DEV-1323, so I did this in the meantime.

### 👀 Preview steps

Testing:
1. Know how defining OpenAPI on Backend should be done
2. Verify I didn't do something weird

### DIFF
--- a/kpi/views/v2/asset.py
+++ b/kpi/views/v2/asset.py
@@ -6,7 +6,12 @@ from operator import itemgetter
 from django.db.models import Count
 from django.http import Http404
 from django.shortcuts import get_object_or_404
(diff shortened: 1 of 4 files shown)


kpi/views/v2/asset.py
noliveleger, line 1: LGTM for this file

PR #6630, opened 2026-01-21
fix(userReports): rename `/api/v2/user-reports/` API fields DEV-1616 DEV-1617

rajpatel24:
### 📖 Description
This PR updates the `UserReports` materialized view to rename two fields and add retention context. These changes ensure the API response field names accurately reflect the underlying data and logic.

Changes:
- `deployed_asset_count` -> `active_project_count`: The previous name implied a count of all deployed assets, but the metric specifically tracks projects that are currently active and not archived. `active_project_count` is a more semantically accurate label for this data point.

- `total_submission_count.all_time` -> `total_submission_count.cumulative`: The label `all_time` was misleading. This value is calculated using the `DailyXFormSubmissionCounter`, from which old records are periodically purged based on the `DAILY_COUNTERS_MAX_DAYS` setting. Consequently, the count reflects a retention window (~12 months) rather than the true history since inception.

- Added `retention_days`: Added a new field to explicitly expose the `DAILY_COUNTERS_MAX_DAYS` setting, allowing super users to understand the time window represented by the cumulative count.


kpi/schema_extensions/v2/user_reports/serializers.py
```diff
@@ -33,6 +33,6 @@
         'current_service_usage': ServiceUsageResponse(),
         'account_restricted': serializers.BooleanField(),
         'asset_count': serializers.IntegerField(),
-        'deployed_asset_count': serializers.IntegerField(),
+        'active_project_count': serializers.IntegerField(),
```
noliveleger, line 50: I cannot comment on other parts of the code but schema and examples in documentation are still showing old properties. 
Moreover (not related to this PR but could be fixed at the same time) - `current_service_usage` has been renamed to `service_usage`. 
Please update example and the schema

noliveleger (changes requested):

PR #6640, opened 2026-01-23
fix(openapi): fix data supplement response qualitative items data schema DEV-1632

Guitlle:
### 🗒️ Checklist

1. [x] run linter locally

### 💭 Notes
Fixes the qualitative data items definitions in the data supplement response schema.

### DIFF
--- a/kpi/schema_extensions/v2/data/extensions.py
+++ b/kpi/schema_extensions/v2/data/extensions.py
@@ -11,122 +11,16 @@ from drf_spectacular.plumbing import (
 from drf_spectacular.types import OpenApiTypes
 
 from kpi.schema_extensions.v2.generic.schema import (
+    GENERIC_DATETIME_SCHEMA,
     GENERIC_INT_SCHEMA,
     GENERIC_STRING_SCHEMA,
-    UUID_STR,
+    GENERIC_UUID_SCHEMA,
 )
-from kpi.utils.schema_extensions.mixins import ComponentRegistrationMixin
+from kpi.utils.schema_extensions.mixins import (
+    ComponentRegistrationMixin,
+    QualComponentsRegistrationMixin,
+)
 from kpi.utils.schema_extensions.url_builder import build_url_type
-
-
-def _register_common_components(component_registrator, auto_schema):
-    references = {}
-    # ---------------------------------------------------------------------
-    # qualInteger
-    #   properties: { value: integer | null }
-    # ---------------------------------------------------------------------
-    references['qual_integer'] = component_registrator._register_schema_component(
-        auto_schema,
-        'DataSupplementManualQualDataInteger',
-        {
-            'type': 'object',
-            'properties': {
-                'value': {
-                    'type': 'integer',
-                    'nullable': True,
-                },
-                'uuid': UUID_STR,
-            },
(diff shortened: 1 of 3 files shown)


kpi/schema_extensions/v2/data/extensions.py
```diff
@@ -165,133 +274,26 @@ def map_serializer(self, auto_schema, direction):
...
-                'properties': {
-                    'value': {
-                        'type': 'array',
-                        'items': {'type': 'string'},
-                    },
-                },
-            },
-        }
-
+    def _qual_schema(self, references):
```
noliveleger, line 182: You could keep it as a property when `_register_common_components()` becomes a protected method of the class.

kpi/schema_extensions/v2/data/extensions.py
```diff
@@ -409,26 +413,25 @@ def _build_transcription_schema(cls, include_status=False):
         version_item = build_object_type(
             additionalProperties=False,
             properties={
-                '_dateCreated': cls.DATETIME,
-                '_uuid': cls.UUID_STR,
-                '_dateAccepted': cls.DATETIME,
-                '_data': cls._get_data_content_schema(include_status=include_status),
+                '_dateCreated': self.DATETIME,
+                '_uuid': self.UUID_STR,
```
noliveleger: Since you created `UUID_STR` inside constants.py, do the same for `DATETIME` (found a better name that only `DATETIME`) and use them everywhere.

kpi/schema_extensions/v2/data/extensions.py
```diff
@@ -13,11 +13,122 @@
 from kpi.schema_extensions.v2.generic.schema import (
     GENERIC_INT_SCHEMA,
     GENERIC_STRING_SCHEMA,
+    UUID_STR,
 )
 from kpi.utils.schema_extensions.mixins import ComponentRegistrationMixin
 from kpi.utils.schema_extensions.url_builder import build_url_type
 
 
+def _register_common_components(component_registrator, auto_schema):
```
noliveleger: Could it be a protected method in a mixin or a base class instead?

kpi/schema_extensions/v2/data/extensions.py
```diff
@@ -393,8 +398,7 @@ def _automatic_transcription_schema(self):
     def _automatic_translation_schema(self):
         return self._build_translation_schema(include_status=True)
 
-    @classmethod
-    def _build_transcription_schema(cls, include_status=False):
+    def _build_transcription_schema(self, include_status=False):
```
noliveleger, line 291: Why did you change all class methods to instance methods? 
We don't the instance anywhere, do we?

noliveleger: As per our discussion: for consistency!

kpi/schema_extensions/v2/data/extensions.py
noliveleger: As per discussion: moved to a mixin.

kpi/schema_extensions/v2/data/extensions.py
noliveleger: done!

PR #6642, opened 2026-01-23
fix(openapi): update `/api/v2/advanced-features` PATCH API schema DEV-1627

rajpatel24:
### 📣 Summary
Add `action` and `question_xpath` to the advanced-features PATCH API OpenAPI schema.

### DIFF
--- a/kpi/schema_extensions/v2/subsequences/fields.py
+++ b/kpi/schema_extensions/v2/subsequences/fields.py
@@ -17,7 +17,13 @@ class AdvancedFeatureCreateResponseParamsField(serializers.JSONField):
 
 class AdvancedFeatureActionField(serializers.ChoiceField):
     def __init__(self, *args, **kwargs):
-        kwargs.setdefault('choices', ACTION_IDS)
-        kwargs.setdefault('allow_null', False)
-        kwargs.setdefault('allow_blank', False)
-        super().__init__(*args, **kwargs)
+        if any(key in kwargs for key in ('choices', 'allow_null', 'allow_blank')):
+            raise TypeError('choices/allow_null/allow_blank cannot be overridden')
+
+        super().__init__(
+            choices=ACTION_IDS,
+            allow_null=False,
+            allow_blank=False,
+            *args,
+            **kwargs,
+        )


kpi/schema_extensions/v2/subsequences/serializers.py
```diff
@@ -14,7 +14,12 @@
 )
 
 AdvancedFeaturePatchRequest = inline_serializer_class(
-    name='AdvancedFeaturePatchRequest', fields={'params': AdvancedFeatureParamsField()}
+    name='AdvancedFeaturePatchRequest',
+    fields={
+        'action': AdvancedFeatureActionField,
```
noliveleger: I think it should be an instance, shouldn't be? `AdvancedFeatureActionField()`? 
It seems that `AdvancedFeatureResponse` is also "wrong".

rajpatel24: For `AdvancedFeatureActionField`, we can't use `AdvancedFeatureActionField()` because that variable already holds an instantiated `ChoiceField`. Calling it would raise `TypeError: 'ChoiceField' object is not callable.`

noliveleger: ~~We learn everyday :-D~~
Would you mind to fix the conflicts before I approve it? 

☝️ , `AdvancedFeatureActionField` is misleading, it should be subclass of a CharField with variables set in its constructor. Otherwise, it should not be capitalized as a class name. 

So please update it!

kpi/schema_extensions/v2/subsequences/fields.py
```diff
@@ -7,6 +7,9 @@ class AdvancedFeatureParamsField(serializers.JSONField):
...
 
 
-AdvancedFeatureActionField = serializers.ChoiceField(
-    choices=ACTION_IDS, allow_null=False, allow_blank=False
-)
+class AdvancedFeatureActionField(serializers.ChoiceField):
+    def __init__(self, *args, **kwargs):
+        kwargs.setdefault('choices', ACTION_IDS)
+        kwargs.setdefault('allow_null', False)
+        kwargs.setdefault('allow_blank', False)
```
noliveleger: I don't think we want those fields to be overwritten.
Why don't pass the values directly to `super()`? 

e.g.: 
``` 
     # remove choices, allow_null, allow_blank if any (or raise an error)
     super().__init__(choices=ACTION_IDS, allow_null=False, allow_blank=False, *args, **kwargs):

```

PR #6644, opened 2026-01-23
fix(api): support repeating groups in bulk submission updates

### LINKED ISSUE #6638: API bulk update returns status 500 when try update repeater fields
Hello! I'm trying to update submitted entries that have repeater fields but this doesn't work and return status 500.

`PATCH https://eu.kobotoolbox.org/api/v2/assets/{asset_uuid}/data/bulk/`

```json

{
  "payload": {
    "submission_ids": [
      1
    ],
    "data": {
      "info_gerais\/status": "01",
      "entrevista\/parte_03\/qd_moradores": [
        {
          "entrevista\/parte_03\/qd_moradores\/c4": "5",
          "entrevista\/parte_03\/qd_moradores\/position": "1",
          "entrevista\/parte_03\/qd_moradores\/sem_estudo": "0",
          "entrevista\/parte_03\/qd_moradores\/fora_escola": "0",
          "entrevista\/parte_03\/qd_moradores\/fora_creche": "0"
        },
       {
          "entrevista\/parte_03\/qd_moradores\/c4": "9",
          "entrevista\/parte_03\/qd_moradores\/position": "2",
          "entrevista\/parte_03\/qd_moradores\/sem_estudo": "0",
          "entrevista\/parte_03\/qd_moradores\/fora_escola": "0",
          "entrevista\/parte_03\/qd_moradores\/fora_creche": "1"
        }
      ]
    }
  }
}

```

Response: 
```json
{
	"error": "Server Error (500)"
}
```

Removing repeater fields from payload works as expected.

emaia:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
The PATCH /api/v2/assets/{asset_uuid}/data/bulk/ endpoint was returning 500 error when the payload contained repeating group fields (arrays). Modified edit_submission_xml() to handle list values by replacing existing repeat elements with new ones from the payload. Fix https://github.com/kobotoolbox/kpi/issues/6638

### 📖 Description
Modified edit_submission_xml() to handle list values by replacing existing repeat elements with new ones from the payload. Added helper functions _edit_repeat_group() and _extract_relative_path() to process nested repeat group structures recursively.

### 💭 Notes
 - Fix 500 error when bulk updating submissions with repeating group fields
 - Add support for array values in the bulk update payload to update repeat groups
 - Maintain backwards compatibility with existing simple field updates

### DIFF
--- a/kpi/tests/test_utils.py
+++ b/kpi/tests/test_utils.py
@@ -814,6 +814,215 @@ class XmlUtilsTestCase(TestCase):
         """
         self.__compare_xml(xml_tostring(xml_parsed), xml_expected)
 
+    def test_edit_submission_xml_repeat_group(self):
+        """Test editing repeat groups in XML"""
+        initial_xml = """
+            <root>
+                <group1>
+                    <repeat_group>
+                        <field1>old1</field1>
+                        <field2>old2</field2>
+                    </repeat_group>
+                    <repeat_group>
+                        <field1>old3</field1>
+                        <field2>old4</field2>
+                    </repeat_group>
+                </group1>
+            </root>
+        """
+        xml_parsed = fromstring_preserve_root_xmlns(initial_xml)
+
+        update_data = [
+            {
+                'group1/repeat_group/field1': 'new1',
+                'group1/repeat_group/field2': 'new2',
+            },
+            {
+                'group1/repeat_group/field1': 'new3',
+                'group1/repeat_group/field2': 'new4',
+            },
+            {
+                'group1/repeat_group/field1': 'new5',
+                'group1/repeat_group/field2': 'new6',
+            },
+        ]
+
+        edit_submission_xml(xml_parsed, 'group1/repeat_group', update_data)
+
+        expected_xml = """
+            <root>
+                <group1>
+                    <repeat_group>
+                        <field1>new1</field1>
+                        <field2>new2</field2>
+                    </repeat_group>
+                    <repeat_group>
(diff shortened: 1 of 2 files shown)


noliveleger:
@greptileai review

noliveleger:
Hello @emaia,

Thanks for digging into this, but I don't think the remove-and-rebuild approach in `_edit_repeat_group` is the right fix.

It deletes **all** existing repeat elements for the given tag and rebuilds them from scratch out of the payload:

```python
existing_repeats = parent_el.findall(repeat_tag)
for existing in existing_repeats:
    parent_el.remove(existing)
```

Each new element is built bare, so only the fields explicitly present in the payload dict survive. Anything else on the original instance, another field not included in that particular edit, a sub-element, an attribute, is silently dropped. Looking at the payload in #6638 itself: it includes `c4`, `position`, `sem_estudo`, `fora_escola`, `fora_creche` for each row. If a future edit only needs to touch `c4` for one row, or if the form has more fields in that group than the ones listed, this implementation wipes the rest instead of leaving it untouched.

We already have a way to edit repeat instances safely without this risk: indexed paths through the existing `get_or_create_element` (`ElementTree.find()`), e.g. `group[3]/repeated`, edit a single node in place and leave every other instance and field alone. That's the behavior we want for a bulk-edit endpoint working from partial payloads.

Please close this PR and rework the fix around indexed paths (`group[n]/field`) instead of remove-and-rebuild. That fixes the 500 from #6638 without the data-loss risk.

PR #6653, opened 2026-01-27
feat(assetNavigator): migrate to TypeScript and remove searches DEV-1323 DEV-1322

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 💭 Notes

`AssetNavigator` is the piece of code that we use in Form Builder to "Add from Library". Previously it was using `searches` and mix of other Reflux actions to display the search with tags and collection filtering. Now it uses orval (react-query). 

It looks a bit different than before, beacuse:

1. It was old and ugly
2. I am using Mantine :)

Changes here:
- Deleted `jsapp/js/components/list.js` as we no longer use any of components from it. They were used only single time in the `assetNavigator.js`
- Deleted bunch of (now) unused bem components (and their css styles)
- `assetNavigator.js` is now `AssetNavigator.tsx` and uses react-query
- Added `AssetNavigatorCard` for displaying small draggable asset in the search results
- Deleted one obsolete reflux action (`listTags`)
- Moved obscure Sentry call to `AssetNavigator`
- Improved `parseTags` to work with `Asset` object (orval's `AssetResponse`)
- Deleted unusued `assetNavExpanded` from `PageStateStore`
- Deleted obsolete `stores.tags` store
- Added a hackfix in `view.surveyApp.coffee` to avoid UI crashing. Not sure why the changes in this PR caused this to happen, given CoffeeScript code was not touched. I spent some time debugging this (crash was easily reproducible), but ended up just having a fix that works, rather than spending more and more hours trying to understand Backbone x CoffeeScript x JQuery split :)
  - The bug was: after creating a new question, the "+" button is focused, you delete the question (while keeping "+" focused) = you get UI crash
- Used our existing `<Badge>` with new `xs` size, as it was quicker than setting up Mantine's `Badge`

### 👀 Preview steps

1. ℹ️ have an account and bunch of projects in the Library:
   - A single question
   - A block of questions
   - A template
   - A collection with single question and block of questions
   - Few of above items with tags
2. Go to Form Builder (either through "NEW" project, or editing some exisisting one)
3. In the header bar on the right use "Add from Library" to open the sidebar
4. 🟢 Compare it to main and see that it still works as expected:
   - Filtering by tags work
   - Filtering through search phrase works
   - Filtering by collection works
   - Drag and dropping search result assets into main area works
   - Expanding details work (each search result will display a list of questions inside and assigned tags)


jsapp/js/editorMixins/AssetNavigator.tsx
```diff
@@ -0,0 +1,215 @@
...
+      ...args: any[]
+    ): JQuery
+  }
+}
+
+const SORTABLE_ITEM_CLASS_NAME = 'asset-navigator-sortable-item'
+
+export default function AssetNavigator() {
+  const [searchQuery, setSearchQuery] = useState('')
+  const [debouncedSearch] = useDebouncedValue(searchQuery, 300)
```
duvld: Why 300? Just curious if this is the standard as mantine's example uses 200

jsapp/js/editorMixins/AssetNavigator.tsx
```diff
@@ -0,0 +1,215 @@
...
+  const [selectedTags, setSelectedTags] = useState<string[]>([])
+  const [selectedCollection, setSelectedCollection] = useState<string | null>(null)
+  const [isExpanded, setIsExpanded] = useState(false)
+
+  // Step 1. Fetch Tags for the MultiSelect filter
+  // Note: if `limit` is too big (e.g. `9999`) it causes a deadly timeout whenever Form Builder displays the aside Library
+  // search, so we use `100`.
+  const { data: tagsResponse } = useTagsList({ limit: 100 })
+  const tagsOptions = useMemo(() => {
```
duvld: What's does using `useMemo` do when the `useTagsList` is already being done through react query?

jsapp/js/editorMixins/AssetNavigator.tsx
```diff
@@ -0,0 +1,215 @@
...
+    }
+  }
+
+  // Step 2. Fetch Collections for the Select filter
+  const { data: collectionsResponse } = useAssetsList({
+    q: COMMON_QUERIES.c,
+    limit: 200,
+    ordering: 'name',
+  })
+  const collectionOptions = useMemo(() => {
```
duvld: Same question as above

jsapp/js/editorMixins/AssetNavigator.tsx
```diff
@@ -0,0 +1,215 @@
...
+    if (!collectionsResponse?.data?.results) return []
+    return collectionsResponse.data.results.map((c: Asset) => ({
+      value: c.uid,
+      label: c.name || t('Unnamed collection'),
+    }))
+  }, [collectionsResponse])
+
+  // Step 3. Fetch Main Assets List
+  const assetQueryParams = useMemo(() => {
+    const parts: string[] = []
```
duvld: Perhaps we can give this variable a more specific name, or at least something like `fields`?

jsapp/js/editorMixins/AssetNavigator.tsx
```diff
@@ -0,0 +1,215 @@
...
+
+    // Include search phrase
+    if (debouncedSearch) {
+      parts.push(`(${debouncedSearch})`)
+    }
+
+    // Include tags filtering
+    if (selectedTags.length > 0) {
+      // BUG: this doesn't work correctly - it does filter one tag, but if multiple are selected it returns zero values
+      // See discussion: https://chat.kobotoolbox.org/#narrow/channel/4-Kobo-Dev/topic/Filtering.20assets.20by.20tags/near/772253
```
duvld: I think we should leave this as DEV-1581 (the linear issue) rather then a direct link to a zulip chat

jsapp/js/editorMixins/AssetNavigator.tsx
```diff
@@ -0,0 +1,215 @@
...
+
+    // Ensure we are only getting library items that make sense here (questions, blocks, and templates)
+    parts.push(COMMON_QUERIES.qbt)
+
+    return {
+      q: parts.join(' AND '),
+    }
+  }, [debouncedSearch, selectedTags, selectedCollection])
+
+  const {
```
duvld, line 97: I think this is some TS stuff I don't really understand. Are we actually assigning anything here to the return of `useAssetsList`, or are we just specifying the object that it returns (I don't think it's this as it sounds pointless)?

duvld: I tried my hand at this in the [TS playground](https://www.typescriptlang.org/play/?#code/MYewdgzgLgBA3gEwIZSQLhhEBbAplACwEswBzAGhiIgBkQkESKqIBRAJ3ZHYF8YBeeMlQY46GAHIAbhJ6VqdBkwwAzJABsIueW07dVGrTwBQoSCHW4AdOpCkAFFjyEmASiA) and I think I'm almost understanding this. Is this a way to cherry pick whatever is in  `data` out of the entire return of `useAssetsList` and assign it to a variable, in my case `something`, in this case `assetsResponse`?

jsapp/js/editorMixins/AssetNavigator.tsx
```diff
@@ -0,0 +1,215 @@
...
+    q: assetQueryParams.q,
+    limit: 200,
+    ordering: '-date_modified',
+  })
+
+  // Step 4. Setup drag and drop of found assets
+  const assetsListRef = useRef<HTMLDivElement>(null)
+  useEffect(() => {
+    activateSortable()
+  }, [assetsResponse])
```
duvld: Assuming I'm correct above, does this mean that `assetsResponse` is acting as a global variable here? What's the difference of doing it this way vs. adding it to the state?

jsapp/js/editorMixins/AssetNavigator.tsx
```diff
@@ -0,0 +1,215 @@
...
+  function activateSortable() {
```
duvld: Is the reason we're not refactoring this jquery logic out becasue it's needed for `sortable` to work?

jsapp/js/editorMixins/AssetNavigator.tsx
```diff
@@ -0,0 +1,215 @@
...
+      deactivate: () => {
+        $el.sortable('cancel')
+      },
+    })
+  }
+
+  return (
+    <Stack gap='sm' h='100%'>
+      {/* Searchbox */}
+      <TextInput
```
duvld, line 136: We have a small regression here, pressing "enter" used to select the first item on the dropdown, if one exists. This functionality is missing now

duvld (commented):
Image

I think we should change the background colour of the tags here.

jsapp/js/editorMixins/AssetNavigator.tsx
```diff
@@ -0,0 +1,215 @@
...
+  const [selectedTags, setSelectedTags] = useState<string[]>([])
+  const [selectedCollection, setSelectedCollection] = useState<string | null>(null)
+  const [isExpanded, setIsExpanded] = useState(false)
+
+  // Step 1. Fetch Tags for the MultiSelect filter
+  // Note: if `limit` is too big (e.g. `9999`) it causes a deadly timeout whenever Form Builder displays the aside Library
+  // search, so we use `100`.
+  const { data: tagsResponse } = useTagsList({ limit: 100 })
+  const tagsOptions = useMemo(() => {
+    if (tagsResponse?.data && 'results' in tagsResponse.data) {
```
Akuukis: please prefer `tagsResponse?.data.status === 200` typeguard, as it reads better and potentially can be expanded to handle each error status as well.

also, same below

jsapp/js/editorMixins/AssetNavigator.tsx
Akuukis: `useMemo` is good for expensive transformations. Imho mapping up to 100 elements is cheap, this one is borderline cheap transformation imho, it's a slight complication and internal comparison also costs. 

Anyway, prefer `select` prop from react-query to achieve the same effect with more optimizations, especially when transformation is a pure function on the response. Search for example `select: transformOrganizationsService,`.

jsapp/js/editorMixins/AssetNavigator.tsx
magicznyleszek: I don't remember :D Looking at the old code, I changed this to 500 to match what we had.

jsapp/js/editorMixins/AssetNavigator.tsx
magicznyleszek: `useMemo` stores the `collectionOptions` array and avoids rerunning that `.map` with every render, it's possibly a bit overkill, but it also keeps some logic encapsulated inside `useMemo`, so it kinda helps out with reding the code :)

jsapp/js/editorMixins/AssetNavigator.tsx
magicznyleszek: Same as below - encapsulate some code and avoid running `.map` on every render.

jsapp/js/editorMixins/AssetNavigator.tsx
magicznyleszek: Renamed to `queryParts`, as it keeps all the parts for the API query, i.e. the x, y, z of `q=x AND y AND z` :)

jsapp/js/editorMixins/AssetNavigator.tsx
magicznyleszek: Good idea!

jsapp/js/editorMixins/AssetNavigator.tsx
magicznyleszek: Yeah it is a shortcut to assign object props to variables. I'm not a big fan of these, but they seem to be prefered in the community. I did copy paste this so it's there :shrug:

jsapp/js/editorMixins/AssetNavigator.tsx
magicznyleszek: It is a variable scoped inside `function AssetNavigator`. Doing it this way makes it easier to reference it than digging it from `useAssetsList` response. We could add it to the state, but then we would need to import and use `useState` and store it inside of `const [x, setX] = setState` - unnecessarily duplicating, as it would give us nothing, and require more code

jsapp/js/editorMixins/AssetNavigator.tsx
magicznyleszek: `select` is interesting! Changed to use it.

jsapp/js/editorMixins/AssetNavigator.tsx
magicznyleszek: `select` is interesting! Changed to use it.

jsapp/js/editorMixins/AssetNavigator.tsx
magicznyleszek: Unfortunately it gives me `Property 'status' does not exist on type 'ErrorDetail'` and requires more typeguards

jsapp/js/editorMixins/AssetNavigator.tsx
magicznyleszek: You mean this?

1. Have multiple items in library, one of them "foo"
2. Focus on "Search…" input
3. Type "foo"
4. 🟢 Notice that there is only one result in the list ("foo" library item)
5. Press "ENTER"
6. 🟢 Notice that "foo" was added to Form Builder

Because it's not workin for me like that on kf.beta

duvld: Just the tags field! I think I highlighted the wrong one here, I'm able to reproduce it on this field here:
Screenshot 2026-02-10 at 17 24 09

Also now that I look at it, do you think we should match the styles of the other fields? I think it'll be a while before we refactor this entire sidebar to match the mantine styles

jsapp/js/editorMixins/AssetNavigator.tsx
Akuukis: Ahh, at first I didn't notice that `data: tagsResponse` is deconstructed above! Let's not do deconstruction because that implicitly disables [`react-query` tracked property optimizations](https://tanstack.com/query/latest/docs/framework/react/guides/render-optimizations#tracked-properties).

This would work then:
```ts
  const response = useTagsList()
  if(response.data?.status === 200) {
    console.log(response.data.data.results) // no TS errors.
  }
```

jsapp/js/editorMixins/AssetNavigator.tsx
magicznyleszek: Yeah it is a shortcut to assign object props to variables. I'm not a big fan of these, but they seem to be prefered in the community. I did copy paste this so it's there 🤷

jsapp/js/editorMixins/AssetNavigator.tsx
magicznyleszek: It is a variable scoped inside `function AssetNavigator`. Doing it this way makes it easier to reference it than digging it from `useAssetsList` response. We could add it to the state, but then we would need to import and use `useState` and store it inside of `const [x, setX] = setState` - unnecessarily duplicating, as it would give us nothing, and require more code

jsapp/js/editorMixins/AssetNavigator.tsx
magicznyleszek: Yes. I mean it would probably be possible to ditch jQuery and rewrite the sortable usage - but let's leave this for other time

jsapp/js/editorMixins/AssetNavigator.tsx
magicznyleszek: I have adjusted the styles to match :)

magicznyleszek: As for the difference in how tags input operates - this is the deafault of Mantine, so I will leave adjusting that to some later time if we decide we don't want the default behaviour. Previous component was a bit better UX wise, but with MultiSelect user can still navigate with keyboard. I will let Tessa know

magicznyleszek: Here: https://chat.kobotoolbox.org/#narrow/channel/7-Kobo-UX.2FUI.2FProduct/topic/Mantine.20Styling.20Questions/near/787931 :)

jsapp/js/editorMixins/AssetNavigator.tsx
magicznyleszek: Ok, changed both places to not use `select` since it's suboptimal. Now the code is simpler :+1: 

Do you think I should also rely on the `.data.status` rather than `isError`?

PR #6655, opened 2026-01-27
fix(projectOwnership): only restart transfers that have been accepted DEV-1643

RuthTurk:
### 📣 Summary
Ensure that project transfers only begin after the recipient has accepted the invitation, preventing projects from being moved before the transfer is authorized.

### Notes
This PR also contains an adjustment to ensure that invitations that are auto-accepted are marked as accepted as soon as they are created to avoid getting stuck in "pending" if the task fails.

### 👀 Preview steps

1. ℹ️ have an account (userA) and a project
2. Transfer the project to userB
3. Do NOT click on the link in the email userB received for the transfer 
4. Wait for several minutes 
5. 🔴 [on release] Refresh userB's project list and see the project has been transferred 
6. 🔴 [on release] Try to click on the link (logged in as userB) and receive an error that the invite has expired 
7. 🟢 [on pr] Refresh userB's project list and see the project has NOT been transferred 
8. 🟢 [on pr] Try to click on the link (logged in as userB) and successfully transfer the project 

To test org invitations:
1. Have an account with an mmo
2. Have another account with a project
3. Invite userB to the mmo
4. Update `project_ownership.utils.create_invite` to error just outside of the transaction block (eg by adding `x = 1/0`)
5. Restart the kpi_worker
6. As userB, accept the invite
7. Remove the error line
8. Restart the kpi_worker
9. Notice the transfer is restarted and successfully completes

### DIFF
--- a/kobo/apps/project_ownership/tasks.py
+++ b/kobo/apps/project_ownership/tasks.py
@@ -12,6 +12,7 @@ from django.urls import reverse
 from django.utils import timezone
 from django.utils.translation import gettext as t
 
+from kobo.apps.project_ownership.models.invite import InviteType
 from kobo.celery import celery_app
 from kpi.utils.mailer import EmailMessage, Mailer
 from .exceptions import AsyncTaskException, TransferStillPendingException
@@ -275,23 +276,30 @@ def task_restarter():
     # PENDING tasks were never started so don't consider them "stuck" even if they are
     # very old
 
-    is_accepted = Q(transfer__invite__status=InviteStatusChoices.ACCEPTED)
-
     stopped_in_progress_tasks = Q(
         date_modified__lte=resume_threshold,
         date_created__gt=stuck_threshold,
         status=TransferStatusChoices.IN_PROGRESS,
     ) & ~Q(status_type=TransferStatusTypeChoices.GLOBAL)
 
+    excluded_pending_user_ownership_invites = Q(
+        transfer__invite__invite_type=InviteType.USER_OWNERSHIP_TRANSFER,
+        transfer__invite__status=InviteStatusChoices.PENDING,
+    )
+
     # Don't start pending attachment tasks independently.
     # The submission task should take care of it
-    still_pending_tasks = Q(
-        status=TransferStatusChoices.PENDING,
-        date_modified__lte=resume_threshold,
-    ) & ~Q(status_type=TransferStatusTypeChoices.ATTACHMENTS)
+    still_pending_tasks = (
+        Q(
+            status=TransferStatusChoices.PENDING,
+            date_modified__lte=resume_threshold,
+        )
+        & ~Q(status_type=TransferStatusTypeChoices.ATTACHMENTS) # noqa
+        & ~excluded_pending_user_ownership_invites # noqa
+    )
 
     queryset = TransferStatus.objects.filter(
-        (stopped_in_progress_tasks | still_pending_tasks), is_accepted
+        stopped_in_progress_tasks | still_pending_tasks
     ).order_by('date_modified')
 
     for transfer_status in queryset:
--- a/kobo/apps/project_ownership/tests/test_tasks.py
+++ b/kobo/apps/project_ownership/tests/test_tasks.py
@@ -9,6 +9,7 @@ from freezegun import freeze_time
 from model_bakery import baker
 
 from kobo.apps.project_ownership.models.choices import InviteStatusChoices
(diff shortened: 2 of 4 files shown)


kobo/apps/project_ownership/utils.py
```diff
@@ -30,7 +30,10 @@ def create_invite(
     TransferStatus = apps.get_model('project_ownership', 'TransferStatus')
 
     with transaction.atomic():
-        invite = InviteModel.objects.create(sender=sender, recipient=recipient)
+        invite = InviteModel(sender=sender, recipient=recipient)
+        if invite.auto_accept_invites:
+            invite.status = InviteStatusChoices.ACCEPTED
+        invite.save()
```
noliveleger: Good idea but unfortunately, It's bypassing the manager so the invite is saved as "User Ownership Transfer" instead of "Org Membership" .

It has to be reverted to `.objects.create`, and maybe pass the status in the manager. 


```diff
@@ -170,7 +170,12 @@ class Invite(AbstractTimeStampedModel):
 class OrgMembershipAutoInviteManager(models.Manager):

     def create(self, **kwargs):
-        return super().create(invite_type=InviteType.ORG_MEMBERSHIP, **kwargs)
+        _ = kwargs.pop('status', InviteStatusChoices.ACCEPTED)
+        return super().create(
+            invite_type=InviteType.ORG_MEMBERSHIP,
+            status=InviteStatusChoices.ACCEPTED,
+            **kwargs
+        )
```

kobo/apps/project_ownership/utils.py
```diff
@@ -45,9 +48,7 @@ def create_invite(
                 )
         TransferStatus.objects.bulk_create(statuses)
 
-    if invite.auto_accept_invites:
-        update_invite(invite, status=InviteStatusChoices.ACCEPTED)
-    else:
+    if not invite.auto_accept_invites:
```
noliveleger: Unfortunately `update_invite` needs to be called. Otherwise, the transfers are not processed until `task_restarter` catches up. 

Should be reverted to what it was to fire transfer as soon as possible.

kobo/apps/project_ownership/tasks.py
```diff
@@ -288,9 +290,11 @@ def task_restarter():
         date_modified__lte=resume_threshold,
     ) & ~Q(status_type=TransferStatusTypeChoices.ATTACHMENTS)
```
noliveleger: I think we should filter down `still_pending_tasks` and not rely on `is_accepted` because as soon as the invite is accepted , it goes in `in_progress` state (whatever the type of the invite), so no tasks would be restarted. 

Something like this.

```python
excluded_pending_user_ownership_invites = Q(
    transfer__invite__invite_type=InviteTypeChoices.USER_OWNERSHIP,
    transfer__invite__status=InviteStatusChoices.PENDING,
)

... 

still_pending_tasks = (
    Q(
        status=TransferStatusChoices.PENDING,
        date_modified__lte=resume_threshold,
    )
    & ~Q(status_type=TransferStatusTypeChoices.ATTACHMENTS)
    & ~excluded_pending_user_ownership_invites
)

queryset = (
    TransferStatus.objects
     .filter(stopped_in_progress_tasks | still_pending_tasks)
     .order_by('date_modified')
)

```

What do you think?

kobo/apps/project_ownership/utils.py
rgraber: I think that's a better idea. It makes sense to default OrgMembershipAutoInvites to accepted on creation

kobo/apps/project_ownership/tests/test_tasks.py
```diff
@@ -32,18 +34,65 @@ def setUp(self):
...
+        )
         return Transfer.objects.create(invite=invite, asset=transfer_asset)
 
-    def test_task_restarter_restarts_pending_transfers(self):
+    def test_task_restarter_ignores_unaccepted_pending_transfers(self):
+        """
+        Ensure that even if a transfer is old/stale, it is ignored
+        if the recipient has not yet clicked the 'Accept' link.
+        """
+        resume_cutoff = timezone.now() - timedelta(minutes=15)
```
noliveleger, line 51: We should also freeze the resume setting on this test to be sure it's not changed in constance with `@override_config`

noliveleger: 🤦  already set at the class level.

noliveleger (approved):
LGTM

PR #6657, opened 2026-01-28
feat(processing): handle new subsequences API on frontend

Akuukis:
### 💭 Notes

Given the breaking changes of API, let's merge the whole frontend changes at once. 

### 👀 Preview steps

1. ℹ️ have an account and a project and a submission with audio
2. click "open" next to the audio
3. click and look around. Add/edit/remove transcription, translation(s) and qualitative analysis questions and answers.

### DIFF
--- a/kobo/settings/dev.py
+++ b/kobo/settings/dev.py
@@ -28,6 +28,3 @@ CALCULATED_HASH_CACHE_EXPIRATION = 5
 # with option `--print-sql`
 SHELL_PLUS_PRINT_SQL_TRUNCATE = None
 RUNSERVER_PLUS_PRINT_SQL_TRUNCATE = None
-
-# manually enable automated transcription/translation
-GS_BUCKET_NAME = 'kobo-transcription-test'


kobo/settings/dev.py
```diff
@@ -28,3 +28,6 @@ def show_toolbar(request):
 # with option `--print-sql`
 SHELL_PLUS_PRINT_SQL_TRUNCATE = None
 RUNSERVER_PLUS_PRINT_SQL_TRUNCATE = None
+
+# manually enable automated transcription/translation
+GS_BUCKET_NAME = 'kobo-transcription-test'
```
Akuukis: @noliveleger from the filename `dev` I assume it doesn't hurt to leave around for easier testing. Or should we remove it?

noliveleger: It does not really hurt indeed but I prefer to have specific kobo settings outside of the repo, so I would remove it (I'm keeping it in my "docker-compose.frontend.custom.yml).

PR #6660, opened 2026-01-28
feat(submissions): support `rootUuid` as `{id}` parameter for data detail endpoints DEV-1657

noliveleger:
### 📣 Summary
Allow data detail endpoints to be accessed using rootUuid as the primary identifier.

### Description
This feature adds support for using `rootUuid` as the `{id} parameter when accessing data detail endpoints. This makes it possible to retrieve a submission directly by its root UUID. The change improves flexibility and consistency when working with submission identifiers, without altering existing behavior for clients that continue to use the original primary key format.

### 👀 Preview steps

2. submit data
3. edit data
4. Go to `/api/v2/assets/{uid_asset}/data/{rootUuid}.json`
4. 🔴 [on main] Receive a 500
5. 🟢 [on PR] Receive a 200 and the correct document

### DIFF
--- a/kpi/views/v2/data.py
+++ b/kpi/views/v2/data.py
@@ -1,5 +1,6 @@
 import copy
 import json
+from typing import Union
 
 import requests
 from django.conf import settings
@@ -242,8 +243,11 @@ class DataViewSet(
 
     def destroy(self, request, pk, *args, **kwargs):
         deployment = self._get_deployment()
-        # Coerce to int because back end only finds matches with same type
-        submission_id = positive_int(pk)
+        submission = self._get_submission_by_id_or_root_uuid(
+            pk, request, fields=['_id'], as_owner=True
+        )
+        # Coerce to int because the back end only finds matches with the same type
+        submission_id = int(submission['_id'])
 
         if deployment.delete_submission(submission_id, user=request.user):
             response = {
@@ -270,7 +274,7 @@ class DataViewSet(
         """
 
         deployment = self._get_deployment()
-        original_submission = self._get_submission_by_id_or_uuid(
+        original_submission = self._get_submission_by_id_or_root_uuid(
             pk, request, fields=['_id'], as_owner=True
         )
         submission_id = original_submission['_id']
@@ -326,7 +330,7 @@ class DataViewSet(
     )
     def enketo_edit(self, request, pk, *args, **kwargs):
 
-        submission = self._get_submission_by_id_or_uuid(
+        submission = self._get_submission_by_id_or_root_uuid(
             pk, request, fields=['_id'], as_owner=True
         )
         submission_id = submission['_id']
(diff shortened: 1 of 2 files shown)


kpi/views/v2/data.py
```diff
@@ -845,6 +831,72 @@ def _get_enketo_link(
...
+        except ValueError:
+            submission_id_or_uuid = add_uuid_prefix(submission_id_or_uuid)
+
+            try:
+                query = json.loads(filters.pop('query', '{}'))
+            except json.JSONDecodeError:
+                raise serializers.ValidationError(
+                    {'query': t('Value must be valid JSON.')}
+                )
+            # TODO Remove `$or` when rootUuid are all backfilled.
```
rgraber: I'm assuming this is still not the case? It would be nice to be able to call this method _get_submission_by_id_or_root_uuid

kpi/tests/test_mock_data_exports.py
```diff
@@ -580,10 +580,10 @@ def test_csv_export_filter_fields(self):
             ]
         }
         expected_lines = [
-            '"start";"end";"Do you descend from an ancestral unicellular organism?";"_uuid";"_index"',
-            '"2017-10-23T05:40:39.000-04:00";"2017-10-23T05:41:13.000-04:00";"No";"48583952-1892-4931-8d9c-869e7b49bafb";"1"',
-            '"2017-10-23T05:41:14.000-04:00";"2017-10-23T05:41:32.000-04:00";"No";"317ba7b7-bea4-4a8c-8620-a483c3079c4b";"2"',
-            '"2017-10-23T05:41:32.000-04:00";"2017-10-23T05:42:05.000-04:00";"Yes";"3f15cdfe-3eab-4678-8352-7806febf158d";"3"',
+            '"start";"end";"Do you descend from an ancestral unicellular organism?";"_uuid";"meta/rootUuid";"_index"',
+            '"2017-10-23T05:40:39.000-04:00";"2017-10-23T05:41:13.000-04:00";"No";"48583952-1892-4931-8d9c-869e7b49bafb";"uuid:48583952-1892-4931-8d9c-869e7b49bafb";"1"',
```
rgraber, line 584: Why did this change?

noliveleger: Short answer: because `meta/rootUuid` is now a required field. 

When you filter the call to the API with `?fields=['mycustomfield']`, `_id`, `_uuid` are always present.
With this PR `meta/rootUuid` too. 

Why? Because we always inject `meta/rootUuid` in the document on the fly if it's not present. We use `_uuid` to set it which could end up with a wrong value when `meta/rootUuid` is not present in the query param `?fields=`. 
On a long term, we want to remove `_uuid` from the list and use `meta/rootUuid` all of the time.

PR #6661, opened 2026-01-28
fix(longRunningMigrations): restart job with smaller batches and improved logging DEV-570

noliveleger:
### 💭 Notes

This fix restarts the long-running migration job using smaller batch sizes after it was stopped due to excessive load on the production database. By processing data in smaller chunks, the migration should reduce pressure on the DB and should lower the risk of performance degradation. Additional logging has been added to provide clearer progress tracking and make it easier to monitor and troubleshoot the job while it runs.


rajpatel24 (approved):
LGTM

PR #6662, opened 2026-01-29
fix(processing): optimistically update automated transx DEV-1675

Akuukis:
### 💭 Notes

Frontend can't guess what backend would return on success, so always pretend that the initial response is "in_progress".

It's ok in both scenarios where backend instantly returns and "overwrites" the latest version, or actually goes into "in_progress" and pushes a new latest version, because frontend cares only about the latest version.

It fixes nothing on it's own, but enables unblocks other issues that will depend on this.

### 👀 Preview steps

1. ℹ️ have an account and a project and a submission with an audio
2. add this console.log to `processing/index.tsx`:
    ```
      console.log(supplement['Use_the_camera_s_mic_ne_to_record_a_sound']['automatic_google_transcription'])
    ```
2. create automatic transcription
4. 🔴 [on main] notice that nothing logs while it loads.
5. 🟢 [on PR] notice that it instantly logs another version with "in_progress" status.

### DIFF
--- a/jsapp/js/api/mutation-defaults/survey-data.ts
+++ b/jsapp/js/api/mutation-defaults/survey-data.ts
@@ -291,6 +291,20 @@ queryClient.setMutationDefaults(
           }
         }
       },
+      /**
+       * Good example of a Direct Update, for cases when mutation returns the whole response.
+       *
+       * Note: use `onSettled` instead of `onSuccess` to be executed AFTER global invalidation logic
+       * in order to cancel it. See more `onSettledInvalidateSnapshots`.
+       *
+       * See more at https://tkdodo.eu/blog/mastering-mutations-in-react-query#direct-updates
+       */
+      onSettled: async (response, error, { rootUuid, uidAsset }, _context) => {
+        if (error) return
+        const queryKey = getAssetsDataSupplementRetrieveQueryKey(uidAsset, rootUuid)
+        queryClient.cancelQueries({ queryKey, exact: true })
+        queryClient.setQueryData(queryKey, response)
+      },
     },
   }),
 )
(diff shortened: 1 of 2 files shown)


jsapp/js/api/mutation-defaults/survey-data.ts
```diff
@@ -204,19 +208,87 @@ queryClient.setMutationDefaults(
...
+                              ...response?.data?.[questionXpath]?.[action],
+                              _versions: [
+                                {
+                                  _uuid: '<server-generated-not-used>',
+                                  _data: {
+                                    ...(datum as PatchedDataSupplementPayloadOneOfAutomaticGoogleTranscription),
+                                    status: 'in_progress',
+                                  },
+                                  _dateCreated: new Date().toISOString(),
+                                }, // Note: this is the actual optimistally added object.
```
magicznyleszek, line 233: I'm testing that `  console.log(supplement['Use_the_camera_s_mic_ne_to_record_a_sound']['automatic_google_transcription'])`. 

When I create automatic transcript, just before the PATCH to `…/supplement/` is done, I see the artifical object being added :+1:

jsapp/js/api/mutation-defaults/survey-data.ts
magicznyleszek, line 88: Previous fix for preselecting newly created automated translation worked on `onSuccess` mutation callback, if we want to have it, should we add it in this file? AFAIK `onSuccess` response contains whole fresh supplement data object.

Akuukis: Right... we should do both! This is an [optimistic update](https://tkdodo.eu/blog/mastering-mutations-in-react-query#optimistic-updates), and let's add back the [direct update](https://tkdodo.eu/blog/mastering-mutations-in-react-query#direct-updates).

PR #6672, opened 2026-01-30
fix(processing): prev/next arrows for edited submissions DEV-1645

Akuukis:
### 💭 Notes

Use `meta/rootUuid` instead of `_uuid` to handle edited submissions.


jamesrkiger (approved):
lgtm

PR #6679, opened 2026-02-01
fix(map): display correct results when filtering map DEV-1446 DEV-1555 DEV-1553 DEV-1564

duvld:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Fixes the map display to show only submissions from the selected question

### 💭 Notes
To save myself the time of reiterating every discovery from the time I started working on this I'll leave a link to the zulip discussion: 
https://chat.kobotoolbox.org/#narrow/channel/7-Product/topic/Map.20view.2C.20and.20harmfulness.20of.20.60_geolocation.60/near/783626
https://chat.kobotoolbox.org/#narrow/channel/7-Product/topic/Map.20view.2C.20and.20harmfulness.20of.20.60_geolocation.60/near/793770

This PR fixes 4 bugs (and therefore has many preview steps to check) because it removes the frontend dependency on `_geolocation` caused multiple issues. 

### 👀 Preview steps

**It may be clearer to follow the reproduction steps outlined in the respective linear tasks as it comes with a survey and a detailed explanation of the issue, but the following steps should also work**

1. ℹ️ have an account and **create a new project** (this step is important to ensure steps 6. and 7. go as planned)
2. create 2 distinct geopoint questions, let's call them question A and question B
3. create 3 submissions, all 3 answering question A and **only** one submission answering question B
    -- to make some later steps more clear, answer question A with points very close together, like clicking on the name of a city for example
5. Navigate to the map data display
6. Click in the settings cog
7. 🔴 [on main] notice that the modal doesn't have any radio buttons selected by default
8. 🟢 [on PR] notice that the modal has the first question selected by default (DEV-1553)
9. Select **question A** and save
7. 🔴 [on main] notice that saving the modal does nothing on first save, and another one is required to actually update the points
8. 🟢 [on PR] notice that clicking "Save" once updates the points immediately (DEV-1564) 
11. Do step 8 again
6. 🔴 [on main] notice that the points have come together to a single point
13. 🟢 [on PR] notice that points stay in the expected position (DEV-1555)
12. Select question B
9. 🔴 [on main] notice that the map doesn't only display the one submissions (displays submissions from both questions)
13. 🟢 [on PR] notice that the map displays only the one submission that answers B (DEV-1446)


jsapp/js/components/map/index.tsx
```diff
@@ -414,7 +414,16 @@ class FormMap extends React.Component<FormMapProps, FormMapState> {
...
+
+    // Note: the selected question is considered a "map style" and actually patched into the asset itself. This creates
+    // the possibility that a question can be pre-selected, and not display all geopoints by default.
+    let selectedQuestion: string | null = null
+    if (this.state.overridenStyles?.selectedQuestion) {
+      selectedQuestion = this.state.overridenStyles.selectedQuestion
+    } else {
+      // After fixing DEV-1446, this line began to work "properly" and sets the previous selected question as default
+      selectedQuestion = this.props.asset.map_styles.selectedQuestion || null
+    }
```
Akuukis: With `??` it's much cleaner now.

```suggestion
    // Note: the selected question is considered a "map style" and actually patched into the asset itself. This creates
    // the possibility that a question can be pre-selected, and not display all geopoints by default.
    // After fixing DEV-1446, the second line began to work "properly" and sets the previous selected question as default
    let selectedQuestion = this.state.overridenStyles?.selectedQuestion ?? this.props.asset.map_styles.selectedQuestion ?? null
```

jnm:
@duvld IDK if it would be a hassle, but if it's easy to get Leaflet to show the coordinates on the markers (maybe by stuffing the coords into `title` as a string?), that might really help when verifying the behavior:
https://github.com/kobotoolbox/kpi/blob/efb66b17c529f5c105a77cbdf230e2cbe7bce86a/jsapp/js/components/map/index.tsx#L581-L585

Leaflet docs for marker title: https://leafletjs.com/reference.html#marker-title

jsapp/js/components/map/index.tsx
```diff
@@ -697,11 +715,12 @@ class FormMap extends React.Component<FormMapProps, FormMapState> {
   buildHeatMap(map: L.Map) {
     const heatmapPoints: Array<[number, number, number]> = []
     this.state.submissions.forEach((item) => {
-      if (checkLatLng(item._geolocation)) {
-        const geo0 = item._geolocation[0]
-        const geo1 = item._geolocation[1]
-        if (geo0 !== null && geo1 !== null) {
-          heatmapPoints.push([geo0, geo1, 1])
+      let parsedCoordinates: string[] = []
```
jnm: could be a const and directly assigned as it's declared, i think?

jsapp/js/components/map/index.tsx
```diff
@@ -697,11 +715,12 @@ class FormMap extends React.Component<FormMapProps, FormMapState> {
...
+      parsedCoordinates = String(item[this.state.foundSelectedQuestion as string]).split(' ')
+
+      if (this.state.foundSelectedQuestion && checkLatLng(parsedCoordinates)) {
+        if (!!parsedCoordinates.length) {
+          heatmapPoints.push([Number.parseFloat(parsedCoordinates[0]), Number.parseFloat(parsedCoordinates[1]), 1])
```
jnm: since the split, length check, and float parsing happens twice (and maybe will need to happen more in the future?) i think it'd be helpful to standardize best-practice logic for doing this within a helper function, maybe called `parseLatLng()`. you could have it return the floats if they're present or some other appropriate and known value if they're not, so that basically that one helper could do the job of `checkLatLng()` plus the splitting and parsing stuff

jsapp/js/components/map/index.tsx
```diff
@@ -414,18 +418,37 @@ class FormMap extends React.Component<FormMapProps, FormMapState> {
...
+      // 2. If the user has selected a question before (at any point), the `map_styles` value of the asset is patched
+      //    and we should use this if it exists. Will happen on every refresh if the user has ever selected a question
+      selectedQuestion = this.props.asset.map_styles.selectedQuestion
+    } else if (this.props.asset.content?.survey) {
+      // 3. If the user has never selected a question before, a "default" needs to be selected. Since after DEV-1446 we
+      //    don't use `_geolocation`, the frontend has to find the first geopoint question and set it as the default
+      //    regardless of if that question is answered.
+      const firstGeopoint = this.props.asset.content.survey.find(
+        (question) => question.type && question.type === 'geopoint',
+      )
```
jnm: i think i'd argue for a helper for this since the logic is repeated and it might be something used again by other parts of the code, and we should standardize on your new logic here

jsapp/js/components/map/index.tsx
```diff
@@ -551,7 +563,11 @@ class FormMap extends React.Component<FormMapProps, FormMapState> {
 
     this.state.submissions.forEach((item) => {
       let markerProps = {}
-      if (checkLatLng(item._geolocation)) {
+
+      const parsedCoordinates: number[] = parseLatLng(item, this.state.foundSelectedQuestion)
+      // Safe to cast `null` as a string here as this will result in Array['undefined'] if there are no geopoint submissions
```
magicznyleszek: I don't understand this comment. Is this some leftover? I don't see casting, null or string anywhere here :D

jsapp/js/components/map/MapSettings.tsx
```diff
@@ -98,6 +98,17 @@ export default class MapSettings extends React.Component<MapSettingsProps, MapSe
       Object.assign(mapStyles, this.props.overridenStyles)
     }
 
+    // Force first geopoint question to be selected if there otherwise would be none
+    if (!mapStyles.selectedQuestion) {
+      const firstGeopoint = props.asset.content?.survey?.find(
+        (question) => question.type && question.type === 'geopoint',
```
magicznyleszek: Shouldn't you use `findFirstGeopoint` here?

PR #6682, opened 2026-02-03
fix(processing): update audio player source when media url change DEV-1710

pauloamorimbr:
### 💭 Notes
This PR fixes the issue where the audio file is not being updated when navigating through records inside of the processing view by using the up and down buttons in the header.

### 👀 Preview steps
1. Have a project with an audio question
2. Have multiple records
3. Open an audio for NLP
4. Use the up and down buttons in the header to navigate to other records
4. 🔴 [on base] notice that the audio file won't update
5. 🟢 [on PR] notice that the audio file updates as expected


Akuukis (approved):
LGTM, not tested.

PR #6685, opened 2026-02-04
fix(organizations): fix permission revocation logic and OOM errors in long running migration 0015 DEV-1714

rajpatel24:
### 📖 Description
Migration `0015_fix_duplicate_organizations` stalled on production due to two primary issues:
- Logic Error: The permission revocation logic incorrectly targeted organization owners during duplicate resolution, causing them to lose access to their own projects.
- Performance Bottleneck: Processing large permission sets (e.g., ~90k records) in a single transaction caused memory exhaustion (OOM) and continuous pod restarts.

To resolve these issues, the following changes were implemented:
- Added a check in `revoke_org_asset_perms()` and the 0015 job to ensure owners are excluded from the revocation process.
- Refactored the permissions deletion logic to process permission removals in chunks of 1000, preventing memory spikes and stabilizing the task environment.

### DIFF
--- a/kobo/apps/organizations/tests/test_fix_duplicate_organizations.py
+++ b/kobo/apps/organizations/tests/test_fix_duplicate_organizations.py
@@ -161,10 +161,26 @@ class TestFixDuplicateOrgs(TestCase):
         asset = Asset.objects.create(name='Owner Project', owner=self.someuser)
 
         # Get all permissions for the asset
-        initial_perms = set(ObjectPermission.objects.filter(
+        initial_perms = ObjectPermission.objects.filter(
             user=self.someuser,
-            asset=asset
-        ).values_list('permission_id', flat=True))
+            asset=asset,
+            deny=False,
+        ).values('permission_id', 'permission__codename')
+
+        expected_permissions = [
+            'add_submissions',
+            'change_asset',
+            'change_submissions',
+            'delete_submissions',
+            'manage_asset',
+            'validate_submissions',
+            'view_asset',
+            'view_submissions',
+        ]
+        initial_codenames = sorted([
+            obj_perm['permission__codename'] for obj_perm in initial_perms
+        ])
+        assert expected_permissions == initial_codenames
 
         job.run()
 
@@ -176,9 +192,13 @@ class TestFixDuplicateOrgs(TestCase):
         # Verify that someuser still has all their initial permissions on the asset
         final_perms = set(ObjectPermission.objects.filter(
             user=self.someuser,
-            asset=asset
+            asset=asset,
+            deny=False,
         ).values_list('permission_id', flat=True))
-        self.assertEqual(initial_perms, final_perms)
+
+        assert set([
+            obj_perm['permission_id'] for obj_perm in initial_perms
+        ]) == final_perms
 
     def _create_organization_for_user(self, user, mmo_override=False):
         org = Organization.objects.create(name='Org', mmo_override=mmo_override)
--- a/kobo/settings/base.py
+++ b/kobo/settings/base.py
@@ -1963,8 +1963,6 @@ ENDPOINT_CACHE_DURATION = env.int('ENDPOINT_CACHE_DURATION', 60 * 15)  # 15 minu
 
 ENV = None
 
-DEFAULT_BATCH_SIZE = 1000
-
 # The maximum size in bytes that a request body may be before a
 # SuspiciousOperation (RequestDataTooBig) is raised
 DATA_UPLOAD_MAX_MEMORY_SIZE = 10485760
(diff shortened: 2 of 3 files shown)


kobo/apps/long_running_migrations/jobs/0015_fix_duplicate_organizations.py
```diff
@@ -103,7 +103,9 @@ def run():
 
             for org in to_remove:
                 with transaction.atomic():
-                    revoke_org_asset_perms(org, [uid])
+                    if org.owner_user_object.pk != uid:
```
noliveleger: I forgot to mention it in previous PR but `uid` is confusing. Let's change it to `user_id`.

kobo/apps/organizations/tests/test_fix_duplicate_organizations.py
```diff
@@ -149,6 +149,28 @@ def test_asset_transfer_to_kept_organization(self):
...
+        """
+        Verify that when a duplicate organization is removed, if the user being
+        processed is the owner of that org, they DO NOT lose their own permissions
+        """
+        # Give someuser a second organization where they are also the owner
+        self._create_organization_for_user(user=self.someuser)
+
+        # Create an asset owned by someuser and assign them a permission
+        asset = Asset.objects.create(name='Owner Project', owner=self.someuser)
+        asset.assign_perm(self.someuser, 'view_asset')
```
noliveleger: This should be useless IMO. When the asset is created, permissions are calculated and added to the owner.

kobo/apps/organizations/tests/test_fix_duplicate_organizations.py
```diff
@@ -149,6 +149,28 @@ def test_asset_transfer_to_kept_organization(self):
...
+
+        job.run()
+
+        # Verify that only one organization remains for someuser
+        self.assertEqual(
+            OrganizationUser.objects.filter(user=self.someuser).count(), 1
+        )
+
+        # Verify that someuser still has their permission on the asset
+        self.assertTrue(asset.has_perm(self.someuser, 'view_asset'))
```
noliveleger: We should validate that all permissions for owner are present.

kobo/apps/organizations/utils.py
```diff
@@ -45,9 +45,23 @@ def revoke_org_asset_perms(organization: Organization, user_ids: list[int]):
     the organization.
     """
     Asset = apps.get_model('kpi', 'Asset')  # noqa
+    owner_id = organization.owner_user_object.pk
+
+    # Filter out the owner from user_ids to prevent them from
+    # revoking their own access
+    safe_user_ids = [uid for uid in user_ids if uid != owner_id]
```
noliveleger: Do not use `uid` for `user_id`. we use `uid` everywhere for Unique identifier, it's confusing IMO.

kobo/apps/organizations/utils.py
```diff
@@ -45,9 +45,23 @@ def revoke_org_asset_perms(organization: Organization, user_ids: list[int]):
...
-    ObjectPermission.objects.filter(
-        asset_id__in=subquery, user_id__in=user_ids
-    ).delete()
+    perms_to_delete = ObjectPermission.objects.filter(
+        asset_id__in=subquery, user_id__in=safe_user_ids
+    )
+
+    BATCH = 1000
+    while perms_to_delete.exists():
+        pks_to_delete = list(perms_to_delete.values_list('pk', flat=True)[:BATCH])
```
noliveleger: Let's just use 1000 instead BATCH. The constant is not reused anywhere. 

Otherwise, let's move `DEFAULT_BATCHSIZE = 1000` from MongoHelper.py to base.py and use it as the DEFAUT_BATCH_SIZE (<- watchout the extra underscore, to be consistent with other settings).

kobo/apps/organizations/tests/test_fix_duplicate_organizations.py
rajpatel24: Right!

kobo/apps/organizations/tests/test_fix_duplicate_organizations.py
```diff
@@ -168,8 +173,12 @@ def test_owner_does_not_lose_permissions_on_removal(self):
...
         )
 
-        # Verify that someuser still has their permission on the asset
-        self.assertTrue(asset.has_perm(self.someuser, 'view_asset'))
+        # Verify that someuser still has all their initial permissions on the asset
+        final_perms = set(ObjectPermission.objects.filter(
+            user=self.someuser,
+            asset=asset
+        ).values_list('permission_id', flat=True))
+        self.assertEqual(initial_perms, final_perms)
```
noliveleger: I would prefer to see them hardcoded because if `initial_perms` is wrong, `final_perms` would be too.

kobo/settings/base.py
```diff
@@ -1963,6 +1963,8 @@ def dj_stripe_request_callback_method():
 
 ENV = None
 
+DEFAULT_BATCH_SIZE = 1000
```
noliveleger: Please group it with others `*_BATCH_SIZE` variables.

noliveleger: ✅

kobo/apps/organizations/tests/test_fix_duplicate_organizations.py
noliveleger: ✅

PR #6686, opened 2026-02-04
fix(subsequences): fix translation blocking to allow transcript replacement DEV-1715

rajpatel24:
### 📣 Summary
Prevents new translations from being created when a transcript has been deleted,  while ensuring they remain available  if a new transcript is provided later.

### 📖 Description
This PR introduces a robust validation rule for transcriptions and translations to ensure that deleted data is never accidentally used for new translations.

The system now utilizes a timestamp based arbitration logic. It compares the most recent time a transcript was deleted against the most recent time a transcript was accepted across all sources (Manual or Automatic).
- If the latest event is a deletion: The system considers the question to have no valid transcript and will block any new translation requests.
- If the latest event is an acceptance: Even if a deletion occurred previously, the system recognizes that the data has been replaced and allows translations to proceed normally using the newest valid text.

This ensures data integrity by preventing the use of ghost transcripts while maintaining a flexible workflow for users who need to delete or re-transcribe their audio.

### 👀 Preview steps

1. ℹ️ Have an account and a project with an audio submission.
2. Using the UI or Swagger, first add an automatic transcript.
3. Create a translation (e.g., French) based on that transcript.
4. Delete the automatic transcript and then add a new manual transcript.
5. Attempt to add a new translation (e.g., Spanish).
6. 🔴 [on base branch] Notice that this throws a "Cannot translate without a transcription" error.
7. 🟢 [on PR] Notice that the translation is added correctly because the system recognizes the new manual transcript.

### DIFF
--- a/kobo/apps/subsequences/actions/mixins.py
+++ b/kobo/apps/subsequences/actions/mixins.py
@@ -69,55 +69,47 @@ class RequiresTranscriptionMixin:
         Raises:
           - TranscriptionNotFound: if no accepted transcript is available.
         """
+        latest_version = latest_accepted_dt = latest_version_action_id = None
+        latest_deletion_dt = None
+
         if 'value' in action_data and action_data['value'] is None:
             return action_data
 
-        global_latest_version, global_latest_dt = None, None
         for action_id, action_supplemental_data in self._action_dependencies.items():
-
             versions = action_supplemental_data.get(self.VERSION_FIELD) or []
             if not versions:
                 continue
 
-            current_latest = versions[0]
-            created_raw = current_latest.get(self.DATE_CREATED_FIELD)
-            if not created_raw:
-                continue
-
-            created_dt = parser.parse(created_raw)
-            if global_latest_dt is None or created_dt > global_latest_dt:
-                global_latest_dt = created_dt
-                global_latest_version = current_latest
-
-        if global_latest_version:
-            latest_data = global_latest_version.get(self.VERSION_DATA_FIELD, {})
-            is_deleted = (
-                latest_data.get('status') == 'deleted' or
-                ('status' not in latest_data and latest_data.get('value') is None)
-            )
-
-            # Do not consider deleted transcriptions
-            if is_deleted:
-                raise TranscriptionNotFound
-
-        latest_version = latest_accepted_dt = latest_version_action_id = None
-
-        for action_id, action_supplemental_data in self._action_dependencies.items():
-            versions = action_supplemental_data.get(self.VERSION_FIELD) or []
             for version in versions:
-                # Skip versions without an acceptance timestamp.
-                accepted_raw = version.get(self.DATE_ACCEPTED_FIELD)
-                if not accepted_raw:
-                    continue
+                version_data = version.get(self.VERSION_DATA_FIELD, {})
 
-                accepted_dt = parser.parse(accepted_raw)
+                is_deleted = (
+                    ('status' not in version_data and version_data.get('value') is None)
+                    or version_data.get('status') == 'deleted'
+                )
 
-                if latest_accepted_dt is None or accepted_dt > latest_accepted_dt:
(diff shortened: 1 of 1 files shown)


kobo/apps/subsequences/actions/mixins.py
```diff
@@ -69,28 +69,41 @@ def attach_action_dependency(self, action_data: dict):
...
             )
 
             # Do not consider deleted transcriptions
             if is_deleted:
                 raise TranscriptionNotFound
 
+        latest_version = latest_accepted_dt = latest_version_action_id = None
+
+        for action_id, action_supplemental_data in self._action_dependencies.items():
+            versions = action_supplemental_data.get(self.VERSION_FIELD) or []
```
noliveleger: Let's avoid a double loop. I think we can do everything inside the same loop. 
We can try to do the same logic we did for the data table. Take dateCreated into account when it's a deletion and compare it to the most recent accepted version. 


What about this diff (against the `refactor-subsequences-2025` branch)

```diff
diff --git a/kobo/apps/subsequences/actions/mixins.py b/kobo/apps/subsequences/actions/mixins.py
index afa841e13..74faca1ff 100644
--- a/kobo/apps/subsequences/actions/mixins.py
+++ b/kobo/apps/subsequences/actions/mixins.py
@@ -70,6 +70,7 @@ class RequiresTranscriptionMixin:
           - TranscriptionNotFound: if no accepted transcript is available.
         """
         latest_version = latest_accepted_dt = latest_version_action_id = None
+        latest_deletion_dt = None
 
         # If deletion has been requested, we do not want to attach any dependency.
         if 'value' in action_data and action_data['value'] is None:
@@ -81,30 +82,36 @@ class RequiresTranscriptionMixin:
             if not versions:
                 continue
 
-            latest_data = versions[0].get(self.VERSION_DATA_FIELD)
-            is_deleted = (
-                latest_data.get('value') is None
-                or latest_data.get('status') == 'deleted'
-            )
-
-            # Do not consider deleted transcriptions
-            if is_deleted:
-                raise TranscriptionNotFound
-
             for version in versions:
-                # Skip versions without an acceptance timestamp.
-                accepted_raw = version.get(self.DATE_ACCEPTED_FIELD)
-                if not accepted_raw:
-                    continue
-
-                accepted_dt = parser.parse(accepted_raw)
-
-                if latest_accepted_dt is None or accepted_dt > latest_accepted_dt:
-                    latest_accepted_dt = accepted_dt
-                    latest_version = version
-                    latest_version_action_id = action_id
-
-        if latest_version is None:
+                version_data = version.get(self.VERSION_DATA_FIELD, {})
+                is_deleted = (
+                    'status' not in version_data and version_data.get('value') is None
+                    or version_data.get('status') == 'deleted'
+                )
+
+                if is_deleted:
+                    # Track the latest deletion date
+                    created_raw = version[self.DATE_CREATED_FIELD]
+                    created_dt = parser.parse(created_raw)
+                    if latest_deletion_dt is None or created_dt > latest_deletion_dt:
+                        latest_deletion_dt = created_dt
+                else:
+                    # Skip versions without an acceptance timestamp.
+                    accepted_raw = version.get(self.DATE_ACCEPTED_FIELD)
+                    if not accepted_raw:
+                        continue
+
+                    accepted_dt = parser.parse(accepted_raw)
+
+                    if latest_accepted_dt is None or accepted_dt > latest_accepted_dt:
+                        latest_accepted_dt = accepted_dt
+                        latest_version = version
+                        latest_version_action_id = action_id
+
+        if (
+            latest_version is None
+            or (latest_deletion_dt and latest_deletion_dt > latest_accepted_dt)
+        ):
             raise TranscriptionNotFound
 
         latest_version_data = latest_version.get(self.VERSION_DATA_FIELD, {})

```

rajpatel24: I think this is more efficient and actually more robust. Updated the PR accordingly!

PR #6696, opened 2026-02-05
fix(sidebar): settings styling and options sorting

pauloamorimbr:
### 💭 Notes
- Sidebar settings style was adjusted to be closer to current production
- Order of the items in sidebar settings was adjusted to match current production 

### 👀 Preview steps
1. ℹ️ have translations and questions
2. open sidebar settings
4. 🔴 [on base] notice that questions border is darker and spaces are smaller
5. 🟢 [on PR] notice that questions border is lighter and spaces are larger
6. do that another thing
4. 🔴 [on base] notice that options order is `Submission Data`, `Audio`, ...translations
5. 🟢 [on PR] notice that options order is `Audio`, `Submission Data`, ...translations

### DIFF
--- a/jsapp/js/components/processing/SingleProcessingSidebar/sidebarDisplaySettings.tsx
+++ b/jsapp/js/components/processing/SingleProcessingSidebar/sidebarDisplaySettings.tsx
@@ -213,8 +213,8 @@ export default function SidebarDisplaySettings({
                         <ScrollArea
                           h={160}
                           p='xs'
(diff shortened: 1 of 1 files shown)


pauloamorimbr:
> On base
> 
> Image
> on branch
> 
> Image
> I think the border has to be more lighter!

Good catch. For some reason it changes when is opened but is reverting when closed and reopened! Let me see what's going on! Thanks for the catch!

PR #6697, opened 2026-02-05
fix(subsequences): fix null bucket name infinite processing DEV-1720

rajpatel24:
### 📖 Description
This PR fixes an issue where automatic transcription and translation subsequences could remain indefinitely in the "in_progress" state when `GS_BUCKET_NAME` was not configured.

The backend now detects missing Google Cloud Storage configuration early and returns a clear failed status instead of triggering async processing and background polling. This prevents infinite retries caused by configuration errors and makes failures explicit and actionable.

### DIFF
--- a/kobo/apps/subsequences/integrations/google/google_transcribe.py
+++ b/kobo/apps/subsequences/integrations/google/google_transcribe.py
@@ -126,11 +126,6 @@ class GoogleTranscriptionService(GoogleService):
         return attachment.get_transcoded_audio('flac', include_duration=True)
 
     def process_data(self, xpath: str, params: dict) -> dict:
-        if settings.GS_BUCKET_NAME is None:
-            return {
-                'status': 'failed',
-                'error': 'GS_BUCKET_NAME not configured'
-            }
         source_language = params.get('locale') or params['language']
         cache_key = self._get_cache_key(xpath, source_language, target_lang=None)
         if cache.get(cache_key):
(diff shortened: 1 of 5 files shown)


kobo/apps/subsequences/integrations/google/google_transcribe.py
```diff
@@ -126,6 +126,11 @@ def get_converted_audio(
         return attachment.get_transcoded_audio('flac', include_duration=True)
 
     def process_data(self, xpath: str, params: dict) -> dict:
+        if settings.GS_BUCKET_NAME is None:
+            return {
+                'status': 'failed',
+                'error': 'GS_BUCKET_NAME not configured'
+            }
```
noliveleger: Please make a private method in `GoogleService` base class ~~which could be called in both classes~~.

EDIT: I made the change and called it from the constructor.

kobo/apps/subsequences/integrations/google/google_translate.py
```diff
@@ -172,6 +172,11 @@ def process_data(self, xpath: str, params: dict) -> dict:
         """
         Translates the value for a given xpath and its json values.
         """
+        if settings.GS_BUCKET_NAME is None:
+            return {
+                'status': 'failed',
+                'error': 'GS_BUCKET_NAME not configured'
+            }
```
noliveleger: Please make a private method in `GoogleService` base class ~~which could be called in both classes~~.

EDIT: I made the change and called it from the constructor.

jamesrkiger (approved):
Tested with frontend locally. Looks good to me.

PR #6705, opened 2026-02-09
feat(releases): branch new release after prev tag DEV-1468

Akuukis:
### 💭 Notes

Instead of cutting a new release every 4 weeks, let's cut a new release dynamically next Tuesday after the previous release is tagged. To "scrap" the release, just delete the branch manually, and the next release will get cut next Tuesday.

### 👀 Preview steps

Can't really preview outside of a release branch. Please read it carefully.

### DIFF
--- a/.github/workflows/release-1-branch.yml
+++ b/.github/workflows/release-1-branch.yml
@@ -14,30 +14,34 @@ on:
 jobs:
 
 
-  is-release-week:
+  current-branch:
     runs-on: ubuntu-24.04
     outputs:
-      is_release_week: ${{ steps.week.outputs.is_release_week }}
+      current_branch: ${{ steps.branch.outputs.current_branch }}
     steps:
-    - name: check if week number is even
-      id: week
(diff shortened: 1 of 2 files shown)


.github/workflows/release-1-branch.yml
```diff
@@ -14,74 +14,48 @@ on:
...
 
   version:
     needs:
-      - create-current-branch
+      - current-branch
     secrets: inherit
     uses: './.github/workflows/find-releases.yml'
     with:
-      current_branch: ${{ needs.create-current-branch.outputs.current_branch }}
+      current_branch: ${{ needs.current-branch.outputs.current_branch }}
```
Akuukis, line 37: Note: the main change is that here the `find-releases.yml` is called with a not-yet created branch. That will work, you can try that locally with `./scripts/find_releases.sh "release/2.026.07"`. Afterwards, the branch will get actually created, or not, based on the outcome. And then zulip notified accordingly. This actually simplifies the workflow.

PR #6588, opened 2026-01-07
refactor(formBuilder): migrate editableForm and connected files to TypeScript DEV-1520 DEV-1338

magicznyleszek:
### 📣 Summary

Internal refactor of piece(s) of code that connects Form Builder (Backbone x CoffeeScript app) with the rest of the UI (React x JavaScript x TypeScript app).

### 💭 Notes

The touched parts are the connection between Backbone and React, so I've started with migrating `editableForm.js` (now `EditableForm.tsx`) and then all the related parts, mainly:
- `cascadeMixin.js` - integrated into `EditableForm.tsx`
- `formEditors.tsx` are using EditableForm in a new way (EditableForm is no longer a React mixin)
- `formBuilderUtils.tsx`
- `_xlform.init.coffee` 
  - is being imported into `EditableForm.tsx` and is itself a sort of root gathering of all coffee code
  - to make it useful (i.e. avoid it being `any`) I've added `.d.ts` file for it and for all the files that it is importing etc. (this is the main reason this PR adds so much LOC)
    - the types were build partially by hand and partially with AI, because I don't really really need them being nice, I just need them be Good Enough™ and I feel they are
- `surveyScope.tsx` is a companion piece to `EditableForm.tsx` and Backbone code is using it
- Improved types for `actions.js` (`actions.d.ts` file)
- Tiny type improvements in `dataInterface.ts`
- In `stores.d.ts` 
  - I removed `sessions` as it is a separate store (change made long time ago, this is a cleanup that was omitted)
  - Improved types for `surveyState`
- Marked some more dead code in coffee code (with comments) 

History: This started during Christmas time when most developers had ooo time and as preparation to fix a bug. Bug was fixed accidently by doing other work, but nevertheless I've finished the TS migration work.

### 👀 Preview steps

This doesn't touch any particular functionality, but rather a wrapper for Form Builder, so any Form Builder related functionalities might be affected  by this PR. It woul be best to test FB thoroughly. By looking at the code, the parts that for sure needs to be tested are:
- Creating a new Library Item
- Importing a Library Item into form
- Cascading select: https://support.kobotoolbox.org/cascading_select.html
- Locking profiles and templates: https://support.kobotoolbox.org/library_locking.html
- Creating new form
- Editing existing form
- Editing form style (from Layout & Settings)
- Grouping rows
- Reordering rows
- Changing form name through Form Builder
- Previewing form
- Question matrix: https://support.kobotoolbox.org/matrix_response.html

PR #6708, opened 2026-02-10
refactor(editableForm): change into functional component DEV-1338 DEV-1337

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 💭 Notes

This is a followup to #6588. Previous PR moved things to TypeScript and merged some code into single component. This one changes that component into a functional component to allow for easier use of orval

Notable changes:
- Changed `routerWillLeave` to `useBeforeUnload` and `useBlocker`
- Changed `setState` to `useState`
  - Had some weird issues with using `setState({...state})`, but they went away after switching to `setState((state) => ({...state}))`
- Changed `componentWillMount` to `useEffect`

### 👀 Preview steps

Similarly to #6588 this doesn't touch any particular functionality, but rather a wrapper for Form Builder, so any Form Builder related functionalities might be affected  by this PR. It would be best to test FB. By looking at the code, the parts that for sure needs to be tested are:
- Creating a new Library Item
- Importing a Library Item into form
- Cascading select: https://support.kobotoolbox.org/cascading_select.html
- Locking profiles and templates: https://support.kobotoolbox.org/library_locking.html
- Creating new form
- Editing existing form
- Editing form style (from Layout & Settings)
- Grouping rows
- Reordering rows
- Changing form name through Form Builder
- Previewing form
- Question matrix: https://support.kobotoolbox.org/matrix_response.html


jsapp/js/editorMixins/EditableForm.tsx
```diff
@@ -158,52 +157,55 @@ interface EditableFormState extends SurveyStateStoreData {
...
+    preventNavigatingOut: false,
+    showCascadePopup: false,
+    surveyAppRendered: props.surveyAppRendered,
+    surveyLoadError: undefined,
+    surveySaveFail: false,
+    isNewAsset: props.isNewAsset,
+    backRoute: props.backRoute === null ? undefined : props.backRoute,
+    groupButtonIsActive: false,
+    multioptionsExpanded: props.multioptionsExpanded,
+  })
```
pauloamorimbr, line 182: I think this is a good step towards to modernization and functional component, but I'm not a big fan of huge objects in state. Maybe we can take a second look at this at some point, also because really splitting this into multiple `useState` would mean probably a logic refactor as well.

pauloamorimbr (approved):

PR #6719, opened 2026-02-12
feat(signup): block user signup for blacklisted email addresses DEV-1741

rajpatel24:
### 📣 Summary
Server admins can now block specific email domains from creating new accounts, preventing unauthorized sign-ups from restricted organizations or providers.

### 📖 Description
This update introduces a new domain blacklist feature in the system config. Admins can now specify a list of email domains (e.g., `bad-domain.com`) that are explicitly forbidden from registering on the server. This is useful for organizations that want to prevent their staff from creating accounts on a public server using their official email addresses, or to block known disposable email providers. If a user tries to sign up with a blacklisted domain, they will see an error message, which can be customized by the admin.

### 👀 Preview steps

1. Add a domain (e.g., example.com) to the `REGISTRATION_BLACKLIST_EMAIL_DOMAINS` constant config.
2. Sign up using an email address with the blacklisted domain (e.g., jack@example.com).
3. 🔴 [On main] There is no feature to add a blacklisted domain.
4. 🟢 [On PR] After clicking "Create Account", verify that the error message is displayed and registration is blocked.
5. Also, try changing the error message in `REGISTRATION_BLACKLIST_ERROR_MESSAGE` and test with an empty blacklist. Verify that it works correctly in all possible scenarios.

### DIFF
--- a/kpi/tests/test_registration.py
+++ b/kpi/tests/test_registration.py
@@ -146,3 +146,4 @@ class RegistrationTestCase(TestCase):
             b'Please contact your organization support team for assistance.',
             response.content
         )
+        self.assertFalse(User.objects.filter(username='alice').exists())


kpi/tests/test_registration.py
```diff
@@ -71,3 +71,91 @@ def test_disallowed_domain_cannot_register(self):
...
+    @override_config(
+        ENABLE_PASSWORD_MINIMUM_LENGTH_VALIDATION=False,
+        ENABLE_PASSWORD_USER_ATTRIBUTE_SIMILARITY_VALIDATION=False,
+        ENABLE_MOST_RECENT_PASSWORD_VALIDATION=False,
+        ENABLE_COMMON_PASSWORD_VALIDATION=False,
+        ENABLE_PASSWORD_CUSTOM_CHARACTER_RULES_VALIDATION=False,
+        REGISTRATION_BLACKLIST_EMAIL_DOMAINS='bad-domain.com\nmalicious.net',
+        REGISTRATION_BLACKLIST_ERROR_MESSAGE='Go away!'
+    )
+    def test_blacklisted_domain_cannot_register(self):
```
rgraber, line 86: Can we also check no new user is created

kpi/tests/test_registration.py
```diff
@@ -71,3 +71,91 @@ def test_disallowed_domain_cannot_register(self):
...
+        ENABLE_PASSWORD_MINIMUM_LENGTH_VALIDATION=False,
+        ENABLE_PASSWORD_USER_ATTRIBUTE_SIMILARITY_VALIDATION=False,
+        ENABLE_MOST_RECENT_PASSWORD_VALIDATION=False,
+        ENABLE_COMMON_PASSWORD_VALIDATION=False,
+        ENABLE_PASSWORD_CUSTOM_CHARACTER_RULES_VALIDATION=False,
+        REGISTRATION_BLACKLIST_EMAIL_DOMAINS='bad-domain.com'
+    )
+    def test_blacklisted_domain_with_no_error_message(self):
+        """
+        Test that if a blacklisted domain is used and no error message is set,
```
rgraber: Is this true? We *don't* block registration if there isn't an error message?

kpi/tests/test_registration.py
```diff
@@ -71,3 +71,91 @@ def test_disallowed_domain_cannot_register(self):
...
+        registration should not be blocked
+        """
+        data = self.valid_data.copy()
+        data['email'] = 'user@bad-domain.com'
+
+        response = self.client.post(
+            reverse('account_signup'), data=data
+        )
+        self.assertIn(b'', response.content)
```
rgraber: Same as above, can we validate no user was created

kpi/tests/test_registration.py
rajpatel24: Sorry, This test is incorrect and not required. We do not allow user signup regardless of whether an error message is present. If no error message is configured, a default message will be displayed.

kpi/tests/test_registration.py
rajpatel24: Same as above, This test is incorrect and not required.

kpi/tests/test_registration.py
```diff
@@ -147,6 +129,11 @@ def test_empty_blacklist_allows_registration(self):
         REGISTRATION_BLACKLIST_EMAIL_DOMAINS='bad-domain.com'
     )
     def test_default_blacklist_error_message(self):
```
rgraber, line 131: Can we also add the check here to make sure no user was created

PR #6732, opened 2026-02-17
fix(submissions): prevent edit link for duplicate submissions DEV-1751

noliveleger:
### 📣 Summary
Users can no longer edit submissions that have duplicate UUIDs to prevent data integrity issues.

### 📖 Description
When attempting to edit a submission that has a duplicate UUID (usually from legacy data without `root_uuid`), the system now returns an error immediately instead of allowing the user to make edits that would fail upon submission in Enketo. This prevents users from wasting time editing a submission only to encounter an error when trying to save their changes.

### 👀 Preview steps
1. ℹ️ have an account and a project with submissions
2. Create a situation with duplicate submission UUIDs (this typically happens with legacy data)
3. Identify a submission that has a duplicate UUID
4. Attempt to get the edit link for this submission via API: `GET /api/v2/assets/{asset_uid}/data/{submission_id}/enketo/edit/`
5. 🔴 [on release branch] The API returns an edit URL, but editing will fail when submitting changes in Enketo
6. 🟢 [on PR] The API immediately returns 409 CONFLICT with message: "A duplicate submission has been detected. This submission cannot be edited at the moment."
Screenshot 2026-02-17 at 16 33 14
7. 🟢 Notice that only duplicates are blocked from editing

### DIFF
--- a/kpi/tests/api/v2/test_api_submissions.py
+++ b/kpi/tests/api/v2/test_api_submissions.py
@@ -14,6 +14,7 @@ import lxml
 import pytest
 import responses
 from constance.test import override_config
+from ddt import data, ddt, unpack
 from django.conf import settings
 from django.core.files.base import ContentFile
 from django.db import connection
@@ -1214,6 +1215,7 @@ class SubmissionApiTests(SubmissionDeleteTestCaseMixin, BaseSubmissionTestCase):
             assert response.data['_submitted_by'] == 'anotheruser'
 
 
+@ddt
 class SubmissionEditApiTests(SubmissionEditTestCaseMixin, BaseSubmissionTestCase):
     """
     Tests for editing submissions.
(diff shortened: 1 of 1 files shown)


kpi/tests/api/v2/test_api_submissions.py
```diff
@@ -2108,6 +2108,167 @@ def test_submitted_by_persists_for_shared_submitter(self):
...
+            ).count()
+            == 2
+        )
+
+        # Attempt to get the edit link - should fail with 409
+        submission_edit_link_url = reverse(
+            self._get_endpoint('submission-enketo-edit'),
+            kwargs={
+                'uid_asset': self.asset.uid,
+                'pk': self.submission['_id'],
```
jnm: ah, what the heck, let's make sure it also returns a 409 for `duplicate.pk`

kpi/tests/api/v2/test_api_submissions.py
```diff
@@ -2108,6 +2108,167 @@ def test_submitted_by_persists_for_shared_submitter(self):
...
+                xform_id=self.asset.deployment.xform_id,
+            ).count()
+            == 1
+        )
+
+        submission_edit_link_url = reverse(
+            self._get_endpoint('submission-enketo-edit'),
+            kwargs={
+                'uid_asset': self.asset.uid,
+                'pk': self.submission['_id'],
```
jnm, line 2179: Let's also check that trying to edit the second instance returns a 409

kpi/tests/api/v2/test_api_submissions.py
```diff
@@ -2108,6 +2108,167 @@ def test_submitted_by_persists_for_shared_submitter(self):
         )
         assert ph_bulk.metadata['submission']['submitted_by'] == deleted_username
 
+    @responses.activate
+    def test_get_edit_link_fails_with_duplicate_uuid_null_root_uuid(self):
```
jnm: @noliveleger suggests using DDT to handle the combinations in these 3 tests:
1. null uuid, null root uuid
2. not-null uuid, null root uuid
3. not-null for both both, but they mismatch

maybe the name for the whole test could be `def test_cannot_edit_duplicate_submissions` and the docstring could explain why testing these 3 scenarios is useful

kpi/tests/api/v2/test_api_submissions.py
```diff
@@ -2108,6 +2108,167 @@ def test_submitted_by_persists_for_shared_submitter(self):
...
+        assert (
+            Instance.objects.filter(
+                uuid=submission_uuid,
+                root_uuid__isnull=False,
+                xform_id=self.asset.deployment.xform_id,
+            ).count()
+            == 2
+        )
+
+        # Get the edit link - should succeed because the original has a non-null
```
jnm: already being fixed :)

jnm:
Tested locally :heavy_check_mark:

jnm (approved):

PR #6735, opened 2026-02-19
feat(automatedQA): integrate AI answer generation api DEV-739

pauloamorimbr:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
This PR implements the flow for activating the AI response generation feature and calling the actual generation endpoint

### 💭 Notes
- For this to work some changes were necessary in the current `TabAnalysis` flow, so we could have access to the automated feature as well. 
- Now the `manual_qual` feature is created whenever a question is set and it's still not there. In the common flow it will happen once the first question is added. I chose that path instead of automatically creating the feature when `TabAnalysis` was opened, which i considered another viable option. This way we won´t create the feature unless the user really adds a qaQuestion.
- There's a verification logic to decide whether to create the advanced feature for the AI generation, to patch it or do nothing if all qaQuestions are already set for it. This is important in the case of new questions being added.
- Tags and Notes questions don't have `Generate with AI` enabled since it's not supported for them
- The actual API call to generate the answer is working, but the generated data is not being parsed and displayed in this PR. We will take care of this in a follow up work.

### 👀 Preview steps
1. ℹ️ have an account and a project with transcription
2. have QA questions
3. have `autoQAEnabled` feature flag enabled (`?ff_autoQAEnabled=true`)
4. Press Generate with AI button in the analysis tab
5. 🔴 [on main] notice that this won't trigger any action
6. 🟢 [on PR] notice that this triggers activation of the feature for the existing questions and the API call for generating the answer with AI.

### DIFF
--- a/jsapp/js/components/processing/SingleProcessingContent/TabAnalysis/AnalysisContent/AnalysisQuestionsList/AnalysisQuestionListItem/ResponseForm.tsx
+++ b/jsapp/js/components/processing/SingleProcessingContent/TabAnalysis/AnalysisContent/AnalysisQuestionsList/AnalysisQuestionListItem/ResponseForm.tsx
@@ -19,7 +19,10 @@ interface Props {
   disabled: boolean
   onEdit: (qaQuestion: ResponseQualActionParams) => unknown
   onDelete: (qaQuestion: ResponseQualActionParams) => Promise<unknown>
-  /** Adds a Generate with AI button */
+  /**
+   * Adds a Generate with AI button. API handling is being served by parent(s), as this component doesn't have all
+   * the required data and it's easier to push this one function up than all the small pieces down.
+   */
   onGenerateWithAI?: () => Promise<unknown>
 }
 
--- a/jsapp/js/components/processing/SingleProcessingContent/TabAnalysis/AnalysisContent/AnalysisQuestionsList/index.tsx
+++ b/jsapp/js/components/processing/SingleProcessingContent/TabAnalysis/AnalysisContent/AnalysisQuestionsList/index.tsx
@@ -89,12 +89,13 @@ export default function AnalysisQuestionsList({
     })
   }, [])
 
-  const localAdvancedFeature: AdvancedFeatureResponseManualQual | undefined = advancedFeatureManual
-    ? {
-        ...advancedFeatureManual,
-        params: localParams,
-      }
-    : undefined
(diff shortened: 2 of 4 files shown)


jsapp/js/components/processing/SingleProcessingContent/TabAnalysis/AnalysisContent/AnalysisQuestionsList/AnalysisQuestionListItem/ResponseForm.tsx
```diff
@@ -108,16 +129,19 @@ export default function ResponseForm({ qaQuestion, children, onClear, disabled,
...
-      {!disabled && ffAutoQAEnabled && (
+      {!disabled && ffAutoQAEnabled && onGenerateWithAI && (
         <Group pl={'40px'}>
           <ButtonNew
             variant='transparent'
             h='fit-content'
             size='md'
             p={0}
-            disabled={disabled}
+            disabled={disabled || isGenerating}
```
magicznyleszek: I think this is unnecessary. If we have `loading` prop, then it already handles disabling of any interactions. And with both `loading` and `disabled`, the button loader dots appear half transparent.

jsapp/js/components/processing/SingleProcessingContent/TabAnalysis/AnalysisContent/AnalysisQuestionsList/index.tsx
```diff
@@ -62,13 +99,70 @@ export default function AnalysisQuestionsList({
...
+        data: {
+          action: ActionEnum.automatic_bedrock_qual,
+          question_xpath: questionXpath,
+          params: uncoveredQuestions,
+        },
+      })
+    }
+  }
+
+  const handleGenerateWithAI = async (qaQuestionParam: ResponseQualActionParams) => {
```
magicznyleszek, line 150: We should add some better error handling. When I was getting 500 (missing `frontend.custom.yml` configuration), the UI wasn't very helpful - I got a huge error notificataion with HTML inside of it :P The question is: is this an error on FE side, or is BE not providing a nice message

pauloamorimbr: Did it!

PR #6736, opened 2026-02-19
build(api)!: remove django-rest-framework frontend files DEV-1718

RuthTurk:
### 📣 Summary
DRF's browsable HTML API, which is no longer used by our application, has been replaced with a dependency-free, standalone HTML renderer to address security vulnerabilities associated with outdated third-party libraries like Bootstrap. 

### 👀 Preview steps

2. 🔴 [on main] navigate to `http://kf.kobo.local/static/rest_framework/js/bootstrap.min.js` and see:
```
/*!
 * Bootstrap v3.4.1 (https://getbootstrap.com/)
 * Copyright 2011-2019 Twitter, Inc.
 * Licensed under the MIT license
 */
 ...
```
3. Navigate to `/api/v2/assets/`
4. 🔴 [on main] Notice a styled HTML page with DRF's browsable HTML API
5. Checkout the pr branch and completely rebuild your docker containers from scratch. 
6. 🟢 [on PR] navigate to `http://kf.kobo.local/static/rest_framework/js/bootstrap.min.js` and get a 404 Not Found.
7. Navigate to `/api/v2/assets/`
8. 🟢 [on PR] Notice that the page renders with a different style as before but with the same content
9. 🟢 [on PR] Check out the console and ensure that there are no 404 errors trying to render any HTML
ex: `GET /static/rest_framework/css/bootstrap.min.css HTTP/1.0" 404`
10. Check out other endpoints and ensure that they do not render or make any HTML requests
11.  Ensure that nothing is broken and everything looks/functions the same in the application

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -132,14 +132,14 @@ ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 COPY ./dependencies/pip/requirements.txt "${TMP_DIR}/pip_dependencies.txt"
 RUN uv pip sync "${TMP_DIR}/pip_dependencies.txt" 1>/dev/null
 
+RUN rm -rf ${VIRTUAL_ENV}/lib/python*/site-packages/rest_framework/static/rest_framework
+
 #####################################
 #                                   #
 # 🧰 KPI production image 'kpi-app' #
 #                                   #
 #####################################
 FROM ghcr.io/astral-sh/uv:python3.10-bookworm-slim AS kpi-app
-
-RUN rm -rf ${VIRTUAL_ENV}/lib/python*/site-packages/rest_framework/static/rest_framework
 
 ENV DEBIAN_FRONTEND=noninteractive \
     LANG=en_US.UTF-8 \


Dockerfile
```diff
@@ -127,7 +129,9 @@ RUN npm run build:app
 # Organize static assets. #
 ###########################
 
-RUN python manage.py collectstatic --noinput
+RUN python manage.py collectstatic --noinput --ignore rest_framework
+
+RUN rm -rf ${KPI_SRC_DIR}/staticfiles/rest_framework/
```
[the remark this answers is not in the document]
noliveleger: `collectstatic --ignore rest_framework` **does** ignore the whole directly entirely. that's correct.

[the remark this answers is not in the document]
noliveleger: `RUN rm -rf ${KPI_SRC_DIR}/staticfiles/rest_framework/` can be removed because `collectstatic --ignore rest_framework` already prevents these files from being copied to the static root, so the directory never exists in the first place.

Additionally, if `staticfiles` is mounted as a volume, the host directory overrides the container's filesystem at that path anyway, making any `rm -rf` inside the image ineffective at runtime.

However, if `staticfiles` is mounted as a volume in development, old `rest_framework` files from a previous build may still be present on the host and would be picked up by rsync.

Instead, you could modify the rsync command in `docker/entrypoint.sh` to explicitly exclude them:

```
rsync -aq --delete --delete-excluded --exclude="rest_framework" --chown=www-data "${KPI_SRC_DIR}/staticfiles/" "${NGINX_STATIC_DIR}/"
```

This ensures stale rest_framework static files are never served by nginx, regardless of what persists on the host.

Dockerfile
```diff
@@ -99,6 +99,8 @@ ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 COPY ./dependencies/pip/requirements.txt "${TMP_DIR}/pip_dependencies.txt"
 COPY --from=build-python "$VIRTUAL_ENV" "$VIRTUAL_ENV"
 
+RUN rm -rf ${VIRTUAL_ENV}/lib/python*/site-packages/rest_framework/static/rest_framework
```
noliveleger: `RUN rm -rf ${VIRTUAL_ENV}/lib/python*/site-packages/rest_framework/static/rest_framework` could be removed.
          
AFAIK, In a multi-stage build, `COPY --from=build-python` introduces the virtualenv into the final stage as a new layer (including the DRF static files). 

A subsequent `RUN rm -rf` only adds another layer that masks those files; it does not remove them from the previous layer. The files still exist in the image and the total image size is not reduced.                                                                                                                  
                             
To actually exclude those files, they would need to be deleted in the build-python stage before the `COPY --from instruction.` 

Otherwise, see comments below, the files will never be exposed in nginx, neither present into `staticfiles` thanks to other changes you made.

kpi/templates/base_no_static.html
noliveleger, line 1: I love your template! It's way nicer than the DRF one ;-)

kpi/templates/base_no_static.html
```diff
@@ -0,0 +1,19 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <title>{% block title %}KoboToolbox API{% endblock %}</title>
+    <style>
+        body { font-family: sans-serif; line-height: 1.6; padding: 20px; max-width: 1200px; margin: 0 auto; color: #333; }
+        pre { background: #f4f4f4; padding: 15px; border-radius: 5px; overflow-x: auto; white-space: pre-wrap; word-wrap: break-word; }
+        .deprecated-warning { background: #fff3cd; border: 1px solid #ffeeba; color: #856404; padding: 15px; margin-bottom: 20px; border-radius: 4px; }
+        a { color: #2095f3; text-decoration: none; }
+        a:hover { text-decoration: underline; }
+        hr { border: 0; border-top: 1px solid #eee; margin: 20px 0; }
+    </style>
```
noliveleger, line 14: Should be tested on a staging environment with CSP headers activated. I'm wondering if the inline style could be blocked.

kpi/templates/rest_framework/api.html
```diff
@@ -1,9 +1,50 @@
...
 {% block title %}KoboToolbox Form Building API{% endblock %}
 
-{% block branding %}
-    <a class='navbar-brand' rel="nofollow" href='http://kobotoolbox.org/'>
-        KoboToolbox Form Building API
-    </a>
+{% block body %}
+    <style>
+        /* No-static API styling */
+        body { font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helvetica, Arial, sans-serif; color: #2c3e50; line-height: 1.5; }
```
noliveleger, line 8: Should be tested on a staging environment with CSP headers activated. I'm wondering if the inline style could be blocked.

PR #6737, opened 2026-02-19
fix(hooks): cache attachment XPaths and optimize single-submission retrieval DEV-1762

noliveleger:
### 📣 Summary
 REST Services hooks now process submissions faster by caching attachment XPath lookups in Redis and avoiding unnecessary database queries when retrieving individual submissions.

### 📖 Description
When a REST Service hook fires for a single submission, the system previously computed attachment XPaths by iterating over all deployed form versions — an expensive operation involving multiple database queries. This change introduces Redis caching for attachment XPaths and an optimized path for single-submission retrieval that derives XPaths from only the submission's specific form version when the cache is cold.

### 💭 Notes
 - Three-tier caching strategy for get_all_attachment_xpaths: 
   - in-memory attribute on the Asset instance
   - Redis cache keyed by asset_uid:latest_deployed_version_uid
   - full recomputation from all deployed versions as a fallback.
 - fetch_one optimization in `__get_submissions_in_json`: When retrieving a single submission (e.g., for hook processing), the method first attempts a cache-only lookup (only_cached_data=True). On a cache hit,
 it skips all version queries. On a cache miss, it reads the submission's __version__ field and computes XPaths from that single version only, rather than scanning all deployed versions.
 - `get_attachment_xpaths_from_version` is also cached individually per version UID, so repeated calls for the same version (across hooks or requests) are served from Redis.
 - Added structured logging to `ServiceDefinitionInterface.send()`to trace hook submission lifecycle (start, data check, pre-send, result).
 
 
### 👀 Preview steps

1. Have an account with a deployed project that has attachment questions (audio, image, or video), a configured REST Service, and many deployed versions
2. Using code before this PR, make a submission and notice the the low-priority Celery worker takes a long time to execute the `service_definition_task`
3. Now, using the code in this PR, make another submission and notice that `service_definition_task` completes quickly
4. Look for new log messages in the Celery worker log
5. Look for new `all_attachment_xpaths:*` and `attachment_xpaths:*` keys in Redis

### DIFF
--- a/kobo/apps/hook/models/service_definition_interface.py
+++ b/kobo/apps/hook/models/service_definition_interface.py
@@ -78,6 +78,10 @@ class ServiceDefinitionInterface(metaclass=ABCMeta):
         Raise an exception if something is wrong. Retries are only allowed
         when `HookRemoteServerDownError` is raised.
         """
+
+        # TODO consider changing "logging.info"  to "logging.debug" when
+        #   DEV-1762 is reviewed & merged.
+
         logging.info(
             'service_json.ServiceDefinition.send: '
             f'Starting hook submission processing - Hook #{self._hook.uid} - '
--- a/kpi/deployment_backends/openrosa_backend.py
+++ b/kpi/deployment_backends/openrosa_backend.py
@@ -1571,8 +1571,8 @@ class OpenRosaDeploymentBackend(BaseDeploymentBackend):
                     ).first()
 
                 # Compute attachment xpaths from that specific form version.
-                attachment_xpaths = (
-                    self.asset.get_attachment_xpaths_from_version(submission_version)
+                attachment_xpaths = self.asset.get_attachment_xpaths_from_version(
+                    submission_version
                 )
         else:
             attachment_xpaths = self.asset.get_all_attachment_xpaths()
--- a/kpi/tests/test_mock_data.py
+++ b/kpi/tests/test_mock_data.py
@@ -535,7 +535,7 @@ class AttachmentXPathCacheTestCase(BaseSubmissionTestCase):
         with patch.object(
             self.asset,
             'get_all_attachment_xpaths',
-            return_value=['Photo', 'Audio', 'Video']
+            return_value=['Photo', 'Audio', 'Video'],
         ) as mock_get_xpaths:
 
             # Retrieve all submissions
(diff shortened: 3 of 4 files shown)


kobo/apps/hook/models/service_definition_interface.py
```diff
@@ -78,15 +78,30 @@ def send(self) -> bool:
...
         when `HookRemoteServerDownError` is raised.
         """
-
+        logging.info(
+            'service_json.ServiceDefinition.send: '
+            f'Starting hook submission processing - Hook #{self._hook.uid} - '
+            f'Submission #{self._submission_id}'
+        )
         if not self._data:
+            logging.warning(
```
jnm: meh, this deletion would happen by user action and so the log message is fine to remain at `info` level

kpi/deployment_backends/openrosa_backend.py
```diff
@@ -1536,12 +1541,46 @@ def __get_submissions_in_json(self, **params) -> Generator[dict, None, None]:
...
+                    # Wrap the single submission in a list so it can still be
+                    # iterated over later in the final loop.
+                    mongo_cursor = [submission]
+
+                # Retrieve the deployed version corresponding to the submission.
+                submission_version = None
+                if '__version__' in submission:
+                    submission_version = self.asset.asset_versions.filter(
+                        uid=submission['__version__'], deployed=True
+                    ).first()
```
jnm, line 1571: maybe reuse some of the old `_infer_version_id()` logic, but unfortunately that function would have to be broken apart

kpi/tests/test_mock_data.py
```diff
@@ -498,3 +506,165 @@ def test_has_version_and_submissions(self):
...
+            submissions.append(submission)
+
+        v_uid = self.asset.latest_deployed_version.uid
+        for submission in submissions:
+            submission.update({'__version__': v_uid})
+
+        self.asset.deployment.mock_submissions(submissions)
+        return submissions
+
+
```
jnm: rebase problem; will go away

jnm (changes requested):
(outdated, but for the record)

kpi/deployment_backends/openrosa_backend.py
jnm: will try to get the :robot: to do this or will open an issue for it later. let's not block this PR

jnm (approved):

PR #6740, opened 2026-02-20
fix(formBuilder): ensure cloned choices and rows ids are unique DEV-1759

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

When cloning a question in Form Builder the result is now ensured to be a copy with unique ids. Previously clone was using original row ids leading to unexpected behaviour.

### 💭 Notes

The problem was happening for every cloned row. The `$kuid` was being cloned too. Luckily we don't rely on `$kuid` much, but it is an obvious error that could be potentially problematic.

The visible issue was happening for cloning of few types: `select_one`, `select_multiple`, `rank` or `score`. When cloning such row, code was also cloning the list of choices for it. Before the fix the cloned list of choices was wrongly being synced between two rows (the original and the clone). This leads to confusing issues as the syncing isn't obvious and might get unnoticed and lead to unwanted changes in the form.

Changes in this PR:
- In `clone` function of choices list (`jsapp/xlform/src/model.choices.coffee`), I ensure that the unique ids are stripped and regenerated - for the list itself and for each option. I also register the new list in global list of `choices`
- In `clone` (look for `rr.clone`) function of `ScoreRankMixin` (`jsapp/xlform/src/model.row.coffee`) I similarly ensure unique ids are stripped and regenerated - for the row itself and for the score/rank levels
- In `clone` function of `row.Row` (`jsapp/xlform/src/model.row.coffee`) I introduced safety measure to ensure unique properties are not being copied and `select_one` and `select_multiple` are properly copying only non-unique things
- Added one `TODO` comment for a weird single use `_isCloned` flag
- Added comment about us purposefuly not allowing `kobomatrix` cloning (as I thought initially it was a mistake/bug)
- Some minor code readability improvements (coffee syntax is not nice)
- Added some comments here and there :) 

### 👀 Preview steps

I recommend this setup:
- Create new empty form and save (unselect "start date" and "end date" in sidebar
- Open two tabs
- First tab at `http://kf.local.kbtdev.org/#/forms/<uid>/edit`
- Second tab at `http://kf.local.kbtdev.org/#/forms/<uid>/json`

1. `/edit` tab → Add `select_one` row with one choice
2. `/edit` tab → Save
3. `/json` tab → Refresh and notice `survey` has one item, and `choices` have one item
4. `/edit` tab → Use "Duplicate Question" icon button (blue one on right side of row)
5. `/edit` tab → Save
6. `/json` tab → Refresh and notice `survey` has two items, and `choices` have two items:
   1. 🔴 [on main] notice that two `survey` and two `choice` items have identical `$kuid`s (they all should be unique) and `select_from_list_name` (pairing with `list_name`) is pointing to the same list (second item should point to different list)
   2. 🟢 [on main] notice that two `survey` and two `choice` items have different `$kuid`s and `select_from_list_name` are pointing to different lists

Do a similar test with cloning:
- a `select_mulitple` question (test is identical to `select_one`, but we need to test it separately, as code paths are not identical)
- a `text` question (`$kuid`s should be unique)
- a `rank` question (`$kuid`s should be unique, `kobo--rank-items` should be pointing to different `list_name`)
- a `score` question (same as `rank`, but `kobo--score-choices` is the name of property that should point to `list_name`)

### DIFF
--- a/jsapp/xlform/src/model.choices.coffee
+++ b/jsapp/xlform/src/model.choices.coffee
@@ -156,7 +156,10 @@ module.exports = do ->
 
       clonedList = new choices.ChoiceList(json)
 
-      return clonedList
+      return _.assign(
+        new choices.ChoiceList(json),
+        collection: @collection
+      )
 
     toJSON: ()->
       @finalize()


jsapp/xlform/src/model.choices.coffee
```diff
@@ -142,13 +142,26 @@ module.exports = do ->
...
+      # Strip out $kuid from each option so they are treated as brand new entities
+      if json.options
+        for optionAttr in json.options
+          delete optionAttr['$kuid']
+
+      clonedList = new choices.ChoiceList(json)
+
+      # If the original list belongs to a collection, add the clone to it.
+      # Backbone's `.add()` will automatically set `clonedList.collection` for us.
+      if @collection
```
duvld: Why is this guard introduced now? previously we didn't check for existence and it doesn't seem to be related to the bug (the `$kuid` part is handled above)

jsapp/xlform/src/model.row.coffee
```diff
@@ -401,6 +423,8 @@ module.exports = do ->
       return
     getTypeId: ->
       return @get('type').get('typeId')
+
+    # Clones given row in a deep way (i.e. also clones list of choices etc.)
```
duvld: "given" as in a row as a parameter, or do you mean "this" row?

jsapp/xlform/src/model.row.coffee
```diff
@@ -410,15 +434,33 @@ module.exports = do ->
         remove: false
         silent: true
 
+      _.each @attributes, (value, key) =>
+        # Prevent copying the hardcoded list reference and the unique row id. They will be autofilled with proper values
+        # in the lines below
+        if key not in ['$kuid', 'select_from_list_name']
```
duvld, line 440: Just trying to understand everything here, is this essentially doing the same as the "stripping" of `$kuid` like the lines above?

jsapp/xlform/src/model.row.coffee
```diff
@@ -410,15 +434,33 @@ module.exports = do ->
...
+          attributes[key] = @getValue key
 
-      _.each @attributes, (value, key) => attributes[key] = @getValue key
-
-      newRow = new row.Row attributes, options
+      newRow = new row.Row(attributes, options)
 
+      # For few distinct types we need to do a bit more work
```
duvld: Are "distinct types" are just "select one" and "select multiple"?

jsapp/xlform/src/model.row.coffee
```diff
@@ -410,15 +434,33 @@ module.exports = do ->
...
-        newRowType.set 'list', @getList().clone()
-        newRowType.set 'listName', newRowType.get('list').get 'name'
+        # Clone the choices list (as it is a distinct entity)
+        clonedList = @getList().clone()
+
+        # Ensure new list has a unique name
+        listName = txtid()
+        # We use `silent` to not cause unnecessary re-renders. We are going to `trigger('change')` few lines below after
+        # cloning is done.
+        clonedList.set('name', listName, {silent: true})
```
duvld: Wondering about this too, why does mutating a variable we just defined through a setter cause a re-render?

jsapp/xlform/src/view.row.coffee
```diff
@@ -202,15 +202,17 @@ module.exports = do ->
     _cleanupExpandedRender: ->
       @$('.card__settings').detach()
 
+    # Clones the row with all the related parts
```
duvld: I think this comment is a bit vague and is better explained below by your next ones anyway, maybe we don't need this one here?

jsapp/xlform/src/view.row.coffee
```diff
@@ -202,15 +202,17 @@ module.exports = do ->
...
-        model = @model.clone()
-      else if @model.get('type').get('typeId') in ['rank', 'score']
+      # When we clone one of these types, we can't simply add another row like it, we need to also use inner clone
+      # function that will ensure all related parts are cloned too (e.g. list of choices)
+      # Note: we omit `kobomatrix` here on purpose
+      if @model.get('type').get('typeId') in ['select_one', 'select_multiple', 'rank', 'score']
         model = @model.clone()
 
-      @model.getSurvey().insert_row.call parent._parent, model, parent.models.indexOf(@model) + 1
+      @model.getSurvey().insert_row.call(parent._parent, model, parent.models.indexOf(@model) + 1)
```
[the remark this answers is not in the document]
duvld: I think Copilot is onto something here. In the preview steps, you say to test cloning a `text` question. I noticed a couple things about this:
1. Why just a `text` question? It's exhaustive, but every question seems to have a `$kuid`, so we should test them all?
2. (This one is related to copilot) I don't think the bug is fixed for non choice questions, like `text`:

https://github.com/user-attachments/assets/81353c44-662f-4a58-8b26-1f28c5cbe087

The duplication bug doesn't really have any effects on these questions, as they don't have a choice list to point to, but we should probably fix this anyways

jsapp/xlform/src/model.row.coffee
```diff
@@ -410,15 +434,33 @@ module.exports = do ->
...
+
+        # Register the cloned list with the global choices collection
+        @getSurvey().choices.add(clonedList)
+
```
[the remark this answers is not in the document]
magicznyleszek: That makes sense. I've fixed that by deduplicating registration

jsapp/xlform/src/view.row.coffee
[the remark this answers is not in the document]
magicznyleszek: That is a good catch! I simplified the code to use clone method for all types :)

jsapp/xlform/src/view.row.coffee
```diff
@@ -202,15 +202,17 @@ module.exports = do ->
...
     clone: (event) ->
       parent = @model._parent
       model = @model
-      if @model.get('type').get('typeId') in ['select_one', 'select_multiple']
-        model = @model.clone()
-      else if @model.get('type').get('typeId') in ['rank', 'score']
+      # When we clone one of these types, we can't simply add another row like it, we need to also use inner clone
+      # function that will ensure all related parts are cloned too (e.g. list of choices)
+      # Note: we omit `kobomatrix` here on purpose
+      if @model.get('type').get('typeId') in ['select_one', 'select_multiple', 'rank', 'score']
```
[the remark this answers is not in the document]
magicznyleszek: Already changed the code, plus matrix cloning is not allowed in the UI. This is good enough

jsapp/xlform/src/model.row.coffee
```diff
@@ -173,23 +178,40 @@ module.exports = do ->
...
           for item in rr.getList().options.models
-            r2._rankLevels.options.add(item.toJSON())
+            itemJson = item.toJSON()
+            # Strip $kuid from items
+            delete itemJson['$kuid']
+            r2._rankLevels.options.add(itemJson)
+
           r2.set('kobo--rank-items', r2._rankLevels.get('name'))
           @convertAttributesToRowDetails()
           r2.get('type').set('list', r2._rankLevels)
```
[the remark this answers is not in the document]
magicznyleszek: This is not related to the PR, but still good catch.

jsapp/xlform/src/model.choices.coffee
magicznyleszek: This code was indeed unnecessary, I think a leftover from figuring things out :)

jsapp/xlform/src/model.row.coffee
magicznyleszek: Improved the comment

jsapp/xlform/src/model.row.coffee
magicznyleszek: Yes, we are stripping `$kuid` (copying all the other attributes), so then Backbone sees it missing and creates a new one

jsapp/xlform/src/model.row.coffee
magicznyleszek: Yes, I think this comment is useless, will remove it :)

jsapp/xlform/src/model.row.coffee
magicznyleszek: I think this one was also some leftover. We are setting things on a cloned thing that is not attached anywhere. I did remove the silent param :)

jsapp/xlform/src/view.row.coffee
magicznyleszek: Agree, removing it

duvld (approved):
Code lgtm, tested `text` question again and it works now with different `$kuid`. Select one still works as expected

PR #6743, opened 2026-02-23
fix(qual): link automatic question uuids to manual params DEV-1763

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Update the advanced features API to only take uuids instead of full question specifications as parameters when updating an automatic QA action.

### 📖 Description
Change the `params` field for automatic QA actions to have parameters in the form `{'uuid': 'question_uuid'}` instead of `{'uuid': 'question_uuid', 'type': 'qualInteger', 'labels':...}`

Contains a migration for updating current QuestionAdvancedFeature objects. This feature has barely been released (API only) so it should only update very few objects, likely on test servers only.

### 💭 Notes

Before this refactor, all question definitions had to be duplicated between the `manual_qual` and `automatic_bedrock_qual` actions and kept in sync. This has the obvious issue of lack of a single source of truth and the possibility of the params to diverge. With this new PR, question definitions will only live inside the `manual_qual` action, and `automatic_bedrock_qual` params will consist only of uuids. When processing an automatic QA request, we will look up the actual question text/choices in the `manual_qual` feature for the same question.

To make this cleaner, there is an additional refactor to get rid of `get_action_dependencies` and instead an option to pass pre-fetched dependencies in the creator. This allows us to easily pass new dependencies to new actions instead of having to continually update the signature of get_action_dependencies. In addition, the `__init__` method of the automatic qual action is overridden to fetch its dependencies if they were not initially passed in. We did this instead of creating a new abstract method because right now it only applies to automatic qual and it would be unnecessarily confusing to add it to base actions. If in the future we have other actions requiring this kind of post-init manipulation, we can reconsider.

This has a few notable consequences:
1. `automatic_bedrock_qual` actions are effectively useless without `manual_qual` actions 
2. when updating QA question text or choices, the updates only happen on the `manual_qual` action
3. The param schemas for automatic and manual qual actions are now different
4. creating an automatic qual action without any prefetched data will always hit the database

This PR covers:
- Doing the actual refactor
- Migrating any old `automatic_bedrock_qual` actions hanging around to just have the uuids
- Updating the drf-spectacular generated schemas to reflect the new params
- Updating the swagger examples
- Updating the frontend to use the new names from the new schemas

Important decisions:
We chose not to make this a new schema version since this feature should be in extremely limited use if at all (only available through the API) and a new schema version would make it unnecessarily complicated and confusing. Similarly, we chose to go with a django migration rather than an on-the-fly migration since it should be updating very few objects, mostly on test servers.

`uuid`s are not verified on creation of an `automatic_bedrock_qual` action. This would require another db call. If we end up with nonsensical uuids in the params, they still cannot be used to submit questions to the LLM and trying to do so will result in an error saying there is no matching question. This does mean our schema is a little less strict than maybe we would like but it reduces complexity and time without much risk.

### 👀 Preview steps
Regression test adding/answering QA questions

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -185,7 +185,7 @@
         "sass": "^1.92.1",
         "sass-loader": "^16.0.5",
         "speed-measure-webpack-plugin": "^1.5.0",
-        "storybook": "^9.1.19",
+        "storybook": "^9.0.0-beta.11",
         "storybook-addon-remix-react-router": "^3.1",
         "storybook-dark-mode": "^4.0.2",
         "style-loader": "^3.3.4",
@@ -7616,12 +7616,6 @@
         "@types/ms": "*"
       }
     },
-    "node_modules/@types/deep-eql": {
-      "version": "4.0.2",
-      "resolved": "https://registry.npmjs.org/@types/deep-eql/-/deep-eql-4.0.2.tgz",
-      "integrity": "sha512-c9h9dVVMigMPc4bwTvC5dxqtqJZwQPePsWjPlpSOnojbor6pGqdk541lfA7AqFQr5pB1BRdq0juY9db81BwyFw==",
-      "dev": true
-    },
     "node_modules/@types/doctrine": {
       "version": "0.0.9",
       "resolved": "https://registry.npmjs.org/@types/doctrine/-/doctrine-0.0.9.tgz",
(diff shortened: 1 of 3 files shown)


kobo/apps/subsequences/actions/automatic_bedrock_qual.py
```diff
@@ -201,6 +207,24 @@ def external_data_schema(self):
...
+        super().get_action_dependencies(question_supplemental_data, all_features)
+        from kobo.apps.subsequences.models import QuestionAdvancedFeature
+        from . import ManualQualAction
+
+        try:
+            manual_qual_action = all_features.filter(
+                question_xpath=self.source_question_xpath, action=ManualQualAction.ID
+            ).first()
+        except QuestionAdvancedFeature.DoesNotExist as e:
+            raise ManualQualNotFound from e
```
jnm: We won't reach this since `filter(…non-matching criteria…).first()` returns `None` instead of raising an exception

kobo/apps/subsequences/actions/base.py
```diff
@@ -374,6 +378,7 @@ def transform_data_for_output(
         return {}
 
     def validate_external_data(self, data):
+        logging.info(f'external_data_schema: {self.external_data_schema}')
```
jnm: just checking to see if you wanted this to go into `main`

kobo/apps/subsequences/actions/base.py
```diff
@@ -474,6 +479,7 @@ def revise_data(
             # the validation process.
             dependency_supplemental_data = action_data.pop(self.DEPENDENCY_FIELD, None)
             action_data.update(service_response)
+            logging.info(f'{action_data=}')
```
jnm: check for intention :)

kobo/apps/subsequences/tests/test_automatic_bedrock_qual.py
```diff
@@ -113,46 +119,32 @@ def _add_manual_transcription(self, submission_uuid) -> dict:
...
             with pytest.raises(jsonschema.exceptions.ValidationError):
                 AutomaticBedrockQual.validate_params([param])
 
+    def test_update_params(self):
+        current_params = [copy.deepcopy(param) for param in self.action.params]
+        new_uuid = str(uuid.uuid4())
+        self.action.update_params(
+            [{'uuid': BEDROCK_QUAL_TEXT_UUID}, {'uuid': new_uuid}]
+        )
+        assert self.action.params == [*current_params, {'uuid': new_uuid}]
```
jnm, line 155: as a general principle, i'd appreciate tests being more explicit about the entire desired state when that's feasible. in this case, could something like what follows below work?

```python
assert self.action.params == [
    {'uuid': BEDROCK_QUAL_INTEGER_UUID},
    {'uuid': BEDROCK_QUAL_SELECT_MULTIPLE_UUID},
    {'uuid': BEDROCK_QUAL_SELECT_ONE_UUID},
    {'uuid': BEDROCK_QUAL_TEXT_UUID},
    {'uuid': new_uuid},
]
```

kobo/apps/subsequences/models.py
```diff
@@ -88,7 +90,10 @@ def revise_data(asset: 'kpi.Asset', submission: dict, incoming_data: dict) -> di
                 action_supplemental_data = question_supplemental_data.setdefault(
                     action_id, {}
                 )
-                action.get_action_dependencies(question_supplemental_data)
+                action.get_action_dependencies(
+                    question_supplemental_data, feature_configs_for_this_question
+                )
+                pprint({'dependencies': action._action_dependencies})
```
jnm: check for intention (this and the import)

kpi/schema_extensions/v2/subsequences/examples.py
```diff
@@ -131,8 +114,21 @@ def get_advanced_features_create_examples() -> list[OpenApiExample]:
                 ],
             },
             request_only=True,
-            description='`<action_id>` may be either `manual_qual` '
-            'or `automatic_bedrock_qual`',
+        ),
+        OpenApiExample(
+            'Automatic Qualitative Analysis - Any  Question',
```
jnm: extra space ;)

kpi/views/v2/data.py
```diff
@@ -591,6 +592,10 @@ def supplement(self, request, root_uuid, *args, **kwargs):
             raise serializers.ValidationError({'detail': 'No response to verify'})
         except SubsequenceAcceptanceError:
             raise serializers.ValidationError({'detail': 'No response to accept'})
+        except ManualQualNotFound:
+            raise serializers.ValidationError(
+                {'detail': 'No matching QA question found'}
```
jnm: this could happen in 2 scenarios, right? one where the `manual_qual` action is missing altogether, and another where one of the uuid references is invalid? if that's right, it'd be nice to differentiate that to the stumbling api user. also, "QA" as "qualitative analysis" should probably be kept to internal use only since the rest of the world likely associates different words with those letters at first glance.

kpi/schema_extensions/v2/subsequences/examples.py
```diff
@@ -131,8 +114,21 @@ def get_advanced_features_create_examples() -> list[OpenApiExample]:
...
+            value={
+                'action': 'manual_qual',
```
jnm: should be automatic?

package-lock.json
```diff
@@ -49,7 +49,6 @@
         "lodash.random": "^3.2.0",
         "lodash.throttle": "^4.1.1",
         "lodash.union": "^4.6.0",
-        "lodash.values": "^4.3.0",
```
jnm: should these removals be here?

kobo/apps/subsequences/actions/automatic_bedrock_qual.py
```diff
@@ -201,6 +207,24 @@ def external_data_schema(self):
         to_return['allOf'] = common['allOf']
         return to_return
 
+    def get_output_fields(self) -> list[dict]:
+        return []
+
+    def get_action_dependencies(
```
jnm: this might be my only architectural comment: this method is kind of doing double duty if i understand correctly. it's grabbing the QA responses [following an old pattern of grabbing transcripts when needed] but now is also grabbing the params from the sibling `manual_qual` what were previously duplicated in the automatic action params.

i wonder if that new (second) job should be somewhere else? maybe even in a new method that's called by the action constructor, plus other places it's needed. i mention the constructor because can the action be useful at all without those params being filled in?

- - -

sleepy while writing; disregard as you see fit—but what i'd like to avoid is having to remember to call `get_action_dependencies()` from outside the class whenever this action is used, e.g. in places like `SubmissionSupplement.retrieve_data()` and `BaseAutomaticBedrockQualTestCase.setUp()`

kobo/apps/subsequences/tests/test_automatic_bedrock_qual.py
rgraber: In this case I prefer to keep the test more resilient against changes in the setup params. It's not completely independent obviously, but this way if we add or remove questions in the setup, unless it relates to BEDROCK_QUAL_TEXT_UUID specifically, this test wont fail.

jnm: Fair enough 🫡

kobo/apps/subsequences/tasks.py
```diff
@@ -86,7 +86,7 @@ def poll_run_external_process_failure(sender=None, **kwargs):
         question_xpath=question_xpath, action=action_id
     )
     action = feature.to_action()
-    action.get_action_dependencies(supplemental_data[question_xpath])
+    action.get_revision_dependencies(supplemental_data[question_xpath])
```
jnm: I can't find `get_revision_dependencies` defined anywhere. If it doesn't exist, is there test coverage for this logic?

jnm (changes requested):
thanks for the changes! hopefully one last & small thing

kobo/apps/subsequences/tasks.py
rgraber: yikes, good catch

rgraber: i'd like to make a new task for these tests

rgraber: Removed this method, verified automatic transcription still works. Automatic translation was broken beforehand so not dealing with that. There is now a task to add tests.

PR #6745, opened 2026-02-23
fix(assets): exclude assets from inactive accounts from asset list DEV-1768

noliveleger:
### 📣 Summary

Assets shared by a deactivated user no longer appear in other users' project lists.

### 📖 Description

When a user account was deactivated (`is_active=False`), assets they owned and had shared with other users were still appearing in those users' asset lists, causing a 500 error when the list was fetched. The fix filters out assets whose owner is inactive when resolving explicitly shared permissions.

### 💭 Notes

- Only one line changed in `kpi/filters.py` (added `asset__owner__is_active=True` filter).
- Two regression tests added in `AssetListApiTests` to cover the shared-asset visibility and the inactive-owner exclusion scenarios.

### 👀 Preview steps

1. ℹ️ Have two accounts: **someuser** (owner) and **anotheruser** (grantee). someuser shares a project with anotheruser.
2. 🟢 [on PR] Fetch anotheruser's asset list — the shared project appears correctly.
3. Deactivate someuser's account (`is_active=False`).
4. 🔴 [on main] Fetch anotheruser's asset list — results in a 500 error or includes the shared project incorrectly.
5. 🟢 [on PR] Fetch anotheruser's asset list — the shared project is no longer listed; anotheruser's own projects are unaffected.

### DIFF
--- a/kpi/filters.py
+++ b/kpi/filters.py
@@ -365,7 +365,9 @@ class KpiObjectPermissionsFilter(filters.BaseFilterBackend):
             perms = ObjectPermission.objects.filter(
                 deny=False,
                 user=user,
-                permission_id=view_asset_perm_id)
+                permission_id=view_asset_perm_id,
+                asset__owner__is_active=True,
+            )
 
         return perms.values('asset')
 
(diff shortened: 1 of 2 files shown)


rgraber (approved):
LGTM

PR #6829, opened 2026-03-16
chore(deps): remove temporary Python pins DEV-1793

jnm:
### 💭 Notes

Pins were added so that `./pip-compile.sh --upgrade` could be run without upgrading any major versions of dependencies (see #6773). Packages that were *not* pinned simply had no new major versions available. Catching up with new major versions and streamlining periodic updates will be handled separately, without these pins.


noliveleger (approved):
LGTM

PR #6863, opened 2026-03-24
chore(deps): upgrade dj-stripe to 2.10 DEV-1918

noliveleger:
### 💭 Notes

Many breaking changes in `2.10`. Please read https://github.com/dj-stripe/dj-stripe/releases/tag/2.10.0 

dj-stripe 2.10 consolidated nearly all Stripe model fields into a single `stripe_data` JSONField. Direct DB column access is replaced by `@property` accessors reading from `stripe_data`, so attribute access on model instances still works, but ORM `.filter()` / `.values()` / `F()` calls must use `stripe_data__<field>` JSON lookups.

**ORM filter/annotation changes:**
- All `status`, `start_date`, `billing_cycle_anchor`, `current_period_start`, `current_period_end`, `ended_at`, `unit_amount`, `recurring`, `refunded`, `quantity` lookups now go through `stripe_data__<field>`
- `Max('prices__unit_amount')` replaced with `Max(Cast(KeyTransform('unit_amount', 'prices__stripe_data'), output_field=IntegerField()))`
- `F('start_date')` replaced with `Cast(F('stripe_data__start_date'), output_field=BigIntegerField())` since timestamps are now Unix ints
- `charge.payment_intent.status` → `charge.payment_intent.stripe_data.get('status')`
- `charge.refunded` → `charge.stripe_data.get('refunded')`
- `subscription.plan.product` no longer works; replaced with iteration via `subscription.items.select_related('price__product')`
- `.select_related('schedule')` removed (field moved to `stripe_data`)

**Raw SQL (materialized view):**
- `user_reports/utils/migrations.py`: all column references rewritten to `stripe_data->>'field'` or `stripe_data->'field'` JSON operators
- `default_tax_rates` subquery joining `djstripe_djstripesubscriptiondefaulttaxrate` M2M table replaced with `sub.stripe_data->'default_tax_rates'` (M2M table dropped in 2.10)
- Old migration `0006` emptied; new `0007_djstripe_update` recreates the materialized view with updated SQL

**Migration strategy:**
- `stripe.0001_initial` dependency changed from `('djstripe', '0011_2_7')` → `('djstripe', '0001_initial')` (exists in all djstripe versions)
- New empty `stripe.0005_djstripe_2_10` depends on `('djstripe', '0002_2_10')` as a bridge migration — this avoids `InconsistentMigrationHistory` on upgrade since djstripe 2.10 reset migrations without Django's `replaces` mechanism

**Test fixtures:**
- All `baker.make()` calls updated: fields like `status`, `unit_amount`, `recurring`, `billing_cycle_anchor`, etc. are now `@property` with no setter — values must be passed via `stripe_data={...}` with Unix timestamps
- Datetime comparisons truncated to `replace(microsecond=0)` since Unix timestamps lose microsecond precision


kobo/apps/user_reports/migrations/0007_djstripe_update.py
```diff
@@ -0,0 +1,47 @@
...
+            """.replace(
+                'CREATE UNIQUE INDEX', 'CREATE UNIQUE INDEX CONCURRENTLY'
+            )
+        )
+        return
+
+    # This pulls the *latest* SQL from your updated migrations.py
+    schema_editor.execute(DROP_MV_SQL)
+    schema_editor.execute(CREATE_MV_SQL)
+    schema_editor.execute(CREATE_INDEXES_SQL)
```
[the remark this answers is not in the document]
noliveleger: On purpose!

jsapp/js/api/models/subscriptionItem.ts
```diff
@@ -15,11 +15,5 @@ export interface SubscriptionItem {
   /** @maxLength 255 */
   id: string
   price: PriceWithProduct
```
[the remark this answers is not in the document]
noliveleger: Problem with OpenAPI 3.0 (drf-spectacular) and expected OpenAPI 3.1 in Orval. 
Won't fix in this PR.

kobo/apps/user_reports/migrations/0007_fix_mfa_is_active_new_table.py
```diff
@@ -1,47 +1,9 @@
...
-    schema_editor.execute(CREATE_INDEXES_SQL)
-
 
 class Migration(migrations.Migration):
-    atomic = False
     dependencies = [('user_reports', '0006_fix_org_subscriptions_missing_metadata')]
 
     operations = [
-        migrations.RunPython(apply_fix, reverse_code=migrations.RunPython.noop),
+        migrations.RunPython(migrations.RunPython.noop, migrations.RunPython.noop),
```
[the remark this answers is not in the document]
noliveleger: Not true since it's handled by "0002_reset_mv_for_djstripe_update.py" migration.

kobo/apps/stripe/utils/billing_dates.py
```diff
@@ -202,18 +216,28 @@ def get_current_billing_period_dates_for_active_plans(
...
         .values(
             org_id=F('id'),
-            anchor=F('djstripe_customers__subscriptions__billing_cycle_anchor'),
-            start=F('djstripe_customers__subscriptions__current_period_start'),
-            end=F('djstripe_customers__subscriptions__current_period_end'),
+            anchor=F(
+                'djstripe_customers__subscriptions__stripe_data__billing_cycle_anchor'
+            ),
+            start=F(
+                'djstripe_customers__subscriptions__stripe_data__current_period_start'
```
jamesrkiger, line 228: I'm pretty sure these current_period... lookups are wrong. As far as I can tell, current_period_end and current_period_start are nested (very unhelpfully) under `stripe_data["items"]["data"][*]`. That is an issue in itself. But I'm more concerned that it doesn't seem to have been caught by any of our tests, because we use a lot of mocked data for Stripe-side objects in our tests. And that makes me worry about anything else we might miss.

jnm: Interloper comment, but I think we're okay with these ORM lookups based on dj-stripe's own use of the `stripe_data` dictionary: https://github.com/dj-stripe/dj-stripe/blob/stable/2.10/djstripe/models/billing.py#L1250-L1258, e.g. `self.stripe_data.get("current_period_end")`

> we use a lot of mocked data for Stripe-side objects in our tests

This certainly is a weakness. I wonder if there's any way for us to use work that dj-stripe is presumably already doing to have up-to-date fixtures for their own tests? Could we compare the shape of those fixtures to some expectation as part of our unit tests?

jamesrkiger: @jnm Good catch. I did not realize they still have that property there, but now I'm even more confused because as far as I can see it doesn't work:
```
In [9]: sub.current_period_start

In [10]: "current_period_start" in sub.stripe_data
Out[10]: False
```

The subscription-level billing period data was removed in a [Stripe API release](https://docs.stripe.com/changelog/basil/2025-03-31/deprecate-subscription-current-period-start-and-end) last year. So presumably these properties could be there for some backwards compatibility? But when I boot up KPI locally and sync Stripe data, djstripe uses a more recent API version that doesn't actually have those subscription-level keys anymore.

jnm: Oo :eyes: I'll butt out as I don't have time to fire up everything locally right now with this branch. If you and @noliveleger want to have a discussion about Stripe testing overall and think I'd be helpful from an architectural standpoint, LMK on Zulip

PR #6867, opened 2026-03-25
refactor(xml): consolidate processing and remove defusedxml DEV-1932

jnm:
### 🗒️ Checklist

1. [x] run linter locally

### 💭 Notes

Remove defusedxml and consolidate XML parsing (from strings) and serialization (to strings) into common utility functions. Files that use minidom DOM manipulation extensively are intentionally left with direct imports. The `*requirements.txt` Python dependency files continue to show defusedxml since it remains a dependency of djangorestframework-xml and pyxform.

Rationale: Python 3.12's Expat and lxml 5.4 are hardened against XXE, entity expansion bombs, and external DTD fetching by default, making defusedxml unnecessary. The recommendation to use defusedxml has been removed from the official Python documentation in https://github.com/python/cpython/pull/135294.

**No changes to behavior** are expected.

### DIFF
--- a/kobo/apps/audit_log/tests/test_project_history_logs.py
+++ b/kobo/apps/audit_log/tests/test_project_history_logs.py
@@ -4,7 +4,6 @@ import io
 import json
 import uuid
 from unittest.mock import patch
-from xml.etree import ElementTree as ET
 
 import responses
 from ddt import data, ddt, unpack
@@ -1852,7 +1851,7 @@ class TestProjectHistoryLogs(BaseAuditLogTestCase):
             'formhub': {'uuid': self.asset.deployment.xform.uuid},
             '_uuid': str(uuid_),
         }
-        xml = ET.fromstring(
+        xml = fromstring_preserve_root_xmlns(
             dict2xform(submission_data, self.asset.deployment.xform.id_string)
         )
         xml.tag = self.asset.uid
@@ -1873,7 +1872,7 @@ class TestProjectHistoryLogs(BaseAuditLogTestCase):
             endpoint,
             kwargs=kwargs,
         )
-        data = {'xml_submission_file': SimpleUploadedFile('name.txt', ET.tostring(xml))}
+        data = {'xml_submission_file': SimpleUploadedFile('name.txt', xml_tostring(xml).encode())}
         # ensure anonymous users are allowed to submit
         self.asset.assign_perm(perm=PERM_ADD_SUBMISSIONS, user_obj=AnonymousUser())
 
--- a/kobo/apps/audit_log/tests/test_project_history_logs.py
+++ b/kobo/apps/audit_log/tests/test_project_history_logs.py
@@ -1872,7 +1872,11 @@ class TestProjectHistoryLogs(BaseAuditLogTestCase):
             endpoint,
             kwargs=kwargs,
         )
-        data = {'xml_submission_file': SimpleUploadedFile('name.txt', xml_tostring(xml).encode())}
+        data = {
+            'xml_submission_file': SimpleUploadedFile(
+                'name.txt', xml_tostring(xml).encode()
+            )
+        }
         # ensure anonymous users are allowed to submit
         self.asset.assign_perm(perm=PERM_ADD_SUBMISSIONS, user_obj=AnonymousUser())
 
--- a/kobo/apps/hook/tests/test_parser.py
+++ b/kobo/apps/hook/tests/test_parser.py
@@ -1,11 +1,9 @@
 import json
 import re
 
-from lxml import etree
-
 from kpi.constants import SUBMISSION_FORMAT_TYPE_XML
 from kpi.utils.strings import to_str
-from kpi.utils.xml import check_lxml_fromstring
+from kpi.utils.xml import fromstring_preserve_root_xmlns, xml_tostring
 from .base import BaseHookTestCase
 
 
@@ -46,7 +44,7 @@ class ParserTestCase(BaseHookTestCase):
         submission_id = submissions[0]['_id']
         submission_uuid = submissions[0]['_uuid']
         service_definition = ServiceDefinition(hook, submission_id)
-        expected_etree = check_lxml_fromstring(
+        expected_elements = fromstring_preserve_root_xmlns(
             f'<{self.asset.uid} id="{self.asset.uid}">'
             f'   <group1>'
             f'      <q3>¿Cómo está en el grupo uno la segunda vez?</q3>'
@@ -64,17 +62,15 @@ class ParserTestCase(BaseHookTestCase):
             f'</{self.asset.uid}>'
         )
 
-        expected_xml = etree.tostring(
-            expected_etree,
-            pretty_print=True,
+        expected_xml = xml_tostring(
+            expected_elements,
             xml_declaration=True,
-            encoding='utf-8',
         )
 
         def remove_whitespace(str_):
-            return re.sub(r'>\s+<', '><', to_str(str_))
+            return re.sub(r'>\s+<', '><', to_str(str_)).strip()
 
         self.assertEqual(
             remove_whitespace(service_definition._get_data()),
-            remove_whitespace(expected_xml.decode()),
+            remove_whitespace(expected_xml),
         )
--- a/kobo/apps/openrosa/apps/api/utils/xml.py
+++ b/kobo/apps/openrosa/apps/api/utils/xml.py
@@ -27,9 +27,15 @@ def extract_confirmation_message(xml_string: str) -> Optional[str]:
         return
 
     confirmation_message_xpath = confirmation_message_xpath.strip()
-    _, submit_message_root_tag, *other_parts = (
-        confirmation_message_xpath.split('/')
-    )
+    parts = confirmation_message_xpath.split('/')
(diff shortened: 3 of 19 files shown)


kobo/apps/audit_log/tests/test_project_history_logs.py
```diff
@@ -1873,7 +1872,7 @@ def test_add_submission(self, user_type, v1):
             endpoint,
             kwargs=kwargs,
         )
-        data = {'xml_submission_file': SimpleUploadedFile('name.txt', ET.tostring(xml))}
+        data = {'xml_submission_file': SimpleUploadedFile('name.txt', xml_tostring(xml).encode())}
```
jnm: `tostring` sounds like it should return a `str` maybe

kobo/apps/hook/tests/test_parser.py
```diff
@@ -64,17 +62,15 @@ def test_xml_parser(self):
...
         )
 
-        expected_xml = etree.tostring(
-            expected_etree,
-            pretty_print=True,
+        expected_xml = xml_tostring(
+            expected_elements,
             xml_declaration=True,
-            encoding='utf-8',
         )
```
jnm, line 68: what's the return type here? are we pretty printing all the time?

kobo/apps/hook/tests/test_parser.py
```diff
@@ -64,17 +62,15 @@ def test_xml_parser(self):
...
 
         def remove_whitespace(str_):
-            return re.sub(r'>\s+<', '><', to_str(str_))
+            return re.sub(r'>\s+<', '><', to_str(str_)).strip()
 
         self.assertEqual(
             remove_whitespace(service_definition._get_data()),
-            remove_whitespace(expected_xml.decode()),
+            remove_whitespace(expected_xml),
```
jnm, line 75: i guess it's `bytes` vs. `bytes`? that could be fine, but still maybe confusing based on the method name

kobo/apps/openrosa/apps/api/utils/xml.py
```diff
@@ -1,48 +1,40 @@
...
-
-    namespaces = root.nsmap
-
-    # Extract the kobo:submitMessage attribute from the root element
     try:
-        confirmation_message_xpath = root.xpath(
-            '@kobo:submitMessage', namespaces=namespaces
+        root = fromstring_preserve_root_xmlns(xml_string)
+    except ParseError:
+        logging.error(
```
jnm, line 16: this is intended as a stand-in for `recover=True`; i should learn more about what that actually does

kobo/apps/audit_log/tests/test_project_history_logs.py
jnm: d'oh, it does return a string :facepalm:

kobo/apps/hook/tests/test_parser.py
jnm: nope, `str` vs. `str`.

kobo/apps/hook/tests/test_parser.py
jnm: it's a `str`. the pretty-printing is ultimately irrelevant because of `remove_whitespace()` being called on the result, but `xml_tostring()` does *not* pretty print by default

kobo/apps/openrosa/apps/api/utils/xml.py
jnm: i'm not sure the rationale behind the old `recover=True`, but it wouldn't be enough to handle malformed XML, since other submission parsing logic would still crash. it doesn't seem helpful to preserve this one-off isolated leniency

kobo/apps/openrosa/apps/api/utils/xml.py
```diff
@@ -1,48 +1,40 @@
...
+            'Failed to parse XML for confirmation message', exc_info=True
         )
-    except (etree.XPathEvalError, TypeError):
         return
 
+    # Extract the kobo:submitMessage attribute from the root element
+    confirmation_message_xpath = root.attrib.get('{http://kobotoolbox.org/xforms}submitMessage')
```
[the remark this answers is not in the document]
jnm: Thanks! Yes, for this custom confirmation feature, the verbatim namespace here must be used.

kobo/apps/hook/tests/test_parser.py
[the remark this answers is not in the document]
jnm: Well, the test does pass :wink: but let's not be satisfied by that alone.

Allow Claude to retort:

Greptile claims that `ET.tostring(el, encoding='unicode', xml_declaration=True)` raises `LookupError` on Python 3.12. It doesn't — here's why.

The relevant code in CPython's `ElementTree.write()` is:

~~~python
if method == "xml" and (xml_declaration or
        (xml_declaration is None and
         encoding.lower() != "unicode" and
         declared_encoding.lower() not in ("utf-8", "us-ascii"))):
    write("<?xml version='1.0' encoding='%s'?>\n" % (declared_encoding,))
~~~

When `xml_declaration=True`, the `or` short-circuits immediately. The `encoding.lower() != "unicode"` check is only in the `xml_declaration is None` branch, which is never reached. The declaration is written unconditionally with `declared_encoding='utf-8'` (derived from `StringIO` having no `.encoding` attribute, falling back to `'utf-8'` in `_get_writer`).

Verified empirically on the project's actual Python 3.12.3:

~~~python
>>> ET.tostring(ET.fromstring('<a/>'), encoding='unicode', xml_declaration=True)
"<?xml version='1.0' encoding='utf-8'?>\n<a />"
~~~

No `LookupError`. The test passes legitimately. Furthermore, the output is equivalent to what the old lxml code produced — `etree.tostring(el, encoding='utf-8', xml_declaration=True)` returned `bytes` with the same `encoding='utf-8'` declaration, which was then `.decode()`d to `str`. Both before and after, the comparison value is a Python `str` containing an `encoding='utf-8'` declaration, and both are whitespace-stripped before assertion anyway.

kobo/apps/openrosa/apps/api/utils/xml.py
```diff
@@ -1,48 +1,40 @@
...
+
     if not confirmation_message_xpath:
         return
 
-    confirmation_message_xpath = confirmation_message_xpath[0].strip()
     _, submit_message_root_tag, *other_parts = (
         confirmation_message_xpath.split('/')
     )
-    confirmation_message_xpath = f'/{root.tag}/' + '/'.join(other_parts)
+    confirmation_message_xpath = './' + '/'.join(other_parts)
```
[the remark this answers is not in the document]
jnm: Thanks! I restored the `strip()`

kobo/apps/hook/tests/test_parser.py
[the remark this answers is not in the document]
jnm: (will mark as resolved: as we've learned, Greptile edits its original comment; it does not post new comments or replies)

kobo/apps/openrosa/apps/logger/xform_instance_parser.py
```diff
@@ -8,14 +8,14 @@
...
 import six
-from defusedxml import minidom
 from django.utils.encoding import smart_str
 from django.utils.translation import gettext as t
 from pyxform.survey_element import SurveyElement
 
 from kobo.apps.openrosa.apps.logger.exceptions import InstanceEmptyError
 from kobo.apps.openrosa.libs.utils.common_tags import XFORM_ID_STRING
 from kpi.utils.log import logging
+from kpi.utils.xml import minidom_parsestring
```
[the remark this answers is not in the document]
jnm: From Opus 4.6:
> Wrong. CodeRabbit claims Python's stdlib minidom doesn't disable external entity resolution by default. This is incorrect for Python 3.12 — the underlying Expat parser does not resolve external entities and has entity expansion limits. We discussed this at the start of the conversation. The whole point of this PR is that defusedxml is unnecessary on modern Python. CodeRabbit is repeating stale security advice.

kpi/utils/xml.py
```diff
@@ -115,27 +62,23 @@ def edit_submission_xml(
     element.text = value
 
 
+def minidom_parsestring(text: Union[str, bytes]) -> minidom.Document:
```
noliveleger: Would you mind to keep the ordering (A->Z) inside the file please 🙏

kpi/utils/xml.py
```diff
@@ -255,15 +198,20 @@ def process_node(node_: etree._Element, xpath_matches_: list):
...
+def xml_tostring(el: ET.Element, **kwargs) -> str:
     """
     Thin wrapper around `ElementTree.tostring()` as a step toward a future
     where all XML handling is done in this file
     """
+    if 'encoding' in kwargs:
+        raise NotImplementedError(
+            'Do not pass an `encoding` argument. The returned encoding is'
+            ' always `unicode`.'
+        )
```
noliveleger, line 210: ❤️

noliveleger (changes requested):
nitpick for function ordering. Otherwise, LGTM. Nice job!

kpi/utils/xml.py
jnm: @noliveleger done at last!

noliveleger (approved):
LGTM

PR #6906, opened 2026-04-01
fix(datatable): apply frozen column left offset to header styles and reset on unfreeze DEV-147

tesster7:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Fixed a bug in the Data Table where unfreezing a column after scrolling left the column header stuck in the wrong position.

### 📖 Description
When a column was frozen in the Data Table and the table was scrolled horizontally, unfreezing the column would leave the column header pinned at the scrolled position while the rest of the column snapped back to its correct place — causing a visual mismatch. The header now correctly returns to its original position when a column is unfrozen.

### 💭 Notes
https://github.com/user-attachments/assets/f60072ba-4ec4-4bad-b206-df27566cfb3f

### 👀 Preview steps
1. ℹ️ Have an account and a project with multiple columns in the Data Table
2. Go to the Data > Table view
3. Right-click a column header and freeze a column
4. Scroll the table horizontally to the right
5. Unfreeze the column
6. 🔴 [on main] notice the column header stays stuck at the scrolled position while the column body snaps back
7. 🟢 [on PR] notice the column header and body both return to their correct position together


magicznyleszek (approved):
Generally works and fixes the issue. There is another issue found by greptile. I tried to quickly find a way to fix it and it will require adding more complexity to the code. It is more of an edge case and is fixed as soon as user scrolls, so I would say let's leave it.

tiritea:
I was just playing around with freeze/unfreeze and ran across this, but seemingly also other issues as well. Can you please review the following short vid and check whether all the observed issues are going to be addressed by this fix (or do we need more tickets, at least to track if not fix?). There seems to be several issues with the table header getting messed up when you freeze and/or unfreeze something, which seems to be related somewhat to the specific alignment of the columns. Its hard to precisely pin down what exactly the trigger is, but if you play around with horizontally scrolling the table, and freezing/unfreezing stuff, then all sorts of weird things soon start to happen...

https://github.com/user-attachments/assets/4c34fc3b-97d4-4559-9195-cc238503423f

attn @tesster7

PR #6911, opened 2026-04-01
fix(tests): restore `_get_endpoint` helper usage for API endpoint resolution DEV-1952

rajpatel24:
### 📣 Summary
Restore usage of _get_endpoint in unit tests to ensure URLs are generated using the API namespace helper.

### 📖 Description
This PR restores the use of the `_get_endpoint` helper method in multiple unit tests.

The helper constructs endpoint names using the configured API namespace defined in `BaseTestCase`, ensuring URLs are generated correctly via `reverse()`. Removing it caused tests to bypass the namespace logic and made them less flexible.

Keeping `_get_endpoint` allows the same tests to continue working if new API versions are introduced (e.g., v3) while maintaining consistent endpoint resolution across test cases.

The original PR [kpi#6585](https://github.com/kobotoolbox/kpi/pull/6585) was merged before this feedback was received, so this PR reintroduces the helper usage accordingly.

### DIFF
--- a/kobo/apps/accounts/tests/test_api_tos.py
+++ b/kobo/apps/accounts/tests/test_api_tos.py
@@ -9,7 +9,7 @@ from kpi.tests.base_test_case import BaseTestCase
 
 class TOSTestCase(BaseTestCase):
     def setUp(self) -> None:
-        self.url = reverse('tos')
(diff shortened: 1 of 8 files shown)


noliveleger (approved):
LGTM

PR #6895, opened 2026-03-30
feat(metadata): add extra project metadata inputs to project settings DEV-1529 

RuthTurk:
### 🗒️ Checklist

1. [X] run linter locally
2. [X] update developer docs (API, README, inline, etc.), if any
3. [X] for user-facing doc changes create a Zulip thread at `#Support Docs Updates`, if any
4. [X] draft PR with a title `<type>(<scope>)<!>: <title> DEV-1234`
5. [X] assign yourself, tag PR: at least `Front end` and/or `Back end` or `workflow`
6. [X] fill in the template below and delete template comments
7. [X] review thyself: read the diff and repro the preview as written
8. [X] open PR & confirm that CI passes & request reviewers, if needed
9. [ ] delete this section before merging

### 📣 Summary
Superuser-defined custom project metadata fields are now supported on the frontend and appear in the project settings.

### 👀 Preview steps

1. ℹ️ Have several different types (text, single, multi select) of "Extra Project Metadata Fields" in the Django admin. Use options and labels with translations, as in:
Label:
```json
{"pt": "Cor preferida", "default": "Favorite color"}
```

Options:
```json
[{"name": "red", "label": {"pt": "Vermelho", "default": "Red"}}, {"name": "blue", "label": "Blue"}, {"name": "green", "label": "Green"}]
```

3. Log into a regular user account
4. Create a new project
5. 🟢 [on PR] notice that each custom metadata field is present in the Project Settings modal and takes up one line in the form
6. Play around and ensure that validation and translations work correctly
7. 🟢 Ensure that all the regular project metadata fields are rendering correctly

PR #6923, opened 2026-04-06
feat(metadata): update project summary to include extra project metadata DEV-1530

RuthTurk:
### 🗒️ Checklist

1. [X] run linter locally
2. [X] update developer docs (API, README, inline, etc.), if any
3. [X] for user-facing doc changes create a Zulip thread at `#Support Docs Updates`, if any
4. [X] draft PR with a title `<type>(<scope>)<!>: <title> DEV-1234`
5. [X] assign yourself, tag PR: at least `Front end` and/or `Back end` or `workflow`
6. [X] fill in the template below and delete template comments
7. [X] review thyself: read the diff and repro the preview as written
8. [X] open PR & confirm that CI passes & request reviewers, if needed

### 📣 Summary

Update the Project Summary page to show the custom metadata fields configured during the project setup process.

### 👀 Preview steps

1. ℹ️ Have several different types (text, single, multi select) of "Extra Project Metadata Fields" in the Django admin
2. Log into a regular user account
3. Create a new project with some extra project metadata
5. 🟢 [on PR] navigate to the Project Summary after the survey has been created and notice that the extra project metadata are displayed
7. 🟢 Ensure that the data is displayed correctly and that multi-select options are displayed as a comma-separated list

### DIFF
--- a/jsapp/js/components/formSummary/formSummaryProjectInfo.tsx
+++ b/jsapp/js/components/formSummary/formSummaryProjectInfo.tsx
@@ -5,11 +5,12 @@ import { getCountryDisplayString, getSectorDisplayString } from '#/assetUtils'
 import bem from '#/bem'
 import AssetStatusBadge from '#/components/common/assetStatusBadge'
 import Avatar from '#/components/common/avatar'
+import { EXTRA_PROJECT_METADATA_FIELD_TYPES } from '#/constants'
 import type { AssetResponse, PaginatedResponse, SubmissionResponse } from '#/dataInterface'
 import { dataInterface } from '#/dataInterface'
 import envStore from '#/envStore'
 import sessionStore from '#/stores/session'
-import { formatTime } from '#/utils'
+import { currentLang, formatTime } from '#/utils'
 
 interface FormSummaryProjectInfoProps {
   asset: AssetResponse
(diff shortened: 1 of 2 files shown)


jsapp/js/components/formSummary/formSummaryProjectInfo.tsx
```diff
@@ -169,6 +172,31 @@ export default function FormSummaryProjectInfo(props: FormSummaryProjectInfoProp
             )}
           </bem.FormView__group>
         )}
+
+        {/* extra metadata */}
+        {extraFieldDefinitions?.length > 0 && (
+          <bem.FormView__group m='items'>
+            {extraFieldDefinitions.map((field) => {
+              const valueEntry = settings[field.name]
```
jamesrkiger, line 180: Similar to my comment over [here](https://github.com/kobotoolbox/kpi/pull/6895/changes#r3089232576), I think this is just a bit too much JS for the template. Can you find a way to do this work before getting to the template? Even a simple `getExtraMetadataDisplayValue` function defined above, for example, would at least keep template/logic concerns a bit more separate.

jamesrkiger (changes requested):

jsapp/js/components/formSummary/formSummaryProjectInfo.tsx
```diff
@@ -42,6 +42,9 @@ export default function FormSummaryProjectInfo(props: FormSummaryProjectInfoProp
   // Support custom labels for project metadata, if defined
   const metadata = envStore.data.getProjectMetadataFieldsAsSimpleDict()
 
+  const extraFieldDefinitions = envStore.data.extra_project_metadata_fields
+  const settings = (props.asset.settings || {}) as Record<string, any>
```
jamesrkiger, line 46: Let's not define this separately. The rest of the component just accesses props.asset.settings, so I think we should stick with that unless we want to do a bigger rewrite.

jamesrkiger (approved):
lgtm

PR #6928, opened 2026-04-08
fix(mfa): prevent superusers from deactivating MFA when `SUPERUSER_AUTH_ENFORCEMENT` is active DEV-1926

RuthTurk:
### 📣 Summary
Ensure administrative accounts remain protected by preventing superusers from disabling two-factor authentication when a security enforcement configuration is enabled. 

### 💭 Notes
- Adds `EnforceSuperuserMFA` permission to ensure that if `SUPERUSER_AUTH_ENFORCEMENT` is active, the deactivation endpoint will return a 403 Forbidden
- Does not let the user open the MFA modal to deactivate on the Security page 
- Exposes `is_staff` and `is_superuser `(read-only) via the `/me/` endpoint and adds `superuser_auth_enforcement` to the `/environment/` endpoint. These additions allow the frontend to verify these values on the Security page when determining if the MFA button needs to be disabled. 

### 👀 Preview steps

2. Login to the Django admin and activate the `SUPERUSER_AUTH_ENFORCEMENT` under **Constance > Config**
4. Login to a regular user account
5. Activate Two-factor authentication under **Account Settings > Security** 
6. Login to the Django admin again as the superuser account and promote the regular user to superuser
7. Log back in as the newly promoted user and navigate to **Account Settings > Security** 
8. 🔴 [on main] Toggle the "Disabled" button for Two-factor authentication and it should successfully disable
9. 🟢 [on PR] The "Disabled" button for Two-factor authentication is grayed out with a tooltip when hovering over the button that says "Superusers cannot deactivate their MFA."
10. Test and ensure these different scenarios work:
- IF user MFA is OFF -> user can't be made superuser
- IF user MFA in ON -> user can be made superuser
- IF superuser -> MFA can't be deactivated, either in the interface or in Django, by the user or an admin.


kpi/serializers/current_user.py
```diff
@@ -60,6 +60,8 @@ class Meta:
             'first_name',
             'last_name',
             'email',
+            'is_staff',
```
noliveleger: I don't think it's useful to expose this file. 
I don't find it in the front-end code either?
I would remove it if we don't need it.

kobo/apps/accounts/mfa/tests/test_api.py
```diff
@@ -141,3 +141,50 @@ def test_mfa_deactivation_syncs_user_profile(self):
...
+            response.json()['detail'],
+            'Superusers cannot deactivate MFA while SUPERUSER_AUTH_ENFORCEMENT'
+            ' is active.',
+        )
+
+        mfa_method = MfaMethodsWrapper.objects.get(user=superuser)
+        self.assertTrue(mfa_method.is_active)
+
+    @override_config(SUPERUSER_AUTH_ENFORCEMENT=True)
+    def test_superuser_deactivation_blocked_when_enforced(self):
```
noliveleger: It seems to test the same thing as `test_superuser_deactivation_prohibited_when_enforced` (except for the message).

hub/admin/extend_user.py
```diff
@@ -31,9 +31,9 @@
 from .mixins import AdvancedSearchMixin
 
 
-def validate_superuser_auth(obj) -> bool:
+def validate_superuser_auth(obj, is_superuser_requested: bool) -> bool:
```
noliveleger: The name of this method is not really clear IMO. 
I would change to something to "can_save_as_superuser()" which would make it more obvious. Especially, in the tests. 

What do you think? 
for the records: 🤖 suggests `superuser_mfa_requirement_met`

kobo/apps/accounts/mfa/admin.py
```diff
@@ -20,8 +50,12 @@ def has_add_permission(self, request, obj=None):
 
     def get_readonly_fields(self, request, obj=None):
         readonly_fields = list(super().get_readonly_fields(request, obj))
-        if obj and not obj.is_active:
-            readonly_fields.append('is_active')
+
+        if obj and obj.pk:
+            persisted_obj = MfaMethodsWrapper.objects.get(pk=obj.pk)
+            if not persisted_obj.is_active:
```
noliveleger: Let's leave a comment here to explain the goal of this extra query. (TBH, not entirely sure why it's needed here).

It could be simplified and more efficient with 

```diff
-            persisted_obj = MfaMethodsWrapper.objects.get(pk=obj.pk)
-            if not persisted_obj.is_active:
+           MfaMethodsWrapper.objects.filter(pk=obj.pk, is_active=True).exists():
```

kobo/apps/accounts/mfa/permissions.py
```diff
@@ -6,3 +6,18 @@
...
+class EnforceSuperuserMFA(BasePermission):
+    message = (
+        'Superusers cannot deactivate MFA while '
+        'SUPERUSER_AUTH_ENFORCEMENT is active.'
+    )
+
+    def has_permission(self, request, view):
+        if (
+            getattr(request.user, 'is_superuser', False)
+            and getattr(config, 'SUPERUSER_AUTH_ENFORCEMENT', False)  # noqa: W503
```
noliveleger: No need to use getattr. 

`SUPERUSER_AUTH_ENFORCEMENT` is always present in Constance.  

Same comment for getattr(request.user, 'is_superuser'). It's a property of the class User and AnonymousUser.

jsapp/js/account/security/mfa/mfaSection.component.tsx
```diff
@@ -181,13 +201,17 @@ export default class SecurityRoute extends React.Component<{}, SecurityState> {
...
-        <div className={styles.options}>
-          <ToggleSwitch
-            label={this.state.isMfaActive ? t('Enabled') : t('Disabled')}
-            checked={this.state.isMfaActive}
-            onChange={this.onToggleChange.bind(this)}
-          />
-        </div>
+        {isDisabled ? (
+          <Tooltip
+            text={t('Superusers cannot deactivate their MFA.')}
```
noliveleger: nitpick: I think we should remove the final dot in the tooltip.

noliveleger (changes requested):

noliveleger (approved):
LGTM
As per discussed: Please retry stuck CI tasks before merging it.

PR #6938, opened 2026-04-13
fix(subsequences): bypass usage limits for LLM answer verification DEV-1969

rajpatel24:
### 📣 Summary
This PR ensures that users can verify or delete existing automated qualitative analysis responses without triggering LLM usage limit checks, as these actions do not invoke external APIs or consume tokens.


rgraber (approved):
LGTM

PR #6939, opened 2026-04-14
feat(mfa): enable MFA unconditionally for all users DEV-1799

rajpatel24:
### 📣 Summary
This PR removes the `MfaAvailableToUser` whitelist model and all subscription-based restrictions tied to Multi-Factor Authentication. MFA is now universally available to all KoboToolbox users, governed only by the global `MFA_ENABLED` configuration.

### 📖 Description
Historically, MFA was gated behind paid subscriptions or a specific per-user whitelist (`MfaAvailableToUser`). As part of our push to make security features available to everyone, this PR completely removes those restrictions.

This PR:
- Deletes the `MfaAvailableToUser` model and its corresponding Django Admin interface.
- Removes the subscription-check logic from `MfaAdapter.is_mfa_enabled()`.
- Simplifies `permissions.py` to allow MFA access based solely on the global `config.MFA_ENABLED` setting.
- Cleans up `environment.py` and related API endpoints that previously exposed the whitelist status to the frontend.
- Refactors unit tests in `test_login.py`, `test_mfa_login.py`, and `test_api_environment.py` to remove whitelist/subscription matrices and verify universal access.

### 👀 Preview steps
1. Log in to the application as any standard, free-tier user.
2. Navigate to the Account Settings/Security page.
3. 🟢 [On PR] Notice that you can set up MFA via the Authenticator App without needing a paid subscription or superuser intervention.
4. Log in as a superuser and go to the Django Admin panel.
5. 🟢 [On PR] Notice that the `per-user availabilities` table is completely gone.


magicznyleszek (approved):
FE part looks good

kobo/apps/accounts/mfa/adapter.py
```diff
@@ -23,9 +21,7 @@ def is_mfa_enabled(self, user, types=None) -> bool:
             and MfaMethodsWrapper.objects.filter(user=user, is_active=True).first()
             is not None
         )
-        mfa_allowed = mfa_allowed_for_user(user)
-        inactive_subscription = user_has_inactive_paid_subscription(user.username)
-        return mfa_active and (mfa_allowed or inactive_subscription)
+        return mfa_active and config.MFA_ENABLED
```
noliveleger, line 24: We could avoid to call 
```python
mfa_active_super = super().is_mfa_enabled(user, types)
        mfa_active = (
            mfa_active_super
            and MfaMethodsWrapper.objects.filter(user=user, is_active=True).first()
            is not None
        )
```

if we know that config.MFA_ENABLED is False. 

What about 

```python 
if not config.MFA_ENABLED:
    return False 

mfa_active_super = super().is_mfa_enabled(user, types)
        mfa_active = (
            mfa_active_super
            and MfaMethodsWrapper.objects.filter(user=user, is_active=True).exists()  # small optimization.
        )
```

noliveleger (changes requested):

kobo/apps/accounts/mfa/adapter.py
rajpatel24: Thanks, that makes sense!

noliveleger (approved):
LGTM

PR #6943, opened 2026-04-14
fix(qual): default to false if verification missing DEV-1982

rgraber:
### 🗒️ Checklist

1. [ ] run linter locally

### 📣 Summary
Fix an error loading the data table with older submissions

### 📖 Description
Loading a data table for a survey with older submissions that had QA data from before verification was released caused a 500 and displayed a long error message.

### 💭 Notes
Defaults the verification column to 'False' if not present. Also fixes a related issue in the migration code that was setting `_dateAccepted,` which is not a field used for qual questions. This PR does not address the issue of qual questions that have already been migrated. There is other work in progress to do a large-scale migration and that work should probably be combined with fixing this issue for pre-existing data.

### 👀 Preview steps
Start out on the 2.026.07 branch to create data using the old code

1. Create a project with an audio question and add a submission
2. Add a QA question and answer to the submission
3. Switch to main
4. Navigate to the data table
5. 🔴 [on main] Error
6. 🟢 [on PR] the table loads with the verified column set to False

### DIFF
--- a/kobo/apps/subsequences/actions/qual.py
+++ b/kobo/apps/subsequences/actions/qual.py
@@ -434,7 +434,7 @@ class BaseQualAction(BaseAction):
                 'xpath': self.source_question_xpath,
                 'labels': qual_question.get('labels', {}),
                 SORT_BY_DATE_FIELD: selected_version[self.DATE_CREATED_FIELD],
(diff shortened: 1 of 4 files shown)


RuthTurk (approved):
LGTM

PR #6945, opened 2026-04-16
fix: prevent IntegrityError when running `create_kobo_superuser` management command DEV-1987

noliveleger:
### 📣 Summary

Running `create_kobo_superuser` could crash with an `IntegrityError` when trying to create the user profile.

### 💭 Notes

- `UserProfile.objects.create` and `EmailAddress.create` replaced with `get_or_create` to avoid `IntegrityError` on duplicate records.
- Secondary fix: when `ProgrammingError` is raised inside `create_superuser` (KC DB not ready, signal-level failure), the `user` variable was never assigned, causing a `NameError` on the subsequent `EmailAddress.objects.get_or_create(user=user, ...)` call. Fix: retrieve the already-saved user via `User.objects.get(username=super_username)`.
- Three unit tests added: happy path, `ProgrammingError` recovery (mocked), and early-exit when user already exists.

### Preview steps:
1. from release branch, run `KOBO_SUPERUSER_USERNAME=foobar ./manage.py create_kobo_superuser` and receive an IntegrityError
2. from this branch, receive the message "Superuser xxx successfully created."

### DIFF
--- a/kpi/management/commands/create_kobo_superuser.py
+++ b/kpi/management/commands/create_kobo_superuser.py
@@ -35,13 +35,18 @@ class Command(BaseCommand):
             )
             sys.exit(1)
 
-        # Fix any superuser missing a user profile or email address
+        # Validate superuser email address
         EmailAddress.objects.get_or_create(
-            user=user, email=user.email, verified=True, primary=True
+            user=user,
+            email=user.email,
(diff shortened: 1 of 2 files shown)


rajpatel24 (approved):
LGTM

PR #6947, opened 2026-04-16
feat(notification): add new notification with themes and stories DEV-1438

duvld:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Adds mantine Notification component and apply our theme to it

### 👀 Preview steps

1. ℹ️ have storybook working locally
2. Navigate to "Design system" > "Notification"
5. 🟢 [on PR] notice that the notification theme matches the mockups


jsapp/js/components/common/Notification.stories.tsx
```diff
@@ -0,0 +1,43 @@
+import { Notification } from '@mantine/core'
+import type { Meta, StoryObj } from '@storybook/react-webpack5'
+import { IconNames } from '#/k-icons'
+import { recordKeys } from '#/utils'
+import Icon from './icon'
```
magicznyleszek, line 5: Given the new icon PR https://github.com/kobotoolbox/kpi/pull/6952 let's wait for it to be merged and use the new icon

jsapp/js/theme/kobo/Notification.module.css
```diff
@@ -0,0 +1,8 @@
+.root {
+  background-color: var(--mantine-color-gray-1);
+}
+
+.title {
+  font-size: small;
```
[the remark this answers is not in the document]
magicznyleszek: Yeah, let's use mantine's variable

jsapp/js/components/common/Notification.stories.tsx
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export default meta
+
+type Story = StoryObj<typeof Notification>
+
+export const Default: Story = {
+  args: {
+    title: 'Your transcripts are on their way!',
+    children: 'Click here to monitor your progress or to cancel this job',
```
magicznyleszek, line 40: Let's have a link in the text, so it would be testable as a styled thing :)

jsapp/js/theme/kobo/Notification.module.css
```diff
@@ -0,0 +1,8 @@
...
+  font-weight: 600;
+}
```
magicznyleszek, line 8: Also the icon should be smaller (Figma has 18x18px circle icon as example). Also let's use mantine's variable if it matches the 18px (or is very close)

magicznyleszek (changes requested):

jsapp/js/theme/kobo/Notification.module.css
duvld: `Notification` needs a jsx element as a prop for the icon, see the story implementation. The size should be specified through there, but I changed the story to `s` instead of `m` (which is at 18px)

magicznyleszek: Ok, so the icon size is something we can change by using icon param. But I see `mantine-Notification-icon` (the blue circle) is still 28px and should be 18px. I think we should change it in `Notification.module.css` and then use `xxs` icon in the story?

jsapp/js/components/common/Notification.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+type Story = StoryObj<typeof Notification>
+
+export const Default: Story = {
+  args: {
+    title: 'Your transcripts are on their way!',
+    icon: 'check',
+  },
+  render: (args) => (
+    <Notification {...args}>
+      <a href='#'>Click here</a> to monitor your progress or to cancel this job
```
magicznyleszek: The link should be lighter blue and underlined

jsapp/js/components/common/Notification.stories.tsx
magicznyleszek: Or maybe just see how it goes with the other PR… Like I don't think we should block this PR if it will be ready before the icon one

magicznyleszek (changes requested):

jsapp/js/theme/kobo/Notification.module.css
duvld: Ah I misunderstood, I set the blue circle to a fixed width now. I'm not sure why `xxs` for the icon though, I think `s` fits the designs better, but instead of going back forth on this I adjusted the story to allow changing the icon size

magicznyleszek (approved):
Added a commit that allows setting the message. LGTM now

jsapp/js/components/common/Notification.stories.tsx
```diff
@@ -0,0 +1,57 @@
...
+export const Default: Story = {
+  args: {
+    title: 'Your transcripts are on their way!',
+    message: '<a href="#">Click here</a> to monitor your progress or to cancel this job',
+    iconName: 'check',
+    iconSize: 's',
+  },
+  render: (args) => (
+    <Notification {...args} icon={args.iconName ? <Icon name={args.iconName} size={args.iconSize} /> : undefined}>
+      <span dangerouslySetInnerHTML={{ __html: DOMPurify.sanitize(args.message) }} />
```
jamesrkiger: @duvld In answer to your question about the DOMPurify package, I think we should step back a bit. Why do we need to use dangerouslySetInnerHTML here? Is the intention that we will be using it when the component is actually used in the app? That should be avoided. We will ideally want to get rid of existing uses of dangerouslySetInnerHTML in the app itself at some point in the future. But in the meantime we should avoid new uses of it. 

If we're not planning to use dangerouslySetInnerHTML with this component in the app itself, then do we really need to use it here? Is it just so someone using storybook can enter different text that includes links? Maybe just let them pick between a few pre-set HTML child options that you can pass as an arg instead. Or just stick with strings and don't let storybook users include links in this example. 

So my initial thought is that we don't need to add DOMPurify just for this use case. Let me know if I missed anything.

duvld: Makes sense, revert the changes and hard coding the notification message

PR #6955, opened 2026-04-21
feat(nlp): upgrade automatic Google transcription to Speech-to-Text v2 batch API DEV-1498

rajpatel24:
### 📣 Summary
This PR upgrades the automatic Google transcription service from the Speech-to-Text v1 API to the v2 API.

The implementation uploads audio files to Google Cloud Storage, submits a batch recognition job, and retrieves the resulting transcripts from JSON output files written by Google.

The service now dynamically resolves the correct region and model using the new languages configuration introduced in the languages app.

### 📖 Description
The new implementation follows the v2 architecture:
1. Audio files are uploaded to Google Cloud Storage
2. A BatchRecognize job is created
3. Google processes the audio asynchronously
4. Transcription results are written as JSON files to Cloud Storage
5. The service reads the output JSON and returns the transcript

The implementation supports two mechanisms for tracking long-running operations:
- Cache-based tracking: Used for single transcription requests.
- Database-based tracking: Future bulk transcription operations will store the Google operation id on `SubsequenceBulkActionItem`.

### 👀 Preview steps
Test the automatic Google transcription of the submitted audio, it should work without any issues.

!! Please do not forget to import the new list of languages in Django admin: 
[new_langauge_list.csv](https://github.com/user-attachments/files/28041494/new_langauge_list.csv)


kobo/apps/subsequences/integrations/google/google_transcribe.py
```diff
@@ -19,95 +27,116 @@
...
-            for result in results:
-                alternatives = result['alternatives']
-                transcript.append(alternatives[0]['transcript'])
-        else:
-            for result in response.results:
-                alternatives = result.alternatives
-                transcript.append(alternatives[0].transcript)
-        result_string = ' '.join(transcript)
-        return result_string
+        transcripts = []
```
rgraber, line 71: A comment here about what we expect the response to look like would be helpful, especially since it seems like there may be multiple formats

kobo/apps/subsequences/integrations/google/google_transcribe.py
```diff
@@ -147,47 +204,363 @@ def process_data(self, xpath: str, params: dict) -> dict:
...
+                value = self._read_batch_result(xpath, source_language)
+                self._clear_operation_reference(xpath, source_language, bulk_action_uid)
+                return {
+                    'status': 'complete',
+                    'value': value,
+                }
+            except TimeoutError:
+                return {'status': 'in_progress'}
+            except TranscriptionResultNotFound as err:
+                logging.error(f'No transcriptions found for {xpath=}: {err}')
```
rgraber, line 272: Can we add a hint here, like "it's possible Google was unable to transcribe the audio"

kobo/apps/subsequences/integrations/google/google_transcribe.py
```diff
@@ -147,47 +204,363 @@ def process_data(self, xpath: str, params: dict) -> dict:
...
             }
+        except (GoogleAPIError, GoogleCloudError, Exception) as err:
+            # Unable to reach Google to check the operation status.
+            # The transcription may still be running, so return "in_progress"
+            # to allow Celery to retry instead of marking the task as failed
+            logging.error(
+                f'Google infrastructure error while polling transcription '
+                f'for {xpath=}: {err}'
+            )
+            return {'status': 'in_progress'}
```
rgraber, line 351: Is there any risk of an infinite in-progress status if we have bad credentials or something?

kobo/apps/subsequences/integrations/google/google_transcribe.py
```diff
@@ -147,47 +204,363 @@ def process_data(self, xpath: str, params: dict) -> dict:
...
+    def _get_google_language_config(self, requested_language: str):
+        """
+        Resolve the speech configuration for the user request
+        """
+        try:
+            transcription_lang_service = TranscriptionService.objects.get(
+                code=GOOGLE_CODE
+            )
+        except TranscriptionService.DoesNotExist:
+            raise LanguageNotSupported(
```
rgraber: This seems a weird error to raise since it's not language specific

kobo/apps/subsequences/tests/test_automatic_google_transcription.py
```diff
@@ -86,6 +86,34 @@ def test_valid_automatic_transcription_data_passes_validation():
         action.validate_external_data(data)
 
 
+def test_plain_language_request_without_locale_can_complete():
```
rgraber, line 89: is this a result of this PR or a separate test?

kobo/apps/subsequences/integrations/google/google_transcribe.py
rajpatel24: Added a new except block to handle these things.

kobo/apps/subsequences/integrations/google/google_transcribe.py
rajpatel24: Updated the exception class

kobo/apps/subsequences/tests/test_automatic_google_transcription.py
rajpatel24: Yes, this test was added as part of this PR. Because the new v2 integration relies on dynamically resolving the Google region/model using the languages configuration app, I added this test to ensure that a basic language request (like 'en') safely falls back and completes even if the user doesn't provide a specific locale (like 'en-US').

I have also added a few new tests in `test_google_transcribe.py‎` to better test the flow.

kobo/apps/subsequences/integrations/google/google_transcribe.py
rgraber: The use of a generic Exception handler here still worries me. My instinct is that an exception we don't know about should fail the process rather than keep it in progress.

rgraber (commented):
Still a little concerned about the generic exception handler

rajpatel24:
> Still a little concerned about the generic exception handler

@rgraber  My original rationale was to protect the long-running transcription job from being killed by temporary, generic network blips (like a base socket timeout or connection reset) that might not be properly wrapped in a GoogleCloudError. Since these jobs can take hours, I wanted to rely on the Celery task's max-retry limit to ride out temporary hiccups rather than failing the whole transcription immediately.

However, I think you make a great point. Catching a generic Exception here risks swallowing real bugs, masking the stack trace, and trapping it in a useless retry loop. I'll remove the generic Exception catch so that only explicit Google infrastructure errors trigger a retry, and unexpected errors fail fast

rgraber (approved):
LGTM

PR #6968, opened 2026-04-27
fix(constance): `JsonSchemaFormField.clean()` must return parsed object DEV-2027

noliveleger:
### 📣 Summary

Saving certain admin settings (MFA help text, password guidance, user/project metadata fields) via the Django admin could cause crashes for end users, such as a broken MFA login flow or broken user profile pages.

### 💭 Notes

`JsonSchemaFormField.clean()` was returning `value` (raw JSON string) instead of `instance` (parsed Python object). Constance encodes the return value of `clean()` directly, so it stored a string in the DB. At runtime, code accessing these settings as dict/list received a string instead, causing `TypeError: string indices must be integers`.

Four constance keys are affected: 
- `MFA_LOCALIZED_HELP_TEXT`
- `CUSTOM_PASSWORD_GUIDANCE_TEXT` 
- `USER_METADATA_FIELDS`
- `PROJECT_METADATA_FIELDS`

A data migration (`hub/migrations/0020_fix_constance_json_field_corruption.py`) repairs existing corrupted values:
- Depends on `constance.0003_drop_pickle` so it runs after the pickle→JSON migration on prod
- Double-parses any value that is a JSON-encoded string to recover the actual dict/list
- Flushes the constance cache for the affected keys after repair
- Skips values that are already correct (dict, list, or `lazy_json_serializable` envelope)

### 👀 Preview steps

1. ℹ️ Have a superuser account
2. Go to Django admin → Constance → Config
3. Edit `MFA_LOCALIZED_HELP_TEXT` — save without changing anything
4. 🔴 [on release/2.026.13] Log out, try MFA login → crash (`string indices must be integers`)
5. 🟢 [on PR] Same flow → MFA login works normally

### DIFF
--- a/hub/migrations/0020_fix_constance_json_field_corruption.py
+++ b/hub/migrations/0020_fix_constance_json_field_corruption.py
@@ -1,0 +1,108 @@
+import json
+
+from constance import settings as constance_settings
+from django.core.cache import caches
+from django.db import migrations
+
+from kpi.utils.log import logging
+
+# These four constance keys use JsonSchemaFormField subclasses and must be
+# stored as dict/list. A bug in clean() caused them to be stored as plain JSON
+# strings instead. The constance 0003_drop_pickle migration preserved the
+# corruption: a pickled str becomes a JSON-encoded string in the DB, so
+# constance still reads back a str instead of a dict/list.
+KEYS_TO_FIX = [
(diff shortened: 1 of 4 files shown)


rajpatel24 (approved):
LGTM

PR #6973, opened 2026-04-28
fix(constance): repair migration 0020 for constance 4 envelope format DEV-2027

noliveleger:
### 📣 Summary

Follow-up to #6968. Migration `0020_fix_constance_json_field_corruption` was silently skipping all records modified by Constance own migration and not repairing corrupted constance values, causing a 500 error on `/environment/`.

### 💭 Notes

Constance 4 wraps every stored value in `{"__type__": "...", "__value__": ...}` (introduced by `constance.0003_drop_pickle`). The previous migration logic checked `if not isinstance(outer, str)` at the root level — since constance 4 always stores a dict envelope at the root, all records were skipped.

The corruption lives one level deeper, in `__value__`, which holds a JSON-encoded string instead of the actual dict/list.

This PR fixes the migration to:
- Detect the constance 4 envelope and inspect `__value__`
- Parse and repair `__value__` when it is a string (double-encoded)
- Keep the legacy constance 3 path (plain double-encoded string at root) as fallback

Also fixes a `SyntaxWarning: invalid escape sequence '\d'` in `0003_copy_organization_uid_to_id.py` (Python 3.12).

### 👀 Preview steps
1. Checkout `release/2.016.12`
2. Modify `MFA_LOCALIZED_HELP_TEXT` (for example)
3. Checkout `release/2.016.13` (wait for the migrations to be applied)
4. Try to login as a regular user and receive a 500 from `/environment/`
5. Checkout this branch
6. Rollback migration hub 0020, then reapplyt it.
3. Verify all 4 keys return `dict`/`list`:
   ```python
   from constance import config
   for key in ['MFA_LOCALIZED_HELP_TEXT', 'CUSTOM_PASSWORD_GUIDANCE_TEXT', 'USER_METADATA_FIELDS', 'PROJECT_METADATA_FIELDS']:
       val = getattr(config, key)
       print('OK' if isinstance(val, (dict, list)) else 'BROKEN', key, type(val).__name__)
   ```
4. 🟢 All 4 lines print `OK`
5. 🟢 `/environment/` returns 200

### DIFF
--- a/hub/migrations/0020_fix_constance_json_field_corruption.py
+++ b/hub/migrations/0020_fix_constance_json_field_corruption.py
@@ -25,13 +25,13 @@ def fix_constance_json_field_corruption(apps, schema_editor):
     of a JSON object/array.
 
     Root cause: JsonSchemaFormField.clean() returned the raw input string
-    instead of the parsed Python object. Constance encoded that string as JSON,
-    so the DB ended up with e.g. '"{\\"default\\": ...}"' for a field that
-    should contain '{"default": ...}'.
+    instead of the parsed Python object.
 
-    After this migration, each affected key is stored as a proper JSON
-    object/array, and the constance cache is invalidated so the corrected
-    values are picked up immediately.
+    Constance 4 wraps every stored value in {"__type__": ..., "__value__": ...}.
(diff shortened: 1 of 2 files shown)


rajpatel24 (approved):
LGTM

PR #6983, opened 2026-05-01
fix(icon): change way of identifying legacy icons DEV-2069

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Fixes a problem with some icon rendering.

### 💭 Notes

Some icons were rendering as empty/transparent because the code that identifies legacy svg icons was wrong - it was setting `stroke` attribute for all icons rather than just tabler ones.

### 👀 Preview steps

Unfortunately the issue happens only on deployed storybook (see https://storybook.kbtdev.org/?path=%2Fstory%2Fdesign-system-koboicon--mantine-integration-examples) and I haven't found a way to reproduce it locally. I suspect the problem happens with production build, and previous way of identifying the legacy svgs stopped working because of code minification (that renamed some things we we're relying on).


jamesrkiger (approved):
lgtm

PR #6936, opened 2026-04-13
refactor(connectProjects): migrate to react-query DEV-1968

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally
2. [x] update developer docs (API, README, inline, etc.), if any
3. [x] for user-facing doc changes create a Zulip thread at `#Support Docs Updates`, if any
4. [x] draft PR with a title `<type>(<scope>)<!>: <title> DEV-1234`
5. [x] assign yourself, tag PR: at least `Front end` and/or `Back end` or `workflow`
6. [x] fill in the template below and delete template comments
7. [x] review thyself: read the diff and repro the preview as written
8. [x] open PR & confirm that CI passes & request reviewers, if needed
9. [ ] delete this section before merging

### 📣 Summary

Internal code improvement. 

### 💭 Notes

Followup to #6931

Changes here:
- Removed `dataShareActions.js`
- `ConnectProjects` component no longer relies on Reflux actions, now runs on orval generated react-query hooks
- `ConnectProjectsImports` renders heading outside `<ul>` to make the HTML code semantic
- `ConnectProjectsSelect` small adjustments due to parent component changes
- `DataAttachmentColumnsForm` is a functional component now and uses orval react-query hooks instead of Reflux actions

### 👀 Preview steps

Use same preview steps as in #6931

PR #6985, opened 2026-05-01
feat(frontend): setup default invalidations DEV-2010

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 💭 Notes

A code change for future. An aftermath of #6936. Added default invalidations for few orval generated API hooks.

Changes here:
- `survey-data.ts` → `useAssetsPairedDataDestroy` invalidates the paired data list for the asset on settled
- `manage-projects-and-library-content.ts` → new file; `useAssetsPartialUpdate` invalidates both the paginated assets list and the specific asset item on settled

### 👀 Preview steps

Setup:

1. Use DevTools → **Network** tab
2. Open a project that has **Connect Projects** enabled

Test 1 → `useAssetsPairedDataDestroy` invalidates the paired-data list

1. Navigate to Project → Settings → Connect Projects
2. Make sure at least one source project is already attached
3. Clear the Network log
4. Click Remove on an attached source and confirm
5. 🟢 You should see a `DELETE /api/v2/assets/<uid>/paired-data/<paired-uid>/` request complete with `204`
6. 🟢 Immediately after, a `GET /api/v2/assets/<uid>/paired-data/` request fires automatically — this is the invalidation refetch
7. 🟢 Notice the removed source disappears from the list without a manual page reload

Test 2 → `useAssetsPartialUpdate` invalidates the asset list and item

1. Stay on the same Connect Projects panel
2. Clear the Network log
3. Toggle the Share data switch (on → off, or off → on).
4. 🟢 A `PATCH /api/v2/assets/<uid>/` request fires with `200`
5. 🟢 After it completes, two refetch requests fire:
   - `GET /api/v2/assets/` (the paginated list — `invalidatePaginatedList`)
   - `GET /api/v2/assets/<uid>/` (the specific asset — `invalidateItem`)
6. 🟢 Notice the sharing state reflects the server response. Other parts of the UI that depend on the asset would also pick up the refreshed data

### DIFF
--- a/jsapp/js/api/mutation-defaults/manage-projects-and-library-content.ts
+++ b/jsapp/js/api/mutation-defaults/manage-projects-and-library-content.ts
@@ -14,7 +14,7 @@ queryClient.setMutationDefaults(
         const assetsListQueryKey = getAssetsListQueryKey()
 
         invalidatePaginatedList(assetsListQueryKey)
-        // Invaliate infinite queries
+        // Invalidate infinite queries
         queryClient.invalidateQueries({
           predicate: ({ queryKey }) =>
             queryKey.length > assetsListQueryKey.length &&
--- a/jsapp/js/api/mutation-defaults/manage-projects-and-library-content.ts
+++ b/jsapp/js/api/mutation-defaults/manage-projects-and-library-content.ts
@@ -14,6 +14,7 @@ queryClient.setMutationDefaults(
         const assetsListQueryKey = getAssetsListQueryKey()
 
         invalidatePaginatedList(assetsListQueryKey)
+        // Invaliate infinite queries
         queryClient.invalidateQueries({
           predicate: ({ queryKey }) =>
             queryKey.length > assetsListQueryKey.length &&
--- a/jsapp/js/api/mutation-defaults/manage-projects-and-library-content.ts
+++ b/jsapp/js/api/mutation-defaults/manage-projects-and-library-content.ts
@@ -11,7 +11,15 @@ queryClient.setMutationDefaults(
   getAssetsPartialUpdateMutationOptions({
     mutation: {
       onSettled: (_data, _error, { uidAsset }) => {
-        invalidatePaginatedList(getAssetsListQueryKey())
+        const assetsListQueryKey = getAssetsListQueryKey()
+
+        invalidatePaginatedList(assetsListQueryKey)
+        queryClient.invalidateQueries({
+          predicate: ({ queryKey }) =>
+            queryKey.length > assetsListQueryKey.length &&
+            assetsListQueryKey.every((keyPart, index) => queryKey[index] === keyPart) &&
+            queryKey[queryKey.length - 1] === 'infinite',
+        })
         invalidateItem(getAssetsRetrieveQueryKey(uidAsset))
       },
     },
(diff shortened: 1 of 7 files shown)


jsapp/js/api/mutation-defaults/manage-projects-and-library-content.ts
```diff
@@ -0,0 +1,28 @@
...
+
+queryClient.setMutationDefaults(
+  getAssetsPartialUpdateMutationOptions().mutationKey!,
+  getAssetsPartialUpdateMutationOptions({
+    mutation: {
+      onSettled: (_data, _error, { uidAsset }) => {
+        const assetsListQueryKey = getAssetsListQueryKey()
+
+        invalidatePaginatedList(assetsListQueryKey)
+        // Invaliate infinite queries
```
pauloamorimbr: typo

```suggestion
        // Invalidate infinite queries
```

jsapp/js/api/mutation-defaults/survey-data.ts
```diff
@@ -378,3 +380,16 @@ queryClient.setMutationDefaults(
...
 )
+
+queryClient.setMutationDefaults(
+  getAssetsPairedDataDestroyMutationOptions().mutationKey!,
+  getAssetsPairedDataDestroyMutationOptions({
+    mutation: {
+      onSettled: (_data, _error, { uidAsset }) => {
+        const queryKey = getAssetsPairedDataListQueryKey(uidAsset)
+        invalidateItem(queryKey)
+        invalidatePaginatedList(queryKey)
```
pauloamorimbr: I may be missing context here, but I think we only need to call the `invalidatePaginatedList` here.

I see other sets of defaults up in this file using it like that, and it makes sense. These two functions are targeting the same queries... i think!


```suggestion
        invalidatePaginatedList(queryKey)
```

jsapp/js/api/mutation-defaults/manage-projects-and-library-content.ts
```diff
@@ -0,0 +1,28 @@
...
+        queryClient.invalidateQueries({
+          predicate: ({ queryKey }) =>
+            queryKey.length > assetsListQueryKey.length &&
+            assetsListQueryKey.every((keyPart, index) => queryKey[index] === keyPart) &&
+            queryKey[queryKey.length - 1] === 'infinite',
```
pauloamorimbr: nothing really important, but you can use `.at(-1)` to get the last item

```suggestion
            queryKey.at(-1) === 'infinite',
```

pauloamorimbr (approved):
Left a few nitpick comments for your considertation, but none blocking.
LGTM! :clap:

jsapp/js/api/mutation-defaults/manage-projects-and-library-content.ts
```diff
@@ -0,0 +1,28 @@
...
+        })
```
Akuukis: I suggest to create a new helper `invalidateInfiniteList` like the other helpers - it will surely be reused in future. Furthermore, it would be good to document/standardize the `infinite` constant used here, either as jsdoc on this helper or in a more appropriate place.

jsapp/js/api/mutation-defaults/survey-data.ts
magicznyleszek: I think you're right. I left `incalidateItem` as it is more broad

jsapp/js/api/mutation-defaults/manage-projects-and-library-content.ts
magicznyleszek: I've created `invalidateInfiniteList` helper and introduced a const to be used instead of `'infinite'` literal string :+1:

PR #7000, opened 2026-05-05
fix(nlp): update v1 speech recognition model to 'latest_long' DEV-2086

tinok:
### 📣 Summary

Improved automatic transcription quality by specifying the correct speech recognition model (v1 STT).

### 📖 Description

Audio transcriptions now use Google's `latest_long` model instead of the unspecified default, which produced lower quality results.

### 💭 Notes

The `RecognitionConfig` did not specify a `model`, causing Google to use a legacy default. 

### 👀 Preview steps

1. Have an account with a project containing an audio question
2. Submit an audio response (ideally conversational speech, 30-60s)
3. Trigger automatic transcription
4. 🔴 [on main] transcription is severely truncated/inaccurate
5. 🟢 [on PR] transcription captures full content of the audio


noliveleger (approved):
LGTM

PR #7016, opened 2026-05-05
ci(packages): Schedule weekly deletion of kpi images older than 90 days INFRA-431

olive-KTB:
### 💭 Notes

This cleans up images older than 90 days in ghcr, it is scheduled to be cleaned Sundays at 00:00. 

[Here](https://github.com/kobotoolbox/kpi/actions/runs/25375137643) is the dry run of images to be deleted. The deletion cuts off at [1c1524649e57db7e2e3b8a939f0936044bcf2cb3](https://github.com/orgs/kobotoolbox/packages/container/kpi/670229637?tag=1c1524649e57db7e2e3b8a939f0936044bcf2cb3)[1c1524649e57db7e2e3b8a939f0936044bcf2cb3](https://github.com/orgs/kobotoolbox/packages/container/kpi/670229637?tag=1c1524649e57db7e2e3b8a939f0936044bcf2cb3)


.github/workflows/nonprod-releases.yml
```diff
@@ -84,7 +84,7 @@ jobs:
         id: app-token
         uses: actions/create-github-app-token@v3
         with:
-          app-id: ${{ secrets.KOBO_BOT_APP_ID }}
+          client-id: ${{ secrets.KOBO_BOT_APP_ID }}
```
ScottMillard: In order to maintain backwards compatibility a new secret for client_id should be added so there is not confusion in the future which value this is.

.github/workflows/ghcr-cleanup.yml
```diff
@@ -0,0 +1,28 @@
...
+jobs:
+  cleanup-old-images:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Generate GitHub App Token
+        id: app-token
+        uses: actions/create-github-app-token@v3
+        with:
+          client-id: ${{ secrets.KOBO_BOT_APP_ID }}
+          private-key: ${{ secrets.KOBO_BOT_PRIVATE_KEY }}
```
ScottMillard, line 17: In order to maintain backwards compatibility a new secret for client_id should be added so there is not confusion in the future which value this is.

.github/workflows/ghcr-cleanup.yml
```diff
@@ -0,0 +1,28 @@
...
+          owner: kobotoolbox
+          repositories: "kpi"
+
+      - uses: snok/container-retention-policy@v3.0.0
+        with:
+          account: kobotoolbox
+          token: ${{ steps.app-token.outputs.token }}
+          image-names: kpi
+          image-tags: "!main !public-beta !feature-*"
```
ScottMillard: I don't believe we have images on GHCR with these kinds of tags so this may not be needed

ScottMillard (changes requested):

.github/workflows/ghcr-cleanup.yml
olive-KTB: We have images with those tags in ghcr like [this](https://github.com/kobotoolbox/kpi/pkgs/container/kpi/757579876?tag=feature-tri)

.github/workflows/nonprod-releases.yml
olive-KTB: I think you created KOBO_BOT_CLIENT_ID already so I'll use instead of creating a new one

.github/workflows/ghcr-cleanup.yml
olive-KTB: I think you created KOBO_BOT_CLIENT_ID already so I'll use instead of creating a new one

.github/workflows/ghcr-cleanup.yml
ScottMillard: I don't believe we are using tags like that any longer https://github.com/kobotoolbox/kpi/blob/main/.github/workflows/nonprod-releases.yml#L76

olive-KTB: I have removed the filter

ScottMillard (approved):

PR #7017, opened 2026-05-05
fix(settings): apply AWS settings without explicit credentials DEV-2002

noliveleger:
### 💭 Notes

Related: https://github.com/kobotoolbox/kobo-install/pull/274

**S3 mock in tests (`test(storage)` commit)**

A `mock_s3_when_needed` session-scoped fixture is added to the root `conftest.py` (moved from `kobo/conftest.py`). When the default storage backend is `S3Boto3Storage`, it activates moto's `mock_aws` context for the entire test session and pre-creates the configured bucket. When `InMemoryStorage` is used, the fixture is a no-op. `moto[s3]` is added to `dev_requirements`.

`testing.py` overrides bucket names to predictable values (`test-bucket`, region `us-east-1`) when S3 is configured, so the fixture can create them without a real AWS connection. The three Bedrock settings previously hard-coded to `''` in `testing.py` are removed — they are now always set from `base.py` with proper defaults.

`AttachmentApiTests` is wrapped in `@override_settings(KOBOCAT_DEFAULT_FILE_STORAGE=FileSystemStorage)` because `ffmpeg` (a subprocess) cannot follow moto-intercepted presigned URLs. `setUpClass` resets `default_storage._backend` to force reinitialisation with the overridden setting.

**Settings fix (`fix(settings)` commit)**

All AWS settings except the credentials themselves were gated behind `if env.str('AWS_ACCESS_KEY_ID', False)`. This meant Bedrock, SES, and S3 settings were silently skipped when credentials came from `~/.aws/credentials` or an instance profile. Now only `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` are conditionally set; the rest are always applied with their safe defaults.


rgraber (approved):
LGTM

PR #7019, opened 2026-05-06
refactor(frontend): drop react-debounced-input in favor of mantine based component DEV-2073

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 💭 Notes

Replaced the unmaintained `react-debounce-input` package with an internal `DebouncedTextInput` component, unblocking the React 19 / Mantine v9 migration.

Changes here:
- `react-debounce-input` does not support React 19 so we dropped it
- The new `DebouncedTextInput` wraps the existing `TextInput` (Mantine) and uses `useDebouncedCallback` from `@mantine/hooks`
- Debounce timing (defaults to 300 ms), immediate flush on `Enter`, and immediate flush on `blur` are all preserved
- Storybook stories include three `play`-based interaction tests covering: debounce delay, Enter flush, and blur flush
- Data Table header filter inputs now use default 300 ms debounce instead of previously using 750 ms. 300 seems to be a good standard, and 750 was a bit to slow

### 👀 Preview steps

1. ℹ️ Have an account with a project that has at least one text/number question and some submissions
2. Open Project → Data → Table
3. Type a few characters in a text column filter.
   - 🟢 Results update after ~300 ms of inactivity (debounce fires once, not on every keystroke)
5. Type again, then press **Enter**
   - 🟢 Results update immediately without waiting for the timer
6. Type again, then click elsewhere (blur the input)
   - 🟢 Results update immediately on blur
7. Open Storybook → **Design system / DebouncedTextInput**
   - 🟢 `Default`, `CustomTimeout`, and `NoImmediateFlush` stories each show a _"Last committed value"_ indicator that updates only when the debounce fires
   - 🟢 `TestDebounceDelay`, `TestEnterFlush`, and `TestBlurFlush` play tests all pass green


jsapp/js/components/common/DebouncedTextInput.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+export interface DebouncedTextInputProps extends Omit<TextInputProps, 'value' | 'onChange'> {
+  value?: string
+  onChange: (value: string) => void
+  debounceTimeout?: number
+  forceNotifyByEnter?: boolean
+  forceNotifyOnBlur?: boolean
+}
+
+const DEFAULT_DEBOUNCE_TIMEOUT = 750
```
duvld: We denounce all lot (if not all) our search text inputs, should they all have the same timeout? The NLP language selector is at 300 for example and that one is searching the database

jsapp/js/components/common/DebouncedTextInput.tsx
```diff
@@ -0,0 +1,84 @@
...
+    flushOnUnmount: true,
+  })
+
+  const onInputChange = useCallback(
+    (event: React.ChangeEvent<HTMLInputElement>) => {
+      const nextValue = event.target.value
+      setInputValue(nextValue)
+      debouncedOnChange(nextValue)
+    },
+    [debouncedOnChange],
```
duvld, line 58: When would `debouncedOnChange` actually change? `debouncedOnChange` doesn't have a dependency array and depends on the `onChange` prop, so if that changes, the whole component would re-render anyway right?

jsapp/js/components/common/DebouncedTextInput.stories.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+/** Typing fires onChange only after the debounce delay, not on every keystroke. */
+export const TestDebounceDelay: Story = {
+  args: {
+    label: 'Debounce delay test',
+    placeholder: 'Type to search…',
+    debounceTimeout: 200,
+    onChange: fn(),
+  },
+  play: async ({ args, canvasElement }) => {
```
duvld, line 109: Why do we do this choreographed delay + changes for the story instead of using the component?

duvld (commented):
Some questions, but the functionality is LGTM

jsapp/js/components/common/DebouncedTextInput.tsx
magicznyleszek: I will change this to 300. Also madea Linear task to unify those times: https://linear.app/kobotoolbox/issue/DEV-2110/unify-debounce-times-used-on-frontend

jsapp/js/components/common/DebouncedTextInput.stories.tsx
magicznyleszek: I don't understand the question. This story here confirms that the debounce is working correctly, so after 30ms onChange should have not been called, because debounce time is 200. After 1000ms it surely should've been called.

jsapp/js/components/common/DebouncedTextInput.stories.tsx
```diff
@@ -0,0 +1,193 @@
...
+  args: {
+    label: 'Disabled',
+    placeholder: 'Not editable',
+    disabled: true,
+    value: 'pre-filled value',
+  },
+}
+
+/** Shows the last debounced value committed to a parent component. */
+export const WithCommittedValueDisplay: Story = {
```
magicznyleszek: This is a leftover before I introduced `WithCommittedValue`, will delete it

jsapp/js/components/common/DebouncedTextInput.tsx
magicznyleszek: `debouncedOnChange` can change when `debounceTimeout` prop changes. Unlikely, but better safe than sorry :P

jsapp/js/components/common/DebouncedTextInput.stories.tsx
duvld: I'm not sure what I meant here, but the storybook does work as intended, marking as resolved

jsapp/js/components/common/DebouncedTextInput.tsx
duvld: Looks like greptile found an inconsistency with this change, but thanks for making the linear issue!

jsapp/js/components/common/DebouncedTextInput.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+export interface DebouncedTextInputProps extends Omit<TextInputProps, 'value' | 'onChange'> {
+  value?: string
+  onChange: (value: string) => void
+  debounceTimeout?: number
+  forceNotifyByEnter?: boolean
+  forceNotifyOnBlur?: boolean
+}
+
+const DEFAULT_DEBOUNCE_TIMEOUT = 300
```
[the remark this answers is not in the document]
magicznyleszek: I've updated the description to now mean 300, which is the amount I've ultimately stuck with

jsapp/js/components/submissions/table.tsx
```diff
@@ -1037,12 +1037,11 @@ export class DataTable extends React.Component<DataTableProps, DataTableState> {
...
         col.Filter = ({ filter, onChange }) => (
-          <DebounceInput
+          <DebouncedTextInput
             value={filter ? filter.value : undefined}
-            debounceTimeout={750}
-            onChange={(event) => onChange(event.target.value)}
-            className='table-filter-input'
+            onChange={onChange}
             placeholder={t('Search')}
+            size='xs'
```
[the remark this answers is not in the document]
magicznyleszek: Updated PR description

jsapp/js/components/common/DebouncedTextInput.tsx
```diff
@@ -0,0 +1,84 @@
...
+    forceNotifyByEnter = true,
+    forceNotifyOnBlur = true,
+    ...restProps
+  } = props
+
+  const [inputValue, setInputValue] = useState(value ?? '')
+
+  useEffect(() => {
+    setInputValue(value ?? '')
+  }, [value])
```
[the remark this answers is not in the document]
magicznyleszek: Fixed

magicznyleszek:

duvld (approved):
LGTM preview works and tests pass

PR #7020, opened 2026-05-06
fix(formBuilder): null language on multilingual XLSForm save DEV-2095

noliveleger:
### 📣 Summary

Opening the Form Builder fails with a "language needs to be configured" error after saving a multilingual form (one with labels in multiple languages and a `default_language` setting).

### 📖 Description

When a multilingual XLSForm (with labels in two or more languages and a `default_language` set) is uploaded and then edited in the Form Builder, saving the form once corrupts the language configuration. On the next attempt to open the Form Builder, an error appears indicating a language needs to be configured. Visiting "Manage Languages" shows an unknown blank language has been created, and the form becomes uneditable.

### 💭 Notes

Root cause — frontend (unnullifyTranslations in formBuilderUtils.ts)

When the Form Builder loads, inputParser.parse() mutates its input object to add translations_0 (the default language name), nullify translations[0], and normalize settings from an array to an object. Survey.loadDict was passing state.asset.content directly, so those mutations landed on the live state.asset.content object. React Query's background re-fetch then replaced state.asset with a fresh API response, losing those mutations. launchAppForSurveyContent early-returns because app is already initialised, so the mutations never run again. At save time, unnullifyTranslations was reading state.asset.content, found translations_0 = undefined, set defaultLang = null, and skipped the rename — plain label keys reached the backend.

**Fix — frontend**

Pass clonedeep(assetContent) to Survey.loadDict so inputParser.parse() mutates only the copy. state.asset.content stays as the raw API object; any React Query re-fetch replaces it with identical data and no mutations are lost.

At both save paths (previewForm and saveForm), pass app.survey._initialParams to unnullifyTranslations instead of state.asset.content. _initialParams is assigned at Survey construction time directly from the already-mutated parse result and is stable for the lifetime of the Survey instance.

Guard both call sites on app.survey._initialParams?.translations_0 rather than state.asset?.content, so the rename only runs when a default language is known — correctly skipping new forms and monolingual forms.

**Fix — backend defensive (`__normalize_settings_and_translations` in `asset.py`)**

`expand_content_in_place` already flags a plain `label` column as `UNTRANSLATED` and appends `None` to the translations list, producing a spurious unnamed language. The method already converts plain `hint` strings to `hint::default_lang`; the same is now done for `label`, but only when `len(existing_translations) > 1` to leave single-language forms (where a plain `label` is intentional) unchanged. Choices rows are also covered.

### 👀 Preview steps

1. ℹ️ Create an XLSForm with two language columns (`label::English (en)`, `label::French (fr)`) and `default_language: English (en)` in the settings sheet.
2. Upload it as a new project (or replace an existing form).
3. Open the Form Builder — it loads without error.
4. Modify any question label, then save and close.
5. 🔴 [on main] Re-opening the Form Builder shows an error about an unnamed language; "Manage Languages" reveals a blank/unknown language entry.
6. 🟢 [on PR] Re-opening the Form Builder works correctly; both original languages are present.

### DIFF
--- a/jsapp/js/components/formBuilder/formBuilderUtils.ts
+++ b/jsapp/js/components/formBuilder/formBuilderUtils.ts
@@ -46,7 +46,12 @@ export function unnullifyTranslations(surveyDataJSON: string, assetContent: Asse
   // set default_language
   let defaultLang = assetContent.translations_0
   if (!defaultLang) {
-    defaultLang = null
+    // Fallback: after a background re-fetch, state.asset.content is replaced with a
+    // fresh API object that doesn't have translations_0 (a frontend-only mutation).
+    // launchAppForSurveyContent early-returns in that case (app already initialized),
+    // so the mutation never runs again. Derive the default language from
+    // settings.default_language to avoid sending plain `label` keys to the backend.
+    defaultLang = assetContent.settings?.default_language || null
   }
   if (!surveyData.settings[0].default_language && defaultLang !== null) {
     surveyData.settings[0].default_language = defaultLang
--- a/jsapp/js/components/formBuilder/formBuilderUtils.ts
+++ b/jsapp/js/components/formBuilder/formBuilderUtils.ts
@@ -32,27 +32,19 @@ export function surveyToValidJson(survey: Survey) {
 /**
  * This function reverses what `nullifyTranslations` did to the form data.
  */
-export function unnullifyTranslations(surveyDataJSON: string, assetContent: AssetContent) {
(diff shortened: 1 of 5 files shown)


RuthTurk (approved):
Backend code lgtm

pauloamorimbr (approved):
LGTM! 👏🏻

tiritea:
FWIW we're getting a lot of traffic on the Forum from users about translation issues (I assume probably stemming this bug?) so it would be probably be good to get this deployed as soon as possible to see if it clears up most of their problems.

PR #7030, opened 2026-05-07
fix(transcription): submission data not updating on navigation DEV-2097

pauloamorimbr:
### 📣 Summary
Fixes an issue with using arrow navigation to switch to other submissions in audio processing view would not fully update submission data, causing possible data loss. Also, adds confirmation message when there is unsaved data in the view and a user tries to navigate to another question or submission.

### 💭 Notes
- A key (used submission._uuid) is needed to be given to components so they know when to update, or else react will likely retain the current state.
- `hasUnsavedData` was integrated into `SelectQuestion` and `SelectSubmission` to protect against navigation and data loss when unsaved data is present.

### 👀 Preview steps
1. ℹ️ have an account and a project with audio question data
2. enter the audio processing view
3. create an AUTOMATIC transcription
4. dont save nor discard, leaving the transcription in the validation state
5. use the top-right arrows to navigate to another submission
4. 🔴 [on main] notice that the submission data is not updated and, if the save button is pressed, data is saved in the current submission
5. 🟢 [on PR] notice that the submission data updates, displaying the correct state and data for the current submission
6. start a MANUAL transcription and fill in some text. Do not save nor discard
6. select another question from the top-left question selection
4. 🔴 [on main] notice that navigation occurs and unsaved data is lost
5. 🟢 [on PR] notice that a confirmation message appears to avoid data loss
6. start a MANUAL transcription and fill in some text. Do not save nor discard
6. press the arrow buttons on top right to navigate to another submission
4. 🔴 [on main] notice that navigation occurs and unsaved data is lost
5. 🟢 [on PR] notice that a confirmation message appears to avoid data loss

### DIFF
--- a/jsapp/js/components/processing/SingleProcessingContent/TabTranscript/index.tsx
+++ b/jsapp/js/components/processing/SingleProcessingContent/TabTranscript/index.tsx
@@ -34,7 +34,7 @@ export default function TranscriptTab({
 }: Props) {
   const transcriptVersion = getLatestTranscriptVersionItem(supplement, questionXpath)
 
-  // We need a key to force remounting the component when switching between submissions,
+  // We need a key to force remounting the component when switching between submissions.
   const submissionKey = submission._uuid
 
   if (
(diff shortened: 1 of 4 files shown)


jamesrkiger (approved):
LGTM. I tested the repro steps and the fix works. I'd like to test more when we resolve DEV-2104, but let's get this merged for now.

PR #7034, opened 2026-05-08
fix(map): blank map for questions in group fix DEV-2093

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally
3. [x] for user-facing doc changes create a Zulip thread at #Support Docs Updates, if any
4. [x] draft PR with a title <type>(<scope>)<!>: <title> DEV-1234
5. [x] assign yourself, tag PR: at least Front end and/or Back end or workflow

### 📣 Summary

Fixed the Project → Data → Map so geopoint answers inside groups are now shown correctly instead of incorrectly showing "No geopoint responses have been received".

### 💭 Notes

Changes here:
- Map component: index.tsx
  - Normalized selected geopoint question names to their flat-path field keys before building the submissions query.
  - Reused the same normalized key when parsing coordinates, so grouped geopoint values are read from the correct submission field.
  - Added a safe fallback in grouped-field path resolution so non-grouped names continue to work as expected.

#7041 fixes this bug on an older code

### 👀 Preview steps

1. ℹ️ Have an account with access to a deployed project.
2. Create or open a form that has one geopoint question inside a group.
3. Submit at least one response with a valid location for that grouped geopoint question.
4. Open the project and go to Data → Map.
5. 🔴 [on main] Notice the map may say "No geopoint responses have been received".
6. 🟢 [on PR] Notice the location point(s) are displayed on the map.
7. Optional sanity check: open a form with a top-level geopoint (not in a group) and confirm it still displays correctly.

### DIFF
--- a/jsapp/js/components/map/index.tsx
+++ b/jsapp/js/components/map/index.tsx
@@ -481,7 +481,7 @@ class FormMap extends React.Component<FormMapProps, FormMapState> {
   updateWrapperQuery(selectedQuestion: string | null, nextViewBy: string, pageLimit: number) {
     const fq = ['_id']
     if (selectedQuestion) {
-      fq.push(selectedQuestion)
+      fq.push(this.nameOfFieldInGroup(selectedQuestion))
     }
     if (nextViewBy) {
       fq.push(this.nameOfFieldInGroup(nextViewBy))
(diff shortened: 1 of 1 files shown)


duvld (approved):
LGTM

PR #7039, opened 2026-05-11
refactor(languageSelector): update and replace old language selector DEV-1401

duvld:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Update old language selector into a new component

### 💭 Notes
Replaced the old language selector with a new one. This one uses react-query and orval generated types so it has no more reliance on the `languageStore`. Removing this store is beyond the scope of replacing the language selector, but the selector is the main user of that store.

There are some intended "regressions" because of the switching from a complex component to a simple dropdown. The new language selector uses Mantine `Select` and is meant to be used in multiple places. See DEV-1401 for more info. But here are some notable changes:
1. No more infinite scroll (`Select` does allow for this in the future, if needed)
2. The display of the languages is all contained in the dropdown
3. The dropdown contains groups
    - There is a new "recently selected" group
4. Storybook is updated
    - The specific "Select 'Sweedish' language" story wasn't working, so it wasn't replicated. All props are available to manipulate 
5. The old language selector advertised a "From source" feature. I couldn't find anywhere that was using this, no prop was ever directly populated, so I didn't recreate this feature (I don't know what it was doing anyways).
6. The new component is a controlled component, I removed the old global event handling to navigate the UX

All functional capabilities of the old language selector is ported over

### 👀 Preview steps

1. ℹ️ have an account and a project with an audio response so you can open the NLP view
7. Navigate to the NLP view
9. 🟢 notice that the functionality is the same as before
    - This means a lot of things, it will give a better idea of how the props are interpreted by the new component if you look at the new storybook implementation
    - In the NLP view, notice that all the buttons surrounding this component still work as intended (back button, manual, automatic), because the props were written in a very decoupled way, the NLP steps seems to work seamlessly 
11. 🟢 notice that the functionality and styles match the mockups


jsapp/js/components/languages/languageSelector.stories.tsx
```diff
@@ -17,71 +50,15 @@ const meta: Meta<typeof LanguageSelector> = {
...
-      await userEvent.click(clearSearchButton)
-      await waitFor(async () => {
-        const noSearchItem2 = await canvas.findByText(/^Arabic/)
-        await expect(noSearchItem2).toBeInTheDocument()
-      })
-    })
+    required: true,
+    titleOverride: 'Select a language',
+    suggestedLanguages: ['ar', 'bn'],
+    hiddenLanguages: ['', ''],
```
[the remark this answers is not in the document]
duvld: I have it set up this way as I can't seem to get a better way to show the storybook viewer that we should add items the same was as the `suggestedItems`. Doing it the above way results in this:
image

And adding `"en"`, `en`, `0: "en"` doesn't get saved

jsapp/js/components/languages/languageSelector.tsx
```diff
@@ -1,414 +1,158 @@
-import './languageSelector.scss'
```
magicznyleszek, line 1: Given we are touching the files, could we rename this and stories file to PascalCase?

jsapp/js/components/processing/SingleProcessingContent/components/StepSelectLanguage.tsx
```diff
@@ -90,6 +90,7 @@ export default function StepSelectLanguage({
         onLanguageChange={handleChangeLanguage}
         hiddenLanguages={hiddenLanguages}
         suggestedLanguages={suggestedLanguages}
+        value={languageCode}
```
magicznyleszek, line 93: I notices on short screen that it goes outside (the top of) the screen:

Image

jsapp/js/components/languages/languageSelector.stories.tsx
```diff
@@ -1,13 +1,49 @@
...
 const meta: Meta<typeof LanguageSelector> = {
   title: 'Components/LanguageSelector',
   component: LanguageSelector,
-  argTypes: {},
+  decorators: [
+    (Story) => (
+      <QueryClientProvider client={queryClient}>
+        <Story />
+      </QueryClientProvider>
+    ),
```
magicznyleszek: I think we can use `queryClientDecorator` from `jsapp/js/query/queryClient.mocks.tsx` instead of this custom one

jsapp/js/components/languages/LanguageSelector.stories.tsx
```diff
@@ -17,70 +53,108 @@ const meta: Meta<typeof LanguageSelector> = {
...
+    suggestedLanguages: ['ar', 'bn'],
+    hiddenLanguages: ['', ''],
+  },
+}
 
-/**
- * This test is searching for "Swedish" language, selecting it, and then undoing both the selection and the search.
- */
-export const TestSearchAndSelection: Story = {
+export const InteractionTest: Story = {
```
magicznyleszek, line 60: Let's bring the comment back and update it to mention all the things test does.

jsapp/js/components/languages/languageSelector.tsx
```diff
@@ -1,414 +1,158 @@
...
-   * not selectable from the list.
-   */
-  sourceLanguage?: LanguageCode
   /**
-   * A list of languages that should be displayed in front of other languages.
-   * Most possibly these languages were already chosen for other parts of given
-   * feature or can be found in existing data.
+   * TODO: This should be typed as (language: ListLanguage | null) => void but there's a type incompatibility
+   * between what StepSelectLanguage expects and what the orval-generated API types are.
+   * Specifically, `transcription_services` and `translation_services` arrays are readonly in new types.
```
magicznyleszek: Do we know which types in particular are mismatched? Could you create a Linear task for the fix in the API docs? And then mention it here via linear id?

jsapp/js/components/languages/languageSelector.tsx
```diff
@@ -1,414 +1,158 @@
...
+  onLanguageChange: (language: any) => void
+  titleOverride?: string
+  /** The following props are from the old languageSelector, and is useful to maintain it's UX*/
   hiddenLanguages?: LanguageCode[]
-  /** Triggered after language is selected or cleared. */
-  onLanguageChange: (selectedLanguage: DetailedLanguage | ListLanguage | null) => void
+  suggestedLanguages?: LanguageCode[]
+  /** The currently selected language code, used for the UX of single processing view */
+  value?: LanguageCode | null
   isDisabled?: boolean
```
magicznyleszek: Let's rename it to `disabled`, as this seems to be the standard in Mantine etc.

jsapp/js/components/languages/languageSelector.tsx
```diff
@@ -1,414 +1,158 @@
...
-   * feature or can be found in existing data.
+   * TODO: This should be typed as (language: ListLanguage | null) => void but there's a type incompatibility
+   * between what StepSelectLanguage expects and what the orval-generated API types are.
+   * Specifically, `transcription_services` and `translation_services` arrays are readonly in new types.
    */
-  suggestedLanguages?: LanguageCode[]
-  /** A list of languages that should be omitted from display. */
+  onLanguageChange: (language: any) => void
+  titleOverride?: string
+  /** The following props are from the old languageSelector, and is useful to maintain it's UX*/
```
magicznyleszek: Don't we want to maintain those in new LanguageSelector as well? I mean this comment sounds like "this is old legacy thing, but we keep it", and AFAIU this is something we do want to still have in the component, right?

jsapp/js/components/languages/languageSelector.tsx
```diff
@@ -1,414 +1,158 @@
...
   hiddenLanguages?: LanguageCode[]
-  /** Triggered after language is selected or cleared. */
-  onLanguageChange: (selectedLanguage: DetailedLanguage | ListLanguage | null) => void
+  suggestedLanguages?: LanguageCode[]
+  /** The currently selected language code, used for the UX of single processing view */
```
magicznyleszek: But what does it do actually? Does it pre-select the language when component is first rendered? Let's comment on what it does rather than mention the single processing - as maybe we would want/need to use it in other places in UI too

jsapp/js/components/languages/LanguageSelector.tsx
```diff
@@ -1,414 +1,158 @@
...
+  const [searchValue, setSearchValue] = useState('')
+  const [debouncedSearch] = useDebouncedValue(searchValue, SEARCH_DEBOUNCE_MS)
+  // Keep UI grouping in sync with what user is typing right now.
+  // This avoids a stale "Results" group after clearing the input.
+  const isSearching = searchValue.length >= MINIMUM_SEARCH_LENGTH
+  // Keep API calls debounced to avoid network churn while typing.
+  const isDebouncedSearching = debouncedSearch.length >= MINIMUM_SEARCH_LENGTH
+
+  // "Main" language list (either featured languages or result of searching)
+  const { data, isLoading } = useLanguagesList(isDebouncedSearching ? ({ q: debouncedSearch } as any) : undefined)
```
magicznyleszek, line 40: The `as any` part is needed because orval types are imperfect, right? Let's create Linear task and mention it

jsapp/js/components/languages/languageSelector.tsx
```diff
@@ -1,414 +1,158 @@
...
+  // Keep API calls debounced to avoid network churn while typing.
+  const isDebouncedSearching = debouncedSearch.length >= MINIMUM_SEARCH_LENGTH
+
+  // "Main" language list (either featured languages or result of searching)
+  const { data, isLoading } = useLanguagesList(isDebouncedSearching ? ({ q: debouncedSearch } as any) : undefined)
+  const languages = data?.status === 200 ? data.data.results : []
+
+  // Create a query for each suggested language
+  // NOTE: Suggested languages are broken but handled by parent. Once fixed, the parent will pass correct props (this
+  // component's logic is already ready)
```
magicznyleszek: Broken how? And "handled by parent" means "parent gives us a list of codes as a fallback"?

jsapp/js/components/languages/languageSelector.tsx
```diff
@@ -1,414 +1,158 @@
...
-            loader={<LoadingSpinner message={false} key='loadingspinner' />}
-            useWindow={false}
+  return (
+    <Select
+      searchValue={searchValue}
+      onSearchChange={setSearchValue}
+      data={languageOptions}
+      onChange={onLanguageSelected}
+      value={props.value ?? null}
+      label={props.titleOverride}
```
magicznyleszek: Would we ever use anything else than `t('Select a language')`? I.e. do we need a prop for overriding?

jsapp/js/components/languages/languageSelector.tsx
```diff
@@ -1,414 +1,158 @@
...
-        {!this.state.selectedLanguage && this.renderSearchForm()}
-      </bem.LanguageSelector>
-    )
-  }
+            <KoboIcon icon={IconInfoCircleFilled} size='sm' />
+            <Text>{t('No matching languages found. Try another spelling or language name')}</Text>
+          </Group>
+        )
+      }
+      renderOption={(option) => <Text>{option.option.label}</Text>}
```
magicznyleszek: Could we rename `(option)` to something else to avoid that funny `option.option`?

jsapp/js/components/processing/SingleProcessingContent/components/StepSelectLanguage.tsx
```diff
@@ -60,10 +60,10 @@ export default function StepSelectLanguage({
   }
 
   const handleClickBack = () => {
-    // When clicking "back" we either unselect the language (inner component "back" action) through a special component
-    // function , or if no language is selected, we let the parent know (the parent flow "back" action).
+    // When clicking "back" we either unselect the language (inner component "back" action) by clearing it,
+    // or if no language is selected, we let the parent know (the parent flow "back" action).
     if (languageCode !== null) {
-      resetAllLanguageSelectors()
+      setLanguageCode(null)
```
magicznyleszek, line 66: We no longer need to reset the selector with that funky `resetAllLanguageSelectors`, because it reset itself due `value` being passed, right?

magicznyleszek (changes requested):

jsapp/js/components/languages/languageSelector.tsx
duvld: I think I had some other idea for what this component was doing when I wrote it, but I don't think it makes sense anymore. Removed the comment

jsapp/js/components/languages/languageSelector.tsx
duvld: This is fixed by #7059 , so I'll just remove this comment

jsapp/js/components/languages/languageSelector.tsx
duvld: Probably not, but it was set up this way in the old component. I'll leave the option there, but set the default to `t('Select a language')`

jsapp/js/components/processing/SingleProcessingContent/components/StepSelectLanguage.tsx
duvld: I'm not sure what you mean, the code removes `resetAllLanguageSelectors`? Setting the language to null this way is the 'controlled component' solution to clearing the Select component

duvld: I've removed the `TODO HACKFIX` comment in case you were referring to that :P

jsapp/js/components/languages/LanguageSelector.tsx
duvld: Added task and updated comment

jsapp/js/components/languages/languageSelector.tsx
duvld: ~~Added task and updated comment~~ Used `LanguageCode` instead here

magicznyleszek (approved):

PR #7040, opened 2026-05-11
feat(frontend): create a reflux to react-query (orval) bridge DEV-2107

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally
3. [x] for user-facing doc changes create a Zulip thread at #Support Docs Updates, if any
4. [x] draft PR with a title <type>(<scope>)<!>: <title> DEV-1234
5. [x] assign yourself, tag PR: at least Front end and/or Back end or workflow
8. [x] open PR and confirm that CI passes and request reviewers, if needed

### 💭 Notes

Adds a bridge between new react-query code and old Reflux code by ensuring Reflux listeners are still notified when new code performs API calls. This lets us migrate incrementally without breaking UI behavior during the transition.

For example, if a migrated flow previously depended on removeAssetPermission.completed, we have two options:
1. Migrate only that flow and let the bridge notify legacy listeners.
2. Migrate that flow plus every dependent Reflux listener path immediately, which can force deeper migration work than planned.

Changes here:
1. `orval.mutator.ts` calls bridge dispatchers at three lifecycle points, so compatibility stays centralized.
2. `index.ts` remains the dispatch entrypoint and has a clear top-level explanation of bridge behavior.
3. `shared.ts` contains shared parsing and normalization helpers plus endpoint pre-match logic.
4. Endpoint handler matching:
   1. Handlers configs use endpoint as the source of truth for method and path shape.
   2. Separate method fields were removed from handler tables to reduce drift.
   3. `doesEndpointMatchRequest` is used as a pre-check before handler-specific matches predicates.
   4. `refluxAction` is kept as a diagnostic label (not a matcher), so logs remain easy to read and trace.
   5. `start-handlers.ts` keeps request-time started callbacks.
   6. `success-handlers.ts` handles success callbacks, including split PATCH deployment behavior for redeploy vs set-active.
   7. `failure-handlers.ts` mirrors the same endpoint family for failed callbacks.
   8. `bridge-handlers.tests.ts` includes handler-flow coverage plus endpoint matcher tests.
5. `jsapp/js/library/managedCollectionsStore.ts`, `jsapp/js/library/myLibraryStore.ts`, `jsapp/js/library/publicCollectionsStore.ts`, and `jsapp/js/library/singleCollectionStore.ts` remove dependence on `assetType` in delete completion handlers to simplify their code, and make life easier for handlers.
6. Improved some types in `actions.d.ts`

I also added and adjusted comments to reduce confusion for future maintainers.

### 👀 Preview steps

This is a prerequisite for upcoming work that migrates Manage Languages to Mantine plus React Query.

### DIFF
--- a/jsapp/js/api/reflux-bridge/bridge-routes.tests.ts
+++ b/jsapp/js/api/reflux-bridge/bridge-routes.tests.ts
@@ -34,6 +34,7 @@ import {
   bridgeOrvalStartToLegacyActions,
   bridgeOrvalSuccessToLegacyActions,
 } from './index'
+import { doesEndpointMatchRequest } from './shared'
 
 describe('reflux bridge route flow', () => {
   beforeEach(() => {
@@ -105,5 +106,25 @@ describe('reflux bridge route flow', () => {
     // Make sure this request did not spill into report-related routes.
     chai.expect(reportActions.setStyle.completed.mock.calls.length).to.equal(0)
     chai.expect(reportActions.setCustom.completed.mock.calls.length).to.equal(0)
+  })
+})
+
+describe('endpoint pattern matching', () => {
+  it('matches dynamic asset endpoint pattern', () => {
+    chai
+      .expect(doesEndpointMatchRequest('PATCH /api/v2/assets/:uid/', 'PATCH', '/api/v2/assets/abc123/'))
+      .to.equal(true)
+  })
+
+  it('does not match a different path shape', () => {
+    chai
+      .expect(doesEndpointMatchRequest('PATCH /api/v2/assets/:uid/', 'PATCH', '/api/v2/assets/abc123/deployment/'))
+      .to.equal(false)
+  })
+
+  it('does not match when method differs', () => {
+    chai
+      .expect(doesEndpointMatchRequest('PATCH /api/v2/assets/:uid/', 'POST', '/api/v2/assets/abc123/'))
+      .to.equal(false)
   })
 })
--- a/jsapp/js/api/reflux-bridge/bridge-routes.tests.ts
+++ b/jsapp/js/api/reflux-bridge/bridge-routes.tests.ts
@@ -1,14 +1,13 @@
 import chai from 'chai'
 
-// The bridge keeps a few legacy Reflux actions alive while write paths move
-// over to Orval/react-query. That lets old UI code keep reacting to API work
-// without forcing the whole migration to happen in one step.
+// The bridge keeps a few legacy Reflux actions alive while write paths move over to Orval/react-query.
+// That lets old UI code keep reacting to API work without forcing the whole migration to happen in one step.
 
 var mockedActions: any
 
 jest.mock('#/actions', () => {
-  // Keep the mock small. Pulling in the real actions module would drag a lot
-  // of unrelated Reflux setup into a test that only cares about bridge routing.
+  // Keep the mock small. Pulling in the real actions module would drag unrelated Reflux setup
+  // into a test that only cares about bridge routing.
   mockedActions = {
     reports: {
       setStyle: { completed: jest.fn(), failed: jest.fn() },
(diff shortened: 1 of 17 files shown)


jsapp/js/api/reflux-bridge/bridge-routes.tests.ts
```diff
@@ -0,0 +1,130 @@
+import chai from 'chai'
+
+// The bridge keeps a few legacy Reflux actions alive while write paths move over to Orval/react-query.
+// That lets old UI code keep reacting to API work without forcing the whole migration to happen in one step.
+
+var mockedActions: any
```
jamesrkiger, line 3: Should probably just use let here, right?

jsapp/js/api/reflux-bridge/failure-routes.ts
```diff
@@ -0,0 +1,100 @@
...
+ */
+export const BRIDGE_FAILURE_ROUTES: ReadonlyArray<BridgeFailureRoute> = [
+  {
+    endpoint: 'PATCH /api/v2/assets/:uid/',
+    refluxAction: 'actions.resources.updateAsset.failed',
+    matches: ({ assetUid, requestBody }) =>
+      Boolean(
+        assetUid &&
+          requestBody &&
+          !('report_styles' in requestBody) &&
```
jamesrkiger: It took me a bit to figure out what is going on here, as this at first seems oddly specific. What would you think about something like:
```
enum SpecializedAssetPatchField {
  ReportStyles = 'report_styles',
  ReportCustom = 'report_custom',
  MapStyles = 'map_styles',
}
```

And then you could use the values below and here you could match on `!Object.values(SpecializedAssetPatchField).some((field) => field in requestBody)`. Doesn't change what's happening, but might make it easier to understand (and hypothetically update) for future devs.

jsapp/js/api/reflux-bridge/failure-routes.ts
```diff
@@ -0,0 +1,100 @@
+import { actions } from '#/actions'
+import type { BridgeFailureRoute } from './shared'
+
+/**
+ * Response-time failure routes. Keep this table focused on legacy `*.failed` callbacks.
+ */
+export const BRIDGE_FAILURE_ROUTES: ReadonlyArray<BridgeFailureRoute> = [
```
jamesrkiger: I'm wondering if 'route' is the best word choice here, since we're not only checking routes.

magicznyleszek: I agree, "route" is not best. I was wondering about using some other word, but didn't come up with something good. Any ideas? :)

jsapp/js/api/reflux-bridge/failure-routes.ts
magicznyleszek: Good idea, added it to `shared.ts` and using in different files too

jsapp/js/api/reflux-bridge/bridge-routes.tests.ts
magicznyleszek: Correct!

jsapp/js/api/reflux-bridge/failure-routes.ts
magicznyleszek: Maybe "handlers"?

jsapp/js/api/reflux-bridge/bridge-routes.tests.ts
magicznyleszek: Actually not correct. We need to use `var` because of how `jest.mock` works. I've added a comment in there :)

jsapp/js/api/reflux-bridge/failure-routes.ts
jamesrkiger: Handlers works for me.

jamesrkiger (approved):
LGTM, thanks for planning this one out, Leszek.

jamesrkiger (approved):
lgtm

pauloamorimbr (approved):
LGTM! 

Very nice idea and implementation.
We just need to have a plan somewhere to get rid of Reflux for once at some point.

PR #7043, opened 2026-05-11
feat(nlp): add models for subsequence bulk action and item tracking  DEV-1421

rajpatel24:
### 📣 Summary
This PR introduces the database tables required to support bulk transcription and translation jobs within the subsequences app.

### 📖 Description
This PR implements the models necessary for orchestrating and tracking bulk processing tasks.

- `SubsequenceBulkAction`: A parent model is created to represent a batch execution, storing deterministic parameters and global status.   

- `SubsequenceBulkActionItem`: A child model is implemented to track individual submission processing within a bulk job, including an external `service_id` for tracking Google operations.   

- Idempotency via Hashing: A deterministic SHA-256 hash of job parameters is utilized to prevent duplicate active jobs for the same submission and action.   

- Integrity Guarantees: Atomic creation logic is provided via `create_with_items`, and status propagation is implemented to synchronize parent and child states.   

- Database Constraints: A `UniqueConstraint` is enforced to ensure only one active job item exists per submission/action/hash combination.

### 👀 Preview steps

Please use the shell or rely on unit tests for testing.


rajpatel24:
@greptileai `BulkActionStatus` intentionally does not include a FAILED state. Even if individual child records (`SubsequenceBulkActionItem`) fail, the parent job itself should still be treated as COMPLETE, with failures reflected at the item level.

RuthTurk (approved):
LGTM

PR #7046, opened 2026-05-11
fix(subsequences): migrate qpath to xpath in SubmissionSupplements DEV-2094

rgraber:
### 🗒️ Checklist

1. [ ] run linter locally

### 📣 Summary
Ensure old NLP data and QA question responses get exported.

### 📖 Description
Fixes a bug wherein older transcripts, translations, and qa question responses were not showing up in exports or the data table.

### 💭 Notes
A previous migration from `qpath`s to `xpath`s was done on the fly, so submission supplements that were not accessed were not updated. However, the `advanced_features` dict of the corresponding assets were updated to have xpaths, and no reference to the old qpaths.

The subsequent migration to the new subsequence version (`20240820`) looked at old submission supplements
to gather `(xpath, action)` pairs to determine what QAFs needed to be created, then used the `advanced_features` dict on the asset to populate the `params` of the new QAF. This led to the creation of QAFs that had a qpath instead of an xpath in `question_xpath` and an empty `params` list.

This PR queries all submission supplements for anything with a hyphen in the question name using `extra` and `rawSQL`. This is because there is no Django-native way of querying for keys in a JSONField that match a particular regex. While this does mean using postgres specific code, it prevents us from having to use a whole new schema version for what is essentially an edge case. 

The PR also creates any missing QAFs that may have not been created if every submission supplement for an asset used the old qpath. For safety, the QAFs are created in an `atomic` block while updating the submission supplement that necessitated them, so if something kills the update mid-process, we will still know they need to be created. We use a simple set to keep track of which QAFs were already created so we don't hit the DB more than necessary.

There is one small risk where we could end up with the migration infinitely in progress:
If there is a submission supplement with a qpath that has no xpath equivalent (ie the question was removed), we just keep the qpath as it is. This means it will be picked up every time we run the initial query to find the supplements that need migrating. If the job keeps getting interrupted and never actually returns, it will be infinitely marked "in progress". However this seems pretty unlikely unless there are lots and lots of submission supplements with deleted questions. Given that there are not that many submission supplements on prod to begin with, the risk seems acceptable. 

### 👀 Preview steps
It's a pill to actually get old data to test this, but you can make mock data that reasonably acts like old data. Make sure to start on the release branch so all supplements will have the old version.

1. ℹ️ have an account and a project with an audio question inside a group
2. ℹ️ have at least one submission with a transcript
3. Add a QA question and answer (manual)
4. Open a django shell
5. 
```
asset = Asset.objects.get(uid='<uid>')
for question in asset.content['survey']:
        if '$xpath' in question:
                question['$qpath'] = question['$xpath'].replace('/','-')  # imitate old qpaths
asset.save()

ss = SubmissionSupplement.objects.get(submission_uuid='<uuid>', asset=asset)
ss_content_copy = copy.deepcopy(ss.content)
for key in [ k for k in ss.content.keys() if '/' in k]:
        ss_content_copy[key.replace('/','-')] = copy.deepcopy(ss_content_copy[key])
        del ss_content_copy[k]
ss.content = ss_content_copy
ss.save()
```
6. Export data with all columns
7. 🔴 [on main] notice that the answer to the QA question is missing for your submission
8. Run LRM 24
9. 🟢 [on PR] Should report 1 supplement with old qpaths was updated
10. Export data with all columns
11. 🟢 notice QA answer is now present

### DIFF
--- a/kobo/apps/long_running_migrations/jobs/0024_migrate_submission_supplement_qpaths.py
+++ b/kobo/apps/long_running_migrations/jobs/0024_migrate_submission_supplement_qpaths.py
@@ -59,7 +59,16 @@ def run(dry_run: bool = False):
                     continue
                 for action_id in actions_for_xpath.keys():
                     if (asset.uid, xpath, action_id) not in existing_qafs:
-                        params = build_params(asset, xpath, action_id)
+                        try:
+                            params = build_params(asset, xpath, action_id)
+                        except ValueError as ve:
+                            print(
+                                f'{logging_prefix} - failed to build params '
+                                f'for {xpath}, {action_id}: {ve}'
+                            )
+                            existing_qafs.add((asset.uid, xpath, action_id))
+                            continue
+
                         if not dry_run:
                             _, created = QuestionAdvancedFeature.objects.get_or_create(
                                 asset=asset,
--- a/kobo/apps/long_running_migrations/jobs/0024_migrate_submission_supplement_qpaths.py
+++ b/kobo/apps/long_running_migrations/jobs/0024_migrate_submission_supplement_qpaths.py
@@ -57,7 +57,7 @@ def run(dry_run: bool = False):
         )
         with transaction.atomic():
             for xpath, actions_for_xpath in new_content.items():
-                if xpath == '_version':
+                if not isinstance(actions_for_xpath, dict):
                     continue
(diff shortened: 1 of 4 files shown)


RuthTurk (approved):
LGTM

PR #7049, opened 2026-05-12
feat(userreports): allow configurable page size for userreport API DEV-2112

rgraber:
### 🗒️ Checklist

1. [ ] run linter locally

### 📣 Summary
Set the max page size for `/api/v2/user-reports` in Constance.

### Notes
We don't want to make a habit of overriding the max page size, so right now the class factory only looks at a new USER_REPORTS_PAGE_SIZE_LIMIT configuration. If eventually we want to do this on more pages we can genericize the method.

### 👀 Preview steps

1. ℹ️ have several accounts (at least 10)
2. In a Django shell, run `refresh_user_report_snapshots`
3. In Constance, set `USER_REPORTS_PAGE_SIZE_LIMIT` to 5
4. Navigate to `/api/v2/user-reports`
5. 🟢 [on PR] notice that only 5 results are returned
6. Navigate to `/api/v2/user-reports?limit=2`
7. 🟢 [on PR] notice that only 2 results are returned
8. Navigate to `/api/v2/user-reports?limit=10`
9. 🟢 notice only 5 results are returned

### DIFF
--- a/kobo/apps/user_reports/tests/test_user_reports.py
+++ b/kobo/apps/user_reports/tests/test_user_reports.py
@@ -3,6 +3,7 @@ from unittest.mock import patch
 
 import pytest
 from allauth.socialaccount.models import SocialAccount
+from constance.test import override_config
 from django.conf import settings
 from django.core.cache import cache
 from django.db import connection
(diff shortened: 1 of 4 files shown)


rajpatel24 (approved):
LGTM

PR #7053, opened 2026-05-13
feat(stripe): automated Stripe manual invoicing setup from Django admin DEV-1913

rajpatel24:
### 📣 Summary
This PR automates the initial setup for manual invoicing by allowing the product team to create a Stripe customer and a baseline "community plan" subscription directly from the `Organization` admin panel.

### 📖 Description
Currently, setting up manual invoicing requires the product team to manually create customers and subscriptions in the Stripe dashboard, which is time-consuming and prone to data entry errors.

Solution:
This PR introduces a dedicated action within the Django admin that allows the product team to initiate the invoicing process with a single click. By selecting an organization, the system automatically creates a new Stripe customer and attaches a free "community plan" subscription. This automation ensures that all relevant metadata including organization IDs and owner user details are correctly synced between KoboToolbox and Stripe without manual data entry. This feature is restricted to organizations that do not already have an active Stripe subscription.

Changes:
- Added `create_manual_subscription` to handle the creation of Stripe objects with proper metadata (Org ID, Owner Username, etc.).
- Modified `OrgAdmin` to inject a "Create Stripe Subscription" button into the organization change form.

### 👀 Preview steps

1. ℹ️ Have an admin account, an organization (orgA), and ensure `STRIPE_ENABLED` is True in your environment.
2. Go to Admin -> Organizations -> orgA. The "Create Stripe Subscription" button should be visible at the bottom of the form.
3. 🔴 [on release]: Notice that this feature is not present on the organization change page.
4. 🟢 [on PR]: Notice that upon clicking the button, a Stripe customer and subscription are correctly created with the appropriate metadata.
5. Test further with an organization that already has an active subscription, or try triggering errors (e.g., an organization with no owner).
6. 🟢 Notice that the button is either disabled or returns a clear error message as intended.

### DIFF
--- a/kobo/apps/organizations/admin/organization.py
+++ b/kobo/apps/organizations/admin/organization.py
@@ -70,10 +70,13 @@ class OrgAdmin(BaseOrganizationAdmin):
         if '_create_manual_invoice_subscription' not in request.POST:
             return super().response_change(request, obj)
 
-        if not settings.STRIPE_ENABLED:
+        if (
+            not settings.STRIPE_ENABLED
+            or not config.ENABLE_MANUAL_INVOICE_SUBSCRIPTIONS
+        ):
             self.message_user(
                 request,
-                'Stripe is disabled. Manual invoicing cannot be started.',
+                'Manual invoicing subscriptions are currently disabled.',
                 messages.ERROR,
             )
             return HttpResponseRedirect('.')
--- a/kobo/apps/organizations/admin/organization.py
+++ b/kobo/apps/organizations/admin/organization.py
@@ -1,4 +1,3 @@
-from constance import config
 from django.conf import settings
 from django.contrib import admin, messages
 from django.db import transaction
@@ -13,10 +12,10 @@ if settings.STRIPE_ENABLED:
     from kobo.apps.stripe.exceptions import (
         DefaultCommunityPlanNotFoundError,
         ManualInvoicingSetupError,
-        ManualInvoicingSubscriptionExistsError,
+        ManualSubscriptionExistsError,
     )
-    from kobo.apps.stripe.utils.manual_invoicing import (
-        create_manual_invoicing_subscription,
+    from kobo.apps.stripe.utils.manual_subscription import (
+        create_manual_subscription,
         organization_can_start_manual_invoicing,
     )
 
@@ -67,13 +66,10 @@ class OrgAdmin(BaseOrganizationAdmin):
         return super().change_view(request, object_id, form_url, extra_context)
 
     def response_change(self, request, obj):
-        if '_create_manual_invoice_subscription' not in request.POST:
+        if '_create_manual_subscription' not in request.POST:
             return super().response_change(request, obj)
 
-        if (
-            not settings.STRIPE_ENABLED
-            or not config.ENABLE_MANUAL_INVOICE_SUBSCRIPTIONS
-        ):
+        if not settings.STRIPE_ENABLED:
             self.message_user(
                 request,
                 'Manual invoicing subscriptions are currently disabled.',
@@ -90,8 +86,8 @@ class OrgAdmin(BaseOrganizationAdmin):
             return HttpResponseRedirect('.')
 
         try:
-            subscription = create_manual_invoicing_subscription(obj)
-        except ManualInvoicingSubscriptionExistsError as err:
+            subscription = create_manual_subscription(obj)
+        except ManualSubscriptionExistsError as err:
             self.message_user(request, str(err), messages.ERROR)
         except DefaultCommunityPlanNotFoundError as err:
             self.message_user(request, str(err), messages.ERROR)
@@ -107,8 +103,8 @@ class OrgAdmin(BaseOrganizationAdmin):
             self.message_user(
                 request,
                 (
-                    'Created Stripe customer and subscription '
-                    f'{subscription.id} for manual invoicing.'
+                    'Created Stripe customer and community subscription '
+                    f'{subscription.id} for manual setup.'
                 ),
                 messages.SUCCESS,
             )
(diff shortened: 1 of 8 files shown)


kobo/apps/organizations/admin/organization.py
```diff
@@ -51,6 +66,52 @@ def change_view(self, request, object_id, form_url='', extra_context=None):
...
+                f'Manual invoicing setup failed unexpectedly. {e}',
+                messages.ERROR,
+            )
+        else:
+            self.message_user(
+                request,
+                (
+                    'Created Stripe customer and subscription '
+                    f'{subscription.id} for manual invoicing.'
+                ),
```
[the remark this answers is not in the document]
rajpatel24: I intentionally removed the "community" word from the message because I think it should not be hardcoded. In the future, there may be other subscription plans as well, so keeping the message generic would be better.

Also, since admins are performing this action, they already know what they are doing, and the help text is there to clarify which subscription will be created.

kobo/settings/base.py
```diff
@@ -253,6 +253,21 @@
         ),
         'positive_int',
     ),
+    'ENABLE_MANUAL_INVOICE_SUBSCRIPTIONS': (
+        False,
+        'Enable manual Stripe subscription creation button in Organization admin.',
+    ),
+    'MANUAL_INVOICE_SUBSCRIPTION_HELP_TEXT': (
```
jamesrkiger: I don't think we need to make the text for this feature configurable. This feature is really only for Kobo's internal use. If we ever need to change it we can just update the code.

I'm also wondering if we want to include configuration for enabling/disabling the button. I can at least see the value in being able to turn off the feature if some bug crops up and we don't want anyone to click the button for a while. But I'm not sure that is worth adding config around. Currently the constance config is going to be configurable by the same set of people who would be able to press the subscription creation button, so it's not actually controlling access so much as adding extra steps. What do you think? Are there any use cases I'm missing?

rajpatel24: @jamesrkiger I agree that we probably don’t need Constance config for the help/disabled text.

Regarding the enable/disable configuration for the button, my only thought process was around cases where we might not want to show this button on certain servers or on private/community instances that have Stripe enabled. Otherwise, I agree we may not need that configuration either.

Do you think there could be cases where we would want to hide this button on some instances, or am I overthinking that use case?

jamesrkiger: @rajpatel24 For all intents and purposes, Stripe-related features are only relevant to our two public instances: global and eu. Theoretically private maintainers could set up Stripe for themselves, but we're not going to be particularly interested in facilitating that. As such, I would say you're probably overthinking it. Really the only question here is how we want things to work for our own admins on our own public instances.

jamesrkiger (changes requested):
This looks good so far (though I haven't tested it yet). I would also suggest rethinking the "manual_invoicing" terminology. This is all to facilitate various manual interventions on the part of the sales team, but manual invoicing sounds like a very particular outcome that may not be relevant in all cases. In some cases it may just be an admin is is going to set up a non-standard subscription with normal invoicing for a customer that can't be accessed through our normal user-facing Stripe flow. Maybe something like "manual_subscription" might be a bit more accurate. The AI has also suggested "direct_billing" but that doesn't mean much to me.

kobo/apps/organizations/admin/organization.py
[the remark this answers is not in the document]
rajpatel24: You were right. I’ve updated it accordingly.

kobo/settings/base.py
rajpatel24: Thanks James! I see what you mean, I was definitely overthinking. I have updated the PR accordingly.

kobo/apps/stripe/utils/manual_subscription.py
```diff
@@ -0,0 +1,105 @@
...
+    stripe_customer = stripe.Customer.modify(
+        customer.id,
+        name=customer.name or organization.name,
+        email=owner.email,
+        description=organization.name,
+        metadata={
+            'kpi_owner_username': owner.username,
+            'kpi_owner_user_id': owner.id,
+            'organization_id': organization.id,
+            'manual_invoicing': 'true',
```
jamesrkiger: I'm inclined to say we should remove this manual_invoicing metadata field for now, but if you think it's important to keep track of we could rename it something like "created_in_dj_admin" or something like that.

kobo/apps/organizations/admin/organization.py
```diff
@@ -123,3 +183,27 @@ def _transfer_user_ownership(self, request: 'HttpRequest', new_members: 'QuerySe
...
+    def _get_manual_invoicing_extra_context(self, organization: Organization | None):
+        context = {
+            'show_manual_invoicing_button': False,
+            'can_create_manual_subscription': False,
+            'manual_invoicing_help_text': '',
+        }
+        if not settings.STRIPE_ENABLED or not organization:
+            return context
+
+        can_create = organization_can_start_manual_invoicing(organization)
```
jamesrkiger: Would probably be good to switch out "manual invoicing" for "manual subscription" everywhere for consistency.

kobo/apps/stripe/utils/manual_subscription.py
```diff
@@ -0,0 +1,105 @@
...
+            'request_url': settings.KOBOFORM_URL,
+        },
+    )
+    customer.sync_from_stripe_data(stripe_customer)
+
+    # Re-check against Stripe directly before creating the subscription. This is
+    # a safety net for cases where local dj-stripe data is stale or webhooks have
+    # not synced yet
+    existing_subscriptions = stripe.Subscription.list(
```
jamesrkiger, line 82: Nitpick: I think if you drop status="all" here it won't return cancelled subscriptions and we don't want those anyways.

jamesrkiger (changes requested):
Left a few minor comments, but I've tested it now and it is working well.

kobo/apps/stripe/utils/manual_subscription.py
rajpatel24: I think we should keep this in the metadata, as it could be helpful for identifying the subscription source in case something goes wrong later.

That said, I agree we should rename it. I’ve updated it to `provisioned_via_kobo_admin`.

jamesrkiger (approved):
lgtm

PR #7060, opened 2026-05-15
fix(superuserStats): fix mixed-type Concat error in reports DEV-2128, DEV-2129

noliveleger:
### 📣 Summary

Admin report generation was failing with a Django ORM error, preventing CSV exports from being downloaded.

### 💭 Notes

Django 4.2+ requires all expressions inside `Concat` to have the same `output_field`. `MonthlyXFormSubmissionCounter.year` and `.month` are `IntegerField`s, which conflicted with the `Value('-')` `CharField` separators.

Fix: wrap `F('year')` and `F('month')` with `Cast(..., output_field=CharField())` in all three report tasks that use this pattern:
- `generate_domain_report`
- `generate_continued_usage_report`
- `generate_user_statistics_report`

Also adds a new test suite (`kobo/apps/superuser_stats/tests/test_tasks.py`) covering all four report generation tasks, including regression tests for the Concat issue and NLP counter aggregation fallback.

Note: `_deployment_status` is computed by `set_deployment_status()` on `save()`, so tests that require a deployed status use `Asset.objects.filter(...).update(...)` to bypass the model method.

### 👀 Preview steps

1. ℹ️ Have a superuser account on an instance with submission data
2. Go to `/superuser_stats/reports/`
3. 🔴 [on main] Generating any of the domain, continued-usage, or user-statistics reports raises a 500 / task error with "Expression contains mixed types: CharField, IntegerField"
4. 🟢 [on PR] Reports generate and download successfully

### DIFF
--- a/kobo/apps/superuser_stats/tasks.py
+++ b/kobo/apps/superuser_stats/tasks.py
@@ -95,7 +95,13 @@ def generate_continued_usage_report(output_filename: str, end_date: str):
         )
         submissions_count = MonthlyXFormSubmissionCounter.objects.annotate(
             date=Cast(
-                Concat(F('year'), Value('-'), F('month'), Value('-'), 1),
+                Concat(
+                    Cast(F('year'), output_field=CharField()),
+                    Value('-'),
+                    Cast(F('month'), output_field=CharField()),
+                    Value('-'),
+                    Value('1'),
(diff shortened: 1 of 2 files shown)


rajpatel24 (approved):
LGTM

PR #7015, opened 2026-05-05
build(jquery): bump to v4 DEV-2089

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally
2. [x] update developer docs (API, README, inline, etc.), if any
3. [x] for user-facing doc changes create a Zulip thread at `#Support Docs Updates`, if any
4. [x] draft PR with a title `<type>(<scope>)<!>: <title> DEV-1234`
5. [x] assign yourself, tag PR: at least `Front end` and/or `Back end` or `workflow`
6. [x] fill in the template below and delete template comments
7. [x] review thyself: read the diff and repro the preview as written
8. [x] open PR & confirm that CI passes & request reviewers, if needed
9. [ ] delete this section before merging

### 💭 Notes

Changes here:
- Bumped jQuery to v4
- Bumped jquery-ui to version that supports v4
- Fixed Storybook building after upgrade
- Add explicit imports

Closes #7009

### 👀 Preview steps

1. Make sure UI builds (click around a bit)
2. Go to Form Builder
3. Add few questions and reorder them
4. Drag and drop questions from Library

PR #7061, opened 2026-05-15
fix(formbuilder): force select2 compatibility with jquery DEV-2134

jamesrkiger:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Fixes incompatibility between outdated select2 dependency and jquery v4 via polyfills

### 💭 Notes
Bumping jquery to v4 (#7015) broke the select2 component we use in formbuilder. Select2 is an abandoned dependency we will most likely be stuck with until we rewrite the formbuilder. As such, our choices seem to be: Don't upgrade jquery or add a hack to keep select2 working. I went with the latter, but I'm open to other approaches.

### 👀 Preview steps
1. In formbuilder, try to open "settings" for an individual question 
2. 🔴 [on main] It errors and does nothing
3. 🟢 [on PR] Settings view opens and the tags and appearance selects in "question options" and dropdowns under "skip logic" work correctly.

### DIFF
--- a/jsapp/js/main.js
+++ b/jsapp/js/main.js
@@ -30,7 +30,7 @@ window.$ = window.jQuery = $
 // TODO: remove when select2 is replaced.
 $.isArray = Array.isArray
 $.isFunction = (value) => typeof value === 'function'
-$.trim = (value) => (value == null ? '' : String.prototype.trim.call(value))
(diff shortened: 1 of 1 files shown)


pauloamorimbr (approved):
LGTM! ⭐ 

(and we should get rid of all jQuery in code asap)

PR #7062, opened 2026-05-15
fix(CI): switch all automatic uses npm install to npm ci DEV-2118

jamesrkiger:
### 🗒️ Checklist

1. [x] run linter locally
9. [x] delete this section before merging

### 📣 Summary
Switch all automated uses of `npm install` to `npm ci` to ensure version consistency and security.

### 💭 Notes
npm ci installs strictly from package-lock.json without resolving ranges or modifying the lockfile, whereas npm install can silently pull in newer patch/minor versions that weren't tested. This closes the gap between what developers test locally and what CI/Docker builds install.

Note: the global npm install for check-dependencies in Dockerfile is intentionally left as-is — npm ci cannot install a single package by git URL. Hopefully we will remove that old dependency at some point.

### DIFF
--- a/.github/workflows/biome.yml
+++ b/.github/workflows/biome.yml
@@ -38,11 +38,11 @@ jobs:
         run: npm run copy-fonts
 
       # Cache miss: If node_modules has not been cached,
-      #             `npm install`
+      #             `npm ci`
       #   (This includes `npm run copy-fonts` as post-install step)
-      - name: Install JavaScript dependencies (npm install)
+      - name: Install JavaScript dependencies (npm ci)
         if: steps.cache-nodemodules.outputs.cache-hit != 'true'
-        run: npm install
+        run: npm ci
 
       # Check for Biome formatting errors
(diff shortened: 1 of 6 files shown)


jamesrkiger:
> ### Greptile Summary

> ### Confidence Score: 4/5
> Safe to merge; the CI workflow changes are clean and the Docker script changes are correct under normal conditions.
> 
> All six changed files make the same mechanical substitution and the logic is sound. The only practical risk is in the two Docker development scripts where `npm ci`'s hard requirement for a synced `package-lock.json` differs from `npm install`'s more forgiving behaviour — this could silently break a developer's mid-update workflow but won't affect CI or production builds.
> 
> `docker/webpack-dev-server.sh` and `docker/entrypoint.sh` — the stricter lock-file enforcement introduced by `npm ci` may affect local developer workflows.
>

I don't think this is particularly concerning. If you are adding or updating dependencies, you should know to update the package-lock.

p2edwards (approved):

PR #7064, opened 2026-05-18
feat(subsequences): implement subsequence bulk action cancellation DEV-1589

rajpatel24:
### 📣 Summary
Users can now cancel ongoing bulk transcription and translation jobs directly through the interface, preventing unwanted automated processing.

### 📖 Description
This PR introduces a cancellation endpoint for bulk processing tasks. When a bulk job is cancelled, all pending sub-tasks are halted before they begin, and any active requests currently processing with Google Cloud are instructed to stop. Completed sub-tasks are preserved exactly as they are, and the system safely logs who performed the cancellation.

### 👀 Preview steps

2. Use POST API to create a `SubsequenceBulkAction` along with multiple pending child items.
3. Manually update one child item to have a status of `in_progress` and assign it a dummy string for its `service_id`.
4. 🔴 [on main] notice that sending a PATCH request to `/api/v2/assets/{uid_asset}/advanced-features/bulk-actions/{uid_bulk_action}/` with the payload `{"status": "cancelled"}` fails or is not supported.
5. 🟢 [on PR] notice that this endpoint is active and accepts the cancellation instruction payload.
6. Issue a PATCH request to the bulk action detail URL with `{"status": "cancelled"}`.
7. 🟢 notice that the parent job and pending/in-progress child tasks transition to `cancelled` simultaneously, while any completed tasks remain unchanged.


RuthTurk (commented):
Tested and lgtm. I will approve after the comments by greptile have been addressed :)

RuthTurk (approved):

PR #7067, opened 2026-05-21
fix(languages): correct Orval-generated types for language services fields and LanguagesListParams DEV-2142 DEV-2143

rajpatel24:
### 📣 Summary
This PR corrects two OpenAPI schema inaccuracies in the language endpoints that caused incorrect Orval-generated TypeScript types for the frontend.

### 📖 Description
This PR addresses two schema issues reported by the frontend against the language API endpoints.

**Issue 1: Incorrect schema for `transcription_services` / `translation_services` on the retrieve endpoint**

The list (`/api/v2/languages/`) and retrieve (`/api/v2/languages/{code}/`) endpoints return different shapes for the services fields, but both serializers were annotated with the same `@extend_schema_field(ServicesField)`, causing drf-spectacular to emit one schema for both, which matched only the list endpoint. This led Orval to generate an incorrect array type for the retrieve endpoint's response.

This is resolved by introducing a dedicated `ServicesDetailField` marker class and a corresponding `ServicesDetailFieldExtension` that emits the correct nested-object schema (`{ [serviceCode]: { [regionCode]: string } }`). 

`LanguageSerializer` (retrieve) is updated to use `ServicesDetailField`, while `LanguageListSerializer` (list) continues to use the existing `ServicesField`.

**Issue 2: Missing `q` query parameter in `LanguagesListParams`**

The `LanguageViewSet` relied on `BaseViewSet`'s `SearchFilter` backend to handle the `q` parameter at runtime, but drf-spectacular requires explicit `OpenApiParameter` declarations to include query params in the schema. As a result, `q` was absent from the generated `LanguagesListParams` type, forcing the frontend to cast hook calls with `as any`. This is resolved by adding an explicit `OpenApiParameter(name='q', ...)` declaration to the `list` action schema, consistent with the pattern already used in `TranscriptionServiceViewSet`.

### 👀 Preview steps

Check `/api/v2/docs` for the languages endpoints.

### DIFF
--- a/jsapp/js/api/models/language.ts
+++ b/jsapp/js/api/models/language.ts
@@ -10,8 +10,8 @@ The endpoints are grouped by area of intended use. Each category contains relate
 **General note**: All projects (whether deployed or draft), as well as all library content (questions, blocks, templates, and collections) in the user-facing application are represented in the API as "assets".
  * OpenAPI spec version: 2.0.0 (api_v2)
  */
-import type { LanguageTranscriptionServicesItem } from './languageTranscriptionServicesItem'
-import type { LanguageTranslationServicesItem } from './languageTranslationServicesItem'
+import type { LanguageTranscriptionServices } from './languageTranscriptionServices'
+import type { LanguageTranslationServices } from './languageTranslationServices'
 
 export interface Language {
   /** @maxLength 200 */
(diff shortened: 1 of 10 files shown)


RuthTurk (approved):
LGTM

PR #7070, opened 2026-05-21
feat(formSummary): always display extra metadata fields in summary DEV-2151

pauloamorimbr:
### 📣 Summary
Summary now displays extra metadata fields even with the value is null or empty

### 💭 Notes
Fidl check changed to always display the extra metadata fields
Extra metada section was moved to right below the common metadata, avoiding other data, as languages, to be between the metadatas (common and extra)

### 👀 Preview steps
1. ℹ️ have extra metadata setup on admin
2. have different projects created
3. have a project with unset values for extra metadata fields
4. 🔴 [on main] notice that the fields with no value are not displayed
5. 🟢 [on PR] notice that the fields appear with no value


RuthTurk (approved):
LGTM

PR #7072, opened 2026-05-21
refactor(libraryAssetForm): change component to functional and use proper metadata DEV-2126

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Library asset create and edit forms now fully honor instance-configured project metadata fields, and the select controls used in that form were modernized and stabilized (new shared Select and MultiSelect wrappers, Mantine input migration, shared theme base styles, and UI parity fixes).

### 📖 Description

Until now, the create/edit form for library templates and collections rendered a fixed set of metadata fields (description, sector, country) regardless of what the instance had configured. If an admin disabled one of those, it still showed up; if they enabled `operational_purpose`, `collects_pii`, or any custom extra metadata field, it was nowhere to be found. The form now behaves the same way Project Settings already does and renders exactly the fields the environment exposes.

### 💭 Notes

Changes here:

Core DEV-2126 behavior and TypeScript migration:
- LibraryAssetForm.tsx
  - Replaced legacy class and mixin flow with a functional TypeScript component.
  - Metadata visibility is driven by environment config via getProjectMetadataFieldsAsSimpleDict.
  - Supports operational_purpose, collects_pii, and extra metadata fields.
  - Uses shared ExtraProjectMetadataFields and keeps submit payload metadata-complete to avoid silent clobbering.
  - Reflux listeners are centralized with explicit cleanup.
  - Migrated text inputs from TextBox to Mantine TextInput and Textarea.
  - Migrated select usage to shared Select and MultiSelect wrappers.
- jsapp/js/components/modalForms/extraProjectMetadataFields.js to ExtraProjectMetadataFields.tsx
  - Ported to TypeScript with typed props and same functional behavior.
- jsapp/js/components/modalForms/libraryAssetForm.js removed.
- LibraryAssetForm.module.scss renamed from lowercase filename.
- projectSettings.js and bigModal.js
  - Import path updates for migrated and renamed files.

Select and MultiSelect follow-up work added:
- Select.tsx and MultiSelect.tsx (common)
  - Introduced dedicated MultiSelect wrapper (separate from Select).
  - Added stable right-section behavior and chevron handling.
  - MultiSelect defaults to withinPortal false to avoid ResizeObserver loop issues observed in modal contexts.
    - An error was happening when selecting enough countries to cause them to wrap in the new line
  - select.types.ts
- useSelectChevron.tsx
  - DRY extraction of shared select types and chevron and open-state logic.
- SelectBase.module.css, Select.module.css, MultiSelect.module.css, Select.tsx, MultiSelect.ts (theme)
  - Consolidated shared Select and MultiSelect styles into SelectBase.
  - Kept component-specific styles local (pill style, placeholder differences, svg sizing).
  - Ensured Select and MultiSelect default height parity and stable chevron and clear alignment.
- wrappedSelect.tsx
  - Added deprecation note directing consumers to shared Select wrapper.
- eslint.config.mjs
  - Added restricted import rule for Mantine MultiSelect to encourage usage of the shared wrapper.

It was easier to fix the issue while migrating to functional component than doing these two things separately.

### 👀 Preview steps

1. ℹ️ have an account, log in
2. as a superuser, go to **Django admin → Constance → Config** and configure project metadata fields — set `operational_purpose` and `collects_pii` to required (or just visible), and add at least one **extra** project metadata field (e.g. a single-select called `funding_source`)
3. back in KoboToolbox, go to **Library** in the left navigation
4. click **NEW → Collection** (or **Template**)
5. 🔴 [on main] notice that only Name, Description, Organization, Sector, Country, and Tags are shown — `operational_purpose`, `collects_pii`, and your custom `funding_source` field are missing in action
6. 🟢 [on PR] notice that Operational Purpose, "Does this project collect personally identifiable information?", and your custom `funding_source` dropdown all show up, in the same shape as on Project Settings
7. fill in the form and click **Create** — open the new collection/template settings and confirm all values were saved
8. now go to an existing library template, open **Settings** modal
9. 🟢 confirm the same fields appear pre-filled with whatever was saved
10. as the superuser, **disable** `description` in the constance config
11. open the library form again
12. 🟢 notice that the Description field is gone (and isn't quietly clobbering saved values on submit)


jsapp/js/components/modalForms/LibraryAssetForm.tsx
```diff
@@ -17,308 +16,319 @@ import { ASSET_TYPES, type AssetTypeName } from '#/constants'
...
 
-          <ExtraProjectMetadataFields values={this.state.extraMetadataFields} onChange={this.onExtraFieldChange} />
-
+        {metadataFields.collects_pii && (
           <bem.FormModal__item>
-            <KoboTagsInput
-              tags={this.state.fields.tags}
-              onChange={this.onAnyFieldChange.bind(this, 'tags')}
-              label={t('Tags')}
+            <WrappedSelect
```
duvld: This label and the extra metadata label opens upwards:

Image

Do you think we can swap to mantine Select here (for it directional combobox dropdown prop) since this is now a functional component and change all of the other selects?

duvld (commented):
Just one UX question, functionality is LGTM

magicznyleszek:

jsapp/js/components/modalForms/LibraryAssetForm.tsx
magicznyleszek: They open upwards if there is no enough space below, otherwise they open downwards (`react-select` works that way). I will migrate to mantine one though :)

magicznyleszek: Also marked `WrappedSelect` as deprecated

jsapp/js/components/modalForms/LibraryAssetForm.tsx
```diff
@@ -1,324 +1,359 @@
...
+              data={[
+                { value: 'Yes', label: t('Yes') },
+                { value: 'No', label: t('No') },
+              ]}
+              clearable
             />
           </bem.FormModal__item>
+        )}
+
+        <ExtraProjectMetadataFields values={extraMetadataFields} onChange={onExtraFieldChange} />
```
duvld: This component still uses the old Select component:

Image

jsapp/js/components/modalForms/LibraryAssetForm.tsx
```diff
@@ -1,324 +1,359 @@
...
+        {metadataFields.country && (
           <bem.FormModal__item>
-            <WrappedSelect
-              label={t('Primary Sector')}
-              value={this.state.fields.sector}
-              onChange={this.onAnyFieldChange.bind(this, 'sector')}
-              options={SECTORS}
-              isLimitedHeight
-              isClearable
+            <MultiSelect
```
duvld, line 297: UI nitpick, but there's a small gap between Multi and regular Select. Looking at the bulk processing mocks (that use `Select` now for the language selector), I think we should go with no gap like the regular Select

Regular:
Image

Multi:
Image

duvld (commented):
Metadata is displaying correctly (preview steps work), some UI points and the missing `styles` is breaking webpack

jsapp/js/components/modalForms/LibraryAssetForm.tsx
magicznyleszek: Updated extra metadata to use new components

jsapp/js/components/modalForms/LibraryAssetForm.tsx
magicznyleszek: Adjusted them to look the same :)

magicznyleszek:

duvld (commented):
Preview steps look good, but it looks like the required functionality doesn't work:

Image

The above is with everything in `PROJECT_METADATA_FIELDS` in constance having `required: true`. This is the case on main as well, so we can create a new linear issue but I think it's relevant enough to do it here

magicznyleszek:
> Preview steps look good, but it looks like the required functionality doesn't work:
> Image
> 
> The above is with everything in `PROJECT_METADATA_FIELDS` in constance having `required: true`. This is the case on main as well, so we can create a new linear issue but I think it's relevant enough to do it here

I have tested it and reproduced it. But it turns out that we can create a template/collection with blank required fields - i.e. backend allows it. So I think this is either a cross-team bug or expected functionality. I asked it here https://chat.kobotoolbox.org/#narrow/channel/4-Dev/topic/Required.20user.20metadata.20fields.20issue/near/844610 and will follow it in a separate linear issue if needed.

PR #7080, opened 2026-05-22
feat(bulkProcessing): implement background execution and polling for bulk actions DEV-1494

rajpatel24:
📣 Summary
Users can now successfully run bulk transcription and translation jobs in the background, with real-time progress tracking and automatic recovery if a network interruption occurs.

📖 Description
This PR implements the core background processing engine for SubsequenceBulkAction using Celery. It ensures that large-scale audio processing and translation tasks run asynchronously without blocking the main API threads.

**Key Architectural Highlights:**
* **Job Orchestration:** Added `start_bulk_item_job` to safely delegate individual submissions to the existing `SubmissionSupplement.revise_data()` flow, ensuring complete compatibility with existing NLP schemas.
* **Progress Polling:** Implemented `update_batch_status` utilizing pessimistic database locking (`select_for_update(skip_locked=True)`) to safely calculate and update the parent job's overall progress percentage without triggering deadlocks.
* **Feature Auto-Provisioning:** Updated the creation serializer to automatically provision the necessary `QuestionAdvancedFeature` (the backend column) if it does not already exist for the requested language.
* **System Resilience:** Introduced a `resume_stuck_bulk_actions` watchdog task (run every 5 minutes) to automatically resume polling for jobs that experience worker restarts, timeouts, or pod terminations.
* **Async Guard Fix:** Updated the execution guard in `base.py` to ensure long-running Google Cloud operations properly trigger the downstream polling mechanism without getting trapped in an infinite `in_progress` state.
* `BULK_ACTION_STATUS_POLL_INTERVAL` (Default: `30` seconds): Determines how frequently the background task queries individual child submissions to compute the overall parent batch progress percentage.
* `BULK_ACTION_STUCK_THRESHOLD` (Default: `300` seconds / 5 minutes): The expiration window after which an active bulk action job is flagged as "stuck" due to potential worker pod crashes or unexpected container restarts.

### 👀 Preview steps

1. ℹ️ Have an account, a project, and a few submissions with audio questions.
2. Make sure to restart the containers after checking out the PR branch.
3. Trigger a bulk transcription job: From Postman or Swagger Docs (`/api/v1/docs`), hit the `POST /api/v2/assets/{uid_asset}/advanced-features/bulk-actions/` API after adding multiple submission UUIDs.

Example payload:

```json
{
    "action_id": "automatic_google_transcription",
    "question_xpath": "q1",
    "submission_uuids": [
        "3ac65b39-9a07-4f68-aad0-66fde7e15442",
        "aa50c534-a84d-43fd-b3af-0d0d2b0932ac"
    ],
    "params": {
        "language": "en"
    }
}
```

4. 🟢 [On PR] Notice that the API immediately returns a `201 Created` response with a status of `in_progress`.
5. Monitor your Celery worker logs and observe `start_bulk_item_job` processing individual files in the background.
6. Query the bulk action detail API endpoint periodically (`/api/v2/assets/{uid_asset}/advanced-features/bulk-actions/<action_uid>/`).
7. 🟢 [On PR] Notice that the `progress` field increments successfully as Google returns processed results, eventually transitioning the parent job to `complete`.

Note: Use a longer audio file (>30–40 seconds) to properly test this flow, as it helps verify that the background jobs keep retrying while waiting for results from Google.


kobo/apps/subsequences/serializers.py
```diff
@@ -314,19 +314,58 @@ def create(self, validated_data):
...
+                bulk_action.start_batch()
+            return bulk_action
         except IntegrityError as err:
             raise serializers.ValidationError(
                 'One or more submissions already have an active matching bulk action.'
             ) from err
 
+    def _ensure_question_advanced_feature(
+        self,
+        *,
```
noliveleger, line 343: Did you add this in purpose?

kobo/apps/subsequences/tasks.py
```diff
@@ -120,3 +152,454 @@ def poll_run_external_process_failure(sender=None, **kwargs):
...
+        action_data=action_data,
+        status='failed',
+    )
+
+
+@celery_app.task(
+    autoretry_for=(SubsequenceTimeoutError,),
+    retry_backoff=5,
+    retry_backoff_max=60,
+    max_retries=482,
```
noliveleger, line 210: Would be great to leave a comment to explain why 482!

kobo/apps/subsequences/tasks.py
```diff
@@ -120,3 +152,454 @@ def poll_run_external_process_failure(sender=None, **kwargs):
...
+    """
+    SubsequenceBulkActionItem = apps.get_model(  # noqa: N806
+        'subsequences',
+        'SubsequenceBulkActionItem',
+    )
+
+    item = (
+        SubsequenceBulkActionItem.objects.select_related('parent', 'parent__asset')
+        .get(pk=bulk_action_item_id)
+    )
```
noliveleger: Do we need to load Asset.content here?

kobo/apps/subsequences/tasks.py
```diff
@@ -120,3 +152,454 @@ def poll_run_external_process_failure(sender=None, **kwargs):
...
+
+    # Handle early dequeues (race conditions) and ignore terminal states
+    if item.status == 'pending':
+        if item.parent.status != 'in_progress':
+            return
+        item.status = 'in_progress'
+        item.save(update_fields=['status', 'date_modified'])
+    elif item.status != 'in_progress':
+        return
```
noliveleger: We should avoid using the string values instead of the model constants, e.g. : `progress` vs `BulkActionStatus.IN_PROGRESS`  .

Please change everywhere ;-)

kobo/apps/subsequences/migrations/0012_add_progress_field_to_subsequencebulkaction.py
```diff
@@ -0,0 +1,18 @@
...
+        ('subsequences', '0011_add_cancelled_by_field_to_subsequencebulkaction'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='subsequencebulkaction',
+            name='progress',
+            field=models.PositiveSmallIntegerField(default=0),
+        ),
+    ]
```
noliveleger, line 18: I don't see this field in the serializer? Is it in purpose?

kobo/apps/subsequences/models.py
```diff
@@ -501,6 +524,36 @@ def cancel(self, *, cancelled_by: str | None = None) -> 'SubsequenceBulkAction':
...
+
+        delay_between_jobs = self._get_delay_between_jobs()
+        item_ids = list(self.items.values_list('pk', flat=True))
+        for index, item_id in enumerate(item_ids):
+            countdown = (index * delay_between_jobs) if delay_between_jobs else 0
+            start_bulk_item_job.apply_async(args=(item_id,), countdown=countdown)
+
+        update_batch_status.apply_async(
+            args=(self.pk,),
+            countdown=getattr(settings, 'BULK_ACTION_STATUS_POLL_INTERVAL', 30),
```
noliveleger: Please stop using `getattr(settings.` (looks like a 🤖 thing :-P). It's there in many places. 
Since it's declared in the settings, there is no need to use this safety net. It makes the code harder to read and make lines longer than needed. 

Moreover `settings.BULK_ACTION_STATUS_POLL_INTERVAL` already has a default. 

```python
BULK_ACTION_STATUS_POLL_INTERVAL = env.int(
    'BULK_ACTION_STATUS_POLL_INTERVAL',
    30,
)
```

kobo/apps/subsequences/tasks.py
```diff
@@ -120,3 +152,454 @@ def poll_run_external_process_failure(sender=None, **kwargs):
...
+
+    action = item.parent
```
noliveleger: I would prefer to rename this var to `bulk_action` to avoid confusion with NLP actions ? What do you think?

kobo/apps/subsequences/tasks.py
```diff
@@ -120,3 +152,454 @@ def poll_run_external_process_failure(sender=None, **kwargs):
...
+    submission_uuid = remove_uuid_prefix(submission[SUBMISSION_UUID_FIELD])
+    return SubmissionSupplement.retrieve_data(
+        asset,
+        submission_root_uuid=submission_uuid,
+    )
+
+
+def _extract_bulk_item_status(
+    asset,
+    action,
```
noliveleger: I think we should rename this parameter (`action`) to something else and we should also use typing aliases to make it clear.  I would also add the type to `Asset` even it's obvious. 

Everywhere, we use `action` , we should consider to think about its name to make things obvious.

kobo/apps/subsequences/tasks.py
```diff
@@ -120,3 +152,454 @@ def poll_run_external_process_failure(sender=None, **kwargs):
...
+        'SubsequenceBulkAction',
+    )
+    threshold_seconds = getattr(settings, 'BULK_ACTION_STUCK_THRESHOLD', 300)
+    cutoff = timezone.now() - timedelta(seconds=threshold_seconds)
+    action_ids = SubsequenceBulkAction.objects.filter(
+        status='in_progress',
+        date_modified__lt=cutoff,
+    ).values_list('pk', flat=True)
+    for action_id in action_ids:
+        update_batch_status.delay(action_id)
```
noliveleger: Once again: I would use `bulk_action_ids` instead of `action_ids`

kobo/apps/subsequences/tasks.py
```diff
@@ -120,3 +152,454 @@ def poll_run_external_process_failure(sender=None, **kwargs):
...
     )
+    _update_bulk_action_item_status(
+        submission=submission,
+        action_data=action_data,
+        status='failed',
+    )
+
+
+@celery_app.task(
+    autoretry_for=(SubsequenceTimeoutError,),
```
noliveleger: Are we sure that `SubsequenceTimeoutError` can be raised by the code executed by `start_bulk_item_job`? 
Sounds like a copy/paste?

kobo/apps/subsequences/tasks.py
```diff
@@ -120,3 +152,454 @@ def poll_run_external_process_failure(sender=None, **kwargs):
...
+        'subsequences',
+        'SubsequenceBulkActionItem',
+    )
+
+    item = (
+        SubsequenceBulkActionItem.objects.select_related('parent', 'parent__asset')
+        .get(pk=bulk_action_item_id)
+    )
+
+    # Handle early dequeues (race conditions) and ignore terminal states
```
noliveleger: I don't see any lock mechanism on the item. 
Could a second `start_bulk_item_job` occurence start with the same uid and change the status? 
maybe a select_for_update could be useful?

kobo/settings/base.py
```diff
@@ -1265,6 +1265,23 @@ def dj_stripe_request_callback_method():
...
+}
+
+BULK_ACTION_STATUS_POLL_INTERVAL = env.int(
+    'BULK_ACTION_STATUS_POLL_INTERVAL',
+    30,
+)
+BULK_ACTION_STUCK_THRESHOLD = env.int(
+    'BULK_ACTION_STUCK_THRESHOLD',
+    BULK_ACTION_STATUS_POLL_INTERVAL * 10,
+)
```
noliveleger, line 1316: nitpick: I would add a blank line here, just to separate the block.

kobo/apps/subsequences/tasks.py
```diff
@@ -32,24 +40,47 @@ def poll_run_external_process(
...
-
     Asset = apps.get_model('kpi', 'Asset')  # noqa: N806
     SubmissionSupplement = apps.get_model(
         'subsequences', 'SubmissionSupplement'
     )  # noqa: N806
+    request_action_data = _get_bulk_action_request_data(action_data)
     incoming_data = {
         '_version': SCHEMA_VERSIONS[0],
-        question_xpath: {action_id: action_data},
+        question_xpath: {action_id: dict(request_action_data)},
```
noliveleger: _get_bulk_action_request_data already returns a copy of  `action_data`. 
I think this `dict(request_action_data)` is not needed.

kobo/apps/subsequences/tasks.py
```diff
@@ -120,3 +152,454 @@ def poll_run_external_process_failure(sender=None, **kwargs):
...
+def _get_bulk_action_request_data(action_data: dict) -> dict:
+    """
+    Return action input data without service result fields
+
+    Automatic actions mutate their input dict by merging service responses into
+    it before writing supplement versions. Celery retries must receive only the
+    original request fields, otherwise schema validation rejects result-only
+    keys such as status, error, or value on the next poll.
+    """
+    request_action_data = dict(action_data)
```
noliveleger: Would `deepcopy` be better here? 

```
  original = {'language': 'en', 'options': {'model': 'chirp'}}

  shallow = dict(original)
  shallow['options']['model'] = 'latest_long'
  print(original['options']['model'])  # 'latest_long'
```

noliveleger (changes requested):

kobo/apps/subsequences/serializers.py
rajpatel24: Not sure if you’re referring to the `_ensure_question_advanced_feature` function in general, but I added this on purpose because `QuestionAdvancedFeature` is required to run the NLP actions. I also think `QuestionAdvancedFeature` should be there even when `SubsequenceBulkAction` is present.

kobo/apps/subsequences/tasks.py
rajpatel24: No longer needed.

kobo/apps/subsequences/tasks.py
rajpatel24: Not really. Updated the code to use `.defer('parent__asset__content')`.

kobo/apps/subsequences/tasks.py
rajpatel24: Updated to `action_context`

kobo/apps/subsequences/tasks.py
rajpatel24: Thanks for catching it!  I originally added `autoretry_for=(SubsequenceTimeoutError,)` in `start_bulk_item_job`, following the same pattern as `poll_run_external_process`, where the error indicates that Google hasn’t finished and the task should be retried. However, retrying `start_bulk_item_job` itself is not the right approach here. When `GoogleTranscriptionService` raises `SubsequenceTimeoutError`, it already indicates an `in_progress` state, so we should instead handle that and trigger `poll_run_external_process` rather than retrying the whole task. I've updated the code accordingly.

kobo/apps/subsequences/tasks.py
rajpatel24: Added row lock

kobo/apps/subsequences/tasks.py
rajpatel24: Indeed!

kobo/apps/subsequences/migrations/0012_add_progress_field_to_subsequencebulkaction.py
rajpatel24: I was going to cover that in the documentation PR, but I’ve added it in this one now

noliveleger:
@greptileai re-review

kobo/apps/subsequences/serializers.py
noliveleger: @rajpatel24 , sorry! I was speaking about the `*` to force positional arguments.

kobo/apps/subsequences/tasks.py
```diff
@@ -175,163 +172,111 @@ def start_bulk_item_job(bulk_action_item_id: str):
...
 
-    item = (
-        SubsequenceBulkActionItem.objects.select_related('parent', 'parent__asset')
-        .get(pk=bulk_action_item_id)
-    )
+    with transaction.atomic():
+        item = (
+            SubsequenceBulkActionItem.objects.select_related('parent', 'parent__asset')
+            .defer('parent__asset__content')
+            .select_for_update()
```
noliveleger: Make sure to release the lock as soon as possible. 
I think the select_for_update is only needed to change the status. 

You can use another `transaction.atomic()` block if you need it. 
```
with transaction.atomic():
     item = S...select_for_update()
     item.status = ...
     item.save()
     
with transaction.atomic():     
    # other stuff we want to run atomically 
```

Otherwise, you can also return as soon as the lock is not acquired to avoid making the Celery worker wait for the lock to be released.

noliveleger (changes requested):

noliveleger:
@rajpatel24 I guess the real endpoint is `/api/v2/assets/{uid_asset}/advanced-features/bulk-actions/`?

rajpatel24:
> @rajpatel24 I guess the real endpoint is `/api/v2/assets/{uid_asset}/advanced-features/bulk-actions/`?

Yes! Sorry, I didn’t include the full endpoint in the preview steps. I thought it was implied since it’s an extension of `advanced_features`. 🤦‍♂️

rajpatel24:
@greptileai re-review

rajpatel24:
@greptileai re-review

noliveleger (approved):
LGTM

rajpatel24:
@greptileai re-review

rajpatel24:
@greptileai Regarding `_schedule_batch_tasks` only filtering for `IN_PROGRESS`: This is intentional and safe. In `start_batch()`, we atomically transition the parent and all pending child items to `IN_PROGRESS` before the `transaction.on_commit` hook ever schedules the tasks. Therefore, by the time `_schedule_batch_tasks` runs, it is mathematically impossible for a child item to still be in the PENDING state.

PR #7090, opened 2026-05-26
feat(asset): allow owners to delete empty assets DEV-1988

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Allow users to delete assets they have created if there are no submissions.

### 📖 Description
Extends permission to delete an asset to a user if:
1. They created the asset
2. They have the manage_asset permission
3. The asset has no submissions

### 💭 Notes
Both single delete and bulk delete go through the bulk serializer, so the change had to be made there in addition to the permissions check for the single DELETE.
Adds the ability to pass a prefetched list of ObjectPermissions to pass to `has_perm` to avoid multiple round trips to the DB for each asset.

### 👀 Preview steps

2. Delete all submissions from the project
4. Add the account to an MMO with a different owner so the project is transferred
5. From a terminal, DELETE to `api/v2/assets/<project_uid>` as the account user (not the organization owner)
7. 🔴 [on main] 403
8. 🟢 [on PR] the asset is deleted
9. Create another asset and make sure it has 0 submissions
10. From a terminal, POST to `/api/v2/assets/bulk` with
```
{
"payload": {
     "asset_uids": ["<asset uid>"],
     "action": "delete"
}
}
```
7. 🔴 [on main] 403
8. 🟢 [on PR] the asset is deleted


kpi/serializers/v2/asset.py
```diff
@@ -249,14 +249,35 @@ def _has_perms(self, payload: dict, asset_uids: list[str]):
...
+            # note: prefetch_related('permissions') doesn't help here because has_perm
+            # gets ObjectPermissions and then filters by asset rather than going
+            # through asset.permissions
+            for asset in Asset.objects.filter(uid__in=asset_uids).defer('content'):
+                has_submissions = (
+                    asset.has_deployment and asset.deployment.submission_count > 0
+                )
+                if (
+                    asset.created_by != self.__user.username
+                    or not asset.has_perm(self.__user, PERM_MANAGE_ASSET)
```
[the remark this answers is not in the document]
rgraber: `has_perm` does work in addition to the ObjectPermission check so we can't just replace it with our own check

kpi/serializers/v2/asset.py
```diff
@@ -249,14 +249,49 @@ def _has_perms(self, payload: dict, asset_uids: list[str]):
...
             ).count()
+            if objects_count != len(asset_uids):
+                raise exceptions.PermissionDenied()
         else:
             objects_count = Asset.objects.filter(
                 owner__in=user_filter,
                 uid__in=asset_uids,
             ).count()
+            if objects_count == len(asset_uids):
+                # all assets are owned by the user, so they can delete all of them
```
noliveleger: This comment is not really correct: Something like would be better 

> All assets are owned by the user (or by their organization if the user is an org admin).

kpi/serializers/v2/asset.py
```diff
@@ -249,14 +249,49 @@ def _has_perms(self, payload: dict, asset_uids: list[str]):
...
+            ):
+                has_submissions = (
+                    asset.has_deployment and asset.deployment.submission_count > 0
+                )
+                if (
+                    asset.created_by != self.__user.username
+                    or not asset.has_perm(self.__user, PERM_MANAGE_ASSET)
+                    or has_submissions
+                ):
+                    raise exceptions.PermissionDenied()
```
noliveleger: I would like to avoid the loop and the O(n) queries. 

What about something like? 

```python

          owner_ids = {u.pk for u in user_filter}  # set() 

          # Fetch all matching assets in a single query. 
          # Cast as a list to avoid fetching data over and over again
          asset_objects = list(
              Asset.objects.filter(uid__in=asset_uids)
              .only('uid', 'owner_id', 'created_by', '_deployment_data')
          )

          # At least one UID doesn't exist, keep consistent with the 403
          # raised when all other assets are owned by the requester.
          if len(asset_objects) != len(asset_uids):
              raise exceptions.PermissionDenied()

          # All assets are owned by the user (or by their organization if the user is an org admin).
          non_owned = [a for a in asset_objects if a.owner_id not in owner_ids]
          if not non_owned:
              return

          # Non-owners can only delete assets they created
          if any(a.created_by != self.__user.username for a in non_owned):
              raise exceptions.PermissionDenied()

          # Non-owners must have manage_asset on every asset
          code_names = get_cached_code_names(Asset)  # those two lines could be at the top of the method 
          perm_dict = code_names[PERM_MANAGE_ASSET]  # because they are also used for archiving.
          non_owned_ids = [a.pk for a in non_owned]
          perm_count = ObjectPermission.objects.filter(
              user=self.__user,
              permission_id=perm_dict['id'],
              asset_id__in=non_owned_ids,
              deny=False,
          ).count()
          if perm_count != len(non_owned_ids):
              raise exceptions.PermissionDenied()

          # Last check, assets with submissions cannot be deleted by non-owners
          formids = [  # retrieve XForm.pks from backend reponse
              a._deployment_data.get('backend_response', {}).get('formid')
              for a in non_owned
              if 'backend' in a._deployment_data
          ]
          if formids:  # if empty, there are no deployed projects, safe to be deleted. 
              # import XForm

              if XForm.all_objects.filter(pk__in=formids, num_of_submissions__gt=0).exists():
                  raise exceptions.PermissionDenied()
```

Pros:
- Goes from O(n) queries to 3 fixed queries (assets, permissions, xform).

Cons: 
- A little bit more verbose (but mitigated with comments)

kpi/tests/api/v2/test_api_asset_bulk_actions.py
```diff
@@ -454,3 +468,72 @@ def test_users_cannot_undelete(self):
...
+        if not has_manage_asset:
+            a1.remove_perm(someuser, PERM_MANAGE_ASSET)
+        if not is_empty:
+            a1.deployment.mock_submissions([{}])
+        self.client.force_login(someuser)
+        response = self._create_send_payload([a0.uid, a1.uid], 'delete')
+        a0.refresh_from_db()
+        a1.refresh_from_db()
+        if can_delete:
+            assert response.status_code == 200
```
noliveleger: Let's use restframework constants  status.HTTP_200_OK and status.HTTP_403_FORBIDDEN (like you did in test_deleting_mixed_assets_owned_and_created)

kpi/tests/api/v2/test_api_assets.py
```diff
@@ -691,6 +692,60 @@ def test_shared_asset_hidden_when_owner_inactive(self):
...
+            asset.remove_perm(someuser, PERM_MANAGE_ASSET)
+        if not is_empty:
+            asset.deployment.mock_submissions([{}])
+        self.client.force_login(someuser)
+        response = self.client.delete(
+            reverse(self._get_endpoint('asset-detail'), args=(asset.uid,))
+        )
+        asset.refresh_from_db()
+        if can_delete:
+            assert response.status_code == 204
```
noliveleger: Same comments for status code

kpi/permissions.py
```diff
@@ -165,6 +166,11 @@ def has_object_permission(self, request, view, obj):
             and user_has_project_view_asset_perm(obj, user, PERM_VIEW_ASSET)
         ):
             return True
+        if method == 'DELETE':
+            if obj.has_perm(user, PERM_MANAGE_ASSET):
+                if not obj.has_deployment or obj.deployment.submission_count == 0:
+                    if obj.created_by == user.username:
+                        return True
```
noliveleger: Could the 4 `if`s  be in the same `if` statement?
Moreover, why keeping the `DELETE` outside the first `if` and not be consistent with `GET` and `PATCH`?

noliveleger (changes requested):

kpi/serializers/v2/asset.py
rgraber: I don't think this actually matches the logic of `asset.has_perm(user,<perm>)`. I wonder if I can update `has_perm` to accept a precalculated list of ObjectPermissions instead of always making a DB round trip. I think that would be safer.

kpi/serializers/v2/asset.py
```diff
@@ -252,17 +253,24 @@ def _has_perms(self, payload: dict, asset_uids: list[str]):
             if objects_count != len(asset_uids):
                 raise exceptions.PermissionDenied()
         else:
-            objects_count = Asset.objects.filter(
-                owner__in=user_filter,
-                uid__in=asset_uids,
-            ).count()
-            if objects_count == len(asset_uids):
+            user_filter_as_ids = [user.id for user in user_filter]
+            all_assets = Asset.objects.filter(uid__in=asset_uids).select_related(
```
noliveleger: I think here we could optimize the fields we need.  No need to load big fields such as `content`, `summary`etc..

```python
 all_assets = Asset.objects.filter(uid__in=asset_uids).select_related(
      'owner__organization'  # useful in `has_perm` 
  ).only(
      'uid', 'created_by', '_deployment_data', 'asset_type', 'owner'
  )
```

kpi/serializers/v2/asset.py
```diff
@@ -249,14 +250,79 @@ def _has_perms(self, payload: dict, asset_uids: list[str]):
...
+            if any(a.created_by != self.__user.username for a in unowned_assets):
+                raise exceptions.PermissionDenied()
+
+            # check 2: does the user have manage_asset?
+            all_perms = [
+                o
+                for o in ObjectPermission.objects.filter(
+                    asset__uid__in=asset_uids, user__in=user_filter
+                )
+                .select_related('permission')
```
noliveleger: I don't think this is needed since we are returning a dict, not a list of objects.

noliveleger (changes requested):

kpi/tests/api/v2/test_api_assets.py
```diff
@@ -691,6 +692,60 @@ def test_shared_asset_hidden_when_owner_inactive(self):
...
+            assert asset.pending_delete
+        elif is_creator:
+            assert response.status_code == 403
+            assert not asset.pending_delete
+        else:
+            assert response.status_code == 404
+            assert not asset.pending_delete
```
noliveleger: Please use DRF status constant (e.g. HTTP_204_NO_CONTENT) everywhere.

noliveleger (changes requested):

noliveleger (approved):
LGTM

PR #7091, opened 2026-05-26
refactor(formLanguagesManager): improve flaky tests DEV-2063

magicznyleszek:
### 💭 Notes

Improved `FormLanguagesManager` Storybook `BasicFlow` test stability in CI by replacing brittle selectors with scoped, deterministic UI interactions.

Changes here:
- FormLanguagesManager story
  - [jsapp/js/project/FormLanguagesManager/FormLanguagesManager.stories.tsx](jsapp/js/project/FormLanguagesManager/FormLanguagesManager.stories.tsx)
  - Replaced global icon-class pagination targeting with table-scoped pagination lookup, so the test no longer depends on fragile DOM/class timing.
  - Removed exact text-node dependency on `Page 1 of 2` (which may be split across elements in CI).
  - Scoped translation input selection to the `Question 11` table row to avoid ambiguous textbox selection when multiple inputs are rendered.
  - Added concise inline comments near changed blocks to explain why these selectors/waits are more robust.

Why this helps:
- CI flakiness came from selector brittleness and timing assumptions rather than functional behavior.
- The flow now waits on stable content and interacts only within the relevant container, which reduces random retries/failures.

### 👀 Preview steps

1. Open Storybook and navigate to Features → FormLanguagesManager.
2. Run the BasicFlow story play test.
3. 🔴 On main: test may fail sometimes and require rerunning to pass.
5. 🟢 On this PR: test never fails on CI.

### DIFF
--- a/jsapp/js/project/FormLanguagesManager/FormLanguagesManager.stories.tsx
+++ b/jsapp/js/project/FormLanguagesManager/FormLanguagesManager.stories.tsx
@@ -198,11 +198,27 @@ export const BasicFlow: Story = {
     })
 
     await step('Go to next translations page', async () => {
-      const nextPageButton = document.querySelector('.k-icon-angle-right')?.closest('button') as
-        | HTMLButtonElement
-        | undefined
+      // Wait for first-page row content instead of matching pagination text,
+      // because the "Page X of Y" label can be split across nested elements.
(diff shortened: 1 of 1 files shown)


jamesrkiger (approved):
lgtm

PR #7096, opened 2026-05-27
fix(admin): fix adding users in Django Admin DEV-2180

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Fixes a bug in Django admin that was preventing admins from adding users.

### 💭 Notes
The Django 5 upgrade introduced a new concept of `usable_password`, which is a setting that can allow/forbid users to authenticate with a password (as opposed to SSO or LDAP or something). This field is included by default in the UserAdmin, but not present on the model. Instead, it is created by a mixin that is present in `AdminUserCreationForm` but not regular `UserCreationForm.` 
The solution is to update our `UserCreationForm` to use `AdminCreationUserForm` as a base and just hide the `usable_password` field by overriding the `fieldsets`.

### 👀 Preview steps

2. Go to Django admin -> Users
3. Add user
4. 🔴 [on release] 500
5. 🟢 [on PR] you are able to successfully add a new user

### DIFF
--- a/hub/admin/extend_user.py
+++ b/hub/admin/extend_user.py
@@ -6,9 +6,9 @@ from django.contrib import admin, messages
 from django.contrib.admin import SimpleListFilter
 from django.contrib.auth.admin import UserAdmin
 from django.contrib.auth.forms import (
-    UserChangeForm as DjangoUserChangeForm,
     AdminUserCreationForm,
 )
+from django.contrib.auth.forms import UserChangeForm as DjangoUserChangeForm
 from django.core.exceptions import ValidationError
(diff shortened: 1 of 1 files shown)


rgraber:
The original greptile comment about removing `usable_password` from the fieldset making it always False is incorrect. The `usable_password` field gets set to the empty string, which `validate_passwords` treats as True. However, that is an internal implementation detail for Django that could change, so it does make sense to explicitly make it True (which will just make it do the normal password validation flow).

noliveleger (approved):
LGTM

PR #7110, opened 2026-06-03
feat(deleteAsset): migrate asset delete modal to Mantine DEV-1994

pauloamorimbr:
### 📣 Summary
Deleting a project now uses the new Mantine ported confirmation modal flow while keeping the same safety checks as before.

### 💭 Notes
Where to test the changed modal:
- Project Table quick actions (single selected project): click Delete
- Asset row actions (Library/Forms listing): More actions → Delete
- Project Settings modal (existing project): Delete Project / Delete Project and Data

The same existing checks were kept:
- deployed project with submissions: all required confirmations must gate Delete
- deployed project without submissions: form/recovery confirmations gate Delete
- draft project: draft warning text, no deployed-only checkboxes
- non-survey library item: library warning text, no deployed-only checkboxes

### 👀 Preview steps
1. ℹ️ Have an account and project
2. Select a project from the list
3. Click the delete action button.
4. 🔴 [on main] observe legacy alertify confirmation UX.
5. 🟢 [on PR] observe Mantine/Kobo modal confirmation UX.
6. In the PR modal, verify Delete stays disabled until required checkboxes are selected.
7. Confirm deletion and verify success toast + project removal from list.
8. Go into the project > settings and press the `Delete Project` button
9. 🟢 Confirm it uses the same updated modal with the same behavior as before
10. ℹ️ Have library items
11. Open the library assets list and click on `... (more actions)` > Delete
12. 🟢 Confirm it uses the same updated modal


jsapp/js/components/modalForms/projectSettings.js
```diff
@@ -1138,20 +1143,33 @@ class ProjectSettings extends React.Component {
...
+        content = this.renderStepProjectDetails()
+        break
       default:
         throw new Error(`Unknown step: ${this.state.currentStep}!`)
     }
+
+    return (
+      <>
+        {content}
+        <DeleteAssetPromptHookBridge onReady={this.onDeleteAssetPromptReady} />
```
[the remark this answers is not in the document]
pauloamorimbr: @greptile-apps  The `autoBind(this)` is present in the constructor and working. So it is in fact binding and triggering the action properly.

jsapp/js/components/assetsTable/assetActionButtons.tsx
```diff
@@ -19,7 +19,6 @@ import {
   cloneAsset,
```
magicznyleszek, line 19: The PR title should include DEV-1994 and I think ~~"KoboPrompt"~~ → "Mantine Modal". And I think "delete-asset" should be camelCase.

jsapp/js/assetQuickActions.tsx
```diff
@@ -43,81 +43,26 @@ export function openInFormBuilder(uid: string) {
...
-    actions.resources.deleteAsset(
-      { uid: asset.uid, assetType: asset.asset_type },
-      {
-        onComplete: () => {
-          notify(t('##ASSET_TYPE## deleted permanently').replace('##ASSET_TYPE##', assetTypeLabel))
-          if (typeof callback === 'function') {
-            callback(asset.uid)
-          }
-        },
+  actions.resources.deleteAsset(
```
magicznyleszek, line 52: I wonder… since we are migrating the whole thing, should we try using orval instead of reflux actions?

jsapp/js/hooks/useDeleteAssetPrompt.hook.tsx
```diff
@@ -0,0 +1,174 @@
...
+    closeDeleteAssetPrompt: handleClosePrompt,
+    deleteAssetPrompt,
+  }
+}
+
+interface DeleteAssetPromptHookBridgeProps {
+  onReady: (openDeleteAssetPrompt: OpenDeleteAssetPrompt) => void
+}
+
+export function DeleteAssetPromptHookBridge(props: DeleteAssetPromptHookBridgeProps) {
```
magicznyleszek: The hook flow is kinda unusual. Can't we use `modals.open` instead? The package is already there and as far as I see, it could be used in the places you use the hook and the bridge.

magicznyleszek: Regardless if we keep it as is, or use `modals.open`, I would expect this:
1. Comments in here to describe each exported thing, to make it clear how this should be operated, etc. (especially if "as is").
2. I don't like it living in the `js/hooks` directory. This modal seems like a project/asset related thing, and `js/hooks` feels more like a generic/util directory. Maybe `js/asset/` or `js/project`?
3. We should have used the index way, so a directory `…/DeleteAssetPrompt/`, with `index.ts` inside that re-exports partials. See for example https://github.com/kobotoolbox/kpi/pull/7131

magicznyleszek (changes requested):
Works correctly, have some comments about implementation idea :)

jsapp/js/assetQuickActions.tsx
pauloamorimbr: No time budget to do it on this project! 👎🏻

jsapp/js/hooks/useDeleteAssetPrompt.hook.tsx
pauloamorimbr: Just changed to implement the same patter used in other parts, using `modals.open`. Thank you for the input on this! ⭐

magicznyleszek (approved):

PR #7113, opened 2026-06-03
feat(bulkProcessing): hide feature based on admin setting DEV-2212

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Only users with ASR/MT enabled can see or use bulk transcription and translation in the data table.

### 💭 Notes

This update adds the `asr_mt_features_enabled` environment gate to the bulk-processing UI, so the bulk transcription and translation actions stay hidden unless the feature is enabled for the instance. It also keeps the bulk-actions data hook from polling or exposing bulk-processing state when the feature is off.

Changes here:
- tableColumnSortDropdown.tsx
  - Added the environment gate alongside the existing feature flag check.
  - Bulk transcription / translation menu items now render only when both gates are on.
- useDataTableBulkActions.ts
  - Added the same gate to bulk-action loading and active-job derivation.
  - Prevents bulk-processing banners, polling, and table-side bulk-job state from showing when ASR/MT is disabled.
- useDataTableBulkActions.tests.ts
  - Added coverage for the new env-level gate.

### 👀 Preview steps

To disable ASR/MT go to `/admin` → Constance → Config and set `ASR_MT_INVITEE_USERNAMES` to empty. To enable it add usernames or `*` wildcard.

1. Open a KPI instance where ASR/MT is disabled in the environment.
2. Go to a project with audio questions and submissions.
3. Open the column menu on an audio question or transcript column.
4. Notice that bulk transcription / translation actions do not appear.
5. Enable ASR/MT for the instance.
6. Reload the data table and open the same column menu again.
7. Notice that the bulk actions are available again, and bulk-processing state can load normally.

### DIFF
--- a/jsapp/js/components/submissions/tableColumnSortDropdown.tsx
+++ b/jsapp/js/components/submissions/tableColumnSortDropdown.tsx
@@ -10,6 +10,7 @@ import { PERMISSIONS_CODENAMES } from '#/components/permissions/permConstants'
 import { userCan } from '#/components/permissions/utils'
 import { SortValues } from '#/components/submissions/tableConstants'
 import type { AssetResponse } from '#/dataInterface'
+import envStore from '#/envStore'
 import { FeatureFlag, useFeatureFlag } from '#/featureFlags'
 
 interface TableColumnSortDropdownProps {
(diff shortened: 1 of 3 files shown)


magicznyleszek:
@greptile-apps please do code review

magicznyleszek:
Note to self @magicznyleszek 
> I can't reproduce the expected behaviour. The dropdown (option) is still available
> I did the following in constance:
> - Allow no users in ASR
> - Allow one user in ASR
> - I also removed the GS_BUCKET from dev.py and I still see the dropdown option. I think I'm just not "turning off ASR" correctly, what did you do to test this without making a new kobo-install instance perhaps :D

Edit: Updated PR description (ASR_MT_INVITEE_USERNAMES)

duvld (approved):
LGTM

PR #7119, opened 2026-06-04
fix(admin): fix timeout when viewing tokens DEV-2149

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Fix a timeout error when attempting to view individual user tokens in Django admin.

### 💭 Notes
The built in TokenAdmin has `user` as an editable field, which requires loading all users to create the selection dropdown. The PR makes the `user` field read-only except on creation. 
Since the original TokenAdmin is registered in DRF app code, we have to unregister it before re-registering it with our edited version.

### 👀 Preview steps
Regression test only. Best to test with `--print-sql` running

2. Go to django admin -> Auth Token -> Token
3. Select an arbitrary token to open the editing view
4. 🔴 [on release] notice there is a query that queries the entire auth_user table (using a cursor)
5. :red_circle: [on release] notice you are able to change the user
6. 🟢 [on PR] notice all queries to auth_user are filtered
7. Delete the token
8. Create a new token for the same user
9. 🟢 Ensure new token is created successfully

### DIFF
--- a/kpi/admin.py
+++ b/kpi/admin.py
@@ -63,11 +63,13 @@ class ExtraProjectMetadataFieldAdmin(admin.ModelAdmin):
 
 
 class TokenAdmin(DRFTokenAdmin):
+    autocomplete_fields = ['user']
     def get_readonly_fields(self, request, obj=...):
         if obj is None:
             return ()
         else:
             return ('user',)
+
 
 # Register your models here.
 admin.site.register(AuthorizedApplication)
--- a/kpi/admin.py
+++ b/kpi/admin.py
@@ -65,7 +65,7 @@ class ExtraProjectMetadataFieldAdmin(admin.ModelAdmin):
 class TokenAdmin(DRFTokenAdmin):
     def get_readonly_fields(self, request, obj=...):
         if obj is None:
-            return ('',)
+            return ()
         else:
             return ('user',)
 
--- a/kpi/admin.py
+++ b/kpi/admin.py
@@ -1,10 +1,13 @@
 from django.contrib import admin
 from django.utils.translation import gettext_lazy as _
+from rest_framework.authtoken.admin import TokenAdmin as DRFTokenAdmin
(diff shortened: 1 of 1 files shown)


kpi/admin.py
```diff
@@ -58,5 +61,15 @@ def label_display(self, obj):
 
     label_display.short_description = _('Default Label')
 
+
+class TokenAdmin(DRFTokenAdmin):
+    def get_readonly_fields(self, request, obj=...):
+        if obj is None:
+            return ('',)
+        else:
+            return ('user',)
```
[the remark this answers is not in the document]
rgraber: Local testing suggests this is not the case (the Add form worked perfectly well) but an empty tuple does make more sense

kpi/admin.py
```diff
@@ -58,5 +61,15 @@ def label_display(self, obj):
...
 
+
+class TokenAdmin(DRFTokenAdmin):
+    def get_readonly_fields(self, request, obj=...):
+        if obj is None:
+            return ()
+        else:
+            return ('user',)
+
 # Register your models here.
```
noliveleger, line 76: Missing one blank line ;-)

kpi/admin.py
```diff
@@ -58,5 +61,15 @@ def label_display(self, obj):
 
     label_display.short_description = _('Default Label')
 
+
+class TokenAdmin(DRFTokenAdmin):
+    def get_readonly_fields(self, request, obj=...):
```
noliveleger, line 69: I would add

```python
    autocomplete_fields = ['user']
```

to avoid receiving a 504 when a superuser creates a token for a user. It uses lazy loading to display the list of users.

Otherwise, we remove entirely the "add" button. Users would be forced to login and recreate one themselves.

noliveleger (changes requested):

noliveleger (approved):
LGTM

PR #7122, opened 2026-06-05
feat(auditLogs): util for getting max lookback days DEV-1466

rgraber:
### 🗒️ Checklist

1. [ ] run linter locally

### 💭 Notes
Dev-only change to add a util that will calculate the maximum number of days a user can look back for audit logs according to their subscription. 

On stripe-enabled instances, this will be part of the subscription plans. On non-stripe instances, the maximum lookback is however long we retain audit logs for.

### 👀 Preview steps
Make sure to have stripe enabled and sync the most recent test Stripe data

1. ℹ️ have an account on the community/default plan
2. Open a django shell
3. 
```
from kobo.apps.audit_log.utils import get_max_lookback_days
u = User.objects.get(username='<username>')
get_max_lookback_days(u)
```
4. 🟢 [on PR] should return 60
5. Upgrade your user's subscription to the Professional plan (make sure to resync using sync_subscriber so the local db gets updated)
7. ```get_max_lookback_days(u)```
8. 🟢 should now return 90

### DIFF
--- a/kobo/apps/audit_log/utils.py
+++ b/kobo/apps/audit_log/utils.py
@@ -31,8 +31,4 @@ def get_max_lookback_days(user, **kwargs) -> int:
     limit = get_organization_subscription_limit(
         organization=user_org, usage_type=UsageType.LOG_LOOKBACK_DAYS
     )
-    # this should only happen if the default subscription is missing a log lookback
-    # limit for some reason, just default to the minimum lifespan of a log
-    if limit == inf:
-        return min(settings.ACCESS_LOG_LIFESPAN, settings.PROJECT_HISTORY_LOG_LIFESPAN)
     return int(limit)
--- a/kobo/apps/audit_log/utils.py
+++ b/kobo/apps/audit_log/utils.py
@@ -34,8 +34,4 @@ def get_max_lookback_days(user, **kwargs) -> int:
     # this really shouldn't happen, but handle it just in case
     if limit == inf:
         return max(settings.ACCESS_LOG_LIFESPAN, settings.PROJECT_HISTORY_LOG_LIFESPAN)
-    return int(
-        get_organization_subscription_limit(
-            organization=user_org, usage_type=UsageType.LOG_LOOKBACK_DAYS
-        )
-    )
+    return int(limit)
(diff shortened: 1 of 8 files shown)


rgraber:
@greptileai

kobo/apps/audit_log/utils.py
```diff
@@ -11,3 +17,13 @@ class SubmissionUpdate:
...
+
+
+@requires_stripe
+def get_max_lookback_days(user) -> int:
+    user_org = user.organization
+    return int(
+        get_organization_subscription_limit(
+            organization=user_org, usage_type=UsageType.LOG_ACCESS_DAYS
+        )
+    )
```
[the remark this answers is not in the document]
rgraber: This is normal with @stripe_enabled and will be handled with the caller

noliveleger:
@rgraber it does not return 30 but 60 in my case.

kobo/apps/audit_log/utils.py
```diff
@@ -11,3 +21,17 @@ class SubmissionUpdate:
...
+    if is_user_anonymous(user):
+        return 0
+    user_org = user.organization
+    limit = get_organization_subscription_limit(
+        organization=user_org, usage_type=UsageType.LOG_LOOKBACK_DAYS
+    )
+    # this really shouldn't happen, but handle it just in case
+    if limit == inf:
+        return max(settings.ACCESS_LOG_LIFESPAN, settings.PROJECT_HISTORY_LOG_LIFESPAN)
+    return int(limit)
```
noliveleger: The comment says: 
`# this really shouldn't happen, but handle it just in case`

but in case, if we reach those lines, I think we should not provide the maximum but the minimum log lookback days.

noliveleger (changes requested):

kobo/apps/audit_log/utils.py
rgraber: What if instead we just set the default lookback to use the environment variable for log retention? again this really shouldn't happen, it's only if the community plan is missing a log lookback limit.

noliveleger (approved):
LGTM

PR #7124, opened 2026-06-05
fix(qual): export qual questions correctly DEV-2233

Guitlle:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Fixes the qualitative manual multiple select data export

### 📖 Description
The one hot data export for manual qualitative multiple select data was [fixed on formpack](https://github.com/kobotoolbox/formpack/pull/347). In `import_export_task.py`, the `SubmissionExportTaskBase.get_export_object`  we pass for_output=True down to `get_submissions`  in order to resolve the choice UUIDs into their corresponding human-readable labels. We have also added `_uuid` and `meta/rootUuid` to the additional_fields list in `_get_fields_and_groups()` to retrieve the qualitative payload correctly. 

### 👀 Preview steps
1. Create a project with an active qualSelectMultiple question and ensure there are submissions with multiple tags/options selected (You can achieve this by having an audio question, then clicking on "Open", in the Data view, in the audio cell, and then in the "Analysis" tab you can add a multiple choice question and select a few options)
3. Go to the project's Data > Downloads
4. Select an export format (e.g., XLS) and check the "Single and separate columns" advanced option
5. Generate and download the export
6. Open the export file and verify that the Qual multiple-choice field has successfully expanded into multiple binary columns, and that the single column version displays the selected labels separated by spaces instead of commas

### DIFF
--- a/kpi/models/import_export_task.py
+++ b/kpi/models/import_export_task.py
@@ -44,10 +44,7 @@ from kobo.apps.reports.report_data import build_formpack
 from kobo.apps.storage_backends.base import default_kpi_private_storage
 from kobo.apps.subsequences.exceptions import SupplementMigrationInProgress
 from kobo.apps.subsequences.models import SubmissionSupplement
-from kobo.apps.subsequences.utils.supplement_data import (
-    get_analysis_form_json,
-    stream_with_supplements,
-)
+from kobo.apps.subsequences.utils.supplement_data import get_analysis_form_json
 from kpi.constants import (
     ASSET_TYPE_COLLECTION,
     ASSET_TYPE_EMPTY,
@@ -1079,9 +1076,6 @@ class SubmissionExportTaskBase(ImportExportTask):
                 raise SupplementMigrationInProgress(
                     'Supplement data migration in progress, please retry later.'
                 )
-            submission_stream = stream_with_supplements(
-                source, submission_stream, for_output=True
-            )
 
         pack, submission_stream = build_formpack(
             source, submission_stream, self._fields_from_all_versions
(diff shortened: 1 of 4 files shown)


Guitlle:
We are waiting on merging the formpack fix in order to update our dependency files

kpi/models/import_export_task.py
```diff
@@ -1079,9 +1082,6 @@ def get_export_object(
                 raise SupplementMigrationInProgress(
                     'Supplement data migration in progress, please retry later.'
                 )
```
noliveleger, line 1077: Maybe kind of out of scope for this PR, but since it's part of your optimization, we could move this block right before `submission_stream = source.deployment.get_submissions(` to avoid fetching submissions for nothing.  
What do you think?

kpi/models/import_export_task.py
```diff
@@ -1064,6 +1066,7 @@ def get_export_object(
             fields=fields,
             submission_ids=submission_ids,
             query=query,
+            for_output=True,
```
noliveleger: 👀

noliveleger (changes requested):

kpi/models/import_export_task.py
Guitlle: Ok, sure

kpi/models/import_export_task.py
Guitlle: because this parameter ends up in the the supplement retrieve_data call, which uses this flag for exporting the data

noliveleger (approved):
LGTM

PR #7126, opened 2026-06-05
fix(accounts): avoid IntegrityError on SSO reconnect DEV-2226

noliveleger:
### 📣 Summary

Reconnecting a social (SSO) account failed with a database error when the user's email address was already stored.

### 💭 Notes

- The crash occurred in `update_email` (signal on `social_account_added`) at `primary.set_as_primary()`, which called `save()` on an unsaved `EmailAddress` instance — triggering an INSERT that violated the `unique(user_id, email)` constraint.
- Fix: before calling `set_as_primary()`, if `primary.pk` is `None`, we attempt `EmailAddress.objects.get(user=social_user, email=primary.email)`. If the row already exists, we reuse that instance (with its pk) so `save()` does an UPDATE instead of INSERT.
- The lookup is scoped to `user=social_user`, so two users sharing the same email address cannot interfere with each other.
- Two new tests cover: (1) reconnecting SSO with an already-stored email, (2) isolation between users who share an email.

### 👀 Preview steps

1. ℹ️ Have an account whose email is already stored as an `EmailAddress` (e.g. previously connected via SSO)
2. Connect a social/SSO account from the Account Settings → Security page
3. Disconnect it
4. 🔴 [on main] Reconnecting the SSO account raises a 500 error (`IntegrityError: duplicate key value violates unique constraint`)
5. 🟢 [on PR] Reconnecting the SSO account succeeds without error

### DIFF
--- a/kobo/apps/accounts/signals.py
+++ b/kobo/apps/accounts/signals.py
@@ -23,7 +23,16 @@ def update_email(*args, **kwargs):
     emails, primary = cleanup_email_addresses(request, all_email_addresses)
 
     # cleanup_email_addresses doesn't actually call set_as_primary on the primary
-    # email so do that now
+    # email so do that now.
+    # If primary has no pk, a matching record may already exist in the DB
+    # (e.g. reconnecting an SSO account whose email is already stored).
+    # In that case we must reuse the existing row; otherwise set_as_primary()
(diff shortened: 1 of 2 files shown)


rgraber (approved):
LGTM

PR #7127, opened 2026-06-08
fix(frontend): translate coffeescript files DEV-2252

Akuukis:
### 📣 Summary

Enable some non-translated strings to be translated.

### 💭 Notes

Instead of maintaining a manually curated `i18nMissingStrings.js` let's just scan Coffeescript files for translation strings.

### 👀 Preview steps

1. ℹ️ run `generate_locale.sh` from base
2. ℹ️ `git add locale/en/LC_MESSAGES/djangojs.po -f`
3. ℹ️ run `generate_locale.sh` from PR
4. 🟢 see the diff for yourself: it mostly replaces `i18nMissingStrings.js` with the correct coffeescript reference, but also removes and adds a few.


jamesrkiger (approved):
lgtm. Minor note: I found the generated file under locale/en/LC_MESSAGES/, rather than just locale/en/

PR #7136, opened 2026-06-09
fix(map): include start-geopoint in Project → Data → Map UI DEV-2240

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Map now shows submissions that only have `start-geopoint` data. Before this, the map stayed empty if your project collected locations via `start-geopoint` but didn't have any geopoint questions.

### 💭 Notes

Changes here:

**`constants.ts`**
- Added `MAP_DISPLAYABLE_GEOPOINT_TYPES` array that includes both `'geopoint'` and `'start-geopoint'` question types
- Added `isMapDisplayableGeopointType()` type guard function to check if a question type can be shown on the map
- This gives us a single place to define which question types the map should recognize (easier to maintain if we add more geo types later, like `'background-geopoint'`)

**`utils.ts`**
- Updated `findFirstGeopoint()` to use the new type guard instead of only checking for `'geopoint'`
- Now finds either regular geopoint questions or start-geopoint

**`components/map/index.tsx`** (the main map component)
- Constructor now uses `isMapDisplayableGeopointType()` to check if the project has any map-compatible questions
- This fixes the "no geopoint responses" message showing up when start-geopoint data actually exists

**components/map/FormMapWrapper.stories.tsx:**
- New Storybook stories with play tests that verify:
  - Map renders for forms with only start-geopoint
  - Map Settings shows both geopoint and start-geopoint when both exist

**.storybook/preview-head.html**
- Mock `window.t()` before modules load (fixes "t is not defined" errors when components use t() at module scope), as `preview.tsx` `t()` definition wasn't working

**`components/map/MapSettings.tsx`** (the map settings dialog)
- Updated to include start-geopoint questions in the question selector dropdown
- Without this, users couldn't manually switch to start-geopoint even if it had data

**storybookUtils.tsx**
- New `withMinHeightWrapper()` decorator for stories that need vertical space (maps, menus, modals)
- Updated `Menu.stories.tsx`, `FormLanguagesManager.stories.tsx`, and `FormMapWrapper.stories.tsx` to use shared decorator instead of duplicated inline styles

**components/common/modal.tsx** (common Modal component):
- Added `role="dialog"` and `aria-label` to the deprecated Modal component (because maps are tricky enough without fighting the test harness 🗺️)
  - `jsapp/js/components/map/MapSettings.tsx` - Added `role="tab"` and `aria-selected` to tab buttons

The root cause was simple: three different places in the code were checking `type === 'geopoint'` and ignoring `'start-geopoint'`. Now they all call the same function that knows about both types. No more copy-paste checks that can drift apart.

### 👀 Preview steps

I wasn't able to have actual `start-geopoint` responses in submissions (doesn't work in browser, even though it asks for location permissions). Added storybook tests to cover that instead of fighting with emulating collection through mobile device (tried this few months ago and lost tons of time)

**Test with start-geopoint only:**
1. ℹ️ Create a project that doesn't have an explicit geopoint question but does collect start-geopoint metadata
2. Submit a few responses from a mobile device (so start-geopoint gets captured)
3. Go to Project → Data → Map
4. 🔴 [on main] See "This project does not include geographical data" message
5. 🟢 [on PR] Map renders with submission markers at the locations where the form was started

**Test with both types:**
6. ℹ️ Add an explicit geopoint question to the form (like "Where are you?")
7. Submit responses that have both start-geopoint and the explicit geopoint question answered
8. Go to Project → Data → Map and click the Map Settings button (gear icon)
9. Switch to the "geopoint question" tab
10. 🟢 [on PR] See both questions listed: "Where are you?" and "start-geopoint"
11. 🟢 Switch between them to see different marker positions (form start location vs. explicit answer location)


tiritea:
>Before this, the map stayed empty if your project collected locations via start-geopoint but didn't have any geopoint questions.

Are you sure about this? 😕  I thought I had tested this scenario a while back, and recall that start-geopoints were showing on the map, just that you couldnt then flip to an actual geopoint question. Or is it a separate 'filter' such that unless you had an actual explicit geopoint question, the map would consider the submission non-georeferenced to begin with (so you never got the chance to pick another...)

I'm just confused because the submission JSON suggests that the submission was already getting picked up as georeferenced (so why wasnt the map using this location...); ie the _geolocation was correctly populated

magicznyleszek:
@copilot fix biome error

magicznyleszek:
> > Before this, the map stayed empty if your project collected locations via start-geopoint but didn't have any geopoint questions.
> 
> Are you sure about this? 😕 I thought I had tested this scenario a while back, and recall that start-geopoints were showing on the map, just that you couldnt then flip to an actual geopoint question. Or is it a separate 'filter' such that unless you had an actual explicit geopoint question, the map would consider the submission non-georeferenced to begin with (so you never got the chance to pick another...)
> 
> I'm just confused because the submission JSON suggests that the submission was already getting picked up as georeferenced (so why wasnt the map using this location...); ie the _geolocation was correctly populated

You're right that `_geolocation` gets populated correctly in the submission JSON (backend handles that). The issue is that the frontend map component was making its decision before ever looking at the submission data. What was happening:

#### Before the fix (form has only start-geopoint, no regular geopoint):

1. Constructor sets `hasGeoPoint = false` because it only checked for `type === 'geopoint'`.
2. `createDataQuery()` calls `findFirstGeopoint()` which returns `null`. 
3. Map decides "no geo questions exist" and shows the empty blocked map.
4. The map never even tries to fetch or display the submission data.

So the `_geolocation` field was correctly populated in the submission JSON, but the map component quit early after checking the form definition, not the submission data.

#### After the fix:

1. `hasGeoPoint = true` (because frontend code now checks for both geopoint types).
2. `findFirstGeopoint()` finds the `start-geopoint`.
3. Map says "we have geo questions!" and proceeds to display the data.
4. The points show up using `start-geopoint` as the selected question

The map doesn't use `_geolocation` directly anymore (see #6679). It needs to know which specific question field to read from the submission JSON, and it picks that by scanning the form definition for geopoint-type questions. Now it checks two of the types instead of just one.

tiritea:
>The map doesn't use _geolocation directly anymore (see https://github.com/kobotoolbox/kpi/pull/6679). 

That makes sense, thanks for the explanation! So is it possible that _before_ (when we were using _geolocation?) that it might have been picking up start-geopoints? Or is it the case that start-geopoints have **never** been being displayed? Just that a vaguely recall these start-geopoints were showing up in the past...

>It needs to know which specific question field to read from the submission JSON, and it picks that by scanning the form definition for geopoint-type questions.

Perhaps a stupid question but... then what purpose does _geolocation now provide, if its not actually being used to display anything on maps?

magicznyleszek:
> > The map doesn't use _geolocation directly anymore (see #6679).
> 
> That makes sense, thanks for the explanation! So is it possible that _before_ (when we were using _geolocation?) that it might have been picking up start-geopoints? Or is it the case that start-geopoints have **never** been being displayed? Just that a vaguely recall these start-geopoints were showing up in the past...
> 
> > It needs to know which specific question field to read from the submission JSON, and it picks that by scanning the form definition for geopoint-type questions.
> 
> Perhaps a stupid question but... then what purpose does _geolocation now provide, if its not actually being used to display anything on maps?

I think it is possible that start-geopoint worked at some point. We touched map code with multiple PRs recently, upgrading parts of it.

As to _geolocation, AFAIK it is "bad" and we want to stop using it. John wrote something about it here https://github.com/kobotoolbox/kpi/pull/6679#pullrequestreview-3741036332

duvld (approved):
Code LGTM. Tested on storybook but could not get `start-geopoint` working locally

PR #7139, opened 2026-06-09
fix(editableForm): stale closure and cache mutability issues DEV-2260

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Fixed two bugs in Form Builder: event listeners weren't cleaning up on unmount (memory leak), and the react-query cache was exposed to direct mutation from Backbone code.

### 💭 Notes

Changes here:

**jsapp/js/editorMixins/EditableForm.tsx**
- Added `appRef` to hold the SurveyApp instance
  - The cleanup function was closing over `app=undefined` from the initial render, so all the `.off()` calls were no-ops on unmount
  - The ref gives cleanup access to whatever app instance is actually alive when it runs
- Deep clone the asset from react-query before storing it in state
  - The old stores.allAssets code had an explicit warning comment about this: "Form Builder code is already changing the content of the object"
  - After the refactor, we were handing the cache reference straight to Backbone, which meant any push/splice on `survey.availableFiles` would corrupt the cache
  - Restored the clone that was there before
- Updated cleanup to read from the ref
- Store the app in both state and ref when creating it

Joke from AI: The cleanup function was like trying to turn off lights in a room after the door's been closed - it couldn't see in anymore. The ref is the window.

### 👀 Preview steps

The bugs are not really testable/reproducible. We are making the code more robust (to avoid hard to debug bugs in future) based on analysis from greptile and verifying its findings by looking at the code.


jsapp/js/editorMixins/EditableForm.tsx
```diff
@@ -204,11 +205,14 @@ export default function EditableForm(props: EditableFormProps) {
...
       const assetDataCast = assetData as unknown as AssetResponse
+      // Deep clone the react-query cache reference to prevent mutations to survey.availableFiles
+      // from corrupting the cache. Form Builder code mutates asset properties during initialization.
+      const clonedAsset = clonedeep(assetDataCast)
       setState((currentState) => ({
         ...currentState,
         // TODO: storing asset that we already have in `assetQuery` is not nice. I left it like this to avoid requiring
         // too much refactor in here.
-        asset: assetDataCast,
+        asset: clonedAsset,
```
pauloamorimbr: Since it's only used here I think we can call the function in here directly

```suggestion
        asset: clonedeep(assetDataCast),
```

pauloamorimbr (approved):
LGTM!

Left 1 comment. Up to you!

PR #7142, opened 2026-06-10
feat(bulkProcessing): skip already-processed submissions in bulk actions DEV-2239  

rajpatel24:
### 📣 Summary
 Previously, if any submission in a bulk action request had an existing result or an active conflicting job, the entire request was rejected with a 400. This PR changes the behavior so eligible submissions are processed and ineligible ones are silently skipped, with their UUIDs returned in `skipped_uuids` on the POST response. A 400 is only returned if every submission is ineligible.  

### 📖 Description
Previously `_validate_no_existing_results` and `_validate_no_active_bulk_conflicts` raised a 400 if any submission was ineligible. Now, 400 is only raised if all submissions are ineligible.  The POST response includes `skipped_uuids` so callers know which submissions were excluded.                                                                                                                                                 

### 👀 Preview steps

1. ℹ️ have an account and a project with audio submissions
2. Create a bulk action for automatic transcription of audio submissions.
3. Create another bulk action that includes one of the submission UUIDs from the previous bulk action.
4. 🔴 [on main] notice that the second request is rejected entirely.
5. 🟢 [on PR] notice that the request is processed successfully and that `skipped_uuids` contains the UUID of the skipped submission.


RuthTurk (approved):
LGTM

PR #7144, opened 2026-06-10
refactor(deleteAccountBanner): make stories more stable DEV-2264

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 💭 Notes

Fixed flaky Storybook test for DeleteAccountBanner that randomly failed in CI.

Changes here:
- `DeleteAccountBanner.tsx`
  - Swapped out the manual `useEffect` + `fetchGet` pattern for Orval/react-query `useAssetsList` hook (works better with MSW in Storybook)
  - Fixes the race condition where tests would sometimes run before the mock API responded
  - Added proper `isPending` handling so the button's disabled state is consistent
- `DeleteAccountBanner.stories.tsx`
  - Added `play` functions to all 4 stories (Default, UserHasAssets, UserHasNoAssets, UserOwnsMMO)
  - Each one waits for auth AND data loading before running assertions
  - Actually tests the UI behavior (button states, messages) instead of just "did it render?"

AI joke: No more Schrödinger's test – it's either passing or failing, never both! 🐱

The bug: the component fetched user assets async, but tests could run before that finished. Sometimes the mock was fast, sometimes slow. React Query + explicit waits = no more flaky tests.

### 👀 Preview steps

1. Run `npm run test:storybook` locally
2. 🟢 All DeleteAccountBanner stories pass consistently across chromium, firefox, and webkit
3. 🟢 No more random CI failures on this component (try rerunning the successful job?)

### DIFF
--- a/jsapp/js/account/DeleteAccountBanner.stories.tsx
+++ b/jsapp/js/account/DeleteAccountBanner.stories.tsx
@@ -1,5 +1,6 @@
 import type { Meta, StoryObj } from '@storybook/react-webpack5'
 import type { DecoratorFunction } from '@storybook/types'
+import { expect, waitFor, within } from 'storybook/test'
 import { withRouter } from 'storybook-addon-remix-react-router'
 import assetsMock from '#/endpoints/assets.mocks'
 import organizationMock from '#/endpoints/organization.mocks'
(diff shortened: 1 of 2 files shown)


jsapp/js/account/DeleteAccountBanner.stories.tsx
```diff
@@ -22,15 +20,63 @@ const meta: Meta<typeof DeleteAccountBanner> = {
...
+    expect(deleteButton).toBeDisabled()
+  },
+}
+
+export const UserHasAssets: Story = {
+  play: async ({ canvasElement }) => {
+    const canvas = within(canvasElement)
+
+    // Wait for the delete account button to appear
+    const deleteButton = await canvas.findByRole('button', { name: /delete account/i }, { timeout: 5000 })
```
jamesrkiger: These tests are being run synchronously with mocked data and don't have to wait for any actual network responses. Are the tests failing on a timeout? If so, that tells me there's a deeper issue with the tests. If not, then let's not exceed the default timeouts unnecessarily.

This is mostly a matter of principle. This isn't going to break our pipeline or anything, but our testing pipeline across frontend and backend is slow enough already, so I'd prefer we avoid setting any precedents that could make it slower over time.

jamesrkiger (changes requested):
Left one comment, good to merge if you decide to take out the extended timeouts. Otherwise maybe we can hop on a call just so you can help me understand the situation better.

jamesrkiger (approved):
lgtm

PR #7151, opened 2026-06-12
feat(api): add endpoint to return the total length of all audio DEV-2254

rajpatel24:
### 📣 Summary
This PR adds a new endpoint that returns the duration of audio attachments, allowing the front-end to calculate total transcription time without downloading and processing each file.

### 📖 Description
This PR introduces a new endpoint, `POST /api/v2/assets/{asset_uid}/attachments/audio-duration/`, which accepts a list of attachment UIDs and returns the duration of each audio file along with the total duration.

A new `audio_length` field has been added to the `Attachment` model to cache audio durations. When a request is made:

* If `Attachment.audio_length` is already populated, the stored value is returned directly.
* Otherwise, the duration is extracted using `ffprobe`, saved to `audio_length`, and then returned in the response.

This caching mechanism improves performance and avoids repeatedly processing the same files. The endpoint also returns per-attachment durations to support future front-end features that may need to display file-level duration information.

### 👀 Preview steps

1. ℹ️ Have an account and a project containing one or more audio attachments.
2. Send a `POST` request to `/api/v2/assets/{asset_uid}/attachments/audio-duration/` with a list of attachment UIDs.
3. 🔴 **[on main]** The endpoint is not available.
4. 🟢 **[on PR]** The endpoint returns the duration (in seconds) for each requested attachment and the total duration across all attachments.
5. 🟢 Verify that repeated requests return the same durations and benefit from the cached `audio_length` values.
6. 🟢 Verify that the endpoint is documented in the Swagger/OpenAPI documentation.


RuthTurk (approved):
LGTM

PR #7128, opened 2026-06-08
feat(nlp): implement Google API US/EU region routing for Speech-to-Text and Translation DEV-2199

rajpatel24:
### 📣 Summary
This PR implements the server-wide Google ASR/MT location setting to enforce strict EU data residency or utilize global US infrastructure for Speech-to-Text (v2) and Translation (v3) APIs.

### 📖 Description
Following the parallel switch to Google Cloud Speech-to-Text v2 and Translation v3, the previous ASR/MT region control setting (from DEV-1943) was no longer functional. This PR re-establishes the server-wide US/EU location toggle and ensures requests are routed to the correct regional or global endpoints based on that configuration.

#### Speech-to-Text (ASR) Updates:
- The Google location is dynamically set to `us` or `eu` depending on the server-wide `ASR_MT_GOOGLE_REGION` setting.

#### Translation (MT) Updates:
- EU traffic is explicitly routed through the dedicated `translate-eu.googleapis.com` multi-regional gateway and processed in `europe-west1`. This guarantees that TLS termination, processing, and data-in-transit remain entirely within European borders.
- US traffic is routed through `translate-us.googleapis.com` + `us-west1`.

#### Constance setting:
- Consolidated the legacy `ASR_MT_GOOGLE_TRANSLATION_LOCATION` setting (which tracked granular locations like `us-central1` or `europe-west1`) into a new `ASR_MT_GOOGLE_REGION` Constance setting that provides a straightforward US or EU toggle.

#### Region code fallback fix (pre-existing bug):

- `TranscriptionService.get_language_code()` and `get_configuration()` previously raised `LanguageNotSupported` for any region code not explicitly listed in the database (e.g. `fr-BE`, `de-AT`). The check `language_set.filter(code='fr-BE')` always returned false because `fr-BE` is a region code, not a language code.
- Fixed: both methods now extract the parent language code (`fr` from `fr-BE`) as a fallback. If the parent language is supported, the original region code is passed through to Google STT unchanged, allowing it to apply the best available model for that regional variant.

### 👀 Preview steps
1. ℹ️ Have a KoboToolbox account with a project that has audio submissions
2. Set the `ASR_MT_GOOGLE_REGION` constance value to `US` (Django admin → Constance → Config)

**Spreadsheet**:
You can download the updated CSV from here: [kobo-import.csv](https://github.com/user-attachments/files/28942188/kobo-import.csv)
[...]

PR #7154, opened 2026-06-15
feat(nlp): ensure maximum language support with per-language STT routing and GLOBAL/EU toggle DEV-2277

rajpatel24:
### 📣 Summary
Extends the Google ASR/MT location handling from [kpi#7128](https://github.com/kobotoolbox/kpi/pull/7128) to support every language Google offers. Instead of a fixed `us`/`eu` endpoint for all languages, the code now routes each language to whichever Google endpoint hosts its best available model. A new `GLOBAL`/`EU` Constance toggle replaces the old `US`/`EU` toggle: `GLOBAL` (default) uses per-language routing for maximum coverage; `EU` restricts all processing to EU-hosted endpoints for data residency compliance.

### 📖 Description
[kpi#7128](https://github.com/kobotoolbox/kpi/pull/7128) re-implemented the server-wide Google ASR/MT region setting  (`ASR_MT_GOOGLE_REGION`) and simplified routing to use a single fixed  location: `us` for all languages on non-EU servers and `eu` on EU servers.

However, this caused 24 languages to lose ASR support, languages like Irish,  Belarusian, Bosnian, and Somali that only have `chirp_2` model support in specific Google sub-regions (`us-central1`, `europe-west4`) and no model at  all in the `us`/`eu` multi-region endpoints.

Because we cannot lose support for these 24 languages, we decided to implement the following approach instead:
1. Global mode (default): Support every language that Google supports by routing each language to the endpoint that provides the best available model.
2. EU mode: Use the best available model within EU-hosted endpoints only. Languages that are not available in any EU endpoint are considered unsupported on EU servers.

In `GLOBAL` mode, the code reads `location_code` per language from the  database (populated by the language spreadsheet, which stores the correct  global location for each model). In EU mode, the code overrides the DB  location at runtime using the table above via `EU_LOCATION_BY_MODEL` in `locations.py`.

### 👀 Preview steps

1. ℹ️ Have an account with a project containing audio submissions.
2. Ensure `ASR_MT_GOOGLE_REGION` is set to `GLOBAL` in Django admin → Constance → Config.

Test GLOBAL mode with `chirp_3` language (e.g. French):
3. 🟢 Transcribe a French audio submission → should succeed.

Test GLOBAL mode with `chirp_2` language (e.g. Irish):
4. Transcribe an Irish audio submission → should succeed.

Test EU mode:
5. Change `ASR_MT_GOOGLE_REGION` to `EU`.
6. 🟢 Transcribe a French submission → should succeed using location `eu`.
7. 🟢 Transcribe an Irish submission → should succeed using location `europe-west4`.

NOTE: Import this updated language list through the Django admin [kobo-import.csv](https://github.com/user-attachments/files/28955677/kobo-import.csv)

### DIFF
--- a/kobo/apps/subsequences/integrations/google/google_transcribe.py
+++ b/kobo/apps/subsequences/integrations/google/google_transcribe.py
@@ -40,7 +40,7 @@ from ...exceptions import (
     TranscriptionResultNotFound
 )
 from .base import GoogleService
-from .locations import get_speech_location
+from .locations import get_speech_location, get_speech_location_for_model
 
 # https://cloud.google.com/speech-to-text/docs/quotas
 ASYNC_MAX_LENGTH = timedelta(minutes=479)
@@ -64,7 +64,6 @@ class GoogleTranscriptionService(GoogleService):
         class. It uses Google Cloud Speech-to-Text v2 batch API.
         """
         super().__init__(submission=submission, asset=asset, *args, **kwargs)
-        self.speech_location = get_speech_location()
 
     def adapt_response(self, response: Union[dict, list]) -> str:
         """
@@ -110,6 +109,7 @@ class GoogleTranscriptionService(GoogleService):
         content: Any,
         *,
         model_code: str | None = None,
+        speech_location: str | None = None,
     ) -> tuple[object, int]:
         """
         Set up a batch transcription operation
@@ -122,19 +122,19 @@ class GoogleTranscriptionService(GoogleService):
             )
 
         speech_model = model_code or DEFAULT_SPEECH_MODEL
-        speech_client = self._get_speech_client(self.speech_location)
+        speech_client = self._get_speech_client(speech_location)
         input_path, output_prefix = self._get_batch_paths(xpath, source_lang)
 
         logging.info(
             'Starting Google automatic transcription for '
             f'{self.submission_root_uuid=}, {xpath=}, {source_lang=}, '
-            f'{self.speech_location=}, {speech_model=}'
+            f'{speech_location=}, {speech_model=}'
         )
         self._cleanup_batch_files(xpath, source_lang)
         gcs_input_uri = self.store_file(flac_content, input_path)
 
         request = speech.BatchRecognizeRequest(
-            recognizer=self._get_recognizer_name(self.speech_location),
+            recognizer=self._get_recognizer_name(speech_location),
             config=speech.RecognitionConfig(
                 auto_decoding_config=speech.AutoDetectDecodingConfig(),
                 language_codes=[source_lang],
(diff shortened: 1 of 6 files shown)


kobo/apps/subsequences/integrations/google/google_transcribe.py
```diff
@@ -213,6 +208,12 @@ def process_data(
             return {'status': 'failed', 'error': message}
 
         source_language = language_config.language_code
+        speech_location = (
+            get_speech_location_for_model(language_config.model_code)
+            or language_config.location_code
+            or get_speech_location()
+        )
```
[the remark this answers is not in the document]
rajpatel24: I don't think this is required.

kobo/apps/subsequences/migrations/0013_rename_google_region_us_to_global.py
```diff
@@ -0,0 +1,75 @@
...
+        ('subsequences', '0012_migrate_google_region_constance_key'),
+        ('constance', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            rename_us_to_global,
+            reverse_code=rename_global_to_us,
+        ),
+    ]
```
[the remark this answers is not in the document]
rajpatel24: I don't think this is required.

kobo/apps/subsequences/integrations/google/locations.py
```diff
@@ -56,8 +56,18 @@ def get_google_region() -> str:
...
+    in which case the caller must use location_code from the database
+    """
+    if get_google_region() == GOOGLE_REGION_EU:
+        return EU_LOCATION_BY_MODEL.get(model_code, 'eu')
+    return None
+
+
 def get_speech_location() -> str:
-    return SPEECH_LOCATION_BY_REGION[get_google_region()]
+    return 'eu' if get_google_region() == GOOGLE_REGION_EU else 'us'
```
noliveleger, line 70: I did not notice the first time but `get_google_region()` should return a lower string like any other slug, i.e.  `GOOGLE_REGION_EU` and `GOOGLE_REGION_GLOBAL`  should be a lower case string.

kobo/apps/subsequences/integrations/google/locations.py
```diff
@@ -56,8 +56,18 @@ def get_google_region() -> str:
...
+    """
+    Return the EU speech location for the given model, or None in GLOBAL mode,
+    in which case the caller must use location_code from the database
+    """
+    if get_google_region() == GOOGLE_REGION_EU:
+        return EU_LOCATION_BY_MODEL.get(model_code, 'eu')
+    return None
+
+
 def get_speech_location() -> str:
```
noliveleger: I think I would make this name more obvious, e.g. : `get_speech_location_for_region`

kobo/apps/subsequences/migrations/0013_rename_google_region_us_to_global.py
```diff
@@ -0,0 +1,75 @@
...
+        return ''
+    try:
+        return str(json.loads(raw))
+    except (json.JSONDecodeError, TypeError):
+        return str(raw)
+
+
+def rename_us_to_global(apps, schema_editor):
+    """
+    Rename the legacy constance value 'US' to 'GLOBAL'.
```
noliveleger: Same comments for slugs. Let's use lower strings. (i.e. `global`)

noliveleger (changes requested):

noliveleger (approved):
LGTM

PR #7158, opened 2026-06-16
fix(formbuilder): duplicate list name DEV-2269

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Fixed a bug where adding the same select question from the Library multiple times made all copies share one choice list, so editing one changed all the others.

### 💭 Notes

Changes here:
- **jsapp/xlform/src/model.survey.coffee**
  - `_ensure_row_list_is_copied()` - Now generates a unique `list_name` for each library select question using `txtid()`, same as the Duplicate button does. This is the fix.
  - `insertSurvey()` - Added comments explaining how it handles both single questions and groups from the library
  - `insert_row()` - Added comments clarifying this is for the Duplicate button (different code path)
- **jsapp/js/models/surveyScope.tsx**
  - `handleItem()` and `addExternalItemAtPosition()` - Added JSDoc comments documenting the async fetch and insert flow
- **jsapp/js/editorMixins/AssetNavigator.tsx**
  - Added comments explaining how drag-drop from library works and where items can be dropped
- **jsapp/xlform/src/view.surveyApp.coffee**
  - Added comments to both receive handlers (one for main survey, one for groups) explaining why we need two separate sortable instances
- **test/xlform/survey.tests.coffee**
  - Added 2 tests that checks the flow of the bug (one for single question, one for group of questions)
  - Added test that checks a problem introduced while working on this task (select_one questions added from Library were losing the options on save)

The comments trace the full path from dragging a library item to giving it a unique choice list.

### 👀 Preview steps

1. ℹ️ Create a library question (`select_one` type with non-default choices) and library block (a group containting some questions and a `select_one` question)
2. Create or open a form in Formbuilder
3. Click "Add from Library" in the right panel
4. 🟡 [on main] Drag the select question into your form twice
5. Edit the first select: add a new option "Grape"
6. Save the form, then reopen Formbuilder
7. 🔴 [on main] The second select also has "Grape" (they share the same list)
8. Export to XLS and check the choices sheet - both selects reference the same `list_name`
9. ♻️ [on PR] Repeat steps 4-8
10. 🟢 The second select still only has the original 3 options
11. 🟢 Export to XLS - each select has its own unique `list_name`
12. ♻️ Repeat steps 4-8 for a block with select question (there is a different code path for cloning single question and cloning a group of questions)


magicznyleszek:
@greptile-apps please rereview

pauloamorimbr (approved):
LGTM! Nice job! ⭐

PR #7160, opened 2026-06-17
fix(tests): fix failing google transcribe test DEV-2298

rajpatel24:
### 📣 Summary
Fix failing google transcribe test


rajpatel24:
Merged without review due to the small scope of the change and prior internal discussion.

PR #7162, opened 2026-06-17
fix(reports): disable restricted report actions DEV-2292

duvld:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Disables custom report buttons when the user doesn't have the correct permissions.

### 💭 Notes
The issue mentions that the missing check is for `manage_asset` but it should be `change_asset`

### 👀 Preview steps
1. ℹ️ have 2 accounts and a project with submissions
2. give the second account every permission except Edit form and Manage asset
3. on the second account, click these buttons:
    - Override Graph Style
    - Create New Report 
    - Edit Report Questions
5. 🔴 [on main] notice that you can make changes and save the modal and the backend will give an error in a toast
6. 🟢 [on PR] notice that all these buttons are disabled


magicznyleszek (approved):
Biome issuess, but other than that LGTM

PR #7171, opened 2026-06-19
fix(scim): address final review comments DEV-2332

Guitlle:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Addressed various minor comments for ensuring data integrity and appropriate defaults accross the SCIM kobo plugin.

### 📖 Description
This PR centralises SCIM exception handling into a new exceptions.py module with a unified ScimException class and scim_exception_handler, and addresses several data-integrity and audit-log gaps identified in the previous review round in #7141. Replaces the infinite loop and the prior ValueError in generate_unique_scim_username with a bounded loop that raises ScimException on exhaustion, so the centralised handler can return a structured 409 instead of an opaque 500. Introduces ScimExceptionHandlerMixin applied to all SCIM view classes, adds an explicit _create_provisioning_audit_log call in perform_update for reactivation events, adds .distinct() to the SSO-linked reactivation query to avoid duplicate processing, and sets instance.is_active = False after the bulk targets.update(is_active=False) call in perform_destroy to keep the in-memory object in sync with the database.


RuthTurk (approved):
Good to merge after the above comment from greptile is addressed ⬆️

PR #7176, opened 2026-06-23
feat(bulkProcessing): add bulk-accept endpoint for NLP results DEV-2324

rajpatel24:
### 📣 Summary
This PR implements a new POST `/api/v2/assets/{uid_asset}/data/supplements/bulk/` endpoint that accepts transcription or translation results for multiple submissions in a single request.

### 📖 Description
Users previously had to accept ASR/MT results one submission at a time. This PR adds a dedicated bulk-accept endpoint so the UI can approve all selected submissions in a single request. 

 - `BulkAcceptSerializer.accept()` fetches all targeted `SubmissionSupplement` rows in one query, stamps `_dateAccepted` on the latest `_versions[0]` entry in memory, then flushes with `bulk_update`, no per-submission ORM round-trips.                                                                                                
 - Skips submissions with no supplement, missing action data, or a null-value version (in-progress / deleted); those are excluded from `accepted_count`.

### 👀 Preview steps

1. ℹ️ have an account and a project with a few audio submissions
2. run a bulk transcription and translation on those audio files
3. To accept the generated result, send a POST request to `/api/v2/assets/{uid_asset}/data/supplements/bulk/` for both actions
5. 🟢 [on PR] notice that the results are now accepted and visible in the UI data table. Also, the `SubmissionSupplement.content` field has a `dateAccepted` timestamp.
6. Test the endpoint with a single submission, a submission with an NLP action still in progress, and a submission with a null NLP result.
7. 🟢 notice that the endpoint handles all of these cases gracefully.


kobo/apps/subsequences/serializers.py
```diff
@@ -377,6 +379,96 @@ def _get_question_advanced_feature_params(self, params: dict) -> list[dict]:
...
+            version_data = latest_version.get('_data', {})
+            # Only accept versions that have actual content
+            if version_data.get('value') is None:
+                continue
+
+            latest_version['_dateAccepted'] = now_str
+            target['_dateModified'] = now_str
+            to_update.append(supplement)
+
+        if to_update:
```
RuthTurk: Should we wrap this `bulk_update` in a transaction? 
```
with transaction.atomic():
    if to_update:
        SubmissionSupplement.objects.bulk_update(to_update, ['content'])
```

RuthTurk (dismissed):
One small comment

kobo/apps/subsequences/serializers.py
rajpatel24: Ah right!

rajpatel24:
@greptileai re-review

rajpatel24:
@greptileai I intentionally omitted `select_for_update()` here because I believe the risk it mitigates is largely theoretical, while the locking it introduces could have real operational costs for our NLP pipeline.

This code path is triggered by a user action from the UI. Concurrent execution of the same action by the same user is highly unlikely. Even if two different users were to trigger it simultaneously, the worst-case outcome is that `_dateAccepted` is updated with slightly different timestamps. The end state remains correct, with the latest timestamp winning and no data corruption occurring.

By contrast, adding `select_for_update()` would lock all affected `SubmissionSupplement` rows (potentially 100+) for the duration of the transaction. That could block other legitimate concurrent operations on those rows, including NLP processing tasks that may need to update supplement data.

Given the low likelihood and limited impact of the race condition versus the potential performance and contention costs of row-level locking, I opted not to introduce `select_for_update()` in this case.

Guitlle (approved):
LGTM

PR #7182, opened 2026-06-24
feat(subsequences): add `pendingReview` flag to `_supplementalDetails` for unaccepted NLP results DEV-2333

rajpatel24:
### 📣 Summary
Updates the data output transformation for the `/api/v2/assets/{asset-uid}/data` endpoint to surface a `pendingReview: true` flag in `_supplementalDetails` when a transcription or translation is completed but awaiting manual acceptance.

### 📖 Description
Previously, the endpoint only returned data if it had a `_dateAccepted` timestamp. Because of this, the frontend couldn't detect when an automatic transcription was finished and ready for review.

Additionally, this fixes a precedence issue: if a user generated a new auto-transcription on top of an older, accepted manual transcription, the API would still serve the old manual version. Now, the newest action takes precedence, accurately surfacing the pending review status so the UI can display the "Review" button.

What changed:
- Updated `transform_data_for_output` within the transcription and translation mixins.
- The data endpoint now evaluates the most recently created version of an action, rather than just the most recently accepted one.
- If the latest version lacks a `_dateAccepted`, the API returns `pendingReview: True` and omits the text value.
- Retained the check to ensure that if the latest version is deleted (where value is explicitly None), it correctly returns an empty object.

### 👀 Preview steps

1. ℹ️ Have an account and a project with a few audio submissions.
2. Perform an automatic transcription of an audio file and do not accept the transcription result.
3. 🔴 [On main] Notice that `_supplementalDetails` in the `/api/v2/{asset_uid}/data/` endpoint returns an empty object (`{}`).
4. 🟢 [On PR] Notice that `_supplementalDetails` in the `/api/v2/{asset_uid}/data/` endpoint now returns `pendingReview: true`.
5. Perform a manual transcription, then use the API to run an automatic transcription on the same audio.
6. 🟢 Notice that `_supplementalDetails`, which previously contained the manual transcription data, now contains the unaccepted automatic transcription data with `pendingReview: true`.
7. 🟢 Notice that `_supplementalDetails` remains empty when the automatic transcription is still in progress or when the transcription has been deleted.
8. Repeat the same steps to test translations.


kobo/apps/subsequences/actions/mixins.py
```diff
@@ -119,29 +119,44 @@ def col_type(self):
     def transform_data_for_output(
         self, action_data: dict
     ) -> SimplifiedOutputCandidatesByColumnKey:
```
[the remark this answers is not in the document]
rajpatel24: Already added.

kobo/apps/subsequences/actions/mixins.py
```diff
@@ -119,29 +119,44 @@ def col_type(self):
...
-        # they should already be in order but there's no way to guarantee it, so
-        # sort just in case
+        if not versions:
+            return {}
+
+        # Sort by _dateCreated (newest first) so we always evaluate the latest
+        # version, whether it is accepted or still pending review
         versions_sorted = sorted(
             versions,
-            key=lambda x: self._get_date_field_value(x),
```
noliveleger, line 128: I think `_get_date_field_value()` is dead code now

noliveleger (approved):
LGTM 

Just validate `_get_date_field_value()` is really dead code or not.

kobo/apps/subsequences/actions/mixins.py
rajpatel24: You're right! It's dead code. I'll remove it

PR #7184, opened 2026-06-25
fix(bulkProcessing): fix bulk action re-processing after deletion and language switching DEV-2346 DEV-2338

rajpatel24:
### 📣 Summary
This PR fixes two bugs in the bulk transcription and translation feature that prevented users from re-processing audio submissions after they had already interacted with them.

### 📖 Description
(DEV-2346): When a user switched the interface language and then tried to run a bulk transcription or translation on their submissions, the entire batch would fail with an error. This happened because the language-switching caused an internal mismatch that broke the bulk processing request before it even started.

(DEV-2338): When a user transcribed a submission, then deleted that transcription (via the "Discard" button in the Single Processing view), and then tried to transcribe the same submission again using the same language, they would get a "400 - All submissions are already processed or currently being processed" error. The system was not correctly recognising that the previous transcription had been deleted and was treating the submission as if it was still processed.

What was fixed?                                                                                                                                                                                     Both bugs have been resolved so that users can now:                                                                                                                               
- Switch languages freely without disrupting bulk processing requests.                                                                                                            
- Delete a transcription and immediately re-transcribe the same submission in the same language without errors.

### 👀 Preview steps

1. ℹ️ Have an account and a project with a few audio submissions.
2. Enable bulk actions in the UI using the `bulkProcessingEnabled` feature flag.
3. Select an audio submission and click "Transcribe selected audio files" to generate a transcription. Use the "Open" button to access the Single Processing view, then click "Discard" on the generated transcription. Finally, click "Done" to return to the Data Table.
4. Request a bulk transcription again for the same submission.
5. 🔴 [On main] Notice that the request fails with a `400` error and the notification "All submissions are already processed or currently being processed." appears.
6. 🟢 [On PR] Notice that the bulk transcription proceeds successfully.
7. Repeat steps 3 and 4, but select a different language for the second transcription request.
8. 🔴 [On main] Notice that the transcription fails (visible in the GET API response).
9. 🟢 [On PR] Notice that the transcription is successfully generated in the requested language.

### DIFF
--- a/kobo/apps/subsequences/exceptions.py
+++ b/kobo/apps/subsequences/exceptions.py
@@ -108,7 +108,14 @@ class SupplementMigrationInProgress(Exception):
 
 
 class TranscriptionNotFound(DependencyNotFound):
-    pass
+    def __init__(
+        self,
+        message=(
+            'No accepted transcription found. '
+            'Accept the transcription before running translation.'
+        )
+    ):
+        super().__init__(message)
 
 
 class TranslationAsyncResultAvailable(Exception):
--- a/kobo/apps/subsequences/serializers.py
+++ b/kobo/apps/subsequences/serializers.py
@@ -365,8 +365,9 @@ class BulkActionCreateSerializer(serializers.Serializer):
             return
 
         action = feature.to_action()
+        params_before_update = list(feature.params)
         action.update_params(feature_params)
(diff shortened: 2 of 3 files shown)


RuthTurk (approved):
LGTM

PR #7195, opened 2026-06-29
feat(qualitativeAnalysis): add `QATagTracker` model for QA tag autocomplete DEV-1469

rajpatel24:
### 📣 Summary
This PR introduces the `QATagTracker` model to support autocomplete suggestions for QA tag-type questions.

### 📖 Description
This PR adds the `QATagTracker` model to the subsequences app as the database foundation for the autocomplete-for-QA-tag-questions feature. Each row represents a unique tag value used for a specific QA question on a specific asset. 

The model includes a foreign key to `Asset`, a `question_uuid` field identifying the QA question (stored as a plain char field since QA questions are not persisted as database objects), and a value field holding the tag string. A unique constraint on (asset, question_uuid, value) ensures no duplicate tags are stored per question.

### 👀 Preview steps
It's just a model, so no preview steps are needed.


kobo/apps/subsequences/models.py
```diff
@@ -287,6 +287,29 @@ def to_action(self, prefetched_dependencies: dict | None = None) -> Action:
         )
 
 
+class QATagTracker(models.Model):
+    asset = models.ForeignKey(
+        'kpi.Asset',
+        related_name='qa_tag_trackers',
+        on_delete=models.CASCADE,
+    )
+    question_uuid = models.CharField(max_length=36, db_index=True)
```
noliveleger: I'm not sure we need `db_index=True` here since we are using the composite index. 
Are we searching for tag without asset context?  
What do you think?

noliveleger (changes requested):

kobo/apps/subsequences/models.py
rajpatel24: I don't think we'd ever search for a tag without an asset context, so you're right! We don't need `db_index=True` since the composite index already covers that. I'll remove it.

noliveleger (approved):
LGTM

PR #7198, opened 2026-06-29
refactor(formMap): hide map tiles to avoid uncontrollable chromatic test failures

magicznyleszek:
### 💭 Notes

The map Storybook tests kept failing Chromatic checks because OpenStreetMap tiles would render with tiny pixel differences between runs. A few dozen pixels would shift around, causing false positive failures.

Changes here:
- `FormMapWrapper.stories.tsx`: Added a decorator that hides the tile layer with CSS (makes `.leaflet-tile-pane` invisible) and replaces it with a solid light blue background. Also added 1000ms delay in Chromatic config so the map has time to initialize before snapshots.

### 👀 Preview steps

1. ℹ️ Open Storybook locally: `npm run storybook`
2. Navigate to Features → FormMap → WithOnlyStartGeopoint or WithBothGeopointTypes
3. 🟢 Notice the map background is now solid light blue instead of showing OpenStreetMap tiles
4. 🟢 Confirm all UI elements still work: settings button, layer toggle, fullscreen, markers, legend

### DIFF
--- a/jsapp/js/components/map/FormMapWrapper.stories.tsx
+++ b/jsapp/js/components/map/FormMapWrapper.stories.tsx
@@ -142,8 +142,32 @@ const meta: Meta<typeof FormMapWrapper> = {
       routing: { path: ROUTES.FORM_DATA },
     }),
     a11y: { test: 'todo' },
+    chromatic: {
+      delay: 1000,
+    },
   },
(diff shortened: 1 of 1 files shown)


jamesrkiger (approved):
I think Greptile's suggestion is probably a better approach here, but I'll leave it up to you to decide.

PR #7199, opened 2026-06-29
feat(qualitativeAnalysis): track QA tags on supplement save DEV-1470

rajpatel24:
### 📣 Summary
This PR implements tag tracking for `qualTags` type QA questions by creating `QATagTracker` rows whenever tags are saved via the supplement PATCH endpoint. 

### 📖 Description
This PR adds the logic to populate `QATagTracker` on every successful supplement save. When the `manual_qual` action processes a `qualTags` question, `SubmissionSupplement._sync_qual_tag_trackers()` is called after the supplement data is committed. It looks up the question type from the `QuestionAdvancedFeature` params, and if the type is `qualTags`, it bulk-creates tracker rows one per tag value using `ignore_conflicts=True` so duplicate tags across submissions are silently skipped without any additional query overhead.

### 👀 Preview steps

1. ℹ️ Have an account and a project with audio submissions.
2. Perform a transcription or translation of an audio submission.
3. Go to the 'Analysis' tab, add a 'Tags' question, and then add a tag to that question.
4. 🔴 [On main] Notice that no entry is created anywhere.
5. 🟢 [On this PR] Save the tag and verify that an entry is created in `QATagTracker`.
6. Try adding the same tag again, either from the same submission or from another submission for the same question.
7. 🟢 Verify that no duplicate entry is created in `QATagTracker`.


rgraber (commented):
This is a broader question, but do we need to keep track of the question xpath as well? We could have multiple `manual_qual` QAFs on a single asset if they correspond to different xpaths, so just keeping track of the QA question xpath may not be enough.

rajpatel24:
> This is a broader question, but do we need to keep track of the question xpath as well? We could have multiple `manual_qual` QAFs on a single asset if they correspond to different xpaths, so just keeping track of the QA question xpath may not be enough.

@rgraber From what I understood from your question, I don't think we need to track the xpath separately. We're already using `question_uuid`, which is more specific than xpath. It uniquely identifies an individual `qualTags` question regardless of which xpath's QAF it lives under, since each qual question gets its own UUID when it's created (independent of the xpath it's attached to). So two `qualTags` questions on different xpaths would naturally have different `question_uuids` already. Let me know if I'm missing something here!

rgraber:
> @rgraber From what I understood from your question, I don't think we need to track the xpath separately. We're already using `question_uuid`, which is more specific than xpath. It uniquely identifies an individual `qualTags` question regardless of which xpath's QAF it lives under, since each qual question gets its own UUID when it's created (independent of the xpath it's attached to). So two `qualTags` questions on different xpaths would naturally have different `question_uuids` already. Let me know if I'm missing something here!

Are we ever going to have to look up a QAF object from a QATagTracker object? Because then we'd have to query on the `params` list which might be funky.

rajpatel24:
> > @rgraber From what I understood from your question, I don't think we need to track the xpath separately. We're already using `question_uuid`, which is more specific than xpath. It uniquely identifies an individual `qualTags` question regardless of which xpath's QAF it lives under, since each qual question gets its own UUID when it's created (independent of the xpath it's attached to). So two `qualTags` questions on different xpaths would naturally have different `question_uuids` already. Let me know if I'm missing something here!
> 
> Are we ever going to have to look up a QAF object from a QATagTracker object? Because then we'd have to query on the `params` list which might be funky.

@rgraber I don't think we'll need the reverse lookup. All of our current flows go from QAF to tracker. We create trackers from the params during a supplement save, and we read them back by `asset` + `question_uuid` for autocomplete.

Even for cleanup on deletion, if we add it in the future, we'd still be starting from the QAF side. We'd already have the UUIDs of the removed questions, so we could simply do:

`QATagTracker.objects.filter(asset=asset, question_uuid__in=removed_uuids).delete()`

So the only time we'd need the JSONB-on-params query is if we started with an orphan tracker and needed to find its owning QAF, which none of our current flows require. There's also no requirement for tracker deletion or validation at the moment.

rgraber (approved):
LGTM

PR #7208, opened 2026-07-01
fix(bulkProcessing): skip already-accepted versions in bulk-accept DEV-2385

rajpatel24:
### 📣 Summary
The bulk-accept endpoint (`POST /api/v2/assets/{uid_asset}/data/supplements/bulk/`) was re-stamping `_dateAccepted` and counting already-accepted submissions on every call, making `accepted_count` misleading. Added a guard in `BulkAcceptSerializer.accept()` to skip versions where `_dateAccepted` is already set.

### 📖 Description
Calling the bulk-accept endpoint multiple times for the same submission always returned `accepted_count: 1` (or more), even though no new acceptance action was performed. The caller had no way to distinguish "just accepted" from "already accepted".

This PR adds an early continue in the per-supplement loop when latest_version.get(`_dateAccepted`) is already truthy. Already-accepted versions are now silently skipped `_dateAccepted` is not modified and the supplement is excluded from `to_update`, so `accepted_count` correctly reflects only newly accepted records.

### 👀 Preview steps

1. ℹ️ Have an account with a project that contains audio submissions.
2. Create a transcription for an audio submission, but do not accept it from the UI. Instead, use the `POST /api/v2/assets/{uid_asset}/data/supplements/bulk/` endpoint from `/api/v2/docs` to accept it. Verify that the response contains `accepted_count: 1`, and confirm in the UI that the transcription is marked as accepted.
3. Send the same API request again using the exact same payload.
4. 🔴 On main: Notice that the API still returns `accepted_count: 1`.
5. 🟢 On this PR: Notice that the API now returns `accepted_count: 0` and does not update the `dateAccepted` timestamp.
6. Repeat the same steps for translations.

### DIFF
--- a/kobo/apps/subsequences/serializers.py
+++ b/kobo/apps/subsequences/serializers.py
@@ -473,6 +473,10 @@ class BulkAcceptSerializer(serializers.Serializer):
                 if version_data.get('value') is None:
                     continue
 
+                # Skip versions that are already accepted
+                if latest_version.get('_dateAccepted'):
+                    continue
+
                 latest_version['_dateAccepted'] = now_str
                 target['_dateModified'] = now_str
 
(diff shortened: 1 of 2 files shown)


RuthTurk (approved):
LGTM

PR #7210, opened 2026-07-01
fix(maps): overflow for lists in map disaggregation dropdown DEV-2388

jamesrkiger:
### 📣 Summary
Adds scrollable overflow for map disaggregation dropdown so long lists of questions can be displayed.

### 👀 Preview steps

1. Upload this form: 
[long-list-geopoint.xlsx](https://github.com/user-attachments/files/29561883/long-list-geopoint.xlsx)
2. Make a couple submissions
3. 🔴 [on main] notice that list takes up whole screen and can't be scrolled
4. 🟢 [on PR] notice that list height is capped and list can be scrolled


duvld (approved):
LGTM

PR #7217, opened 2026-07-06
feat(subsequences): background polling for async Google translations DEV-2090

rajpatel24:
### 📣 Summary
This PR enables background polling support for asynchronous automated Google Translate operations.

### 📖 Description
Enables `allow_async=True` on `AutomaticGoogleTranslationAction`, so long-running translation batch jobs now reuse the same `poll_run_external_process` Celery task already used for automatic transcription, instead of relying on the user to manually re-request the translation after a timeout.

`GoogleTranslationService` now returns status: `in_progress` (instead of status: `failed` with the `SYNC_RETRY_LATER_ERROR` message) whenever the batch translation job is still running or a transient Google infrastructure error occurs while starting/polling it. The Google operation reference is preserved in these cases so the next poll resumes the same job rather than starting a duplicate one.

### 👀 Preview steps

1. ℹ️ Have an account and a project with audio submissions
2. Set `MAX_SYNC_CHARS` in `GoogleTranslationService` to a lower value to trigger the async operation with a small audio file.
3. Trigger an automatic Google translation.
4. 🔴 [on main] Notice that the request processes for some time and then returns a `failed` status, asking you to try again later.
5. 🟢 [on PR] Notice that the API returns an `in_progress` status and, after some time, returns the actual translated object.
6. Note that async support is not yet implemented on the frontend, so you will need to either revisit the page later to check whether the result is ready or call the API directly using curl or another API tool.

### DIFF
--- a/kobo/apps/subsequences/actions/automatic_google_translation.py
+++ b/kobo/apps/subsequences/actions/automatic_google_translation.py
@@ -13,6 +13,7 @@ class AutomaticGoogleTranslationAction(TranslationActionMixin, BaseAutomaticNLPA
         automatic=True,
         action_data_key='language',
         review_type=ReviewType.ACCEPTANCE,
+        allow_async=True,
     )
 
     def get_nlp_service_class(self) -> NLPExternalServiceClass:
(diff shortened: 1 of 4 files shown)


RuthTurk (approved):
LGTM

PR #7218, opened 2026-07-07
fix(pairedData): return null source for deleted parent projects DEV-2034

platreth:
### 📣 Summary
When a source project used for a connected (paired) dataset is deleted, it is now clearly reported as deleted rather than still appearing as a normal source.

### 📖 Description
Connected projects pull data from a "source" project. If that source project was deleted, the paired-data endpoint kept listing it as if it were a live source, which the connected-projects UI couldn't reliably detect. The deleted source's `source` link is now returned as `null`, giving the frontend a dependable signal.

### 💭 Notes
`paired_data` is stored as a JSON blob on the destination `Asset` keyed by source uid, with **no cascading FK**, so deleting a source asset leaves a dangling reference (DEV-2034). `PairedDataSerializer.to_representation` now emits `source: null` when the source Asset row no longer exists, detected by reusing the `source__names` mapping the viewset already builds (uid absent ⇒ row deleted) — no extra query on the list endpoint, with a direct `.exists()` fallback when context is absent. Runtime-only change: no OpenAPI schema drift, no migration.

### 👀 Preview steps
1. ℹ️ have an account with two projects; enable data sharing on project **A** (the source).
2. Connect project **B** to **A** (paired data) and `GET /api/v2/assets/<B_uid>/paired-data/`.
3. Delete project **A**.
4. `GET /api/v2/assets/<B_uid>/paired-data/` again.
5. 🔴 [on main] the deleted source still appears with a populated `"source"` URL (only `"source__name"` is null).
6. 🟢 [on PR] that entry's `"source"` is `null`.

### DIFF
--- a/kpi/serializers/v2/paired_data.py
+++ b/kpi/serializers/v2/paired_data.py
@@ -3,6 +3,7 @@ import os
 import re
 
 from django.utils.translation import gettext as t
+from django_request_cache import cache_for_request
 from drf_spectacular.utils import extend_schema_field
 from rest_framework import serializers
 from rest_framework.reverse import reverse
@@ -41,6 +42,9 @@ class PairedDataSerializer(serializers.Serializer):
         return self.__save(validated_data)
 
     def get_source__name(self, paired_data):
+        if self._source_is_deleted(paired_data):
+            return None
+
         # To avoid multiple calls to DB, try to retrieve source names from
         # the context.
         source__name = None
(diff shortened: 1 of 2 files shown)


kpi/serializers/v2/paired_data.py
```diff
@@ -62,8 +62,11 @@ def get_source__name(self, paired_data):
         return source__name
 
     def to_representation(self, instance):
+        source_deleted = self._source_is_deleted(instance)
```
noliveleger: Let's move this line inside `__get_source_asset_url()` and `get_source__name()` and return None (change their signature accordingly). 

You can decorate `_source_is_deleted` with `@cache_for_request` to avoid calling the DB twice. 

in both methods, if `self._source_is_deleted()` returns None, we return right away, in the representation, no need to add conditions anymore. It would be handled inside other methods.

On the side note. That's not great. I think we should move away from the JSON Field and give PairedData its own db model and use FKs. Would you mind to create a linear issue for that?

noliveleger (changes requested):

noliveleger (approved):
LGTM

PR #7220, opened 2026-07-07
fix(dataTable): duplicate column and broken Open button DEV-2404

magicznyleszek:
### 📣 Summary

Fixed Single Processing opening for audio responses when attachment metadata has an empty question path, and prevented duplicate audio columns in Project → Data → Table when old submission paths differ from the current form schema.

### 💭 Notes

Changes here:

- `index.tsx`: for audio cells, fallback to the schema xpath when attachment question path is empty, so Open does not navigate to a broken processing route.
- `tableUtils.ts`: in `getAllDataColumns`, dedupe stale audio keys from submissions when the current survey already contains the matching audio question path.
- `tableUtils.tests.ts`: added regression coverage to ensure current audio xpath is preferred and stale duplicate xpath is excluded.

### 👀 Preview steps

1. Open a project with audio questions and submissions where at least one attachment has an empty question path value (not sure how to achieve it yet).
2. Go to Project → Data → Table.
3. Find an audio response row and use "Open" button.
4. 🔴 On main: for affected attachments, "Open" may navigate to a broken processing URL.
5. 🟢 On this PR: Open navigates to the expected Single Processing route using a safe xpath fallback.
6. In the same table, check columns for an audio question that exists under an old submission path and a current schema path.
7. 🔴 On main: duplicate audio columns may appear.
8. 🟢 On this PR: only the current survey audio column is shown; stale duplicate path column is not shown.

### DIFF
--- a/jsapp/js/components/submissions/DataTableCell/index.tsx
+++ b/jsapp/js/components/submissions/DataTableCell/index.tsx
@@ -72,7 +72,9 @@ export default function DataTableCell(props: DataTableCellProps) {
       ) {
         if (mediaAttachment !== null && props.question.$xpath !== undefined) {
           const audioXpath =
-            typeof mediaAttachment === 'string' ? props.question.$xpath : mediaAttachment.question_xpath
+            typeof mediaAttachment === 'string' || !mediaAttachment.question_xpath
+              ? props.question.$xpath
+              : mediaAttachment.question_xpath
           return (
             <AudioCell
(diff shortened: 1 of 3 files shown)


pauloamorimbr (approved):
LGTM! 👏🏻 

Looked at code, tested for regression, couldn't "organically" simulate the issue.
I've applied a fix for a small issue greptile found (talked to Leszek about it first)

PR #7222, opened 2026-07-08
fix(supplement): add UUID query fallback to supplement view DEV-2320

Guitlle:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Use the fallback query in the supplement view to avoid throwing a 404 error for submissions that didn't get the rootUuid backfilled in mongo.

### 📖 Description
In `kpi/views/v2/data.py`, the method `DataViewSet.supplement` I replaced the strict `meta/rootUuid` MongoDB query  with the method `_get_submission_by_id_or_root_uuid` that searches both `meta/rootUuid` and `meta/instanceID` as a fallback. Also added logic to safely inject the matched uuid into `submission['meta/rootUuid']` if it was missing.

### 👀 Preview steps
Hard to preview, but you could manually remove the `meta/rootUuid` from a mongodb instance record and then request the supplement endpoint to test the fallback: 

1. Have a deployed asset and submit a record to it.

2. Execute in shell_plus the following snippet to manually strip the `meta/rootUuid` from MongoDB for that submission:

```
   
   asset = Asset.objects.get(uid='ASSET_UID')
   submission = list(asset.deployment.get_submissions(asset.owner))[0]
   instance_id = submission['_uuid']
   mongo_id = submission['_id']

   settings.MONGO_DB.instances.update_one(
       {'_id': mongo_id}, 
       {'$unset': {'meta/rootUuid': ""}}
   )
   
   print(f"Use this Instance ID for the URL: {instance_id}")
```
3. Request the supplement endpoint using the instanceID printed above. You can do this via curl, or by navigating in the browser to: `https://kf.local.kbtdev.org/api/v2/assets/{YOUR_ASSET_UID}/data/{INSTANCE_ID}/supplement/`. 

5. Before this PR the endpoint would return a 404 Not Found because it strictly queried meta/rootUuid. After this PR it will return a 200 OK because it successfully falls back to matching the provided UUID against `meta/instanceID`

### DIFF
--- a/kobo/apps/long_running_migrations/jobs/0028_sync_mongo_root_uuid.py
+++ b/kobo/apps/long_running_migrations/jobs/0028_sync_mongo_root_uuid.py
@@ -69,18 +69,18 @@ def _process_batch(docs: list):
     doc_ids = [doc['_id'] for doc in docs]
 
     # Only fetch Postgres records that actually have a root_uuid populated
-    instances = Instance.objects.only('pk', 'root_uuid').filter(
-        pk__in=doc_ids, root_uuid__isnull=False
+    instances_map = dict(
+        Instance.objects.filter(
+            pk__in=doc_ids, root_uuid__isnull=False
+        ).values_list('pk', 'root_uuid')
     )
-    instances_map = {inst.pk: inst.root_uuid for inst in instances}
 
     mongo_updates = [
         UpdateOne(
-            {'_id': doc_id},
-            {'$set': {'meta/rootUuid': add_uuid_prefix(instances_map[doc_id])}},
+            {'_id': pk},
+            {'$set': {'meta/rootUuid': add_uuid_prefix(root_uuid)}},
         )
-        for doc_id in doc_ids
-        if doc_id in instances_map
+        for pk, root_uuid in instances_map.items()
     ]
 
     if mongo_updates:
--- a/kobo/apps/long_running_migrations/jobs/0028_sync_mongo_root_uuid.py
+++ b/kobo/apps/long_running_migrations/jobs/0028_sync_mongo_root_uuid.py
@@ -1,0 +1,86 @@
+from django.conf import settings
+from django.db.models import Q
+from pymongo import UpdateOne
+
+from kobo.apps.long_running_migrations.exceptions import (
+    LongRunningMigrationDependencyError,
+)
+from kobo.apps.long_running_migrations.models import LongRunningMigration
+from kobo.apps.openrosa.apps.logger.models import Instance
+from kobo.apps.openrosa.apps.logger.xform_instance_parser import add_uuid_prefix
+from kpi.utils.log import logging
+
+CHUNK_SIZE = settings.LONG_RUNNING_MIGRATION_SMALL_BATCH_SIZE
+
+
+def run():
+    """
+    Syncs `meta/rootUuid` to MongoDB for submissions that had it backfilled in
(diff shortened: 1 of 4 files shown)


kpi/views/v2/data.py
```diff
@@ -562,18 +562,21 @@ def supplement(self, request, root_uuid, *args, **kwargs):
...
-        except IndexError:
-            raise Http404
+        fetched_root_uuid = submission.get(deployment.SUBMISSION_ROOT_UUID_XPATH)
+
+        if not fetched_root_uuid:
+            fetched_root_uuid = submission.get('meta/instanceID') or submission.get(
+                '_uuid'
+            )
+            fetched_root_uuid = add_uuid_prefix(fetched_root_uuid)
+            submission[deployment.SUBMISSION_ROOT_UUID_XPATH] = fetched_root_uuid
```
noliveleger: AFAIK, `meta/instanceID` is always present and is always prefixed.

What about?:

```python 
        submission = self._get_submission_by_id_or_root_uuid(
            submission_root_uuid, request
        )
        # deployment = self._get_deployment() <- Remove this line

        # TODO remove the block below when LRM 0028 has completed

        root_uuid_key = OpenRosaDeploymentBackend.SUBMISSION_ROOT_UUID_XPATH
        instance_id_key = OpenRosaDeploymentBackend.SUBMISSION_CURRENT_UUID_XPATH     
        if root_uuid_key not in submission:
            submission[root_uuid_key] = submission[instance_id_key]

        submission_root_uuid = remove_uuid_prefix(submission[root_uuid_key])
```

kobo/apps/long_running_migrations/jobs/0028_sync_mongo_root_uuid.py
```diff
@@ -0,0 +1,86 @@
...
+    instances_map = {inst.pk: inst.root_uuid for inst in instances}
+
+    mongo_updates = [
+        UpdateOne(
+            {'_id': doc_id},
+            {'$set': {'meta/rootUuid': add_uuid_prefix(instances_map[doc_id])}},
+        )
+        for doc_id in doc_ids
+        if doc_id in instances_map
+    ]
```
noliveleger: I think we can use `values_list` here to avoid loading objects in memory (even if we use `only`). 

Morever, `instances` already filter doc_ids. We can remove `if doc_id in instances_map`. We can iterate over `instances_map` directly, no? 


```python

      instances_map = dict(                                                                                                                
          Instance.objects.filter(                                                                                                         
              pk__in=doc_ids, root_uuid__isnull=False                                                                                      
          ).values_list('pk', 'root_uuid')                                                                                                 
      )                                                                                                                                    
                                                                                                                                           
      mongo_updates = [                                                                                                                    
          UpdateOne(                                                                                                                       
              {'_id': pk},                                                                                                                 
              {'$set': {'meta/rootUuid': add_uuid_prefix(root_uuid)}},                                                                     
          )                                                                                                                                
          for pk, root_uuid in instances_map.items()                                                                                       
      ]  
```

noliveleger (changes requested):

kobo/apps/long_running_migrations/jobs/0028_sync_mongo_root_uuid.py
Guitlle: yeah, that's better syntaxis

noliveleger (approved):
LGTM

PR #7224, opened 2026-07-08
fix(accounts): restrict users to one linked SSO account DEV-1878

platreth:
### 📣 Summary

You can now only connect one single-sign-on (SSO) provider to your account at a time; attempting to link a second while one is already connected is blocked with a clear error.

### 📖 Description

Previously the "connect SSO" action was only hidden in the UI once an account was linked, but the underlying link was still reachable, so a user could connect a second SSO provider by navigating to it directly. This change enforces the one-SSO-per-account rule on the server, so the restriction holds regardless of how the request is made.

Adds a `SocialAccountAdapter` (`kobo.apps.accounts.adapter`) overriding `pre_social_login`, registered via the new `SOCIALACCOUNT_ADAPTER` setting. On a `connect` flow, if the authenticated user already has a `SocialAccount` for a *different* provider/uid, the link is rejected and a dedicated error page is rendered via allauth's own `render_authentication_error()` helper (the same helper allauth uses for its other SSO errors) explaining that only one SSO account can be linked at a time. Reconnecting the same account, normal SSO login, and signup are unaffected. The guard sits in allauth's `complete_login` before the account is created, so it covers both the classic redirect flow and the headless API (`/api/v2/allauth/…`).

### 💭 Notes

- Root cause was a client-side-only restriction: `ssoSection.component.tsx` hides the connect button once an account is linked, but `allauth.urls` keeps `/accounts/<provider>/login/?process=connect` live. Reported case: a user linked multiple SSOs via the hidden URL on the ACF server.
- No frontend change: the UI already hides the button correctly; only the server guard was missing.
- Covered by 4 new tests in `test_social_account.py` (block second, allow first, allow same-account reconnect, ignore non-connect). Full `kobo/apps/accounts` suite green.
- The error is a dedicated page today on both the classic and headless URLs. `allauth.headless` isn't in `INSTALLED_APPS` yet (see the project plan), so allauth's headless-specific error response is globally inactive regardless of which endpoint is hit. No FE change is needed for this PR.

Follow-up: once `allauth.headless` is enabled later in the Authentication Redesign project, this same code will start redirecting headless/SPA callers with `?error=multiple_sso_not_allowed` instead of rendering the page. We should verify that the SPA reads and displays it correctly at that point.

### 👀 Preview steps

1. ℹ️ Have an account and a configured SSO provider (a `SocialApp`). 
_OR_
Log in as your test user. Instead of linking a first SSO account through the UI, simulate it directly: in Django admin, go to `/admin/socialaccount/socialaccount/add/` and create a `SocialAccount` for your user with any placeholder provider/uid (e.g. provider: fake-idp, uid: fake-uid-123). This instantly puts the user in the "already has one SSO linked" state and complete a real sign-in for this part.

2. Now attempt to link a second, real SSO account: visit the connect URL for your configured provider: `/accounts/<provider>/login/?process=connect&next=%2F%23%2Faccount%2Fsecurity` and complete that provider's real sign-in (e.g. Google's consent screen) (refer https://app.notion.com/p/kobotoolbox/Setting-up-SSO-3f85a206af1349dc8c4fc4c0c919e58e)
4. 🔴 [on main] After signing in, you land back on `/#/account/security` with no error shown, and your account now has two linked SSO providers.
5. 🟢 [on PR] After signing in, you land on a dedicated page titled "SSO account already linked":

   > You can only link one SSO account at a time. Disconnect your existing SSO account before linking a new one.

   No second account is created.


Guitlle (approved):
LGTM, though, users with existing more than one SSO can still log in with their SSO accounts.

rajpatel24:
> LGTM, though, users with existing more than one SSO can still log in with their SSO accounts.

@Guitlle Yes! We're not restricting existing users. They'll still be able to log in with their existing SSO accounts, they just won't be able to add a new one.

PR #7225, opened 2026-07-08
feat(qualitativeAnalysis): add list endpoint for tag trackers DEV-1472

rajpatel24:
### 📣 Summary
This PR adds a list endpoint for tag trackers, nested under the relevant asset and QA question:

`GET /api/v2/assets/{uid_asset}/qual-questions/{uid_qa_question}/tags/`

The endpoint returns previously tracked tag values for a given QA tags question, intended to power frontend autocomplete suggestions when submitting answers to QA questions. 

### 📖 Description
-  The endpoint uses `AdvancedSubmissionPermission`, since only users who can submit answers to QA questions should be able to see this data.
- Results are filtered by asset + `question_uuid` and ordered by `value`.

### 👀 Preview steps

1. ℹ️ Have an account and a project with audio submissions and their transcriptions/translations.
2. Add a tag question under Qualitative Analysis and assign a value to it.
3. 🔴 [on main]: Notice that these tags are not returned anywhere.
4. 🟢 [on PR]: Call this API and verify that it returns the tags.
5. Try adding duplicate tags or the same tag under different questions.
6. 🟢 Verify that the endpoint returns the correct tags based on both the asset and the question.


kobo/apps/subsequences/views.py
```diff
@@ -371,3 +381,60 @@ def create(self, request, *args, **kwargs):
...
+            location=OpenApiParameter.PATH,
+            required=True,
+            description='UUID of the QA tags question',
+        ),
+    ],
+)
+@extend_schema_view(
+    list=extend_schema(
+        description=read_md('subsequences', 'subsequences/qa_tag_trackers_list.md'),
+        responses=open_api_200_ok_response(
```
rgraber: I don't think we need the `responses` or `examples` here since the serializer is pretty simple. I did a quick test locally with just

`list=extend_schema(description=read_md(<stuff>)` and the docs look ok to me:

Image

kpi/schema_extensions/v2/subsequences/examples.py
```diff
@@ -738,6 +738,16 @@ def get_bulk_accept_examples() -> list[OpenApiExample]:
     ]
 
 
+def get_qa_tag_trackers_list_examples() -> list[OpenApiExample]:
```
rgraber: see previous comment, i think we can remove this

kpi/schema_extensions/v2/subsequences/serializers.py
```diff
@@ -192,6 +192,13 @@ def __init__(self, *args, **kwargs):
     },
 )
 
+QATagTrackerResponse = inline_serializer_class(
```
rgraber: see previous comment, i think we can remove this

rgraber (changes requested):
The functionality looks good but I think we can simplify the drf-spectacular stuff

rgraber (approved):
LGTM

PR #7230, opened 2026-07-09
refactor(settings): standardize Constance env overrides to CONSTANCE_ prefix DEV-1976

platreth:
### 🔗 Related PRs
Part of a 3-repo Constance env-var standardization:
- **kobotoolbox/kobo-install#279** — same rename in the kobo-install env template
- **kobotoolbox/kobo-docker#373** — same rename in the kobo-docker default env file

Backward compatibility is preserved (see below), so these can land independently — the deploy-repo PRs simply complete the migration to the new names.

### 📣 Summary
Standardize all Constance override env vars on a single `CONSTANCE_` prefix.

### 📖 Description
The env vars that override Constance defaults used three conventions (bare, `KOBO_`, `CONSTANCE_`). This unifies them on `CONSTANCE_`, matching the existing `CONSTANCE_ASR_MT_GOOGLE_*` precedent.

Renamed:
- `KOBO_SUPPORT_EMAIL` → `CONSTANCE_SUPPORT_EMAIL`
- `KOBO_SUPPORT_URL` → `CONSTANCE_SUPPORT_URL`
- `KOBO_ACADEMY_URL` → `CONSTANCE_ACADEMY_URL`
- `KOBO_COMMUNITY_URL` → `CONSTANCE_COMMUNITY_URL`
- `USAGE_LIMIT_ENFORCEMENT` → `CONSTANCE_USAGE_LIMIT_ENFORCEMENT` (env fallback in `base.py` + pytest `env` in `pyproject.toml`)

**Backward compatible:** each override reads the new `CONSTANCE_`-prefixed variable and falls back to the deprecated name (`KOBO_*` / bare) when it's unset, emitting a `DeprecationWarning` so deployments know to migrate. Old env vars keep working — no hard break.

The fallback logic lives in a new `constance_env` helper in `kobo/settings/utils.py`, moved out of `base.py` along with `dj_stripe_request_callback_method`.

### ✅ Test plan
- [x] `test_api_environment.py` — both settings axes
- [x] usage-limit-enforcement tests (KPI submissions, OpenRosa, subsequences)
- [x] deprecated env var still read + `DeprecationWarning` fired

### DIFF
--- a/kobo/settings/base.py
+++ b/kobo/settings/base.py
@@ -190,6 +190,24 @@ if os.environ.get('DEFAULT_FROM_EMAIL'):
 # must use `constance.config.THE_SETTING` instead of
 # `django.conf.settings.THE_SETTING`
 
+
+def _constance_env(getter, new_key, deprecated_key, default):
+    """
+    Read a Constance override from the `CONSTANCE_`-prefixed environment
+    variable, falling back to its deprecated (unprefixed / `KOBO_`) name.
+
+    Emit a `DeprecationWarning` when the deprecated name is still set so
+    deployments know to migrate to the `CONSTANCE_` prefix.
+    """
+    if deprecated_key in os.environ:
+        warnings.warn(
+            f'{deprecated_key} is renamed {new_key}, '
+            f'update the environment variable.',
+            DeprecationWarning,
+        )
+    return getter(new_key, getter(deprecated_key, default))
+
+
 CONSTANCE_CONFIG = {
     'REGISTRATION_OPEN': (
         True,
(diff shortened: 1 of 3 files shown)


kobo/settings/base.py
```diff
@@ -225,21 +225,24 @@
         'in the user interface',
     ),
     'SUPPORT_EMAIL': (
-        env.str('KOBO_SUPPORT_EMAIL', env.str('DEFAULT_FROM_EMAIL', 'help@kobotoolbox.org')),
```
noliveleger, line 228: Let's not break the compatibility here. 
Just read from the new variable but keep reading from the deprecated one too. 

For any deprecated variable. Please log a Deprecation warning instance.
See https://github.com/kobotoolbox/kpi/blob/39416ad759d3fd1551c429930dd3d9be4f86def1/kobo/settings/base.py#L2007-L2014

noliveleger (changes requested):

kobo/settings/base.py
platreth: Yes, that's way better! Done

kobo/settings/base.py
```diff
@@ -190,6 +190,24 @@
 # must use `constance.config.THE_SETTING` instead of
 # `django.conf.settings.THE_SETTING`
 
+
+def _constance_env(getter, new_key, deprecated_key, default):
```
noliveleger: Would you mind to create an utility file where this could be moved out the settings files. 
You can also move other helpers present in that file (e.g.: `dj_stripe_request_callback_method`)

noliveleger (changes requested):
You should also update the PR description.
It does not match what's happening anymore 🙏

kobo/settings/base.py
platreth: Done in 078be154f — moved `constance_env` (renamed from `_constance_env`) and `dj_stripe_request_callback_method` into `kobo/settings/utils.py`. Also updated the PR description to match the current backward-compatible approach.

noliveleger (approved):
LGTM

PR #7232, opened 2026-07-09
feat(bulkProcessing): near limit alert DEV-1399

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Adds a new Near Limit error alert to bulk transcription and bulk translation modals, shown when a user still has quota left but not enough to process all selected submissions.

### 💭 Notes

Changes here:
- Near-limit evaluator is now implemented in the bulk alerts pipeline.
- Near-limit is treated as an error (blocking), so users cannot continue until they reduce selection or increase quota.
- Both modals now provide near-limit required amounts into the shared alert hook.
- Unit coverage was extended for near-limit scenarios (show/hide behavior and computed message values).

### 👀 Preview steps
1. ℹ️ Have a project with audio/transcript data.
2. Go to Project → Data → Table
3. Select submissions and open either:
   - Bulk transcription modal, or
   - Bulk translation modal.
4. If your account naturally has low remaining quota, use that directly and continue to step 6.
5. Optional local-testing shortcut (temporary code tweak to force near-limit):
   - In BulkTranscriptionModal.tsx, temporarily multiply required seconds:
      - set near-limit required value (`const nearLimitRequiredSeconds`) to something like `totalSelectedAudioDuration * 1000`.
   - In BulkTranslationModal.tsx, temporarily multiply required characters:
      - set near-limit required value to something like `nearLimitRequiredCharacters * 1000`.
6. Reload the page and reopen the modal with selected submissions.
7. 🔴 [on main] Near-limit alert is not appearing.
8. 🟢 [on PR] Near-limit error alert appears with remaining amount.


magicznyleszek:
@greptile-apps review please

magicznyleszek:
@greptile-apps review again

jsapp/js/components/submissions/BulkProcessingModals/BulkTranscriptionModal/BulkTranscriptionModal.tsx
duvld, line 331: Should we change the logic here to also display this button if the user is requesting more than their ASR allowance? Or say something like "your selection will surpass the amount of ASR time you have left, please purchase an add-on or select fewer submissions"?

jsapp/js/components/submissions/BulkProcessingModals/BulkTranslationModal/BulkTranslationModal.tsx
duvld: Same as the BulkTranscriptionModal. Should we remove this comment and update the logic here?

jsapp/js/components/submissions/BulkProcessingModals/alerts/alertEvaluators.ts
```diff
@@ -27,13 +27,46 @@ export function evaluateReachedLimit(context: AlertEvaluationContext): AlertEval
...
+  const balance =
+    actionType === 'transcript' ? serviceUsageData.balances.asr_seconds : serviceUsageData.balances.mt_characters
+
+  if (!balance) {
+    return createInactiveResult('error')
+  }
+
+  const remainingAmount = balance.balance_value
+
+  if (remainingAmount <= 0 || remainingAmount >= nearLimitRequiredAmount) {
```
duvld: I don't think I'm reading this right. Why do we return an error if the user's remaining amount of time is greater than the audio duration time?

jsapp/js/components/submissions/BulkProcessingModals/alerts/types.ts
```diff
@@ -23,6 +23,8 @@ export interface AlertEvaluationContext {
   /** For transcription only */
   selectedRegion?: string
   actionType: BulkActionType
+  /** Required amount for full job in base units: seconds (transcription) or characters (translation). */
+  nearLimitRequiredAmount?: number
```
duvld: I think we should rename this variable, at least it's a bit confusing for me. When we say "required amount for full job" we mean the return of the duration form `useCalculateAudioDuration` right? I think "near limit" adds some ambiguity when this is the actual duration number

jsapp/js/components/submissions/BulkProcessingModals/BulkTranscriptionModal/BulkTranscriptionModal.tsx
```diff
@@ -109,18 +115,40 @@ export function BulkTranscriptionModal(props: BulkTranscriptionModalProps) {
...
+    duration: totalSelectedAudioDuration,
+    isLoading: isTotalSelectedAudioDurationLoading,
+    isError: isTotalSelectedAudioDurationError,
+  } = useCalculateAudioDuration({
+    selectedSubmissions: transcribableSubmissions,
+    fieldId: sourceRowPath,
+    assetUid: props.assetUid,
+  })
+
+  const nearLimitRequiredSeconds =
```
duvld: Same here about the variable name, I think it reads better if we drop the "near limit" prefix

duvld: From what I understand in the preview steps, we want this number to be big to simulate not having enough ASR minutes left(?) I tried setting this to very large numbers (1,000,000,000,000) and `Number.POSITIVE_INFINITY` and the alert still doesn't show.

duvld (commented):
Some questions

magicznyleszek:

jsapp/js/components/submissions/BulkProcessingModals/BulkTranscriptionModal/BulkTranscriptionModal.tsx
magicznyleszek: When user has not enough minutes left there will be an alert stating what are the options. Ignacio confirmed we don't want to display "Purchase add-on" button here - mu guess is that since user has some minutes left after all, we don't want to be forcing a purchase on them just yet
Screenshot 2026-07-13 at 11 21 49

jsapp/js/components/submissions/BulkProcessingModals/BulkTranslationModal/BulkTranslationModal.tsx
magicznyleszek: I removed both legacy alerts, since BulkProcessingAlerts are handling those cases already :)

jsapp/js/components/submissions/BulkProcessingModals/alerts/alertEvaluators.ts
magicznyleszek: This is actually correct, I have added a comment to explain it :)

magicznyleszek: I will also simplify the code to not return that weird `createInactiveResult` but instead simple `null`

jsapp/js/components/submissions/BulkProcessingModals/alerts/types.ts
magicznyleszek: Renamed to `requiredAmount` :)

jsapp/js/components/submissions/BulkProcessingModals/BulkTranscriptionModal/BulkTranscriptionModal.tsx
magicznyleszek: Renamed :+1:

magicznyleszek: Weird, if I change this line:

```
const requiredSeconds =
    isTotalSelectedAudioDurationLoading || isTotalSelectedAudioDurationError ? undefined : totalSelectedAudioDuration
```

to this

```
const requiredSeconds =
    isTotalSelectedAudioDurationLoading || isTotalSelectedAudioDurationError ? undefined : totalSelectedAudioDuration * 1000
```

then it works for me

duvld (approved):
Code LGTM, unable to test it locally though

PR #7234, opened 2026-07-10
fix(parsers): preserve raw filenames beyond 255 chars DEV-1856

platreth:
### 📣 Summary

Fixed a server error that blocked KoboCollect submissions for projects with very long names.

### 📖 Description

Submitting a form from KoboToolbox Collect could fail with a server error when the project name was very long. Submissions now succeed regardless of name length, so no data is lost.

### 💭 Notes

The submission filename is derived from the form title. Long titles overflowed the parser's base64-encoded filename past Django's 255-char cap on `file.name`, raising `binascii.Error` → HTTP 500. The parser now keeps the raw filename in a side-channel dict instead of `file.name`, so it survives any length. No limit change, no migration.

Tests: unit round-trip (`kpi/tests/test_parsers.py`) + e2e `test_post_submission_with_long_filename` (reproduces the 500 pre-fix). Submission and snapshot suites green on both settings axes.

### 👀 Preview steps

1. deploy a project and rename it to a 255+ character title
2. submit a form for it from KoboToolbox Collect
3. 🔴 [on main] submission fails with HTTP 500
4. 🟢 [on PR] submission succeeds

### DIFF
--- a/kpi/tests/test_parsers.py
+++ b/kpi/tests/test_parsers.py
@@ -1,0 +1,36 @@
+from io import BytesIO
+
+from django.core.files.uploadedfile import SimpleUploadedFile
+from django.test import TestCase
+from rest_framework.test import APIRequestFactory
+
+from kpi.parsers import RawFilenameMultiPartParser
+
+
+class RawFilenameMultiPartParserTestCase(TestCase):
+    def _parse_single_file(self, filename: str):
+        upload = SimpleUploadedFile(filename, b'<data/>', content_type='text/xml')
+        request = APIRequestFactory().post(
+            '/', {'xml_submission_file': upload}, format='multipart'
+        )
+        parser = RawFilenameMultiPartParser()
+        result = parser.parse(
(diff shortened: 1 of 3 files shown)


kpi/tests/test_parsers.py
```diff
@@ -0,0 +1,36 @@
...
+
+from django.core.files.uploadedfile import SimpleUploadedFile
+from django.test import TestCase
+from rest_framework.test import APIRequestFactory
+
+from kpi.parsers import RawFilenameMultiPartParser
+
+
+class RawFilenameMultiPartParserTestCase(TestCase):
+    def _parse_single_file(self, filename: str):
```
RuthTurk: Style nit: (same as in https://github.com/kobotoolbox/kpi/pull/7238) protected methods should be listed under public ones.

RuthTurk (changes requested):
LGTM other than small style nit.

RuthTurk (approved):

PR #7236, opened 2026-07-10
feat(KMLExport): route KML exports through KPI export tasks DEV-1591

RuthTurk:
### 📣 Summary
KML exports now run through KPI export tasks and no longer use the legacy OpenRosa KML export endpoint.

### 💭 Notes
- Updated formpack pin to the latest [commit] that implements kml exports (https://github.com/kobotoolbox/formpack/commit/3bf65b74fe876a514cff839fb10c24470dd63de2) 
- Added kml as a valid export type in SubmissionExportTask pipeline.
- Writes KML output via export.to_kml(..., flatten=True).
- Added geo-question field auto-inclusion logic for KML when fields filtering is used, so geometry columns are still fetched.

### 👀 Preview steps

1. ℹ️ Have an account, a project, and a deployed survey with at least one of each geo type question (geopoint, geotrace, and geoshape).
2. Submit at least one response containing geo data.
3. Navigate to `/api/v2/docs` and create an export task `/api/v2/assets/{uid_asset}/exports/` with payload:
```
{
  "type": "kml",
  "lang": "_default",
  "group_sep": "/",
  "hierarchy_in_labels": false,
  "fields_from_all_versions": false,
  "multiple_select": "both"
}
```
4. 🟢 Should succeed with HTTP 201 Created. Navigate to {Project Name} > Data > Downloads you should see an export generated in the export list with no type yet (as the frontend hasn't been implemented).
Screenshot 2026-07-10 at 2 11 27 PM
5. 🟢 Download the export and validate that the kml is correct by uploading the file to google earth. 
6. 🟢 Check that requests to the openRosa endpoint are bypassed by trying to generate a kml export through the legacy kml exports. You should get an error:
Screenshot 2026-07-10 at 2 09 28 PM

### DIFF
--- a/dependencies/pip/dev_requirements.txt
+++ b/dependencies/pip/dev_requirements.txt
@@ -264,7 +264,7 @@ flake8-quotes==3.4.0
     # via -r dependencies/pip/dev_requirements.in
 flower==2.0.1
     # via -r dependencies/pip/requirements.in
-formpack @ git+https://github.com/kobotoolbox/formpack.git@7e92685993f76a14b7772363edd9882dc50da18f#egg=formpack
+formpack @ git+https://github.com/kobotoolbox/formpack.git@3bf65b74fe876a514cff839fb10c24470dd63de2
     # via -r dependencies/pip/requirements.in
 freezegun==1.5.5
     # via -r dependencies/pip/dev_requirements.in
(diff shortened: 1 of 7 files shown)


Guitlle (approved):
Looks good, and works as expected

PR #7238, opened 2026-07-13
perf(asset): fetch permissions once in Asset detail endpoint DEV-1833

platreth:
### 📣 Summary

Opening a single project or collection is a bit faster — the server does less redundant work per request.

### 📖 Description

When you open a project or collection, the server was preparing its permission information twice. It now does that once, so detail pages load slightly faster. Nothing changes in what you see.

### 💭 Notes

- The Asset **detail** endpoint fetched permission assignments **twice** — once in `get_permissions()`, once in the `get_access_types()` fallback. Added a per-request memoized helper so both share **one** query (verified 2→1 with `CaptureQueriesContext`).
- Output is unchanged: the shared fetch is a superset each method narrows in-memory. The list path is untouched (it uses the viewset cache).

Added a regression test (`test_detail_access_types_reuses_permission_fetch`) that fails if the fields stop sharing the fetch. Green on both settings axes, darker clean, no schema drift.

### 👀 Preview steps

1. ℹ️ have an account with a project or collection, and turn on SQL query logging (Django debug toolbar, or log `django.db.backends`)
2. `GET /api/v2/assets/<uid>/` for that asset
3. 🔴 [on main] the object-permission table is queried **twice** for that request
4. 🟢 [on PR] it is queried **once**, and the response body is identical

### DIFF
--- a/kpi/serializers/v2/asset.py
+++ b/kpi/serializers/v2/asset.py
@@ -870,28 +870,6 @@ class AssetSerializer(serializers.HyperlinkedModelSerializer):
         request = self.context.get('request')
         return reverse('paired-data-list', args=(asset.uid,), request=request)
 
-    def _get_object_permission_assignments(self, asset):
-        """
-        Fetch the asset's non-denied assignments once per request (memoized), so
-        get_permissions and get_access_types share one query in the detail view.
-        """
-        cache = self.__dict__.setdefault(
-            '_object_permission_assignments_cache', {}
-        )
-        if asset.pk not in cache:
-            cache[asset.pk] = list(
-                asset.permissions.filter(deny=False)
-                .values(
-                    'uid',
-                    'user_id',
-                    'user__username',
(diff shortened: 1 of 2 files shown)


kpi/serializers/v2/asset.py
```diff
@@ -870,23 +870,42 @@ def get_paired_data(self, asset):
         request = self.context.get('request')
         return reverse('paired-data-list', args=(asset.uid,), request=request)
 
+    def _get_object_permission_assignments(self, asset):
```
RuthTurk: Style nit: _get_object_permission_assignments() is a protected helper, so it should be grouped with the other protected methods after the public serializer methods (and kept in alphabetical order within that section).

RuthTurk (changes requested):
Looks great overall, just one small requested change :)

RuthTurk (approved):

PR #7239, opened 2026-07-13
fix(supplement): include null value for deleting automatic transcription/translation DEV-1722

platreth:
### 📣 Summary

API-schema fix

### 📖 Description

The API reference for automatic transcription and translation now documents that a result can be deleted by sending an empty (null) value. This only affects the API documentation and generated API client; the delete itself already worked.

### 💭 Notes

- `_nlp_automatic_action_schema` now includes an optional, null-only `value` (`{enum: [null], nullable: true}` — drf-spectacular's NullEnum idiom, since OpenAPI 3.0.3 has no `type: null`). Deleting already worked server-side (`actions/base.py`: "Deletion is triggered by passing `{value: null}`").
- Regenerated `static/openapi/*` + the orval client. The two automatic `Patched…` types now carry `value?: null`, so the `as any` casts and `TODO … DEV-1722` markers in `survey-data.ts` are gone.
- No runtime/behavior change, no migration.

### 👀 Preview steps

Observable via the API docs / generated types (no user-facing UI change):

1. ℹ️ Open the automatic transcription PATCH schema — Swagger at `/api/v2/docs/`, or `jsapp/js/api/models/patchedDataSupplementPayloadOneOfAutomaticGoogleTranscription.ts`.
2. 🔴 [on main] `value` is absent from the type/schema.
3. 🟢 [on PR] the type/schema includes `value?: null`.

### DIFF
--- a/jsapp/js/api/models/patchedDataSupplementPayloadOneOfAutomaticGoogleTranscription.ts
+++ b/jsapp/js/api/models/patchedDataSupplementPayloadOneOfAutomaticGoogleTranscription.ts
@@ -9,9 +9,12 @@ The endpoints are grouped by area of intended use. Each category contains relate
 **General note**: All projects (whether deployed or draft), as well as all library content (questions, blocks, templates, and collections) in the user-facing application are represented in the API as "assets".
  * OpenAPI spec version: 2.0.0 (api_v2)
  */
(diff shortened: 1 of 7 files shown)


noliveleger (approved):
LGTM

PR #7248, opened 2026-07-14
fix(nlp): add configurable Google STT language code overrides and improve batch error handling DEV-2403

rajpatel24:
### 📣 Summary
Adds configurable language code overrides for Google Speech-to-Text v2 (e.g. mapping Swahili to `auto`) and improves batch transcription error handling by surfacing per-file Google API errors instead of generic missing-output failures.

#### Problem

Google Speech-to-Text v2 currently fails to transcribe Swahili when using the standard language codes (`sw` or `sw-KE`), even though the language is supported. However, using `auto` allows Google to correctly detect and transcribe the audio.

Additionally, batch transcription failures could sometimes be reported as "No transcription output written to GCS" even though Google had already returned a more specific per-file error.

#### Solution

This PR introduces configurable language code overrides for Google Speech-to-Text and improves error handling for batch transcription responses.

#### Changes

- Added a configurable `ASR_LANGUAGE_CODE_OVERRIDES` Constance setting with environment variable support.
- Added parsing for comma-separated language mappings (e.g. `sw:auto,sw-KE:auto`).
- Applied language overrides only within the Google Speech-to-Text integration before sending requests to Google, keeping Kobo language definitions unchanged.
- Added informational logging whenever a configured language override is applied.
- Improved Speech-to-Text v2 batch error handling by detecting per-file errors returned by Google before attempting to read output files from Cloud Storage.
- Added clearer logging and user-facing error messages for per-file transcription failures.

This PR introduces a default language-code override for Google Speech-to-Text:
- sw -> auto
- sw-KE -> auto

This workaround is enabled by default because Google Speech-to-Text v2 currently fails to transcribe Swahili when explicitly using `sw` or `sw-KE`, while automatic language detection (`auto`) succeeds.

The default can be overridden or disabled by setting the `CONSTANCE_ASR_LANGUAGE_CODE_OVERRIDES` environment variable (or by updating the Constance setting after deployment).

This workaround is intended to be temporary and should be removed once Google resolves the underlying Speech-to-Text issue.

### 👀 Preview steps

1. ℹ️ Use an account with a project that contains a long audio submission (>20 minutes). You can also download an existing user's audio for testing.
2. Request a transcript in **Swahili (`sw`)**.
3. 🔴 [on release]: Observe that the transcription takes an excessively long time to complete (more than 5 hours).
4. 🟢 [on PR]: Observe that the transcription is generated within approximately 5 minutes. Also verify that the logs contain a message similar to:
   `"Starting Google automatic transcription... source_lang='auto' ..."`


platreth (commented):
Looks good. Just couple things:

I think we should have a small test on `get_asr_language_code_overrides`. It runs on every transcription and if the parsing breaks we're back to the sw timeout. Same for _extract_per_file_error!

Also the docstring says it returns {} on a parse error but it actually just skips the bad entries and keeps the good ones

And maybe a # TODO(DEV-2403): remove when Google fixes sw on the default so we don't forget it's temporary and we can remove it easily later :)

noliveleger:
> Looks good. Just couple things:
> 
> I think we should have a small test on `get_asr_language_code_overrides`. It runs on every transcription and if the parsing breaks we're back to the sw timeout. Same for _extract_per_file_error!
> 
> Also the docstring says it returns {} on a parse error but it actually just skips the bad entries and keeps the good ones
> 
> And maybe a # TODO(DEV-2403): remove when Google fixes sw on the default so we don't forget it's temporary and we can remove it easily later :)

I personally addressed Hugo's comments so we could merge the PR (already approved) and avoid delaying the new release deployment.

PR #7249, opened 2026-07-15
fix(versions): allow date_deployed to be false in version responses DEV-1823

platreth:
### 📣 Summary

Corrects the deployment date the API reports for form versions that were never deployed.

### 📖 Description

In a form's version history, versions that were never deployed have no deployment date. The API already returns this correctly, but its published type description wrongly claimed the deployment date is always a date. This corrects the description so the value is understood as either a date or "not deployed" — matching what the API actually returns. There is no visible change in the app.

### 💭 Notes

- Root cause: `AssetVersionListSerializer.get_date_deployed` returns `obj.deployed and obj.date_modified`, i.e. a datetime when deployed and the literal `false` otherwise. The OpenAPI schema declared it as a plain `date-time`, which propagated to the orval client as `date_deployed: string`.
- Fix is schema-only: `DateDeployedField` now maps to `oneOf: [date-time, false]` in `kpi/schema_extensions/v2/versions/extensions.py`. The committed `static/openapi/schema_v2.*` and `jsapp/js/api/*` were regenerated (not hand-edited); orval collapses the boolean `enum:[false]` to `boolean`, so the emitted type is `string | boolean`.
- Frontend: removed the now-unneeded `as string | false` casts (and the `TODO … DEV-1823`) in `FormHistory.tsx`.

### 👀 Preview steps

1. ℹ️ open the version **History** of a project that has a deployed version and a later undeployed form change
2. the table shows "Deployment" for deployed rows and "Form change" for undeployed rows — unchanged by this PR
3. 🔴 [on main] at `/api/v2/docs/`, `VersionListResponse.date_deployed` is typed as a `date-time` string only, even though the API returns `false` for undeployed versions
4. 🟢 [on PR] `date_deployed` is typed as a `date-time` **or** `false`, matching the real response


rgraber (approved):
LGTM

PR #7255, opened 2026-07-15
fix(constance): add new setting to fieldsets DEV-2178

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Fixes an error that was preventing saves to Constance.

### 💭 Notes
Constance throws an error if a field is present in the config but not present in fieldsets. We don't actually want people editing the TOS last update field, so we disable it and make a note that it should not be edited manually. It would be nice to remove the 'reset to default' option but that's more work than this small fix requires, and it's only for admins so it's not a huge deal. 

### 👀 Preview steps

1. ℹ️ have an admin account
2. In admin, attempt to change a constance config variable
3. 🔴 [on main] Error
4. 🟢 [on PR] update is successful

### DIFF
--- a/kobo/settings/base.py
+++ b/kobo/settings/base.py
@@ -219,7 +219,14 @@ CONSTANCE_CONFIG = {
         'if field is not blank'
     ),
     'TERMS_OF_SERVICE_URL': ('', 'URL for terms of service document'),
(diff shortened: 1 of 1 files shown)


noliveleger (approved):
LGTM

PR #7260, opened 2026-07-16
feat(data): add _date_modified field to submission data DEV-2861

### LINKED ISSUE #7259: Data API: no way to detect edited submissions
## Description

The submission JSON served by `GET /api/v2/assets/{uid}/data/` carries no last-modified timestamp. API consumers replicating data can't detect edits incrementally: we see no other choice but to re-fetch all submissions on every sync and hash the payloads.

Ask: add a `_date_modified` key to the served document, updated on every path that changes it (Enketo edits, validation-status changes, attachment changes), and accept it in the `query` parameter so consumers can sweep `?query={"_date_modified": {"$gt": "..."}}`.

## Steps to Reproduce

1. Submit a record and fetch it via the data API. Note `_submission_time` and the full JSON.
2. Edit the record in Enketo (or change its validation status, or remove an attachment).
3. Fetch again. The content changed, but no timestamp reflects it: `_submission_time` is unchanged and no `_date_modified` key exists.

## Expected behavior

A machine-readable timestamp that moves whenever the served document changes, usable for incremental sync.

## Actual behavior

`_submission_time` is `Instance.date_created` and never moves. The Mongo document built by `ParsedInstance.to_dict_for_mongo()` (`kobo/apps/openrosa/apps/viewer/models/parsed_instance.py`) serializes no modified time at all.

The Postgres model already tracks `Instance.date_modified` via `AbstractTimeStampedModel`. It's just never written into the document the API serves.

Two wrinkles for an implementation to handle:
- The bulk validation-status path updates instances via a queryset `.update()` (`kobo/apps/openrosa/apps/logger/utils/instance.py`), which bypasses `save()` and won't bump `date_modified` today.
- Attachment deletion updates Mongo directly via `ParsedInstance.bulk_update_attachments()`.

## Additional details

We replicate ~35 forms / ~230k submissions nightly into a SQL warehouse. Post-submission edits are routine (enumerator corrections, validation workflow), so we fetch everything every run and hash each payload to find the ~0.1% that changed. It works, but it's a full re-download per sync to catch a tenth of a percent of changes.

Existing signals don't cover this. `_uuid` changes on Enketo edits (previous value in `meta/deprecatedID`), but noticing that still requires fetching every record, and it doesn't move on validation-status or attachment-only changes. The project-history logs API (`/api/v2/assets/{uid}/history/`) records submission modifications but requires manage-project permission and has had tier-limited lookback since #7137.

Related: kobotoolbox/kpi#2781 (webhook on submission update), and two community threads where staff confirmed in 2021 that `_submission_time` doesn't update on edit:
- https://community.kobotoolbox.org/t/expose-submission-version-history-via-api-v2/70635
- https://community.kobotoolbox.org/t/api-to-get-history-of-submission/17760

RuneO:
### 📣 Summary
Submissions returned by the data API now carry a `_date_modified` timestamp that
updates whenever the submission changes (edits, validation status, attachments),
so integrations can finally detect changed records without re-downloading everything.

### 📖 Description

Until now there was no way to tell that a submission had changed: `_submission_time`
is the creation time and never moves, and no modified-time field existed. Anyone
keeping an external copy of their data (reporting databases, dashboards) had to
re-fetch every submission on every sync. Each submission now reports
`_date_modified`, and you can query it, e.g.
`?query={"_date_modified": {"$gt": "2026-07-16T00:00:00"}}`.
- `ParsedInstance.to_dict_for_mongo()` now serialises the existing PostgreSQL
  `Instance.date_modified` into the Mongo document as `_date_modified`
  (same `MONGO_STRFTIME` format as `_submission_time`).
- Two paths that bypassed `AbstractTimeStampedModel.save()` now refresh
  `date_modified` explicitly with one shared timestamp for PostgreSQL and Mongo:
  the bulk validation-status path (`set_instance_validation_statuses`) and
  `ParsedInstance.bulk_update_attachments()`.
- Documents written before this change lack the key; at read time
  `_inject_properties()` falls back to `_submission_time` (same pattern as the
  `meta/rootUuid` injection). Mongo-side backfill happens naturally on the next
  write per submission (or via `sync_mongo`), so `$gt` queries cover legacy
  documents progressively.
- `_date_modified` is added to `AbstractDataFrameBuilder.IGNORED_COLUMNS`, so
  legacy CSV/XLS exports are unchanged.
- No schema migrations. Fixes #7259.

### 💭 Notes
- Side benefit: bulk validation-status changes now correctly invalidate the legacy
  export cache (single-status changes already did, since they bump `date_modified`
  via `save(update_fields=...)`; bulk changes previously left stale cached exports
  even though `_validation_status` is an export column).
- Trade-off: trashing an attachment now also invalidates that cache; the
  regenerated export is identical (attachments aren't export columns). Happy to
  scope the bump differently if you prefer.
- Known limitation: submissions edited *before* this feature deploys report the
  fallback value until their next change. Recovering historical edit times would
  need a Postgres→Mongo backfill, which felt out of scope for a first PR.
- `DataResponse` (kpi/schema_extensions/v2/data/serializers.py) doesn't yet list
  the field. I left it untouched because the openapi/orval CI jobs fail on
  uncommitted regenerated artifacts, and I can't run `generate_api.sh` locally.
  Glad to add it + regenerate if you want it in the published schema.
- Tests added in `kobo/apps/openrosa/apps/viewer/tests/test_date_modified.py`;
  linted with your black/isort/ruff settings on changed lines. I could not run
  the full suite locally (no kobo-install stack). Relying on CI here.

### 👀 Preview steps
1. ℹ️ have an account and a deployed project with one submission
2. `GET /api/v2/assets/{uid}/data/`: `_date_modified` equals


kpi/deployment_backends/base_backend.py
```diff
@@ -832,13 +837,27 @@ def _inject_properties(
...
         submission = self._inject_root_uuid(submission)
+        submission = self._inject_date_modified(submission)
         submission['_validation_status'] = (
             submission.get('_validation_status', None) or {}
         )
         submission.pop('_tags', None)
         submission.pop('_notes', None)
         return submission
 
+    def _inject_date_modified(self, submission: dict) -> dict:
```
Guitlle, line 849: `MongoHelper._get_cursor_and_count()` applies a projection containing only _id, _date_modified, and mandatory UUID fields. Since _submission_time was not requested, when `_inject_date_modified` runs on a legacy submission (which lacks `_date_modified` in Mongo), `submission.get(SUBMISSION_TIME)` returns None so `_date_modified` is omitted from the response

kobo/apps/openrosa/apps/viewer/models/parsed_instance.py
```diff
@@ -308,6 +310,7 @@ def to_dict_for_mongo(self):
             self.STATUS: self.instance.status,
             GEOLOCATION: [self.lat, self.lng],
             SUBMISSION_TIME: self.instance.date_created.strftime(MONGO_STRFTIME),
+            DATE_MODIFIED: self.instance.date_modified.strftime(MONGO_STRFTIME),
```
Guitlle: It'd be safer to do: `(self.instance.date_modified or self.instance.date_created).strftime(MONGO_STRFTIME)`

kobo/apps/openrosa/apps/viewer/tests/test_date_modified.py
```diff
@@ -0,0 +1,100 @@
...
+from kobo.apps.openrosa.apps.main.tests.test_base import TestBase
+from kobo.apps.openrosa.apps.viewer.models.parsed_instance import ParsedInstance
+from kobo.apps.openrosa.libs.utils.common_tags import (
+    DATE_MODIFIED,
+    MONGO_STRFTIME,
+    SUBMISSION_TIME,
+)
+
+
+class TestDateModified(TestBase):
```
Guitlle, line 17: Good unit tests, but we should also have API level tests that 
* assert that _date_modified is included in the response
* test the filter works and
* the property is injected when missing

The tests could go in kpi/tests/api/v2/test_api_submissions.py

Guitlle (changes requested):
About the OpenAPI changes needed, we do want to have this included in this PR. I could help running the generate_api.sh script.

RuneO:
Thanks for the review @Guitlle. All four points are addressed in the latest push:

- `MongoHelper._get_cursor_and_count()` now adds `_submission_time` to the projection whenever `_date_modified` is requested, so the legacy fallback works with `fields` too.
- `to_dict_for_mongo()` falls back to `date_created` when `date_modified` is null.
- `_date_modified` is added to `DataResponse`. I patched `schema_v2.yaml`, `schema_v2.json` and `dataResponse.ts` by hand to mirror `_submission_time`, since I cannot run `generate_api.sh` locally. If the openapi job shows drift, I would gladly take you up on the offer to run the script.
- Three API-level tests in `kpi/tests/api/v2/test_api_submissions.py` cover presence, `$gt`/`$lt` filtering and injection when the key is missing (with and without `fields`).

PR #7262, opened 2026-07-16
fix(subsequences): make translation _dependency optional in schema DEV-1721

platreth:
### 📣 Summary

Corrects the data API documentation so it no longer claims that every translation always has a linked source transcript.

### 📖 Description

In the API, each translation records which transcript it was based on. Deleted translations don't have one. The documentation wrongly listed that link as always present — it's now marked optional, matching what the API actually returns.

### 💭 Notes

- Drops `_dependency` from the `required` list of `SupplementalDataVersionItemAutomaticWithDep` and `SupplementalDataVersionItemManualWithDep` (still in `properties`); the generated orval type becomes `_dependency?`.
- No serializer/runtime change — `_dependency` is still returned for every non-deleted revision. Brings these two components in line with `SupplementalDataVersionItemQualAutomatic`, which was already optional.
- `static/openapi/*` + `jsapp/js/api/*` are regenerated artifacts. Branch sits a few commits behind `main` because the local dev container needs an older `formpack` than the current `main` tip; regeneration is unaffected.

### 👀 Preview steps

1. ℹ️ have an asset with a translation that has a deleted revision
2. `GET /api/v2/assets/<uid>/data/<submission>/supplemental/`
3. 🔴 [on main] the generated type marks `_dependency` as required, so a client assumes it is always present — even on a deleted revision where it is absent
4. 🟢 [on PR] the generated type marks `_dependency?` as optional, matching the actual response

### DIFF
--- a/jsapp/js/api/models/supplementalDataVersionItemAutomaticWithDep.ts
+++ b/jsapp/js/api/models/supplementalDataVersionItemAutomaticWithDep.ts
@@ -16,6 +16,6 @@ export interface SupplementalDataVersionItemAutomaticWithDep {
   _dateCreated: string
   _uuid: string
   _dateAccepted?: string
-  _dependency: SupplementalDataDependency
+  _dependency?: SupplementalDataDependency
   _data: SupplementalDataContentAutomatic
 }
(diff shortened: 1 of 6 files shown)


RuthTurk (approved):
LGTM

PR #7265, opened 2026-07-16
refactor(fonts): replace webfonts-generator DEV-2443

jamesrkiger:
### 💭 Notes
Replaces outdated webfonts-generator package with fantasticon. Webfonts-generator hadn't been updated in years and was locking us into transitive dependencies that were triggering security alerts. Also removes replace-in-file dependency, which was used for a caching hackfix that is no longer needed.

### 👀 Preview steps
Load app/storybook and visually check icons. Update an icon, run generate-icons again, then confirm change is reflected.


scripts/generate_icons.js
```diff
@@ -15,110 +17,60 @@ console.warn(
...
 
-  fs.writeFile(`${destDir}/k-icons.ts`, fileContent, (err) => {
-    if (err) {
-      throw new Error('Fail!', err)
-    }
+    console.info('Done.')
+  })
+  .catch((error) => {
+    throw new Error('Font generation failed', { cause: error })
   })
```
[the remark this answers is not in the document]
jamesrkiger: @greptile Check for yourself, do any code paths still inline the css directly?

package.json
```diff
@@ -135,7 +135,7 @@
     "@types/reflux": "^6.4.6",
     "@types/rtl-detect": "^1.0.3",
     "@types/zxcvbn": "^4.4.4",
-    "@vusion/webfonts-generator": "^0.8.0",
+    "fantasticon": "^4.1.0",
```
magicznyleszek: Are we dropping Node v20.19.0 support? :)

```
npm warn EBADENGINE Unsupported engine {
npm warn EBADENGINE   package: 'fantasticon@4.1.0',
npm warn EBADENGINE   required: { node: '>= 22.0' },
npm warn EBADENGINE   current: { node: 'v20.19.0', npm: '10.8.2' }
npm warn EBADENGINE }
```

scripts/generate_icons.js
```diff
@@ -15,110 +17,61 @@ console.warn(
...
+    css: path.join(root, 'jsapp/k-icons-css-template.hbs'),
+    html: undefined,
   },
-)
+})
+  .then(({ codepoints }) => {
+    console.info('Copying k-icons.css to SCSS file…')
+    fs.copyFileSync(
+      path.join(destDir, 'k-icons.css'),
+      path.join(destDir, 'k-icons.scss'),
```
magicznyleszek: Previous comment said that we need scss because of `@extend`. I see now we only have one `@extend .k-icon` and it seems we could delete it and only use css file?

magicznyleszek (approved):
Generally everything still works, just two questions before merging

package.json
jamesrkiger: Good catch. I have dropped fantasticon down to 3.0 for the moment. We can upgrade when we drop node 20 (which I have opened a linear issue for).

scripts/generate_icons.js
jamesrkiger: Agreed.

PR #7267, opened 2026-07-16
fix(longRunningMigrations): do not tag XForm as failed on Celery timeout DEV-2450

noliveleger:
### 📣 Summary

A background data-repair task could permanently skip a form's submissions, either when it ran out of time or when it hit an old submission it did not know how to handle. It now resumes on the next cycle and handles those old submissions instead of giving up for good.

### 💭 Notes

Long-running migration `0027_backfill_remaining_root_uuid` calls the `clean_duplicated_submissions_root_uuid` management command when it hits an `IntegrityError` while backfilling `root_uuid`. If Celery interrupted that command (`SoftTimeLimitExceeded` / `TimeLimitExceeded`), the broad `except Exception` in `_process_instances_batch()` caught the timeout, logged it as a clean failure, and tagged the XForm with `FAILED_TAG` (`kobo-root-uuid-failed-0027`). That permanently skipped it on every subsequent run even though nothing was actually wrong with the data.

Both timeout exceptions are now re-raised before that handler so they propagate to `LongRunningMigration.execute()`, which already treats them as "took too long, resume on the next Celery beat cycle" without marking the migration as failed.

Production logs then showed the timeout was sometimes swallowed one layer deeper: `clean_duplicated_submissions_root_uuid` wrapped it into a `CommandError` (`command has completed with errors: SoftTimeLimitExceeded()`), which the direct re-raise did not catch, so the XForm was tagged failed anyway. Two changes close that gap. The wrapper command now re-raises `SoftTimeLimitExceeded` / `TimeLimitExceeded` instead of swallowing them into a `CommandError`. As defense in depth, `_process_instances_batch()` walks the exception `__cause__` / `__context__` chain (`_find_timeout_in_chain`) and, if it finds a timeout wrapped in any other exception type, re-raises the timeout instead of tagging the XForm. `FAILED_TAG` is now reserved for genuine, unrecoverable data errors.

A separate blocker also surfaced in production: `clean_duplicated_submissions._replace_duplicates()` crashed with `instanceID node not found` on legacy submissions whose XML has no `meta/instanceID` node, aborting the whole form's cleaning. `set_meta()` is now guarded (the DB `uuid` update already resolves the collision, so the missing node is fine) and `_populate_root_uuid()` falls back to the submission `uuid`, matching the tolerance already used by `_resolve_conflict_by_pk` and LRM 0027. This does not fabricate data into archived XML.

For visibility, LRM 0027 now logs when it starts and finishes cleaning duplicates for an XForm, so a long clean is no longer mistaken for a frozen task.

### 👀 Preview steps

Reproducing a real Celery timeout mid-command is contrived, so this needs a forced timeout.

1. ℹ️ have a form whose submissions still have a null `root_uuid` and at least one duplicate that makes `clean_duplicated_submissions_root_uuid` raise, so LRM 0027 enters the clean path
2. force the clean command to be interrupted, for example by setting `CELERY_LONG_RUNNING_TASK_SOFT_TIME_LIMIT` very low so Celery raises `SoftTimeLimitExceeded` while it runs
3. run the long-running migrations and let LRM 0027 process that form
4. check the form's tags and the migration status
5. 🔴 [on main] the form gets tagged `kobo-root-uuid-failed-0027` and is skipped on every later run
6. 🟢 [on PR] the form is not tagged, the migration stays `in_progress` and resumes on the next cycle

### DIFF
--- a/kobo/apps/long_running_migrations/jobs/0027_backfill_remaining_root_uuid.py
+++ b/kobo/apps/long_running_migrations/jobs/0027_backfill_remaining_root_uuid.py
@@ -1,5 +1,6 @@
 # Generated on 2026-06-23
 
+from celery.exceptions import SoftTimeLimitExceeded, TimeLimitExceeded
 from django.conf import settings
 from django.core.management import call_command
 from django.db import IntegrityError
@@ -145,6 +146,12 @@ def _process_instances_batch(
                     xform=xform.id_string,
                     verbosity=2,
                 )
+            except (SoftTimeLimitExceeded, TimeLimitExceeded):
+                # Celery interrupted the command mid-run: this is not an
+                # unrecoverable data error, so do not tag the XForm as failed.
+                # Let it propagate to `execute()`, which resumes on the next
+                # Celery beat cycle.
+                raise
             except Exception as e:
                 logging.error(
                     f'[LRM 0027] - Failed to clean duplicated submissions: {str(e)}'
--- a/kobo/apps/long_running_migrations/jobs/0027_backfill_remaining_root_uuid.py
+++ b/kobo/apps/long_running_migrations/jobs/0027_backfill_remaining_root_uuid.py
@@ -118,6 +118,25 @@ def _check_lrm_0005_is_completed():
         )
 
 
+def _find_timeout_in_chain(exc: BaseException) -> BaseException | None:
+    """
+    Walk the exception's cause/context chain and return the first Celery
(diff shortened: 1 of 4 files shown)


platreth (changes requested):
Looks good! 
Would be nice to have a test for `_find_timeout_in_chain ` and a regression test mocking call_command

noliveleger:
> Would be nice to have a test for `_find_timeout_in_chain ` and a regression test mocking call_command

Done!

platreth (approved):
LGTM

PR #7268, opened 2026-07-16
feat(dataTable): deduplicate non-audio attachment columns DEV-2407

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Data Table now safely de-duplicates stale attachment columns not only for audio, but also for image, video, and file questions, so users no longer see duplicate columns after group/path changes.

### 💭 Notes

Changes here:
- Expanded stale/current attachment de-duplication from audio-only to audio, background-audio, image, video, and file
- Added metadata-aware safeguards using attachment question paths, so we only collapse columns when evidence strongly suggests they represent the same field
- Kept behavior conservative for ambiguous cases, including same filename appearing on both paths
- Added tests
- Added helpful 🤞 comments

### 👀 Preview steps

1. ℹ️ Have a deployed project with `image`, `video`, `file`, and `audio` questions inside a group
2. Add a submission
3. Edit form - rename group (name and label)
4. Redeploy
5. Go to Project → Data → Table
6. 🔴 On main: notice duplicate columns can appear for the same attachment question (old path + current path)
  - Note: we already fixed `audio` on `main`, so for that type it should work already. We test it again here as the deduplicating code changed for all types
7. 🟢 On this PR branch: notice duplicate stale attachment columns are removed when they are true mirrors


magicznyleszek:
@greptile-apps review and have a nice day

magicznyleszek:
@greptile-apps review

pauloamorimbr (approved):
LGTM! 👏🏻 

Tested and  it works as described! 

Bot raised a valid point tho: The dedup logic is only considering the current page, so we may end up having different fields being displayed for each page depending on submissions done. (I don't consider this to be bad, though. But it's worth to think about)

PR #7269, opened 2026-07-16
fix(account): delete button disabled check DEV-2445

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Fixed an issue where users who deleted all their projects could still be blocked from deleting their account because non-project assets were counted by mistake.

### 💭 Notes

Changes here:

- `DeleteAccountBanner.tsx` updated the asset lookup query to only count projects by filtering with asset_type:survey.
  - This aligns the logic with the banner copy, which refers to projects, not all asset types.
- `DeleteAccountBanner.stories.tsx` - Added a regression story-test `UserHasOnlyNonProjectAssets`

### 👀 Preview steps

1. ℹ️ Have an account with a template and a project
2. ℹ️ Have `ALLOW_SELF_ACCOUNT_DELETION` enabled in `/admin` Constance → Config
3. Go to Account Settings → Profile and scroll down
4. 🟢 Notice that "Delete account" button is disabled
5. Go to My Library
6. 🟢 Notice that user owns some library items (at least one is enough)
7. Go to My Projects
8. Delete all your projects
9. Go to Account Settings → Profile and scroll down
10. 🔴 On main: "Delete account" button is disabled
13. 🟢 On this PR: "Delete account" button is enabled and the message confirms you can delete your account

### DIFF
--- a/jsapp/js/account/DeleteAccountBanner.tsx
+++ b/jsapp/js/account/DeleteAccountBanner.tsx
@@ -18,7 +18,7 @@ export default function DeleteAccountBanner() {
 
   const username = session.currentLoggedAccount.username
   const { data: assetsData, isPending } = useAssetsList({
-    q: `(owner__username:${username})`,
+    q: `(owner__username:${username} AND asset_type:survey)`,
     limit: 1,
   })
 
--- a/jsapp/js/constants.ts
+++ b/jsapp/js/constants.ts
@@ -150,6 +150,8 @@ export enum AssetTypeName {
   template = 'template',
   survey = 'survey',
   collection = 'collection',
+  // Sometimes an empty asset is possible to be created by some funky import flows
+  // empty = 'empty'
 }
 
 export interface AssetTypeDefinition {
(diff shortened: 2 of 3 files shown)


jsapp/js/constants.ts
```diff
@@ -150,6 +150,8 @@ export enum AssetTypeName {
   template = 'template',
   survey = 'survey',
   collection = 'collection',
+  // Sometimes an empty asset is possible to be created by some funky import flows
+  // empty = 'empty'
```
pauloamorimbr, line 154: Shouldn't we have the actual `empty` key uncomented then, since it can exist at some point for a check? 🤔

jsapp/js/account/DeleteAccountBanner.tsx
```diff
@@ -18,7 +18,7 @@ export default function DeleteAccountBanner() {
 
   const username = session.currentLoggedAccount.username
   const { data: assetsData, isPending } = useAssetsList({
-    q: `(owner__username:${username})`,
+    q: `(owner__username:${username} AND asset_type:survey)`,
```
pauloamorimbr, line 21: Just checking, any other asset type won't block the deletion then, right?

pauloamorimbr (approved):
LGTM! 👏🏻

jsapp/js/constants.ts
magicznyleszek: I don't know and didn't want to decide. Plus I am not sure what code is using those enums, and didn't want to break anything

jsapp/js/account/DeleteAccountBanner.tsx
magicznyleszek: Yes, we only block by surveys ("projects")

PR #7275, opened 2026-07-17
fix(openapi): type validation error responses as ErrorValidation DEV-1220

platreth:
### 📣 Summary

The API reference now documents validation errors with their real shape — a map of field name to error messages — instead of an untyped placeholder.

### 📖 Description

When a request fails validation, the KoboToolbox API returns each field's errors as a list of messages. Until now the API reference described this `400` response as an untyped `detail` object, which didn't match what the API actually sends. It is now documented accurately, so anyone building against the API — or relying on the types generated from its schema — gets the correct shape for validation errors.

### 💭 Notes

- Renamed the OpenAPI type `ErrorObject` (`{ detail: unknown }`) → `ErrorValidation` (`Record<string, string[]>`), matching DRF's unwrapped serializer `ValidationError` payloads (there is no `detail` envelope — the custom exception handler delegates to stock DRF). A new `ErrorValidationSerializer` plus a drf-spectacular `OpenApiSerializerExtension` emit `additionalProperties: string[]`; the `400` example was flattened to match.
- Regenerated `static/openapi/*` and the orval client. `ServerError` no longer `implements` the renamed type and the `orval.types.d.ts` `Error` augmentation for it was dropped — a validation map isn't an `Error`.
- The two list views (`MembersRoute`, `usageProjectBreakdown`) now type their table error as `ErrorDetail` (their GET hooks never return a validation map); also fixed a stray `schema-utils` `ErrorObject` import in `UniversalTable`.
- No runtime behavior change. Gates green: tsc, biome, ruff, darker, schema generation (no drf-spectacular warnings).

### 👀 Preview steps

1. ℹ️ open the API docs at `/api/v2/docs/` on an endpoint that can return `400` (e.g. create asset)
2. look at the documented `400 Bad Request` response body and its example
3. 🔴 [on main] the schema is `ErrorObject` = `{ detail: <unknown> }` and the example wraps the errors under `detail`
4. 🟢 [on PR] the schema is `ErrorValidation` = `{ "<field>": ["<message>"] }` — a flat field→messages map matching the actual API response

### DIFF
--- a/jsapp/js/UniversalTable/index.tsx
+++ b/jsapp/js/UniversalTable/index.tsx
@@ -1,7 +1,6 @@
 import React, { useMemo } from 'react'
 
 import type { UseQueryResult } from '@tanstack/react-query'
-import type { ErrorObject } from 'schema-utils/declarations/validate'
 import type { ErrorDetail } from '#/api/models/errorDetail'
 import UniversalTableCore from './UniversalTableCore'
 import type { UniversalTableColumn } from './UniversalTableCore'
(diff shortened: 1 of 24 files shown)


platreth:
Created a follow up issue : https://linear.app/kobotoolbox/issue/DEV-2467/date-deployed-mock-can-return-true-schema-only-allows-false

rgraber (approved):
LGTM

PR #7277, opened 2026-07-17
fix(project settings): remove deployment check for showing anonymous submission permission option DEV-2460

jamesrkiger:
### 📣 Summary
Removes a check for project deployment status that determined whether or not the checkbox for enabling anonymous users to view submission data was visible, so users can now view/toggle option for undeployed projects.

### 👀 Preview steps
1. Have a project with some submissions 
2. Archive the project
3. Go the the sharing view in project settings 
4. 🔴 [on main] notice that there is no checkbox for "Anyone can view submissions made to this form", regardless of whether or not that option was previously checked.
5. 🟢 [on PR] Verify that checkbox is always visible, regardless of deployment status


pauloamorimbr (approved):
LGTM! 👍🏻

PR #5994, opened 2025-07-25
feat(tag): migrate `tag` endpoints to API `v2`  DEV-800

ZoeGodin:
### 📣 Summary
Migrate the `tag` endpoint from the now deprecated `/tags/` to the `/api/v2/tags/`.

### 👀 Preview steps
1. Execute the container and enter `./manage.py show_urls | grep tags`
2. Make sure the api/v1 and api/v2 appears (`tags/` and `api/v2/tags`)
3. Visit `http://kf.kobo.local/api/v2/tags/`, an api page for tag appears
4. The page at `http://kf.kobo.local/tags/` is still present and suggests the user to visit the `api/v2`

PR #7278, opened 2026-07-17
fix(tags): fix tags API to list only users assets DEV-2464

Guitlle:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
The detail view (/api/v2/tags/{uid}/) incorrectly iterates through the UIDs of all assets in the entire database and generates detail URLs for them. It should only enumerate the specific assets that are associated with the requested tag AND that the requesting user has permission to view.

### 📖 Description
* Modified the `get_assets` method so that it scopes the search space solely to assets the user is authorized to view (`get_objects_for_user(user, PERM_VIEW_ASSET, Asset)`).
* Filter that safe queryset to only return the UIDs for the assets tagged with the specific tag currently being serialized 
* Added an early exit if the user is unauthenticated or there is no request object (returning an empty list).
* Add a unit test to check this behavior

### 👀 Preview steps
1. Have multiple users, assets and tags
2. Go to the detail API view for a tag ` /api/v2/tags/<tag_uid>/` and verify that only the user assets are listed

### DIFF
--- a/kpi/serializers/v2/tag.py
+++ b/kpi/serializers/v2/tag.py
@@ -4,8 +4,10 @@ from rest_framework import serializers
 from rest_framework.reverse import reverse
 from taggit.models import Tag
 
+from kpi.constants import PERM_VIEW_ASSET
 from kpi.models import Asset, TagUid
 from kpi.schema_extensions.v2.tags.fields import ParentUrlField, TagUrlField
+from kpi.utils.object_permission import get_database_user, get_objects_for_user
 
 
 class TagSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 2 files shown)


kpi/serializers/v2/tag.py
```diff
@@ -25,9 +27,15 @@ def get_parent(self, obj):
     @extend_schema_field(OpenApiTypes.OBJECT)
     def get_assets(self, obj):
         request = self.context.get('request', None)
+        if not request or not request.user.is_authenticated:
+            return []
+
```
noliveleger: I would add the permission class to the viewset instead "IsAuthenticated"

noliveleger (changes requested):

jnm:
The problem this PR is trying to fix was introduced by #5994. I've run Greptile on it: https://github.com/kobotoolbox/kpi/pull/5994#issuecomment-5008438450

kpi/serializers/v2/tag.py
```diff
@@ -4,15 +4,19 @@
...
 from kpi.schema_extensions.v2.tags.fields import ParentUrlField, TagUrlField
+from kpi.utils.object_permission import get_objects_for_user
 
 
 class TagSerializer(serializers.ModelSerializer):
     url = serializers.SerializerMethodField()
     assets = serializers.SerializerMethodField()
     parent = serializers.SerializerMethodField()
     uid = serializers.ReadOnlyField(source='taguid.uid')
+    permission_classes = (IsAuthenticated,)
```
jnm, line 19: @greptileai does this do anything on a serializer class?

jnm:
Closing in favor of #7280

PR #7280, opened 2026-07-20
feat(tags)!: remove tag detail endpoint DEV-2464 DEV-2463

jnm:
### 🗒️ Checklist

1. [x] run linter locally
3. [ ] ~~for user-facing doc changes create a Zulip thread at `#Support Docs Updates`, if any~~
10. [x] delete this checklist section from the final squash commit before merging
11. [x] mark Linear issues as done since this doesn't require input from the QA team

### 📣 Summary

Remove `/api/v2/tags/<tag uid>` detail endpoint from the API.

### 📖 Description

Remove `/api/v2/tags/<tag uid>` detail endpoint from the API. All information that was available there can be obtained either from the list endpoint, `/api/v2/tags`, or by querying the asset endpoint, e.g. `/api/v2/assets/?q=tags__name:<tag name>`. Note that `tags` is plural in `tags__name`.

The tags detail API was unused by the application and likely unused by external parties given that it was broken in most cases: whenever a tag was associated with multiple assets accessible to the current user, a 500 error was returned due to missing `distinct()` on the queryset after permission joins.

The detail view also provided an `assets` array that started leaking the URLs of all assets on the server after #5994 was merged.

### 👀 Preview steps

The detail endpoint `/api/v2/tags/<tag uid>` should exist before the changes in this PR and should not exist after them.


static/openapi/schema_v2.yaml
```diff
@@ -17640,8 +17594,8 @@ components:
         result:
           type: string
           format: uri
-          example: https://kf.kobotoolbox.orguser/export/NEW PROJECT - all versions
-            - False - 2025-01-01-01-01.csv
+          example: https://kf.kobotoolbox.orguser/export/NEW PROJECT - all versions - False
```
jnm: Hm, I don't know why this change is here or why there was already a missing `/` between `org` and `user` 🤔

Guitlle: It comes from kpi/schema_extensions/v2/export_tasks/extensions.py 

```
class ResultFieldExtension(OpenApiSerializerFieldExtension):
    target_class = 'kpi.schema_extensions.v2.export_tasks.fields.ResultField'

    def map_serializer_field(self, auto_schema, direction):
        return build_url_type(
            'serve_asset_private_file',
            path='user/export/NEW PROJECT - all versions - False - 2025-01-01-01-01.csv',  # noqa
        )
```

Then in kpi/utils/schema_extensions/url_builder.py the urls_pattern_mapping value doesn't have a trailing `/` ..

jnm:
@greptileai please review

kpi/serializers/v2/tag.py
```diff
@@ -1,45 +1,18 @@
...
-    assets = serializers.SerializerMethodField()
-    parent = serializers.SerializerMethodField()
-    uid = serializers.ReadOnlyField(source='taguid.uid')
+class TagListSerializer(serializers.ModelSerializer):
+    uid = serializers.SerializerMethodField()
 
     class Meta:
         model = Tag
-        fields = ('name', 'url', 'assets', 'parent', 'uid')
+        fields = ('name', 'uid')
```
Guitlle, line 12: Should we keep the parent link?

Guitlle (changes requested):

kpi/serializers/v2/tag.py
jnm: @Guitlle no, it's okay to trash, because `parent` in the old detail endpoint was just a link back to the list view

Guitlle (approved):

PR #7281, opened 2026-07-20
test(storybook): concise CI output DEV-2275

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 💭 Notes

Storybook tests on CI now show only the summary and failures instead of listing every passing test.

Changes here:
- `.storybook/summary-reporter.js` - custom reporter that only prints fails and summary
- Changed `a11y: { test: 'todo' }` → `a11y: { disable: true }` across all story files
  - This prevents a11y warnings from showing up for stories where we're not ready to fix violations yet
  - Stories without this override will still fail if they have a11y violations (the default behavior)
  - `.storybook/preview.tsx` - Added comment clarifying that `a11y: { disable: true }` is the way to go
- Also using quiet flag to avoid hundrets of `"<s> [webpack.Progress] 1%"` lines

The main goal was to stop CI from printing hundreds of "PASS" lines. Now it just shows how many tests passed/failed and how long it took. If a test fails, you still get the full error details - we're only suppressing the noise from passing tests.

We basically got down from around 7800 lines to 4200 (storybook output change) and then further down to 3100 (webpack output change)

### 👀 Preview steps

1. ℹ️ Make a change to any Storybook story file
2. Run `npm run test:storybook` locally
3. 🟢 Notice the new output - no more listing every individual passing test, just summary
4. 🔴 [optional] Break a story to see failures still show up with full details


jamesrkiger (approved):
lgtm

PR #7282, opened 2026-07-20
fix(xlsForm): keep leading "=" in labels as text instead of a formula DEV-1235

platreth:
### 📣 Summary

Question and choice labels that start with an equals sign (e.g. `=foo`) now display and export correctly instead of turning into `0` or a spreadsheet error.

### 📖 Description

If a label began with `=`, it was mistaken for a spreadsheet formula. The deployed form showed `0` in place of the label in Enketo, and exporting the form to XLSForm produced an error (`Err:520`) where the label should be. Labels (and other cells) starting with `=` are now kept as plain text everywhere.

### 💭 Notes

- Root cause: `Asset.to_xlsx_io()` builds the workbook with `xlsxwriter`, whose generic `write()` routes any `=`-prefixed string to `write_formula()`. The deploy path feeds the same in-memory xlsx to pyxform, so one bug corrupted **both** the XLSForm export and the deployed XForm (pyxform reads the formula's cached `0`).
- Fix: pass `{'strings_to_formulas': False}` to the `xlsxwriter.Workbook`. XLSForm cells are never Excel formulas, so this is safe; it's workbook-scoped, so it also covers `+`/`-`/`@` prefixes and the survey, choices and settings sheets.
- Added regression test `test_to_xlsx_io_preserves_leading_equals_in_label`.

### 👀 Preview steps

2. add a text question with the label `=foo`, then deploy the form
3. 🔴 [on main] open the form in Enketo → the label shows `0`; export the form to XLSForm → the cell shows `Err:520`
4. 🟢 [on PR] open the form in Enketo → the label shows `=foo`; export the form to XLSForm → the cell shows `=foo`

### DIFF
--- a/kpi/mixins/xls_exportable.py
+++ b/kpi/mixins/xls_exportable.py
@@ -80,7 +80,11 @@ class XlsExportableMixin:
             # achieve this isolation.
             ss_dict = self.ordered_xlsform_content(**kwargs)
             output = BytesIO()
-            with xlsxwriter.Workbook(output) as workbook:
+            # Keep values like `=foo` as literal strings; XLSForm cells are
+            # never Excel formulas (DEV-1235).
+            with xlsxwriter.Workbook(
+                output, {'strings_to_formulas': False}
+            ) as workbook:
                 for sheet_name, contents in ss_dict.items():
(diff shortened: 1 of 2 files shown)


RuthTurk (approved):
LGTM

PR #7292, opened 2026-07-21
fix(logger): resolve `root_uuid` conflicts in dedup command DEV-2450

noliveleger:
### 📣 Summary

A background data-repair task could fail on forms that contain certain duplicate submissions, leaving those forms unrepaired. It now resolves the conflict and continues.

### 💭 Notes

The `clean_duplicated_submissions` management command (run by long-running migration 0027) could raise `unique_root_uuid_per_xform` while assigning `root_uuid`, which aborted the whole form and left it flagged as failed.

Two root causes are fixed:
- `_replace_duplicates` gave a divergent duplicate a new unique `uuid` but left its `root_uuid` pointing at the shared value, because `_populate_root_uuid()` is a no-op once `root_uuid` is already set. It now sets `root_uuid` to the new uuid and updates the XML `rootUuid` node when that node exists.
- `_delete_duplicates` backfilled the reference `root_uuid` without handling a collision with another submission in the same form.

Both now fall back to a shared helper, `resolve_root_uuid_conflict()` in `logger.utils.instance`, which assigns a unique `CONFLICT-<uuid4>-...` `root_uuid` (hashed when it would exceed the column length, and never using the sequential pk) and keeps Postgres, the XML `rootUuid` node (when present) and the Mongo `meta/rootUuid` in sync. `_resolve_conflict_by_pk` in `clean_duplicated_submissions_root_uuid` now uses the same helper instead of duplicating the logic.

The conflicting `.update()` runs in its own `kc_transaction_atomic` block so a caught `IntegrityError` cannot poison an enclosing transaction: it is a savepoint when the command runs inside a transaction (for example under `TestCase`, which is what surfaced this), or its own transaction when run standalone in autocommit. Without it, the conflict resolution below would run on an aborted transaction. In `_replace_duplicates`, only the identity fields change, so after the Postgres bulk update Mongo is synced with a single bulk `$set` on `_uuid`, `meta/instanceID` and `meta/rootUuid` (on the existing docs) instead of a full `update_mongo()` per instance. Syncing after Postgres also means a failed write never leaves Mongo ahead of the database.

Mongo `meta/rootUuid` is built from the DB `root_uuid` field, not from the XML, so the backfill stays correct for submissions that have no `rootUuid` node in their XML (submissions that were never edited).

### 👀 Preview steps

It's easier to refer to unit tests.

### DIFF
--- a/kobo/apps/openrosa/apps/logger/tests/test_clean_duplicated_submissions.py
+++ b/kobo/apps/openrosa/apps/logger/tests/test_clean_duplicated_submissions.py
@@ -436,20 +436,29 @@ class TestCleanDuplicatedSubmissions(TestBase):
         # Hash fallback: "CONFLICT-" + a 64-char sha256 hexdigest.
         self.assertEqual(len(inst.root_uuid), len('CONFLICT-') + 64)
 
-    def test_replace_duplicates_syncs_mongo_root_uuid(self):
+    def test_replace_duplicates_syncs_mongo_identity_fields(self):
         """
-        Even without a rootUuid node in the XML (a non-edited submission), the
-        Mongo `meta/rootUuid` must be backfilled from the new root_uuid, since
-        it is built from the DB field, not the XML.
+        The renamed identity fields (`_uuid`, `meta/instanceID`,
+        `meta/rootUuid`) must be updated on the existing Mongo documents.
         """
 
         self._make_group_b()
+        # Submissions already exist in Mongo before the repair runs.
+        for pk in (self.b1.pk, self.b2.pk, self.b3.pk):
+            Instance.objects.get(pk=pk).parsed_instance.update_mongo(
+                asynchronous=False
+            )
+
         call_command('clean_duplicated_submissions', xform=self.xform.id_string)
 
         for pk in (self.b1.pk, self.b2.pk, self.b3.pk):
             instance = Instance.objects.get(pk=pk)
             doc = settings.MONGO_DB.instances.find_one({'_id': pk})
             self.assertIsNotNone(doc)
+            self.assertEqual(doc['_uuid'], instance.uuid)
+            self.assertEqual(
+                doc['meta/instanceID'], add_uuid_prefix(instance.uuid)
+            )
             self.assertEqual(
                 doc['meta/rootUuid'], add_uuid_prefix(instance.root_uuid)
             )
--- a/kobo/apps/openrosa/apps/logger/tests/test_clean_duplicated_submissions.py
+++ b/kobo/apps/openrosa/apps/logger/tests/test_clean_duplicated_submissions.py
@@ -1,6 +1,7 @@
 import os
 import uuid as uuid_module
 
+from django.conf import settings
 from django.core.files.base import ContentFile
 from django.core.files.uploadedfile import SimpleUploadedFile
 from django.core.management import call_command
@@ -13,6 +14,8 @@ from kobo.apps.openrosa.apps.logger.models import (
     SurveyType,
 )
 from kobo.apps.openrosa.apps.logger.models.attachment import Attachment
+from kobo.apps.openrosa.apps.logger.utils import resolve_root_uuid_conflict
+from kobo.apps.openrosa.apps.logger.xform_instance_parser import add_uuid_prefix
 from kobo.apps.openrosa.apps.main.tests.test_base import TestBase
 from kobo.apps.openrosa.apps.viewer.models.parsed_instance import ParsedInstance
 from kobo.apps.openrosa.libs.utils.logger_tools import publish_xls_form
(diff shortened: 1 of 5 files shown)


kobo/apps/openrosa/apps/logger/tests/test_clean_duplicated_submissions.py
```diff
@@ -84,39 +90,32 @@ def _create_instance(self, xml, instance_uuid):
...
+    def _make_xml(self, instance_uuid, name='Test'):
+        """
+        Minimal valid XML submission for the tutorial form.
+        """
+        return (
+            f'<?xml version="1.0" ?>'
+            f'<{self.xform.id_string} id="{self.xform.id_string}">'
+            f'<meta><instanceID>uuid:{instance_uuid}</instanceID></meta>'
+            f'<name>{name}</name>'
+            f'</{self.xform.id_string}>'
```
noliveleger, line 102: No changes in this PR. Only A->Z reordering

kobo/apps/openrosa/apps/logger/tests/test_clean_duplicated_submissions.py
```diff
@@ -51,18 +54,21 @@ def setUp(self):
...
+            instance=instance,
+            xform=self.xform,
+            media_file=fake_file,
+            mimetype='text/plain',
+            media_file_size=7,
+        )
+
+    def _add_parsed_instance(self, instance):
+        parsed, _ = ParsedInstance.objects.get_or_create(instance=instance)
+        return parsed
```
noliveleger, line 71: No changes in this PR. Only A->Z reordering

kobo/apps/openrosa/apps/logger/tests/test_clean_duplicated_submissions.py
```diff
@@ -84,39 +90,32 @@ def _create_instance(self, xml, instance_uuid):
...
+
+        return (
+            f'<?xml version="1.0" ?>'
+            f'<{self.xform.id_string} id="{self.xform.id_string}">'
+            f'<meta>'
+            f'<instanceID>uuid:{instance_uuid}</instanceID>'
+            f'<rootUuid>uuid:{root_uuid}</rootUuid>'
+            f'</meta>'
+            f'<name>{name}</name>'
+            f'</{self.xform.id_string}>'
```
noliveleger, line 118: New code from this PR.

kobo/apps/openrosa/apps/logger/tests/test_clean_duplicated_submissions.py
```diff
@@ -144,6 +143,24 @@ def _make_group_b(self):
...
+            user_id=self.user.pk,
+            xform_id=self.xform.pk,
+            defaults={'counter': count},
+        )
+        DailyXFormSubmissionCounter.objects.update_or_create(
+            date=today,
+            user_id=self.user.pk,
+            xform_id=self.xform.pk,
+            defaults={'counter': count},
+        )
```
noliveleger, line 163: No changes in this PR. Only A->Z reordering

kobo/apps/openrosa/apps/logger/tests/test_clean_duplicated_submissions.py
```diff
@@ -324,3 +341,115 @@ def test_replace_duplicates_tolerates_missing_parsed_instance(self):
         self.assertEqual(updated.count(), 2)
         for instance in updated:
             self.assertTrue(instance.uuid.startswith('DUPLICATE-'))
+
+    # ------------------------------------------ root_uuid conflict resolution
```
noliveleger, line 345: New tests below!

kobo/apps/openrosa/apps/logger/utils/instance.py
```diff
@@ -228,6 +231,47 @@ def remove_validation_status_from_instance(instance: Instance) -> bool:
...
+    else:
+        fields['xml'] = instance.xml
+        fields['xml_hash'] = instance.get_hash(instance.xml)
+
+    Instance.objects.filter(pk=instance.pk).update(**fields)
+
+    doc = settings.MONGO_DB.instances.find_one({'_id': instance.pk})
+    if doc is not None:
+        doc['meta/rootUuid'] = add_uuid_prefix(conflict_root_uuid)
+        settings.MONGO_DB.instances.replace_one({'_id': instance.pk}, doc)
```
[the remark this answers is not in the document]
noliveleger: @greptileai Fixed in [5bbba80](https://github.com/kobotoolbox/kpi/pull/7292/commits/5bbba802f0d6e164c9b64c4ee34965bbf77d78d4)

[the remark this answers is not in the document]
noliveleger: @greptileai it matches with the logic of the management commands which handle `root_uuid` backfill. 
The long-running migration 0028 handle cases where `meta/rootUuid` is absent from MongoDB documents but present in PostgreSQL.

noliveleger:
@greptileai review

Guitlle (approved):
LGTM

PR #7293, opened 2026-07-21
feat(designSystem): use mantine Tabs component and deprecate our old component DEV-2418

duvld:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Replace old tabs component with updated one

### 💭 Notes
Replaced old stories and added new variant "bubble". Replaced the only usage of the old tabs component (`usageTopTabs.tsx`) with the mantine component.

### 👀 Preview steps

1. ℹ️ have an account and a project with stripe enabled
2. navigate to Settings -> Usage
5. 🟢 [on PR] notice that the new tabs function the same
7. 🟢 notice that the new tabs match the design system mocks


jsapp/js/theme/kobo/Tabs.ts
```diff
@@ -0,0 +1,21 @@
+import type { TabsVariant } from '@mantine/core'
+import { Tabs } from '@mantine/core'
+import classes from './Tabs.module.css'
```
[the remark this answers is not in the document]
magicznyleszek: @greptile-apps jsapp/js/theme/kobo/Tabs.ts follows the existing pattern in jsapp/js/theme/kobo/ directory. Can you tell if there is some difference here, or if all the files have similar problem?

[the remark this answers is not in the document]
magicznyleszek: @greptile-apps sure

jamesrkiger:
A couple things about the style here:

1. The lighter coloring on the line under the non-selected tab is visually confusing. I don't think the designs require this, do they? 

image

It looks a lot clearer with just two colors like in the [Mantine docs](https://mantine.dev/core/tabs/):

image

2. We also seem to not have any color change on these tabs on hover (compare to mantine docs). I think that would be good to have.

jamesrkiger (changes requested):
Left a comment on styling

jsapp/js/theme/kobo/Tabs.ts
```diff
@@ -0,0 +1,21 @@
...
+
+type TabsVariantCustom = Extract<TabsVariant, 'default' | 'pills'> | 'bubble'
+type TabsSizeCustom = 'sm' | 'md' | 'lg'
```
magicznyleszek: Should we export those types and use them in stories file?

magicznyleszek (changes requested):
Generally looks fine

jsapp/js/theme/kobo/Tabs.ts
duvld: I thought this would be trivial to change, but I couldn't get it to work. Maybe I don't understand what you mean, but I don't think we can, according to the robot:
>TypeScript types are erased at compile time, so you can't generate a runtime array from a type.
The relationship only works the other way: make the as const array the source of truth and derive the type from it with typeof … [number]

Which probably isn't what we want

magicznyleszek: I just thought about this: https://github.com/kobotoolbox/kpi/pull/7293/commits/4de1a4eeef6016dfa965b602e6b2551a7c65f408

magicznyleszek (approved):

jamesrkiger (approved):
lgtm

PR #7294, opened 2026-07-22
fix(projectOwnership): stop false transfer failures and improve admin transfer details DEV-2080

platreth:
### 📣 Summary

Project ownership transfers are no longer reported as failed when the transfer actually completed successfully.

### 📖 Description

Transferring projects could end as "failed" even though everything had moved correctly — a single already-deleted file among thousands was enough. A file that no longer exists cannot be moved, so it no longer counts as a failure, and the same now applies when an internal retry re-runs a step that had already finished. Genuine problems are still reported, and a new log page shows what happened for a given project.

### 💭 Notes

- `ExtendedFieldFile.move()` raises `SourceFileMissingError` when the source is gone; the two movers skip it instead of failing. A genuine error still fails the task.
- `TransferStatus.update_status()` makes `success` terminal, so a late or duplicate retry cannot downgrade a finished sub-task.
- A missing source can also mean an interrupted move (file relocated, row never saved). `_recover_moved_file()` checks the target and completes it — otherwise the record would keep pointing at a path holding no file.
- `TransferStatusError.level` (migration `0008`) marks skips as `info`: not counted on the invite page, not shown on the transfer page, listed in the log admin. Indexed — skips vastly outnumber errors, so filtering on `error` is selective — with the index gated on `SKIP_HEAVY_MIGRATIONS`.
- 1000-transfer check: the invite summary stays at 2 queries, the changelist status column at 0.

`validate_org_transfer.py` is deliberately not included; it stays a draft.

### 👀 Preview steps

1. ℹ️ have a superuser and an invite whose transfers include both a skipped file and a genuine error
2. go to `/admin/project_ownership/invite/<id>/change/`
3. 🔴 [on main] every transfer is dumped inline with its statuses and up to 100 error records
4. 🟢 [on PR] one line — `N transfer(s): X failed · Y success — Z error record(s).` — plus `View transfers` and `View logs`
5. open a transfer that succeeded but skipped files
6. 🔴 [on main] the skipped files are listed as red errors
7. 🟢 [on PR] every sub-status simply reads `success`
8. click `View logs`
9. 🟢 [on PR] a read-only log filtered to that invite, filterable by level and searchable per project


kobo/apps/project_ownership/models/transfer.py
```diff
@@ -461,3 +463,10 @@ class TransferStatusError(AbstractTimeStampedModel):
         TransferStatus, related_name='errors', on_delete=models.CASCADE
     )
     error = models.CharField(null=True, blank=True)
+    # No index: two values is not selective, and `CREATE INDEX` would lock
+    # writes on a large table.
```
noliveleger: We can use "SKIP_HEAVY_MIGRATIONS"  and use "CREATE INDEX CONCURRENTLY ..." 
Several examples in the base code.

kobo/apps/project_ownership/admin.py
```diff
@@ -1,12 +1,22 @@
...
 from .models.invite import InviteType
-from .models.transfer import TransferStatusTypeChoices
+from .models.transfer import (
+    TransferStatus,
+    TransferStatusError,
+    TransferStatusTypeChoices,
+)
+
+# Max records rendered inline per status.
+MAX_DISPLAYED_RECORDS = 100
```
noliveleger: Let's move this to settings (and make the name more obvious).

kobo/apps/project_ownership/admin.py
```diff
@@ -152,3 +162,168 @@ def _get_cached_asset_names_by_invite(
...
+        'date_created',
+    )
+    list_select_related = ('asset', 'invite')
+    search_fields = ('uid', 'asset__uid', 'asset__name', 'invite__uid')
+    readonly_fields = ('uid', 'asset', 'invite', 'invite_type', 'get_statuses')
+    fields = ('uid', 'asset', 'invite', 'invite_type', 'get_statuses')
+
+    def lookup_allowed(self, lookup, value, request=None):
+        # Allow the deep-link from InviteAdmin.get_transfers.
+        return lookup == 'invite__id__exact' or super().lookup_allowed(
```
noliveleger: FYI: `invite__id__exact` , you could use `invite_id=` instead of `invite__id__exact`. 

beware of the double underscore in front the `id`. Django is smart enough to avoid to create the JOIN but the, by convention, I would use only `invite_id` instead of `invite__id` if we don't need to traverse the `Invite` model.

kobo/apps/project_ownership/utils.py
```diff
@@ -117,23 +122,37 @@ def move_attachments(transfer: 'project_ownership.Transfer'):
...
+                            f'(data already gone)',
+                            level=TransferStatusErrorLevelChoices.INFO,
+                        )
+                        logging.warning(
+                            f'Source file {media_file_path} (#{attachment.pk}) '
+                            f'no longer exists — skipped'
+                        )
+                if moved:
                     update_fields.append('media_file')
                     _delete_thumbnails(media_file_path)
```
noliveleger: _delete_thumbnails should run even if the file was not moved. Just in case there are still present on the storage.

kpi/fields/file.py
```diff
@@ -51,9 +58,9 @@ def move(self, target_folder: str, reraise_errors: bool = False) -> bool:
             self.storage.delete(old_path)
             success = True
         except FileNotFoundError as fe:
-            logging.error(fe, exc_info=True)
+            logging.warning(f'Source file {old_path} no longer exists: {fe}')
```
noliveleger, line 63: I think we are logging twice quite the same error: 

```
worker_low_priority-1  | WARNING 2026-07-22 17:57:29,933 file 30 281473196761120 Source file user/attachments/42597303a4904c9782a5aa0171dac76b/edc312c61a6b4b05acb97d2bc6251491/image_3.png no longer exists: An error occurred (404) when calling the HeadObject operation: Not Found
worker_low_priority-1  | WARNING 2026-07-22 17:57:29,976 utils 30 281473196761120 Source file user/attachments/42597303a4904c9782a5aa0171dac76b/edc312c61a6b4b05acb97d2bc6251491/image_3.png (#28976) no longer exists — skipped
```


even if the message is a little bit different, it's redundant. Let's remove one of the other (leave a comment to explain why and where it's logged).

kobo/apps/project_ownership/admin.py
```diff
@@ -152,3 +162,168 @@ def _get_cached_asset_names_by_invite(
...
+                        status_type=TransferStatusTypeChoices.GLOBAL
+                    ),
+                    to_attr='_global_statuses',
+                )
+            )
+        )
+
+    @admin.display(description='Asset')
+    def get_asset(self, obj):
+        return f'{obj.asset.name} #{obj.asset.uid}'
```
noliveleger, line 214: Be sure to not load `asset.content`  (especially if `get_asset` is called in a loop). We try to avoid loading this field if it's not needed since it can be pretty big.

noliveleger (changes requested):

noliveleger:
I cannot really comment in the PR but I would use more obvious "verbose_names" in the menu

Screenshot 2026-07-22 at 15 21 06

What about:
- Invites
- Transfers
- Logs 

To match the links you just added in the PR? 

- Remove the "Projects" column in Invites, use the uid of the invite instead (maybe projects count?)
- Replace the recipient username with the Organization name when the type is Org membership

kobo/apps/project_ownership/models/choices.py
```diff
@@ -22,6 +22,16 @@ class TransferStatusChoices(models.TextChoices):
...
 
 
+class TransferStatusErrorLevelChoices(models.TextChoices):
+    """
+    `INFO` records an intentional skip, e.g. a source file that no longer
+    exists. It is not a failure.
+    """
+
+    ERROR = 'error'
+    INFO = 'info', 'Skipped'
```
noliveleger: I think we should have "Warning" and use this level instead of "Info" when files are skipped.
I would keep "Info" as the label. It could be use for something else in the future.

kobo/apps/project_ownership/migrations/0008_alter_invite_options_alter_transfer_options_and_more.py
```diff
@@ -0,0 +1,80 @@
...
+                default='error',
+                max_length=5,
+            ),
+        ),
+        (
+            migrations.RunPython(
+                manually_create_indexes_instructions,
+                manually_drop_indexes_instructions,
+            )
+            if settings.SKIP_HEAVY_MIGRATIONS
```
noliveleger: I know I told you we were using `SKIP_HEAVY_MIGRATIONS` but do you have any idea if it is really needed? 
Will the index creation take a long time on production? (more than 5 minutes). If not, let's Django handle it.

kobo/apps/project_ownership/models/transfer.py
```diff
@@ -415,14 +416,10 @@ def update_status(
...
                 transfer_id=transfer_id, status_type=status_type
             )
-            if (
-                transfer_status.status == TransferStatusChoices.SUCCESS
-                and status != TransferStatusChoices.SUCCESS
-            ):
-                cls._add_error(
-                    transfer_status,
-                    f'Updating status of previously successful transfer to {status}',
-                )
```
noliveleger, line 425: What about logging a warning here at least (instead of adding the error)?

noliveleger (changes requested):

kobo/apps/project_ownership/migrations/0008_alter_invite_options_alter_transfer_options_and_more.py
platreth: Yeah you're right, I benchmarked it locally to be sure: a scratch clone of the table at 5M rows builds both indexes in ~7s total on my local stack, ~1.6s at 1M, so I dropped the SKIP_HEAVY_MIGRATIONS ceremony and kept a plain db_index=True!

noliveleger (approved):
LGTM

PR #7300, opened 2026-07-23
feat(assetVersion): add version_number to `/api/v2/assets/<uid>/versions/` DEV-1825

rajpatel24:
### 📣 Summary
This PR adds a `version_number` field to the asset versions API, returning "12" for deployed versions and "11.4" for undeployed versions (numbered relative to the deployment that precedes them), computed server-side so the numbering stays correct regardless of pagination.

#### Why this needed to be done in the backend
The Form History UI (Project → Form tab) needs to label each version like v12, v11.4, etc. - deployed versions get a whole "major" number, and undeployed (draft) versions get a "minor" number counting up since the last deployment.

The frontend currently derives this from array index, which only works because the full list is loaded at once. We're moving this list to paginated infinite scroll, and a version's correct number depends on the entire history of the asset before it - not just the versions on the current page. The frontend has no way to know, from a single page of results, how many deployments happened earlier in history, so it can't reliably compute this itself once pagination is introduced. 

#### What changed
Added a `version_number` field to the versions serializer. For each request, we fetch the asset's full version history as a lightweight (id, deployed) list in a single ordered query, then walk it once in Python to build a map of version id → label: deployed versions increment the major number and reset the minor, undeployed versions increment the minor. The serializer looks up each row's label from that map, so both the paginated list endpoint and retrieve return identical, pagination-independent numbers with no per-row queries.

#### Why computed on the fly instead of stored on the model
I considered storing these numbers as columns on AssetVersion, populated at save time like the existing _content_hash field, but ruled it out because the table has millions of rows and would need a long, risky backfill migration for a display-only field. Version counts per asset are in the low hundreds in practice, so fetching two integer columns for one asset's history and looping over them is negligible work - it requires no migration and is automatically correct for all existing data. If per-asset version counts ever grow into the tens of thousands, the stored-column approach becomes worth revisiting; at current scale it isn't.

### 👀 Preview steps

1. ℹ️ have an account and a form
2. deploy the form, edit it a few times (saving each time without deploying), then deploy it again, and make one more edit afterward without deploying
4. 🔴 [on main] notice that `/api/v2/assets/<uid>/versions/` doesn't return anything regarding version numbers
5. 🟢 [on PR] notice that `/api/v2/assets/<uid>/versions/` returns a version_number field on each version - a whole number (e.g. "2") for deployed versions, and a major.minor number (e.g. "2.1") for undeployed versions, numbered relative to the deployment that precedes them
6. Verify the retrieve endpoint as well.

### DIFF
--- a/kpi/serializers/v2/asset_version.py
+++ b/kpi/serializers/v2/asset_version.py
@@ -26,10 +26,10 @@ class AssetVersionListSerializer(serializers.Serializer):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        # Memo of `{asset_id: {version_id: label}}`, built lazily per asset by
-        # `get_version_number`. The asset is taken from `obj`, not the view, so
-        # this works both for the versions endpoints and when this serializer is
-        # the nested `deployed_versions` field of an asset (whose view is
+        # Memo of `{asset_id: {version_id: label}}`, filled lazily per asset by
+        # `_build_version_labels`. The asset is taken from `obj`, not the view,
+        # so this works both for the versions endpoints and when this serializer
+        # is the nested `deployed_versions` field of an asset (whose view is
         # `AssetViewSet`, which has no `asset_uid`). Keyed by `asset_id` so one
         # instance stays correct even if reused across assets
         self._version_labels = {}
@@ -37,7 +37,7 @@ class AssetVersionListSerializer(serializers.Serializer):
     def get_date_deployed(self, obj):
         return obj.deployed and obj.date_modified
 
-    def get_version_number(self, obj):
+    def get_version_number(self, obj) -> str:
         """
         Human-readable version number (see `_build_version_labels`), consistent
         across pages, `?deployed=` filters, and the `retrieve` action
@@ -46,11 +46,10 @@ class AssetVersionListSerializer(serializers.Serializer):
         history) and memoized on this serializer instance, so serializing a
         whole page adds one query, not one per row. The lookup is keyed on `id`.
         """
-        labels = self._version_labels.get(obj.asset_id)
-        if labels is None:
-            labels = self._version_labels[obj.asset_id] = (
-                self._build_version_labels(obj.asset_id)
-            )
+        try:
+            labels = self._version_labels[obj.asset_id]
+        except KeyError:
+            labels = self._build_version_labels(obj.asset_id)
         return labels[obj.id]
 
     def get_url(self, obj):
@@ -60,11 +59,11 @@ class AssetVersionListSerializer(serializers.Serializer):
             request=self.context.get('request', None),
         )
 
-    @classmethod
-    def _build_version_labels(cls, asset_id):
+    def _build_version_labels(self, asset_id: int) -> dict[int, str]:
         """
-        Build `{version_id: "1.2"}` for every version of the asset, in a single
-        query and one ordered pass over its full history
+        Build, cache, and return `{version_id: "1.2"}` for every version of the
+        asset, from a single query and one ordered pass over its full history.
+        The result is stored in `self._version_labels[asset_id]`
 
         Deployed versions get a major number, e.g. "12": the 1st deployment
         (chronologically) is "1", the 12th is "12". Undeployed versions get a
@@ -91,6 +90,7 @@ class AssetVersionListSerializer(serializers.Serializer):
             else:
                 minor += 1
                 labels[version_id] = f'{major}.{minor}'
+        self._version_labels[asset_id] = labels
         return labels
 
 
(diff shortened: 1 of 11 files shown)


kpi/serializers/v2/asset_version.py
```diff
@@ -21,10 +22,66 @@ class AssetVersionListSerializer(serializers.Serializer):
...
+        # Per-request memo of each asset's full-history label map, filled lazily
+        # by `get_version_number`. Always a single asset in practice, but keyed
+        # by `asset_id` to stay correct regardless
+        self._version_labels = {}
 
     def get_date_deployed(self, obj):
         return obj.deployed and obj.date_modified
 
+    @classmethod
+    def _labels_for_asset(cls, asset_id):
```
noliveleger: Move it after public methods ;-)

kpi/serializers/v2/asset_version.py
```diff
@@ -21,10 +22,66 @@ class AssetVersionListSerializer(serializers.Serializer):
...
+
+        The full-history label map is computed once per serializer instance and
+        memoized per asset (see `__init__`), so serializing a whole page costs
+        one extra query, not one per row. The lookup is keyed on `id`.
+        """
+        if obj.asset_id not in self._version_labels:
+            self._version_labels[obj.asset_id] = self._labels_for_asset(
+                obj.asset_id
+            )
+        return self._version_labels[obj.asset_id][obj.id]
```
noliveleger: I think it's confusing: At the first glance I thought it would enter the condition more than once. 
Actually, the `if` is only to initialize `self._version_labels` .

What about renaming `_labels_for_asset` to `version_labels` , make it a property of the class. It `self._version_labels` has an instance cache.  

Something like 

```diff
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        # Per-request memo of each asset's full-history label map, filled lazily
        # by `get_version_number`. Always a single asset in practice, but keyed
        # by `asset_id` to stay correct regardless
        self._version_labels = None

-    @classmethod
-    def _labels_for_asset(cls, asset_id):
+    @property 
+    def version_labels(self, asset_id):
+         if self._version_labels is not None:
+              return self._version_labels
+
           rows = ( 
      ... 

      def get_version_number(self, obj):
-        if obj.asset_id not in self._version_labels:
-            self._version_labels[obj.asset_id] = self._labels_for_asset(
-                obj.asset_id
-            )
-        return self._version_labels[obj.asset_id][obj.id]
+        return self.version_labels[obj.asset_id][obj.id]
      
```


Moreover, you could even avoid `obj.asset_id`  and use `self.context['view'].asset_uid` directly in `version_labels` since `AssetVersionListSerializer` is called from the ViewSet which inherits from `AssetNestedObjectViewsetMixin`. 

so something like 

```
        asset_uid = self.context['view'].asset_uid
        rows = (
            AssetVersion.objects.filter(asset__uid=asset_uid)
            .order_by('date_modified', 'id')
            .values_list('id', 'deployed')
        )
```

noliveleger (changes requested):

kpi/serializers/v2/asset_version.py
rajpatel24: Using `version_labels` as a property with `self.context['view'].asset_uid` works great for the `/versions/` endpoints, but it breaks on GET `/api/v2/assets/<uid>/`. This serializer is also used there as the nested `deployed_versions` field, where the view in context is `AssetViewSet` which doesn't inherit `AssetNestedObjectViewsetMixin` and so has no `asset_uid`. That makes `context['view'].asset_uid` raise `AttributeError`, so the asset detail endpoint 500s. Reading the `asset` from `obj.asset_id` instead avoids the coupling and works in both places.

noliveleger: Sorry about that ;-)

kpi/serializers/v2/asset_version.py
```diff
@@ -26,16 +26,42 @@ class AssetVersionListSerializer(serializers.Serializer):
...
+        The label map is built once per asset (a single query over its full
+        history) and memoized on this serializer instance, so serializing a
+        whole page adds one query, not one per row. The lookup is keyed on `id`.
+        """
+        labels = self._version_labels.get(obj.asset_id)
+        if labels is None:
+            labels = self._version_labels[obj.asset_id] = (
+                self._build_version_labels(obj.asset_id)
+            )
+        return labels[obj.id]
```
noliveleger: Let's avoid this dual assignment. 

```python
try: 
    labels = self._version_labels[obj.asset_id]
except KeyError:
    labels = self._build_version_labels(obj.asset_id)  # init `self._version_labels` inside this method

return labels[obj.id] 

... 


def _build_version_labels(cls, asset_id: int) -> dict[str]:
-    labels = {}
...
-   labels[version_id] = str(major)
+   self._version_labels[version_id] = str(major)

```

noliveleger (changes requested):

noliveleger (approved):
LGTM

PR #7301, opened 2026-07-23
refactor(map): remove leaflet-omnivore DEV-857

jamesrkiger:
### 📣 Summary
Replaces defunct leaflet-omnivore package with more targeted dependencies for parsing map overlay data, while also removing support for .wkt files in this context. 

### 💭 Notes
GeoJSON overlays are currently broken on main. The validation path used fetchGetUrl, which only parses the response body when the Content-Type is application/json. GeoJSON files are served as application/geo+json, so the body was silently discarded, causing validation to always fail. Switching to fetch + response.text() fixes this.

Security: Map overlay popups previously passed feature property values directly to Leaflet's bindPopup as HTML strings. Property values from uploaded files are now set via textContent, eliminating an XSS vector.

### 👀 Preview steps

1. Create a project with at least one geo question and make a submission
2. Go to data table and click on "Map"
3. Add the files in this archive as overlays:
[test-overlays.zip](https://github.com/user-attachments/files/30307199/test-overlays.zip)
4. Verify that overlays are properly imported, loaded and displayed (three of the files share the same data, so you will want to toggle the overlays on/off to see each one).
5. Try adding some invalid files from this archive:
[error-tests.zip](https://github.com/user-attachments/files/30308432/error-tests.zip)
6. Verify that errors result in a simple toast notification as before and do not break the map or valid overlays.


jsapp/js/components/map/index.tsx
```diff
@@ -1,10 +1,8 @@
 import { Dialog, Group, Stack, type TooltipProps } from '@mantine/core'
-// Leaflet
-// TODO: use something diifferent than leaflet-omnivore as it is not maintained
-// and last realease was 8(!) years ago.
-import omnivore, { type OmnivoreFunction } from '@mapbox/leaflet-omnivore'
+import { kml as toGeoJsonKml } from '@tmcw/togeojson'
 import JSZip from 'jszip'
-import L, { type LayerGroup } from 'leaflet'
+import L from 'leaflet'
+import Papa from 'papaparse'
```
magicznyleszek, line 5: Lol :D

jsapp/js/components/map/index.tsx
```diff
@@ -334,126 +339,125 @@ class FormMap extends React.Component<FormMapProps, FormMapState> {
...
-            .then((zip) => zip.file('doc.kml')?.async('string'))
-            .then((kmlContent) => {
-              if (kmlContent && this.state.map) {
-                // We don't need to react to `.on('ready')` here, as KML file is already loaded and we just need to
-                // parse it (works synchronously)
-                const parsedOverlayLayer = omnivore.kml.parse(kmlContent)
-                this.onOmnivoreLayerReady(parsedOverlayLayer, layer.description)
+      this.loadOverlayLayer(layer).catch((err) => {
+        console.error(err)
+        notify.error(OVERLAY_ERROR.replace('##name##', layer.description))
```
magicznyleszek: I see in console that the errors provide some useful information, can we use them here?

magicznyleszek (changes requested):

magicznyleszek:
Generally works great and code looks fine, one thing to improve

PR #7307, opened 2026-07-24
fix(submissions): use correct Mongo indexes on `root_uuid`/`_uuid` fields INFRA-551

noliveleger:

This is the application-side counterpart to kobotoolbox/kobo-docker#381, which reworks the MongoDB indexes on the `instances` collection (drops indexes that only backed dead code, adds indexes matching how the app actually queries `meta/rootUuid`/`_uuid`). Both PRs should land together on an instance.

Two changes here:

- `kobo/apps/long_running_migrations/jobs/0028_sync_mongo_root_uuid.py`: simplifies the "still missing root_uuid" query to a plain `meta/rootUuid: {$exists: False}` (was also matching `None`/`''`, which the backfill never produces), matching the shape of the new `id_rootUuid_idx` index from kobo-docker#381. Also tightens `_check_lrm_0027_is_completed` to require LRM 0027's status to be strictly `completed` before 0028 runs, instead of accepting `completed` or `failed`.
- `kpi/views/v2/data.py` (`_get_submission_by_id_or_root_uuid`): the legacy UUID-lookup fallback now queries `_uuid` instead of `meta/instanceID`, matching the `{_userform_id, _uuid}` index added in kobotoolbox/kobo-docker#381 rather than relying on an index we're retiring.

### 💭 Notes

Related to kobotoolbox/kobo-docker#381.

### DIFF
--- a/kobo/apps/long_running_migrations/jobs/0028_sync_mongo_root_uuid.py
+++ b/kobo/apps/long_running_migrations/jobs/0028_sync_mongo_root_uuid.py
@@ -1,5 +1,4 @@
 from django.conf import settings
-from django.db.models import Q
 from pymongo import UpdateOne
 
 from kobo.apps.long_running_migrations.exceptions import (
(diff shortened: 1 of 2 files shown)


platreth (approved):
LGTM

PR #7308, opened 2026-07-24
chore(node): drop support for node 20 DEV-2451

jamesrkiger:
### 💭 Notes
Drop support for Node 20 for more efficient frontend dependency management.

Changes:
- Dockerfile: updates both build-stage base images (npm-install, webpack-build-prod), the nodesource apt repo (setup_20.x → setup_22.x), and the pinned apt package version
- .nvmrc / .node-version: updated to v22.23.1
- package.json engines: ^20.19.0 || ^22.12.0 → ^22.23.1
- scripts/hints.js: updated the hardcoded ok_node/ok_npm values that were still checking for Node 20 and warning on mismatch
- All CI workflows now use `node-version-file: '.node-version'` instead of hardcoded version strings, making .node-version the single source of truth. The one exception is storybook-deploy.yml, which uses a Docker container image directly and can't use node-version-file. npm-test.yml's build matrix is removed since we no longer need to test multiple version.


.node-version
```diff
@@ -1 +1 @@
-v20.19.0
+v22.17.0
```
magicznyleszek: Shouldn't we use v22.18.0? Or try using v24.18.0 and see if it's working?

```
npm i
npm warn EBADENGINE Unsupported engine {
npm warn EBADENGINE   package: 'orval@7.21.0',
npm warn EBADENGINE   required: { node: '>=22.18.0' },
npm warn EBADENGINE   current: { node: 'v22.17.0', npm: '10.9.2' }
npm warn EBADENGINE }
```

magicznyleszek (changes requested):

.node-version
magicznyleszek: Also given we have the node version in so many files, do you think there is a way to reuse .node-version in some of them?

magicznyleszek: I spend some more time looking at that and pushed a commit that reuses `.node-version` file

magicznyleszek: Unfortunately we still need to keep both `.nvmrc` and `.node-version` as there is no consensus in the community as to what should be the standard. There is a 10 years old (ongoing) NodeJS issue about this: https://github.com/nodejs/version-management/issues/13 and 9 years old issue at nvm: https://github.com/nvm-sh/nvm/issues/794

PR #7314, opened 2026-07-27
feat(openapiValidator): add OpenAPI request/response validation middleware DEV-1740

platreth:
### 📣 Summary

The API and its published documentation are now checked against each other automatically, so the API documentation stays accurate.

### 📖 Description

KoboToolbox publishes a description of its API that integrations and third-party tools rely on. Until now nothing verified that the API really matches it, so the two could drift apart unnoticed. Requests and responses are now compared against that description automatically and any difference is reported to the team, so it gets fixed instead of reaching the documentation. Nothing changes in how the product behaves today.

### 💭 Notes

Proof of concept for DEV-1740, picked up from Olivier's branch (~700 commits behind main at the time).

- New app `kobo.apps.openapi_validator`: middleware validating requests and responses against `static/openapi/schema_v2.json`.
- Test settings only. `OPENAPI_VALIDATION` is hard-coded off in `base.py` and on in `testing.py`, and the middleware refuses to load unless `settings.TESTING` is true (`MiddlewareNotUsed`, so Django drops it from the stack): a deployed environment never runs it whatever the flags say and pays nothing. Enabling it on live traffic is DEV-1740's follow-up, not this PR.
- Strict in test settings: a mismatch fails the test that triggered it unless listed in `OPENAPI_KNOWN_MISMATCHES` — 85 `(error_code, route, method)` entries, each one a real schema bug to fix and delete.
- Entries are keyed by endpoint, not by test, so no test file is touched: the 1189 per-test entries the suite produces are only 85 distinct bugs, and endpoint keys survive test renames. Trade-offs: a new test hitting an already-listed endpoint passes silently, and the key ignores the status code, so a listed `response-validation` silences drift on every status that route returns. Adding the status to the key needs a full-suite regeneration — follow-up.
- Request-side errors are reported only when the API accepted the request. A test that posts a bad payload on purpose to check the 400 is the contract being honored, not a mismatch; without this, every such test on main turns this PR red (DEV-2294's `email-confirmations` did). Some of the 19 `request-payload-validation` entries in the list probably exist only for that reason and can go at the next regeneration.
- Regenerating the list is append-only: new entries land in a marked block, existing entries and comments stay untouched, and running it twice is a no-op. The middleware's own tests never feed the CSV and patch the list empty, so they do not depend on what is whitelisted.
- Also fixed in the middleware: two crashes in non-strict mode (`NameError` → 500 instead of a log line); route lookups ignore regex anchors so `re_path()` → `path()` refactors don't silently invalidate entries; the body is only read for JSON payloads (touching `request.body` on a multipart upload buffers it in memory and trips `DATA_UPLOAD_MAX_MEMORY_SIZE`, which Django skips for streamed uploads); `$ref`s resolve through `referencing` instead of the deprecated `jsonschema.RefResolver`; a broken `$ref` or schema fragment now propagates instead of being reported as an endpoint bug.
- Also hardened `kobo/apps/audit_log/middleware.py`: it crashed with an `AttributeError` on any response produced before URL resolution, where `request.resolver_match` is `None`. CORS preflights are the concrete case (`CorsMiddleware` answers them from `process_request`), so on main every cross-origin preflight is a 500. Guarded, with a regression test.
- The `schema_v2.*` diff is only the pagination `example` URL (`/accounts/` → `/api-endpoint/`, from `kpi/paginators.py`); nothing structural changed.
- Not covered, for a follow-up ticket: request headers, query-parameter types, undocumented status codes, DataDog.
- If we want zero impact on other PRs at first, this can land with `OPENAPI_VALIDATION_STRICT = False` and be flipped on in a one-line follow-up.

### 👀 Preview steps

1. ℹ️ backend only, no UI change — run these in a shell inside the kpi container
2. run `pytest -q kobo/apps/accounts/tests/test_current_user.py`
3. 🟢 [on PR] passes — every API test now also validates its requests and responses against the OpenAPI schema
4. remove the `('response-validation', 'me/', 'GET')` entry from `kobo/apps/openapi_validator/constants.py` and re-run
5. 🟢 `test_social_accounts` now fails with `AssertionError: OpenAPI validation error for /me/ [GET]: Response validation failed - None is not of type 'string'` — a real, currently undocumented mismatch; restore the file afterwards
6. 🔴 [on main] neither exists: the schema is never checked and drift goes unnoticed


kobo/apps/openapi_validator/README.md
```diff
@@ -0,0 +1,149 @@
...
+```python
+OPENAPI_VALIDATION = True
+OPENAPI_VALIDATION_STRICT = False
+OPENAPI_VALIDATION_BUILD_WHITELIST_LOG = True
+```
+
+Delete any stale `scripts/openapi_errors.csv`, then run the full test suite:
+
+```bash
+pytest -q -n auto --dist=loadfile
```
noliveleger, line 140: you need to rerun `pytest --lf` right after. Most of the time I get false position with `-n auto` because of concurrent HTTP requests.

kobo/apps/openapi_validator/middleware.py
```diff
@@ -0,0 +1,425 @@
...
+                    f'OpenAPI validation error for {request.path} '
+                    f'[{request.method}]: Missing required request body'
+                )
+                self._handle_validation_error(
+                    request, error_message, 'missing-required-payload'
+                )
+                return None
+
+            if hasattr(request, 'body') and request.body:
+                if 'json' in content_type.lower():
```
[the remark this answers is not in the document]
platreth: Declined on purpose and documented under "Not validated" in the app README, along with the other deliberate gaps (undocumented operations/status codes, headers, query-param types).

platreth:

Akuukis (commented):
I see it's ready for review now and may be merged anytime, please remember to remove the `POC - ` prefix

rgraber (changes requested):
Can we update the PR so that if we're not using the `testing` or `testing_no_stripe` environments the middleware turns itself off? Just to be super safe.

kobo/apps/openapi_validator/scripts/generate_constants.py
```diff
@@ -0,0 +1,113 @@
...
+        f"    ('{error_code}', '{route}', '{method}'),"
+        for error_code, route, method in sorted(triples)
+    )
+    lines.append('})')
+
+    with open(py_path, 'w', encoding='utf-8') as f:
+        f.write(format_source('\n'.join(lines) + '\n'))
+
+
+def run(csv_path: str, out_path: str, resolve: bool = True) -> None:
```
rgraber: nit: calling this method `run` makes it hard to search for. Maybe `regenerate_expected_failures`?

kobo/apps/openapi_validator/middleware.py
```diff
@@ -0,0 +1,439 @@
...
+
+    def __init__(self, get_response=None):
+        super().__init__(get_response)
+        # Test settings only. Validating live traffic would cost a schema
+        # validation per request and turn any bug in here into a 500, for no
+        # benefit the test suite does not already provide. Enabling it on a
+        # deployed environment is DEV-1740's follow-up, not this PoC.
+        if not (settings.TESTING and settings.OPENAPI_VALIDATION):
+            # Remove the middleware from the chain entirely when disabled
+            raise MiddlewareNotUsed
```
noliveleger: Maybe we could pass an explicit error to `Middleware NotUser` 

Note to myself, Django won't raise a 500 with this error, from Django source: 

```
mw_instance = middleware(adapted_handler)
except MiddlewareNotUsed as exc:
    if settings.DEBUG:
        logger.debug("MiddlewareNotUsed: %r", middleware_path)
    continue
```

kobo/apps/openapi_validator/middleware.py
```diff
@@ -0,0 +1,439 @@
...
+        self.schema = self._load_schema()
+        if not self.schema:
+            raise MiddlewareNotUsed
```
noliveleger: Let's pass a string to make a difference the one above.

noliveleger:
@platreth after regenerating the whitelist, it adds 

+        ('invalid-json-payload', 'api/v2/assets/', 'POST'),
+        ('invalid-json-response', 'api/v2/assets/', 'GET'),
+        ('missing-required-payload', 'api/v2/asset_subscriptions/', 'POST'),

Moreover, your comments are lost.  

This makes 4 tests to crash because we rely on the same endpoint that have just been whitelisted.

FAILED kobo/apps/openapi_validator/tests/test_middleware.py::OpenAPIValidationMiddlewareTestCase::test_empty_response_body_fails_in_strict_mode - AssertionError: AssertionError not raised
FAILED kobo/apps/openapi_validator/tests/test_middleware.py::OpenAPIValidationMiddlewareTestCase::test_malformed_json_response_body_fails_in_strict_mode - AssertionError: AssertionError not raised
FAILED kobo/apps/openapi_validator/tests/test_middleware.py::OpenAPIValidationMiddlewareTestCase::test_missing_required_request_body_fails_in_strict_mode - AssertionError: AssertionError not raised
FAILED kobo/apps/openapi_validator/tests/test_middleware.py::OpenAPIValidationMiddlewareTestCase::test_undecodable_response_body_fails_in_strict_mode - AssertionError: AssertionError not raised


It's going to be a chicken-and-egg problem I think, but we cannot rely on those endpoints to be always wrong.
Or maybe in `read_triples` , ignore lines which come from `kobo/apps/openapi_validator/tests/` ?

kobo/apps/openapi_validator/middleware.py
```diff
@@ -0,0 +1,488 @@
...
+        # if the API accepted the request anyway. A 4xx means it was rejected
+        # as the schema says it should be
+        pending = getattr(request, '_openapi_pending_error', None)
+        if pending and response.status_code < 400:
+            self._report_validation_error(request, *pending)
+
+        # Only JSON responses can be checked against a schema; file downloads,
+        # XML and HTML are served by the same endpoints and are out of scope
+        content_type = response.get('Content-Type', '').lower()
+        if 'json' not in content_type:
```
rgraber, line 194: Curiosity: Why not `content_type == 'application/json'`?

kobo/apps/openapi_validator/middleware.py
```diff
@@ -0,0 +1,488 @@
...
+
+        # Resolve $ref references
+        if '$ref' in schema:
+            if ref_schema := self._resolve_schema_ref(schema['$ref']):
+                return ref_schema
+            raise OpenAPIComponentRefNotFoundError
+
+        return schema
+
+    def _get_response_schema(
```
rgraber, line 304: nit: a fair bit of duplicated code here

rgraber (approved):
One question and one nit, but nothing blocking. LGTM

kobo/apps/openapi_validator/middleware.py
platreth: Because it's not always application/json: responses are application/scim+json (kobo_scim/renderers.py)

PR #7317, opened 2026-07-27
fix(admin): allow creating ExtraUserDetail without a manual uid DEV-960

rajpatel24:
### 📣 Summary
 Superusers can now add a user's extra details from the Django admin without having to invent an ID by hand.

### 📖 Description
Adding a new "extra user detail" record from the Django admin was effectively impossible. The form demanded a `uid` that the admin was supposed to generate automatically, and it rejected the empty `{}` value for the `data` field even though that is the field's own default. Both fields now behave as expected: the `uid` is generated on save and is no longer editable, and `{}` is accepted for `data`. 

### 👀 Preview steps

1. ℹ️  log in as a superuser and go to `/admin/hub/extrauserdetail/add/`
2. 🔴 [on main] notice the required `Uid` field, and that submitting with `{}` in `Data` fails with "This field is required"
3. 🟢 [on PR]  notice `Uid` is no longer an editable field on the add form
6. pick a user, leave `Data` as `{}`, save 
7. 🟢 notice the record saves successfully and has an auto-generated `uid` starting with `u`


noliveleger (approved):
LGTM

PR #7320, opened 2026-07-27
feat(bulkProcessing): consistent audio duration in UI DEV-2494

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Audio durations in Project → Data → Table now match the durations used by bulk transcription and translation.

### 💭 Notes

The duration shown next to each audio file in Table View came from the browser, while bulk processing gets its durations from the backend. They disagreed. E.g. 34 seconds in Table View vs 32 seconds in the bulk transcription modal.

Table View now asks the server too. The value shown is the one your processing minutes are actually counted against.

- The backend durations are only fetched for users with ASR/MT enabled and only for visible columns. I did it this way to avoid getting backend audio durations for users who will never see the difference (never will see the bulk processing modal).
- Moved `useCalculateAudioDuration.*` out of `jsapp/js/hooks/` into `components/submissions/`. Felt like better place.
- Tests: 6 on the hook and 6 on the provider.

### 👀 Preview steps

1. ℹ️ Have an account with ASR/MT enabled and a project with an audio question
2. Submit 4 recordings with fractional lengths (8.7s, 8.4s, and so on)
3. Go to Project → Data → Table, note the duration on each audio cell
4. Select those submissions, open the bulk transcription modal, note the total
5. 🔴 [on main] the per-cell durations add up to more than the modal's total
6. 🟢 [on PR] the per-cell durations add up to exactly the modal's total
7. Hide the audio column in table settings, reload, 🟢 no `attachments/audio-duration/` request in the network tab
8. 🟢 On an account with ASR/MT disabled, durations still render from the browser and no request fires


duvld (approved):
LGTM, preview steps work, love the comments

PR #7323, opened 2026-07-27
chore(dependencies): upgrade immutable package DEV-2504

jamesrkiger:
### 💭 Notes
Upgrades the 'immutable' package to remove security warnings from outdated version.

### 👀 Preview steps
This package was only being used on the Kobo Matrix component, which means that is all we need to test. However, the testing will need to be rather thorough because we had to make changes to rather basic state management functionality. You will want to create, preview, and edit a form with a kobo matrix on both main and this branch to ensure there are no changes in behavior.


magicznyleszek (approved):
LGTM. I don't see any differences in how Matrix works here vs main

PR #7325, opened 2026-07-27
feat(bulkProcessing): disable bulk option if no data DEV-2496 DEV-2499

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

The `Transcribe selected audio files` and `Translate selected transcriptions` options in the data table column menu are now greyed out when none of the selected submissions has anything to work with.

### 💭 Notes

Changes here:

- `bulkProcessingUtils.ts`
  - new `hasTranscribableAudio()` / `hasAnyTranscribableAudio()` — a non-deleted attachment for the given question xpath
  - new `hasTranslatableTranscript()` / `hasAnyTranslatableTranscript()` — a transcript with an actual `value`.
- `alerts/alertEvaluators.ts`
  - `evaluateNoSource()` now calls those shared helpers instead of doing its own inline checks.
- `tableColumnSortDropdown.tsx`
  - new `isBulkTranscriptionDisabled` and `isBulkTranslationDisabled` props, added into each item's existing `disabled` condition.
- `table.tsx`
  - extracted `getSelectedSubmissions()` (the column header needs the selected rows now, not just the modal opener) and passes both new props per column.
- `bulkProcessingUtils.tests.ts`
  - unit tests for all four helpers, including the deleted-attachment and pending-review cases.

### 👀 Preview steps

1. ℹ️ have an account and a project with an audio question, transcription/translation enabled in admin
2. add `?ff_bulkProcessingEnabled=true` to the URL to turn on the bulk processing feature flag
3. submit three responses: **A** and **B** with an audio recording, and **C** with the audio question skipped
4. go to Project → Data → Table
5. tick the checkbox for **C** only, then open the column menu (the caret in the header) of the audio question
6. 🔴 [on main] `Transcribe selected audio files` is clickable — click through the modal and get told there are no eligible submissions at the end
7. 🟢 [on PR] `Transcribe selected audio files` is greyed out
8. now tick **A** only and use `Transcribe selected audio files`, run the job, and wait for the transcript column to appear — leave the new transcript unapproved (it shows a `Review` button)
9. with **A** still the only selected row, open the column menu of the transcript column
10. 🔴 [on main] `Translate selected transcriptions` is clickable, and the modal only reveals at the end that nothing can be translated
11. 🟢 [on PR] `Translate selected transcriptions` is greyed out
12. tick **B** only (it has audio but no transcript at all) and open the transcript column menu
13. 🔴 [on main] `Translate selected transcriptions` is clickable
14. 🟢 [on PR] `Translate selected transcriptions` is greyed out
15. open **A**'s transcript and approve it, then go back to the table and select **A** again
16. 🟢 `Translate selected transcriptions` is now enabled and works as before — the option waits its turn instead of promising things it can't deliver

### DIFF
--- a/jsapp/js/components/submissions/tableColumnSortDropdown.tsx
+++ b/jsapp/js/components/submissions/tableColumnSortDropdown.tsx
@@ -30,6 +30,16 @@ interface TableColumnSortDropdownProps {
   onApproveSelectedSubmissions?: (fieldId: string) => void
   isBulkProcessingDisabled?: boolean
   /**
+   * `true` when none of the selected submissions has an audio file in this
+   * column, so there is nothing to transcribe.
+   */
+  isBulkTranscriptionDisabled?: boolean
+  /**
+   * `true` when none of the selected submissions has an approved transcript in
+   * this column, so there is nothing to translate.
+   */
+  isBulkTranslationDisabled?: boolean
+  /**
    * To be put inside trigger, before the predefined content. Please note that
    * the trigger as a whole is clickable, so this additional content would need
    * stopPropagation to be clickable.
@@ -153,7 +163,7 @@ export default function TableColumnSortDropdown(props: TableColumnSortDropdownPr
               {canTranscribeSelectedAudioFiles && (
                 <Menu.Item
                   className='sort-dropdown-menu-button'
-                  disabled={props.isBulkProcessingDisabled}
+                  disabled={props.isBulkProcessingDisabled || props.isBulkTranscriptionDisabled}
                   onClick={transcribeSelectedAudioFiles}
                   leftSection={<Icon name='qt-audio' size='inherit' />}
                 >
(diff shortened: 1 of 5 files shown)


magicznyleszek:
@greptile-apps review

jsapp/js/components/submissions/tableColumnSortDropdown.tsx
```diff
@@ -29,6 +29,16 @@ interface TableColumnSortDropdownProps {
   onTranslateSelectedTranscriptions?: (fieldId: string) => void
   onApproveSelectedSubmissions?: (fieldId: string) => void
   isBulkProcessingDisabled?: boolean
+  /**
+   * `true` when none of the selected submissions has an audio file in this
```
jamesrkiger: Minor quibble: This component doesn't get to choose its props, so the comments here seem a little misleading: the props are just true whenever the parents passes in true values. Alternatives: rewrite to make the comments imperative (i.e. Should be true), or drop comment and rename to something like  transcribableAudioAvailable (which will be self-documenting when it gets passed to 'disabled').

jamesrkiger (approved):
Left a minor quibble about some comments, but other than that LGTM and works well.

PR #7329, opened 2026-07-27
feat(organizations): enable sorting for organizations members API DEV-2454

Guitlle:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Adds an ordering parameter to the organization members API

### 📖 Description
The combined list of members and invitees is sorted in `OrganizationMemberViewSet.get_queryset`. The new parameter is also documented and the schema docs were updated. Since the get_queryset method converts the data to lists, the sorting is not done with the ORM but with a call to the sorted funcion. 

### 👀 Preview steps
1. ℹ️ have an account, an organization with multiple users. 
2. With the organization uid make API requests to test sorting and SSO status:

```
GET /api/v2/organizations/{uid}/members/
GET /api/v2/organizations/{uid}/members/?ordering=user__username
GET /api/v2/organizations/{uid}/members/?ordering=status
GET /api/v2/organizations/{uid}/members/?ordering=-role
GET /api/v2/organizations/{uid}/members/?ordering=-user__has_sso_enabled
```

3. Verify the sorting works as expected with your list of users


kobo/apps/organizations/views.py
platreth: The break in paginate_queryset assumes members always sort before invites. That held when (model_type, ordering_date) was the only sort. It doesn't now, so the loop stops at the first invite and every member after it comes back with invite: null. 



Can you add a test on ?ordering=-status asserting invite is still populated? Nothing catches this at the moment.



```

                for obj in page:
    if obj.model_type == '0_organization_user':
        members_user_ids.append(obj.user_id)
```

kobo/apps/organizations/views.py
```diff
@@ -526,29 +565,46 @@ def get_queryset(self):
...
                 model_type=Value('0_organization_user', output_field=CharField()),
+                username_sort=Lower(F('user__username')),
+                status_sort=Value('active'),
+                date_joined_sort=F('created'),
+                role_sort=Lower(role_annotation),
+                sso_sort=Case(
+                    When(Exists(sso_subquery), then=Value(1)),
+                    default=Value(0),
+                    output_field=IntegerField(),
+                ),
```
platreth: `sso_sort` and `has_sso_enabled` are the same Exists twice, so every row runs two identical subqueries. `role_sort` has the same problem against role (the Case already returns lowercase), and `date_joined_sort` is a copy of `ordering_date`. Can the map point at the existing annotations instead?

kobo/apps/organizations/views.py
```diff
@@ -526,29 +565,46 @@ def get_queryset(self):
...
-                    output_field=CharField(),
-                ),
+                role=role_annotation,
                 has_mfa_enabled=Exists(mfa_subquery),
+                has_sso_enabled=Exists(sso_subquery),
                 invite=Value(None, output_field=CharField()),
                 ordering_date=F('created'),
                 model_type=Value('0_organization_user', output_field=CharField()),
+                username_sort=Lower(F('user__username')),
+                status_sort=Value('active'),
```
platreth, line 598: status only has two values, so ties fall back to user id order, which is worse than the current default. Would attrgetter(sort_attr, 'model_type', 'ordering_date') work here?

platreth (changes requested):

kobo/apps/organizations/views.py
Guitlle: thank you, great catch

platreth (approved):

PR #7331, opened 2026-07-28
feat(projectViews): show organization IDs in the admin organization selector DEV-2500

platreth:
### 📣 Summary

Organizations are now shown with their ID in the Django admin, so organizations that share the same name can be told apart.

### 📖 Description

Many organizations share a name while being completely separate teams. Until now the Organizations selector of a Project View listed only names, so there was no way to know which entry was the intended one. Organizations are now displayed as `Name (org_id)` everywhere the Django admin shows one: in the Organizations suggestions, in the entries already selected, and in the other organization pickers.

### 💭 Notes

- `Organization.__str__` now returns `f'{self.name} ({self.id})'`.
- Doing it on the model covers both label paths of the admin autocomplete in one place: `AutocompleteJsonView.serialize_result()` for the AJAX suggestions, and `ModelChoiceField.label_from_instance()` for the values already selected. Fixing it at the widget level meant overriding both, in four different admins, and keeping them in sync.
- Side effect: the admin audit log embeds `str(obj)` in `object_repr`, so its messages now read `updated organization 'Foo (org_x)' (pk: org_x)`. Two assertions were updated for it.
- No leak outside the admin: the organization invite emails pass `organization.name`, and the organization user import/export binds `ForeignKeyWidget(Organization, field='name')`.

### 👀 Preview steps

1. ℹ️ be a superuser, and have two organizations sharing the exact same name
2. go to `/admin/project_views/projectview/add/`
3. type that shared name into the **Organizations** field
4. 🔴 [on main] both suggestions read just the name — no way to tell which is which
5. 🟢 [on PR] each suggestion reads `Name (org_…)`
6. pick one, fill in the rest of the form and save, then reopen the project view
7. 🟢 the selected organization still reads `Name (org_…)`

### DIFF
--- a/kobo/apps/audit_log/tests/test_signals.py
+++ b/kobo/apps/audit_log/tests/test_signals.py
@@ -354,7 +354,8 @@ class TestAdminAuditLogIntegration(BaseTestCase):
         self.assertEqual(latest_log.user, self.admin)
         self.assertEqual(latest_log.model_name, 'organization')
         self.assertIn(
-            "adminuser updated organization 'Updated Test Org Name' (pk: org999)",
+            'adminuser updated organization '
+            "'Updated Test Org Name (org999)' (pk: org999)",
             latest_log.metadata['message']
         )
         self.assertIn(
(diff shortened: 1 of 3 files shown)


rajpatel24 (approved):
LGTM

PR #7332, opened 2026-07-28
fix(organizations): return displayable error details for invites and members DEV-1218

platreth:
### 📣 Summary

Error messages shown when managing team or organization members now explain what actually went wrong, instead of a generic "there was an error" message.

### 📖 Description

Some screens in the **Members** and **Usage** sections replaced the server's explanation with their own wording — changing the role on an invitation that had already been accepted just said *"There was an error updating this invitation."* They now show the specific reason. When the server can't be reached at all, a single plain message is shown instead of a technical status code.

### 💭 Notes

- Removes the nine `// TODO: update message in backend (DEV-1218)` overrides; errors now flow through `onErrorDefaultHandler`, with one generic fallback there for when nothing displayable exists (empty 500, dropped connection).
- Root cause was shape, not wording: DRF keys `validate_<field>()` errors under the field name (`{"status": [...]}`), which `getApiErrorMessage()` can't read. `InvalidMembershipRequest` (400 `APIException`) isn't re-keyed, so messages arrive as a plain `detail` string.
- ⚠️ **API contract:** invite/member `400` bodies go from field-keyed to `{"detail": "…"}`. Status codes and the OpenAPI schema are unchanged (verified zero diff).
- Drive-by fixes: `validate_role` double-nested into `{"role": {"role": [...]}}`; `OrgInviteModal`'s `miscError` branch was unreachable and could render a stale error after a failure→success; `accessLogsSection` set `statusText` but `handleApiFail()` reads `responseText`. Resend rate-limit message now reports minutes (was "wait for 847 seconds"); `Retry-After` stays in seconds.
- Two strings had grammar fixed since they become user-visible.
- Every message constant in `organizations/constants.py` is now wrapped with `gettext_lazy` at its definition instead of `t(CONSTANT)` at the call site — `makemessages` only extracts literal arguments, so the old form left all 14 out of the catalogs (English in all 51 locales) despite already reaching the UI. Verified all 14 now extract.
- Tests pin both halves: they fail if the raises revert to `ValidationError`, or the frontend fallback reverts to `String(error)`.

### 👀 Preview steps

1. ℹ️ have an account owning a multi-member organization, with one **pending** invitation
2. from `/api/v2/docs/`, `PATCH /api/v2/organizations/<org_id>/invites/<guid>/` with `{"status": "expired"}`
3. 🔴 [on main] body is `{"status": ["`expired` is reserved and cannot be set"]}` — field-keyed, so the UI can't display it
4. 🟢 [on PR] body is `{"detail": "`expired` is reserved and cannot be set"}`
5. open **Account → Members** in two tabs; in tab A accept the invitation as the invitee
6. in tab B, change that same invitation's role with the dropdown
7. 🔴 [on main] toast reads "There was an error updating this invitation."
8. 🟢 [on PR] toast reads "Role cannot be changed after acceptance"


rajpatel24 (approved):
@platreth Backend changes look good to me. For the frontend changes, I think we should get a review from someone on the frontend team

jamesrkiger (approved):
LGTM

PR #7334, opened 2026-07-28
fix(messages): global variable replacement DEV-2510

jamesrkiger:
### 📣 Summary
Fixes a bug that resulted in only the first instance of a template variable (i.e. ##username##) being replaced in in-app messages.

### 👀 Preview steps
1. Make an in-app message in the django admin like with the following snippet and message:
```
username 1 ##username##
username 2 ##username##
```
2. 🔴 [on target branch] notice that only first instance of username variable is replaced
3. 🟢 [on PR] notice that both instances are replaced

### DIFF
--- a/jsapp/js/components/support/helpBubble.tsx
+++ b/jsapp/js/components/support/helpBubble.tsx
@@ -116,9 +116,9 @@ class HelpBubble extends React.Component<HelpBubbleProps, HelpBubbleState> {
 
   interpolateMessageUserInfo(htmlString: string) {
     return htmlString
-      .replace('##username##', this.props.username)
-      .replace('##user_uid##', this.props.userUid)
-      .replace('##user_full_name##', this.props.userFullName || t('KoboToolbox user'))
+      .replace(/##username##/g, this.props.username)
+      .replace(/##user_uid##/g, this.props.userUid)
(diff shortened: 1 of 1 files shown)


jsapp/js/components/support/helpBubble.tsx
```diff
@@ -116,9 +116,9 @@ class HelpBubble extends React.Component<HelpBubbleProps, HelpBubbleState> {
 
   interpolateMessageUserInfo(htmlString: string) {
     return htmlString
-      .replace('##username##', this.props.username)
-      .replace('##user_uid##', this.props.userUid)
-      .replace('##user_full_name##', this.props.userFullName || t('KoboToolbox user'))
+      .replace(/##username##/g, () => this.props.username)
+      .replace(/##user_uid##/g, () => this.props.userUid)
+      .replace(/##user_full_name##/g, () => this.props.userFullName || t('KoboToolbox user'))
```
magicznyleszek: You could also use [`replaceAll`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll), which makes it a tiny bit clearer?

magicznyleszek (approved):
LGTM, one small comment

PR #7337, opened 2026-07-29
fix(bulkProcessing): keep activity log progress in sync with bulk items DEV-2498

rajpatel24:
### 📣 Summary
This PR fixes bulk transcription and translation jobs reporting stale progress on the project activity page.

#### Problem
The data table and the project activity page report bulk processing progress from two different sources. The data table reads the bulk action item rows directly, so it is always current. The activity page reads a snapshot stored on the project history log, which is written when the job starts and refreshed only as a side effect of a periodic Celery task. That made the activity page only as fresh as the periodic task.

#### Solution 
The history log is now refreshed as each submission reaches a terminal state, rather than waiting for the next poll, so the activity page and the data table move together. Because several submissions can finish at once and refreshing the log is a read-modify-write of a single JSON field, the refresh takes a row lock so concurrent updates cannot be lost.

### 👀 Preview steps

1. ℹ️  Have an account and a project with at least 10 audio submissions
2. Enable bulk processing via the feature flag (`?ff_bulkProcessingEnabled=true`)
3. Bulk select the submissions and, from the data table's submission question column, hit **Transcribe selected audio files**
4. Open **Project Settings → Activity** in a second tab and watch the counter (e.g. `0/10`) under *Recent project activity* while keeping an eye on the data table
5. 🔴 [on main] notice that the data table shows the **Review** button for several processed submissions while the activity counter lags well behind (e.g. `1/10` when the data table already shows 5 **Review** buttons)
6. 🟢 [on PR] notice that the two stay in sync - the activity counter advances as each submission finishes
7. Repeat a few times to confirm it holds

### DIFF
--- a/kobo/apps/subsequences/audit.py
+++ b/kobo/apps/subsequences/audit.py
@@ -44,7 +44,17 @@ def sync_bulk_action_history_log(bulk_action) -> None:
     """
     Refresh the bulk-specific metadata on the existing history log row
     """
-    transaction.on_commit(lambda: _sync_bulk_action_history_log(bulk_action.pk))
+    sync_bulk_action_history_log_by_uid(bulk_action.pk)
+
+
+def sync_bulk_action_history_log_by_uid(bulk_action_uid: str) -> None:
+    """
+    Refresh the history log for a bulk action known only by its uid
+
+    Callers that update a child item with `queryset.update()` never load the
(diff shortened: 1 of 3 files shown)


rajpatel24:
@greptileai re-review

RuthTurk (approved):
LGTM

PR #7339, opened 2026-07-29
feat(projectHistoryLogs): log when users view data DEV-28

rgraber:
### 🗒️ Checklist

1. [x] run linter locally
2. [x] update developer docs (API, README, inline, etc.), if any
3. [x] for user-facing doc changes create a Zulip thread at `#Support Docs Updates`, if any
4. [x] draft PR with a title `<type>(<scope>)<!>: <title> DEV-1234`
5. [x] assign yourself, tag PR: at least `Front end` and/or `Back end` or `workflow`
6. [x] fill in the template below and delete template comments
7. [x] review thyself: read the diff and repro the preview as written
8. [x] open PR & confirm that CI passes & request reviewers, if needed
9. [x] act on any greptile review below a 5/5 score or leave comment explaining why you won't
10. [ ] delete this checklist section from the final squash commit before merging

### 📣 Summary
Log when users view data for a project.

### 📖 Description
Create new project history logs whenever a user hits the `/api/v2/assets/{uid}/data/` endpoint.

### 👷 Description for instance maintainers
This is likely to add a ton more audit logs. It shouldn't be enough to affect performance but it's worth noting.

### 💭 Notes
As noted above, this will probably add a ton more audit logs. However, logging hits to the endpoint was decided with product to be the best way to log that 'X viewed Y data'. For now there's a manual exception in the middleware to log a GET request if it's to the data endpoint specifically. If eventually we end up logging more GET requests we may consider a refactor (maybe a boolean field on AuditLoggedViewSet) but for one endpoint it seems like overkill.

### 👀 Preview steps

1. ℹ️ have an account and a project that has been deployed
2. Go to the project data table
3. Go to Settings > Activity
4. 🟢 [on PR] There should be a new `{username} accessed survey data` log

PR #7299, opened 2026-07-22
feat(KMLExport): cleanup legacy KoboCAT KML logic DEV-1635

RuthTurk:
### 📣 Summary
Removed legacy KML export generation from KoBoCAT in favor of routing through our KPI Exports. 

### 💭 Notes

* Rather than attempting to maintain backward-compatibility for `kml_legacy`, all legacy KML code has been fully purged from the backend/OpenRosa layer. A database migration (`viewer.0007_purge_kml_exports.py`) removes any existing `export_type = 'kml'` records from the `export` table and removes `'kml'` from the model choices. This guarantees that `/api/v2/assets/<uid>/exports/` will never return records with `.data.type = 'kml_legacy'` (or `'kml'` from Kobocat), allowing the frontend to safely strip `kml_legacy` logic.

### 👀 Preview steps

1. ℹ️ Have an account, a project, and a deployed survey with at least one of each geo type question (geopoint, geotrace, and geoshape).
2. Submit several responses containing geo data.
3. 🟢 [on PR] test the kml export process and ensure that it works correctly, generates accurate kml, and nothing breaks

PR #7341, opened 2026-07-29
ci(darker): print the diff so formatting failures are self-explanatory DEV-1034

platreth:
### 💭 Notes

Follow-up to #7316. That PR correctly made the `darker` job trust darker's exit code, but the command uses `darker --check`, which signals formatting/isort findings through the **exit code only** and prints nothing. The result: a real formatting failure shows up as a blank red ❌ with no output (see the silent failures on #7299 and #7339).

This adds `--diff` so darker prints the exact reformatting/isort patch it wants. `--diff` does not change the exit code, so #7316's crash-vs-clean behavior is preserved — the job just now explains itself.

Verified locally with the exact CI command against a real failing range:
- findings → `exit 1` **and** the patch is printed to stdout
- clean → `exit 0`, no output
- YAML parses

Devs hitting a red darker can then see what to run (`format-python.sh origin/main`) instead of guessing.

CI/internal only, no behavior change.

### DIFF
--- a/.github/workflows/darker.yml
+++ b/.github/workflows/darker.yml
@@ -57,7 +57,13 @@ jobs:
           # revision, missing linter, …). Do not capture the output — darker
           # writes crash details to stderr, so capturing stdout and testing it
           # for emptiness makes every crash look like a pass.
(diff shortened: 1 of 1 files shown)


RuthTurk (approved):
LGTM

PR #7343, opened 2026-07-29
feat(dataTable): freeze and hide for all DEV-2035

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Users without editing permissions can now hide, show, and freeze columns in Project → Data → Table for their own session, without changing what anyone else sees.

### 💭 Notes

Previously the "Hide field" and "Freeze field" options, along with the whole "Hide fields" dropdown, were only shown to users who could edit the project. That's because picking one of them saves the choice into the project settings for everybody.

Now everyone gets these options. Users who can edit the project keep saving their choice for all project users, exactly as before. Everyone else gets the same options applied only to their own browser session — nothing is saved, and reloading the page brings the original layout back. The message at the top of the "Hide fields" dropdown now says which of the two is happening.

Changes here:

- **`tableStore.ts`** — `saveTableSettings()` already picked between saving to the asset and falling back to session-local `overrides`. But `getTableSettings()` and `setOverrides()` only knew four settings, and `selected-columns` / `frozen-column` weren't among them, so un-hiding the buttons alone would have made them no-op.
  - both methods now loop over one shared `OVERRIDABLE_SETTINGS` list
  - `applySetting()` keeps the `undefined` check, not a falsiness one — `null` means "show everything" for `selected-columns` and "nothing frozen" for `frozen-column`, so unfreezing needs it
  - it deep-clones, since `table.tsx` mutates these arrays in place
- **`tableColumnSortDropdown.tsx`** — dropped the `change_asset` wrapper. The `asset` prop existed only for that check, so it went too.
- **`table.tsx`**
  - `ColumnsHideDropdown` is no longer gated — it's the only way back once you've hidden a column, since the header you'd click to unhide it goes with it
  - the overrides check now compares the whole overrides object via `isEqual` instead of `JSON.stringify` on one key. A store trigger is the only signal for users who can't save: no request goes out, so `componentDidUpdate` never sees a fresh asset
- **`columnsHideForm.tsx`** — Apply and Reset showed a spinner until the settings request came back. With no request to wait for, that spinner would still be spinning today, so `applySelection()` closes the dropdown right away on the session-only path.
  - also: "These settings affects" → "affect", plus a session-scoped variant of that message

No tests in this PR — didn't find time.

### 👀 Preview steps

1. ℹ️ have two accounts, and a project with at least a handful of questions and a few submissions
2. as the owner, go to Project → Settings → Sharing and share the project with the second user, granting **View submissions** only (leave "Edit form" unchecked)
3. log in as the second user and open Project → Data → Table
4. 🔴 [on main] notice there's no "Hide fields" button above the table
5. 🟢 [on PR] notice the "Hide fields" button is there
6. click the caret on any column header to open its dropdown
7. 🔴 [on main] notice the dropdown has sorting options only — no "Hide field", no "Freeze field"
8. 🟢 [on PR] notice both "Hide field" and "Freeze field" are listed
9. 🟢 click "Hide field" and notice the column disappears
10. 🟢 click "Freeze field" on the first column, scroll the table sideways, and notice that column stays put
11. 🟢 open the "Hide fields" dropdown and notice the message says these settings only apply to your current session
12. 🟢 reload the page and notice your hidden and frozen columns are back to how the owner left them — nothing was saved
13. now log in as the owner and open Project → Data → Table
14. 🟢 notice the table is untouched by whatever the second user was doing
15. 🟢 hide a column, reload, and notice it stays hidden — the owner's choice still saves for everyone, as before
16. 🟢 back as the second user, reload and notice the owner's hidden column is hidden for you too


magicznyleszek:
@greptile-apps review

pauloamorimbr (approved):
LGTM!

PR #7344, opened 2026-07-29
fix(subsequences): keep value key when pendingReview DEV-2508

noliveleger:
### 📣 Summary

Exporting submission data (XLSX, CSV, etc.) could fail when a transcription or translation was still awaiting review.

### 💭 Notes

`transform_data_for_output` (`kobo/apps/subsequences/actions/mixins.py`) previously omitted the `value` key entirely from the `_supplementalDetails` payload whenever `pendingReview` was `True`, keeping only the `pendingReview` flag.

The `/api/v2/assets/{uid}/data` endpoint (its only intended consumer per DEV-2333 / #7182) only needs the `pendingReview` flag to render the frontend's "Review" button, so this was fine there. 
However `SubmissionSupplement.retrieve_data(for_output=True)` is also reused by the export pipeline (`ExportTask.get_export_object` → `get_submissions(for_output=True)`), and formpack's `QualTranscriptField.get_value_from_entry` reads `response['value']` unconditionally, with no `except KeyError` guard (unlike its translation counterpart, which happens to be guarded). A submission whose latest transcription for a given language was pending review would therefore raise an uncaught `KeyError` while building that export row.

Fix: both methods now always include `value` in the output entry. When `pending_review` is `True`, `value` is set to an empty string instead of being omitted, so the unreviewed content isn't leaked to the frontend, but formpack (and any other consumer) can read the key unconditionally without crashing.

### 👀 Preview steps

1. ℹ️ have an account and a project with an automatic transcription action enabled
2. submit a response and let the automatic transcription complete without accepting it (so it stays pending review)
3. export the submission data (XLSX or CSV)
4. 🔴 [on release/2.026.27] export fails with an uncaught `KeyError`
5. 🟢 [on this PR] export succeeds; the transcript column is blank for that row until the transcription is accepted

### DIFF
--- a/kobo/apps/subsequences/actions/mixins.py
+++ b/kobo/apps/subsequences/actions/mixins.py
@@ -151,7 +151,12 @@ class TranscriptionActionMixin:
         }
 
         if pending_review:
+            # `value` is intentionally blanked out here: the FE only needs
+            # `pendingReview` to render its "Review" button, but formpack
+            # exports read `value` unconditionally and must not crash or
+            # leak an unreviewed result
             entry['pendingReview'] = True
+            entry['value'] = ''
         else:
             entry['value'] = version_data.get('value')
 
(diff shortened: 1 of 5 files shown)


rgraber (approved):
LGTM

PR #7346, opened 2026-07-30
feat(bulkProcessing): improve UI blocking code based on bulk job status(es) DEV-2538

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Fixed transcripts and translations being locked for review/save while other rows from the same bulk job were still being processed.

### 💭 Notes

Root cause: we checked `submission_uuids` (every row in the job) against the *job's* status. A job stays `in_progress` until its slowest row finishes, so one straggler held the whole batch hostage. The per-row statuses in `submission_statuses` were the right source all along — the table view was already using them and got this right, which is why cells correctly flipped to "Review".

Changes here:

- **`jsapp/js/components/submissions/bulkProcessingUtils.ts`** - adding two new utils for per-row status, reusing code
- **`jsapp/js/components/processing/common/conflictingOngoingJob.ts`** - using the helper, fixing all possible consumers; dropped the job-level status filter that's now redundant
- **`jsapp/js/components/submissions/BulkProcessingModals/alerts/alertEvaluators.ts`** - `evaluateConflictingJob()` had the same bug and would ignore rows a running job had already finished
- **Tests** (`conflictingOngoingJob.tests.ts`, `alertEvaluators.tests.ts`) - coverage for the partially-finished job etc, no longer randomizng `submission_statuses`

### 👀 Preview steps

1. ℹ️ have a project with an audio question and at least 4-5 submissions — mix in one clearly longer recording so it finishes last
3. select all the rows
4. in the audio column header dropdown pick "Transcribe selected audio files" and confirm in the modal
5. notice "Processing" appears for every row in the new transcription column
6. wait until at least one row flips to a "Review" button while at least one other row still says "Processing" ⚠️ this is the window we care about
7. click "Review" on one of the finished rows
8. 🔴 [on main] notice the "This submission is already being processed by another job." warning and a disabled "Save" button
9. 🟢 [on PR] notice no warning, and "Save" works — edit the transcript, save it, and the change sticks
10. go back to Project → Data → Table and click "Open" on a row that still says "Processing"
11. follow through with creating a transcript
12. 🟢 notice the warning and disabled buttons are still there for that one

### DIFF
--- a/jsapp/js/components/processing/common/conflictingOngoingJob.tests.ts
+++ b/jsapp/js/components/processing/common/conflictingOngoingJob.tests.ts
@@ -2,6 +2,7 @@ import chai from 'chai'
 import { ActionIdEnum } from '#/api/models/actionIdEnum'
 import type { BulkActionResponse } from '#/api/models/bulkActionResponse'
 import { BulkActionResponseStatusEnum } from '#/api/models/bulkActionResponseStatusEnum'
+import { BulkActionSubmissionStatusResponseStatusEnum } from '#/api/models/bulkActionSubmissionStatusResponseStatusEnum'
 import { getApiV2AssetsAdvancedFeaturesBulkActionsRetrieveResponseMock } from '#/api/react-query/survey-data/msw'
 import { isConflictingOngoingJobForSubmission } from './conflictingOngoingJob'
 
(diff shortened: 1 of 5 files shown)


jamesrkiger (approved):
LGTM

PR #7350, opened 2026-07-31
fix(userReports): replace per-user OR submission query with a single bounded scan DEV-2567

platreth:
### 📣 Summary

User Reports stays reliable on very busy servers: the job that precomputes billing and usage figures no longer risks exhausting the database's memory during large bursts of incoming submissions.

### 📖 Description

The background job behind the admin User Reports screen counts, for every organization, the submissions made during its current billing period. On servers handling a large volume of incoming submissions, the way that calculation queried the database could make it allocate a very large amount of memory at once, which contributed to a recent outage. It now computes the exact same figures with a lighter, more predictable query.

### 💭 Notes

- Removes the per-execution memory blowup behind the DB OOM: the current-period query OR-ed one clause per org (up to `CHUNK_SIZE = 1000`), which PostgreSQL plans as a `BitmapOr` whose memory scales with batch size. Now a single bounded `user_id__in … AND date__range=[min, max]` scan over the union of current-period windows, bucketed per user in Python. Same totals (edges inclusive); no API/schema/serializer change.
- Complements #7348 (DEV-2566) — that PR fixes the task's cadence/lock reliability, this one the query memory. Same incident, no file overlap.
- Follow-up (separate ticket): `logger_dailyxformsubmissioncounter` has a `(date, user)` index — wrong order for per-user lookups. A `(user, date)` index is the durable win but a heavy gated migration, so not here.
- Verified: pytest on both axes (`testing` + `testing_no_stripe`) and `darker`/`flake8` on changed lines, green.

### 👀 Preview steps

1. ℹ️ have `DailyXFormSubmissionCounter` rows for several orgs that each have a current billing period
2. run `refresh_user_report_snapshots` (or call `BillingAndUsageCalculator().calculate_usage_batch(...)`)
3. 🔴 [on main] the current-period query is `WHERE (user_id=… AND date BETWEEN …) OR (…) OR …` — one branch per org; `EXPLAIN` shows a `BitmapOr` of many index scans
4. 🟢 [on PR] it's a single `WHERE user_id IN (…) AND date BETWEEN <min> AND <max>` scan, bucketed per user
5. 🟢 each org's `total_submission_count.current_period` is identical between main and PR

### DIFF
--- a/kobo/apps/user_reports/utils/billing_and_usage_calculator.py
+++ b/kobo/apps/user_reports/utils/billing_and_usage_calculator.py
@@ -1,3 +1,4 @@
+from collections import defaultdict
 from datetime import datetime
 
 from django.db.models import Sum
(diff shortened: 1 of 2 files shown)


kobo/apps/user_reports/utils/billing_and_usage_calculator.py
```diff
@@ -68,19 +76,37 @@ def _get_submission_usage_batch(self, user_ids, date_ranges_by_user):
...
+        if windows:
             rows = (
-                DailyXFormSubmissionCounter.objects.filter(combined_q)
-                .values('user_id')
+                DailyXFormSubmissionCounter.objects.filter(
+                    user_id__in=windows.keys(),
+                    date__range=[
+                        min(start for start, _ in windows.values()),
+                        max(end for _, end in windows.values()),
+                    ],
```
rajpatel24: This PR removes the OR expansion, which was the real mechanism behind the outage.

One thing I'd like your read on. The scan is bounded by `[min(all starts), max(all ends)]` across the entire 1000-org chunk. That stays narrow as long as every org is on a monthly window, but what happens if the interval is `'year'`. An annual plan spans a 365-day window.

So, if one org's annual period started ~10 months ago, it pulls `min(start)` back by 10 months. Likewise, if another org's period ends in 10 months, it pushes `max(end)` forward. A handful of annual customers can therefore widen the union window for the entire chunk by an order of magnitude.

Why I think it's worth catching this now: `DAILY_COUNTERS_MAX_DAYS` is 366, so the table never holds more than ~12 months of data. Once the window exceeds that, the date filter effectively stops excluding anything, and we're back to reading every row for those users. `GROUP BY user_id, date` then has to build far more groups, which could be enough to exceed `work_mem` and spill to disk.

This PR is still hugely better than what caused the outage. The part that nags me is that it's data-dependent: it'll look perfect in staging and on any chunk with no annual customers, then behave differently on the one chunk that happens to have one. That's the kind of thing that tends to resurface at 2 a.m.

One ORM-friendly approach would be to bucket users by their distinct `(start, end)` window and run one query per bucket:

```python
buckets = defaultdict(list)

for uid, (start, end) in windows.items():
    buckets[(start, end)].append(uid)

for (start, end), uids in buckets.items():
    rows = (
        DailyXFormSubmissionCounter.objects.filter(
            user_id__in=uids,
            date__range=[start, end],
        )
        .values("user_id")
        .annotate(total=Coalesce(Sum("counter"), 0))
    )

    for row in rows:
        current[row["user_id"]] = row["total"]
```

Since ~90% of orgs share the default first-of-the-month window, a chunk would only have a few dozen distinct windows. Each query gets a genuinely narrow date range and still aggregates back to `values("user_id")`, so only ~1,000 rows come back instead of one row per user per day. That also makes the annual-plan case no longer matter.

Of course, please correct me if I'm wrong about any of this :)

rajpatel24 (changes requested):

kobo/apps/user_reports/utils/billing_and_usage_calculator.py
platreth: You're right! I will change that, thanks for the feedback!!

rajpatel24 (approved):
LGTM

PR #7353, opened 2026-07-31
feat(frontend): improve Select and MultiSelect UX DEV-2569

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Make all Selects and MultiSelects searchable by default. Also add PageUp, PageDown, Home, and End keyboard navigation to them.

### 💭 Notes

We basically want our Selects to have UX at least as good as native select.

Changes here:
- Make all Selects searchable by default (in theme file)
- Make all Select instances with small amount of options non-searchable
- Common components Select and MultiSelect now have a custom onKeyDown caller that runs a function for handling advanced keyboard navigation
  - This is a case when `no-restricted-imports` config should become useful - when dev imports Select from `@mantine/core`, the advanced keyboard navigation will not work, but they should not import it from there, but rather from `js/components/common`
- Added a complex `js/components/common/selectKeyboardNavigation.ts` that makes the PageUp, PageDown, Home, and End functionality happen. I even added a storybook test to verify it works :+1:

### 👀 Preview steps

1. ℹ️ have an account
2. Go to Account Settings → General
3. Use the Country selector testing all the mentioned functionalities (keyboard keys on MacOS are fn+arrows)
4. 🍏 On PR it works!
5. 🟥 On main it fails


magicznyleszek:
@greptileai review

jamesrkiger (approved):
LGTM

PR #7355, opened 2026-07-31
fix(usage): batch usage queries by billing window DEV-2567

noliveleger:
### 📣 Summary
Fixes a background job that could overload the database and disrupt service when checking which accounts are approaching their usage limits.

### 💭 Notes
Same root cause and fix pattern as the earlier `BillingAndUsageCalculator` fix for #7350  (bitmap-or OOM), applied to the two remaining call sites that had the same shape: `get_submission_counts_in_date_range_by_user_id` and `get_nlp_usage_in_date_range_by_user_id` in `kpi/utils/usage_calculator.py`.

Both functions built one giant `Q()` OR-chain (`user_id=X AND date BETWEEN start AND end` per user) and ran it as a single query. These are reached via `get_submissions_for_current_billing_period_by_user_id` / `get_nlp_usage_for_current_billing_period_by_user_id`, called without an `orgs` filter from `kobo.apps.mass_emails.user_queries.get_users_within_range_of_usage_limit` (the periodic `generate_mass_email_user_lists` task) — so the OR-chain spanned every organization in the instance, not a bounded batch.

Both now group users by their distinct `(start, end)` billing window and issue one `user_id__in=...` query per window instead, matching the existing calculator fix.

Also fixed along the way: `get_submission_counts_in_date_range_by_user_id` was grouping by `.values('counter', 'user_id', 'date')` instead of `.values('user_id')`, so the `Sum('counter')` annotation wasn't actually aggregating per user — the returned dict kept only one arbitrary row per user rather than the true sum. Now grouped by `user_id` only, consistent with the NLP usage function and the calculator.

No behavior change for callers — same return shape (`dict[int, int]` / `dict[int, NLPUsage]`), just fixed aggregation and no more single-query OR-chain.

### 👀 Preview steps

Not user-facing; covered by existing unit tests in `kpi/tests/test_usage_calculator.py` and `kobo/apps/mass_emails/tests/test_usage_limit_user_queries.py`.

### DIFF
--- a/kpi/utils/usage_calculator.py
+++ b/kpi/utils/usage_calculator.py
@@ -1,3 +1,5 @@
+from collections import defaultdict
+from datetime import datetime
 from json import dumps, loads
 from math import inf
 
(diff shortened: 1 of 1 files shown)


rgraber (approved):
Fine for expediency. Eventually we should deduplicate all this.

noliveleger:
Closed in favor of #7358

PR #7356, opened 2026-07-31
feat(designSystem): convert single processing view tabs to mantine DEV-2272

duvld:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Part of series of changes to convert older components to new stylized components. This one affects the tabs in the single processing view

### 💭 Notes
Few notable changes:
- Had to export helper function from processing utils to get the proper route
- A bit hacky of a fix needed to make `Link`'s work with middle-clicking and to use the `safeExecute` protector function
- The theme is inverted of the default mantine tabs theme, but I asked the robot to inverse the theme so it works on the grey background. I've added it to the existing CSS with its own classnames

### 👀 Preview steps

1. ℹ️ have an account and a project with audio submissions
2. go to single processing view
5. 🟢 [on PR] notice that the tabs:
    - match the theme
    - function as before
    - can be middle-clicked into a new tab
    - unsaved work still gets notified to the user if trying to switch to a new tab
6. 🟢 Tab navigation works
    - Note mantine's default behaviour for in-conetxt tab navigation (i.e., navigating to other tabs in the "bunch"), is mean to be done via: Tab into the "bunch" (will land on Transcript) -> use arrow keys -> access the other tabs
8. 🟢 Opening tabs in a new browser tab works


tesster7:
@duvld thanks for creating this PR, let's chat about NLP updates to Mantine on Thursday before we merge the changes 🙌

duvld:
@jamesrkiger Here's the PR for moving the single processing view tabs to mantine and adding accessibility changes. Note the tab navigation has a quirk and I've added it to the preview steps

jsapp/js/components/processing/SingleProcessingContent/index.tsx
```diff
@@ -44,54 +45,61 @@ export default function SingleProcessingContent({
...
-          className={classNames({
-            [styles.tab]: true,
-            [styles.activeTab]: isProcessingRouteActive(PROCESSING_ROUTES.TRANSCRIPT),
-          })}
-          onClick={handleTranscriptClick}
-        >
-          {t('Transcript')}
-        </li>
+      <Tabs
+        mt='8px'
```
magicznyleszek: This is causing the difference between the left content and the right side settings:
Image

I think we should use some shared way so they don't divert again

jsapp/js/components/processing/SingleProcessingContent/index.tsx
```diff
@@ -44,54 +45,61 @@ export default function SingleProcessingContent({
...
+        size='lg'
+        value={activeTab}
+        classNames={{ root: styles.tabsRoot, list: styles.tabsList, tab: styles.tab }}
```
magicznyleszek: I wonder if we shouldn't make a special variant in the Tabs component and use it here… Given the Single Processing UI will probalby not be redesigned anytime soon, this is kinda existing valid variant, not a hack. WDYT?

magicznyleszek (changes requested):

jsapp/js/components/processing/SingleProcessingContent/index.tsx
duvld: I made the variant, I'm fine either way but this is a slight improvement in readability as all the relevant styles are in one place

jsapp/js/components/processing/SingleProcessingContent/index.tsx
duvld: The "Display settings" row is already hard coded to `h={48}`. I added this to the Tabs row to keep it the same. I then added 12px padding to the parent so they're pushed down the same. I don't know if this is the best way to do it though

magicznyleszek (approved):
LGTM

jamesrkiger (approved):
The tabs look much nicer than our old version, but they keep the current style so let's go ahead and merge this. Thanks for working on it.

PR #7358, opened 2026-08-02
fix(usage): batch usage queries by billing window DEV-2567

noliveleger:
### 📣 Summary

Fixes a background job that could overload the database and disrupt service when checking which accounts are approaching their usage limits.

### 💭 Notes

- Same root cause and fix pattern as the earlier `BillingAndUsageCalculator` fix in #7350 (bitmap-or OOM), applied to the two remaining call sites with the same shape: `get_submission_counts_in_date_range_by_user_id` and `get_nlp_usage_in_date_range_by_user_id` in `kpi/utils/usage_calculator.py`.
- Both functions built one giant `Q()` OR-chain (`user_id=X AND date BETWEEN start AND end` per user) and ran it as a single query. These are reached via `get_submissions_for_current_billing_period_by_user_id` / `get_nlp_usage_for_current_billing_period_by_user_id`, called without an `orgs` filter from `kobo.apps.mass_emails.user_queries.get_users_within_range_of_usage_limit` (the periodic `generate_mass_email_user_lists` task), so the OR-chain spanned every organization in the instance, not a bounded batch.
- Both now group users by their distinct `(start, end)` billing window and issue one `user_id__in=...` query per window instead, matching the existing calculator fix.
- Also fixed along the way: `get_submission_counts_in_date_range_by_user_id` was grouping by `.values('counter', 'user_id', 'date')` instead of `.values('user_id')`, so the `Sum('counter')` annotation wasn't actually aggregating per user, the returned dict kept only one arbitrary row per user rather than the true sum. Now grouped by `user_id` only, consistent with the NLP usage function and the calculator.
- No behavior change for callers, same return shape (`dict[int, int]` / `dict[int, NLPUsage]`), just fixed aggregation and no more single-query OR-chain.
- Follow-up cleanup bundled in this PR: the "group users by window" pattern above ended up duplicated three times (`BillingAndUsageCalculator._get_submission_usage_batch`, `get_submission_counts_in_date_range_by_user_id`, `get_nlp_usage_in_date_range_by_user_id`), each with its own copy of the same `_as_date` helper. Extracted the shared logic into `group_user_ids_by_date_range()` in `kpi/utils/usage_calculator.py`, which `billing_and_usage_calculator.py` already imports from. Pure refactor, no behavior change.

### 👀 Preview steps

Not user-facing; covered by existing unit tests in `kpi/tests/test_usage_calculator.py`, `kobo/apps/mass_emails/tests/test_usage_limit_user_queries.py`, and `kobo/apps/user_reports/tests/test_user_reports.py`

### DIFF
--- a/kobo/apps/user_reports/utils/billing_and_usage_calculator.py
+++ b/kobo/apps/user_reports/utils/billing_and_usage_calculator.py
@@ -1,12 +1,12 @@
-from collections import defaultdict
-from datetime import datetime
-
 from django.db.models import Sum
 from django.db.models.functions import Coalesce
 
 from kobo.apps.openrosa.apps.logger.models import DailyXFormSubmissionCounter
 from kobo.apps.organizations.models import Organization
-from kpi.utils.usage_calculator import get_storage_usage_by_user_id
+from kpi.utils.usage_calculator import (
+    get_storage_usage_by_user_id,
+    group_user_ids_by_date_range,
+)
 from ..typing_aliases import OrganizationIterator
 
 
@@ -55,12 +55,6 @@ class BillingAndUsageCalculator:
         except AttributeError:
             return None
 
-    @staticmethod
-    def _as_date(value):
-        # Billing bounds are tz-aware datetimes; `date` is a DateField, so
-        # normalize to a date for the Python-side window comparison below.
-        return value.date() if isinstance(value, datetime) else value
-
     def _get_submission_usage_batch(self, user_ids, date_ranges_by_user):
         if not user_ids:
             return {}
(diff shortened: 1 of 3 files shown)


platreth (approved):
LGTM! 

But for regression pruposes I would add a new test case with a user with two in-range dates and asserting the sum before merging.

Something like this: 

self.add_submissions(count=2, username='someuser', date_override=three_months_ago)
self.add_submissions(count=3, username='someuser', date_override=four_months_ago)
=> assert submissions_by_user[self.someuser.id] == 5

platreth (approved):

PR #7361, opened 2026-08-03
fix(submissions): reject trailing slash on OpenRosa endpoints with 404 DEV-1039

platreth:
### 📣 Summary

Calling an OpenRosa endpoint (form list or submission) with an accidental trailing slash now returns a clear "not found" message instead of a confusing security error.

### 📖 Description

The OpenRosa endpoints are meant to be called without a trailing slash. Adding one (e.g. `/submission/`) used to surface a misleading CSRF error. They now respond with an explicit 404 that names the correct URL to retry. The official clients (Collect, Enketo) are unaffected — they already build the correct slash-less URLs. This only helps people writing their own integrations or curl commands.

New `OpenRosaTrailingSlashMiddleware`, ordered ahead of `CsrfViewMiddleware`, intercepts any request whose path ends in `/submission/` or `/formList/` — covering the authenticated, per-user, data-collector and asset-snapshot variants — and returns a 404 pointing at the slash-less URL. Honoring the slash was deliberately rejected: OpenRosa clients build slash-less URLs, and Django's `APPEND_SLASH` redirect would drop the POST body on submissions.

### 💭 Notes

- Replaces the earlier url-pattern approach after discussion — one middleware covers all eight endpoint variants via two path suffixes, rather than duplicating routes.
- No nginx or k8s change needed; the companion kobo-docker PR #384 is closed.
- Tests: `kobo/apps/openrosa/libs/utils/tests/test_middleware.py` (4 tests / 13 subtests, green on both settings axes). Existing `test_past_bugs.py` still passes, confirming slash-less routing is untouched.

### 👀 Preview steps

1. ℹ️ have an account and a deployed project
2. `curl -i -X POST https://kc.kobo.local/<username>/submission/` (note the trailing slash)
3. 🔴 [on main] a confusing `403` CSRF error comes back
4. 🟢 [on PR] a `404` comes back: `OpenRosa endpoints do not accept a trailing slash. Retry the request at /<username>/submission`
5. 🟢 the slash-less URL (`.../submission`) still reaches the submission view and behaves exactly as before

### DIFF
--- a/kobo/apps/openrosa/libs/utils/middleware.py
+++ b/kobo/apps/openrosa/libs/utils/middleware.py
@@ -8,6 +8,7 @@ from django.http import (
     HttpResponse,
     HttpResponseForbidden,
     HttpResponseNotAllowed,
+    HttpResponseNotFound,
 )
 from django.middleware.locale import LocaleMiddleware
 from django.template import loader
@@ -39,6 +40,34 @@ ALLOWED_VIEWS_WITH_WEAK_PASSWORD = {
         'GET': []
     },
 }
+
+
+class OpenRosaTrailingSlashMiddleware(MiddlewareMixin):
+    """
+    OpenRosa endpoints (submission, formList and their variants) are reached
+    WITHOUT a trailing slash: clients (Collect, Enketo) build slash-less URLs,
+    and Django's APPEND_SLASH redirect would drop the POST body on submissions.
+
+    Without this, a slashed request falls through to CsrfViewMiddleware and
+    surfaces a misleading CSRF error (DEV-1039). Intercept before CSRF runs and
+    return an explicit 404 pointing at the correct, slash-less URL.
+    """
+
+    # Slash-stripped path suffixes that identify an OpenRosa endpoint.
+    OPENROSA_SUFFIXES = ('/submission', '/formList')
+
+    def process_request(self, request):
+        path = request.path
+        if not path.endswith('/'):
+            return None
+
+        target = path.rstrip('/')
+        if target.endswith(self.OPENROSA_SUFFIXES):
+            return HttpResponseNotFound(
+                'OpenRosa endpoints do not accept a trailing slash. '
+                f'Retry the request at {target}'
+            )
+        return None
 
 
 class HTTPResponseNotAllowedMiddleware(MiddlewareMixin):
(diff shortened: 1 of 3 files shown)


platreth:
@greptileai

kobo/apps/openrosa/libs/utils/middleware.py
```diff
@@ -41,6 +42,34 @@
...
+    # Slash-stripped path suffixes that identify an OpenRosa endpoint.
+    OPENROSA_SUFFIXES = ('/submission', '/formList')
+
+    def process_request(self, request):
+        path = request.path
+        if not path.endswith('/'):
+            return None
+
+        target = path.rstrip('/')
+        if target.endswith(self.OPENROSA_SUFFIXES):
```
noliveleger: This seems a little bit fragile, don't you think? Imagine a new endpoint is created called `/api/v2/assets/{uid_asset}/submission/`. It would enter this condition and would return a 404. 

What about using a resolver as a safety guard?

```
      OPENROSA_URL_NAMES = frozenset(                                                                                        
          ['submissions', 'form-list', 'assetsnapshot-submission',                                                           
           'assetsnapshot-form-list']                                                                                        
      )                                                                                                                      
                                                                                                                             
      def process_request(self, request):                                                                                    
          path = request.path
          if not path.endswith('/'):                                                                                         
              return None                                                                                                    
                                                                                                                             
          target = path.rstrip('/') 
          # check pre-filter . no resolver involved                                                                                         
          if not target.endswith(self.OPENROSA_SUFFIXES):                                                                    
              return None                                                                                                    
                                                                                                                             
          # make sure we are check OpenRosa endpoints.
          try:                                                                                                               
              match = resolve(target)                                                                                        
          except Resolver404:                                                                                                
              return None                                                                                                    
                                                                                                                             
          if match.url_name not in self.OPENROSA_URL_NAMES:                                                                  
              return None                                                                                                    
                                                                                                                             
          return HttpResponseNotFound(...) 
```

noliveleger (changes requested):
LGTM

kobo/apps/openrosa/libs/utils/middleware.py
platreth: Agree!

noliveleger (approved):
LGTM

EDIT: Darker is failing

PR #7362, opened 2026-08-03
fix(dataTable): improve select_x data displaying DEV-2476

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Data Table now shows all selected options of a `select_many` question, falling back to the raw value for options that were renamed or removed in a later version of the form.

### 💭 Notes

When a form is edited, the `name` of a select option can change, while submissions collected earlier keep the old one. The DATA table had no label to show for such values, and for `select_many` questions it dropped them from the cell entirely — a response of three animals could show up as just one, with nothing hinting the other two were ever picked.

Those values now appear as-is, next to the labels of the options that still match. This is how `select_one` questions already behaved.

Same fix applies to the single submission modal, which listed unmatched values but in a way that broke the bullet list.

Changes here:
- `jsapp/js/components/submissions/tableUtils.ts`
  - adds `getSelectResponseLabel()` for resolving both `select_one` and `select_many` responses with a per-value fallback to the raw value
- `jsapp/js/components/submissions/DataTableCell/index.tsx`
  - both select types now go through the helper instead of two branches that had quietly drifted apart
  - label lookup is now gated on `translationIndex > -1` ("Display XML Values"). It used to be `=== 0` for `select_many`, so picking any language other than the first one showed raw values instead of labels. Free bug fix 🎁🐛🔨
- `jsapp/js/components/submissions/submissionDataTable.tsx`
  - unmatched values already fell back to the raw value here, but returned a bare string inside the `<ul>`, so it landed outside the `<li>` and without a key (Also 🎁🐛🔨)
- `jsapp/js/components/submissions/tableUtils.tests.ts` - 9 cases for the helper
- `jsapp/js/components/submissions/DataTableCell/DataTableCell.stories.tsx`
  - 2 stories for the renamed-options cell, with labels and with XML values

Not touched: the column filter dropdown still only offers options from the current form version, so these newly visible legacy values can't be filtered on - made an issue though: https://linear.app/kobotoolbox/issue/DEV-2576/data-table-column-filter-only-lists-option-values-from-the-latest-form.

### 👀 Preview steps

2. add a `select_many` question with three options, name them `a`, `b` and `c`, and label them `archaeopteryx`, `badger` and `crocodile`
3. deploy the form and submit one response with all three animals selected
4. edit the form and rename the option names `a` and `b` to `a1` and `b1` — leave the labels alone, and leave `c` as it is
5. redeploy the form
6. go to Project → Data → Table, open 'Table display options' and pick 'XML Values' — all three original values `a b c` show up, as they should
7. switch 'Table display options' to 'Labels'
8. 🔴 [on main] the cell shows only "crocodile" — the archaeopteryx and the badger swam off without a trace
9. 🟢 [on PR] the cell shows "a, b, crocodile"
10. 🟢 open the submission (click the row) and see the same three values listed there
11. if your form has a second language, switch 'Table display options' to 'Labels - <language>'
12. 🔴 [on main] the cell shows raw values `a b c`, no labels at all
13. 🟢 [on PR] the cell shows the labels of that language, with `a` and `b` still raw

### DIFF
--- a/jsapp/js/components/submissions/DataTableCell/DataTableCell.stories.tsx
+++ b/jsapp/js/components/submissions/DataTableCell/DataTableCell.stories.tsx
@@ -132,9 +132,6 @@ const videoSubmission = buildMediaSubmission(
   'attachment-video-1',
 )
 
-// Choices 'a' and 'b' were renamed to 'a1' and 'b1', so only
-// the new names exist in the choice list below while the submission still holds
-// the old ones.
 const renamedChoicesQuestion: SurveyRow = {
   type: QuestionTypeName.select_multiple,
   $kuid: 'renamedChoicesQuestion',
@@ -144,6 +141,9 @@ const renamedChoicesQuestion: SurveyRow = {
   select_from_list_name: 'animals_list',
 }
 
+// Choices 'a' and 'b' were renamed to 'a1' and 'b1', so only
+// the new names exist in the choice list below while the submission still holds
+// the old ones.
 const renamedChoices: SurveyChoice[] = [
   { name: 'a1', label: ['Whale'], list_name: 'animals_list', $autovalue: 'a1', $kuid: 'whaleChoice' },
   { name: 'b1', label: ['Frog'], list_name: 'animals_list', $autovalue: 'b1', $kuid: 'frogChoice' },
--- a/jsapp/js/components/submissions/DataTableCell/DataTableCell.stories.tsx
+++ b/jsapp/js/components/submissions/DataTableCell/DataTableCell.stories.tsx
@@ -132,7 +132,7 @@ const videoSubmission = buildMediaSubmission(
   'attachment-video-1',
 )
 
-// Reproduces DEV-2476: choices 'a' and 'b' were renamed to 'A' and 'B', so only
+// Choices 'a' and 'b' were renamed to 'a1' and 'b1', so only
 // the new names exist in the choice list below while the submission still holds
 // the old ones.
 const renamedChoicesQuestion: SurveyRow = {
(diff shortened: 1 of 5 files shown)


jsapp/js/components/submissions/tableUtils.ts
```diff
@@ -569,6 +569,51 @@ export function buildFilterQuery(
...
+  questionType: AnyRowTypeName | undefined
+  /** The question's `select_from_list_name`. */
+  listName: string | undefined
+  /** All of the form's choices, not just the ones of this question's list. */
+  choices: SurveyChoice[]
+  /**
+   * Index into a choice `label` array. Only pass a non-negative one; negative
+   * means "display raw XML values", where there's nothing to resolve.
+   */
+  translationIndex: number
```
duvld, line 586: I'm a bit confused why -1 doesn't resolve? It seems like -1 should be the fallback option because every response should have an XML value no matter what

jsapp/js/components/submissions/DataTableCell/DataTableCell.stories.tsx
```diff
@@ -132,6 +132,26 @@ const videoSubmission = buildMediaSubmission(
   'attachment-video-1',
 )
 
+// Choices 'a' and 'b' were renamed to 'a1' and 'b1', so only
```
duvld: Nitpick, but shouldn't this comment be about the block below?

duvld (commented):
Tiny nitpick and question

jsapp/js/components/submissions/DataTableCell/DataTableCell.stories.tsx
magicznyleszek: Right!

jsapp/js/components/submissions/tableUtils.ts
magicznyleszek: Improved the comment

duvld (approved):
LGTM

PR #7364, opened 2026-08-03
ci(releases): improve checkout-with-github-app-token action DEV-2578

Akuukis:
### 💭 Notes

Standardize `app-id` and `client-id` usage to `client-id` and bump GHA actions.

Future-proof `checkout-with-github-app-token` action with 3 optional
parameters: ref, owner and repository.

### DIFF
--- a/.github/actions/checkout-with-github-app-token/action.yml
+++ b/.github/actions/checkout-with-github-app-token/action.yml
@@ -51,3 +51,4 @@ runs:
       token: ${{ steps.app-token.outputs.token }}
       ref: ${{ inputs.ref }}
       fetch-depth: "0"
+      fetch-tags: true
--- a/.github/actions/checkout-with-github-app-token/action.yml
+++ b/.github/actions/checkout-with-github-app-token/action.yml
@@ -29,7 +29,7 @@ runs:
     id: app-token
     uses: actions/create-github-app-token@v3
     with:
-      app-id: ${{ inputs.client-id }}
+      client-id: ${{ inputs.client-id }}
       private-key: ${{ inputs.private-key }}
       owner: ${{ inputs.owner }}
       repositories: ${{ inputs.repositories }}
--- a/.github/actions/checkout-with-github-app-token/action.yml
+++ b/.github/actions/checkout-with-github-app-token/action.yml
@@ -2,10 +2,24 @@ name: 'Checkout with Github App token'
 description: 'Checkout repo with a Github App token and set bot config, so that push to the repo would trigger workflows.'
 
 inputs:
-  app-id:
+  client-id:
     required: true
   private-key:
     required: true
+  ref:
+    description: 'Git ref to checkout (default: current ref)'
+    required: false
+  owner:
+    description: 'GitHub organization for cross-repo token scope (optional)'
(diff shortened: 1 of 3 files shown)


.github/actions/checkout-with-github-app-token/action.yml
```diff
@@ -2,21 +2,37 @@ name: 'Checkout with Github App token'
 description: 'Checkout repo with a Github App token and set bot config, so that push to the repo would trigger workflows.'
```
ScottMillard, line 2: non blocking : Might be good to update this comment since it doesn't use the app token any longer

ScottMillard (approved):

.github/workflows/release-1-branch.yml
```diff
@@ -48,10 +48,9 @@ jobs:
     if: ${{ needs.version.outputs.prev_released == 'true' }}
     runs-on: ubuntu-24.04
     steps:
-    - uses: actions/checkout@v6
```
platreth, line 51: This should be kept, because we are calling ./.github/actions/... next line we need to checkout the repo. 

Will get `Can't find 'action.yml'.`

platreth (dismissed):

.github/workflows/release-2-stabilize.yml
```diff
@@ -143,10 +143,9 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
 
-    - uses: actions/checkout@v6
```
platreth, line 146: Here too

.github/workflows/release-2-stabilize.yml
```diff
@@ -187,10 +186,9 @@ jobs:
...
```
platreth, line 190: And here too

.github/workflows/release-1-branch.yml
Akuukis: Hmm weird, see https://github.com/kobotoolbox/kpi/pull/7365#discussion_r3705449906

how do you test GHA in local? I will push a WIP commit and test on Github :crossed_fingers:

platreth: No I haven't tested in local, I wanted to ask if you did a fake release to try this whole workflow? I updated the comment! The WIP commit and test on Github is perfect, will tell us if everything works fine :)

Akuukis: yup, thanks, you are right and robot advice was wrong, here's the [GHA run](https://github.com/kobotoolbox/kpi/actions/runs/30889759218/job/91928961568) failing as you say

platreth: I commented all of them, there is also the one the bot hallucinated to fix!

.github/actions/checkout-with-github-app-token/action.yml
Akuukis: hmm looks correct actually, it doesn't use `app-id` but still returns app token.

.github/workflows/release-1-branch.yml
Akuukis: pushed & resolved

platreth (approved):

PR #7365, opened 2026-08-03
ci(releases): extract changelog generation into reusable workflow DEV-660

Akuukis:
### 💭 Notes

Extract inline changelog generation from release-2-stabilize into a dedicated reusable workflow (changelog.yml). Supports both draft mode (stabilize-time preview) and authoritative mode (at tag time).

Update release-2-stabilize.yml to use the new reusable workflow.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -28,6 +28,7 @@ jobs:
   changelog:
     runs-on: ubuntu-24.04
     steps:
+    - uses: actions/checkout@v6
     - uses: ./.github/actions/checkout-with-github-app-token
       with:
         client-id: ${{ secrets['client-id'] }}
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -28,11 +28,6 @@ jobs:
   changelog:
     runs-on: ubuntu-24.04
     steps:
-    - uses: actions/checkout@v6
-      with:
-        ref: ${{ inputs.ref }}
-        fetch-depth: 0
-        fetch-tags: true
     - uses: ./.github/actions/checkout-with-github-app-token
       with:
         client-id: ${{ secrets['client-id'] }}
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -35,8 +35,8 @@ jobs:
         fetch-tags: true
     - uses: ./.github/actions/checkout-with-github-app-token
       with:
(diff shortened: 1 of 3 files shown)


.github/workflows/changelog.yml
```diff
@@ -0,0 +1,56 @@
...
+        required: true
+      private-key:
+        required: true
+
+
+jobs:
+  changelog:
+    runs-on: ubuntu-24.04
+    steps:
+    - uses: ./.github/actions/checkout-with-github-app-token
```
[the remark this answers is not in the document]
Akuukis: @greptileai are you sure?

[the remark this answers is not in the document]
Akuukis: @greptileai you're wrong after all, see [GHA run](https://github.com/kobotoolbox/kpi/actions/runs/30889759218/job/91928961568) failing

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,56 @@
...
+      with:
+        client-id: ${{ secrets['client-id'] }}
+        private-key: ${{ secrets['private-key'] }}
+        ref: ${{ inputs.ref }}
+
+    - name: Generate changelog
+      env:
+        VERSION: ${{ inputs.version }}
+      run: |
+        npx git-cliff@2.10.0 -u --tag "$VERSION"
```
platreth, line 42: When calling this in https://github.com/kobotoolbox/kpi/issues/7372 with ref = the tag and draft:false (Line 242). Nothing is unrealeased so the output is empty and gets forced push over the draft, making it empty. I checked in local and we should use --current for this case. 

Something like this: 
```
        if [[ "${{ inputs.draft }}" == "true" ]]; then
          npx git-cliff@2.10.0 -u --tag "$VERSION"
        else
          npx git-cliff@2.10.0 --current
        fi
```

platreth (dismissed):

.github/workflows/changelog.yml
Akuukis: that's a problem for #7372, not here

This PR is a simple refactor and should work on it's own

.github/workflows/changelog.yml
platreth, line 32: This will still fail, after that LGTM

Akuukis: another one! thanks, fixed

platreth (approved):

PR #7368, opened 2026-08-03
ci(releases): cleanup branch workflow DEV-2424

Akuukis:
### 💭 Notes

Cleanup branch workflow:
- rename concurrency group `branch-` → `release-branch-` for clarity
- remove unused `id: branch` from the create step
- simplified the Zulip notification
- fix failure notification condition (`failure()` → `!cancelled() && failure()`)

### DIFF
--- a/.github/workflows/release-1-branch.yml
+++ b/.github/workflows/release-1-branch.yml
@@ -63,21 +63,9 @@ jobs:
         git push origin $NEW_BRANCH
 
 
-  changelog:
-    needs:
-      - version
-      - create-current-branch
-    runs-on: ubuntu-24.04
-    steps:
-    - name: create Linear ticket to track a draft of changelog
-      run: |
-        echo 'TODO, for now do it manually'
-
-
   notify-current:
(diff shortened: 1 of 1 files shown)


.github/workflows/release-1-branch.yml
```diff
@@ -63,21 +63,9 @@ jobs:
...
-    steps:
-    - name: create Linear ticket to track a draft of changelog
-      run: |
-        echo 'TODO, for now do it manually'
-
-
   notify-current:
     needs:
       - version
-      - changelog
```
platreth, line 80: We should add create-current-branch here because the TODO job that got deleted was carrying this dependency chain: notify-current → changelog → create-current-branch.

platreth (changes requested):

.github/workflows/release-1-branch.yml
Akuukis: right! fixed

platreth (approved):

PR #7371, opened 2026-08-03
feat(openRosa): block superuser form submissions by default DEV-32

platreth:
### 📣 Summary

Superuser accounts can no longer submit data to forms, preventing accidental damage to projects from test submissions.

### 📖 Description

Submitting data while signed in as a superuser can corrupt a project's data. Superuser accounts are now blocked from submitting to any form — whether through KoboCollect, Enketo web forms, or the API. Regular accounts are unaffected; use one to submit or test data.

This restriction is **on by default**. Self-hosters who deliberately rely on superuser submissions can re-enable them by setting the environment variable `ALLOW_SUPERUSER_SUBMISSIONS=True`. Doing so re-arms a known footgun — superuser submissions can break projects.

### 💭 Notes

- Guard added at the top of `check_submission_permissions` (`logger_tools.py`), the single chokepoint every submission path flows through (Collect, Enketo web, authenticated, data-collector token, and edits).
- Placed **before** the `require_auth` early-return, so superusers are blocked even on forms that accept anonymous submissions.
- Gated by a Django setting, not Constance — a superuser can't lift their own restriction from `/admin/`.
- Internal callers pass `request=None` and are never affected (imports, restores, etc.).
- The OpenRosa error handler flattens every `PermissionDenied` to a generic `Access denied` 403, so that's what the submitter (Collect/Enketo) sees; the more explicit "use a regular account" wording stays in the code/logs.

### 👀 Preview steps

1. ℹ️ have a superuser account and a deployed project
2. authenticate as the superuser and submit to the form (Enketo web form or the OpenRosa submission API)
3. 🔴 [on main] the submission is accepted
4. 🟢 [on PR] the submission is rejected with a `403` ("Access denied")
5. 🟢 [on PR, with `ALLOW_SUPERUSER_SUBMISSIONS=True`] the submission is accepted again

### DIFF
--- a/kobo/apps/audit_log/tests/test_project_history_logs.py
+++ b/kobo/apps/audit_log/tests/test_project_history_logs.py
@@ -65,7 +65,10 @@ from kpi.utils.xml import (
 
 
 @ddt
-@override_settings(DEFAULT_DEPLOYMENT_BACKEND='mock')
+# These tests create submissions as the superuser `adminuser`, which is blocked
+# by default (DEV-32). Opt into it here since the restriction is not what they
+# exercise.
+@override_settings(DEFAULT_DEPLOYMENT_BACKEND='mock', ALLOW_SUPERUSER_SUBMISSIONS=True)
 class TestProjectHistoryLogs(BaseAuditLogTestCase):
     """
     Integration tests for flows that create ProjectHistoryLogs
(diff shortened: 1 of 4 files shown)


platreth:
@greptileai

platreth:
@greptileai

rgraber (approved):
LGTM

PR #7372, opened 2026-08-03
ci(releases): build a release DEV-659 DEV-1916

Akuukis:
### 💭 Notes

Create a GHA workflow that pulls transifex, tags and builds KPI, as well bumps kobo-docker and kobo-install.

### DIFF
--- a/.github/workflows/release-3-tag.yml
+++ b/.github/workflows/release-3-tag.yml
@@ -2,156 +2,429 @@ name: 'CD: tag'
 
 
 concurrency:
-  group: tag-${{ github.ref }}
-  cancel-in-progress: true
+  group: release-tag
+  cancel-in-progress: false
 
 
+# Pipeline: determine version → pull-transifex → tag → build-image → [bump, linear, deploy-to-master] → notify
 on:
-  push:
-    tags:
-      - v2.0**
+  workflow_dispatch:
+    inputs:
+      release_branch:
+        description: 'Release branch to tag (e.g. release/2.026.27). Leave empty to auto-detect the latest.'
+        required: false
+        type: string
+      force:
+        description: 'Skip branch eligibility checks (for security hotfixes on older branches).'
+        required: false
+        type: boolean
+        default: false
 
 
 jobs:
 
-  darker:
-    uses: ./.github/workflows/darker.yml
-
-  biome:
-    uses: ./.github/workflows/biome.yml
-
-  openapi:
-    uses: ./.github/workflows/openapi.yml
-
-  pytest:
-    uses: ./.github/workflows/pytest.yml
-
-  npm-test:
-    uses: ./.github/workflows/npm-test.yml
-
-  version:
+  # Determine which release branch to work on.
+  # If the user provided one, use it. Otherwise, find the latest release branch.
+  current-branch:
     runs-on: ubuntu-24.04
     outputs:
-      version: ${{ steps.version.outputs.version }}
+      current_branch: ${{ steps.branch.outputs.current_branch }}
     steps:
     - uses: actions/checkout@v6
-    - name: Formats the version number
-      id: version
+      with:
+        fetch-depth: 0
+        fetch-tags: true
+    - name: Determine release branch
+      id: branch
+      env:
+        INPUT_BRANCH: ${{ inputs.release_branch }}
       run: |
         set -xe
-        TAG=$GITHUB_REF
-        VERSION=$(echo "$TAG" | sed 's/v//')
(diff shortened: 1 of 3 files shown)


.github/workflows/release-3-tag.yml
```diff
@@ -2,156 +2,429 @@ name: 'CD: tag'
...
+
+  # Bump version references in kobo-docker and kobo-install, then tag them.
+  bump:
+    runs-on: ubuntu-24.04
+    needs:
+      - version
+      - build-image
+    if: ${{ !cancelled() && !failure() }}
+    steps:
+    - uses: ./.github/actions/checkout-with-github-app-token
```
platreth: Here it needs a checkout as well. Better use `    - uses: actions/create-github-app-token@v3`

platreth (dismissed):

.github/workflows/release-3-tag.yml
```diff
@@ -2,156 +2,428 @@ name: 'CD: tag'
...
 
-  version:
+  # Validate that this branch is eligible for tagging:
+  # - The latest minor is always allowed (next_branch == 'main')
+  # - The previous minor is allowed only if the latest minor is unreleased
+  # - Older branches require the `force` input
+  validate:
+    needs:
+      - version
+    if: ${{ !cancelled() && !failure() && inputs.force != true }}
```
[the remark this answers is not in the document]
Akuukis: the other jobs have `if: ${{ !cancelled() && !failure() }}` already, does it solve the issue?

.github/workflows/release-3-tag.yml
```diff
@@ -2,156 +2,428 @@ name: 'CD: tag'
...
+    needs:
+      - version
+      - tag
+    if: ${{ !cancelled() && !failure() }}
+    uses: './.github/workflows/changelog.yml'
+    secrets:
+      client-id: ${{ secrets.KOBO_BOT_CLIENT_ID }}
+      private-key: ${{ secrets.KOBO_BOT_PRIVATE_KEY }}
+    with:
+      ref: refs/tags/${{ needs.version.outputs.current_patch }}
```
platreth: Following this discussion : https://github.com/kobotoolbox/kpi/pull/7365#discussion_r3710823188

Yeah you're right it should be here, I did this: 

```
git worktree add --detach /tmp/cliffw 2.026.27d  

npx git-cliff@2.10.0 -u --tag 2.026.27d   -> empty

npx git-cliff@2.10.0 --current     -> has a correct text
```

Akuukis: yup, and here's the fix [30763e0](https://github.com/kobotoolbox/kpi/pull/7372/commits/30763e0fb2ce629abaedb3dfb9b66a5d3a331e61) (not tested, seemed obvious)

.github/workflows/release-3-tag.yml
```diff
@@ -2,156 +2,428 @@ name: 'CD: tag'
...
+        fi
+        git add .
+        git commit -m "chore: pull transifex translations for $VERSION"
+        git push
+
+
+  # Regenerate changelog with all tags known (authoritative, replaces stabilize draft).
+  changelog:
+    needs:
+      - version
```
platreth, line 172: We need to add `-validate` because if validate fails, the changelog will run anyway. It was not the case before because it was in a different order

platreth:
If we need to rerun this workflow after a successful tag, it will not retry, is that okay? (in case we have a failed bump, we would need a new tag basically)

.github/workflows/release-3-tag.yml
Akuukis: fixed in the new squashed commit

.github/workflows/release-3-tag.yml
```diff
@@ -2,156 +2,429 @@ name: 'CD: tag'
...
     runs-on: ubuntu-24.04
+    steps:
+    - uses: actions/checkout@v6
+      with:
+        ref: ${{ needs.version.outputs.current_branch }}
+    - uses: ./.github/actions/checkout-with-github-app-token
+      with:
+        client-id: ${{ secrets.KOBO_BOT_CLIENT_ID }}
+        private-key: ${{ secrets.KOBO_BOT_PRIVATE_KEY }}
+        ref: ${{ needs.version.outputs.current_branch }}
```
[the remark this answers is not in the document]
Akuukis: @greptileai not a problem. "a repository collaborator" requires write access to repo and those people are vetted. Furthermore, if it has write access, there are easier ways to steal the KOBO_BOT_PRIVATE_KEY.

platreth (approved):

PR #7374, opened 2026-08-03
ci(releases): integrate Linear Releases DEV-2424

Akuukis:
### 💭 Notes

Integrate Linear Releases in the release cycle. See also Handbook PR for the big picture.

### DIFF
--- a/.github/workflows/release-1-branch.yml
+++ b/.github/workflows/release-1-branch.yml
@@ -63,10 +63,36 @@ jobs:
         git push origin $NEW_BRANCH
 
 
-  notify-current:
+  # Create Linear release and link issues from commits unique to this branch.
+  # Uses prev branch tip (not prev tag) so commits to be released as patches on
+  # the previous minor are excluded.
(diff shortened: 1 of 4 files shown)


.github/workflows/release-1-branch.yml
```diff
@@ -61,9 +61,36 @@ jobs:
...
+    needs:
+      - version
+      - create-current-branch
+    runs-on: ubuntu-24.04
+    steps:
+    - uses: actions/checkout@v6
+      with:
+        fetch-depth: 0
+        fetch-tags: true
+    - uses: linear/linear-release-action@v0
```
platreth: Might be better to tag a specific version? Linear could release a buggy version and it could break the pipeline

Akuukis: yes, good point. I already did it but lost that change somewhere mid all rebases

platreth (approved):
LGTM after changing the version of linear

PR #7381, opened 2026-08-04
fix(datatable): fallback values for missing choice names DEV-2573

jamesrkiger:
### 📣 Summary

Fixes the select one/many column filter dropdown on the data table to handle choices with a missing `name` by falling back to `$autoname` then `$autovalue` before dropping value from list.

### 👀 Preview steps

1. Upload the following xlsform as a new project:
[missing-name-select.xlsx](https://github.com/user-attachments/files/30703068/missing-name-select.xlsx)
2. Make a few submissions
3. Open project data table
4. 🔴 [on main] notice the mantine component throws an error that breaks entire view
5. 🟢 [on PR] Verify table renders and dropdown correctly shows choice labels with usable filter values
6. Select a choice from the dropdown
7. 🟢 notice the table filters correctly to rows matching that choice

### DIFF
--- a/jsapp/js/components/submissions/TableDropdownFilter.tsx
+++ b/jsapp/js/components/submissions/TableDropdownFilter.tsx
@@ -3,7 +3,7 @@ import type { Column, Filter, FilterRender } from 'react-table'
 import { getQuestionOrChoiceDisplayName } from '#/assetUtils'
 import Select from '#/components/common/Select'
 import type { TableColumn } from '#/components/submissions/table.types'
-import type { SurveyChoice } from '#/dataInterface'
+import type { LabelValuePair, SurveyChoice } from '#/dataInterface'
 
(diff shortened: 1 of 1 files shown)


jsapp/js/components/submissions/TableDropdownFilter.tsx
```diff
@@ -26,24 +30,20 @@ const TableDropdownFilter: FilterRender = (props: TableDropdownFilterProps) => {
   const selectFromListName = 'selectFromListName' in props.column ? props.column.selectFromListName : undefined
   const translationIndex = 'translationIndex' in props.column ? props.column.translationIndex || 0 : 0
 
-  // Duplicate values can exist in the choices array, so we use a Set to filter them out
+  interface ChoiceOption {
```
magicznyleszek: You could use `LabelValuePair`, but :shrug:

magicznyleszek (approved):
Works, LGTM

PR #7386, opened 2026-08-04
feat(userReports): add LLM usage to user reports DEV-2571

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Add usage summaries for LLM requests to the user reports endpoint.

### 💭 Notes
Requires a small refactor of how we handle dropping/creating the MV with migrations.
Previously, if a migration required changing the MV SQL and recreated it, previous migrations that tried to drop/recreate the MV were changed to noops so they wouldn't try to run new SQL on old models (which often leads to errors because of missing columns).

This PR adds a concept of versioning to the CREATE_MV_SQL constant, and migrations only execute the drop/recreate if the version matches what they expect. This way new instances won't try to create the view until after all previous migrations to adjust necessary columns are completed. It won't matter on old instances because those migrations will have already run. 

This will prevent having to go back and edit old migrations, which is generally considered an anti-pattern in Django.

Also moves the drop/recreate code to a shared file and removes a method that was not being called anywhere.

### 👀 Preview steps
Make sure to run migrations

1. ℹ️ have an account and a project with at least one audio question and add least one submission
2. Add a QA question to the audio question and use `Generate with AI` to fill it in
3. Run `./manage.py manage_user_reports_mv --drop` followed by `./manage.py manage_user_reports_mv --create --force` in the kpi container
4. As a superuser, navigate to `/api/v2/user-reports`
5. 🟢 Notice there is now information on LLM usage, limits, and balances


kobo/apps/user_reports/migrations/0009_billingandusagesnapshot_llm_requests_limit.py
```diff
@@ -0,0 +1,18 @@
...
+        ('user_reports', '0008_drop_mv_before_djstripe'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='billingandusagesnapshot',
+            name='llm_requests_limit',
+            field=models.BigIntegerField(blank=True, null=True),
+        ),
+    ]
```
rajpatel24: 0009 only adds the column to BillingAndUsageSnapshot. Nothing in the PR drops and recreates `user_reports_userreportsmv`, and the API reads exclusively from the MV.               
                                                                                                                                                                                   
On a fresh install this works by accident: `0008_drop_mv_before_djstripe` drops the MV and reschedules the LRM, so the background job rebuilds it after `0009` has run. On any environment where the MV already exists, `0007` and `0008` are already applied, LRM `0019_recreate_user_reports_mv` is already completed, and nothing re-runs.

I think we should add the MV rebuild to `0009` itself, following the approach used in `long_running_migrations/0022_sync_mfa_profiles.py`, drop the MV, then reset LRM `0019` to `created` so the background job rebuilds it.

Something like this:

```python
class Migration(migrations.Migration):
    atomic = False

    dependencies = [
        ('user_reports', '0008_drop_mv_before_djstripe'),
        ('long_running_migrations', '0019_recreate_user_reports_mv'),
    ]

    operations = [
        migrations.AddField(
            model_name='billingandusagesnapshot',
            name='llm_requests_limit',
            field=models.BigIntegerField(blank=True, null=True),
        ),
        migrations.RunSQL(sql=DROP_MV_SQL, reverse_sql=migrations.RunSQL.noop),
        migrations.RunPython(reschedule_lrm_recreate, noop),
    ]
```

I think this also lets us remove `get_create_mv_sql()`. Once `0009` owns the rebuild, `0007` is effectively superseded and can be turned into a no-op, just like `0002`–`0006`, which are already `RunPython.noop`.

What do you think about that?

Also, `./manage.py manage_user_reports_mv --force` from the preview steps is effectively a no-op. It doesn't do anything unless `--drop` or `--create` is passed.

rajpatel24 (changes requested):

kobo/apps/user_reports/migrations/0009_billingandusagesnapshot_llm_requests_limit.py
rgraber: @rajpatel24 good call. i assume i want to go through all the lock-management we do in 0008_drop_mv_before_djstripe instead of just dropping the view, right?

rajpatel24 (approved):
LGTM!

Couple of non-blocking requests: 
- Can we add a short comment for `CREATE_MV_SQL_VERSIONS` explaining what it’s for and how to add a new version? It would make things easier for a new dev working on this. What do you think?
- For consistency, let’s add `llm_requests` to the filterable metric lists in the API docs: `kpi/docs/api/v2/user_reports/list.md` lines 24–27, 40–41, and 76.

PR #7388, opened 2026-08-05
feat(queryParser): filter assets by multiple tags DEV-1581

platreth:
### 📣 Summary

Filtering projects by several tags now returns only the projects that have all of them.

### 📖 Description

Combining tags in a search (e.g. "foo" and "bar") used to return nothing. It now returns the projects tagged with every tag asked for. Single-tag and "any of these tags" searches are unchanged.

### 💭 Notes

The `q` parser merged every `AND` into one `.filter()`; on a to-many like `tags`, `tags__name:foo AND tags__name:bar` asks a single joined row to be both → zero results. Fix: `parse()` now builds every leaf crossing a to-many relation as its own correlated `EXISTS` subquery, i.e. an independent "some related row matches" predicate, so `AND`/`OR`/`NOT` compose with set semantics in any tree shape. To-many detection shares the security allowlist's path walk (`_iter_field_path`). Leaves generated from `default_field_lookups` stay plain lookups — a bare search term expands to one leaf per default lookup, and `tags__name__icontains` is one of them, so a subquery each would make every word of a plain search cost one. The number of explicit to-many lookups a query may contain is capped by `QUERY_PARSER_MAX_TO_MANY_FILTERS`, and the library tag filter is capped to match.

### 👀 Preview steps

1. ℹ️ three projects: one tagged `foo`, one `bar`, one both
2. search `?q=tags__name:foo AND tags__name:bar`
3. 🔴 [main] nothing returned
4. 🟢 [PR] only the project tagged both
5. 🟢 [PR] `?q=(tags__name:foo AND tags__name:bar) OR asset_type:survey` also returns the both-tagged project even when it is not a survey


platreth:
@greptileai

kpi/utils/query_parser/query_parser.py
```diff
@@ -338,16 +327,62 @@ def _validate_field(self, field):
...
+    if model is None:
+        return False
+
+    return any(
+        model_field.many_to_many or model_field.one_to_many
+        for _, _, model_field in _iter_field_path(model, field)
+    )
+
+
+def split_relational_and(q_obj: Q, model: Optional[ModelClass]) -> tuple:
```
noliveleger: Maybe we should specify what kind of tuple. e.g. ` -> tuple[Q, list[Q]]`

kpi/utils/query_parser/query_parser.py
```diff
@@ -338,16 +327,62 @@ def _validate_field(self, field):
...
+        yield current_model, segment, model_field
+
+        if not model_field.is_relation or model_field.related_model is None:
+            return
+
+        current_model = model_field.related_model
+
+
+def _crosses_to_many(model, field):
+    """Whether `field` traverses a to-many relation (many-to-many or reverse FK)."""
```
noliveleger: Would you mind to write it on multiline? 

```
"""
Whether ...
"""

 if model is None
```

Not required, but I think it's better especially if we change the docstring at a later time. It's easier to review.

kpi/filters.py
```diff
@@ -485,13 +490,18 @@ def filter_queryset(self, request, queryset, view):
...
         try:
-            # If we are searching on an n-to-many field, we may get multiple results
-            # from the same model, so we need to de-duplicate with distinct(). Rely
-            # on the view to tell us if this is not necessary
+            # A to-many AND needs chained `.filter()` calls, not one merged
+            # `.filter()` (which matches no rows). See `split_relational_and`.
+            main_q, chained_qs = split_relational_and(q_obj, queryset.model)
+            filtered = queryset.filter(main_q)
+            for chained_q in chained_qs:
+                filtered = filtered.filter(chained_q)
```
noliveleger: This should be restricted. We don't want to allow _n_ joins (imagine `tags__name:foo OR ... tags__name:bar`) with 50 `tags__name`..

noliveleger (changes requested):
Found a problem.
Create a collection with two tags `foo`, `bar`
                                                                                                                              
```                                                                                                                         
q=tags__name:foo AND tags__name:bar                        -> returns it                                                    
q=(tags__name:foo AND tags__name:bar) OR asset_type:survey -> does not return it                                            
```

noliveleger:
@platreth Also add a regression test to be sure that kpi#7243 still works. (It does, but just in case).

platreth:
@noliveleger Treated all the comments! Thanks for the good eye :) Let me know!

kpi/utils/query_parser/query_parser.py
```diff
@@ -349,40 +350,122 @@ def _iter_field_path(model, field):
         current_model = model_field.related_model
 
 
-def _crosses_to_many(model, field):
-    """Whether `field` traverses a to-many relation (many-to-many or reverse FK)."""
+def _to_many_relation_root(model, field):
```
noliveleger: Keep A -> Z . public first, private second. 🙏

noliveleger (changes requested):
Let's remove "// BUG: this doesn't work correctly - it does filter one tag, but if multiple are selected it returns zero values" in the front-end code. 

https://github.com/kobotoolbox/kpi/blob/3c89222943e86779ee3b48d0c4a2c4566a39e270/jsapp/js/editorMixins/AssetNavigator.tsx#L81

kpi/utils/query_parser/query_parser.py
```diff
@@ -338,16 +328,144 @@ def _validate_field(self, field):
...
+
+def _crosses_to_many(model, field):
+    """
+    Whether `field` traverses a to-many relation (many-to-many or reverse FK).
+    """
+    return _to_many_relation_root(model, field) is not None
+
+
+def _count_to_many_leaves(q_obj: Q, model: Optional[ModelClass]) -> int:
+    """Count, across the whole tree, leaves whose lookup crosses a to-many."""
```
noliveleger: Multiline docstring please 🙏

kpi/tests/test_query_parser.py
```diff
@@ -194,3 +199,91 @@ def test_viewset_level_allowlist_overrides():
...
     )
     assert parsed_q is not None
+
+
+def _parse(query):
+    return parse(query, default_field_lookups=['name__icontains'], model=Asset)
+
+
+def _leaf_keys(q):
+    """Flatten a `Q` tree into the list of lookup keys on its leaf tuples."""
```
noliveleger: Multiline docstring please 🙏

kpi/utils/query_parser/query_parser.py
```diff
@@ -338,16 +328,144 @@ def _validate_field(self, field):
...
+        if len(leaves) < 2:
+            result.extend(leaves)
+            continue
+        # `_base_manager` never narrows the row set, so the subquery is a pure
+        # "has a row for each leaf" predicate; the outer queryset keeps its own
+        # scope when it applies `pk__in`.
+        subquery = model._base_manager.all()
+        for leaf in leaves:
+            subquery = subquery.filter(Q(leaf))
+        result.append(('pk__in', subquery.values('pk')))
```
noliveleger: performance wise, this makes me worried 😟 . I think we should remove any `order by` here.  
For example, on Asset. We should try to order thousands of id for nothing. 
What about something like `subquery.order_by().values('pk')`

noliveleger (changes requested):

kpi/utils/query_parser/query_parser.py
platreth: Hmm Django was already dropping it by default! But the query was not optimized (each extra tag added 2 more JOINs to one big subquery). I reworked that! Let me know if you like it :)

noliveleger (approved):
LGTM

PR #7389, opened 2026-08-05
refactor(frontend): SSOT for columns order DEV-1480 DEV-1481 DEV-1296

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Columns are now listed in the same order in every place that shows them: Data Table, "hide fields" menu, Single Submission modal, and the field picker in Downloads.

### 💭 Notes

Single Submission modal used to list a fixed set of metadata rows at the bottom, regardless of the project. It now lists the metadata the project and the submission actually have, in the same order as the table columns.

Changes here:

- **`tableUtils.ts`**
  - new `orderColumns()` — the single source of truth for column order. It's applied at the end of `getAllDataColumns()`, so everything built on top of that function inherits the order instead of re-inventing it
  - it reorders *blocks* (a question plus the supplemental details columns that follow it), so NLP columns stay glued to their source question.
  - new `getMetadataColumns()` — filters `getAllDataColumns()` down to meta questions + back end submission props, rather than hand-building another list that could drift.
- **`tableConstants.ts`** — `LAST_COLUMNS_ORDER` (the trailing metadata order, formerly the `z1`…`z999` prefixes) and `METADATA_COLUMN_LABELS` (formerly hardcoded in the submission modal).
- **`table.tsx`** — deleted the ~100 line `switch (key)` that invented sort keys (`z1`, `y_`, `2_5_supplementalDetails/…`) and the `localeCompare` sort that consumed them. The `index` property is gone from `TableColumn` too — it's not a `react-table` property.
- **`submissionDataTable.tsx`** — metadata rows are mapped from `getMetadataColumns()` instead of 12 hardcoded `renderMetaResponse()` calls.
- **`tableStore.ts`** — `getHideableColumns()` now has proper ordering.
- **`ProjectExportsCreator.tsx`** — export field list goes through `orderColumns()`. The list stays wider than the table's on purpose, some props are worth exporting without being worth a column.
- **`assetUtils.ts` + `submissionUtils.tsx` + `TabAnalysis/common/constants.ts`** — the two supplemental-field filters dropped different sets (`qualNote` here, `qualNote` + `qualSource` there). Both now call one shared `isDisplayableSupplementalField()`, so they can't disagree again.
- **tests** — order and metadata cases in `tableUtils.tests.ts`, supplemental injection cases in `assetUtils.tests.js`.

Also note - the bulk edit modal builds its own list from displayed columns, so it follows along for free.

### 👀 Preview steps

1. ℹ️ have a project with a background-audio question, a few questions inside groups, and metadata enabled in Layout & Settings → Metadata (username, device ID, phone number, today — but **not** audit)
2. ℹ️ have at least one submission
3. go to Project → Data → Table
4. 🟢 notice the column order: background audio first, then `start` and `end`, then the form questions in form order, then username, device ID, phone number, today, and the submission metadata (`__version__`, `_id`, `_uuid`, submission time, submitted by, rootUuid) at the very end
5. click "hide fields"
6. 🔴 [on main] notice the list is in a different order than the table columns
7. 🟢 [on PR] notice the list order matches the table exactly, with Validation at the top
8. hide a couple of columns, freeze a column, sort by a column — 🟢 notice all of it still works
9. back in the table, open a single submission (the expand icon in the first column) and scroll to the metadata rows at the bottom
10. 🔴 [on main] notice `audit` is listed even though this project doesn't collect it, and `instanceID` is there
11. 🟢 [on PR] notice `audit` and `instanceID` are gone, `_uuid` and submission time are listed, and the rows come in the same order as the table columns
12. go to Project → Data → Downloads → "Select questions to be exported"
13. 🟢 notice the field list follows the same order as the table
14. ℹ️ for the qualitative analysis bit: use a project with an audio question that has a transcript and a few analysis questions, including a Note
15. 🟢 notice the Note question isn't a column in the table, isn't a row in the single submission view, and isn't in the export field list — and neither is anything named "source"


magicznyleszek:
@greptileai review please

pauloamorimbr (approved):
LGTM! 

Code looks good. Tested and it works as described! 👍🏻

PR #7390, opened 2026-08-05
 fix(trashBin): restart trash bin tasks which failed on transient errors DEV-2511

rajpatel24:
### 📣 Summary
This PR makes trash bin deletion jobs that fail on a transient infrastructure error retry themselves automatically, by leaving them in-progress so the existing task restarter picks them up, and fixes three pre-existing bugs in that restarter including one that deleted accounts a superuser was supposed to approve first.

#### Problem
Trash bin deletion jobs (accounts, projects, attachments) that died on an infrastructure error (MongoDB unreachable, a PostgreSQL deadlock, a Celery time limit, an OOM kill) were marked FAILED and left there. Nothing retried them, so half-deleted objects lingered until someone noticed and restarted them by hand. 

#### What this changes 
When a job fails, its error is matched against a list of known-transient patterns. If it matches, the object is left `IN_PROGRESS` instead of `FAILED` which is exactly what the existing `task_restarter` already looks for, so it gets restarted with no new machinery. Any other failure still goes to FAILED and waits for a human.

Attempts are counted and capped: once an object exceeds TRASH_BIN_MAX_AUTO_RESTARTS it is flagged FAILED and stops retrying, so it can't loop forever and becomes visible to a human.

#### Three pre-existing bugs on main 
1. The stuck check was effectively dead for started deletions. It required an object to be roughly twice its retention period old before being eligible. (e.g., 7-day-retention project failing on day 7 was only restarted on day 14; 360 days for accounts)
2. Accounts held for manual deletion were deleted automatically. Their scheduled time is computed into the past, and the restarter reads that timestamp rather than the disabled Celery schedule so it force-started the deletion ~75–105 minutes after trashing, voiding the "a superuser must approve this" guarantee.
3. A manual deletion that genuinely got stuck was never restarted. The mirror image: once retention is changed to a normal value, that same past timestamp fails the check, so a superuser-started deletion whose worker died sat half-complete in IN_PROGRESS for months.

The fix for all three: a deletion that is already in progress has by definition passed its scheduled time, so that time is no longer consulted for it. Only objects that never started still get the "is it due yet?" check, and those are additionally never auto-started when they're waiting for a superuser.

### 👀 Preview steps

1. ℹ️ have an account with a project
2.  In the Django shell, move the project to trash, then simulate a transient failure by calling `trash_bin_task_failure()` with a `deadlock detected..` message.
5. 🟢 [on PR] the trash object stays in `in_progress` with the error recorded in `metadata['failure_error']` and `metadata['retryable_failure_count']` at 1
6. ℹ️ it is not restarted immediately, the restarter only picks up objects that have gone quiet for longer than celery hard time limit. Either wait, or backdate the trash object's `date_modified` by more than that
7.  Run `task_restarter()`
8. 🔴 [on main] nothing happens, the object stays `FAILED` forever and the project is never deleted. Only a manual restart from the admin clears it
9. 🟢 [on PR] the restarter picks it up and re-runs the deletion, and the  project is deleted
10. 🟢 [on PR] repeat steps 2–5 with a non-transient message such as `Asset matching query does not exist.`. The object goes to `FAILED`

If you still want the exact shell commands, they're in [test_restart_failed_trash_job.txt](https://github.com/user-attachments/files/30740435/test_restart_failed_trash_job.txt)


rajpatel24:

kobo/apps/trash_bin/tasks/__init__.py
```diff
@@ -61,56 +65,76 @@ def garbage_collector():
...
-        )
-        .order_by('date_modified')[:settings.MAX_RESTARTED_TASKS]
-    )
-    for stuck_account_id in stuck_account_ids:
-        empty_account.delay(stuck_account_id, force=True)
+    # A deletion which has already started has, by definition, passed its
+    # scheduled time: looking at the clocked time again would only delay its
+    # restart by a whole retention period and that time is set in the past
+    # when the object is trashed manually anyway
+    started_objects = Q(status=TrashStatus.IN_PROGRESS)
```
noliveleger, line 98: The behaviour is now changed. It uses to wait for 2 x `CELERY_LONG_RUNNING_TASK_TIME_LIMIT`

`stuck_threshold` uses `CELERY_LONG_RUNNING_TASK_TIME_LIMIT` + 5min as the buffer. `time_limit` is measured from when Celery actually starts executing the task, not from when it's claimed/enqueued here. If the broker queue backs up for more than 5 minutes before the worker picks it up, the total elapsed time (queue wait + execution) can exceed this threshold while the task is still legitimately running. The restarter would then treat it as stuck, claim it, and dispatch a second force=True run in parallel with the first.


Should we add a safety guard by checking in `i.active()` before calling `.delay()` ? Or adding a bigger buffer just in case?

kobo/apps/trash_bin/utils/trash.py
```diff
@@ -279,22 +285,38 @@ def put_back(
...
+
+        # Transient failures (OOM kill, MongoDB unreachable, deadlock, Celery
+        # time limits) are left in progress, so that `task_restarter` grabs them
+        # like any other task whose worker died. Attempts are counted and the
+        # object is only flagged as failed. i.e.: left for a human - once it
+        # has exhausted its budget, so it cannot restart forever
+        obj_trash.status = TrashStatus.FAILED
+        if is_retryable_failure(error):
+            restart_count = (obj_trash.metadata.get('retryable_failure_count') or 0) + 1
+            obj_trash.metadata['retryable_failure_count'] = restart_count
```
noliveleger, line 336: I'm not sure we want to stop trying a big account/project to be deleted. I can see a task to be killed by Celery because of the soft limit more than 5 times.

`TRASH_BIN_MAX_AUTO_RESTARTS` is probably too low IMO

noliveleger (changes requested):
Do we also cover cases when the pod is OOM killed? I guess it just stays in IN_PROGRESS until the `task_restarter` can restart it?

noliveleger:
@rajpatel24 This PR does not cover (i.e. : reset) old tasks which failed on transient errors, right?

kobo/settings/base.py
```diff
@@ -2392,6 +2392,10 @@ def __init__(self, *args, **kwargs):
 MAX_RESTARTED_TASKS = 100
 MAX_RESTARTED_TRANSFERS = 20
 
+# Number of times a trash bin task that failed on a transient (infrastructure)
+# error is automatically restarted before it requires manual intervention
+TRASH_BIN_MAX_AUTO_RESTARTS = 5
```
noliveleger: What about using a env var here and fallback on an integer?

kobo/apps/trash_bin/mixins/admin.py
```diff
@@ -16,6 +17,13 @@ def empty_trash(self, request, queryset, **kwargs):
             messages.SUCCESS,
         )
 
+    @admin.display(description='Automatic restarts')
+    def get_auto_restarts(self, obj):
+        restart_count = obj.metadata.get('retryable_failure_count') or 0
+        if not restart_count:
+            return '-'
+        return f'{restart_count}/{settings.TRASH_BIN_MAX_AUTO_RESTARTS}'
```
noliveleger: No need to add `/{settings.TRASH_BIN_MAX_AUTO_RESTARTS}'  to every line. 

it could be in the head e.g. `RESTARTS (max: {settings.TRASH_BIN_MAX_AUTO_RESTARTS})`

kobo/apps/trash_bin/utils/trash.py
rajpatel24: Agreed. One scenario I was thinking of is that `delete_account()` deletes assets one at a time outside of a wrapping transaction, so each restart still makes real progress. Capping retries at 5 could strand a large account that was steadily working through its deletion. I've increased the limit to 30 for now, but let me know what you think would be a better value

kobo/apps/trash_bin/tasks/__init__.py
rajpatel24: I believe the queue-wait part doesn't apply: `process_deletion()` refreshes `date_modified` when the worker starts executing, so the threshold runs from execution start, not from enqueue.

On the old behaviour: it wasn't 2 × CELERY_LONG_RUNNING_TASK_TIME_LIMIT. `in_progress` objects also needed `clocked_time <= now - retention`, and since `clocked_time` is `trashed_at` + retention, that means an object had to be twice the retention period old before it was eligible.

That said, I do think two tasks can be enqueued for the same object. So I added a lock in `process_deletion()` with a TTL of `CELERY_LONG_RUNNING_TASK_TIME_LIMIT + 60`, shorter than the stuck threshold so a dead worker can't block its own restart. Duplicates can still be enqueued, but the second one exits without doing anything.

I didn't use `i.active()` because it only reports running tasks, a queued restart is invisible to it, so we'd dispatch the duplicate anyway. A bigger buffer only makes it rarer, and delays legitimate recovery.

rajpatel24:
> Do we also cover cases when the pod is OOM killed? I guess it just stays in IN_PROGRESS until the task_restarter can restart it?

@noliveleger  Yes, covered. If Celery survives the OOM it raises Worker exited prematurely, which is in the retryable list, so the object stays `in_progress`. If the whole pod dies, nothing runs and it simply stays `in_progress`. Either way `task_restarter` picks it up once it looks stuck.

> This PR does not cover (i.e. : reset) old tasks which failed on transient errors, right?

Correct, it didn't. But I have now added a long-running migration (`0031_restart_transient_trash_failures`) that puts failed objects with a retryable error back to `in_progress` in small batches, so a backlog trickles out instead of landing at once. Non-transient failures are left alone.

kobo/settings/base.py
```diff
@@ -2394,7 +2394,11 @@ def __init__(self, *args, **kwargs):
 
 # Number of times a trash bin task that failed on a transient (infrastructure)
 # error is automatically restarted before it requires manual intervention
-TRASH_BIN_MAX_AUTO_RESTARTS = 5
+TRASH_BIN_MAX_AUTO_RESTARTS = 30
+
+# How long a trash bin object stays locked while it is being deleted. Must stay
+# shorter than the delay after which `task_restarter` considers a task stuck
+TRASH_BIN_DELETION_LOCK_TTL = CELERY_LONG_RUNNING_TASK_TIME_LIMIT + 60
```
noliveleger: We usually use a buffer of 5 minutes everywhere else. 

In the comment, I would also add, it must be >=  than the celery hard limit of the task.

kobo/settings/base.py
```diff
@@ -2394,7 +2394,11 @@ def __init__(self, *args, **kwargs):
 
 # Number of times a trash bin task that failed on a transient (infrastructure)
 # error is automatically restarted before it requires manual intervention
-TRASH_BIN_MAX_AUTO_RESTARTS = 5
+TRASH_BIN_MAX_AUTO_RESTARTS = 30
```
noliveleger: The more I think about it, the more I think we should have different values for each type of trash bin object? Deleting an attachment won't take as long as an account or a project. Right? 
30 retries may be too much for attachments? 

Moreover... We should not increase the counter when the error is "TimeLimit".  If Celery kills the task because it took longer than 4200s, it's probably because the object is pretty big and needs that time. 
If a query gets stuck, PostgreSQL timeout should be triggerred and then the auto restart counter could be increased. Skipping "TimeLimit" could make use a small maximum. 

What do you think?

kobo/apps/long_running_migrations/jobs/0031_restart_transient_trash_failures.py
```diff
@@ -0,0 +1,47 @@
...
+from django.db.models import Q
+
+from kobo.apps.trash_bin.constants import RETRYABLE_FAILURE_PATTERNS
+from kobo.apps.trash_bin.models import TrashStatus
+from kobo.apps.trash_bin.models.account import AccountTrash
+from kobo.apps.trash_bin.models.attachment import AttachmentTrash
+from kobo.apps.trash_bin.models.project import ProjectTrash
+from kpi.utils.log import logging
+
+CHUNK_SIZE = settings.LONG_RUNNING_MIGRATION_SMALL_BATCH_SIZE
```
noliveleger: I think we can use LONG_RUNNING_MIGRATION_BATCH_SIZE here. It should be a quick update!

noliveleger (changes requested):

kobo/settings/base.py
rajpatel24: Agreed on skipping "TimeLimit." For per-type values, I’d lean towards keeping just one. Now that time limits don’t count, the counter only sees deadlocks, Mongo timeouts, and OOM kills, which indicate that "something is wrong" rather than "this object is big."

I’ve kept it at 10 for now, but let me know what you think. Happy to update it if you disagree.

noliveleger (approved):
LGTM

PR #7391, opened 2026-08-05
refactor(myProjects): replace react-infinite-scroller with in-house solution DEV-2074

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Scrolling through a long list of projects now loads the next batch more reliably, and a failed load offers a Retry button instead of just an error message.

### 💭 Notes

Changing a filter or a sort order while more projects were still loading could mix rows from the old list into the new one. And when loading more projects failed, the only feedback was an error toast, with no way to try again short of reloading the page — there's now a Retry button at the bottom of the list.

Once you've scrolled to the very bottom of a long list, the table says "You've reached the end of the list". Lists that fit on a single page don't get that message, since the end is plain to see.

Swaps `react-infinite-scroller` for our own `InfiniteScrollTrigger` in the projects table. The old plugin never declared React 19 support and (thus) was holding up the Mantine v9 upgrade, so it has been shown the door.

Changes here:

- **`projectsTable.tsx`** — uses `InfiniteScrollTrigger`
- **`customViewStore.ts`** — two bugs were present:
  - `fetchMoreAssets` set no loading flag and aborted any in-flight request, so repeated firing would abort and restart the *same* page fetch in a loop. Now sets `isLoadingNextPage` and ignores re-entrant calls.
  - `fetchAssets` didn't reset `nextPageUrl`, so a filter or order change left a stale "next page" pointing at results we'd just dropped. Now cleared, along with both new flags.
  - New `hasNextPageError` powers the Retry button. Aborts are deliberately not treated as errors, so a request we cancelled ourselves doesn't staple a Retry button onto a list that is loading just fine.
- **`api.ts`** — `isAbortResponse` factored out of `handleApiFail` and exported, for callers keeping their own error state.
- **`projectsTable.module.scss`** — removed now unused styles
- **`customViewStore.tests.ts`** — 5 cases covering the re-entry guard, the `nextPageUrl` reset, and both error paths.

### 👀 Preview steps

1. ℹ️ have an account with **more than 50 projects** (the list loads 50 at a time, so a shorter list won't scroll far enough to prove anything) - you can drag and drop 40 at a time from [many small.zip](https://github.com/user-attachments/files/30743406/many.small.zip)
2. go to **My Projects**
3. scroll to the bottom of the table
4. 🟢 notice a spinner appears briefly and the next 50 projects load
5. keep scrolling to the very bottom of the whole list
6. 🟢 notice the message "You've reached the end of the list"
7. now the mid-flight test: reload the page, scroll down until the spinner shows, and while it's still spinning change a filter (**Filter** button) or click a column header to re-sort
8. 🟢 notice the table resets to a clean first page of results, with no leftover rows from the previous list mixed in
9. 🔴 [on main] the same steps show no end-of-list message, and step 7 can leave the table with a jumbled mix of old and new rows
10. optional, for the pessimists: open devtools → Network, set it to **Offline**, then scroll to the bottom
11. 🟢 notice an error toast plus a **Retry** button at the end of the list; set the network back to **Online**, click Retry, and the projects load
12. repeat steps 2–6 on **Projects → [any custom view]** and on an organization's projects list, which use the same table


magicznyleszek:
@greptileai I demand you review this PR!

duvld (approved):
LGTM

PR #7392, opened 2026-08-05
fix(formbuilder): crash with unsupported type DEV-2616

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Fixed Form Builder failing to open forms that contain question types it can't edit, and no longer dropping those questions when the form is saved.

### 📖 Description

Some valid XLSForm question types have no editor in Form Builder. Opening such a form used to break the editor, and saving it - silently deleted the question.

Now those questions show up as a card explaining that they can't be edited here and will be left alone when you save. The rest of the form works normally. Affected types include `select_one_external`, `email`, `osm`, `percentage`, `phone number`, `number of days in last month`/`six months`/`year`, `q select`, `q select1`, and the `uri:*` metadata types.

A card:

Screenshot 2026-08-05 at 16 16 51

### 💭 Notes

Changes here:

- **`model.configs.coffee`** — registered 16 types in `lookupRowType` with `supportedByUI: false`
- **`model.row.coffee`**
  - `processType` now checks the whole type value against the registry before splitting on spaces. Six of the new types contain spaces, so `"phone number"` was being read as type `phone` + choice list `number`. That's why they were never found, and it also stuck a bogus `listName` on the row.
  - `toJSON2` splits `select_one_external` back into `type` + `select_from_list_name`, since the constructor merges them. Unlike `select_one` it can legitimately have no list name, so we don't invent an empty one.
- **`view.row.coffee`**
  - `_renderRow` branches on the new `hiddenInUI` flag: `background-geopoint` stays invisible as before, everything else gets the notice card.
  - `applyLocking` bails when there's no `@$card` (both implementations). Error rows render no card, and `RowView.initialize` binds this to every survey `change` event, so it was guaranteed to throw.
- **`view.row.templates.coffee` / `view.templates.coffee`** — the notice template. It shows the question *name*, not the label, because the label may not exist.
- **`view.surveyApp.coffee`** — `_notifyIfRowsOutOfOrder` was gathering ids without `includeErrors` while `_reset` renders the DOM with it, so it compared a list that was missing rows against a DOM that had them and reported a mismatch that never happened.
- **`_card.scss`** — amber `.card--unsupported` styles.
- **`survey.tests.coffee`** — 66 tests. Each type: loads without erroring, marked unsupported, survives a save with its type intact, gets no phantom choice list.

Finding the 16 types was done with AI assistance - [pyxform's `question_type_dictionary.py`](https://github.com/XLSForm/pyxform/blob/master/pyxform/question_type_dictionary.py) has 112 types. Our formpack have [some type aliases defined](https://github.com/kobotoolbox/formpack/blob/main/src/formpack/utils/replace_aliases.py), so it filtered out few of them. Running those through `Survey.loadDict` showed which were broken. Out of them 48 were spelling variants of types we already handle, and the 16 without equivalents.

Also `phone_number` in `model.aliases.coffee` seems to be a dead code — it isn't a pyxform type at all. Left it alone for now though.

### 👀 Preview steps

2. Grab an XLSForm (`.xlsx`) and add a row to the **survey** sheet with type `select_one_external` and name `region`. Any other type from the Description works too, `email` is the quickest to type :wink:
3. Upload it (e.g. drag and drop on My Projects).
4. Open the form in Formbuilder
5. 🔴 [on main] app crashes with error
6. 🟢 [on PR] the form loads. The question is an amber card saying it can't be displayed or edited here and will be left unchanged on save.
7. Edit something unrelated (change a question label) and hit **SAVE**
8. 🟢 reopen the form in Form Builder — the amber card is still there
9. Download the form as XLS: Project → Settings → **Download** → XLS
10. 🟢 open it and check the survey sheet: the `region` row still has type `select_one_external`, right where you left it
11. 🔴 [on main] same download after a save — the row is missing entirely, having quietly ghosted your form


magicznyleszek:
@greptileai I order you to review my code

jsapp/xlform/src/model.row.coffee
```diff
@@ -84,10 +84,22 @@ module.exports = do ->
...
+        # back end uses.
         if key is 'type' and val.get('typeId') in ['select_one', 'select_multiple']
           outObj['type'] = val.get('typeId')
           outObj['select_from_list_name'] = val.get('listName')
           continue
+        # Same split for `select_one_external`, which we can't edit (its choices
+        # live in the `external_choices` sheet, which we don't load), so that it
+        # is saved back in the shape it came in. Unlike the types above it can
+        # legitimately arrive without a list name, so we don't invent an empty one.
+        else if key is 'type' and val.get('typeId') is 'select_one_external'
```
jamesrkiger: Might be worth merging this logic with the unchanged select_one and select_multiple code above:

```
if key is 'type' and val.get('typeId') in ['select_one', 'select_multiple', 'select_one_external']
  outObj['type'] = val.get('typeId')
  listName = val.get('listName')
  if listName
    outObj['select_from_list_name'] = listName
  continue
```

jamesrkiger (approved):
Left one minor comment, otherwise LGTM

jnm:
@greptileai, for the 16 added question types, would any of these be allowed into `asset.content` in the first place, if kobotoolbox/formpack has no support for them?

jnm:
@greptileai thanks. Could you explain where "unknown" (by formpack?) types are rejected at deployment time? For example, consider the error `invalid data_type: number of days in last month`

jnm:
@greptileai please confirm my understanding, which slightly differs from yours: "export" of asset JSON content (to XLSForm) is performed as part of the deployment process, which is why these `invalid data_type`s are flagged when clicking the "deploy" button.

It also looks like `select_one_external` is completely undeployable, also failing at the same XLSForm export step: `asset.content improperly formatted for XLS export: ('"label" column is not translated…`.

If you could specify which (if any) of the added types could have ever been used in a Kobo form deployed for data collection and which could not, that would be appreciated.

jnm:
@greptileai, please look at your response https://github.com/kobotoolbox/kpi/pull/7392#issuecomment-5209967944 and reconcile that with your latest comment that `number of days in last month` is deployable.

Let's be clear about `select_one_external`: it's not deployable without an `external_choices` sheet in the XLSForm, and something appears not to handle `label` properly in that sheet, making it undeployable even with that required sheet. The error I reported is not from a multi-language form. It's a simple form with `name`, `type`, and `label` in the choices sheet and `list name`, `value` and `label` in the `external_choices` sheet.

jnm:
@greptileai thanks :+1: Can you see the cause of the `label` handling problem with `select_one_external`s?

jnm:
@greptileai, the error appears without using the form builder, though. In any event, let's conclude our conversation here. Thanks :wave:

PR #7396, opened 2026-08-06
chore(mfa): remove the django-trench backend dependency DEV-852

platreth:
### 📣 Summary

Internal cleanup — removed an old, unused authentication library. No change to how two-step verification (MFA) works.

### 📖 Description

Two-step verification already runs on our current authentication stack; an older library it used to depend on was left behind with no live data still using it. This removes that old library and its leftover database tables. Nothing changes for you — enabling, using, and disabling two-step verification all behave exactly as before.

### 💭 Notes

Removes the `django-trench` dependency. Runtime MFA already runs on `django-allauth[mfa]` + `MfaMethodsWrapper`; trench survived only as a drained, read-only legacy data source, and its trench→allauth data migration is complete on all environments.

- Drop `trench` from `INSTALLED_APPS` + delete `TRENCH_AUTH`. `mfa_code_length` now reads `settings.MFA_TOTP_DIGITS` — same env var `MFA_CODE_LENGTH`, same default `6`, same int response (no API contract change, no schema regen).
- Delete the trench `MfaMethod` model, its admin, the adapter's `migrate_user`; repoint the two reporting queries to `mfa_methods_wrapper__is_active` (`UniqueConstraint(user, name)` → same per-user cardinality).
- Migration `0009` removes `MfaMethod` and drops `accounts_mfa_mfamethod` / `trench_mfamethod`. **Irreversible** (`reverse_sql=noop`) — safe only because legacy data is drained everywhere.
- Historical migrations edited to sever the trench dep + MTI base — safe (Django keys migrations by name, not content); `makemigrations accounts_mfa --check` → "No changes detected".
- `django-trench` gone from `requirements.in` (14 transitive pkgs dropped); `pyotp` re-declared as an explicit **test** dep (MFA tests import it, was transitive via trench).

### 👀 Preview steps

Behavior does not change; this confirms nothing regressed.

1. ℹ️ have an account with MFA enabled
2. 🔴 [on main] `\dt trench_mfamethod` in the kobocat/koboform DB shows the legacy table; `django-trench` is in the lockfile
3. 🟢 [on PR] after migrate, `accounts_mfa_mfamethod` + `trench_mfamethod` are gone; `django-trench` is absent from the lockfile
4. 🟢 enable/disable MFA and log in with a TOTP code — still works
5. 🟢 `/environments/` still returns `mfa_code_length: 6`


platreth:
@greptileai

kobo/apps/long_running_migrations/jobs/0013_migrate_mfa_data.py
```diff
@@ -1,57 +0,0 @@
-from allauth.mfa.adapter import get_adapter
```
Guitlle, line 1: Instead of removing it it should be left with an empty run like so, if you remove the job file it will raise errors when a new instance tries to apply the long running migrations jobs: 

```
def run():
    # Deprecated: The trench mfa library was removed, 
    # so this migration no longer needs to do anything.
    pass
```

Guitlle (changes requested):

kobo/apps/long_running_migrations/jobs/0013_migrate_mfa_data.py
platreth: Ah thanks, good point, I will fix that

Guitlle (approved):

PR #7402, opened 2026-08-06
fix(bulkProcessing): translations tab alert DEV-2623 DEV-2621 DEV-2620

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

The "this submission is already being processed by another job" warning now also appears on the Translations tab, not just on the Transcript tab.

### 💭 Notes

Changes here:

- `common/conflictingOngoingJob.ts` - dropped wrong early return
- `TabTranslations/TranslationAdd/index.tsx` + `TabTranscript/TranscriptCreate/index.tsx`
  - stopped short-circuiting on `languageCode !== null` and pass `languageCode ?? undefined` instead, then forward the flag down to `StepBegin`.
- `TabTranslations/TranslationAdd/StepBegin.tsx` + `TabTranscript/TranscriptCreate/StepBegin.tsx` - render the alert. The transcript one also disables "begin" (any job on that question rewrites the transcript), the translations one only warns to allow translating to different language.
  - also renamed the translations `StepBegin` export from `Foo`, which had been quietly living its best life in there.
- `conflictingOngoingJob.tests.ts` — three cases for the no-language path: ongoing job, completed job, job on a different question.
- `BulkApproveModal` / `BulkTranslationModal` — dropped two `size='sm'` overrides so alert text matches the rest of the alerts, and trimmed a redundant "This may take some time to complete." from the translation modal copy.

### 👀 Preview steps

1. ℹ️ have a project with several submissions containing audio recordings, and automatic transcription/translation enabled
2. go to Project → Data → Table and transcribe a few submissions (needed as the source for translations)
3. select a good handful of those rows — the more the better, the whole point is to catch the job while it's still running — and start a bulk translation to a language (e.g. Spanish)
4. while the job is processing, click the "open" button on one of the affected rows to enter the single processing view
5. 🟢 on the Transcript tab, notice the warning "This submission is already being processed by another job."
6. switch to the Translations tab
7. 🔴 [on main] notice the tab looks perfectly ordinary, with no hint that a translation is being made as you watch
8. 🟢 [on PR] notice the same warning banner as on the Transcript tab
9. 🟢 click "begin" and pick the language the job is translating to (Spanish) — notice the warning again and that both "manual" and "automatic" are disabled
10. 🟢 go back and pick a different language (e.g. French) instead — notice you can proceed normally, since that job isn't in your way

Since the above didn't work for Anji, I came up with alternative steps (that didn't work for Anji either :shrug:):

1. For one of submissions I copy pasted whole page content https://en.wikipedia.org/wiki/2010_United_Kingdom_general_election  as transcript text and saved it
2. Selected that submission row in Data Table
3. Used transcript column's "Translate selected…" dropdown option and chose "French"
4. Quickly used "Open" button
5. Arrived at "Transcript" tab with alert visible (alert visible, edit and delete buttons disabled)
6. Switched to "Translations" tab and also alert was visible (alert visible, begin button disabled)


jsapp/js/components/processing/common/conflictingOngoingJob.tests.ts
```diff
@@ -138,6 +138,59 @@ describe('isConflictingOngoingJobForSubmission', () => {
...
+      ],
+      actionType: 'translation',
+      fieldXpath,
+      submissionUuid,
+    })
+
+    chai.expect(result).to.equal(false)
+  })
+
+  it('returns false for translation with no selected language when the job targets another question', () => {
```
duvld, line 177: All these tests are cases with "no selected language". Should we not have a test case for
1) When the selected language already has a translation
2) When the selected language doesn't have a translation?

jsapp/js/components/processing/SingleProcessingContent/TabTranscript/TranscriptCreate/index.tsx
```diff
@@ -51,15 +51,15 @@ export default function TranscriptCreate({
...
-  const hasConflictingOngoingJob =
-    languageCode !== null &&
-    isConflictingOngoingJobForSubmission({
-      activeBulkActions,
-      actionType: 'transcript',
-      fieldXpath: questionXpath,
-      submissionUuid: getSubmissionRootUuid(submission),
-      selectedLanguage: languageCode,
-    })
+  // Any job on this question rewrites the transcript, so the conflict doesn't
```
duvld: I don't think this comment is equivlent to the logic here. `selectedLanguage: languageCode ?? undefined` implies "whether a language is selected or not", and not "doesn't depend on *which* language is selected`.

The latter should already be the case before the change?

jsapp/js/components/processing/SingleProcessingContent/TabTranslations/TranslationAdd/StepBegin.tsx
```diff
@@ -3,26 +3,36 @@ import React from 'react'
...
 }
 
-export default function Foo({ asset, onNext }: Props) {
+export default function StepBegin({ asset, hasConflictingOngoingJob, onNext }: Props) {
   return (
     <div className={cx(bodyStyles.root, bodyStyles.stepBegin)}>
       <header className={bodyStyles.header}>{t('This transcript does not have any translations yet')}</header>
 
+      {/*
+        We only warn here, without blocking: a job translating to one language
```
duvld: This looks like to me if the user was already on this page and in another tabs creates a conflicting bulk translation job, then the language step blocking access to this page would not stop a conflict. Is disabling the button here not worth maybe a little redundancy?

duvld (commented):
Some questions, I cannot reproduce any of the new changes.

I did the following to try:
1. Have ~24 audio submissions
2. Bulk transcribe all submissions
3. Open activity monitor to enure they are still processing
4. Click "Open" to single processing view
5. 🔴 I don't see any banner in the transcript tab
6. Select a few transcriptions (I made them very long about 80,000 characters) and run a bulk translation job
7. Open activity monitor to enure they are still processing
4. Click "Open" to single processing view
5. 🔴 I don't see any banner in the translation tab

I'm also allowed to select submissions that have English translations, and run a bulk job with English again. Not sure if that's intended though

jsapp/js/components/processing/common/conflictingOngoingJob.tests.ts
magicznyleszek: Added a test

jsapp/js/components/processing/SingleProcessingContent/TabTranscript/TranscriptCreate/index.tsx
magicznyleszek: You're right - dropped the argument and rewrote the comment to say what actually changed

jsapp/js/components/processing/SingleProcessingContent/TabTranslations/TranslationAdd/StepBegin.tsx
magicznyleszek: Right, disabled it now :)

jamesrkiger (approved):
LGTM

PR #7403, opened 2026-08-06
fix(bulkProcessing): hide transcript language from bulk translation modal DEV-2622

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

The language a transcript is already in is no longer offered when creating a translation, so you can't accidentally translate English into English and end up with a blank column you can't get rid of.

### 💭 Notes

Translating a transcript into its own language created an empty column in the data table that couldn't be removed, and nothing appeared under Translations. Fix is to keep that language out of the dropdown in the first place.

Changes here:

- `jsapp/js/components/processing/common/utils.ts`
  - `getTranslationSourceLanguages()` — new. Returns the language codes a new translation must not target for a given question. Picks the source transcript the way the back end does in `RequiresTranscriptionMixin.attach_action_dependency`: newest `_dateAccepted`, skipping versions never accepted and ones whose text is gone. Note this is deliberately *not* `getLatestTranscriptVersionItem()`, which sorts by creation date — when versions are accepted out of order the two pick different versions, and following the wrong one would hide the wrong language and let the bug straight back in. Both helpers stay, doc comments warn against swapping them.
  - `getBlockedTargetLanguages()` — new, and the reason this is a list rather than one code. The back end translates from `locale` when it's set and never checks it against `language`, so `{language: 'es', locale: 'fr-CA'}` is a storable transcript that gets translated from French. Columns are also keyed by base language (back end does `language.split('-')[0]`), so `fr-CA` and `fr` collide in one column. Both codes get blocked.
- `jsapp/js/components/processing/common/utils.tests.ts` — 10 tests over the two new helpers: acceptance-vs-creation order, unaccepted versions, deleted text, locale/language disagreement. Mutation-checked by breaking each rule in turn and confirming the matching test went red.
- `jsapp/js/components/submissions/.../BulkTranslationModal.tsx` — passes `hiddenLanguages` to `<LanguageSelector>`, which it never did. Blocks the column's own language plus every transcript language across the selected rows: eligibility only asks for a transcript with *some* text, so a row transcribed in another language would be translated from that one instead. Reads `regionCode` too, since that's where the flattened transcript keeps its locale.
- `jsapp/js/components/processing/.../TabTranslations/index.tsx` — same gap in single-submission processing. It was hiding existing translation languages but not the transcript's, so English → English was one click away here as well.
- `jsapp/js/components/processing/.../TranslationAdd/index.tsx` — `languagesExisting` → `languagesUnavailable`, since it now carries more than the existing translations. Single caller, so the rename stays put.

The dropdown itself only ever lists base language codes — regions live in a separate model and surface through `RegionSelector`, which translations don't use. Blocking the full locale is cheap insurance in case regional targets ever show up there.

### 👀 Preview steps

2. go to Project → Data → Table and transcribe that audio question in English (Data → open submission → Transcript, or select rows and use the bulk transcribe action)
3. accept the transcript
4. in the table, select the row(s), open the transcript column menu and choose the translate action
5. 🔴 [on main] English is in the language dropdown; pick it, start the translation, and a blank English translation column shows up in the table with no way to delete it and nothing under Translations
6. 🟢 [on PR] English isn't in the dropdown at all — the transcript language quietly excused itself
7. now open a single submission → Translations → add a translation
8. 🔴 [on main] English is offered here too
9. 🟢 [on PR] English is missing here as well, alongside any language already translated into
10. 🟢 pick any other language and confirm translating still works normally

### DIFF
--- a/jsapp/js/components/processing/SingleProcessingContent/TabTranslations/TranslationAdd/index.tsx
+++ b/jsapp/js/components/processing/SingleProcessingContent/TabTranslations/TranslationAdd/index.tsx
@@ -28,7 +28,8 @@ interface Props {
   submission: DataResponse
   supplement: DataSupplementResponse
   activeBulkActions: BulkActionResponse[]
-  languagesExisting: LanguageCode[]
+  /** Languages that can't be picked: already translated into, or the transcript's own language. */
+  languagesUnavailable: LanguageCode[]
   initialStep?: CreateSteps.Begin | CreateSteps.Language
   translationVersions: Array<SupplementalDataVersionItemManual | SupplementalDataVersionItemAutomatic>
   onCreate: (languageCode: LanguageCode, context: 'automated' | 'manual') => void
@@ -43,7 +44,7 @@ export default function TranslationAdd({
   submission,
   supplement,
   activeBulkActions,
-  languagesExisting,
+  languagesUnavailable,
   initialStep,
   translationVersions,
   onCreate,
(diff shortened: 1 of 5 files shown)


jamesrkiger (approved):
lgtm

PR #7404, opened 2026-08-06
feat(organizations): add search parameter to org members API DEV-2458

Guitlle:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Adding the q parameter query parsing solution to the organizations members API so that frontend can have search filters for the members.

### 📖 Description
Inside OrganizationMemberViewSet.get_queryset(), when the q parameter is present, we now run parse() twice:
* For existing members: Parsed q against `OrganizationUser fields (user__username__icontains, user__email__icontains, user__first_name__icontains, user__last_name__icontains)`
* For pending invites: Parsed q against `OrganizationInvitation fields (invitee__username__icontains, invitee__email__icontains, invitee__first_name__icontains, invitee__last_name__icontains, invitee_identifier__icontains)`
We explicitly bypassed the global field denylist for this view by specifying custom allowed_lookup_fields (allowing kobo_auth.user fields like first_name and last_name and others mentioned in the ticket). I also include the parameter definition @extend_schema_view to get the correct documentation for the endpoint auto generated.

Since the members API uses two different models (`user` for active members and `invitee` for pending invites), the lookups work like so:
* To search ONLY active members: ?q=user__email:luis@example.com
* To search ONLY pending invites: ?q=invitee__email:luis@example.com
* To search across both groups simultaneously use the generic query (e.g., `q=luis@example.com` which will match any of the allowed fields using `icontains`) without specifying prefixes. 

### 👀 Preview steps
1. ℹ️ have an account and a organization with multiple members 
2. Using the organization uid do:
```
curl -X GET "http://kf.kobo.local/api/v2/organizations/{uid}/members/?q={search}" \
     -H "Authorization: Token {API_TOKEN}"
```
3. 🟢 [on PR] the search feature should work correctly, i.e., it should filter according to the rules mentioned in the description section.


kobo/apps/organizations/views.py
```diff
@@ -609,8 +624,88 @@ def get_queryset(self):
...
+                        model=queryset.model,
+                        allowed_lookup_fields=user_allowed_fields,
+                        user=self.request.user,
+                    )
+
+                    queryset = queryset.filter(q_obj_user)
+                except QueryParserNotSupportedFieldLookup as e:
+                    user_unsupported_exception = e
+                    queryset = queryset.none()
+                except (ParseError, FieldError, ValueError):
```
platreth: These silently turn invalid queries into an empty 200! Maybe return a 400 for ParseError?

kobo/apps/organizations/docs/api/v2/members/list.md
```diff
@@ -2,6 +2,19 @@
...
+### Searching
+
+Search can be made with the `q` parameter. It will implicitly match against the username, email, first name, last name, and profile name (`extra_details__data__name`) of the members or pending invitations.
+
+```shell
+curl -X GET https://kf.kobotoolbox.org/api/v2/organizations/{uid_organization}/members/?q=luis
+```
+
+> [!WARNING]
+> **Explicit Field Searching:** Because this endpoint combines active members (`OrganizationUser`) and pending invitations (`OrganizationInvitation`), you cannot use an explicit boolean `OR` across different model fields in the same query (e.g., `q=user__email:luis@example.com OR invitee__email:luis@example.com`). Doing so will cause the database query to fail for the invalid fields and return an empty list.
```
platreth: It says that mixing fields from both models will return in an empty list but that's not true, the client gets a 400

platreth (changes requested):

kobo/apps/organizations/docs/api/v2/members/list.md
Guitlle: thank y ou, yeah, it's better returning the 400

magicznyleszek (approved):
Frontend part LGTM

platreth (approved):

PR #7411, opened 2026-08-10
feat(organizations): display SSO column in Members Table DEV-2456

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

The Members table now has an SSO column showing which members sign in through single sign-on.

### 💭 Notes

Changes here:

- `MembersRoute.tsx`
  - New `SSO` column next to `2FA`, driven by `user__has_sso_enabled` from the members endpoint (added in DEV-2455).
  - Column header has a tooltip and is always present. When no SSO provider is configured, the header goes grey and the cells stay empty — the column is there, it just has nothing to say yet.
  - Pulled the check/minus badge out of the 2FA formatter into `renderStatusBadge` so both columns can't drift apart.
  - Component is now wrapped in `observer` so the column wakes up when `envStore` finishes loading.
- `sso.utils.ts` (new)
  - `getSsoProviders`, `isSsoAvailable`, `getConnectedSsoAccounts`. Takes `EnvStoreData` as an argument rather than reaching for the store, matching `organization.utils.tsx`.
- `ssoSection.component.tsx`
  - Uses those helpers instead of its own inline copies.

### 👀 Preview steps

1. ℹ️ have an account and MMO with some members
2. go to Account Settings → Members
3. 🔴 [on main] notice there's only a `2FA` column
4. 🟢 [on PR] notice a new `SSO` column right after `2FA`
5. on a server **without** SSO configured (e.g. a plain local instance): 🟢 the `SSO` header is greyed out, all its cells are empty, and hovering the header shows "Activate SSO add-on to enable"
6. on a server **with** an SSO provider configured: 🟢 hovering the header shows "SSO status", and each member gets a blue check badge if they've connected an SSO account or a grey minus badge if they haven't — same badges as the 2FA column
7. 🟢 check a pending invitee row: both `2FA` and `SSO` cells are empty


pauloamorimbr (approved):
LGTM!

PR #7412, opened 2026-08-11
feat(bulkProcessing): improve language blocking DEV-2661

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Fixed bulk translation blocking languages it shouldn't: when a project has transcripts in more than one language for the same question, all of those languages became unavailable as translation targets.

### 📖 Description

If some submissions were transcribed in English and others in Spanish for the same question, both English and Spanish disappeared from the language list when starting a bulk translation, leaving no way to translate one into the other. Now only the transcript you're actually translating from counts.

### 💭 Notes

Changes here:

- **`bulkProcessingUtils.ts`**
  - `hasTranslatableTranscript` now also checks the transcript language against the column language.
  - `getBlockedBulkTranslationLanguages` — new, moved out of the modal so it can be unit tested. Always blocks the column's own language, and beyond that only considers rows it can genuinely translate from.
- **`BulkTranslationModal.tsx`** — the ~20-line `useMemo` is now a call to the helper above.
- **`dataInterface.ts`** — exported `TransxObject` so tests can type transcript fixtures.
- **tests** — the translation `evaluateNoSource` cases were passing a bare question name as `fieldXpath`; translation always runs off a transcript column, so those fixtures never matched reality. Fixed, plus coverage for the mixed-language case and for the locale/language mismatch.

### 👀 Preview steps

1. ℹ️ Have a project with an audio question
2. Submit at least 3 responses to it
3. Go to Project → Data → Table
4. Create manual transcriptions for submissions - two English, and the third one in Spanish
5. 🟢 Notice you now have two transcript columns: one English, one Spanish
6. Select all three rows, open the column header menu on the **English** transcript column header → "Translate selected transcriptions"
7. 🔴 [on main] neither English nor Spanish appears in the language list, so there's no way forward
8. 🟢 [on PR] Spanish is available, and the modal warns that 1 submission is missing transcription and will be ignored
9. 🟢 Notice English is still unavailable — translating the English transcript into English would leave an empty column behind
10. Pick Spanish and create the translations
11. 🟢 Notice only the two English-transcript rows got translated, and the Spanish row was left alone
12. now open the column header menu on the **Spanish** transcript column header with all three rows still selected
13. 🟢 notice English is available here and Spanish is not — mirror image of step 8

### DIFF
--- a/jsapp/js/components/submissions/BulkProcessingModals/BulkTranslationModal/BulkTranslationModal.tsx
+++ b/jsapp/js/components/submissions/BulkProcessingModals/BulkTranslationModal/BulkTranslationModal.tsx
@@ -18,9 +18,10 @@ import {
 import ButtonNew from '#/components/common/ButtonNew'
 import LanguageSelector from '#/components/languages/LanguageSelector'
 import type { LanguageCode } from '#/components/languages/languagesStore'
-import { getBlockedTargetLanguages, getSuggestedLanguages } from '#/components/processing/common/utils'
+import { getSuggestedLanguages } from '#/components/processing/common/utils'
 import { getSupplementalPathParts } from '#/components/processing/processingUtils'
 import { BulkProcessingWarningModal } from '#/components/submissions/BulkProcessingModals/BulkProcessingWarningModal'
+import { getBlockedBulkTranslationLanguages } from '#/components/submissions/bulkProcessingUtils'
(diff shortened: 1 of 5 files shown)


magicznyleszek:
@greptileai I would be glad if you'd review this

jamesrkiger (approved):
lgtm

PR #7413, opened 2026-08-11
fix(xlsImport): match project languages to XLSForm languages DEV-2657

platreth:
### 📣 Summary

Re-uploading an XLSForm now sets the project languages to exactly the languages of the uploaded file, instead of keeping languages that were removed from it.

### 📖 Description

Uploading a form with three languages and re-uploading it with two used to leave the third language defined with empty strings, and removing all languages created an unnamed language. Project languages now always match the file: a language stays only if at least one `survey` or `choices` column still uses it, and a file with plain `label` columns brings the project back to "no languages defined".

### 💭 Notes

- Cause: `Asset.save()` restores the DB translations when the incoming content declares none (#7020, for the Form Builder). An imported spreadsheet dict has no `translations` key yet, so the stored list was injected before `formpack.expand_content()` ran — which only ever appends to that list.
- Fix: the import path expands the content (`standardize_content_in_place()`) before assigning it, so the language list comes from the file. Covers both import destinations (base64 and URL). Form Builder path untouched; a direct API `PATCH content` still hits the old fallback — separate ticket territory.
- Related: DEV-2656.

### 👀 Preview steps

2. upload an XLSForm with three languages (`label::English (en)`, `label::French (fr)`, `label::Spanish (es)`)
3. remove every `::Spanish (es)` column and re-upload to the same project
4. 🔴 [on main] Spanish still listed as a project language, with empty strings
5. 🟢 [on PR] project lists English and French only
6. replace the remaining language columns with a plain `label` column and re-upload
7. 🔴 [on main] languages still there, plus an unnamed language
8. 🟢 [on PR] project back to "no languages defined"

### DIFF
--- a/kpi/models/import_export_task.py
+++ b/kpi/models/import_export_task.py
@@ -7,11 +7,12 @@ import tempfile
 from collections import defaultdict
 from io import BytesIO
 from os.path import split, splitext
-from typing import Any, Dict, Generator, List, Optional, Tuple
+from typing import Dict, Generator, List, Optional, Tuple
 from zoneinfo import ZoneInfo
 
 import constance
 import dateutil.parser
+import formpack
 import requests
 from django.conf import settings
 from django.contrib.postgres.indexes import BTreeIndex, HashIndex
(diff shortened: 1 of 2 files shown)


platreth:
@greptileai

PR #7240, opened 2026-07-13
fix(import): forbid duplicate names on import DEV-2167

rgraber:
### 🗒️ Checklist

1. [x] run linter locally
2. [x] update developer docs (API, README, inline, etc.), if any
3. [x] for user-facing doc changes create a Zulip thread at `#Support Docs Updates`, if any
4. [x] draft PR with a title `<type>(<scope>)<!>: <title> DEV-1234`
5. [x] assign yourself, tag PR: at least `Front end` and/or `Back end` or `workflow`
6. [x] fill in the template below and delete template comments
7. [x] review thyself: read the diff and repro the preview as written
8. [x] open PR & confirm that CI passes & request reviewers, if needed
9. [ ] act on any greptile review below a 5/5 score or leave comment explaining why you won't
10. [ ] delete this checklist section from the final squash commit before merging

### 📣 Summary
Do not allow users to import XLS files with duplicate names.

### 📖 Description
Previously, users were allowed to import XLS files with duplicate names that were then renamed on deploy. Update the import process to no longer allow this.

### 👀 Preview steps
Make sure to restart the worker after switching branches

1. ℹ️ have an account
2. Upload https://docs.google.com/spreadsheets/d/e/2PACX-1vQAXAkJ9s4PNFyZhI8cmzbGCdApK2u5YcDjArt4gXaxbjKa0oiNHp0-hLW7Jaz-OA/pub?output=xlsx to create a new asset
3. 🔴 [on main] Upload succeeds
4. [on main] Deploy
5. [on main] Add a submission with two audio responses
6. [on main] Navigate to data table and click Open on the audio file in the second group
7. 🔴 [on main] path not found / recognized
8. 🟢 [on PR] upload is forbidden
9. 🟢 [on PR] regression tests: ensure an existing form with duplicates can be exported
10. [on PR] redeploy the asset uploaded when on the main branch
11. 🟢 [on PR] Duplicate name error

PR #7414, opened 2026-08-11
fix(submissions): allow editing submissions with legacy duplicate field names DEV-2597

rgraber:
### 📣 Summary
Fixes a bug blocking users from editing submissions that had been made to older forms with duplicate field names (before they were forbidden).

### 📖 Description
Formpack does not allow duplicate field names. In the past, if an XLS was uploaded with duplicate field names, we silently renamed them for formpack, keeping the old duplicate names around in KPI. We changed this behavior in a previous fix to forbid uploading forms with duplicates, but there were still legacy forms that had been previously uploaded and deployed. Once we started raising an error with duplicates, trying to edit a submission for one of these legacy forms would trigger the error. This PR fixes that.

### Notes
In order to ensure edited responses can still be exported, this PR requires reinstating the autorenaming process under certain circumstances to make sure that edited responses are associated with the same questions they were associated with before we forbade duplicate fields. 

### 👀 Preview steps

1. ℹ️ have an account
2. Check out commit `e966f2782472fc22bee1076062a0920a09925303` (before the previous fix)
3. Create a project from https://docs.google.com/spreadsheets/d/e/2PACX-1vQAXAkJ9s4PNFyZhI8cmzbGCdApK2u5YcDjArt4gXaxbjKa0oiNHp0-hLW7Jaz-OA/pub?output=xlsx
4. Add a submission
5. Go back to release branch
6. Try to edit the submission
7. 🔴 [on release] error
8. 🟢 [on PR] you are able to edit the submission in enketo
9. 🟢 [on PR]  (regression) ensure the edited field appears when you export data

### DIFF
--- a/kpi/deployment_backends/openrosa_backend.py
+++ b/kpi/deployment_backends/openrosa_backend.py
@@ -5,12 +5,12 @@ from contextlib import contextmanager
 from datetime import date, datetime, timedelta
 from typing import Generator, Literal, Optional, Union
 from urllib.parse import urlparse
+from xml.etree.ElementTree import ParseError
 from zoneinfo import ZoneInfo
 
 import redis.exceptions
 import requests
 from constance import config
-from xml.etree.ElementTree import ParseError
 from django.conf import settings
 from django.core.cache.backends.base import InvalidCacheBackendError
 from django.core.files import File
@@ -21,7 +21,6 @@ from django.db.models.query import QuerySet
 from django.utils import timezone
 from django.utils.translation import gettext_lazy as t
 from django_redis import get_redis_connection
-from pyxform.builder import create_survey_from_xls
 from rest_framework import exceptions, status
 
 from kobo.apps.data_collectors.utils import (
@@ -75,11 +74,13 @@ from kpi.fields import KpiUidField
 from kpi.interfaces.sync_backend_media import SyncBackendMediaInterface
 from kpi.models.asset_file import AssetFile
 from kpi.models.paired_data import PairedData
+from kpi.utils.autoname import HandleDuplicatesOptions
 from kpi.utils.files import ExtendedContentFile
 from kpi.utils.log import logging
 from kpi.utils.mongo_helper import MongoHelper
 from kpi.utils.object_permission import get_anonymous_user, get_database_user
 from kpi.utils.xml import fromstring_preserve_root_xmlns, xml_tostring
+from pyxform.builder import create_survey_from_xls
 from ..exceptions import AttachmentUidMismatchException, BadFormatException
 from .base_backend import BaseDeploymentBackend
 from .kc_access.utils import kc_transaction_atomic
@@ -129,6 +130,7 @@ class OpenRosaDeploymentBackend(BaseDeploymentBackend):
                     'form_title': self.asset.name,
                 }
             },
+            handle_duplicates=HandleDuplicatesOptions.RAISE,
         )
         xlsx_file = ContentFile(xlsx_io.read(), name=f'{self.asset.uid}.xlsx')
 
--- a/kpi/deployment_backends/openrosa_backend.py
+++ b/kpi/deployment_backends/openrosa_backend.py
@@ -75,7 +75,6 @@ from kpi.fields import KpiUidField
 from kpi.interfaces.sync_backend_media import SyncBackendMediaInterface
 from kpi.models.asset_file import AssetFile
 from kpi.models.paired_data import PairedData
-from kpi.utils.autoname import HandleDuplicatesOptions
 from kpi.utils.files import ExtendedContentFile
 from kpi.utils.log import logging
 from kpi.utils.mongo_helper import MongoHelper
@@ -130,7 +129,6 @@ class OpenRosaDeploymentBackend(BaseDeploymentBackend):
                     'form_title': self.asset.name,
                 }
             },
-            handle_duplicates=HandleDuplicatesOptions.RAISE,
         )
         xlsx_file = ContentFile(xlsx_io.read(), name=f'{self.asset.uid}.xlsx')
 
--- a/kpi/deployment_backends/openrosa_backend.py
+++ b/kpi/deployment_backends/openrosa_backend.py
@@ -21,6 +21,7 @@ from django.db.models.query import QuerySet
 from django.utils import timezone
 from django.utils.translation import gettext_lazy as t
 from django_redis import get_redis_connection
+from pyxform.builder import create_survey_from_xls
 from rest_framework import exceptions, status
 
 from kobo.apps.data_collectors.utils import (
@@ -80,7 +81,6 @@ from kpi.utils.log import logging
 from kpi.utils.mongo_helper import MongoHelper
 from kpi.utils.object_permission import get_anonymous_user, get_database_user
 from kpi.utils.xml import fromstring_preserve_root_xmlns, xml_tostring
-from pyxform.builder import create_survey_from_xls
 from ..exceptions import AttachmentUidMismatchException, BadFormatException
 from .base_backend import BaseDeploymentBackend
 from .kc_access.utils import kc_transaction_atomic
--- a/kpi/mixins/formpack_xlsform_utils.py
+++ b/kpi/mixins/formpack_xlsform_utils.py
@@ -57,8 +57,10 @@ class FormpackXLSFormUtilsMixin:
         else:
             return False
 
-    def _autoname(self, content, raise_on_error=True):
-        autoname_fields_in_place(content, '$autoname', raise_on_error)
+    def _autoname(self, content, raise_on_error=True, rename_bad_fields=False):
+        autoname_fields_in_place(
+            content, '$autoname', raise_on_error, rename_bad_fields
+        )
         autovalue_choices_in_place(content, '$autovalue')
 
     def _insert_xpath(self, content):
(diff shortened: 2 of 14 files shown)


kpi/tests/api/v2/test_api_submissions.py
```diff
@@ -1665,6 +1665,65 @@ def test_get_edit_link_submission_as_owner(self):
...
+            self.asset.deploy()
+        uuid_ = uuid.uuid4()
+        submission = {
+            'q1': ''.join(random.choice(string.ascii_letters) for letter in range(10)),
+            'q1_001': ''.join(
+                random.choice(string.ascii_letters) for letter in range(10)
+            ),
+            'meta/instanceID': f'uuid:{uuid_}',
+            '_uuid': str(uuid_),
+            '_submitted_by': 'someuser',
```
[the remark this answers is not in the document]
rgraber: Why did darker not catch this?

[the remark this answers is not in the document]
noliveleger: @greptile re-review.

kpi/models/asset_snapshot.py
```diff
@@ -2,8 +2,8 @@
 
 from django.conf import settings
 from django.db import models
-from formpack import FormPack
 
+from formpack import FormPack
```
noliveleger, line 5: I know that darker (flake8?) has a hard time with this import but `formpack` belongs to the third party imports. The previous line was the correct one.

kpi/mixins/formpack_xlsform_utils.py
```diff
@@ -57,8 +57,10 @@ def _standardize(self, content):
         else:
             return False
 
-    def _autoname(self, content, raise_on_error=True):
-        autoname_fields_in_place(content, '$autoname', raise_on_error)
+    def _autoname(self, content, raise_on_error=True, rename_bad_fields=False):
```
noliveleger: `_autoname()` exposes `rename_bad_fields` whereas `autoname_fields_in_place()` exposes `rename_invalid_fields`. What about using the same name for both?


EDIT: this comment is not valid anymore if we use my "enum" idea.

kpi/utils/autoname.py
```diff
@@ -83,7 +83,9 @@ def autoname_fields(surv_content, in_place=False):
         return _content_copy.get('survey')
 
 
-def autoname_fields_in_place(surv_content, destination_key, raise_on_error=True):
+def autoname_fields_in_place(
+    surv_content, destination_key, raise_on_error=True, rename_invalid_fields=False
```
noliveleger: The two booleans are hard to read: if `rename_invalid_fields` is `True`, then `raise_on_error` is unreachable, since the rename loop resolves every duplicate before `_assign_row_to_name` is called. So `raise_on_duplicate_error=False, rename_invalid_fields=True` in the viewset looks like two independent knobs when it is really a single mode, and the fourth combination (`True/True`) does not exist.

What about a single enum, in the spirit of `on_delete=CASCADE`?

```python
class DuplicateNameHandling(Enum):

    RAISE = 'raise'  # reject the content: imports, deployments
    IGNORE = 'ignore'  # keep going, the last row wins
    RENAME = 'rename'  # legacy autorenaming of duplicate and invalid names: `q1`, `q1_001`, ...
```

```diff
-def autoname_fields_in_place(surv_content, destination_key, raise_on_error=True, rename_invalid_fields=False):
+def autoname_fields_in_place(surv_content, destination_key, duplicate_name_handling=DuplicateNameHandling.RAISE):
...
     def _assign_row_to_name(row, suggested_name):
-        if raise_on_error and suggested_name in other_names:
+        if duplicate_name_handling is DuplicateNameHandling.RAISE and suggested_name in other_names:
...
-    if not rename_invalid_fields:
+    if duplicate_name_handling is not DuplicateNameHandling.RENAME:
```

`_autoname()` keeps raising by default:

```python
def _autoname(self, content, duplicate_name_handling=DuplicateNameHandling.RAISE):
    autoname_fields_in_place(content, '$autoname', duplicate_name_handling)
    ...
```

```diff
 class AssetSnapshot(
         self._strip_empty_rows(_source)
-        raise_on_duplicate = kwargs.pop('raise_on_duplicate_error', True)
-        rename = kwargs.pop('rename_invalid_fields', False)
-        self._autoname(
-            _source, raise_on_error=raise_on_duplicate, rename_bad_fields=rename
-        )
+        self._autoname(_source, DuplicateNameHandling.RENAME)
```

Snapshots are ephermal XML for previewing and editing existing content, so there is no case where one should reject a legacy form, right? 

In [`table_view()`](https://github.com/kobotoolbox/kpi/blob/b3e09c0bccaf226ef6453cab26565a232a12b2f7/kpi/views/v2/asset.py#L1067):

```python
md_table = ss_structure_to_mdtable(
    sa.ordered_xlsform_content(duplicate_name_handling=DuplicateNameHandling.IGNORE)
)
```

And in `kpi/mixins/xls_exportable.py`, replacing `raise_on_autoname_error`, keeping `RAISE` as the default so deployments still reject duplicates (`openrosa_backend.py` calls `to_xlsx_io()` without arguments, and `test_to_xlsx_io_with_duplicate_node_names` covers it):

```python
def ordered_xlsform_content(
    self,
    kobo_specific_types=False,
    append=None,
    duplicate_name_handling=DuplicateNameHandling.RAISE,
):
    ...
    self._autoname(content, duplicate_name_handling=duplicate_name_handling)
```

Three remaining call sites need to follow, the first one being a `TypeError` on every `asset.save()` otherwise:

- `kpi/models/asset.py:502`: `self._autoname(self.content, raise_on_error=False)` becomes `DuplicateNameHandling.IGNORE`
- `kpi/renderers.py:423`: `raise_on_autoname_error=False` is forwarded through `to_xlsx_io(**kwargs)`, so it becomes `duplicate_name_handling=DuplicateNameHandling.IGNORE`
- `kpi/tests/test_assets.py:545`: same change

I think this makes the flow more obvious than the two booleans.

noliveleger (changes requested):
It does work as expected but some nitpicks.

Also, (IMO) my idea deserves a discussion

noliveleger (changes requested):
I cannot access the line from this PR but one call has to be changed. 
See https://github.com/kobotoolbox/kpi/blob/b3e09c0bccaf226ef6453cab26565a232a12b2f7/kpi/views/v2/asset.py#L1070 

```
        md_table = ss_structure_to_mdtable(
            sa.ordered_xlsform_content(raise_on_autoname_error=False)
        )
```

`raise_on_autoname_error`, does not exist anymore (and the CI did not catch it because no test on that).

kpi/deployment_backends/openrosa_backend.py
```diff
@@ -129,6 +130,7 @@ def connect(self, active=False):
                     'form_title': self.asset.name,
                 }
             },
+            handle_duplicates=HandleDuplicatesOptions.RAISE,
```
noliveleger: nits: Why passing this one (which is the default) although the two other calls in the same files `to_xlsx_io` are not (and rely on the default value)?

kpi/tests/api/v2/test_api_submissions.py
```diff
@@ -1665,6 +1665,65 @@ def test_get_edit_link_submission_as_owner(self):
...
+        }
+
+        self.asset.deployment.mock_submissions([submission])
+        self._get_edit_link()
```
noliveleger, line 1726: I guess we should assert that `q1_001` is in the `snapsnot.xml` and `snapshot.detail` contains success ( I force an error in `generate_xml_from_source` when saving the XML but the test did pass. 

What do you think?

kpi/models/asset_snapshot.py
```diff
@@ -139,7 +140,10 @@ def save(self, *args, **kwargs):
         self._standardize(_source)
         self._make_default_translation_first(_source)
         self._strip_empty_rows(_source)
-        self._autoname(_source)
+        handle_duplicates = kwargs.pop(
+            'handle_duplicates', HandleDuplicatesOptions.IGNORE
+        )
+        self._autoname(_source, handle_duplicates=handle_duplicates)
```
noliveleger: I think this should could be hard-coded to `self._autoname(_source, HandleDuplicatesOptions.RENAME)`

Snapshots are ephemeral XML for previewing and editing existing content,  so duplicate names are renamed the way they were when the form was deployed. 

To be concrete about why `RENAME` and not `IGNORE` here: on a form deployed before #7240, with two `word_question` in two different groups, the live XForm binds `/group_second/word_question_001` and the submitted data uses that same key.

`IGNORE` regenerates it as `/group_second/word_question`. The preview renders fine, no error, but it is no longer the form the data was collected with. And when the duplicates sit in the same section, whether at the root or inside the same group, `IGNORE` produces no XML at all (`status: failure`, `There are more than one survey elements named 'word_question'`), so the preview fails with an 400.

`RENAME` reproduces the deployed XPaths exactly, in both cases.

Bonus: Asset.snapshot() would not need the `handle_duplicates` argument anymore.

noliveleger (changes requested):

kpi/utils/autoname.py
```diff
@@ -99,7 +113,26 @@ def _assign_row_to_name(row, suggested_name):
...
-        _assign_row_to_name(row, row['name'])
+        _name = row['name']
+        if handle_duplicates != HandleDuplicatesOptions.RENAME:
+            _assign_row_to_name(row, _name)
+            continue
+
+        _attempt_count = 0
+        while not is_valid_node_name(_name) or _name in other_names:
+            # save original name
+            row['$given_name'] = _name
```
platreth: This gets overwritten no? Or is it on purpose?

noliveleger (approved):
LGTM

PR #7415, opened 2026-08-11
feat(bulkProcessing): accept both submission `uuid` and `root_uuid` in bulk endpoints DEV-2663

rajpatel24:
### 📣 Summary
The frontend now sends `root_uuid` when starting and accepting bulk transcription/translation jobs, but falls back to `uuid` for submissions that don't have a `root_uuid` available yet. 

The bulk endpoints only matched on `root_uuid`, so this makes them accept either identifier and resolve it to the `root_uuid` the rest of the pipeline is keyed on, consistent with the existing single-processing endpoint.

### 📖 Description
A submission has two identifiers: `root_uuid`, which remains stable throughout its lifetime, and `uuid`, which changes whenever the submission is edited.

The frontend normally sends root_uuid, but it may not be available yet because the long-running migration that syncs root_uuid to MongoDB is still in progress. In those cases, the frontend only has uuid.

The bulk pipeline remains keyed on root_uuid throughout. This change allows the bulk endpoints to accept either identifier and canonicalise it to root_uuid before processing.

### 👀 Preview steps

1. ℹ️ have an account and a project with audio submissions.
2. trigger a bulk transcription of those audio files. (Use `?ff_bulkProcessingEnabled=true` to enable bulk actions in the UI.)
3. once the transcription is processed, edit one of the submissions and re-trigger bulk transcription.
4. 🔴 [on main] notice that this fails with an "Unknown submission" error.
5. 🟢 [on PR] notice that this now works correctly and the FE sends `root_uuid` in the payload.

### DIFF
--- a/kobo/apps/subsequences/serializers.py
+++ b/kobo/apps/subsequences/serializers.py
@@ -43,14 +43,22 @@ def get_submission_root_uuid_map(asset, submission_uuids: list[str]) -> dict[str
         return {}
 
     requested = [remove_uuid_prefix(uuid) for uuid in submission_uuids]
-    root_uuid_by_identifier = {}
-    for uuid, root_uuid in Instance.objects.filter(
-        Q(root_uuid__in=requested) | Q(uuid__in=requested),
-        xform=asset.deployment.xform,
-    ).values_list('uuid', 'root_uuid'):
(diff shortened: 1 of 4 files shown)


rajpatel24:
No longer required.

PR #7420, opened 2026-08-12
fix(bulkProcessing): use rootUuid in bulk processing code DEV-2660

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Fixes bulk transcription, translation and approval failing with an "Unknown submission UUIDs" error when the selection included submissions that had been edited.

### 💭 Notes

- Fixed at the evaluators, not at the POST site. The uuids in the request body come out of the alert code, so mapping them at the POST would leave the evaluators still comparing `_uuid`s. Root uuid is now the only id in there, and the two modals no longer create the list themselves.
- Prefix-less, matching Single Processing code and how the backend stores it. Custom namespaces like `kobotoolbox.org:…` are preserved; only the default `uuid:` prefix is stripped.
- Consolidated ~20 duplicate `meta/rootUuid || _uuid` expressions plus a fallback-less local helper onto the shared one. :bug: One of them, in `SelectSubmission.tsx`, would have thrown on a submission with no `meta/rootUuid`.
- Why no test caught this: `assetData.factory.ts` hardcoded `meta/rootUuid` independently of `_uuid`, so overriding `_uuid` produced a submission that couldn't exist. Both meta fields now derive from `_uuid`, which makes edited submissions expressible in tests.
- Also fixed: editing a submission mid-job made its Data Table cell drop the progress spinner and look finished.
- 9 new unit tests.

### 👀 Preview steps

1. ℹ️ have an account, and a project with an audio question and the bulk processing feature flag enabled (`ff_bulkProcessingEnabled`)
2. submit at least 3 submissions with audio
3. open Project → Data → Table, open one submission in Edit, change any answer and resubmit it — this is what makes its `_uuid` drift from its root uuid
4. back in Project → Data → Table, select all submissions including the edited one
5. from the bulk actions menu choose "Transcribe selected…", pick a language and start
6. 🔴 [on main] the modal fails with `Unknown submission UUIDs: …` and no job starts for any of the submissions
7. 🟢 [on PR] the job starts for the whole selection, edited submission included, and the cells show progress until it completes
8. repeat steps 4–5 with "Translate selected transcripts" — 🔴 [on main] same error, 🟢 [on PR] works
9. select the finished transcriptions and use the bulk approve action
10. 🔴 [on main] it reports success but the edited submission's transcript stays unapproved
11. 🟢 [on PR] every selected transcript is approved


jamesrkiger (approved):
lgtm

PR #7421, opened 2026-08-12
fix(frontend): notification with multiple error lines layout DEV-2482

magicznyleszek:
### 📣 Summary

Improve layout of import error notification when multiple lines of text
needs to be displayed.

### 💭 Notes

This is something reportert by Becca, and I wasn't able to reproduce it.
But by looking at the code it's clear what was the problem.

Notification element is a flex container (probably to center the
content), so when it receives multiple children, it displays them in a
row. Before passing multiple children, we wrap them in a div.

### 👀 Preview steps

1. ℹ️ Be Becca
2. Cause the error to happen
3. 🟢 Notice that error now renders properly


jamesrkiger (approved):
LGTM

PR #7422, opened 2026-08-12
feat(trashBin): limit restarts to the deletions already running DEV-2618

rajpatel24:
### 📣 Summary
This PR stops `task_restarter` from piling more trash bin deletions onto workers that are already busy: it now asks the workers what they are running, and dispatches only up to the remaining headroom of `MAX_RESTARTED_TASKS`, shared across accounts, projects and attachments.

### 📖 Description
`task_restarter` dispatches up to `MAX_RESTARTED_TASKS` (100) objects per model every 30 minutes without considering tasks already in flight. Trash deletions are DB-intensive, so if they take longer than the beat interval, batches pile up, queues grow, and database pressure can quickly overwhelm servers.

What this PR changes:
Before dispatching, `task_restarter` checks active trash bin deletions and subtracts them from `MAX_RESTARTED_TASKS`. The remaining budget limits `_restart_stuck_tasks()` and shrinks as tasks are dispatched.

Only tasks that are running are counted. Ones still waiting in the queue are deliberately ignored: they hold no locks yet, so they are not what puts the database under pressure. The count is filtered by task name, so unrelated work on the same workers never eats the budget.

### 👀 Preview steps

1. ℹ️ have an account with ~10 projects, and the low priority Celery worker running 
2.  To make the limit observable, temporarily:
- set `MAX_RESTARTED_TASKS = 2` (at 100 it can never trigger, the worker only runs 6 tasks at a time)                                                                                                                                                     
- set the `trash-bin-task-restarter` schedule to `crontab(minute='*/1')`                                                                                                      
- add `time.sleep(120)` at the top of `empty_project()` so deletions stay visibly in flight                                                                                                                                                           
4.  Move the 10 projects to trash, then in the shell make them look stuck
5. 🔴 [on main] notice that this isn't anywhere
6. 🟢 [on PR] notice that the tasks are scheduled one by one.


rajpatel24:

rajpatel24:

rajpatel24:
[@greptileai](https://github.com/greptileai) About the partial inspection responses: valid, but it's inherent to inspect(), since a worker that misses the window is indistinguishable from one that no longer exists. The overshoot is bounded by the concurrency of whichever workers stayed silent rather than by the whole budget, and if none of them answer we skip the run entirely. Failing closed would be worse, since a legitimate scale-down looks identical to a silent worker, so we would skip runs and leave stuck deletions unrecoverable, which is exactly what this task exists to prevent.

noliveleger:
@rajpatel24 : part of the task says:

> The limit should also apply to the admin when asking to empty the trash bin. We should not try to empty them all at once. Maybe we could apply the same logic, i.e. if there is room to empty them right away, go ahead otherwise put them in the queue. 

I don't see anything in this PR for admin model. Are you going to tackle this in another PR?

rajpatel24:
Closed in favor of PR #7514.

PR #7424, opened 2026-08-12
feat(organizations): add sorting to Members Table DEV-2457

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

The Members table columns can now be sorted by name, status, date added, and role.

### 💭 Notes

Frontend for the sorting API added in #7329. Reuses `SortableProjectColumnHeader`, same as the Usage → Projects breakdown table, so the menus look and behave identically.

Changes here (all in `MembersRoute.tsx`):

- `MembersTableOrderableField` — the four fields the endpoint accepts
- `renderSortableHeader()` — swaps the four plain string labels for sortable ones
- `queryParams` — ordering state mapped to the API's `-` prefix convention, merged with pagination. Also passed to `getOrganizationsMembersListQueryKey()`, so each sort order caches separately and `keepPreviousData` keeps the table from flashing empty between sorts.
- `updateOrder()` — resets to page 1

Two columns stay unsorted, both because the API has no ordering field for them:
- **2FA**
- **Name sorts by username, not full name.** No ordering field exists for full name. Since the cell shows full name above username, sorting looks slightly out of order for members whose two differ — expected, not a bug.

### 👀 Preview steps

1. ℹ️ have an account in a team/organization with several members, ideally with mixed roles, a pending invite or two, and enough people to fill more than one page
2. go to Account Settings → (your team name) → Members
3. 🔴 [on main] notice the column headers are just labels — nothing to click, no way to reorder
4. 🟢 [on PR] notice **Name**, **Status**, **Date added** and **Role** now have a caret next to the label
5. click **Name** → choose "Sort A→Z", then "Sort Z→A" → 🟢 rows reorder by username, and a teal arrow marks the sorted column
6. choose "Default sort" → 🟢 the original order returns
7. repeat on **Status**, **Date added** and **Role** → 🟢 each sorts both ways
8. sort by something, then page forward and back with the arrows at the bottom → 🟢 the sort sticks across pages
9. go to page 2, then change the sort → 🟢 you land back on page 1 instead of somewhere in the middle of the newly sorted list
10. 🟢 notice the **2FA** column has no caret — the API can't sort by it yet

### DIFF
--- a/jsapp/js/account/organization/MembersRoute.tsx
+++ b/jsapp/js/account/organization/MembersRoute.tsx
@@ -62,10 +62,15 @@ export default function MembersRoute() {
     queryParams.ordering = `${orderPrefix}${order.fieldName}`
   }
 
-  /** Sorting affects which rows land on which page, so we go back to the first page whenever it changes. */
+  /**
+   * Sorting affects which rows land on which page, so we go back to the first page whenever it changes. Updating from
+   * the latest state, so we don't reset `limit` to a stale page size.
+   */
   function updateOrder(newOrder: SortableColumnOrder<MembersTableOrderableField>) {
     setOrder(newOrder)
-    setPagination({ ...pagination, start: 0 })
+    setPagination((currentPagination) => {
+      return { ...currentPagination, start: 0 }
+    })
   }
 
   const membersQuery = useOrganizationsMembersList(organization.id, queryParams, {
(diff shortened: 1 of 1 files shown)


magicznyleszek:
@greptileai review now!

duvld (approved):
LGTM

PR #7425, opened 2026-08-12
fix(bulkProcessing): review button showing up wrongly DEV-2676

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

The "Review" button for an automatic transcription now appears only in the column of the language it was transcribed into, instead of in every transcript column of that row.

### 💭 Notes

A question can only hold one transcript, so a transcript awaiting approval belongs to a single language column. Previously it was flagged in all of them, including columns that showed no text at all.

The same correction applies to "Approve all selected" in a transcript column header menu: it now stays disabled when the only pending transcripts in the selected rows are in a different language, and the approve confirmation no longer offers to approve them.

Changes here:

- `submissionUtils.tsx`
  - `hasUnacceptedAutomaticContent()` compares the transcript's `languageCode` against the column's before reporting pending content. The translation branch right below already did this.
  - `hasAnyUnacceptedAutomaticContent()` delegates per submission, so the "Approve all selected" menu item and `evaluateAlreadyApproved` pick up the fix for free.
- `bulkProcessingUtils.ts`
  - New `hasTranscriptInAnyLanguage()`. This one ignores the language **on purpose**: any existing transcript gets overwritten by a new transcription regardless of language, so it all counts as content at risk. Documented next to its language-aware sibling `hasTranslatableTranscript()` so nobody files this as the same bug six months from now.
- `alertEvaluators.ts` + `BulkTranscriptionModal.tsx`
  - Both had their own copy of the "does this row already have a transcript" check. They now share the helper above, which keeps the already-transcribed warning count and the quota estimate from drifting apart.
- Tests
  - `submissionUtils.tests.ts`: a pending transcript in one language must not light up another language's column, plus the same for the bulk-approve helper.
  - `bulkProcessingUtils.tests.ts`: pins the intentional language-blindness of the new helper.

No existing test needed changing.

### 👀 Preview steps

1. ℹ️ have a project with an audio question and at least three submissions
3. open the processing view for submission 1's audio question and transcribe it manually into Turkish
4. do the same for submission 2, but into English — the table now has two transcript columns
5. select submission 3, then use the audio question column header menu → Transcribe selected audio files, pick English and start it
6. wait for the bulk transcription to finish and refresh the table
7. 🔴 [on main] notice the "Review" button in submission 3 for **both** the Turkish and English transcript columns, even though the Turkish cell is otherwise empty
8. 🟢 [on PR] notice the "Review" button only in the English transcript column
9. 🟢 with submission 3 selected, open the Turkish transcript column header menu and notice "Approve all selected" is disabled; in the English column it's enabled


jamesrkiger (approved):
LGTM

PR #7428, opened 2026-08-13
fix(formbuilder): duplicate imported question name DEV-2269

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Fixed a bug where adding the same question or block from the Library multiple times gave every copy the same data column name in Formbuilder.

### 💭 Notes

Formbuilder was showing the same data column name for each copy, even though the saved form definition had already renamed them to be unique. Copies now get their unique name right away, so what you see in Formbuilder matches what ends up in the XLSForm.

Blocks were affected more seriously: adding one twice produced genuinely duplicated names in the form definition, not just in the display. Every question inside a block copy now gets a unique name too.

Changes here:

- **jsapp/xlform/src/model.survey.coffee**
  - `_ensure_row_name_is_unique()` - new. Runs `finalize()` (fills an empty name from the label) then `deduplicate()` (appends `_001` on a clash), in that order. Same mechanism the Duplicate button already uses.
  - `_is_row_name_taken()` - new. Guards the rename, because `deduplicate()` re-sluggifies and would truncate long names at 30 characters that had no clash to begin with.
  - `insertSurvey()` - single questions: renames the row that `add()` actually returns. Previously it renamed the library row before `add()` built a fresh one from JSON, so the rename went to an object nobody kept. Groups: renames the group and everything inside it, which was missing entirely.
- **jsapp/xlform/src/model.survey.d.ts** - declarations for the two new methods.
- **test/xlform/survey.tests.coffee** - 7 tests: question twice, block twice, clash with a hand-typed question, label-only library item, model vs. `toCsvJson()` parity, save/reload round trip, and long names left alone. Three of them fail without the fix.

### 👀 Preview steps

2. In the Library, create a `select_one` question named "Choose a fruit" with a few options, and a block containing two or three questions
3. Open a project in Formbuilder, click **Add from Library** in the right panel
4. Drag the "Choose a fruit" question into the form **twice**
5. Open the settings on both copies and compare **Data column name**
6. 🔴 [on main] both say `choose_a_fruit`
7. 🟢 [on PR] the second one says `choose_a_fruit_001`
8. Now drag the block in **twice** as well
9. 🟢 [on PR] every question in the second block copy has its own name, suffixed `_001`
10. SAVE, leave Formbuilder, then re-open it
11. 🟢 the names are still unique and unchanged - no surprise renaming on reload
12. Project → Settings → Download → XLS, and open the file
13. 🟢 the names in the spreadsheet match what Formbuilder showed you
14. Add a new option to the first fruit question, SAVE, re-open
15. 🟢 the second fruit question still has its original options - the two are no longer joined at the hip


pauloamorimbr (approved):
LGTM! 👏🏻

PR #7429, opened 2026-08-13
ci(release): maybe fix a false-positive transifex diff DEV-1916

Akuukis:
### 💭 Notes

git on CI weirdly didn't pick up modified files from transifex.
This change should check for changes better, as well reveal more info in case it doesn't work too.


noliveleger (approved):
LGTM

PR #7430, opened 2026-08-13
fix(organizations): sort role by level of privilege instead of alphabetically DEV-2454

Guitlle:
### 🗒️ Checklist

1. [x] run linter locally
10. [x] delete this checklist section from the final squash commit before merging

### 📣 Summary
Organizations members are now sorted by level of privilege in the members API when using `ordering=role` (previously they were sorted alphabetically)

### 👀 Preview steps
1. ℹ️ have an organization with multiple members with admin and member roles
2. 🟢 [on PR] notice that the members are sorted by level of privilege
    * Ascending Privilege (member → admin → owner): /api/v2/organizations/<UID>/members/?ordering=role
    * Descending Privilege (owner → admin → member): /api/v2/organizations/<UID>/members/?ordering=-role

### DIFF
--- a/kobo/apps/organizations/views.py
+++ b/kobo/apps/organizations/views.py
@@ -471,7 +471,11 @@ class OrganizationViewSet(viewsets.ModelViewSet):
                     'user__has_sso_enabled',
                     '-user__has_sso_enabled',
                 ],
-                description='Which field to use when ordering the results. Note: `role` sorts by privilege rank (`member` → `admin` → `owner`).',
+                description=(
+                    'Which field to use when ordering the results. '
+                    'Note: `role` sorts by privilege rank '
+                    '(`member` → `admin` → `owner`).'
+                ),
             ),
         ],
     ),
--- a/kobo/apps/organizations/views.py
+++ b/kobo/apps/organizations/views.py
@@ -2,17 +2,7 @@ from operator import attrgetter
 
 from allauth.socialaccount.models import SocialAccount
 from django.db import transaction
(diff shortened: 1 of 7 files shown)


kobo/apps/organizations/views.py
```diff
@@ -581,6 +582,13 @@ def get_queryset(self):
             output_field=CharField(),
         )
 
+        role_sort_annotation = Case(
+            When(Exists(owner_subquery), then=Value(3)),
+            When(is_admin=True, then=Value(2)),
+            default=Value(1),
+            output_field=IntegerField(),
+        )
+
```
rajpatel24: The sort happens in Python (line 641 materializes the list, then `sorted(...)`), so I think this `Case` never orders anything in SQL, it just re-runs the owner `EXISTS` a second time per row, on every action including retrieve.

We could derive the rank from the `role` we already have:
```
queryset = list(queryset) + list(invitees)
role_ranks = {'member': 1, 'admin': 2, 'owner': 3}
for row in queryset:
    row.role_sort = role_ranks.get(row.role, 0)
```
What do you think?

rajpatel24 (changes requested):
Please fix the darker and also run `./scripts/generate_api.sh ` to generate json/yaml files

kobo/apps/organizations/views.py
Guitlle: yeah, it's simpler

rajpatel24 (approved):
LGTM

PR #7437, opened 2026-08-17
fix(massEmails): stop OOM when generating send lists at scale DEV-2684

platreth:
## 📣 Summary
Automated emails, like usage limit alerts, go out again on very large servers.

## 📖 Description
On servers with millions of accounts, the daily job that prepares email recipient lists ran out of memory and stopped, so no emails were sent. The job now uses a fraction of the memory and the emails are sent as expected.

## 💭 Notes
- The pipeline loaded full User/Organization objects and built a dict per org for the whole fleet: several GB of memory on large instances, worker OOM-killed, failure silent (transaction rolls back, "done for today" flag never set, send task skips).
- Fix: ids via `values_list`, one shared default limits/dates dict (the addon merge copies before mutating; a test pins that), no giant `IN` clauses (`None` now means "all orgs"; per-user usage queries run in batches via `USAGE_QUERY_USER_ID_BATCH_SIZE`, default 20000 and clamped to at least 1), records enqueued by `user_id` in batches.
- Repro at 50k users: peak went from 94.4 MB to 22.5 MB, same 50,010 records.

## 👀 Preview steps
Backend only. Seed 50k users and run `enqueue_mass_email_records` under `tracemalloc`:
1. 🔴 on the base branch: peak ~94 MB for 50k recipients, growing ~2 KB per user, so several GB at production scale
2. 🟢 on this PR: peak ~22 MB, identical records created

### DIFF
--- a/kobo/apps/mass_emails/tasks.py
+++ b/kobo/apps/mass_emails/tasks.py
@@ -9,7 +9,7 @@ from constance import config
 from django.conf import settings
 from django.core.cache import cache
 from django.db import IntegrityError, transaction
-from django.db.models import Count, Q
+from django.db.models import Count, Q, QuerySet
 from django.utils import timezone
 from django.utils.translation import gettext
 
@@ -56,24 +56,32 @@ def enqueue_mass_email_records(email_config):
     Creates a email job and enqueues email records for users based on query
     """
     job = MassEmailJob.objects.create(email_config=email_config)
-    users = get_users_for_config(email_config)
+    user_ids = get_users_for_config(email_config)
     # edge case: if a one-off email has no recipients, store a warning and turn
     # it off
-    if len(users) == 0 and email_config.type == EmailType.ONE_TIME:
+    if len(user_ids) == 0 and email_config.type == EmailType.ONE_TIME:
         logging.warning(
             f'No recipients found for one-time email config'
             f' {email_config.uid}: {email_config.name}. Turning it off.'
         )
         email_config.live = False
         email_config.save()
-    records = [
-        MassEmailRecord(user=user, email_job=job, status=EmailStatus.ENQUEUED)
-        for user in users
-    ]
-    MassEmailRecord.objects.bulk_create(records)
+
+    # Instantiating millions of unsaved MassEmailRecord objects at once is ~1 GB,
+    # so build and insert them in batches instead.
+    batch_size = 10000
+    total_created = 0
+    for start_idx in range(0, len(user_ids), batch_size):
+        end_idx = start_idx + batch_size
+        records = [
+            MassEmailRecord(user_id=user_id, email_job=job, status=EmailStatus.ENQUEUED)
+            for user_id in user_ids[start_idx:end_idx]
+        ]
+        MassEmailRecord.objects.bulk_create(records, batch_size=batch_size)
+        total_created += len(records)
 
     logging.info(
-        f'Created {len(records)} MassEmailRecord(s) for {email_config.name} '
+        f'Created {total_created} MassEmailRecord(s) for {email_config.name} '
         f'with query {email_config.query}'
     )
 
@@ -392,7 +400,7 @@ def send_emails():
 
 def get_users_for_config(email_config):
     """
-    Get users based on query, excluding recent recipients
+    Get user ids based on query, excluding recent recipients
 
     frequency = -1: One time email
     frequency = 1: Daily emails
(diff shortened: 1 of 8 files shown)


platreth:
@greptileai

kobo/settings/base.py
```diff
@@ -2453,3 +2453,8 @@ def __init__(self, *args, **kwargs):
 LAST_PROJECT_ACTIVITY_THROTTLE_SECONDS = env.int(
     'LAST_PROJECT_ACTIVITY_THROTTLE_SECONDS', 3600  # seconds
 )
+
+# Fleet-wide usage queries filter on `user_id__in=<all users>`; at full scale
+# that is a multi-million-element IN clause. Split it into chunks of this size
+# and run one query per chunk.
+USAGE_QUERY_USER_ID_CHUNK_SIZE = env.int('USAGE_QUERY_USER_ID_CHUNK_SIZE', 20000)
```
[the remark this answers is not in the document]
platreth: Good catch. Clamped to a minimum of 1 at both use sites in 2dbbc77, with a regression test covering the 0 case.

platreth:
@greptileai update the score

kobo/settings/base.py
noliveleger: I would regroup this setting with other "_BATCH_SIZE" variables. See https://github.com/kobotoolbox/kpi/blob/6d6a0c6c7f779fd97d2e1429425a1799d95a3629/kobo/settings/base.py#L2397 
I would also keep the same convention for the name (i.e. `_BATCH_SIZE` vs `CHUNK_SIZE`).

kobo/apps/stripe/utils/subscription_limits.py
```diff
@@ -209,24 +227,41 @@ def get_organizations_effective_limits(
...
+        # `addon_limits` is keyed only by the (small) set of orgs holding addons.
         addon_limits = PlanAddOn.get_organizations_totals(organizations=organizations)
-        for org_id, limits in effective_limits.items():
+        for org_id, org_addons in addon_limits.items():
+            limits = effective_limits.get(org_id)
+            if limits is None:
+                continue
+            # `limits` may be the shared default object reused across orgs; copy
+            # before mutating so we don't corrupt every other org's limits.
+            limits = dict(limits)
```
noliveleger: We should use `deepcopy` here. `limits` may not contain a nested dict but could in the future and it could be mutated with `dict` only.

```python
>>> d = {'foo': {'bar': 'foobar'}}
>>> d_copy = dict(d)
>>> d_copy['foo']['bar'] = 'barfoo'
>>> d
{'foo': {'bar': 'barfoo'}}  # d is mutated

>>> d = {'foo': {'bar': 'foobar'}}
>>> from copy import deepcopy
>>> d = {'foo': {'bar': 'foobar'}}
>>> d_copy = deepcopy(d)
>>> d_copy['foo']['bar'] = 'barfoo'
>>> d
{'foo': {'bar': 'foobar'}} # d stays intact
```

kobo/apps/stripe/utils/subscription_limits.py
```diff
@@ -209,24 +227,41 @@ def get_organizations_effective_limits(
...
-        orgs = organizations or Organization.objects.all()
-        return {org.id: _get_default_usage_limits() for org in orgs}
+        if organizations is not None:
+            org_ids = [org.id for org in organizations]
+        else:
+            # Avoid loading millions of full Organization instances just to read ids.
+            org_ids = Organization.objects.values_list('id', flat=True).iterator()
+        # Share one default-limits object across all orgs; callers that mutate a
+        # returned dict must copy it first.
+        shared_default_limits = _get_default_usage_limits()
```
noliveleger: It's a little bit scary. I would make `shared_default_limits` a not-mutable object. (see `from types import MappingProxyType`)

kobo/apps/stripe/utils/billing_dates.py
```diff
@@ -23,28 +23,28 @@ def get_current_billing_period_dates_by_org(
     first_of_next_month = first_of_this_month + relativedelta(months=1)
 
     if not settings.STRIPE_ENABLED:
+        # Callers only read these dates, so share one object across all orgs and
+        # stream ids to avoid loading millions of full Organization instances.
+        shared_dates = {'start': first_of_this_month, 'end': first_of_next_month}
```
noliveleger: Same comment as `shared_default_limits`. Let's use `types.MappingProxyType` instead.

kobo/apps/stripe/utils/subscription_limits.py
noliveleger: Watch-out if you use `types.MappingProxyType` , you would need to do `deepcopy(dict(limits))` . Otherwise, it will raise an error. `deepcopy` does not work with `types.MappingProxyType`.

noliveleger (changes requested):

kobo/apps/mass_emails/tasks.py
```diff
@@ -56,24 +56,32 @@ def enqueue_mass_email_records(email_config):
...
         email_config.save()
-    records = [
-        MassEmailRecord(user=user, email_job=job, status=EmailStatus.ENQUEUED)
-        for user in users
-    ]
-    MassEmailRecord.objects.bulk_create(records)
+
+    # Instantiating millions of unsaved MassEmailRecord objects at once is ~1 GB,
+    # so build and insert them in batches instead.
+    batch_size = 10000
```
noliveleger: I'd rather use the same setting everywhere:  `USAGE_QUERY_USER_ID_BATCH_SIZE`. 

Sorry, my comment did not make through the first time.

noliveleger (changes requested):

PR #7440, opened 2026-08-17
feat(massEmails): pace sends against the provider rate limit DEV-2681

noliveleger:
### 📣 Summary

Mass email campaigns now pace themselves to stay within the email provider's sending limits, instead of risking being throttled or blocked for sending too fast.

Two new environment variables control mass-email pacing and replace the removed `MASS_EMAIL_SLEEP_SECONDS`:

- `MASS_EMAIL_SEND_RATE_RATIO` (default `0.35`): share of `MASS_EMAIL_THROTTLE_PER_SECOND` actually claimed per second, leaving headroom for transactional email and other instances sharing the same provider account. `0.5` or below is recommended so two full budget windows landing back to back still can't exceed the provider's rate limit; a Django system check only blocks values outside `(0, 1]`, an admin can go above `0.5` at the risk of bursting past the limit near a window boundary.
- `MAILER_CONNECTION_IDLE_TIMEOUT` (default `10`): seconds a reused SMTP connection can sit idle before it's proactively checked and reconnected ahead of the next send.

`MASS_EMAIL_THROTTLE_PER_SECOND` and `MAX_MASS_EMAILS_PER_DAY` keep their existing meaning, but should be reviewed per deployment: they must reflect the actual provider's real limits, which differ between a main SES account, a secondary SES account, SendGrid, or Office365. No migrations.

### 💭 Notes

`Mailer.send()` no longer returns a `bool`; it raises a typed exception on failure (new hierarchy in `kpi/exceptions.py`: `MailerError`, `MailerProviderThrottledError` → `MailerProviderRateThrottledError`/`MailerProviderQuotaExhaustedError`, matched against a per-provider SMTP signature catalog for SES/SendGrid/Office365 sourced from official docs, not live-probed against a real relay — an unmatched response falls back to today's plain-failure behavior). This is a breaking API change; every caller in the repo was updated, including four "solo" send sites (org invites, project-ownership invites) that previously ignored the return value and silently swallowed failures.

`send_day_emails()`'s fixed batch-sleep throttle is replaced by a real 1-second sliding-window budget. Connection reconnect-on-drop logic moved from `mass_emails/tasks.py` into `Mailer.send()` itself so every caller benefits, plus a new idle-timeout-aware proactive check. The existing "don't resend on an ambiguous connection drop" fix (DEV-2675, PR #7435) is preserved unchanged.

`MailerProviderRateThrottledError` currently stops the run exactly like `MailerProviderQuotaExhaustedError` (flagged by Greptile: an earlier version slept `MASS_EMAIL_THROTTLE_COOLDOWN_SECONDS` synchronously, which exceeded the task's own soft time limit). That makes the two exceptions synonyms for now - a real non-blocking cooldown is tracked in DEV-2693.

### 👀 Preview steps

No UI surface here (background Celery task + SMTP). Quick reference, no manual setup:

- `kpi/tests/test_mailer.py::TestClassifySmtpException` — a mocked SES/SendGrid/Office365 throttle response comes back as `MailerProviderRateThrottledError` or `MailerProviderQuotaExhaustedError`
- `kobo/apps/mass_emails/tests/test_celery_tasks.py::TestMassEmailSenderConnection` — a rate-throttled or quota-exhausted record stops the run outright and stays `enqueued` for the next scheduled run (see the Notes section on why these two aren't differentiated yet - DEV-2693)
- `kobo/apps/mass_emails/tests/test_celery_tasks.py::TestMassEmailSender::test_send_is_throttled` — the per-second budget actually paces sends instead of the old fixed-batch sleep

To watch it live end to end, from scratch, in `manage.py shell` (`docker exec -it kobofe-kpi-1 /opt/venv/bin/python manage.py shell`):

1. ℹ️ create ~20 throwaway users whose email points back to your own inbox (Gmail plus-addressing works: `you+test1@gmail.com` etc. all land in `you@gmail.com`)
   ```python
   from kobo.apps.kobo_auth.shortcuts import User

   test_emails = [f'you+test{i}@gmail.com' for i in range(1, 20)]
   for i, email in enumerate(test_emails):
       User.objects.get_or_create(username=f'mass_email_test_{i}', defaults={'email': email})
   ```
2. point the `test_users` query at them (it's the query mass email configs use for exactly this purpose - see `get_all_test_users()`)
   ```python
   from constance import config
   config.MASS_EMAIL_TEST_EMAILS = '\n'.join(test_emails)
   ```
3. create a config targeting them. Note the `date_created` backdate: `generate_mass_email_user_lists()` only picks up configs created before today
   ```python
   from kobo.apps.mass_emails.models import MassEmailConfig

   email_config = MassEmailConfig.objects.create(
       name='Manual test',
       subject='Test mass email',
       template='Hello ##username##, this is a test.',
       query='test_users',
       frequency=1,
       live=True,
       date_created=timezone.now() - timedelta(days=1),
   )
   ```
4. run the real pipeline synchronously (no worker needed, these are plain functions)
   ```python
   from kobo.apps.mass_emails.tasks import generate_mass_email_user_lists, send_emails

   generate_mass_email_user_lists()
   send_emails()
   ```
5. 🟢 watch the logs: with the default settings (`MASS_EMAIL_THROTTLE_PER_SECOND=40`, `MASS_EMAIL_SEND_RATE_RATIO=0.35`, budget = 14/s) with ~20 records you should see 14 `Processing MassEmailRecord(...)` lines, one `sleeping for ...s to stay within the per-second budget` line, then the rest - not the old fixed "N sent, sleep 1s" pattern
6. 🟢 to see it paced more slowly and watchably, drop the budget in the same shell session before re-running (`send_emails()` reads `django.conf.settings` live, no restart needed): `from django.conf import settings; settings.MASS_EMAIL_THROTTLE_PER_SECOND = 2; settings.MASS_EMAIL_SEND_RATE_RATIO = 1.0` gives a budget of 2/s
7. re-running the same day: `generate_mass_email_user_lists()` and the per-config daily quota are both cached for the day and won't re-trigger; reset with:
   ```python
   from kobo.apps.mass_emails.models import EmailStatus, MassEmailRecord
   from kobo.apps.mass_emails.tasks import PROCESSED_EMAILS_CACHE_KEY, MassEmailSender

   MassEmailRecord.objects.filter(email_job__email_config=email_config).update(status=EmailStatus.ENQUEUED)
   cache_key_date = MassEmailSender.get_cache_key_date(timezone.now())
   cache.delete(PROCESSED_EMAILS_CACHE_KEY.format(key_date=cache_key_date))
   cache.delete(f'mass_emails_{cache_key_date.isoformat()}_email_remaining_total')
   cache.delete(f'mass_emails_{cache_key_date.isoformat()}_email_remaining_{email_config.id}')
   ```
8. no real email lands anywhere unless `EMAIL_BACKEND` is pointed at real SMTP credentials - the dev default (`filebased`) just writes messages to `EMAIL_FILE_PATH` instead
9. cleanup: `email_config.live = False; email_config.save()` (or delete jobs/records then the config - `MassEmailJob`/`MassEmailRecord` use `on_delete=PROTECT`, so the config can't be deleted while they exist)

### DIFF
--- a/kobo/apps/mass_emails/tasks.py
+++ b/kobo/apps/mass_emails/tasks.py
@@ -306,9 +306,9 @@ class MassEmailSender:
                         sleep(remaining)
                     window_start = monotonic()
                     spent_in_window = 0
+                self.send_email(email_config, record)
                 self.cache_limit_value(email_config, self.limits[email_config.id] - 1)
                 self.cache_limit_value(None, self.total_limit - 1)
-                self.send_email(email_config, record)
                 spent_in_window += 1
 
     def send_email(self, email_config, record):
--- a/kobo/apps/mass_emails/tasks.py
+++ b/kobo/apps/mass_emails/tasks.py
@@ -345,11 +345,10 @@ class MassEmailSender:
             # run there. The record stays `enqueued` for the next run.
             raise
         except MailerProviderRateThrottledError as e:
-            logging.warning(
-                f'Provider rate limit hit, cooling down for '
-                f'{settings.MASS_EMAIL_THROTTLE_COOLDOWN_SECONDS}s: {e}'
-            )
-            sleep(settings.MASS_EMAIL_THROTTLE_COOLDOWN_SECONDS)
+            logging.warning(f'Provider rate limit hit, stopping this run: {e}')
+            # TODO(DEV-2693): needs its own non-blocking cooldown instead of
+            # being treated exactly like a quota-exhausted stop.
+            raise
         except MailerError as e:
             logging.warning(f'Error sending record {record}: {e}')
             record.status = EmailStatus.FAILED
@@ -386,7 +385,7 @@ def send_emails():
     sender = MassEmailSender()
     try:
         sender.send_day_emails()
-    except MailerProviderQuotaExhaustedError:
+    except (MailerProviderQuotaExhaustedError, MailerProviderRateThrottledError):
         # Already logged in `send_email()`. Nothing left to do this run: the
         # remaining `enqueued` records are picked up by the next hourly run.
         pass
--- a/kobo/apps/mass_emails/tasks.py
+++ b/kobo/apps/mass_emails/tasks.py
@@ -1,7 +1,7 @@
 from datetime import date, datetime, time, timedelta
 from enum import Enum
 from math import ceil
-from time import sleep
+from time import monotonic, sleep
 from typing import Optional
 
 from celery.exceptions import SoftTimeLimitExceeded
@@ -22,14 +22,13 @@ from kobo.apps.mass_emails.models import (
 )
 from kobo.apps.organizations.models import OrganizationUser
 from kobo.celery import celery_app
-from kpi.utils.log import logging
-from kpi.utils.mailer import (
-    EmailMessage,
-    Mailer,
-    is_connection_alive,
-    reset_connection,
-    with_smtp_connection,
+from kpi.exceptions import (
+    MailerError,
+    MailerProviderQuotaExhaustedError,
+    MailerProviderRateThrottledError,
 )
+from kpi.utils.log import logging
+from kpi.utils.mailer import EmailMessage, Mailer, with_smtp_connection
 
 if settings.STRIPE_ENABLED:
     from kobo.apps.stripe.utils.subscription_limits import get_plan_name
@@ -269,9 +268,18 @@ class MassEmailSender:
 
     @with_smtp_connection
     def send_day_emails(self):
-        emails_sent = 0
-
-        batch_size = settings.MASS_EMAIL_THROTTLE_PER_SECOND
+        # Claim only a share of the provider's real per-second limit: it
+        # also carries transactional email, which draws on the same budget
+        # without going through this throttle.
+        budget_per_second = max(
+            1,
+            int(
+                settings.MASS_EMAIL_THROTTLE_PER_SECOND
+                * settings.MASS_EMAIL_SEND_RATE_RATIO
+            ),
+        )
+        window_start = monotonic()
+        spent_in_window = 0
 
         for email_config in self.configs:
             limit = self.limits.get(email_config.id)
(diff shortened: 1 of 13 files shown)


kobo/apps/mass_emails/tasks.py
```diff
@@ -311,38 +329,38 @@ def send_email(self, email_config, record):
...
+            # Let this propagate, same as `SoftTimeLimitExceeded`: it must
+            # not be recorded as a failure, and `send_emails()` stops the
+            # run there. The record stays `enqueued` for the next run.
+            raise
+        except MailerProviderRateThrottledError as e:
+            logging.warning(
+                f'Provider rate limit hit, cooling down for '
+                f'{settings.MASS_EMAIL_THROTTLE_COOLDOWN_SECONDS}s: {e}'
+            )
+            sleep(settings.MASS_EMAIL_THROTTLE_COOLDOWN_SECONDS)
```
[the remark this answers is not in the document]
noliveleger: Addressed in [452f0cb](https://github.com/kobotoolbox/kpi/pull/7440/commits/452f0cbdf0578db84da9676b2af5c0f98b6be904)

kobo/apps/mass_emails/tasks.py
```diff
@@ -285,13 +293,23 @@ def send_day_emails(self):
...
+                    remaining = 1 - (monotonic() - window_start)
+                    if remaining > 0:
+                        logging.info(
+                            f'sleeping for {remaining:.3f}s to stay within '
+                            f'the per-second budget'
+                        )
+                        sleep(remaining)
+                    window_start = monotonic()
+                    spent_in_window = 0
                 self.cache_limit_value(email_config, self.limits[email_config.id] - 1)
```
platreth, line 310: I think it might be better to put this after sending the email, then we don't decrement if the email is not sent

kpi/utils/mailer.py
```diff
@@ -170,47 +131,223 @@ def to_multi_alternative(self):
...
+            # re-sending messages that already went out.
+            raise cls._classify_smtp_exception(e) from e
+
+    @classmethod
+    def _send_single(
+        cls, email_message: EmailMessage, connection: Optional[BaseEmailBackend]
+    ):
+        try:
+            cls._dispatch_single(email_message, connection)
+        except SMTPException as e:
```
platreth: We should also add OSError

kpi/utils/mailer.py
```diff
@@ -170,47 +131,223 @@ def to_multi_alternative(self):
...
+    # - SES: https://docs.aws.amazon.com/ses/latest/dg/manage-sending-quotas-errors.html
+    # - SendGrid: https://twilio.com/docs/sendgrid/for-developers/sending-email/smtp-errors-and-troubleshooting  # noqa
+    # - Office365/Exchange: https://learn.microsoft.com/en-us/troubleshoot/exchange/email-delivery/ndr/non-delivery-reports-in-exchange-online.  # noqa
+    #   Enhanced codes there are listed without a paired basic SMTP code; 550
+    #   is Microsoft's documented convention for a 5.x.x enhanced code
+    #   returned live at submission time.
+    #
+    # An unmatched response keeps today's plain failure behavior
+    # (`MailerError`), no pause.
+    THROTTLE_SIGNATURES = (
```
platreth, line 154: Let's lower case on both side before comparing to void case sensitive errors

platreth (changes requested):

platreth (approved):

PR #7442, opened 2026-08-18
fix(projectDownloads): kml exporting bug DEV-2695

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Fixed an error that blocked KML exports in projects where a GeoJSON export was created earlier.

### 📖 Description

Options that apply to a single export format no longer follow you when you switch format. Before this fix, exporting GeoJSON and then KML sent the GeoJSON-only "Flatten GeoJSON" option along with the KML request, and the server refused it. Projects already in that state recover on their next export, nothing to clean up.

### 💭 Notes

Changes here:

- **`exportsUtils.ts`**
  - `preserveApiOnlySettings()` replaces the inline merge loop that lived in `onSubmit`. When the form overwrites a saved export setting, keep the options the API supports but our form has no field for (`submission_ids` and friends) — except it now skips every key listed in `FORM_MANAGED_EXPORT_SETTINGS`.
  - That list is the actual fix. Each export saves its payload as an export setting (a nameless "Latest unsaved settings" record when you don't save a named preset), and the old loop copied back anything the fresh payload didn't have. Some of those keys are absent on purpose: `flatten` only ships for GeoJSON, `xls_types_as_text` for XLS. So a GeoJSON export left `flatten` in storage for the next KML export to inherit, and Backend rejects `flatten` for KML.
- **`ProjectExportsCreator.tsx`**
  - calls the new util, and `applyExportSettingToState()` now falls back to `DEFAULT_EXPORT_SETTINGS` for the three format-specific options
- **`exportsUtils.tests.ts`** -
  - new file: the GeoJSON → KML regression, plus the settings we still want carried over.

### 👀 Preview steps

1. ℹ️ have an account and a deployed project with a geopoint question and at least one submission
2. go to Project → Data → Downloads
3. set Export type to GeoJSON, click Export, wait for the file to appear in the list below
4. set Export type to KML, click Export
5. 🔴 [on main] error message, no KML file
6. 🟢 [on PR] the KML export is created and can be downloaded
7. switch back to GeoJSON, open Advanced options, untick "Flatten GeoJSON", export it, then export KML once more — 🟢 both still work
8. reload the page so the form restores your last used (KML) settings, then set Export type to GeoJSON and open Advanced options — 🟢 "Flatten GeoJSON" comes back checked, which is its default, instead of blank

### DIFF
--- a/jsapp/js/components/projectDownloads/ProjectExportsCreator.tsx
+++ b/jsapp/js/components/projectDownloads/ProjectExportsCreator.tsx
@@ -31,12 +31,13 @@ import {
   type ExportFormatOption,
   getContextualDefaultExportFormat,
   getExportFormatOptions,
+  preserveApiOnlySettings,
 } from '#/components/projectDownloads/exportsUtils'
 import { openDeleteExportSettingModal } from '#/components/projectDownloads/openDeleteExportSettingModal'
 import { getColumnLabel } from '#/components/submissions/tableUtils'
 import { ADDITIONAL_SUBMISSION_PROPS, SUPPLEMENTAL_DETAILS_PROP } from '#/constants'
 import type { AssetResponse, ExportSetting, ExportSettingRequest, MongoQuery } from '#/dataInterface'
-import { createDateQuery, formatTimeDate, notify, recordEntries, recordKeys, recordValues } from '#/utils'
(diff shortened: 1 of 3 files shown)


jamesrkiger (approved):
LGTM

PR #7445, opened 2026-08-18
fix(massEmails): exclude deactivated users from mass emails DEV-2691

platreth:
### 📣 Summary

Deactivated accounts no longer receive mass emails.

### 📖 Description

Deactivated accounts could still show up as recipients of mass emails. That includes accounts moved to a private server, which stay around for months before their data is deleted. They're now left out of the recipient lists, and an account deactivated after a list was built won't get the email either.

### 💭 Notes

- `get_active_users()` was the only recipient query missing the `is_active` / trash exclusion the others already had.
- Builds on the eligibility recheck from #7448: records whose user is deactivated, deleted or trashed are marked `stale`. `MassEmailSender.__init__` sweeps them up front (keeps counts accurate, lets one-time configs terminate), and a per-record guard in the send loop catches changes that happen mid-run, regardless of the `MASS_EMAIL_STALE_RECORD_RECHECK_HOURS` threshold.
- `is_user_still_eligible()` now returns `False` for a deleted/deactivated/trashed user before consulting the check registry, so the fail-open path can't apply to them.

### 👀 Preview steps

1. ℹ️ have a superuser and a user who logged in within the last year
2. deactivate that user in `/admin/` (uncheck `Active`)
3. in a Django shell: `from kobo.apps.mass_emails.user_queries import get_active_users`
4. run `get_active_users().filter(username='<user>').exists()`
5. 🔴 [on 2.026.30] returns `True`, the deactivated user is still a mass email recipient
6. 🟢 [on PR] returns `False`

### DIFF
--- a/kobo/apps/mass_emails/tasks.py
+++ b/kobo/apps/mass_emails/tasks.py
@@ -122,6 +122,16 @@ class MassEmailSender:
         cache_date = self.get_cache_key_date(send_date=now)
         self.cache_key_prefix = f'mass_emails_{cache_date.isoformat()}_email_remaining'
 
+        # Users deactivated, deleted or trashed after enqueue must not be
+        # emailed. Mark their records stale up front so counts stay accurate
+        # and one-time configs can still finish and be turned off.
+        MassEmailRecord.objects.filter(
+            Q(user__isnull=True)
+            | Q(user__is_active=False)
+            | Q(user__trash__isnull=False),
+            status=EmailStatus.ENQUEUED,
+        ).update(status=EmailStatus.STALE)
+
         self.total_records = MassEmailRecord.objects.filter(
             status=EmailStatus.ENQUEUED
         ).count()
@@ -309,16 +319,20 @@ class MassEmailSender:
             for record in records.iterator():
                 if budget_used >= limit:
                     break
-                # Skip a stale record that no longer matches its config's criteria
-                if (
-                    record.date_created < stale_threshold
-                    and not email_config.is_user_still_eligible(record.user)
+                # A user deactivated, deleted or trashed after enqueue must
+                # never be emailed, no matter how fresh the record is;
+                # `record.user` is loaded here, so this reads their current
+                # state. A record older than the stale threshold is also
+                # re-checked against its config's criteria.
+                if not email_config.is_user_still_eligible(
+                    record.user,
+                    reevaluate_query=record.date_created < stale_threshold,
                 ):
                     record.status = EmailStatus.STALE
                     record.save(update_fields=['status', 'date_modified'])
                     logging.info(
                         f'Skipping stale MassEmailRecord({record}): recipient '
-                        f'no longer matches {email_config.query}'
+                        f'is no longer eligible for {email_config.query}'
                     )
                     continue
                 if spent_in_window >= budget_per_second:
--- a/kobo/apps/mass_emails/user_queries.py
+++ b/kobo/apps/mass_emails/user_queries.py
@@ -25,6 +25,7 @@ def get_active_users(days: int = 365) -> QuerySet:
     Retrieve users who have been active within specified number of days.
 
     A user is considered active if:
+    - They are not deactivated (``is_active=True``) and not in the trash
     - They have logged in within the given period, or were recently created and never
       logged in
     - They have modified or created an asset within the given period
@@ -47,7 +48,7 @@ def get_active_users(days: int = 365) -> QuerySet:
         extra_details__last_project_activity__gt=inactivity_threshold
     )
     return User.objects.filter(recent_login_filter | recently_active).exclude(
-        pk=settings.ANONYMOUS_USER_ID
+        Q(pk=settings.ANONYMOUS_USER_ID) | Q(is_active=False) | Q(trash__isnull=False)
     )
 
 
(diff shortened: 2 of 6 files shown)


platreth:
@greptileai

kobo/apps/mass_emails/user_queries.py
```diff
@@ -46,7 +47,7 @@ def get_active_users(days: int = 365) -> QuerySet:
         extra_details__last_project_activity__gt=inactivity_threshold
     )
     return User.objects.filter(recent_login_filter | recently_active).exclude(
-        pk=settings.ANONYMOUS_USER_ID
+        Q(pk=settings.ANONYMOUS_USER_ID) | Q(is_active=False) | Q(trash__isnull=False)
```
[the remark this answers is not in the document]
platreth: This line is 86 chars, under the 88 limit, and it's exactly how darker formats it (the darker CI check passes). Leaving as is!

platreth:
@greptileai check again

kobo/apps/mass_emails/tasks.py
```diff
@@ -321,6 +332,14 @@ def send_day_emails(self):
 
     def send_email(self, email_config, record):
         logging.info(f'Processing MassEmailRecord({record})')
+        # A run drains its records over several minutes, so a user can be
+        # deactivated (or deleted) after the batch was read but before we
+        # reach their record. `record.user` is loaded here, so this reads the
+        # current state without an extra query.
+        if record.user is None or not record.user.is_active:
+            record.status = EmailStatus.FAILED
+            record.save()
```
noliveleger: ~In #7448 I replaced `.save()` to `.save(update_fields=['status', 'date_modified'])`.~

~Let's be *consistent* 😄~

The comment is useless if we go with https://github.com/kobotoolbox/kpi/pull/7445#discussion_r3807389824

kobo/apps/mass_emails/tasks.py
```diff
@@ -321,6 +332,14 @@ def send_day_emails(self):
 
     def send_email(self, email_config, record):
         logging.info(f'Processing MassEmailRecord({record})')
+        # A run drains its records over several minutes, so a user can be
+        # deactivated (or deleted) after the batch was read but before we
+        # reach their record. `record.user` is loaded here, so this reads the
+        # current state without an extra query.
+        if record.user is None or not record.user.is_active:
```
noliveleger: Look at #7448, we could move this logic in "is_user_still_eligible()"?  directly what do you think? And mark the record as `stale` instead of `failed`? 

EDIT: I think we could merge #7448 work in your branch and use `is_user_active_and_not_trashed` directly inside `is_user_still_eligible()`  . We should also test if user is not None.

kobo/apps/mass_emails/tasks.py
```diff
@@ -122,6 +122,14 @@ def __init__(self):
         cache_date = self.get_cache_key_date(send_date=now)
         self.cache_key_prefix = f'mass_emails_{cache_date.isoformat()}_email_remaining'
 
+        # Users deactivated (or deleted) after enqueue must not be emailed.
+        # Fail their records up front so counts stay accurate and one-time
+        # configs can still finish and be turned off.
+        MassEmailRecord.objects.filter(
+            Q(user__is_active=False) | Q(user__isnull=True),
+            status=EmailStatus.ENQUEUED,
+        ).update(status=EmailStatus.FAILED)
```
noliveleger: I would use the new status from #7448 , i.e. : `stale`. Otherwise, let's create a new status.

kobo/apps/mass_emails/tasks.py
```diff
@@ -293,9 +301,12 @@ def send_day_emails(self):
             limit = self.limits.get(email_config.id)
             if not limit:
                 continue
+            # Guards users deactivated mid-run; those deactivated between runs
+            # are already failed by __init__'s eager update.
             records = MassEmailRecord.objects.filter(
                 status=EmailStatus.ENQUEUED,
                 email_job__email_config=email_config,
+                user__is_active=True,
```
noliveleger: I understand the reptile's comment but I'm wondering if we really want to this filter (which adds a `JOIN` here). Especially if we are re-checking the user's state [inside the loop](https://github.com/kobotoolbox/kpi/pull/7445/changes#diff-7108b9ada285887566a56b06f688203b433d1202ec15f44b375d5a4421f1b973R339). Since we don't use `.select_related(user)`, we do fetch again the user inside the loop (in purpose, to get the user's state as fresh as possible). 

However, if we keep this condition, we may also need to validate that the user is not null (i.e.: User could have been deleted) like we do everywhere else. Besides, we probably also need to check if they are not in trash, right? But what's the performance cost for that?

noliveleger (changes requested):

kobo/apps/mass_emails/tasks.py
noliveleger: See this comment https://github.com/kobotoolbox/kpi/pull/7448#discussion_r3813437126

kobo/apps/mass_emails/tasks.py
```diff
@@ -309,6 +320,21 @@ def send_day_emails(self):
...
                 if budget_used >= limit:
                     break
+                # A user deactivated, deleted or trashed after enqueue must
+                # never be emailed, no matter how fresh the record is.
+                # `record.user` is loaded here, so this reads their current
+                # state. Records that predate this run are already marked
+                # stale by `__init__`; this catches mid-run changes.
+                if record.user is None or not is_user_active_and_not_trashed(
+                    record.user
+                ):
```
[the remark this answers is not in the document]
platreth: This is the design agreed with @noliveleger in the thread above: no state filter at query time, one fresh fetch of the user per record right before sending (no select_related), covering deactivated, deleted and trashed.

noliveleger (approved):
LGTM

PR #7448, opened 2026-08-18
feat(massEmails): re-validate recipient eligibility before sending stale records DEV-2682

noliveleger:
### 📣 Summary

Mass email recipients who no longer meet a campaign's criteria by the time their email is about to be sent are now skipped instead of receiving a message that's no longer relevant to them.

### 📖 Description

A large mass email batch can take days to fully send because of the daily send cap. During that time, a recipient can stop matching the campaign's criteria (for example, buying a storage add-on that resolves the over-limit condition that triggered the email), but until now the system had no way to notice and would send the email anyway, sometimes days later. Before sending any record that's been sitting in the queue for a while, we now re-check that the recipient still matches the campaign's criteria, and skip the send if they don't.

### 💭 Notes

- New Constance setting `MASS_EMAIL_STALE_RECORD_RECHECK_HOURS` (default 12) controls how old a record must be before the recheck kicks in.
- New `USER_ELIGIBILITY_CHECKS` registry in `kobo/apps/mass_emails/models.py`, parallel to the existing `USER_QUERIES`: one cheap single-user check per query type, added to `user_queries.py`.
- New `EmailStatus.STALE` status for records skipped this way.

### 👀 Preview steps

Uses the existing `test_users` query (`MASS_EMAIL_TEST_EMAILS`), so no Stripe/usage setup needed. `MassEmailRecord` isn't registered in the admin, so checking record status needs the shell; everything else can go through the admin.

1. ℹ️ shell: clear the cache first — `send_day_emails()` caches the day's send budget per config id, so a stale cache entry from an earlier test can silently make it skip a newly created config
   ```python
   cache.clear()
   ```
2. admin → Constance config (`/admin/constance/config/`): set `MASS_EMAIL_TEST_EMAILS` to your test user's email, and `MASS_EMAIL_STALE_RECORD_RECHECK_HOURS` to `0` (so the recheck applies immediately, no backdating needed)
3. admin → Mass Email Configs (`/admin/mass_emails/massemailconfig/add/`): create a config with `query=test_users`, `frequency=-1`, `live` checked
4. shell: `generate_mass_email_user_lists()` only picks up configs created before today, so enqueue directly instead of waiting for tomorrow
   ```python
   from kobo.apps.mass_emails.tasks import enqueue_mass_email_records
   from kobo.apps.mass_emails.models import MassEmailConfig
   email_config = MassEmailConfig.objects.get(name='<your config name>')
   enqueue_mass_email_records(email_config)
   ```
5. admin → Constance config: clear `MASS_EMAIL_TEST_EMAILS` — the recipient no longer matches `test_users`
6. shell:
   ```python
   from kobo.apps.mass_emails.tasks import MassEmailSender
   MassEmailSender().send_day_emails()
   ```
7. 🔴 [on main] the record is sent regardless
8. 🟢 [on PR] the record is skipped, confirmed by this run's actual log output:
   ```
   INFO Found 1 enqueued records
   INFO Setting up MassEmailConfig limits for the current day
   INFO Processing 1 records for MassEmailConfig(test)
   INFO Skipping stale MassEmailRecord(test send to user): recipient no longer matches test_users
   ```


kobo/apps/mass_emails/tasks.py
```diff
@@ -296,11 +299,28 @@ def send_day_emails(self):
...
+            # No `[:limit]` slice: a stale record must not burn a slot without
+            # spending budget, so keep pulling records until `limit` is spent.
+            budget_used = 0
+            for record in records.iterator():
+                if budget_used >= limit:
+                    break
+                # Skip a stale record that no longer matches its config's criteria
+                if (
+                    record.date_created < stale_threshold
+                    and not email_config.is_user_still_eligible(record.user)
```
platreth, line 315: We should add a guard for `record.user`, it can be None here

Also looking at `is_user_still_eligible` this is a real check per user that can fail in a lot of ways 

Why not doing something like this: 

```suggestion
                    if record.date_created < stale_threshold:
    try:
        still_eligible = (
            record.user is not None
            and email_config.is_user_still_eligible(record.user)
        )
    except SoftTimeLimitExceeded:
        raise
    except Exception:
        logging.exception( (#or info?)
            f'Eligibility recheck failed for record {record.pk} '
            f'(config {email_config.pk}); skipping this run'
        )
        continue
    if not still_eligible:
        ...mark STALE as before...
```

kobo/apps/mass_emails/tasks.py
```diff
@@ -296,11 +299,28 @@ def send_day_emails(self):
             records = MassEmailRecord.objects.filter(
                 status=EmailStatus.ENQUEUED,
                 email_job__email_config=email_config,
-            )[:limit]
+            )
```
platreth, line 302: I think here removing `[:limit]` can be a problem, because nothing replaced the cap no?

platreth (changes requested):

kobo/apps/mass_emails/tasks.py
noliveleger: something replaces the cap. Instead of doing at the PostgreSQL level, it's done in Python with [these lines](https://github.com/kobotoolbox/kpi/pull/7448/changes/BASE..26d73f42f3d536beff204b10e1ad6397ac05eae9#diff-7108b9ada285887566a56b06f688203b433d1202ec15f44b375d5a4421f1b973R310-R311) 

That's why I'm using also `.iterator()` to avoid loading objects in memory. Django will use a DB cursor. ~~Having that said, I could specify chunk size with `limit`  (e.g.: `.iterator(chunk=limit)`). We want to load _limit_ records at a time maximum.~~ 

EDIT: The limit is usually higher than Django default limit. It's useless.

kobo/apps/mass_emails/tasks.py
noliveleger: IMO, too complicated! `record.user` could be None and passed to `is_user_still_eligible` a

Inside `is_user_still_eligible` the diff would be 

```
    def is_user_still_eligible(self, user: User | None) -> bool:
        """
        Cheap single-user recheck of `self.query`'s criteria, used before
        sending a stale enqueued record to catch a recipient who stopped
        matching while the record was sitting in the queue.
        """

       # Get rid of `is_user_active_and_not_trashed`  or move the code inside it  and call it here. 
       
       if user is None or not user.is_active:
           return False
       elif hasattr(user, 'trash'):
           return False
             
        check = USER_ELIGIBILITY_CHECKS.get(self.query)
        ...
```

(this to be done in #7445 after the merge of this one)

platreth (approved):

PR #7452, opened 2026-08-19
perf(massEmails): fewer per-record queries when sending DEV-2691

noliveleger:
### 📣 Summary

Sending mass emails now does fewer database queries per recipient, and no longer risks crashing if a recipient's account is deleted mid-send.

### 📖 Description

Sending a batch of mass emails looks up each recipient's profile details and their organization's plan name separately for every single email, most of which doesn't need to be a fresh, individual lookup. This change batches those lookups together where it's safe to do so, while still reading a recipient's active status fresh right before sending each email — a large batch can take close to an hour to fully send, so an account status still needs to be checked close to send time, not just once at the start. It also makes sure a recipient whose account was deleted partway through a send doesn't cause the sending process to fail.

### 💭 Notes

- `select_related('user', 'user__extra_details').defer('user__is_active')` on the enqueued-records query in `send_day_emails()` (`kobo/apps/mass_emails/tasks.py`). `is_active` stays deferred so it's still read fresh at per-record processing time, matching the freshness guarantee added for DEV-2691 (a run can span most of an hour under the DEV-2681 pacer, long enough for a recipient to be deactivated mid-run).
- Measured against mass-email configs before merging: consistent ~2 queries/record → ~1 query/record and ~2x wall-clock reduction on this slice of the send loop.
- `get_plan_name()` (`kobo/apps/stripe/utils/subscription_limits.py`) now takes an `Organization` directly instead of an `OrganizationUser`, since it only ever read `.organization.id` off the latter. Removes a redundant per-record `OrganizationUser` query in `send_email()`.
- `_is_user_active_and_not_trashed()` (`kobo/apps/mass_emails/user_queries.py`) now catches `User.DoesNotExist`: with `is_active` deferred, a recipient deleted between the main query running and that specific record being processed would otherwise raise instead of being treated as ineligible. Covered by a new unit test that simulates the deferred-field-on-a-deleted-row scenario directly.

### DIFF
--- a/kobo/apps/mass_emails/tasks.py
+++ b/kobo/apps/mass_emails/tasks.py
@@ -306,10 +306,14 @@ class MassEmailSender:
             limit = self.limits.get(email_config.id)
             if not limit:
                 continue
+            # `user__is_active` stays deferred so it's re-fetched fresh right
+            # when a record is processed, not batched with the rest of user
+            # at query time: a run can span close to an hour, long enough
+            # for a recipient to be deactivated in between.
             records = MassEmailRecord.objects.filter(
                 status=EmailStatus.ENQUEUED,
                 email_job__email_config=email_config,
-            )
+            ).select_related('user', 'user__extra_details').defer('user__is_active')
             logging.info(
                 f'Processing up to {limit} records for MassEmailConfig({email_config})'
             )
--- a/kobo/apps/mass_emails/tasks.py
+++ b/kobo/apps/mass_emails/tasks.py
@@ -324,11 +324,9 @@ class MassEmailSender:
             for record in records.iterator():
                 if budget_used >= limit:
                     break
-                # A user deactivated, deleted or trashed after enqueue must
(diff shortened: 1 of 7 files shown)


kobo/apps/mass_emails/tests/test_eligibility_checks.py
```diff
@@ -105,18 +106,30 @@ def test_false_when_stripe_disabled(self):
             )
         assert result is False
 
-    @override_settings(STRIPE_ENABLED=True)
-    def test_false_for_inactive_user_without_hitting_the_usage_calculator(self):
```
noliveleger, line 109: removed because it was useless. "is_user_within_usage_range" does not use "_is_user_active_and_not_trashed" anymore. It's provided by "is_user_still_eligible"

kobo/apps/mass_emails/tasks.py
platreth: Not in the PR but not true anymore :)

platreth (approved):

PR #7454, opened 2026-08-19
feat(sso): expose managed sso fields in /me endpoint DEV-2408

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Include additional SSO properties `managed` and `managed_domains` for social accounts the `/me` endpoint

### 💭 Notes
There is no direct link between SocialAccounts and SocialApps. The only think we have to connect a SocialAccount to a SocialApp is the provider, which is not a ForeignKey and not actually unique. In practice, however, it is very unlikely for there to be an installation with >1 SSO with the same provider, so we just mirror the logic in allauth's `get_app`, but don't raise an exception for the rare case that there are >1 or 0 apps from the given provider. Instead we just assume that the SSO is not managed and has no associated domains.

### 👀 Preview steps
Make sure SSO is enabled.

1. ℹ️ have an admin user and a user with an SSO account.
2. In Django admin, create a SocialAppCustomData object for the SSO provider with managed=True and the sso user's email domain.
3. Login as the sso user.
4. Navigate to `/me`
5. 🟢 [on PR] notice the response contains `managed: True` and `managed_domains: ['<domain>']` in the `'social_accounts'` field.

### DIFF
--- a/kobo/apps/accounts/tests/test_current_user.py
+++ b/kobo/apps/accounts/tests/test_current_user.py
@@ -1,5 +1,6 @@
 import re
 from datetime import timedelta
+from unittest.mock import patch
 
 import dateutil
 from constance.test import override_config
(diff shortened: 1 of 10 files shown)


kobo/apps/accounts/tests/test_current_user.py
```diff
@@ -224,3 +214,44 @@ def test_validate_extra_detail_no_organization_type(self):
...
+        app = baker.make('socialaccount.SocialApp', provider='provider')
+        app2 = baker.make('socialaccount.SocialApp', provider='provider')
+        baker.make('socialaccount.SocialApp', provider='provider')
+        SocialAppCustomData.objects.create(social_app=app)
+        SocialAppCustomData.objects.create(social_app=app2)
+        baker.make('socialaccount.SocialAccount', user=self.user, provider='provider')
+        response = self.client.get(self.url)
+        assert response.status_code == status.HTTP_200_OK
+        mock_logging.assert_called_once_with(
+            'Multiple social apps returned for provider provider,'
```
noliveleger, line 255: very nitpicky: I thought it was a typo at the first glance. What about changing the provider name to `my_provider` or use backticks in the message. 

Nothing required AT ALL!!! just me being very annoying.

PR #7456, opened 2026-08-19
fix(mailer): classify SES per-session message limit DEV-2675

noliveleger:
### 📣 Summary

A mass email recipient whose send happened to hit SES's per-session message limit was never getting emailed, silently, with no retry.

### 📖 Description

SES closes the SMTP connection after it's carried a certain number of messages in one session — a normal connection-lifecycle event, not a problem with the recipient or the message. The code already reconnects correctly when this happens, but it didn't recognize this specific SES response, so it fell back to treating it like any other send failure: the recipient's record was marked permanently failed and never retried, even though nothing was actually wrong with their address.

### 💭 Notes

- New `MailerConnectionSessionLimitError` (`kpi/exceptions.py`), deliberately not a subclass of `MailerProviderThrottledError`: that hierarchy exists for account-wide throttling (rate limit, daily quota), where stopping the whole send run makes sense. This is per-connection only, and the connection is already reconnected by the time it's raised, so the run should just skip this one record and keep going.
- New signature in `Mailer.THROTTLE_SIGNATURES` (`kpi/utils/mailer.py`): `421` + `"Maximum message count per session reached"` → `MailerConnectionSessionLimitError`.
- `send_day_emails()` (`kobo/apps/mass_emails/tasks.py`) now catches this specifically around the send call: logs a warning, leaves the record `ENQUEUED` for a later run, and moves on to the next record without spending any of the day's send budget on it — same treatment as a stale-skipped record.
- Found while testing the DEV-2681/DEV-2682 mass email pacing work against the real AWS SES mailbox simulator.

### 👀 Preview steps

1. ℹ️ have a mass email record in `ENQUEUED` status
2. mock `Mailer.send()` to raise `MailerConnectionSessionLimitError` for that record, followed by a normal send for the next one
3. run `MassEmailSender().send_day_emails()`
4. 🔴 [on main] the record is marked `FAILED` and never retried
5. 🟢 [on PR] the record stays `ENQUEUED`, and the run continues to the next record instead of stopping

### DIFF
--- a/kobo/apps/mass_emails/tasks.py
+++ b/kobo/apps/mass_emails/tasks.py
@@ -23,6 +23,7 @@ from kobo.apps.mass_emails.models import (
 from kobo.apps.organizations.models import Organization
 from kobo.celery import celery_app
 from kpi.exceptions import (
+    MailerConnectionSessionLimitError,
     MailerError,
     MailerProviderQuotaExhaustedError,
     MailerProviderRateThrottledError,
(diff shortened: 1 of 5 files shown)


rajpatel24 (approved):
LGTM

PR #7458, opened 2026-08-20
feat(subsequences): run automatic qualitative analysis on text questions DEV-2291

platreth:
### 📣 Summary

Automatic qualitative analysis and translation now run on text questions, without needing a transcript first.

### 📖 Description

Both features used to read their input from a transcript, so they only worked on audio and video questions. A text question has nothing to transcribe, so they were unavailable there.

They now read the answer straight from the submission when the question is a text question. Audio and video keep working as before, through the transcript. Enabling an action on a question type that doesn't support it, like transcription on a text question, is now refused up front instead of failing later.

Nobody declares what language a typed answer is in, so translation uses the form's own language when it knows it, and otherwise lets the translation service detect it. Very long answers are the exception: they go through a batch API that cannot detect a language, so those ask for the form language to be set.

### 💭 Notes

- Text sources get a synthetic dependency (`_actionId: 'submission'`, `_uuid: <submission uuid>`) in place of a transcript reference. The sanitizer already keeps only `_actionId` and `_uuid` before saving, so no result schema changed.
- The new serializer check reads `allowed_source_types` off each action class, which reuses the `*_SOURCE_TYPES` constants that were already sitting unused in `constants.py`.
- Prompts are now stored per source question type (transcript or text). Same text in both for now, but product wants to be able to tune them separately, and images are coming.
- The source language for a submission-sourced translation is a hint, not a requirement: the form's `default_language`, falling back to its default translation, and resolved against `TranslationService` rather than `TranscriptionService` (nothing was transcribed here). Anything unusable means no `source_language_code` in the request, which is how Google is asked to detect it. The batch path has no detection, so over `MAX_SYNC_CHARS` it fails with a message naming the fix.
- Resolving a question type means reading `Asset.content`, which the bulk and polling tasks deliberately defer. It is cached per deployed version, like `Asset.get_all_attachment_xpaths` does, so a bulk run reads content once per form version instead of once per submission. `test_deployed_asset_content_is_read_only_once` locks that in.

### 👀 Preview steps

1. ℹ️ have a project with a plain **text** question and at least one submission
2. add a qualitative analysis question on it and run automatic analysis
3. 🔴 [on main] it fails with "No accepted transcription found. Accept the transcription before running translation.", and there is no way to get a transcript on a text question
4. 🟢 [on PR] the analysis completes, based on the submitted text
5. run automatic translation on the same text question
6. 🟢 [on PR] it translates the answer, whether or not the form has a named language
7. 🟢 an audio question is unchanged: analysis and translation still want an accepted transcript first


platreth:
@greptileai

kobo/apps/subsequences/actions/mixins.py
```diff
@@ -2,48 +2,62 @@
...
         latest_deletion_dt = None
 
         if 'value' in action_data and action_data['value'] is None:
             return action_data
 
+        if self.get_source_question_type() == TEXT_SOURCE_TYPE:
+            # Direct-text source: point the dependency at the submission. The
+            # `_question_type` key is in-flight only; the sanitizer in
+            # `revise_data` strips it before persistence.
+            raw_uuid = submission.get(SUBMISSION_UUID_FIELD) or submission['_uuid']
```
rgraber: It's been a minute, so this may be a silly question: is this supposed to be `rootUuid`? or by this point do we know '_uuid' will match the rootUuid?

kobo/apps/subsequences/integrations/tests/test_google_translate.py
```diff
@@ -404,3 +423,135 @@ def test_sync_translation_raises_quota_error_when_quota_is_exhausted(self):
...
+                    'q1',
+                    {
+                        'language': 'fr',
+                        '_dependency': {'_actionId': 'submission'},
+                    },
+                )
+
+        assert response == {'status': 'complete', 'value': 'Bonjour'}
+        # Source language was extracted from the 'English (en)' form language,
+        # and resolved against the translation service, not the ASR one
```
rgraber: I don't understand this comment (specifically the phrase "resolved against the translation service")

kobo/apps/subsequences/integrations/tests/test_google_translate.py
```diff
@@ -404,3 +423,135 @@ def test_sync_translation_raises_quota_error_when_quota_is_exhausted(self):
...
+
+    def test_form_language_falls_back_to_the_default_translation(self):
+        service = self._build_service()
+
+        for content, expected in (
+            ({'settings': {'default_language': 'English (en)'}}, 'en'),
+            ({'settings': {}, 'translations': ['Deutsch (de)']}, 'de'),
+            ({'settings': {}, 'translations': [None]}, None),
+        ):
+            service.asset.content = content
```
rgraber, line 557: How does this test that we use the default if we keep wiping the content? Doesn't that mean it's still only looking at one language at a time rather than a form with a default language as well as translations?

kobo/apps/subsequences/tests/test_utils.py
```diff
@@ -0,0 +1,67 @@
...
+        # This runs on every supplemental revision, including one celery task
+        # per submission for bulk actions, so it must not refetch the deferred
+        # `content` field every time
+        xpath = 'settings_fixture_q1'
+        get_survey_question_type(self.asset, xpath)
+
+        deferred_asset = Asset.objects.only('pk', 'uid').get(pk=self.asset.pk)
+        with self.assertNumQueries(1):
+            # The single query resolves the deployed version for the cache key
+            question_type = get_survey_question_type(deferred_asset, xpath)
```
rgraber: do we need to call this twice to make sure it doesn't make an additional query?

kobo/apps/subsequences/utils/__init__.py
rgraber, line 1: Why are we using an __init__ file instead of a regular utils.py?

rgraber (commented):
Couple questions

kobo/apps/subsequences/actions/mixins.py
platreth: Good catch, `rootUuid` is the right one 🤦‍♂️

kobo/apps/subsequences/integrations/tests/test_google_translate.py
platreth: Yeah, that comment is bad. There are two language tables: 
- `TranscriptionService` => what Google can transcribe
-  `TranslationService` => what it can translate. 

Source language used to be resolved with the transcription one since it always came from a transcript. Typed text never went through ASR, so it uses the translation table now, otherwise we'd reject languages Google translates fine but can't transcribe. I changed the comment!

kobo/apps/subsequences/integrations/tests/test_google_translate.py
platreth: Yup, added one with default_language: 'English (en)' and translations: ['Deutsch (de)']

kobo/apps/subsequences/tests/test_utils.py
platreth: It does call twice already: `get_survey_question_type` then `get_form_language` inside the same `assertNumQueries(1)`

kobo/apps/subsequences/utils/__init__.py
platreth: `utils/` is already a package (supplement_data.py, time.py, versioning.py), so utils.py would mean moving all of those!
I moved it to utils/survey.py thought!

rgraber (approved):
LGTM

PR #7462, opened 2026-08-20
refactor(frontend): replace react-select with Mantine Select DEV-2652

magicznyleszek:
### 📣 Summary

A few dropdowns are now unified to look like all the other dropdowns.

### 💭 Notes

Changes here:

- `react-select` is gone (and some style overrides too)
- Adjusted few components to handle strings not objects (`react-select` passed whole `{value, label}` objects around)
- `ProjectSettingsFields` now holds plain values, but the asset settings still store `LabelValuePair`s → `getSettingsForEndpoint` pairs the values back up with their labels with 5 new tests cover that round-trip
- `renderInBackbone.js` — the form builder mounts `KoboMatrix` into its own React root, which the app's `MantineProvider` never reaches, so that root now gets a provider of its own
- new `Pill` variants

### 👀 Preview steps

1. ℹ️ have a deployed project with a few submissions (at least one validated, at least one photo), plus a second deployed project to import data from; also enable Operational purpose and PII in admin
2. go to **Project → Settings → General**
3. 🟢 notice **Sector**, **Country**, **Operational purpose** and the **PII** dropdown match the other Kobo dropdowns, filter as you type, and clear with the ×
4. 🔴 [on main] notice they open **upwards** and have their own slightly different look
5. set all four, **Save**, reload the page — 🟢 notice every value was stored correctly
6. go to **Projects** list → **Filter** → **Sector** → pick the sector you just saved — 🟢 notice the project is listed, then sort the list by **Sector** — 🟢 notice sorting still works
7. go to **Projects → New → Build from scratch** — 🟢 notice the same four dropdowns behave the same in the **Project details** step
8. go to **Project → Settings → Connect Projects** → open the project dropdown — 🟢 notice it lists your other project, shows a spinner while loading, and says "No projects to connect" when nothing matches your search
9. go to **Project → Data → Table** → open the **Validation** column header filter — 🟢 notice the coloured pills in the list, and that picking one filters the rows
10. in the same table, open the **Validation** dropdown inside a row — 🟢 notice the selected status shows as a coloured pill
11. go to **Project → Data → Downloads** → **Select export type** — 🟢 notice it switches type as before
12. click **Advanced options** — 🟢 notice **Export Select Many questions as…** and **Value and header format** both work
13. save a custom export (tick **Save selection as…**, give it a name, **Export**), then use **Apply saved export settings** — 🟢 notice picking your saved export fills the form back in, and picking **None** resets it to defaults
14. go to **Project → Data → Gallery** → open the **From** dropdown — 🟢 notice **All questions** is preselected and picking a question filters the photos
15. go to **Project → Form → Edit** (form builder) → **Layout & Settings** → 🟢 notice the **Form style** dropdown still switches styles
16. in the same panel under **Metadata**, enable **Enable audio recording in the background** → 🟢 notice the **Audio quality** dropdown appears and changes value
17. still in the form builder, add a **Question Matrix** question, expand a column and open its **Response Type** dropdown
18. 🟢 notice the dropdown renders at the same width as the text fields next to it and changes the column type (with the usual warning when switching away from Select One / Select Many)
19. You're done, thanks <3


jamesrkiger:
Anything but a short, one-word option in the country selector here expands the select component vertically:
image

In the old version you can usually squeeze in two countries before this happens. Can we try to make these pills more compact? The 'x' and chevron icons also look awkwardly large to me.

jamesrkiger:
Admittedly, the question matrix select already didn't match its neighboring inputs before this PR. But it seems to be getting even more incongruous now:

image

Any chance you could make it more compact and add a border here?

jsapp/js/components/formGallery/formGallery.component.tsx
```diff
@@ -26,6 +26,12 @@ bem.GalleryFooter = makeBem(bem.Gallery, 'footer')
 
 const PAGE_SIZE = 20
 
+/**
+ * Stands for "no question filter". `Select` treats an empty string as no value
+ * at all, so the "all questions" option needs a sentinel of its own.
+ */
+const ALL_QUESTIONS_VALUE = '__show_me_all_questions_pretty_please__'
```
jamesrkiger: If our Select wrapper is making us do this, maybe we ought to change the select wrapper. Can we just make it so the wrapper doesn't coerce empty strings to null? Seems like it would be preferable (and more standard practice).

jamesrkiger (changes requested):

jamesrkiger (approved):
LGTM (aside from failing test)

PR #7467, opened 2026-08-21
fix(attachments): match media filenames across unicode normalization forms DEV-2686

platreth:
### 📣 Summary

Media files with accented characters in their name now upload and play back instead of vanishing from the data table.

### 📖 Description

A recording called something like `Guérisseur.ogg` would upload without any error, then land in the data table as plain text with no way to play or download it. Same thing on edits, and re-uploading never helped. Those files stay attached to their question now.

### 💭 Notes

- The filename reaches us twice: as text in the submission XML, and as the name of the uploaded file. Depending on where the file came from (macOS especially), one side is NFD and the other NFC. Every match on that name was byte-exact, so they never lined up.
- Two things broke as a result. `get_soft_deleted_attachments()` didn't find the new upload in the XML basename set and soft-deleted it right after saving, and `question_xpath` came back `''`, so the UI had no question to hang the file on.
- Fix normalizes to NFC when the attachment comes in (basename plus the stored file name) and on both sides of every lookup. `get_attachment()` also accepts NFD so rows already in the DB still resolve.
- One trap: `get_valid_filename()`'s `\w` doesn't match combining marks, so an NFD name quietly loses its accent on the storage path. That's why `f.name` gets normalized too and not just the basename.
- No migration, no backfill. Already-broken submissions stay broken until someone edits or re-submits them, which now saves correctly.

### 👀 Preview steps

1. ℹ️ have a project with an audio (or any file) question, deployed
2. make a file with an NFD-encoded name. On macOS: `python3 -c "import unicodedata,shutil;shutil.copy('sound.m4a','/tmp/'+unicodedata.normalize('NFD','Guérisseur.m4a'))"`
3. open the Enketo form, attach that file, submit
4. open the project's data table
5. 🔴 [on main] the cell shows the filename as text, no player and no download. The attachment got soft-deleted on save
6. 🟢 [on PR] the player shows up and the file plays
7. edit the submission and save again
8. 🟢 file is still there

### DIFF
--- a/kpi/utils/submission.py
+++ b/kpi/utils/submission.py
@@ -41,7 +41,13 @@ def get_attachment_filenames_and_xpaths(
         else:
             if key in attachment_xpaths:
                 try:
-                    value = default_kobocat_storage.get_valid_name(normalize_nfc(value))
+                    # Attachments saved before basenames were normalized are
+                    # stored under the raw name, which `get_valid_name()` strips
+                    # of its combining marks. Key both forms so they resolve.
+                    raw_valid_name = default_kobocat_storage.get_valid_name(value)
+                    nfc_valid_name = default_kobocat_storage.get_valid_name(
+                        normalize_nfc(value)
+                    )
                 except SuspiciousFileOperation:
                     logging.error(f'Could not get valid name from {value}')
                     continue
@@ -56,6 +62,8 @@ def get_attachment_filenames_and_xpaths(
                         group = group_name.split('/')[-1]
                         key = key.replace(group, f'{group}[{group_index}]')
 
-                return_dict[value] = key
+                return_dict[nfc_valid_name] = key
+                if raw_valid_name != nfc_valid_name:
+                    return_dict[raw_valid_name] = key
 
     return return_dict
--- a/kpi/utils/submission.py
+++ b/kpi/utils/submission.py
@@ -3,6 +3,7 @@ from collections import defaultdict
 from django.core.exceptions import SuspiciousFileOperation
 
 from kpi.deployment_backends.kc_access.storage import default_kobocat_storage
+from kpi.utils.files import normalize_nfc
 from kpi.utils.log import logging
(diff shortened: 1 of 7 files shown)


platreth:
@greptileai

platreth:
@greptileai recheck and update score

kpi/utils/submission.py
```diff
@@ -40,7 +41,7 @@ def get_attachment_filenames_and_xpaths(
         else:
             if key in attachment_xpaths:
                 try:
-                    value = default_kobocat_storage.get_valid_name(value)
+                    value = default_kobocat_storage.get_valid_name(normalize_nfc(value))
```
rajpatel24: I think this might break `question_xpath` for some existing rows, wanted to check with you.

The key is now built from the NFC form, but the lookup in `base_backend.py:934` reads the stored path. On S3 that's fine (`get_valid_name()` there never strips anything), but on filesystem storage `get_valid_filename()'s` `\w` drops the combining mark, so an older NFD upload sits on disk as Guerisseur.jpg.

So for an old submission where both the XML value and the filename were NFD (consistent client, never hit the soft-delete bug):

- on main: key Guerisseur.jpg vs lookup Guerisseur.jpg → matched
- with this PR: key Guérisseur.jpg vs lookup Guerisseur.jpg → ''

`_without_suffix()` wouldn't help since the accent is gone rather than decomposed.

Would keying both forms here (`get_valid_name(value)` alongside `get_valid_name(normalize_nfc(value))`) be enough to stay backward compatible? Might be that those rows don't really exist in practice. What do you think?

rajpatel24 (changes requested):

kpi/utils/submission.py
platreth: That's a good point, thanks for catching this! I will fix it

rajpatel24 (approved):
LGTM

PR #7470, opened 2026-08-22
feat(auth): read-only REGISTRATION_SSO_ALLOWED_EMAIL_DOMAINS aggregate in Constance DEV-2410

Guitlle:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Added a display-only constance setting `REGISTRATION_SSO_MANAGED_EMAIL_DOMAINS` that displays a list of all email domains configured across managed SocialApps. It is located under General Options next to the other `REGISTRATION_*` settings in the Constance Config screen

### 📖 Description
Registered `REGISTRATION_SSO_MANAGED_EMAIL_DOMAINS` setting in `CONSTANCE_CONFIG` with field type `disabled_textarea` and a note explaining per-app management in Admin > SocialApp. The `disabled_textarea` field type was added to `CONSTANCE_ADDITIONAL_FIELDS`. The signal ` update_managed_sso_email_domains()` helper function aggregates email domains from managed SocialApp objects (social_app__managed=True) when they are modified and stores them in the new read only setting in constance.
I've also created a data migration to initialize `REGISTRATION_SSO_MANAGED_EMAIL_DOMAINS` for existing database deployments.

### 👀 Preview Steps

1. ℹ️ Log in with a superuser / admin account.
2. Navigate to Django Admin > Constance > Config
3. 🟢 Confirm `REGISTRATION_SSO_MANAGED_EMAIL_DOMAINS` is placed directly after `REGISTRATION_BLACKLIST_ERROR_MESSAGE`
4. Navigate to Account Extras > Social app custom datas
6. Toggle `Managed` for a SocialApp and add an email domain
7. Return to Django Admin > Constance > Config
8. 🟢 Confirm `REGISTRATION_SSO_MANAGED_EMAIL_DOMAINS` lists the new domain in read-only format


kobo/apps/accounts/signals.py
```diff
@@ -51,3 +56,33 @@ def update_email(*args, **kwargs):
...
+                .order_by('domain')
+            )
+            domain_string = '\n'.join(domains)
+            setattr(
+                constance.config,
+                'REGISTRATION_SSO_MANAGED_EMAIL_DOMAINS',
+                domain_string,
+            )
+    except Exception:
+        pass
```
platreth: I think we should either drop the try / except or log, but pass feels weird. 

Also `transaction.atomic():` is useless here!

platreth (commented):
PR title says "REGISTRATION_SSO_ALLOWED_EMAIL_DOMAINS" but in the PR it is "egistration_sso_managed_email_domains"

platreth (changes requested):

kobo/apps/accounts/signals.py
Guitlle: I think there are no potential exceptions, so I removed the try/pass

platreth (approved):

PR #7473, opened 2026-08-24
fix(xlsImport): block translated XLSForm missing language-specific columns DEV-2656

platreth:
### 📣 Summary

Uploading an XLSForm that has translations but leaves some columns untranslated (like a plain `hint` next to `label::English (en)`) is now blocked with a clear error instead of creating a broken project.

### 📖 Description

Until now, importing such a form quietly added an "Unnamed language" to the project, which could break the form builder, or silently dropped the missing translations. The import now fails right away with a message naming the column, e.g. "The `hint` column is not translated". Forms with no translations at all import as before.

### 💭 Notes

- New `ImportTask._ensure_translated_columns()` runs on both import routes (base64 and URL) before any asset is created or modified. It mirrors formpack's `_get_special_survey_cols()` rules, so it blocks exactly the forms that would end up with a null translation: bare `label`/`hint`, bare media columns, or a column present both bare and with a `::language` suffix.
- This also blocks forms that used to import thanks to the save-time repair from DEV-2095 (bare `hint` + `default_language`): that repair silently dropped the hint's other translations, which is the other half of the bug report.
- Second commit makes the import-error toast readable (it rendered in unstyled monospace); applies to all import errors, and the same fix goes to the duplicated toast in the library upload modal.
- `test_import_xls_with_default_language_not_in_translations` used fixture asset pk=2, which itself mixes bare `label` with `label::English` and is now rejected; rebuilt its input so it still covers the default-language mismatch.

- The column and fuzzy-version regexes now come from `formpack.constants` (kobotoolbox/formpack#351) instead of local copies, per review feedback; the duplicates in `kobo/apps/reports/constants.py` are gone too. Depends on kobotoolbox/formpack#351 (now merged); the formpack pin here points at its merge commit.

### 👀 Preview steps

1. ℹ️ have an account
2. make an XLSForm whose survey sheet has `label::English (en)`, `label::French (fr)` and a plain `hint` column (one text question is enough)
3. upload it as a new project (NEW → Upload an XLSForm)
4. 🔴 [on main] the import succeeds and the project gets an Unnamed language (or silently loses the French hint when `default_language` is set)
5. 🟢 [on PR] the import fails with "The `hint` column is not translated" and no project is created
6. rename `hint` to `hint::English (en)`, add `hint::French (fr)`, upload again
7. 🟢 the project imports normally with both languages

### DIFF
--- a/kpi/models/import_export_task.py
+++ b/kpi/models/import_export_task.py
@@ -4,7 +4,7 @@ import os
 import posixpath
 import re
 import tempfile
-from collections import defaultdict
+from collections import OrderedDict, defaultdict
 from io import BytesIO
 from os.path import split, splitext
 from typing import Dict, Generator, List, Optional, Tuple
@@ -22,7 +22,7 @@ from django.db.models.functions import Cast, Coalesce, Concat
 from django.db.models.query import QuerySet
 from django.utils import timezone
 from django.utils.translation import gettext as t
-from formpack.constants import KOBO_LOCK_SHEET
+from formpack.constants import KOBO_LOCK_SHEET, MEDIA_COLUMN_NAMES
 from formpack.schema.fields import (
     IdCopyField,
     NotesCopyField,
@@ -371,6 +371,7 @@ class ImportTask(ImportExportTask):
                 except InvalidFileException:
                     kontent = xls_to_dict(item.readable)
                 self._ensure_valid_node_names(kontent)
+                self._ensure_translated_columns(kontent)
 
                 if not destination:
                     extra_args['content'] = _strip_header_keys(kontent)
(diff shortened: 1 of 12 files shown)


platreth:
@greptileai

kpi/models/import_export_task.py
```diff
@@ -406,6 +407,60 @@ def _load_assets_from_url(self, url, messages, **kwargs):
...
+                column_shortname = mtch.groups()[0]
+                languages.add(mtch.groups()[1])
+                if column_shortname in uniq_cols:
+                    _mark_untranslated(column_shortname)
+
+        if languages and untranslated:
+            if len(untranslated) == 1:
+                raise ValueError(f'The `{untranslated[0]}` column is not translated')
+            cols = ', '.join(f'`{col}`' for col in untranslated)
+            raise ValueError(f'These columns are not translated: {cols}')
```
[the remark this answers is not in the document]
platreth: These messages are raised in a celery worker and stored in ImportTask.messages, so t() would translate them with the worker's locale at save time, not the uploader's + every errors like this are unlocalized so I kept it consistent

kpi/models/import_export_task.py
```diff
@@ -406,6 +407,60 @@ def _load_assets_from_url(self, url, messages, **kwargs):
...
+    @staticmethod
+    def _ensure_translated_columns(survey_dict):
+        """
+        Block importing an XLSForm that mixes translated columns (e.g.
+        `label::English (en)`) with untranslated ones (e.g. a bare `hint`),
+        which would otherwise inject a null "Unnamed language" translation and
+        break the formbuilder (DEV-2656). Mirrors the column classification in
+        `formpack.utils.expand_content._get_special_survey_cols()`.
+        """
+        media_re = '|'.join(MEDIA_COLUMN_NAMES)
```
rgraber: I realize this will require an extra PR for formpack, but can we extract things like this regex and the other column-name regexes into the formpack constants file? it's confusing and risky to constantly copy paste them.

jamesrkiger (commented):
Frontend changes look good to me. Thanks for catching that.

kpi/models/import_export_task.py
```diff
@@ -406,6 +407,60 @@ def _load_assets_from_url(self, url, messages, **kwargs):
...
+            if column_name.startswith('bind:') or column_name.startswith('body:'):
+                continue
+            mtch = re.match(
+                rf'^(media\s*::?\s*)?({media_re})\s*::?\s*([^:]+)$',
+                column_name,
+            )
+            if mtch:
+                languages.add(mtch.groups()[2])
+                continue
+            mtch = re.match(rf'^(media\s*::?\s*)?({media_re})$', column_name)
```
rgraber: some comments here with examples of what we're trying to match would be helpful

rgraber (approved):
LGTM

PR #7477, opened 2026-08-25
ci(releases): changelog job shouldn't be concurrent DEV-2722

Akuukis:
### 💭 Notes

Fix the changelog job between stabilize and tag workflows, by pushing to different branches.

There was an cascading problem: when tag workflow pushes pulled translations,
it triggered stabilize workflow that then overwrites changelog.

There was a similar race-condition problem: while tag workflow runs and a untimely
push triggers stabilize workflow, stabilize's changelog job may overwrite changelog.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -50,8 +50,11 @@ jobs:
     - name: Commit to dedicated branch
       env:
         VERSION: ${{ inputs.version }}
+        DRAFT: ${{ inputs.draft }}
       run: |
-        git checkout -B changelog/$VERSION
+        BRANCH="changelog/$VERSION"
+        if [[ "$DRAFT" == "true" ]]; then BRANCH="$BRANCH-draft"; fi
+        git checkout -B "$BRANCH"
         git add -f CHANGELOG.md
         git commit -m "chore(releases): generate CHANGELOG.md for $VERSION"
-        git push -f --set-upstream origin changelog/$VERSION
+        git push -f --set-upstream origin "$BRANCH"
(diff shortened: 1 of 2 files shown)


.github/workflows/release-3-tag.yml
```diff
@@ -2,17 +2,17 @@ name: 'CD: tag'
 
 
 concurrency:
-  group: release-tag
+  group: release-shared-refs/heads/${{ inputs.release_branch }}
   cancel-in-progress: false
```
noliveleger, line 6: I saw that the value of `cancel-in-progress` was different from other workflows (i.e. `release-2-stabilize.yml`) and now the group is the same.

`github.ref` == `refs/heads/release/2.026.30`  in `stabilize.yml` so it would be equal to `release-shared-refs/heads/${{ inputs.release_branch }}` in `release-3-tag.yml` when the input is `release/2.026.30`. 

It's hard to test the whole mechanism locally, so asked Claude for help to analyze my theory 

Claude notes: 

--- 
`cancel-in-progress` is asymmetric across the shared group: `false` on tag, `true` on stabilize and delete. GitHub reads that flag from the *newly queued* run, so a stabilize run queued while a tag run is in progress will cancel the tag run.

That is not hypothetical, because the tag workflow pushes to the release branch itself. The `pull-transifex` job commits `chore: pull transifex translations for $VERSION` using the KOBO_BOT app token, and app-token pushes do trigger `on: push` workflows. 

From the Aug 20 history:

* tag run 32388587088: `pull-transifex` ends 15:58:50 (pushes the commit), `tag` job starts 16:01:30
* stabilize run 32389453740 / 32389342486 `chore: pull transifex translations for 2.026.30b`: created 15:58:50, same branch, same future group

With the shared group, that tag run would have been cancelled at 15:58:50, before the tag was pushed and before the image was built. The `Rollback: delete tag on failure` step is `if: failure()`, so it does not run on cancellation either. This would fire on every release where translations changed.

Smaller variant of the same thing: while the tag run sits pending, any new push to the release branch cancels it, since GitHub keeps only one pending run per group. The dispatch just disappears.

--- 

This may be a race condition and may never happen but it's worth a consideration, don't you think?

Akuukis: Good point!

I need to think about this, looks like there's a problem either way.

Akuukis:
@greptileai rereview, it's a totally different solution now

Akuukis:
@greptileai rereview, it's a totally different solution again

noliveleger (approved):
LGTM

PR #7483, opened 2026-08-25
feat(sso): restrict account management of sso-managed accounts DEV-2356

rgraber:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Restrict account management for users with SSO-managed accounts.

### 📖 Description
Users with SSO-managed accounts are forbidden from setting or changing a password or unlinking their account. An admin cannot add another social account or set a password either, but they may unlink.

### 💭 Notes
*  It's already the case that a user with one SSO account cannot add another, so this PR ignores that part of the issue.

* determine if a user has a managed social account by linking `SocialAccount.provider` to the `SocialApp.provider_id`. It is important to note that `SocialApp.provider` and `SocialApp.provider_id` are both strings and refer to different properties. The one that ends up in `SocialAccounts` is `provider_id`, but it gets saved on the `SocialAccount` as `provider`.

* We handle password and SSO un/linking in a number of different places which is why there are so many different parts to the PR. In an ideal world we would probably handle unlinking social accounts via adapter.validate_disconnect() rather than deleting `SocialAccount` objects but that would be a heavier lift.

* makes the `provider` field for `SocialAccounts` in admin read-only once set to avoid weird results from changing a provider via admin rather than through the SSO flow.

* some normalization on saving a `SocialAppManagedDomain`

### 👀 Preview steps
Make sure SSO is enabled

1. ℹ️ have an admin account
2. Create a `SocialAppCustomData` object for your social app with `managed=True` and add a usable email domain (`kobotoolbox.org` if using Kobo Google Apps)
3. Register and login as a new user via the managed SSO
4. Navigate to Account Settings > Security
5. Attempt to set a password via Forgot my password
6. 🟢 [on PR] Forbidden
7. Attempt to unlink the SSO
8. 🟢 [on PR] Forbidden
9. Login as admin
10. In Django admin, open the SSO user for editing
11. 🟢 [on PR] Notice there is no place to set a password
12. In Django admin, attempt to create another SocialAccount for the user
13. 🟢 [on PR] Forbidden
14. In Django admin, delete the user's SSO account
15. 🟢 [on PR] Validate you are still able to do this

### DIFF
--- a/hub/admin/extend_user.py
+++ b/hub/admin/extend_user.py
@@ -222,7 +222,7 @@ class ExtendedUserAdmin(AdvancedSearchMixin, UserAdmin):
         css = {'all': ('admin/css/inline_as_fieldset.css',)}
 
     def get_fieldsets(self, request: HttpRequest, obj=...):
-        fieldsets = super().get_fieldsets(request, obj)
+        fieldsets = [fieldset for fieldset in super().get_fieldsets(request, obj)]
         if obj and obj.pk is not None:
             if user_is_managed_by_sso(obj):
                 fieldsets[0][1]['fields'] = ('username',)
--- a/hub/admin/extend_user.py
+++ b/hub/admin/extend_user.py
@@ -18,7 +18,7 @@ from django.utils import timezone
 from django.utils.safestring import mark_safe
 
 from kobo.apps.accounts.mfa.models import MfaMethodsWrapper
-from kobo.apps.accounts.models import SocialAppManagedDomain
+from kobo.apps.accounts.utils import user_is_managed_by_sso
 from kobo.apps.accounts.validators import (
     USERNAME_INVALID_MESSAGE,
     USERNAME_MAX_LENGTH,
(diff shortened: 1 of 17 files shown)


rgraber:
@greptileai ignore the checklist just check the code

rgraber:
@greptileai

rgraber:
@greptileai

jamesrkiger:
Note to @Guitlle and anyone else reviewing this PR, the frontend testing for this is going to get more difficult after #7498 gets merged. You could probably work around that modal by loading the frontend security page while the SSO is not managed, then flipping it to managed in django admin in a separate window before pressing the actual button so the frontend doesn't know it needs to display the modal.

hub/admin/extend_user.py
```diff
@@ -219,6 +223,13 @@ class ExtendedUserAdmin(AdvancedSearchMixin, UserAdmin):
     class Media:
         css = {'all': ('admin/css/inline_as_fieldset.css',)}
 
+    def get_fieldsets(self, request: HttpRequest, obj=...):
```
Guitlle, line 226: What do you think of preferring a more resilient approach like filtering regardless of the fieldsets structure? Django could change the fieldsets and it would break it  

```python
def get_fieldsets(self, request, obj=None):
    fieldsets = copy.deepcopy(super().get_fieldsets(request, obj))
    if obj and obj.pk is not None:
        if user_is_managed_by_sso(obj):
            # Remove password fields from ALL fieldsets, not just [0]
            for fieldset_name, fieldset_options in fieldsets:
                fields = list(fieldset_options.get('fields', []))
                # Filter out password-related fields
                fields = [f for f in fields if f not in ('password', 'usable_password')]
                fieldset_options['fields'] = tuple(fields)
    return fieldsets
```

Guitlle (changes requested):
My only comment is about the get_fieldsets override, the rest  looks good to me and works as expected.

hub/admin/extend_user.py
rgraber: Good call!

Guitlle (approved):

PR #7485, opened 2026-08-25
fix(usage): simplify description text in usage page DEV-2699

duvld:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Remove the dismissable banner from the usage page and have a simple dynamic string explaining the usage page.

### 💭 Notes
Adds a `hasActivePlan` to `useBillingPeriod` hook to be destructured where needed. This is to meet the following requirments:
>Text dynamically displays "month", "year" or "billing period" depending on the user:
> - Free user (or Stripe not enabled) → display "month"
> - Member of a team/org whose owner is paying for an annual plan → display "year"
> - Any paid user (i.e. they dealt with Stripe) → show "billing period"

### 👀 Preview steps

1. ℹ️ have an multiple accounts and a project with stripe enabled
2. create a MMO on one account with another account as a member
3. buy any plan with an account not associated with the above two
4. buy an annual plan as the MMO owner for the org
5.  🟢 On the usage page for the MMO owner it should say "year"
6.  🟢 On the usage page for the MMO member it should say "year"
7.  🟢 On the usage page for a non-MMO user with any paid plan it should say "billing period"
8.  🟢 On the usage page for a free plan user it should say "month"
9.  🟢 Notice that other consumers of the `useBillingPeriod` hook still function as intended


jsapp/js/account/usage/useBillingPeriod.tsx
```diff
@@ -20,25 +27,30 @@ export async function getSubscriptionInterval() {
...
+    queryFn: getSubscriptionInfo,
   })
 
   return {
-    billingPeriod: billingPeriod || 'month',
-    // Default to 'month' if billingPeriod is undefined
+    // Default to 'month'/no plan while the query is still loading.
     // This ensures that the hook always returns a valid billing period
+    billingPeriod: data?.interval || 'month',
+    hasActivePlan: data?.hasActivePlan ?? false,
```
jamesrkiger, line 53: This was built as a very narrowly scoped hook so, while it doesn't hurt anything, I don't like returning this new, unrelated data on an ad hoc basis. But we also wouldn't want another tiny hook just for getting active subscription status. How about this: move this `billingPeriod === 'year' ? t('year') : hasActivePlan ? t('billing period') : t('month'),` into the hook instead and return it in lieu of hasActivePlan. So it would return `billingPeriod` and something like `intervalLabel` (make sure to be clear about what the label is for in the comments).

jamesrkiger (changes requested):

jsapp/js/account/usage/useBillingPeriod.tsx
duvld: @jamesrkiger sorry this slipped through the cracks. I tried implementing what I think you mean by adding a `intervalLabel` alongside the `billingPeriod`. `hasActivePlan` is still used to determine the `intervalLabel`, but it is in a now internal function and the `useBillingPeriod` hook still only returns strings. Let me know if this is what you had in mind

jamesrkiger (approved):
LGTM

PR #7486, opened 2026-08-25
fix(massEmails): disable one-off sends if remaining records are failed DEV-2729

rgraber:
### 🗒️ Checklist

1. [ ] run linter locally

### 📣 Summary
Fix a bug that was causing one-off email campaigns to be sent multiple times.

### 💭 Notes

The bug was caused by the following scenario:
1. one-off email config Config A is set to live
2. the next day, records are enqueued
3. a large email backlog causes enqueued records not to be sent 
4. after several days, the cleanup job marks all remaining enqueued records as 'failed'
5. the next time the sender runs, it sees Config A has no enqueued records and is still live, so it regenerates the campaign

The solution is to expire email configs (ie set `live=False`) immediately after we have marked remaining enqueued records as 'failed', similar to how we do in the sender where we set `live=False` if, after a mass send, we notice there are no more enqueued records.

### 👀 Preview steps
Turn off the kpi_worker while testing so you have control over when the various tasks run, otherwise you risk records being changed underneath you

1. ℹ️ have an admin account and a few others
2. Set the following values in Constance:
* MASS_EMAIL_ENQUEUED_RECORD_EXPIRY = 0
* MASS_EMAIL_TEST_EMAILS = `<newline-delineated list of user emails>`
* MASS_EMAIL_CONDENSE_SEND = True
6. Create an email config with frequency = -1 and query `test_emails` and set it to Live
7. Wait until the next 15m boundary
8. In a django shell, run `generate_mass_email_user_lists`
9. To simulate records getting old and being marked as failed, run `mark_old_enqueued_mass_email_record_as_failed`
10. To simulate it being the next day, clear the cache with `for key in cache.keys('mass_emails*'): cache.delete(key)`
11. Rerun `generate_mass_email_user_lists`
12. 🔴 [on main] Notice new records are enqueued for the one off email
13. 🟢 [on PR] Notice no new records are added for the one off email

### DIFF
--- a/kobo/apps/mass_emails/tasks.py
+++ b/kobo/apps/mass_emails/tasks.py
@@ -106,7 +106,8 @@ def mark_old_enqueued_mass_email_record_as_failed():
         .distinct()
     )
 
-    # this won't actually be evaluated until we call update() so it's safe to declare it here
+    # this won't actually be evaluated until we call
+    # update() so it's safe to declare it here
     configs_to_update = (
         MassEmailConfig.objects.filter(uid__in=affected_one_time_campaigns)
         .annotate(
@@ -119,11 +120,13 @@ def mark_old_enqueued_mass_email_record_as_failed():
 
     with transaction.atomic():
         updated_records_count = records_to_update.update(status=EmailStatus.FAILED)
-        configs_to_update.update(live=False)
+        updated_configs_count = configs_to_update.update(live=False)
 
     logging.info(
-        f'Updated {updated_records_count} MassEmailRecord(s) from `enqueued` to `failed` '
+        f'Updated {updated_records_count} MassEmailRecord(s)'
+        ' from `enqueued` to `failed` '
         f'that were older than {threshold_date}.'
+        f' Turned {updated_configs_count} campaigns off.'
     )
 
 
--- a/kobo/apps/mass_emails/tasks.py
+++ b/kobo/apps/mass_emails/tasks.py
@@ -100,8 +100,7 @@ def mark_old_enqueued_mass_email_record_as_failed():
     )
 
     affected_one_time_campaigns = list(
-        records_to_update.select_related('email_job__email_config')
-        .filter(email_job__email_config__frequency=-1)
+        records_to_update.filter(email_job__email_config__frequency=-1)
         .values_list('email_job__email_config__uid', flat=True)
         .distinct()
     )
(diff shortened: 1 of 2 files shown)


kobo/apps/mass_emails/tasks.py
```diff
@@ -95,14 +95,38 @@ def mark_old_enqueued_mass_email_record_as_failed():
...
-    updated_records = MassEmailRecord.objects.filter(
-        status=EmailStatus.ENQUEUED,
-        date_created__lt=threshold_date
-    ).update(status=EmailStatus.FAILED)
+    records_to_update = MassEmailRecord.objects.filter(
+        status=EmailStatus.ENQUEUED, date_created__lt=threshold_date
+    )
+
+    affected_one_time_campaigns = list(
+        records_to_update.select_related('email_job__email_config')
```
noliveleger: I think `.select_related('email_job__email_config')` is useless since we are using a `.values_list()` in the queryset. No objects are loaded.

kobo/apps/mass_emails/tasks.py
```diff
@@ -95,14 +95,38 @@ def mark_old_enqueued_mass_email_record_as_failed():
...
+        .distinct()
+    )
+
+    # this won't actually be evaluated until we call
+    # update() so it's safe to declare it here
+    configs_to_update = (
+        MassEmailConfig.objects.filter(uid__in=affected_one_time_campaigns)
+        .annotate(
+            enqueued_count=Count(
+                'jobs__records__id', filter=Q(jobs__records__status='enqueued')
```
noliveleger: We should be consistent and use `EmailStatus.ENQUEUED` instead of `'enqueued'` ?

noliveleger:
I did not find `MASS_EMAIL_CONDENSE_SEND` in Constance. It's an env var AFAIK.

noliveleger:
@rgraber, it works great but I may have found edge-cases. 

1) What if an one-off campaign contains 100 recipients

- day 1: 90 emails are sent
- day 2: before the rest of campaign is sent, the last 10 users are deactivated (See https://github.com/kobotoolbox/kpi/blob/00bdbe88bc00ca7fd47b29b750288f57f39b0dcc/kobo/apps/mass_emails/tasks.py#L143) 

I think the email are resent to everyone because [everybody will be re-enqueued ](https://github.com/kobotoolbox/kpi/blob/00bdbe88bc00ca7fd47b29b750288f57f39b0dcc/kobo/apps/mass_emails/tasks.py#L518)

2) The other problem I noticed, if the task is killed (OOM, HPA cool down) before https://github.com/kobotoolbox/kpi/blob/00bdbe88bc00ca7fd47b29b750288f57f39b0dcc/kobo/apps/mass_emails/tasks.py#L497 right after the last record is saved. 

The MassEmailConfig will be still live with 0 records enqueued. The next day, `generate_mass_email_user_lists` see no "enqueued" records, so recreate a new job with everybody (even those who already received the email because of same reason described above).
It's not simple because an one-off could be restarted manually by an admin with "add to daily queue". Hard to distinguish a restarted campaign and a not-closed-properly one by the system. 

Those are not introduced by this PR though. Should it cover them or should we create other PRs?

@greptileai what do you think?

noliveleger (changes requested):

noliveleger (approved):
LGTM

PR #7489, opened 2026-08-26
feat(SSO): hide password/SSO disable for managed SSO users DEV-2361

jamesrkiger:
### 📣 Summary
Hide password update and disable SSO UI for users with a managed SSO social account.

### 💭 Notes

When an admin configures a managed SSO provider, users authenticated through it should not be able to change their password or disconnect their SSO account — both are controlled by the organization.

A few implementation notes:
- The security route now gates on `envStore.isReady` before rendering any sections, so no section needs to handle the pre-load state individually.
- `SsoSection` no longer uses MobX `observer`. It now uses the `useSession` hook, which handles reactivity internally via a MobX reaction — so the component itself is plain React. Unfortunately, `securityRoute` now needs `observer` for tracking `envStore.isReady`.

### 👀 Preview steps

1. ℹ️ Configure a managed SSO provider in Django admin and connect a user account to it
2. Sign in as that user and navigate to **Account → Security**
3. Verify the **Password** section shows "Basic authentication not permitted" with no input for password change
4. Verify the **Single-Sign On** section shows "Connected to {provider name}" with no Disable button
5. Remove managed status from social app in django admin
6. Navigate to security page
7. Notice the Password section and SSO section are back to their normal state


jsapp/js/account/security/password/passwordSection.component.tsx
```diff
@@ -1,41 +1,57 @@
...
+            <NavLink to={`${ACCOUNT_ROUTES.CHANGE_PASSWORD}`}>
+              <Button
+                label={t('Update')}
+                size='m'
+                type='primary'
+                onClick={() => {
+                  /*TODO: Handle NavLink and Button*/
+                }}
+              />
+            </NavLink>
```
magicznyleszek, line 51: This should become:
```
        <ButtonNew
          variant='filled'
          component={Link}
          to={ACCOUNT_ROUTES.CHANGE_PASSWORD}
        >
          {t('Update')}
        </ButtonNeqw>
```

jsapp/js/account/security/sso/ssoSection.component.tsx
```diff
@@ -46,19 +47,27 @@ const SsoSection = observer(() => {
...
           )}
         </div>
-      ) : (
-        <div className={cx(securityStyles.securitySectionBody, styles.body)}>{t('Already connected')}</div>
       )}
 
-      {socialAccounts.length === 0 ? (
+      {connectedAccount ? (
+        <div className={styles.options}>
+          {!isManaged && <Button label={t('Disable')} size='m' type='primary' onClick={disconnectSocialAccount} />}
```
magicznyleszek, line 68: Maybe let's switch to ButtonNew here? :)

magicznyleszek (approved):
LGTM, two small mantineification opportunities noticed

jsapp/js/account/security/password/passwordSection.component.tsx
jamesrkiger: I did consider that, but given the way these buttons are displayed on the page in alignment with the other security route buttons I don't think it makes sense to switch just a couple of them. And I don't think is the right PR for migrating them all.

PR #7496, opened 2026-08-27
feat(sso): add provider detail endpoint DEV-2665

rajpatel24:
### 📣 Summary
This PR adds a public API endpoint that resolves an SSO provider's URL identifier into its display name, so the upcoming React login screens can render the "Log in with …" page that Django templates render today.

### 📖 Description
This PR adds GET `/api/v2/social-apps/<provider_id>/`, which returns the `provider_id` and `name` for a configured Social Application.

Some organizations sign in through an SSO provider that is deliberately not shown on the public login page. Its Social Application has `is_public` unchecked, and they distribute a direct link to their own staff, for example: `https://<kobo-server>/accounts/oidc/nca/login/`. Only the `nca` part varies. Today, that URL renders a shared Django template titled "Log in with {{ SocialApp.name }}". The same template is used for every provider on every server.

The Authentication Redesign deletes that template and rebuilds the screen in the SPA. To render it at a route like `#/login/oidc/<provider_id>`, the frontend needs to turn `provider_id` into a display name and distinguish a real provider from a typo so it can show the appropriate 404. The frontend cannot read the database directly, so it needs this endpoint.

Nothing changes for users yet. The existing login pages are untouched.

### 👀 Preview steps

1. ℹ️ Have a superuser account and be able to run manage.py.
2. Create a hidden SSO provider (provision_server sets is_public=False):
```
./manage.py provision_server socialapp --provider openid_connect \
  --provider_id nca --name "Norwegian Church Aid" \
  --client_id test-client-id --secret test-secret
```     
3. Log out and open `/accounts/login/.`
4. ℹ️ Notice that "Norwegian Church Aid" is not among the SSO buttons because the provider is hidden.
5. Open `/accounts/oidc/nca/login/`.
6. ℹ️ Notice that the "Log in with Norwegian Church Aid" page still renders unchanged.
7. 🔴 On main: GET `/api/v2/social-apps/nca/` -> 404, because the route does not exist.
8. 🟢 On the PR: GET `/api/v2/social-apps/nca/` -> 200 `{"provider_id": "nca", "name": "Norwegian Church Aid"}` even though the provider is hidden. That's the point of the endpoint.

9. 🟢 GET `/api/v2/social-apps/typo/` -> 404.
10. 🟢 GET `/api/v2/social-apps/` -> 404. There is deliberately no list endpoint.
11. 🟢 Notice that the response has exactly two keys. It does not expose `client_id`, `secret`, or `server_url`.


kobo/apps/accounts/views.py
```diff
@@ -128,3 +137,66 @@ class SocialAccountViewSet(
...
+
+    def get_object(self):
+        provider_id = self.kwargs[self.lookup_url_kwarg]
+        # Resolve through allauth's adapter rather than querying `SocialApp`
+        # directly, so this endpoint agrees with the two flows it sits between:
+        # the legacy `/accounts/oidc/{provider_id}/login/` page, and allauth's
+        # headless `auth/provider/redirect`, which the SPA posts to next. Both
+        # look providers up this way, so a 200 here means the redirect will work
+        try:
+            return get_socialaccount_adapter().get_app(self.request, provider_id)
```
noliveleger: You should call self.check_object_permissions(self.request, obj) even if it would do nothing special. Just as a safety guard if the permission classes is changed some day. 
Or at least , leave a comment why you omitted. `permission_classes` is useless atm.

kobo/apps/accounts/views.py
```diff
@@ -128,3 +137,66 @@ class SocialAccountViewSet(
...
+        description=read_md('accounts', 'social_apps/retrieve.md'),
+        responses=open_api_200_ok_response(
+            SocialAppDetailSerializer,
+            require_auth=False,
+            raise_access_forbidden=False,
+            validate_payload=False,
+        ),
+    )
+    def get(self, request, *args, **kwargs):
+        return super().get(request, *args, **kwargs)
```
noliveleger: You can probably set it at the View level. No need to overload the parent for nothing. I think we do have some examples in the  code.

noliveleger (changes requested):

noliveleger (approved):
LGTM

PR #7498, opened 2026-08-27
feat(SSO): confirm modal for managed SSO signup DEV-2360

jamesrkiger:
### 📣 Summary
Adds a confirmation modal for connecting to a managed SSO provider.

### 💭 Notes
Managed SSO providers (`SocialApp.managed: true`) require an extra confirmation step before navigating to the connect URL, since the action is irreversible without direct admin intervention. Non-managed providers are unaffected and continue to navigate directly.

### 👀 Preview steps

1. ℹ️ Have an account with no SSO connected, on a server with a managed SSO provider configured
   - After you configure SSO, you need to go to Django admin → Account Extras → Social app custom datas, and there "Add social app custom data". Make sure it is both managed and public.
2. Navigate to **Account → Security**
3. Click the SSO provider button — a modal appears
4. Verify modal copy matches designs
5. Click **Go back** — modal closes, nothing happens
6. Click the provider button again, then **Proceed** — navigates to the SSO connect flow as before
7. Turn off 'managed' status on SSO provider and verify that button on security page now navigates directly to SSO connection without modal.


jsapp/js/account/security/sso/ssoSection.component.tsx
```diff
@@ -70,19 +73,30 @@ export default function SsoSection() {
...
+              type='primary'
+              onClick={() => {
+                if (socialApp.managed) {
+                  setPendingManagedAppConfirm(socialApp)
+                } else {
+                  window.location.href = providerLink(socialApp)
+                }
+              }}
+            />
           ))}
```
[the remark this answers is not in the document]
jamesrkiger: Based on the UI flow, I don't think there's any reason we need to provide link actions here. It's a link that takes you to the SSO provider for authorization and then sends you back to the app. It's not a link you're supposed to save for later.

jsapp/js/account/security/sso/ManagedSsoConfirmModal.tsx
```diff
@@ -0,0 +1,48 @@
...
+  connectHref: string
+  onClose: () => void
+}
+
+export default function ManagedSsoConfirmModal({ providerName, connectHref, onClose }: ManagedSsoConfirmModalProps) {
+  return (
+    <ModalNew
+      opened
+      onClose={onClose}
+      title={t('Connecting with ##provider##').replace(/##provider##/g, providerName)}
```
magicznyleszek: Why regex? Our standard way is simple string replacement

jsapp/js/account/security/sso/ManagedSsoConfirmModal.tsx
```diff
@@ -0,0 +1,48 @@
...
+        </Alert>
+
+        <Group justify='flex-end'>
+          <ButtonNew size='md' variant='light' onClick={onClose}>
+            {t('Go back')}
+          </ButtonNew>
+
+          <ButtonNew size='md' component='a' href={connectHref}>
+            {t('Proceed')}
+          </ButtonNew>
```
magicznyleszek, line 43: Is it ok that after going to usual flow, "go back" button doesn't really go back? Although it will be fixed in the Auth Redesign project in few months, I just want to make sure this is not required in this project

magicznyleszek (approved):
LGTM, few nitpicks

kobo/apps/accounts/templates/socialaccount/login.html
jamesrkiger, line 1: @rgraber Could you take a quick look at this? Mostly a formality, but it's your end of the repo so I wanted to double check. Leszek pointed out that this "go back" button previously took you to the login screen no matter what, even if you were coming from the security page in the app as an already logged in user. So I'm using the allauth context to determine when to go back to login vs. going back to the security page.

rgraber: looks fine to me, thanks for checking

PR #7501, opened 2026-08-28
feat(auth): main skeleton DEV-1871

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Groundwork for the redesigned sign-in and account creation screens. Hidden behind a feature flag.

### 💭 Notes

Changes here:

- **New `jsapp/js/auth/`** — where the redesigned authentication screens will live
  - `routes.tsx` — `/auth` mounts as a sibling of `<App />`, so none of the logged-in layout (main header, drawer, ToS and password blockers) applies
  - `AuthAppProviders.tsx` — what `basicLayout` sets up, minus the chrome. Kept separate from the container so stories can pass a fresh `QueryClient` instead of the app singleton, which would leak one story's `/environment` response into the next
  - `AuthContainer/AuthContainer.tsx` — the frame: background, logo, language picker, `<Outlet/>`, legal links
  - `AuthContainer/AuthCard.tsx` — Mantine `Paper` plus a real vertical `Divider`; passing `aside` turns on the second column. The stacking is a **container query**, not a media query, so it can be tested in storybook. Plus it's a nice feature ;)
  - `AuthTestRoute/` — placeholder at `#/auth/test` to be replaced by first real route
  - `StandaloneUILanguageSelector/` moved here from `#/components/`
- **`router/router.tsx`** — now wrapping `authRoutes()` and the existing `App` route - everything else is reindentation
- **`router/routerConstants.ts`** — `ROUTES.AUTH_ROOT` and an `AUTH_ROUTES` group shaped like `PROJECTS_ROUTES`
- **`endpoints/environment.mocks.ts`** — `makeEnvironmentMock(override)` so stories can vary `auth_configuration`
- **`stores/session.mocks.ts`** — `setAnonymousSessionForStories()`, for screens a visitor sees before logging in (it has to override the default in `preview.tsx` that applies logged-in state for every story)
- **New `scss/pageFrame.scss`** — the frame shared by the auth screens and the standalone error pages
- **`errorApp/ErrorPage.module.scss`** — uses shared mixin now
- **Stories** — `Features/AuthContainer`: logged out, both themes, the stacked card, and the config that hides the logo or the legal links

### 👀 Preview steps

1. ℹ️ use a private/incognito window so you're logged out, and have Django admin access on the side
2. open `<your instance>/#/auth/test?ff_authRedesignEnabled=true`
3. 🔴 [on main] you get "Section not found"
4. 🟢 [on PR] you get the new frame: KoboToolbox logo top-left, language picker top-right, white card in the middle, and Terms of Service · Privacy Policy at the bottom (each link only shows if it's configured on your instance)
5. 🟢 notice the card is a placeholder with nothing to sign you into — that part arrives later
6. pick another language in the top-right picker → the page reloads in that language and the picker keeps it
7. narrow the browser window down to phone width → (less than 460px) the card padding shrinks and nothing overlaps or clips
8. press Tab repeatedly starting from the address bar → focus moves logo → language picker → card → footer links, with a visible ring on each
9. log in as usual in a normal window, then open the same URL with `?ff_authRedesignEnabled=true` → 🟢 the same screen appears
10. in a fresh private window, open `<your instance>/#/auth/test` without the `?ff_authRedesignEnabled=true` part → 🟢 you land on Projects instead (the flag sticks per browser tab, so it has to be a fresh window)
11. custom theme: Django admin → Hub → Configuration files → add a `login_background` image, then reload the auth URL → 🟢 the photo fills the page behind a dark overlay, the logo turns white, the footer links turn white and the language picker goes dark
12. Django admin → Constance → Config → turn off `SHOW_KOBOTOOLBOX_LOGO` → reload → 🟢 the logo is gone and the language picker stays where it was


jsapp/js/auth/AuthAppProviders.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { cssVariablesResolverKobo, themeKobo } from '#/theme'
+import { Tracking } from '../router/useTracking'
+import ToasterConfig from '../toasterConfig'
+
+/**
+ * Context the authentication screens need, mirroring what `#/router/basicLayout.component` sets up for
+ * the logged in app. They can't reuse that layout - it also renders the main header, the drawer and the
+ * route blockers, none of which belong on a sign-in page.
+ *
+ * Separate from `AuthContainer` so stories can render the container against a fresh `QueryClient`.
```
jamesrkiger: I'd suggest dropping this comment. This component makes sense on its own without storybook (and if storybook was the only reason for separating it out I'd say that's not a good enough reason to do it).

jsapp/js/auth/AuthContainer/AuthContainer.stories.tsx
```diff
@@ -0,0 +1,174 @@
...
+import AuthCard from './AuthCard'
+import AuthContainer from './AuthContainer'
+
+const TERMS_OF_SERVICE_URL = 'https://example.org/terms'
+const PRIVACY_POLICY_URL = 'https://example.org/privacy'
+
+/**
+ * A flat blue square, inlined rather than fetched to not depend on the network.
+ */
+const BACKGROUND_IMAGE_DATA_URI =
```
jamesrkiger: Is there any way we could add a real image for preview purposes in Storybook? Not worth spending a ton of time on, but would definitely be a nice-to-have.

jamesrkiger:
Can we make the non-custom storybook stories use the same logo as the app itself?

image

jamesrkiger (approved):
Left a minor few comments, but otherwise LGTM

PR #7503, opened 2026-08-28
test(storybook): improve map stories DEV-2762

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 💭 Notes

Background: `WithPointsRepeatedInEveryWorldCopy` failed on CI with `expected 0 to be greater than 0` — no markers within 5s. Locally the map plots its points in ~0.7s (chromium), ~0.8s (firefox), ~0.85s (webkit), but one cold run under CPU load took 5.3s (over the timeout set)

Changes here, all in `FormMapWrapper.stories.tsx`:

- **`waitForPlottedPoints()`** — shared wait used by every story: first the map's own "Fetching points…" overlay has to go away, then the markers have to be drawn. Replaces the `#data-map` in-the-document checks, which passed long before anything was plotted.
- **`WAIT`** — one ceiling for the whole file, 5s → 10s. `waitFor` resolves when condition holds, so if nothing is broken it will be much faster than 10
- **`describeMap()`** — attaches the map's state to a failing assertion, so the next CI failure names the state instead of just counting to zero
- **`WithPointsRepeatedInEveryWorldCopy`**
  - the `waitFor` that clicked zoom-out from inside its own callback is now a plain loop that waits for each zoom animation to land
  - new closing step asserts more than one cluster on screen, which is what the story's name has claimed all along
- **`WithOnlyStartGeopoint` / `WithBothGeopointTypes`** — checks that only needed the map loaded now run once, after `waitForPlottedPoints()`

### 👀 Preview steps

1. ℹ️ nothing changes in the app itself — if you want to confirm, Project → Data → Map looks and behaves the same as on main
2. run `npm run storybook` locally
3. go to Features → FormMap → With Points Repeated In Every World Copy
4. 🟢 notice the map is zoomed all the way out, with several copies of the world side by side and a cluster of 4 points in each
5. 🟢 notice the CI passes for this PR
6. Rerun CI multiple (5? 🤣) times, each time ensuring it passes

### DIFF
--- a/jsapp/js/components/map/FormMapWrapper.stories.tsx
+++ b/jsapp/js/components/map/FormMapWrapper.stories.tsx
@@ -216,6 +216,55 @@ function mapHandlers(asset: AssetResponse, submissions: SubmissionResponse[]) {
   ]
 }
 
+/**
+ * Ceiling shared by the waits below. `waitFor` carries on the moment its condition holds, so a high ceiling costs
+ * nothing on a quick machine and only buys patience on a cold CI worker. Kept under the runner's 30s per-story limit,
+ * so a real regression fails on the assertion rather than on the clock.
+ */
+const WAIT = { timeout: 10000 }
+
+/** The markers the map has drawn, clusters included. Scoped to the map, so it also covers the container being there. */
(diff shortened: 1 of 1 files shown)


jamesrkiger (approved):
LGTM

PR #7512, opened 2026-08-30
feat(processing): hide Transcript tab for text questions DEV-2374

pauloamorimbr:
### 📣 Summary

The Processing view now only shows the "Transcript" tab for questions that actually have something to transcribe.

### 💭 Notes

For audio questions, the Processing view still shows all three tabs (Transcript, Translations, Analysis) — no change there. For text questions, the Transcript tab is now hidden entirely (there's no audio/video to transcribe), leaving only Translations (now the default landing tab) and Analysis.

- Added `getAvailableTabsForQuestionType()` in `components/processing/common/utils.ts` to centralize which tabs apply to a given question type.
- `SingleProcessingContent` uses it to hide the Transcript nav item/body for text questions, and also redirects off the Transcript tab (to Translations) if a stale link or a question switch lands on it for a question type that doesn't support it.
- `ProcessingRootRedirect` (the bare `/processing/:xpath/:submissionEditId` route) now picks Transcript or Translations as the default tab based on question type, instead of always defaulting to Transcript.
- Per the tabs appear/disappear (not disabled) per Tessa's comment on the ticket.

### 👀 Preview steps

1. ℹ️ have a project with both an audio question and a text question, with at least one submission answering both
2. open Processing for the audio question (e.g. via "Open" on the audio cell in the table)
3. 🟢 see all three tabs: Transcript, Translations, Analysis (unchanged)
4. open Processing for the text question (e.g. via "Open" on the text cell in the table, behind the `nlpTextActionsEnabled` feature flag)
5. 🔴 [on main] the Transcript tab is shown even though there's nothing to transcribe
6. 🟢 [on PR] only Translations and Analysis are shown, and Translations is the tab you land on by default

---
_Recreated from #7511, which got closed as a side effect of a branch rename._

### DIFF
--- a/jsapp/js/components/processing/SingleProcessingContent/index.tsx
+++ b/jsapp/js/components/processing/SingleProcessingContent/index.tsx
@@ -1,14 +1,16 @@
 import { Tabs } from '@mantine/core'
-import React from 'react'
+import React, { useEffect } from 'react'
 import { Link } from 'react-router-dom'
 import type { AdvancedFeatureResponse } from '#/api/models/advancedFeatureResponse'
 import type { BulkActionResponse } from '#/api/models/bulkActionResponse'
 import type { DataResponse } from '#/api/models/dataResponse'
 import type { DataSupplementResponse } from '#/api/models/dataSupplementResponse'
+import { findRowByXpath } from '#/assetUtils'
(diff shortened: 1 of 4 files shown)


magicznyleszek (approved):
LGTM

PR #7513, opened 2026-08-31
fix(celery): schedule trash bin and project ownership dispatchers on `kpi_queue` DEV-2618

rajpatel24:
### 📖 Description
Moves four scheduled orchestration tasks from `kpi_low_priority_queue` to `kpi_queue`:

- trash_bin.task_restarter
- trash_bin.garbage_collector
- project_ownership.task_restarter
- project_ownership.garbage_collector

These tasks only enqueue work and finish in milliseconds, but they were scheduled onto the same queue as the long-running deletions and transfers they dispatch. When that queue backs up, the recovery path is stuck behind the backlog it exists to clear.

The deletion and transfer tasks themselves are unchanged and stay on `kpi_low_priority_queue`.

### DIFF
--- a/kobo/settings/base.py
+++ b/kobo/settings/base.py
@@ -1543,7 +1543,7 @@ CELERY_BEAT_SCHEDULE = {
     'trash-bin-task-restarter': {
         'task': 'kobo.apps.trash_bin.tasks.task_restarter',
         'schedule': crontab(minute='*/30'),
-        'options': {'queue': 'kpi_queue'}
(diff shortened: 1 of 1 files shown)


noliveleger (dismissed):
Maybe we could leave a comment to explain why such change. Just to be sure another dev doesn't change it back a later time.

Otherwise, LGTM (you can merge it when the comments are added).

noliveleger (approved):
LGTM

PR #7514, opened 2026-08-31
fix(trashBin): give each trash type its own restart budget DEV-2618

rajpatel24:
### 📣 Summary
Splits `MAX_RESTARTED_TASKS` into three per-type settings for trash bin deletions, and stops `task_restarter` from swallowing its own soft time limit.

### 📖 Description
Implements the findings from #7422 and supersedes it.

The investigation showed that limiting how many deletions `task_restarter` enqueues does not limit DB pressure: concurrency is set by the size of the worker pool, not by how many messages are published. Two things from #7422 are worth keeping on their own merits, and this PR is just those two.

1. Per-type restart budgets. _restart_stuck_tasks() sliced all three trash types with settings.MAX_RESTARTED_TASKS, which project_ownership/tasks.py also uses with a different "batch per run" meaning - one shared name, two unrelated behaviours. Each trash type now has its own setting:

- MAX_RESTARTED_ACCOUNT_DELETIONS
- MAX_RESTARTED_PROJECT_DELETIONS
- MAX_RESTARTED_ATTACHMENT_DELETIONS

All default to 100, so behaviour is unchanged. They are separate because the three deletions do not cost the same. 

2. Do not swallow `SoftTimeLimitExceeded`. In the enqueue block it was caught by the generic except Exception and logged as Could not restart <Model> #<id>, which is wrong on both counts: nothing is wrong with the object, and the restarter had already claimed it, so it sat untouched until the next stuck threshold ~76 minutes later. It now restores `date_modified` and re-raises.

### DIFF
--- a/kobo/settings/base.py
+++ b/kobo/settings/base.py
@@ -2402,9 +2402,7 @@ MAX_RESTARTED_TRANSFERS = 20
 # Number of stuck trash bin deletions `task_restarter` re-enqueues per run per type
 MAX_RESTARTED_ACCOUNT_DELETIONS = env.int('MAX_RESTARTED_ACCOUNT_DELETIONS', 100)
 MAX_RESTARTED_PROJECT_DELETIONS = env.int('MAX_RESTARTED_PROJECT_DELETIONS', 100)
-MAX_RESTARTED_ATTACHMENT_DELETIONS = env.int(
-    'MAX_RESTARTED_ATTACHMENT_DELETIONS', 100
-)
+MAX_RESTARTED_ATTACHMENT_DELETIONS = env.int('MAX_RESTARTED_ATTACHMENT_DELETIONS', 100)
 
 # Number of times a trash bin task that failed on a transient (infrastructure)
 # error is automatically restarted before it requires manual intervention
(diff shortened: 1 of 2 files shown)


kobo/settings/base.py
```diff
@@ -2395,10 +2395,15 @@ def __init__(self, *args, **kwargs):
...
 
-# Number of stuck tasks should be restarted at a time
+# Number of stuck project ownership tasks and transfers restarted at a time
 MAX_RESTARTED_TASKS = 100
 MAX_RESTARTED_TRANSFERS = 20
 
+# Number of stuck trash bin deletions `task_restarter` re-enqueues per run per type
+MAX_RESTARTED_ACCOUNT_DELETIONS = env.int('MAX_RESTARTED_ACCOUNT_DELETIONS', 100)
+MAX_RESTARTED_PROJECT_DELETIONS = env.int('MAX_RESTARTED_PROJECT_DELETIONS', 100)
+MAX_RESTARTED_ATTACHMENT_DELETIONS = env.int('MAX_RESTARTED_ATTACHMENT_DELETIONS', 100)
```
noliveleger: Let's change the default: 
What about  50/100/300 or 25/100/300. 
Attachments are pretty small to delete, the server can handle more than 100 at a time. 
What do you think?

noliveleger (approved):
LGTM, please change the defaults before merging to something you think is good.

kobo/settings/base.py
rajpatel24: Yeah, I think 50/100/300 would be better

PR #7515, opened 2026-08-31
feat(organizations): support q search on organization asset usage DEV-2700

platreth:
## 📣 Summary
Projects in the organization usage table can now be searched by name.

## 📖 Description
The usage page lists every project in the organization. This adds name search to the data behind that table, so the search box coming in DEV-2701 can narrow the list instead of always showing everything.

## 💭 Notes
- Reuses the existing `SearchFilter`, chained before `AssetOrganizationUsageFilter` in `OrganizationViewSet.asset_usage`. Bare `q` terms search `name__icontains`.
- Fielded queries still go through the query parser's allowlist, and the queryset was already scoped to the org owner, so nothing new is exposed.
- Schema and orval client regenerated.

## 👀 Preview steps
1. Log in as an owner or admin of an org with a few projects.
2. Open `/api/v2/organizations/<org_id>/asset_usage/?q=<part of a project name>`.

🔴 On main: `q` is ignored and every project comes back.
🟢 On this PR: only projects whose name contains the term (case insensitive). A `q` under 3 characters returns a 400.

### DIFF
--- a/kpi/docs/api/v2/organizations/org_asset_usage.md
+++ b/kpi/docs/api/v2/organizations/org_asset_usage.md
@@ -1,3 +1,5 @@
 ## Retrieve organization asset usage tracker
 
 Tracks the total usage of each asset for the user in the given organization
+
+Use the `q` query parameter to filter by project name (e.g. `?q=household survey`). Bare search terms must be at least 3 characters long. The standard query syntax is supported (e.g. `?q=name:household`).
--- a/kpi/docs/api/v2/organizations/org_asset_usage.md
+++ b/kpi/docs/api/v2/organizations/org_asset_usage.md
@@ -2,4 +2,4 @@
 
 Tracks the total usage of each asset for the user in the given organization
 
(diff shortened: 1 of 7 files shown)


platreth:
@greptileai

kpi/docs/api/v2/organizations/org_asset_usage.md
```diff
@@ -1,3 +1,5 @@
 ## Retrieve organization asset usage tracker
 
 Tracks the total usage of each asset for the user in the given organization
+
+Use the `q` query parameter to filter by project name (e.g. `?q=household survey`). Bare search terms must be at least 3 characters long. The standard query syntax is supported (e.g. `?q=name:household`).
```
rgraber: `?q=name:bananas` didn't work for me, but `?q=bananas` did

rgraber (commented):
Bare terms worked but field name qualifiers didn't

kpi/docs/api/v2/organizations/org_asset_usage.md
platreth: Thanks for catching that! Problem in the doc

rgraber (approved):
LGTM

PR #7516, opened 2026-08-31
feat(sso): update InAppMessage for managed sso DEV-2758

rgraber:
### 🗒️ Checklist

1. [ ] run linter locally

### 💭 Notes
Backend-only change to make parts of managed SSO easier. New field not added to the serializer because the frontend doesn't need the information. Read-only because there's no current use case for a user to change it, it just needs to be set on the backend.


Guitlle (approved):
LGTM

PR #7521, opened 2026-09-01
fix(dataTable): make audio accessible after question rename DEV-2765

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Media files (images, audio, video) attached to a question that was later renamed now show up in the data table and submission modal again, and audio among them can be opened in the processing view.

### 💭 Notes

Renaming a question in the form builder used to hide the media already collected under its old name. The files were never lost, but the data table stopped offering any way to reach them. This is the same limitation #7187 fixed for renamed groups, now fixed for renamed questions too.

In the submission details modal it was worse: a renamed question showed up with a blank response, and renaming a *group* left that group displaying no questions at all — every answer inside it disappeared from the modal, media or not.

Changes here:

- **`submissionMediaUtils.ts`**
  - `findAttachmentByQuestionXpath` — one place that matches an attachment on the path recorded at submission time
  - `getAttachmentQuestionType` — mimetype → row type, with `file` as the fallback. ⚠️ Two documented imprecisions, both harmless for display: `background-audio` is indistinguishable from `audio`, and a `file` question holding a photo reads as `image`
- **`DataTableCell/index.tsx`** — the Data Table fix. `questionType`/`questionXpath` fall back to the attachment when the column has no survey row. Also drops a guard that could never fire (`TABLE_MEDIA_TYPES` has no `text`).
- **`sidebarSubmissionMedia.tsx`** - same fallback for the type switch, so the audio player renders instead of nothing
- **`SelectQuestion.tsx`** - the question selector names the renamed question instead of rendering a blank option
- **`submissionUtils.tsx`** — the modal fix
  - `traverseSurvey` now keeps a ledger of which submission key each row's answer came from, and `addUnaccountedAnswers` then appends a row for every leftover answer
  - `findGroupForUnaccountedAnswer` + `findDisplayGroupByPath` place each recovered row
  - `UNMODELED_META_QUESTION_NAMES` keeps meta questions from surfacing as rows. Broader than `META_QUESTION_TYPES` on purpose, sourced from `q.hiddenTypes()` in `xlform/src/model.aliases.coffee`: `simserial` and `subscriberid` are deprecated in `model.configs.coffee`, and `imei`/`phone_number` are legacy aliases
- **Tests & stories**
  - `submissionMediaUtils.tests.ts` — 10 tests for the two helpers, including two questions whose files share a basename (the case that makes basename matching wrong)
  - `tableUtils.tests.ts` — ensures that the renamed question's column survives `getAllDataColumns`
  - `DataTableCell.stories.tsx` — `RenamedImageQuestion` and `RenamedAudioQuestion`, both with `question: undefined`

**Note: A renamed question appears twice in the modal** — once under its current label with no answer, once with the recovered answer under its old path.

Renames *inside* a repeat group are not covered.

### 👀 Preview steps

1. ℹ️ Have a project with an image question and an audio recording question at the top level (not in a group), plus a group containing two ordinary questions
2. Make one submission filling in everything, with a photo and a recording attached
3. Go to Project → Data → Table and note the baseline: the image cell has an icon that opens the photo, the audio cell has a small player and a button that opens single processing
4. Open the submission from Project → Data → Table by clicking its row, and note the modal shows every question with its answer
5. Rename **both media questions** (e.g. `photo_question` → `photo_question_v2`, `audio_question` → `audio_question_v2`) **and the group** (e.g. `my_group` → `my_group_v2`). Deploy
6. 🔴 [on main] In Project → Data → Table both media cells now show a plain file name and nothing else. The files are on the server, but this screen has forgotten the way there
7. 🟢 [on PR] The image cell has its icon back and opens the photo; the audio cell has its player back
8. 🟢 [on PR] From the audio cell, open single processing: the correct recording plays, and the question picker at the top names the question instead of showing an empty option. Transcribe or translate as usual
9. 🔴 [on main] Open the submission in the modal: both renamed questions show a **blank** response, and the renamed group shows up with **no questions inside it at all** — three answers vanished from this view
10. 🟢 [on PR] The modal shows the group populated with its two answers again, complete with labels and choice labels, and the two media questions render their player and photo
11. ℹ️ Expect each renamed question to be listed **twice** in the modal — a blank row under the new name, and the recovered answer below it. That is intended; see Notes
12. 🟢 Make one more submission after the rename and confirm the new columns and modal rows behave exactly as in step 3–4 — the fallback only speaks up where the form definition has gone quiet


jsapp/js/components/processing/SingleProcessingHeader/SelectQuestion.tsx
```diff
@@ -76,22 +77,41 @@ export default function SelectQuestion({
...
       const foundRow = findRowByXpathOrLeafName(assetContent, submissionXpath)
-      if (!foundRow || !isAudioRow(foundRow.type)) {
+      const attachment = foundRow ? undefined : findAttachmentByQuestionXpath(submission, submissionXpath)
+      const type = foundRow?.type ?? (attachment && getAttachmentQuestionType(attachment))
+
+      if (!type || !isAudioRow(type)) {
         continue
       }
-      result.push(buildOption(submissionXpath, foundRow))
+      result.push(foundRow ? buildOption(submissionXpath, foundRow) : buildOptionWithoutRow(submissionXpath, type))
```
jamesrkiger, line 114: I think it would make more sense to have a buildOption function that receives the necessary args for either case and decides what to do on its own.

jsapp/js/components/processing/SingleProcessingSidebar/sidebarSubmissionMedia.tsx
```diff
@@ -35,7 +36,12 @@ export default function SidebarSubmissionMedia({ asset, xpath, submission }: Sid
     )
   }
 
-  switch (findRowByXpathOrLeafName(asset.content, xpath)?.type) {
+  // Form definition first; the attachment's mimetype keeps the player working
+  // for a question renamed after this submission came in, which leaves no row to
+  // read the type from.
```
jamesrkiger, line 41: I'd suggest trying to keep the comments a bit shorter where possible. For example, you could write something like "Use attachment mimetype as fallback for renamed questions".

jsapp/js/components/submissions/submissionMediaUtils.ts
```diff
@@ -1,7 +1,51 @@
...
+ * Guesses the type of the question that produced an attachment, from its
+ * mimetype.
+ *
+ * Needed when the question is gone from the current form definition (renamed
+ * after this submission came in), leaving no row to read the real type from.
+ * NOTE: Two known imprecisions, both harmless for displaying the file:
+ * `background-audio` looks like `audio`, and a `file` question holding e.g. a
+ * photo reads as `image`.
+ */
+export function getAttachmentQuestionType(
```
jamesrkiger, line 29: I think "get" should be reserved for more reliable retrieval methods. Maybe "inferAttachmentQuestionType"?

jsapp/js/components/submissions/submissionMediaUtils.ts
```diff
@@ -1,7 +1,51 @@
...
+  attachment: Pick<SubmissionAttachment, 'mimetype'>,
+): AnyRowTypeName | undefined {
+  // No mimetype leaves nothing to guess from.
+  if (!attachment.mimetype) {
+    return undefined
+  }
+  if (attachment.mimetype.startsWith('audio/')) {
```
jamesrkiger, line 36: This doesn't catch files with "application/ogg", which we support as audio.

jsapp/js/components/submissions/submissionUtils.tsx
```diff
@@ -381,6 +401,157 @@ export function getSubmissionDisplayData(
...
+
+    // A renamed group leaves the question findable by leaf name, with its real type
+    // and choice list. A renamed question leaves only its attachment's mimetype -
+    // still enough to render media rather than a bare filename.
+    const row = findRowByXpathOrLeafName(assetContent, key)
+    const attachment = row ? undefined : findAttachmentByQuestionXpath(submissionData, key)
+
+    findGroupForUnaccountedAnswer(output, key, row, flatPaths).children.push(
+      new DisplayResponse(
+        row?.type ?? (attachment && getAttachmentQuestionType(attachment)) ?? null,
```
jamesrkiger, line 478: Might be nice to put this in a variable above for readability.

jsapp/js/components/submissions/submissionUtils.tsx
```diff
@@ -381,6 +401,157 @@ export function getSubmissionDisplayData(
...
+  ...recordValues(META_QUESTION_TYPES),
+  ...UNMODELED_META_QUESTION_NAMES,
+])
+
+/**
+ * Adds a row for every answer stored under a path the current form no longer
+ * accounts for - what renaming or removing a question or group leaves behind.
+ * Without it the answer and its file vanish from the modal, while Data Table keeps
+ * the column (see `getAllDataColumns`). Rows never go into the renamed question's
+ * own row, as its old name may belong to another question by now.
```
jamesrkiger, line 433:
```suggestion
 * Adds a row for every answer stored under a path not included in current form.
```
Somewhat subjective, admittedly.

jsapp/js/components/submissions/DataTableCell/index.tsx
```diff
@@ -79,30 +79,37 @@ export default function DataTableCell(props: DataTableCellProps) {
...
-
-      if (
-        props.question.type === QUESTION_TYPES.audio.id ||
-        props.question.type === QUESTION_TYPES['background-audio'].id
-      ) {
-        if (mediaAttachment !== null && props.question.$xpath !== undefined) {
+  // A question renamed after this submission came in is gone from the form
+  // definition, so its column gets no `props.question` - even though the response and
+  // its file are right here. The attachment still records the path used back then.
+  const renamedQuestionAttachment = props.question
```
jamesrkiger, line 85: This isn't just for renamed questions, right? Also deleted?

jamesrkiger (changes requested):

jamesrkiger (approved):
lgtm

PR #7523, opened 2026-09-01
feat(organizations): members table search DEV-2459

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

The members list of a team or organization can now be searched.

### 📖 Description

A search field sits next to the "Members" title. Typing at least three characters filters the list; the search covers name, username and email. When nothing matches, the table says so instead of showing an empty header.

### 💭 Notes

Search relies on the `q` parameter from DEV-2458 (#7404, #7430). It's server-side, so it filters across all pages rather than the current one.

Changes here:
- **`jsapp/js/account/organization/MembersRoute.tsx`**
  - Added search input on top (uses existing members query)
  - Require minimum characters typed to call actual endpoint - below `settings.MINIMUM_DEFAULT_SEARCH_CHARACTERS` we do nothing, or fire a toast if user presses Enter
  - Searching resets to page 1, sharing that with sorting
  - Inline `Alert` when the request fails
- **`jsapp/js/UniversalTable/UniversalTableCore.tsx`, `index.tsx`, `UniversalTableCore.module.scss`**
  - New `emptyMessage` prop, rendered as one centered row in the `tbody` (similarly to what `projectsTable` and `assetsTable` already have)
  - Generic empty message
  - Suppressed while the spinner is up — "nothing here" is a lie mid-fetch
- **`jsapp/js/endpoints/organizationMembers.mocks.ts`** 
  - MSW handler reproducing the parts of the `q` contract callers have to survive: the too-short 400, the query-parser 400, and the filtering itself. Hand-rolled rather than wrapping Orval's `getApiV2OrganizationsMembersListMockHandler`, which hardcodes `status: 200` and can only override the body.
- **Stories** — `MembersRoute.stories.tsx` and `UniversalTableCore.stories.tsx`

### 👀 Preview steps

1. ℹ️ have an account in a team or organization with several members
2. go to Account Settings → Team (or Organization) → Members
3. 🔴 [on main] notice there is no way to search the list
4. 🟢 [on PR] notice a search field with a magnifying glass, top right, on the same row as the "Members" title
5. type at least three characters of a member's name → 🟢 notice the list narrows to matches
6. clear the field → 🟢 notice the full list comes back
7. go to page 2, then type a search that matches someone on page 1 → 🟢 notice you land on page 1 rather than an empty page
8. type one or two characters → 🟢 notice nothing happens and the full list stays put
9. leaving those one or two characters in the field, press Enter → 🟢 notice a warning explaining the three character minimum
10. search for someone who has never worked at your organization, e.g. `voldemort` → 🟢 notice the table says nobody matches, rather than leaving you with a header and a nagging sense of loss
11. search for a phrase containing an apostrophe or a bracket, e.g. `o'brien` → 🟢 notice an error message in place of the table, with the search field still there so the phrase can be fixed
12. run storybook and verify the new stories work correctly


duvld (approved):
I believe the AI has a niche but correct observation, but LGTM otherwise

PR #7531, opened 2026-09-03
feat(sso): optional and customizable in-app message on confirmation page DEV-2416

platreth:
### 📣 Summary
When enabling managed SSO, admins can now choose whether to send the in-app "connect your SSO account" message, and edit its text before confirming.

### 📖 Description
When a superuser enables managed SSO or adds a managed domain, the confirmation screen now has a "Send in-app message" checkbox and an editable message prefilled with the default text. Unchecking the box skips the message. When it is checked, the message cannot be empty.

### 💭 Notes
- The text only reaches users once DEV-2759 (calling `update_users` from the confirmed POST) lands. `update_users` now accepts `send_in_app_message` and `in_app_message_body`. The nightly sweep keeps the defaults.
- Validation reuses the existing interception branch. A confirmed POST with an empty body re-renders the confirmation page with an error instead of saving.
- The toggle script lives in a static file because `script-src` has no `unsafe-inline` when CSP is enabled.
- Custom text still gets `##username##` / `##sso_name##` substitution but is not translated, since it is not in the catalog.

### 👀 Preview steps
1. ℹ️ have a superuser and a social app with a "Social app custom data" entry
2. In Django admin, open the entry, tick "Managed", add a domain, click "Save"
3. 🔴 [on main] the confirmation page only lists the affected-account counts
4. 🟢 [on PR] a checked "Send in-app message to non-linked accounts" box and a prefilled "In-app message" textarea appear, and unchecking the box hides the textarea
5. Clear the textarea, click "Yes, I'm sure"
6. 🟢 "The in-app message cannot be empty." is shown and nothing is saved
7. Type a message, click "Yes, I'm sure": the changes are saved

### DIFF
--- a/kobo/apps/accounts/admin.py
+++ b/kobo/apps/accounts/admin.py
@@ -18,6 +18,7 @@ from kobo.apps.accounts.models import (
     SocialAppCustomData,
     SocialAppManagedDomain,
 )
+from kobo.apps.accounts.tasks import DEFAULT_IN_APP_MESSAGE_BODY
 from kobo.apps.accounts.utils import user_is_managed_by_sso, users_needing_update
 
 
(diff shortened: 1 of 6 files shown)


platreth:
@greptileai

rgraber (approved):
LGTM

PR #7535, opened 2026-09-03
fix(assetFile): validate redirect_url at serve time before redirecting DEV-2778

platreth:
### 📣 Summary

Media files hosted at a web address now only redirect to that address when it is a normal http or https link.

### 📖 Description

When a form uses a media file hosted elsewhere, downloading it from KoboToolbox sends you to that address. Until now the address was followed without any check, so a broken or unwanted one stored on a project could send people anywhere. KoboToolbox now checks it first and refuses to redirect when it is not a plain http or https address, or when it points inside the server's own network. Media hosted on ordinary public web addresses keeps working as before.

### 💭 Notes

- Serve-time check in the `content` action: `URLValidator(schemes=['http', 'https'])` plus `validate_url_against_ssrf()` from DEV-2777, `Http404` on failure. The stored value can predate the serializer's validation (rows created before #7520, or written by `sync_kobocat_form_media`, which stores whatever `media_filename` it gets), so validating at creation alone is not enough.
- `AssetFileSerializer` no longer accepts `ftp`/`ftps`. Django's default `URLValidator` schemes include them.
- Residual: an editor can still point media at any valid public URL for anyone who can view the project. That is the remote media feature itself. Self-hosted instances pointing media at an internal host can allow it through `SSRF_ALLOWED_IP_ADDRESS`, same as DEV-2777.
- Cost: remote URL files only, one regex and one DNS resolution per download. Uploaded files are untouched.
- The OpenRosa media endpoint (`get_media_file_response` in `kobo/apps/openrosa/apps/api/tools.py`) has the same redirect on legacy `MetaData` rows. Out of scope here, ticket to follow.

### 👀 Preview steps

Back end only, nothing changes in the UI.

1. ℹ️ log in as a user who owns a survey project
2. In the kpi container, create a form media row that bypasses the API validation, the way legacy data does, and note the printed URL:
   ```
   python manage.py shell -c "from kpi.models import Asset, AssetFile; a = Asset.objects.filter(asset_type='survey', owner__username='<you>').first(); af = AssetFile.objects.create(asset=a, user=a.owner, file_type='form_media', metadata={'filename': 'x.png', 'hash': 'md5:x', 'mimetype': 'image/png', 'redirect_url': 'http://127.0.0.1/x.png'}); print(af.download_url)"
   ```
3. Open that URL in the browser while logged in
4. 🔴 [on release/2.026.33] the browser is redirected to `http://127.0.0.1/x.png`
5. 🟢 [on PR] the response is a `404` with `{"detail": "Not found."}`
6. Repeat step 2 with `'redirect_url': 'https://example.com/x.png'`
7. 🟢 still a `302` to `https://example.com/x.png`

### DIFF
--- a/kpi/serializers/v2/asset_file.py
+++ b/kpi/serializers/v2/asset_file.py
@@ -251,7 +251,7 @@ class AssetFileSerializer(serializers.ModelSerializer):
         metadata = self._validate_metadata(metadata, validate_redirect_url=True)
 
         redirect_url = metadata['redirect_url']
-        validator = URLValidator()
+        validator = URLValidator(schemes=['http', 'https'])
         try:
             validator(redirect_url)
         except (AttributeError, DjangoValidationError):
--- a/kpi/tests/api/v2/test_api_assets.py
+++ b/kpi/tests/api/v2/test_api_assets.py
@@ -44,6 +44,7 @@ from kpi.tests.base_test_case import (
 )
 from kpi.tests.kpi_test_case import KpiTestCase
 from kpi.tests.utils.mixins import AssetFileTestCaseMixin
+from kpi.tests.utils.mock import patch_ssrf_dns
 from kpi.urls.router_api_v2 import URL_NAMESPACE as ROUTER_URL_NAMESPACE
 from kpi.utils.fuzzy_int import FuzzyInt
 from kpi.utils.hash import calculate_hash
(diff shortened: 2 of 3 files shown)


platreth:
@greptileai

Guitlle (approved):
LGTM, works as expected

PR #7544, opened 2026-09-04
feat(sso): make the signup-closed message configurable DEV-2363

platreth:
### 📣 Summary
Admins can now customize the message shown on the sign-up page when registration is closed.

### 📖 Description
When account registration is turned off, anyone opening the sign-up page sees a fixed "Sign Up Closed" notice. Admins can now replace that text with their own, for example to point people to their organization's SSO login. If nothing is configured, the default notice stays as it was.

### 💭 Notes
- Same pattern as `welcome_message`: the `sitewide_messages` context processor now also looks up a `SitewideMessage` with slug `signup_closed_message` on the signup URL. Localized `signup_closed_message_<lang>` rows take precedence, as before.
- New `account/signup_closed.html` overrides allauth's stock template. The fallback keeps allauth's exact strings so its translations still apply.
- One message for the whole server, not per SSO app, as agreed on the ticket.

### 👀 Preview steps
2. In Django admin > Constance, uncheck `REGISTRATION_OPEN` and save
3. Open `/accounts/signup/` in a private window: the page shows the default "Sign Up Closed" text
4. In Django admin > Sitewide messages, add one with slug `signup_closed_message` and a markdown body
5. Reload `/accounts/signup/`
6. 🔴 [on main] you still see the default text
7. 🟢 [on PR] you see your message instead

### DIFF
--- a/kobo/apps/accounts/templates/account/signup_closed.html
+++ b/kobo/apps/accounts/templates/account/signup_closed.html
@@ -1,0 +1,20 @@
+{% extends "account/base.html" %}
+{% load static %}
+{% load i18n %}
+{% block content %}
+<div class="registration registration--complete">
+  <div class="registration--logo">
+    <a href="/">{% block logo %}{{ block.super }}{% endblock %}</a>
+    {% if not signup_closed_message %}
+      <h1>{% trans "Sign Up Closed" %}</h1>
+    {% endif %}
+  </div>
+  {% if signup_closed_message %}
+    {{ signup_closed_message|safe }}
+  {% else %}
+    <p class="registration__message registration__message--complete">
+      {% trans "We are sorry, but the sign up is currently closed." %}
+    </p>
+  {% endif %}
+</div>
+{% endblock %}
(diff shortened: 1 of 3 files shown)


platreth:
@greptileai

kobo/apps/accounts/templates/account/signup_closed.html
```diff
@@ -0,0 +1,20 @@
+{% extends "account/base.html" %}
+{% load static %}
+{% load i18n %}
+{% block content %}
+<div class="registration registration--complete">
```
rajpatel24: I think this should be `<section>` rather than `<div>` as `div` picks up no card styling here. This renders with no background, centering, padding or max-width, leaving light-grey text directly on the registration background photo. Correct me if I'm wrong or if you're seeing something different.

rajpatel24 (changes requested):

kobo/apps/accounts/templates/account/signup_closed.html
platreth: You're totally right, thanks for catching this!

rajpatel24 (approved):
LGTM

PR #7549, opened 2026-09-05
feat(auth): custom mutator for allauth endpoints DEV-2805

jamesrkiger:
### 📣 Summary
Adds a custom mutator for allauth endpoints to handle expected, non-server-error 4xx responses.

### 💭 Notes
The allauth headless protocol uses HTTP status codes as state signals rather than error indicators. For example, DELETE `/session` always returns 401 to signal the new unauthenticated state after a successful logout. The existing fetchWithAuth mutator throws ServerError on any non-2xx response, which routes these protocol signals to onError rather than onSuccess, making them awkward to handle correctly.

fetchAllauth behaves like fetchWithAuth with two differences: it returns 4xx responses as data instead of throwing, and it omits the Reflux bridge calls (allauth endpoints have no legacy listeners). 5xx responses still throw.

The mutator is wired to allauth endpoints via override.tags in orval.config.js. The tag name ('Authentication (Allauth Headless)') is a raw string from the OpenAPI schema in kpi/utils/spectacular_processing.py. It's not a good tag name and should probably be changed to something more slug-like, but we can deal with that later. If that string changes and orval is re-run without updating the config, the generated hooks silently revert to fetchWithAuth. A wiring test (orval.allauthMutator.wiring.tests.ts) catches this by verifying a representative allauth hook actually calls fetchAllauth.

We may need to refine this mutator down the line when we get more familiar with the types of returns we are going to be dealing with from allauth. My main question for a PR review is if the general approach makes sense.

### 👀 Preview steps
No frontend code consumes allauth endpoints yet.

### DIFF
--- a/jsapp/js/api/orval.allauthMutator.tests.ts
+++ b/jsapp/js/api/orval.allauthMutator.tests.ts
@@ -1,0 +1,119 @@
+var mockedGetCsrfToken: jest.Mock
+jest.mock('#/utils', () => {
+  mockedGetCsrfToken = jest.fn()
+  return { getCsrfToken: mockedGetCsrfToken }
+})
+
+jest.mock('./ServerError', () => {
+  return {
+    ServerError: {
+      new: jest.fn().mockImplementation(async (response: Response) => {
+        const err = Object.assign(new Error(`${response.status}`), { response })
+        return err
+      }),
+    },
+  }
+})
+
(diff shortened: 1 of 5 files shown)


jamesrkiger:
@greptile-apps rereview

magicznyleszek (approved):
LGTM

PR #7553, opened 2026-09-07
fix(trashBin): suspend owner submissions during project deletion DEV-2513

platreth:
### 📣 Summary

Permanently deleting a project or an account no longer fails when that account keeps receiving submissions while the deletion runs.

### 📖 Description

While a project is being permanently deleted, incoming submissions for all projects of that project's owner are paused. Data collection apps get the usual "temporarily unavailable" answer and retry on their own, so no data is lost. The pause is lifted as soon as the deletion ends, whether it succeeded or not. Until now a submission arriving in the middle of a deletion could make the deletion fail and get rescheduled.

### 💭 Notes

- Root cause: the trash task decrements the owner's counters (profile submission count, monthly catch-all counters at XForm `pre_delete`) while live submissions increment the same rows, and Postgres picks the deletion as the deadlock victim. DEV-2511 made that failure retryable, this removes the contention.
- Reuses `UserProfile.submissions_suspended`, enforced in `Instance.check_active()`. Every deletion of the same owner registers a token with a timestamp in a per-user redis hash and only the last live holder releases the flag. Tokens older than the lease (the task soft time limit) count as dead holders, and long loops refresh their token through the yielded `heartbeat()`: `_delete_submissions()` per batch, the storage recount per form.
- Bookkeeping runs under `redis_client.lock()` (`SUBMISSIONS_SUSPENSION_LOCK_TTL`), like the other locks in the repo. If the lock cannot be acquired in time the block runs unlocked and logs an error; if redis is down on entry nothing is suspended, and on exit the flag is released anyway, since a retryable deadlock costs less than an owner locked out of data collection.
- `update_attachment_storage_bytes` is a holder too, so the weekly recount can no longer unlock a running deletion. `fix_stale_submissions_suspended_flag` now calls `release_orphaned_suspensions()`, which frees only profiles with no live holder. That made the old heartbeat hash dead code, so nothing writes it any more. It is still read, so a recount worker on the previous release keeps its lock during a rolling deployment; the read and its key carry a TODO to remove them in the next release. `populate_submission_counters` still releases everything at startup, it is a one-off command.
- Only the deployed branch of `delete_asset()` is wrapped: account deletion is covered per project, drafts and blocks skip it, attachment trash does not touch these counters.
- The deadlock itself needs racing transactions and was not reproduced locally. The mechanism was: with the fix, a submission probe for the owner's other form during `_delete_submissions()` raises `TemporarilyUnavailableError`, without it the probe passes.
- `ProjectTrashTestCase` got a Mongo `tearDown`. Postgres is rolled back between tests, Mongo is not, and the new test that runs a real deletion tripped on a document left behind by `test_move_to_trash`.

### 👀 Preview steps

Back end only, nothing changes in the UI.

1. ℹ️ have the local stack running
2. In the kpi container run the new tests:
   ```
   pytest kobo/apps/trash_bin/tests/test_utils.py -k owner_submissions
   ```
3. 🟢 [on PR] `3 passed`
4. 🔴 [on main] the wiring does not exist there. To see the failure mode, put main's `kobo/apps/trash_bin/utils/project.py` on the branch and rerun step 2: `test_owner_submissions_suspended_during_deletion` fails with `assert False is True`, meaning submissions were still accepted while the project was being deleted. Restore the file afterwards.
5. Live check, optional: while an `empty_project` task runs for your user, `UserProfile.objects.get(user__username='<you>').submissions_suspended` is `True` and a Collect or Enketo submission to any of your forms gets the temporary-unavailable response until the task ends.

### DIFF
--- a/kobo/apps/openrosa/apps/logger/management/commands/update_attachment_storage_bytes.py
+++ b/kobo/apps/openrosa/apps/logger/management/commands/update_attachment_storage_bytes.py
@@ -1,17 +1,19 @@
 from __future__ import annotations
 
-import time
+from collections.abc import Callable
 
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.core.management.base import BaseCommand
 from django.db.models import OuterRef, Subquery, Sum, Value
 from django.db.models.functions import Coalesce
-from django_redis import get_redis_connection
 
-from kobo.apps.openrosa.apps.logger.constants import SUBMISSIONS_SUSPENDED_HEARTBEAT_KEY
 from kobo.apps.openrosa.apps.logger.models.attachment import Attachment
 from kobo.apps.openrosa.apps.logger.models.xform import XForm
+from kobo.apps.openrosa.apps.logger.utils.suspension import (
+    release_orphaned_suspensions,
+    suspend_submissions,
+)
 from kobo.apps.openrosa.apps.main.models.user_profile import UserProfile
 from kpi.utils.django_orm_helper import UpdateJSONFieldAttributes
 
@@ -28,7 +30,6 @@ class Command(BaseCommand):
         self._verbosity = 0
         self._force = False
         self._sync = False
-        self._redis_client = get_redis_connection()
 
     def add_arguments(self, parser):
         parser.add_argument(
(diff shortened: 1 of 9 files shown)


platreth:
@greptileai

platreth:
@greptileai Look at the last commit and update the score accordingly

platreth:
@greptileai check again

kobo/apps/trash_bin/utils/project.py
```diff
@@ -23,8 +24,9 @@ def delete_asset(request_author: settings.AUTH_USER_MODEL, asset: Asset):
     project_exports = []
 
     if asset.has_deployment:
-        _delete_submissions(request_author, asset)
-        asset.deployment.delete()
+        with suspend_submissions(asset.owner):
+            _delete_submissions(request_author, asset)
```
Guitlle: The _delete_submissions function should also use the heartbeat internally, no? I saw it has a potentially long loop that deletes data in batches

kobo/apps/openrosa/apps/logger/utils/suspension.py
```diff
@@ -0,0 +1,158 @@
...
+):
+    """
+    Drop this holder's token and release the flag if no live holder remains.
+
+    Without Redis there is no way to tell whether another holder is alive, so
+    the flag is released anyway: a deadlock is retryable, an owner who cannot
+    collect data until `fix_stale_submissions_suspended_flag` runs is not.
+    """
+    try:
+        with _mutex(redis_client, holders_key):
```
Guitlle, line 155: if _mutex raises a LockError exception, the flag is going to be set to False regardless of any remaining live holders, because LockError inherits from RedisError.

kobo/apps/openrosa/apps/logger/utils/suspension.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def heartbeat():
+        _register_holder(redis_client, holders_key, token, lease, user.username)
+
+    # Redis first: if it fails, nothing has been suspended yet
+    with _mutex(redis_client, holders_key):
+        heartbeat()
+    try:
+        UserProfile.objects.get_or_create(user_id=user.pk)
+        _set_flag(user, True)
```
Guitlle: this _set_flag call should be inside the mutex, to avoid the race condition where a task is about to end and another task is just getting started

kobo/apps/openrosa/apps/logger/utils/suspension.py
```diff
@@ -0,0 +1,158 @@
...
+        for token, registered_at in redis_client.hgetall(holders_key).items()
+        if int(registered_at) + lease <= now
+    ]
+    if dead:
+        redis_client.hdel(holders_key, *dead)
+    return redis_client.hlen(holders_key)
+
+
+@contextmanager
+def _mutex(redis_client, holders_key: str):
```
Guitlle, line 107: what do you think about using something like `redis_client.lock(lock_key, timeout=10, sleep=0.05, blocking_timeout=30)` instead?

kobo/settings/base.py
```diff
@@ -1699,7 +1699,8 @@ def __init__(self, *args, **kwargs):
         ),
         'schedule': crontab(minute='*/15', hour='2-5', day_of_week=0),
         'description': (
-            'Unlock accounts locked by `sync_storage_counters` task'
+            'Unlock accounts left suspended by a storage recount or a trash bin'
```
Guitlle, line 1702: This task should be using the new `release_orphaned_suspensions` function, from your `suspension.py` but it wasn't changed in this PR

Guitlle (changes requested):

kobo/apps/openrosa/apps/logger/management/commands/update_attachment_storage_bytes.py
```diff
@@ -203,52 +155,23 @@ def _get_queryset(self, profile_queryset, username):
...
-        # Release any locks on the users' profile from getting submissions
+        # Release the profiles a previous run left suspended, without touching
+        # the ones held by a running deletion
         if self._verbosity > 1:
             self.stdout.write('Releasing submission locks…')
 
-        UserProfile.objects.all().update(submissions_suspended=False)
+        orphaned = release_orphaned_suspensions()
+        if self._verbosity > 1:
+            self.stdout.write(f'Released submission locks: {len(orphaned)}')
```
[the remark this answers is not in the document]
platreth: These are management command output and log lines, which are operator-facing and not translated anywhere in this repo.

platreth:
@greptile update the score

Guitlle (approved):
LGTM

PR #7555, opened 2026-09-08
fix(massemails): fix edge cases for one-off mass email campaigns DEV-2737

Guitlle:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Fixes edge cases where one-off mass email campaigns (`frequency = -1`) could be accidentally resent if remaining enqueued records transitioned to STALE or if the Celery worker process terminated before the campaign was finalized

### 📖 Description
* `get_users_for_config()` now excludes users who already have a terminal status (SENT, FAILED, STALE) under that one-off config
* Any one-off campaign that has already run and has 0 enqueued records is disabled immediately, even if all remaining records went STALE. `send_emails()` now filters on `jobs__isnull=False` instead of `pk__in=sender.config_ids`. 
* If an unclosed one-off config has records and no eligible recipients remaining, `generate_mass_email_user_lists()` marks `live = False` directly without creating an empty `MassEmailJob` re-enqueueing records
* Included new unit tests to validate that the edge cases are handled correctly

### 👀 Preview steps
1. ℹ️ have two user accounts (one of them should be inactive)
2. Create a mass email config with `frequency = -1`, `query = 'test_users'` and `live = True`
```
from kobo.apps.mass_emails.tasks import generate_mass_email_user_lists, send_emails

cache.clear()

# First call to send_emails to 
send_emails()

config = MassEmailConfig.objects.get(id=<id>)
# It has to be 1 day old to be processed
config.date_created = timezone.now() - timedelta(days=1)
config.save()

generate_mass_email_user_lists()

# 🟢 [on PR] Config is turned off immediately
# The inactive account should have status = STALE
assert not config.live
assert MassEmailRecord.objects.get(user__username='<inactive username>').status == EmailStatus.STALE


send_emails()
```
3. Wait for the next 15-minute window  and clear cache to simulate next run and verify that no new records are created:
```
# Make config be live (in a real scenario this would happen if the job was interrupted)
config.live = True
config.save()

cache.clear()
generate_mass_email_user_lists()
config.refresh_from_db()

# 🟢 [on PR] Config is automatically turned off without creating new records
config.refresh_from_db()
assert not config.live
assert MassEmailRecord.objects.filter(email_job__email_config=config).count() == 2
```


kobo/apps/mass_emails/tasks.py
```diff
@@ -569,7 +574,18 @@ def generate_mass_email_user_lists():
...
             processed_configs.add(email_config.id)
-
+        elif (
+            email_config.frequency == -1
+            and email_records.exists()
+            and not get_users_for_config(email_config)
+        ):
+            logging.info(
+                f'Completing unclosed one-time email config {email_config.id} '
+                f'({email_config.name}) as all records have been processed.'
```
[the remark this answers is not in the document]
Guitlle: Log messages should not be translated

kobo/apps/mass_emails/tasks.py
```diff
@@ -516,7 +515,13 @@ def get_users_for_config(email_config):
         user_ids = [getattr(user, 'id', user) for user in users]
 
     if email_config.frequency == -1:
-        return user_ids
+        terminal_recipients = set(
```
noliveleger: I think we should use the join to avoid loading all MassEmailRecord into a set. 

```python
     users = email_config.get_users_queryset()
     if email_config.frequency == -1 and isinstance(users, QuerySet):                                                                                             users = users.exclude(                                                                                                                              pk__in=MassEmailRecord.objects.filter(
                 email_job__email_config=email_config,                                                                                                           status__in=[
                     EmailStatus.SENT,
                     EmailStatus.FAILED,
                     EmailStatus.STALE,                                                                                                                          ],
             ).values('user_id')                                                                                                                         )

     if isinstance(users, QuerySet):
         user_ids = list(users.values_list('id', flat=True))
     else:
         user_ids = [getattr(user, 'id', user) for user in users]

     if email_config.frequency == -1:
         return user_ids
```

kobo/apps/mass_emails/tasks.py
noliveleger: is it used somewhere? Seems dead code

kobo/apps/mass_emails/tasks.py
```diff
@@ -51,12 +51,13 @@
 SEND_EMAILS_SOFT_LIMIT_BUFFER = 10
 
 
-def enqueue_mass_email_records(email_config):
+def enqueue_mass_email_records(email_config, user_ids=None):
     """
     Creates a email job and enqueues email records for users based on query
     """
     job = MassEmailJob.objects.create(email_config=email_config)
```
noliveleger: A one-off config with no records and no eligible recipients still falls into the else branch below and creates an empty MassEmailJob here, before the empty-user_ids check turns the config off. 

What about creating the job only once user_ids is known to be non-empty?

kobo/apps/mass_emails/tasks.py
```diff
@@ -51,12 +51,13 @@
 SEND_EMAILS_SOFT_LIMIT_BUFFER = 10
 
 
-def enqueue_mass_email_records(email_config):
+def enqueue_mass_email_records(email_config, user_ids=None):
```
noliveleger: Now that the only production caller passes user_ids explicitly, the None fallback is exercised only by `test_enqueue_creates_records_via_id_path`. It is also the one path that can produce a duplicate send: get_users_for_config() excludes terminal recipients but not ENQUEUED ones, and the unique constraint is (email_job, user), so calling it on a config that already has a pending queue creates a second record in a new job. Making the argument required (and updating that test) would remove the trap.

kobo/apps/mass_emails/tasks.py
```diff
@@ -565,11 +570,22 @@ def generate_mass_email_user_lists():
                 f'enqueued records.'
             )
             processed_configs.add(email_config.id)
+            continue
+
+        user_ids = get_users_for_config(email_config)
 
+        if email_config.frequency == -1 and email_records.exists() and not user_ids:
```
noliveleger: A one-off config with no records at all and no eligible recipients still goes through the else branch, where enqueue_mass_email_records() creates a MassEmailJob before finding out user_ids is empty. It ends at live = False just like the new branch does, but leaves an empty job row behind, which is the thing this PR set out to avoid.                                                                                     

Dropping and email_records.exists() from the condition sends both cases through the new branch and removes that empty job, and it also saves a second exists() round trip on the same table one line after the first one. The only thing lost is the distinct "No recipients found for one-time email config" log line; if that distinction is worth keeping, email_records.exists() can pick the log message inside the branch instead of picking the branch itself.

noliveleger (changes requested):

kobo/apps/mass_emails/tasks.py
Guitlle: no,

kobo/apps/mass_emails/tasks.py
Guitlle: interesting edge case

kobo/apps/mass_emails/tasks.py
Guitlle: refactored the logic to fix this

kobo/apps/mass_emails/tasks.py
```diff
@@ -569,16 +566,34 @@ def generate_mass_email_user_lists():
                 f'enqueued records.'
             )
             processed_configs.add(email_config.id)
+            continue
 
-        else:
-            try:
```
platreth, line 574: Shouldn't we keep a warning? Now it's only an info?

kobo/apps/mass_emails/tasks.py
```diff
@@ -514,13 +498,26 @@ def get_users_for_config(email_config):
     # `get_users_queryset()` returns a QuerySet for real queries but a plain list
     # (e.g. `[]`) from the default fallback; normalize to a list of ids without
     # materializing full User instances.
+    if email_config.frequency == -1 and isinstance(users, QuerySet):
```
platreth, line 501: I tried the resume path with a recipient who deleted their account and it turns the config off instead of enqueueing the remaining users. 

What do you think about adding `user__isnull=False` to the subquery filter or check `Exist(...)`

And maybe a small test for this case?

platreth (changes requested):

kobo/apps/mass_emails/tasks.py
Guitlle: The warning was for the integrity error, it's still there but no on line 594, the info messages are used to log that there are no more recipients in the email config

platreth (approved):
LGTM

PR #7556, opened 2026-09-08
fix(trashBin): record a real error when a trash deletion fails without one DEV-2512

platreth:
### 📣 Summary
Deletions from the trash bin that fail now always say why, and the ones that failed silently in the past get one more automatic try.

### 📖 Description
Some projects, accounts and attachments failed to be deleted from the trash bin with a blank error. Superusers could see they had failed, but not why, and nothing retried them. Every failure now records a message, and the deletions that failed with a blank one are put back in the queue so they run again at the usual pace.

### 💭 Notes
- Root cause: `str()` of an exception raised without arguments is `''`. The main producer is `MissingXFormException`, raised bare by the deployment backend and re-raised on purpose in `_delete_submissions()`: a first deletion attempt dies after the XForm is gone but before the asset is, then every later attempt fails at that same spot with a blank message. A bare `TrashTaskInProgressError` does the same on the retry path.
- Fix: `trash_bin_task_failure()` and `trash_bin_task_retry()` fall back to `repr()` when `str()` is empty, and `MissingXFormException` gets a default message like its sibling `InvalidXFormException`.
- LRM `0032` flips `FAILED` rows with an empty or missing `failure_error` back to `IN_PROGRESS`, same shape as `0031` from DEV-2511. `task_restarter` then re-runs them at most `MAX_RESTARTED_TASKS` per run, so the queue cannot be flooded. Rows stuck on the missing-XForm case will fail again, but this time with a readable message, so we can decide what to do with them from real data.

### 👀 Preview steps
1. ℹ️ have a superuser and a project
2. In the Django shell, move the project to trash, then call `trash_bin_task_failure(ProjectTrash, args=[pt.pk], exception=MissingXFormException())`, which is what Celery does when `empty_project` dies on a missing XForm
3. 🔴 [on main] the trash object is `failed` and `metadata['failure_error']` is `''`, and nothing ever restarts it
4. 🟢 [on PR] `failure_error` reads `Deployment links to a KoboCAT XForm that no longer exists`
5. 🟢 [on PR] set a trash row to `failed` with an empty `failure_error` and run job `0032`: the row is back to `in_progress`, while a row with a real error stays `failed`

<details>
<summary>Shell snippet for steps 2 to 4 (rolls everything back)</summary>

```python
from django.contrib.auth import get_user_model
from django.db import transaction
from kobo.apps.trash_bin.models.project import ProjectTrash
from kobo.apps.trash_bin.utils.trash import move_to_trash, trash_bin_task_failure
from kpi.exceptions import MissingXFormException
from kpi.models import Asset

owner = get_user_model().objects.filter(is_superuser=True).first()
try:
    with transaction.atomic():
        asset = Asset.objects.create(owner=owner, name='DEV-2512', asset_type='survey')
        move_to_trash(owner, [{'pk': asset.pk, 'asset_uid': asset.uid, 'asset_name': asset.name}], 1, 'asset')
        pt = ProjectTrash.objects.get(asset=asset)
        trash_bin_task_failure(ProjectTrash, args=[pt.pk], exception=MissingXFormException())
        pt.refresh_from_db()
        print(pt.status, repr(pt.metadata.get('failure_error')))
        raise RuntimeError('rollback')
except RuntimeError:
    pass
```
</details>

### DIFF
--- a/kpi/exceptions.py
+++ b/kpi/exceptions.py
@@ -195,7 +195,10 @@ class MailerConnectionSessionLimitError(MailerError):
 
 
 class MissingXFormException(Exception):
-    pass
+    def __init__(
+        self, message=t('Deployment links to a KoboCAT XForm that no longer exists')
+    ):
+        super().__init__(message)
 
 
 class NotSupportedFormatException(Exception):
--- a/kpi/exceptions.py
+++ b/kpi/exceptions.py
@@ -195,8 +195,9 @@ class MailerConnectionSessionLimitError(MailerError):
 
 
 class MissingXFormException(Exception):
+    # Not translated on purpose: only stored as an ops diagnostic and grepped
     def __init__(
-        self, message=t('Deployment links to a KoboCAT XForm that no longer exists')
+        self, message='Deployment links to a KoboCAT XForm that no longer exists'
     ):
         super().__init__(message)
 
--- a/kobo/apps/long_running_migrations/jobs/0032_restart_empty_error_trash_failures.py
+++ b/kobo/apps/long_running_migrations/jobs/0032_restart_empty_error_trash_failures.py
@@ -35,9 +35,12 @@ def run():
             if not batch:
                 break
 
-            # `date_modified` is deliberately left untouched: these objects have
-            # not been updated for a while, which is what makes `task_restarter`
-            # treat them as stuck and restart them
-            model.objects.filter(pk__in=batch).update(status=TrashStatus.IN_PROGRESS)
-            released += len(batch)
+            # The predicates are reapplied so that a row restarted by a superuser
+            # in the meantime, which may have failed again with a real error by
+            # now, keeps its newer state. `date_modified` is deliberately left
+            # untouched: these objects have not been updated for a while, which
+            # is what makes `task_restarter` treat them as stuck and restart them
+            released += model.objects.filter(
+                empty_errors, pk__in=batch, status=TrashStatus.FAILED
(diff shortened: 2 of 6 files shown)


platreth:
@greptileai

kpi/exceptions.py
```diff
@@ -195,7 +195,10 @@ class MailerConnectionSessionLimitError(MailerError):
 
 
 class MissingXFormException(Exception):
-    pass
+    def __init__(
+        self, message=t('Deployment links to a KoboCAT XForm that no longer exists')
```
rajpatel24: failure_error is an ops/superuser diagnostic, not end-user UI, so I think a plain, untranslated string would be safer and easier to grep in logs. What do you think?

rajpatel24 (changes requested):

kpi/exceptions.py
platreth: Agreed, changed!

kpi/exceptions.py
```diff
@@ -195,7 +195,11 @@ class MailerConnectionSessionLimitError(MailerError):
 
 
 class MissingXFormException(Exception):
-    pass
+    # Not translated on purpose: only stored as an ops diagnostic and grepped
+    def __init__(
```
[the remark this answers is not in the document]
platreth: The reason I did that is on the thread with raj. Look at it

rajpatel24 (approved):
LGTM

PR #7559, opened 2026-09-08
ci(releases): set Linear release name explicitly DEV-2787

Akuukis:
### 💭 Notes

Let's name Linear releases by the patch version to match the release version.


noliveleger (approved):
LGTM

PR #7561, opened 2026-09-08
feat(frontend): improve error handling DEV-2795

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Error messages no longer dump raw server output: you either get the message the server actually wrote, or a plain "An error occurred".

### 💭 Notes

When something went wrong on the server, a notification could contain a whole error page or a traceback. Same for a few screens that print the error instead of toasting it. Those now fall back to a short generic message, with the full response kept in the browser console for support.

Validation errors read as sentences too. A form that failed on a field used to produce `{"name":["This field is required."]}`; it now says `name: This field is required.`

Changes here:
- `api/flattenErrorBody.ts` (new) — parsed body into one line - replaces the old "unwrap a lone `detail` key" trick in `notify()`
- `api/ServerError.ts` — a body that fails `JSON.parse` no longer lands in `detail`, only on `parsedResponse` for logging
- `api/onErrorDefaultHandler.ts` — `getApiErrorMessage` is now a single call to `flattenErrorBody`
- `api.ts` (`handleApiFail`), three things:
  - the docstring already claimed HTML was suppressed, but only the Werkzeug `.errormsg` pluck existed, so production pages fell right through. Order is now Werkzeug pluck (development only) → JSON body → generic
  - Sentry gets `${status} ${statusText}` as the title with the body in `extra`
- `utils.ts` — `getErrorMessage()` is now flattened and escaped
- Fixed screens that print the body instead of toasting it
- Tests added

### 👀 Preview steps

1. ℹ️ have an account and a project with some submissions, and browser DevTools open on the Network tab
2. trigger any error notification you can reproduce — Project → Settings → Sharing and inviting a username that doesn't exist is an easy one
3. 🟢 notice the notification still uses the server's own wording; genuine error messages are untouched
4. ℹ️ now play broken server: right-click that failed request in Network → **Override content**, replace the saved body with `<html><body><h1>Server error (500)</h1>Something broke</body></html>`, then repeat step 2
5. 🔴 [on main] the notification prints the page — headings, angle brackets and all
6. 🟢 [on PR] it reads "An error occurred", with the full body waiting in the console

### DIFF
--- a/jsapp/js/api/getDisplayableErrorText.ts
+++ b/jsapp/js/api/getDisplayableErrorText.ts
@@ -1,0 +1,41 @@
+/**
+ * Keeps response bodies that aren't messages out of the UI.
+ */
+
+/**
+ * An allowlist, because real messages contain angle brackets, like `'must be < 5'`. A tag name has to end the way one
+ * does in markup, or an address such as `<a@b.com>` would read as a link.
+ */
+const HTML_TAG_REGEX =
+  /<\/?(?:!doctype|html|head|body|title|meta|link|script|style|div|span|p|h[1-6]|a|img|br|hr|table|thead|tbody|tr|td|th|ul|ol|li|dl|dt|dd|pre|code|form|input|button|label|iframe|section|article|header|footer|main|nav|strong|em|b|i|small|font|center|blockquote|textarea)(?=[\s/>])[^>]*>/i
+
+/** Whether `text` is page output rather than a message. */
+export function containsHtmlMarkup(text: string): boolean {
+  return HTML_TAG_REGEX.test(text)
+}
+
+/**
+ * The body when it's fit to show a user, or `null` when it's error-page output and the caller should fall back to its
+ * own message. Pass the *raw* body only - structured errors go through `flattenErrorBody`.
+ */
+export function getDisplayableErrorText(body: unknown, status?: number): string | null {
+  if (typeof body !== 'string') {
+    return null
+  }
+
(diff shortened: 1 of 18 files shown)


jsapp/js/api/getDisplayableErrorText.ts
```diff
@@ -0,0 +1,41 @@
...
+/** Whether `text` is page output rather than a message. */
+export function containsHtmlMarkup(text: string): boolean {
+  return HTML_TAG_REGEX.test(text)
+}
+
+/**
+ * The body when it's fit to show a user, or `null` when it's error-page output and the caller should fall back to its
+ * own message. Pass the *raw* body only - structured errors go through `flattenErrorBody`.
+ */
+export function getDisplayableErrorText(body: unknown, status?: number): string | null {
```
jamesrkiger, line 21: Are you aware of cases where we're getting HTML on non-5xx errors? The backend shouldn't be returning HTML on non-5xx errors, so if there are known cases I'd rather treat that as a backend issue than code around it with a regex. And if there aren't, we definitely don't need to code around it.

Ideally I'd like to see the 5xx check moved up the call chain (to ServerError or getApiErrorMessage) and then just assume we're dealing with JSON for anything that passes it — which I think would let this file go away entirely.

jamesrkiger (changes requested):

jamesrkiger (approved):
LGTM

PR #7567, opened 2026-09-09
fix(attachments): accept media questions from every version a submission spans DEV-2766

noliveleger:
### 📣 Summary

Files attached to a submission are no longer dropped when the form's question was renamed while the form was open.

### 📖 Description

Attaching a file to a form that was renamed or reorganised in the meantime used to lose that file. The submission went through, the filename stayed in the data, but the file itself was gone with no error shown anywhere. It happened whenever the question moved between the moment the form was loaded and the moment the record reached the server: a form edited and redeployed in another tab, or a record filled in offline and sent days later.

Those files are kept now. A submission is matched against every version of the form it has been through, not only against the one deployed today.

### 💭 Notes

`get_soft_deleted_attachments()` built its list of accepted media questions from the currently deployed XForm alone, while the submission's nodes carry the names of the version the client actually rendered. Nothing matched, so every attachment of the record was soft deleted moments after being saved, with a 201 still returned to the client. The versions recorded in `meta/formVersions` by #7543 are now resolved through KPI and added to the XForm's own. `soft_delete_orphan_attachments` goes through the same function and is fixed along with it. Nothing to run, no migration, no backfill; records already soft deleted stay that way, that is DEV-2804.

**Known limitation, worth knowing before testing.** When a question's XPath changes, whether a rename or a move in or out of a group, and the record is then edited, Enketo posts back both shapes: the deployed one it renders, and the record's original nodes, which it appends at the end of the document. Nothing on the server ties two names of the same question together across a rename, so the file the edit has just replaced cannot be told apart from a file that simply has no home in the current form. Both are kept, and the stale one stays visible on the submission and has to be deleted by hand. It behaves like other questions, but this could be improved with another PR if we implement a specific unique identifier per question. (`$kuid` is not). 

That is a deliberate choice rather than an oversight. The alternative is to ignore the older names as soon as the deployed ones are answered, which is what release does. It retires the replaced file correctly, but it also retires any other file whose own question changed XPath in the same redeployment and whose deployed node comes back empty because Enketo could not carry the old value over. That second file was never touched by anyone, and it is not hypothetical: it is the shape of the record this was reproduced on, where `second_image` came back empty and the original `image_2` survived only as a leftover node. Neither behaviour is right and no rule can be, since the pairing is undecidable. Keeping is the safer of the two, since a stale file stays visible and can be removed while a file that vanishes is never noticed. Ordinary edits are unaffected either way: a replaced file is retired and the others are left alone. The Enketo behaviour behind all of this deserves its own ticket, it is not introduced here.

### 👀 Preview steps

1. ℹ️ have a deployed form with an image question named `upload_image`
2. open the form in Enketo, online-offline mode, and attach an image
3. in another tab, open the same form in the form builder, rename the question to `upload_image_edited`, save and deploy
4. go back to the Enketo tab and press submit
5. 🔴 [on release] open the submission in the API: the filename is in the data but `_attachments` is empty, the file is gone
6. 🟢 [on PR] the attachment is there, with a `question_xpath` of `upload_image`
7. edit the submission and replace the file on a question that was **not** renamed
8. 🟢 the replaced file is soft deleted as before, the others are untouched

### DIFF
--- a/kpi/models/asset.py
+++ b/kpi/models/asset.py
@@ -629,16 +629,22 @@ class Asset(
         """
 
         if version:
-            content = version.to_formpack_schema()['content']
             cache_key = f'attachment_xpaths:{self.uid}:{version.uid}'
         else:
-            content = self.content
             cache_key = f'attachment_xpaths:{self.uid}:no-version'
 
         cached_xpaths = cache.get(cache_key)
 
         if cached_xpaths is not None:
             return cached_xpaths
+
+        # Read the content only once the cache has been given its chance:
+        # expanding a version's content is the expensive part of this method,
+        # and `self.content` may be a deferred field
+        if version:
+            content = version.to_formpack_schema()['content']
+        else:
+            content = self.content
 
         survey = content['survey']
 
@@ -672,6 +678,31 @@ class Asset(
         cache.set(cache_key, xpaths_list, timeout=settings.ATTACHMENT_XPATHS_CACHE_TTL)
 
         return xpaths_list
+
+    def get_attachment_xpaths_from_version_uids(self, version_uids: list[str]) -> list:
+        """
+        Get the attachment xpaths of several form versions, merged.
+
+        Takes uids, where `get_attachment_xpaths_from_version()` takes an
+        `AssetVersion` and holds the per-version Redis cache both share.
+
+        Versions are merged, never superseded: a question renamed between two
+        of them lives at both xpaths, and keeping only the newest is what makes
+        an attachment collected under the older name unreachable.
+        """
+
+        # A `version_content` weighs several megabytes on a large form.
+        # `iterator()` keeps one version at a time in memory, where iterating
+        # the queryset itself would fill its result cache with all of them
+        versions = self.asset_versions.filter(
+            uid__in=version_uids, deployed=True
+        ).iterator()
+
+        xpaths = set()
+        for version in versions:
+            xpaths.update(self.get_attachment_xpaths_from_version(version) or [])
+
+        return list(xpaths)
 
     def get_filters_for_partial_perm(
         self, user_id: int, perm: str = PERM_VIEW_SUBMISSIONS
--- a/kpi/models/asset.py
+++ b/kpi/models/asset.py
@@ -600,6 +600,10 @@ class Asset(
         cached_xpaths = cache.get(cache_key)
 
         if cached_xpaths is not None:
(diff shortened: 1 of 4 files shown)


kpi/models/asset.py
platreth: I checked and after one call the key for a version with an image question is still None, a no-media version caches [], and get_attachment_xpaths_from_version_uids() re-runs the full query on every call!

What about something like this 
```suggestion
        if xpaths := _get_xpaths(survey):
    cache.set(cache_key, xpaths, timeout=settings.ATTACHMENT_XPATHS_CACHE_TTL)
    return xpaths
```

noliveleger: Good catch, I also found another optimization to add (will be in the next commit).

kpi/models/asset.py
```diff
@@ -625,23 +629,26 @@ def get_attachment_xpaths_from_version(
...
+        invalidates these keys. Reading it is also the cheap case, with no
+        `to_formpack_schema()` to pay for.
         """
 
+        cache_key = None
+
         if version:
             cache_key = f'attachment_xpaths:{self.uid}:{version.uid}'
-        else:
-            cache_key = f'attachment_xpaths:{self.uid}:no-version'
```
noliveleger, line 636: Removed this option because submissions should always have `__version__` node in their XML.  In some cases, `version` can be `None` but it would mean that version has been deleted (from the shell, not possible from the API) OR some 3rd party script send the XML without the `__version__` node. Not the case with Enketo and Collect.

kpi/models/asset.py
```diff
@@ -600,6 +600,10 @@ def get_all_attachment_xpaths(self, only_cached_data: bool = False) -> list | No
         cached_xpaths = cache.get(cache_key)
 
         if cached_xpaths is not None:
+            # Memoize the hit too, otherwise every call on this instance pays a
```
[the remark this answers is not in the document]
noliveleger: The memo predates this change: `_all_attachment_xpaths` was already set on the cache-miss path, and that miss is exactly what a first read after a redeploy hits, since the cache key embeds `latest_deployed_version_uid`. So the sequence described already returns a memoized value today. This only feeds the same memo from the hit path.                                                                                              
                                                                                                                                        
Besides, no caller reads attachment xpaths and then deploys on the same instance: the three of them are submission reads in openrosa_backend, plus LRM 0024. `refresh_from_db()` clears it.

However, I'll reset it in `Asset.save()` in the next commit.

platreth (approved):

PR #7572, opened 2026-09-09
feat(auth): confirm email address screen DEV-1849

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

The link in your sign-up email opens a redesigned screen that confirms your address and then takes you straight into Kobo, or to the login form if your instance asks you to sign in again.

### 💭 Notes

Route for the activation link from the sign-up email. Not hooked up yet - `HEADLESS_FRONTEND_URLS` is unset, so allauth still puts the Django view's URL in the email. Also `/auth` is behind the `authRedesignEnabled` flag.

Changes here:

- `ActivateAccountRoute.tsx` — `/auth/activate/:key`. Looks the key up, asks for confirmation, then sends people either into Kobo or to the login form. `ACCOUNT_LOGIN_ON_EMAIL_CONFIRMATION` decides which, and allauth reports the second case as a 401 that its own docs call a success, so it is treated as one.
- `allauth.mocks.ts` — the four email-verify handlers, since a rejected key arrives as data rather than as an error

### 👀 Preview steps

1. ℹ️ be signed out, on an instance where you can read the outgoing mail
2. sign up, open the activation email, and copy the key out of the link: it is the part after `/accounts/confirm-email/`
3. open `/#/auth/activate/<key>?ff_authRedesignEnabled=true`
4. 🟢 "Confirm E-mail Address", naming your address and username
5. click `Confirm`
6. 🟢 "Email address confirmed", with a button into Kobo — or a `Log in` button, if the instance does not sign people in on confirmation
7. reload that same URL
8. 🟢 "Activation Failed", because activation keys are strictly one and done
9. open `/#/auth/activate/nonsense?ff_authRedesignEnabled=true`
10. 🟢 the same failure screen, with no hint about which part of the link was wrong


jsapp/js/auth/ActivateAccountRoute/ActivateAccountRoute.tsx
```diff
@@ -0,0 +1,199 @@
...
+        {t('Confirm E-mail Address')}
+      </Title>
+      <Text>{t('Checking your activation link…')}</Text>
+    </Stack>
+  )
+}
+
+interface ConfirmPromptPanelProps {
+  email: string
+  /** allauth's `user_display()` value, the same one its Django template shows. */
```
jamesrkiger:
```suggestion
  /** Corresponds to username */
```

jamesrkiger:
Can we keep the size of the activation card stable across the various states? Right now it shrinks and expands vertically when it's transitioning through loading states.

jamesrkiger (changes requested):

jamesrkiger (approved):
LGTM

PR #7575, opened 2026-09-09
feat(singleSubmissionView): add Translate & analyze action for text responses DEV-2373

pauloamorimbr:
### 📣 Summary

Text responses in the Single Submission Modal now have a "..." menu with a "Translate & analyze" action, opening Processing for that response — same as audio already offers.

### 💭 Notes

- Audio's existing "Open" button in this same dialog was replaced by the same "Translate & analyze" menu entry, in its own existing "..." attachment menu, so both response types now share one consistent action, icon, and label.
- Extracted the shared "..." menu (`MoreActionsMenu`) and its optional "Translate & analyze" entry (`processingAction` prop) so both the attachment dropdown and the new text menu use the same implementation, rather than duplicating the trigger/label wiring.
- The audio player in this dialog now fills the row's full width, matching the layout of other response types.
- The text menu resolves the question's static survey xpath by name rather than using the repeat-instance-aware path used for matching media attachments, so the action opens Processing at the correct route even for a text question inside a repeat group.
- Depends on `isNlpSupported`/`isAudioQuestionType`/`isTextQuestionType` from DEV-2288, which this branch is stacked on top of (this PR's base is that branch, not `main`).

### 👀 Preview steps

1. ℹ️ have a project deployed with at least one audio question and one text question, and a submission answering both
2. append `?ff_nlpTextActionsEnabled=true` to the URL and reload
3. open a submission from the table to bring up the Single Submission Modal
4. 🔴 [on main] the text response has no "..." menu at all; the audio response has a separate "Open" button next to its "..." menu
5. 🟢 [on PR] both the audio and the text response show a "..." menu with a "Translate & analyze" item (pencil + star icon)
6. click "Translate & analyze" on the text response and confirm it opens Processing for that exact question
7. do the same for the audio response, and confirm the audio player now spans the full row width
8. reload without the `ff_nlpTextActionsEnabled` param (or use a private window) and confirm the text response's menu no longer appears (audio's "..." menu, with Download/Delete, still does)

### DIFF
--- a/jsapp/js/components/common/MoreActionsMenu.tsx
+++ b/jsapp/js/components/common/MoreActionsMenu.tsx
@@ -1,0 +1,47 @@
+import { Menu } from '@mantine/core'
+import { IconPencilStar } from '@tabler/icons-react'
+import type { ReactNode } from 'react'
+import ActionIcon from '#/components/common/ActionIcon'
+import KoboIcon from '#/components/common/KoboIcon'
+import { goToProcessing } from '#/components/processing/routes.utils'
+
+interface MoreActionsMenuProps {
+  children?: ReactNode
+  className?: string
+  processingAction?: {
+    assetUid: string
+    xpath: string
+    submissionEditId: string
+  }
+}
+
+/** A "…" button that opens a dropdown menu of actions. */
+export default function MoreActionsMenu({ children, className, processingAction }: MoreActionsMenuProps) {
+  return (
+    <Menu withinPortal={false} closeOnClickOutside closeOnItemClick position='bottom-end'>
+      <Menu.Target>
+        <span className={className} style={{ position: 'relative' }}>
+          <ActionIcon size='md' variant='transparent' iconName='more' />
+        </span>
+      </Menu.Target>
+
+      <Menu.Dropdown>
+        {processingAction && (
+          <>
+            <Menu.Item
+              leftSection={<KoboIcon icon={IconPencilStar} size={16} />}
+              onClick={() =>
+                goToProcessing(processingAction.assetUid, processingAction.xpath, processingAction.submissionEditId)
+              }
+            >
+              {t('Translate & analyze')}
+            </Menu.Item>
+            {children && <Menu.Divider />}
+          </>
+        )}
+
+        {children}
+      </Menu.Dropdown>
+    </Menu>
+  )
+}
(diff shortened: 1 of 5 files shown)


jsapp/js/components/submissions/submissionMediaUtils.ts
```diff
@@ -3,6 +3,11 @@ import type { DataResponse } from '#/api/models/dataResponse'
 import { type AnyRowTypeName, QuestionTypeName } from '#/constants'
 import type { SubmissionAttachment, SubmissionResponse } from '#/dataInterface'
 
+/** Strips repeat-instance indices (e.g. `[2]`) to get the static survey xpath. */
+export function stripRepeatIndices(xpath: string): string {
```
magicznyleszek: Should this be placed in `submissionUtils` as it's not really media related?

jsapp/js/components/submissions/submissionDataTable.tsx
```diff
@@ -298,6 +278,28 @@ class SubmissionDataTable extends React.Component<SubmissionDataTableProps> {
     )
   }
 
+  renderTextResponse(text: string, xpath: string) {
+    // Strip any repeat-instance index to get the question's static survey xpath.
+    const questionXpath = stripRepeatIndices(xpath)
+
+    return (
+      <Group wrap='nowrap' align='flex-start'>
```
magicznyleszek: The group needs `w='100%'` because it looks like this with short response:

Image

jsapp/js/components/common/MoreActionsMenu.tsx
```diff
@@ -0,0 +1,47 @@
...
+  className?: string
+  processingAction?: {
+    assetUid: string
+    xpath: string
+    submissionEditId: string
+  }
+}
+
+/** A "…" button that opens a dropdown menu of actions. */
+export default function MoreActionsMenu({ children, className, processingAction }: MoreActionsMenuProps) {
```
magicznyleszek: I feel like this can become a very generic component for all those "more actions" places.

jsapp/js/components/common/MoreActionsMenu.tsx
```diff
@@ -0,0 +1,47 @@
...
+  return (
+    <Menu withinPortal={false} closeOnClickOutside closeOnItemClick position='bottom-end'>
+      <Menu.Target>
+        <span className={className} style={{ position: 'relative' }}>
+          <ActionIcon size='md' variant='transparent' iconName='more' />
+        </span>
+      </Menu.Target>
+
+      <Menu.Dropdown>
+        {processingAction && (
```
magicznyleszek: I'm not feeling this… The component seems to be generic, and this is very specific. I think I would rather see a silly small `MenuItemProcessing` or something like that

magicznyleszek (changes requested):

magicznyleszek (approved):
LGTM

PR #7576, opened 2026-09-10
fix(processing): keep audio files and NLP data reachable after a question moves into a group DEV-158

platreth:
### 📣 Summary

Audio recordings and their transcripts, translations and analysis answers stay visible after an audio question is moved into a group.

### 📖 Description

Moving a question into or out of a group changes its path in the form. Older submissions keep the path of the version they were made with, so their audio and NLP data used to disappear from the single-submission view, and automatic transcription could fail to find the recording. Both now resolve the question by its name when the path changed.

### 💭 Notes

- `GET .../attachments/?xpath=` now accepts either the pre-move or the current path of a question. When the exact path is missing from the submission XML, the candidates come from the form versions the submission spans, its `__version__` plus `meta/formVersions` (#7543), through `Asset.get_attachment_xpaths_from_version_uids()` (#7567): the media xpaths of those versions sharing the requested last segment, with the `[n]` repeat indexes of the request applied by group name. The XML itself is never searched, since on edits Enketo keeps the record's original nodes and a file referenced only by such a leftover may have been replaced. `__version__` is enough for the pre-move case, so this works on existing submissions with no backfill. Automatic transcription and translation use the same lookup, so they work across form versions too. No API shape changes, no migrations.
- `get_submission_form_version_uids()` is the public half of `_get_other_form_version_uids()` in `logger_tools`, extracted so the deployment backend reads the lineage through the same code as the soft-delete rule.

- most of the ticket as written was already covered by DEV-2345 / DEV-2407 (processing view and data table read old submissions by their submission-era path). This PR closes the three spots still comparing paths strictly: the attachment lookup by xpath in the deployment backend, the NLP rows in the single-submission view, and the row lookup for a question moved from one group to another.
- the frontend half follows #7187 and #7521: stored keys are never rewritten, and the browser, which has no version definitions, falls back on the question name.
- in the single-submission view a pre-move submission surfaces its answer through the "unaccounted answers" path (plain groups don't recurse when nothing is stored under the group), so the NLP rows are attached there as well.
- still open, tracked in DEV-2821 because it needs a design decision rather than a fix: advanced-feature rows stay pinned to the xpath they were enabled at, so submissions made after the move need the features re-enabled at the new path, which forks config and analysis questions per era.
- each new test was run against the unfixed code first and fails there.

### 👀 Preview steps

1. ℹ️ have a project with one audio question at the root, deployed, with NLP languages loaded on the instance
2. submit one audio response, open it in the processing view, add a transcript, save
3. in the form builder select the question, click "Create group with selected questions", save, redeploy
4. in DATA, open that submission's record view (eye icon)
5. 🔴 [on main] the audio row shows but no transcript row anywhere
6. 🟢 [on PR] the transcript row appears right under the audio answer
7. 🟢 [on PR] `GET /api/v2/assets/<uid>/data/<id>/attachments/?xpath=<group>/<question>` returns the pre-move recording instead of a 404

### DIFF
--- a/jsapp/js/components/submissions/submissionUtils.tests.ts
+++ b/jsapp/js/components/submissions/submissionUtils.tests.ts
@@ -370,7 +370,33 @@ describe('getSubmissionDisplayData for a question moved into a group', () => {
       .expect(groupResponses.find((response) => response.data === 'This is french transcript text.'))
       .to.not.equal(undefined)
   })
+
+  it('should keep the NLP rows of a submission made before the move', () => {
+    // A submission made before the move stores both the answer and its
+    // `_supplementalDetails` under the pre-move flat name. The current form
+    // only knows the grouped path, so the answer surfaces through the
+    // unaccounted-answers path - which must bring the NLP rows along, or the
+    // transcript silently vanishes from the single-submission view.
+    const asset = withRowMovedIntoGroup(assetWithSupplementalDetails, 'Secret_password_as_an_audio_file', 'grp')
+    const submission: SubmissionResponse = JSON.parse(JSON.stringify(submissionWithSupplementalDetails))
+
+    const allResponses = getAllResponses(getSubmissionDisplayData(asset, 0, submission))
+
+    // (a) the audio answer itself, kept visible under its submission-era path.
+    chai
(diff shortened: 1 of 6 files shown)


platreth:
@greptileai

PR #7578, opened 2026-09-10
feat(accounts): support custom signup fields on headless API DEV-2807

rajpatel24:
### 📣 Summary
Adds the two signup fields allauth doesn't know about, `newsletter_subscription` and `terms_of_service`, to its headless signup endpoint so the SPA registration screen can collect

### 📖 Description
Registration collects two things beyond `username`/`email`/`password` that have to be recorded when the account is created: `newsletter_subscription` and `terms_of_service`. Both were declared only on the form used by the Django HTML page (`ACCOUNT_FORMS['signup']`), which allauth's headless API does not use.

`POST /api/v2/allauth/browser/v1/auth/signup` therefore accepted only three fields, and because Django forms ignore unknown keys, anything else was silently discarded with a success response. The SPA registration screen needs both fields, and consent in particular cannot be deferred until after signup.

This PR moves those two declarations to `ACCOUNT_SIGNUP_FORM_CLASS`, the hook allauth provides for this: it injects the class as a base of every signup form, so the HTML page, the SSO page and the headless API all get the fields. allauth's own schema generator reads the same class, so they land in the OpenAPI schema and the generated orval types with no manual schema work. 

The rest of the profile metadata (`name`, `organization`, `sector`, `country`, `gender`) is deliberately left off the API. As agreed with the frontend team, the SPA collects it after login through `PATCH /me/`, which already validates required fields against `USER_METADATA_FIELDS`. That also covers SSO users, who never see a signup form at all, so both kinds of  user end up completing their profile in the same place. Those fields stay on `KoboSignupMixin`, which only the HTML and SSO forms use, so the Django registration page is unchanged and keeps collecting everything until future work removes it. 

The new class lives in `signup_fields.py` rather than `forms.py` because allauth resolves the setting while `allauth.account.forms` is still importing; pointing it at `forms.py` is a circular import and the app will not boot. `KoboSignupMixin` stays above allauth's classes on the two form classes it is mixed into, which is what lets its `clean_email()` and organization skip-logic keep overriding allauth's behaviour exactly as before.

### 👀 Preview steps

1. ℹ️ From `/api/v1/docs`, try hitting `POST /api/v2/allauth/app/v1/auth/signup` with the two new fields: `newsletter_subscription` and `terms_of_service`
2. 🔴 [on main] notice that this endpoint doesn't exist.
3. 🟢 [on PR] notice that this will create a new user. You'll get a 401 response since the email is not yet verified, but you can see the newly created user in the admin.
4. 🟢 Try the validations for these fields; they should all work as expected.
5. 🟢 Try testing the existing HTML-templated flow; it should not be broken.

NOTE: We get a 401 response from the API because we have set `ACCOUNT_EMAIL_VERIFICATION = 'mandatory'`. This means that after a successful signup, the session remains unauthenticated pending email verification, which allauth reports as 401 rather than 2xx.

### DIFF
--- a/jsapp/js/api/models/baseSignup.ts
+++ b/jsapp/js/api/models/baseSignup.ts
@@ -23,5 +23,5 @@ export interface BaseSignup {
   sector?: string
   country?: string
   newsletter_subscription?: boolean
-  terms_of_service: boolean
+  terms_of_service?: boolean
 }
--- a/jsapp/js/api/models/baseSignup.ts
+++ b/jsapp/js/api/models/baseSignup.ts
@@ -15,4 +15,13 @@ import type { Username } from './username'
 export interface BaseSignup {
   email: Email
   username: Username
+  name?: string
+  organization?: string
+  organization_website?: string
+  organization_type?: string
+  gender?: string
+  sector?: string
+  country?: string
+  newsletter_subscription?: boolean
+  terms_of_service: boolean
 }
--- a/kobo/apps/accounts/adapter.py
+++ b/kobo/apps/accounts/adapter.py
@@ -1,5 +1,4 @@
 from allauth.account.adapter import DefaultAccountAdapter
-from allauth.account.forms import SignupForm
 from allauth.core.exceptions import ImmediateHttpResponse
 from allauth.socialaccount.adapter import DefaultSocialAccountAdapter
 from allauth.socialaccount.helpers import render_authentication_error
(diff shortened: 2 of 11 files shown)


jamesrkiger:
@rajpatel24 FYI, I'm adding @magicznyleszek as a reviewer just to make sure we're all still on the same page about this change. I'm not sure we are actually going to need these changes now that we're splitting up the sign-up and profile forms on the frontend.

platreth (approved):
LGTM

PR #7581, opened 2026-09-11
refactor(frontend): tsify FormSubScreens DEV-2747

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 💭 Notes

Changes here:
- Moved `jsapp/js/components/formSubScreens.js` → `jsapp/js/project/FormSubScreens.tsx`
  - State is typed as `Partial<AssetResponse>`, because `mixins.dmix` only fills it in once the asset lands
  - Dropped unused `autoBind` and the `mixins.contextRouter` mixin
  - The sub-screen render methods now take the asset as an argument, so children get a non-optional `AssetResponse` without casts
- `RESTServices/index.tsx`: `hookUid` is now optional — the services list route doesn't have one, and the component already handled its absence
- `formMapWrapper.tsx`: dropped `WithRouterProps` from its props, as it never read `router` nor `params`
- `router.tsx`: pointed at the new path

### 👀 Preview steps

1. ℹ️ have an account and a deployed project with a handful of submissions
2. Go to Project → Data and click through Table, Gallery, Downloads and Map
3. 🟢 all four open as before
4. On the Map, pick a question from the "Display by" list
5. 🟢 the map redraws for that question, and the question name shows up in the URL
6. Go to Project → Settings and click through General, Media, Sharing, Connect Projects, REST Services and Activity
7. 🟢 all six open as before
8. In REST Services, add a service (any URL will do), then open it to see its logs
9. 🟢 the logs screen opens
10. Switch between a couple of projects and revisit some of the screens above
11. 🟢 each screen shows the project you're actually looking at, not the one you just left


duvld (approved):
LGTM

PR #7585, opened 2026-09-11
feat(usage): project usage breakdown search DEV-2701

duvld:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Adds a search box to the project usage breakdown page to filter by project name

### 💭 Notes
The way margins + padding were done on the usage page created horizontal clipping onto the new search box. To fix this, I tried to mimic MembersRoute.tsx but the usage page is broken down into two steps via its top tabs. I introduced some padding to both the tabs and the table itself to avoid the clipping as well as keeping everything lined up.

MembersRoute.tsx also has the same search box with the same timeout and warning logic. We may even want to refactor the entire search phrase hook but I left it as just the constants. These constants represent the default expectation from the backend

### 👀 Preview steps

1. ℹ️ have an account and many projects (rest of steps taken from #7523)
2. go to Account Settings → Usage → Per Project Total tab
1. type at least three characters of a member's name → 🟢 notice the list narrows to matches
1. clear the field → 🟢 notice the full list comes back
1. go to page 2, then type a search that matches someone on page 1 → 🟢 notice you land on page 1 rather than an empty page
1. type one or two characters → 🟢 notice nothing happens and the full list stays put
1. leaving those one or two characters in the field, press Enter → 🟢 notice a warning explaining the three character minimum
1. search for project that doesn't exist, e.g. voldemort → 🟢 notice the table explains there is no data to display

### DIFF
--- a/jsapp/js/account/organization/MembersRoute.stories.tsx
+++ b/jsapp/js/account/organization/MembersRoute.stories.tsx
@@ -2,12 +2,13 @@ import type { Meta, StoryObj } from '@storybook/react-webpack5'
 import { http, HttpResponse } from 'msw'
 import { toast } from 'react-hot-toast'
 import { expect, userEvent, waitFor, within } from 'storybook/test'
+import { TOO_SHORT_SEARCH_WARNING } from '#/components/common/searchPhrase.constants'
 import organizationMock from '#/endpoints/organization.mocks'
 import organizationMembersMock from '#/endpoints/organizationMembers.mocks'
 import { queryClientDecorator } from '#/query/queryClient.mocks'
 import { RequireOrg } from '#/router/RequireOrg'
 import ToasterConfig from '#/toasterConfig'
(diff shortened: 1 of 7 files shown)


jsapp/js/account/organization/MembersRoute.stories.tsx
```diff
@@ -2,12 +2,13 @@ import type { Meta, StoryObj } from '@storybook/react-webpack5'
 import { http, HttpResponse } from 'msw'
 import { toast } from 'react-hot-toast'
 import { expect, userEvent, waitFor, within } from 'storybook/test'
+import { TOO_SHORT_SEARCH_WARNING } from '#/components/common/searchPhrase.constants'
```
magicznyleszek, line 5: :+1:

magicznyleszek (approved):
LGTM

PR #7587, opened 2026-09-14
fix(exports): drop legacy _notes and _tags columns DEV-2817

platreth:
### 📣 Summary

Data exports no longer include the `_notes` and `_tags` columns.

### 📖 Description

Notes and tags on submissions came from the legacy KoboCAT API, which is gone, so nobody can fill them in anymore. Exports kept the two columns anyway, which was confusing. They are removed from CSV, XLSX and SPSS labels exports, along with the `_submission__notes` and `_submission__tags` columns on repeat group sheets.

### 💭 Notes

- The change is in kpi only. formpack still defines `NotesCopyField` and `TagsCopyField`; kpi simply stops listing them in `SubmissionExportTaskBase.COPY_FIELDS`. There is no API or schema change and no migration.
- Old submissions may still carry `_notes` and `_tags` keys in Mongo. The data API already strips them in `_inject_properties`, exports were the last place still showing them. The test fixtures keep those keys on purpose so the tests prove the columns stay out.
- The copy columns now come out as `_id, _uuid, _submission_time, _validation_status, _status, _submitted_by, __version__, meta/rootUuid, _index`. A script that reads an export by column position instead of by header will be off by one from `_status` on and by two after `__version__`.

### 👀 Preview steps

1. ℹ️ have a deployed project with at least one submission
2. in DATA, open the Downloads tab, export as CSV and open the file
3. 🔴 [on release/2.026.33] the header row has `_notes` between `_validation_status` and `_status`, and `_tags` between `__version__` and `meta/rootUuid`
4. 🟢 [on PR] neither column is there and `_status` directly follows `_validation_status`
5. 🟢 [on PR] with a repeat group, the repeat sheet of an XLSX export has no `_submission__notes` or `_submission__tags` column either


platreth:
@greptileai

rajpatel24 (approved):
LGTM

PR #7589, opened 2026-09-14
fix(queryParser): reserve every Django field lookup in list queries DEV-2469

platreth:
### 📣 Summary

Searching a list field with an operator the list syntax cannot run, for example `summary__languages[]__istartswith:Eng`, now returns an error instead of silently matching nothing.

### 💭 Notes

- `RESERVED_KEYWORDS` is only consulted for the `field[]__X` syntax. It decides whether the segment after `[]__` is a Django lookup or a key inside the list's dicts. It has no say on plain-field operators, which go straight to Django. That is why DEV-2349 was fixed with the field allowlist rather than by trimming operators.
- Renamed to `DJANGO_FIELD_LOOKUPS`, since it now mirrors `Field.class_lookups` minus `in` and `range`, which cannot be passed through `q=`. `JSONFIELD_COMPARISON_OPERATORS` was on the table but `isnull`, `regex` and `contains` are not comparisons. A test pins the set against Django's registry so a Django upgrade that adds a lookup fails loudly. kpi's own global `in_array` lookup is excluded there for the same list-value reason.
- `get_q_for_list` is untouched. A per-element `startswith`, `regex` or `gt` on a JSON list cannot be built from Django's lookups (they cast the whole array to text), so those names now raise `QueryParserNotSupportedFieldLookup` like any other lookup the list syntax does not implement. Before, they were read as a dict key and the query quietly returned no rows. Real support would need a jsonpath (`@?`) lookup and is out of scope here.
- The project views UI only sends `[]__iexact` and `[]__icontains` on list fields, so nothing changes in the app. Only raw API callers can hit the new 400.

### 👀 Preview steps

1. ℹ️ any instance, logged in or not
2. call `GET /api/v2/assets/?q=summary__languages[]__istartswith:Eng`
3. 🔴 [on main] `200` with `"count": 0`, the term was read as a dict key named `istartswith`
4. 🟢 [on PR] `400` with `{"detail": "Not supported field lookup"}`
5. 🟢 [on PR] `GET /api/v2/assets/?q=summary__languages[]__iexact:English` still returns `200` and lists the matching projects


platreth:
@greptileai

rajpatel24 (approved):
LGTM

PR #7592, opened 2026-09-14
ci(releases): don't sometimes skip linear release linking

Akuukis:
### 💭 Notes

Don't skip Linear Issue linking to Linear Release when tests are skipped.


platreth (approved):
LGTM

PR #7597, opened 2026-09-15
fix(backend): return git version number DEV-307

platreth:
### 📣 Summary
Superusers and instances that enable `EXPOSE_GIT_REV` can see the server's git revision in the API again.

### 📖 Description
The `git_rev` block in the `/me/` response came back as all `false` on servers even with the setting turned on. It now contains the actual commit hashes, branch and tag.

### 💭 Notes
Two separate problems:

- git 2.35.2+ refuses to run in a repo owned by another user ("dubious ownership"). In our containers `/srv/src/kpi` (including `.git`) belongs to root while uWSGI and celery run as `kobo`, so every `git rev-parse` in the settings exited with code 128 and `GIT_REV` fell back to all-`false`. Each git call now passes `-c safe.directory=<repo root>`, scoped to those four read-only commands. No image or gitconfig change needed.
- `subprocess.check_output` returned bytes, so the detached-HEAD normalization (`branch == 'HEAD'`) never matched and the values were not JSON-native strings. Fixed with `text=True`.

The loop now lives in `kobo.settings.utils.get_git_rev()` (settings just call it with `BASE_DIR`) so `kpi/tests/test_git_rev.py` can pin the `safe.directory` argv, `text=True`, the `CalledProcessError`/`OSError` fallbacks and the detached-HEAD normalization with a mocked `subprocess`. The ownership mismatch itself can't run in CI (needs root to chown), so the argv assertion is the regression guard.

### 👀 Preview steps
Backend only, no UI.

1. ℹ️ have the dev stack running
2. from kobo-install, run `python3 run.py -cf exec kpi python -c "from django.conf import settings; print(settings.GIT_REV)"`
3. 🔴 [on main] values print as bytes, e.g. `{'short': b'3d3441711', ...}`
4. 🟢 [on PR] plain strings: `{'short': '…', 'long': '…', 'branch': '…', 'tag': False}`
5. to see the production failure mode, run git as an unprivileged user against a root-owned copy of the repo:
   `docker exec <kpi container> bash -c 'mkdir -p /tmp/r && cp -r /srv/src/kpi/.git /tmp/r/.git && cd /tmp/r && su kobo -s /bin/bash -c "git rev-parse --short HEAD; git -c safe.directory=/tmp/r rev-parse --short HEAD"'`
6. 🔴 the plain call fails with `fatal: detected dubious ownership`; 🟢 the `safe.directory` call (what this PR adds) prints the hash


platreth:
@greptileai

platreth:
@greptileai recheck

rajpatel24 (approved):
LGTM

PR #7602, opened 2026-09-15
fix(formbuilder): better handling of importing (un)translated blocks from Library DEV-2836

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary

Fixed adding questions and blocks from the Library into a form: labels are no longer replaced with question names, option labels stay with the language they were written in, and the form still opens in the Formbuilder afterwards.

### 📖 Description

Dragging an item out of the Library into an open form means matching the item's languages to the ones the form already has. Three things went wrong there:

- when the item didn't have the form's default language, that column got the question's name instead of the item's own label,
- when the form had no named language, the item's languages came along anyway and left the form with an unnamed language next to named ones — the Form Builder refuses to open such a form, and the downloaded XLSForm has its translation columns scrambled,
- labels of select options were left behind whenever languages got reordered or added, so they ended up under the wrong language.

Forms already saved in that state aren't repaired here — user needs to name the unnamed language with "Manage translations" from the Project's landing page. Project → Summary now shows "Unnamed language" instead of a blank line, so it's easier to see what's wrong.

### 💭 Notes

This PR targets `release/2.026.33` release branch.

Changes here:

- `formBuilderUtils.ts`, all in `nullifyTranslations()`:
  - it now takes `choices` and rewrites both sheets through one new `rewriteTranslatedValues()` helper
  - the cases are now a single `else if` chain
  - new case: the form has no named language → keep the item's default value and drop its other translations, rather than producing an unnamed language among named ones (with a TODO comment for future improvement)
  - the "form's default language missing from the item" case now uses label and name as fallback
- `model.inputParser.coffee` — passes `choices` through, and merges the item's `translated` list into the form's. A column the form doesn't translate yet (for example `guidance_hint`) would otherwise be saved without a language suffix
- `formSummary/formSummaryProjectInfo.tsx` — a `null` language used to render as an empty cell
- `dataInterface.ts` — comment only. `summary.languages` can contain `null` while the generated OpenAPI type says it can't; the schema fix is DEV-2864.
- tests:
  - `formBuilderUtils.tests.js` — one case each, choices included.
  - `test/xlform/libraryImport.tests.ts` (new) — runs the real path, `Survey.loadDict` → `insertSurvey` → `unnullifyTranslations`, and asserts on what the backend would actually receive, column names and all. The `translated` bug is only visible at that level.
  - `tsconfig.json` — `test/**` wasn't type-checked, and the new test is TypeScript

### 👀 Preview steps

1. ℹ️ have an account and two projects, each with a form containing one question
2. Library → NEW → Block: add a *Select One* question with two options, save it
3. Library → the block's ⋯ menu → "Manage translations": name the existing language `Francais (fr)`, add `Polski (pl)`
4. In "Manage translations" fill in both languages for the question label and both option labels
5. First project → ⋯ → "Manage translations": name its language `English (en)`
6. First project → Edit form → `Add from Library` → drag the block into the form
7. 🔴 [on release/2.026.33] the English column shows the question's *name*, and the option labels sit under the wrong languages
8. 🟢 [on PR] the English column shows the block's French labels, Polski keeps the Polish ones, options included
9. Save, leave, and reopen the form in the Form Builder — 🟢 it opens and the labels are where you left them
10. Now the fun one: leave the second project's language unnamed (don't touch "Manage translations"), drag the same block in, save
11. 🔴 [on release/2.026.33] reopening that form in the Form Builder fails with "There is an unnamed translation in your form definition."
12. 🟢 [on PR] the form opens, the block arrived with its French labels only, and the form still has no language
13. 🟢 for a project already broken this way, Project → Summary lists "Unnamed language" under Languages instead of a blank entry


magicznyleszek:
@greptileai review the code

jamesrkiger (approved):
lgtm

PR #7604, opened 2026-09-15
fix(subscriptions): handle unparseable values DEV-2867

rgraber:
### 🗒️ Checklist

1. [ ] run linter locally

### 💭 Notes
Handles an error when computing subscription limits. If a limit is unparseable, we will set it to inf and log the unparseable value. While this does create a risk of being overly-generous with limits, there is not a consistent way to determine what the person who updated the metadata meant (even a comma could be a decimal separation in some languages) and it's better to air on the side of not preventing users from taking actions.

### 👀 Preview steps
The method in question is called fairly deep in internal processing, so the unit test coverage should suffice rather than trying create the circumstances that will eventually call it.


rgraber:
@greptileai

kobo/apps/stripe/utils/subscription_limits.py
```diff
@@ -389,19 +390,36 @@ def determine_limit(
...
     # for storage, factor in addons if specified
     if usage_type == UsageType.STORAGE_BYTES and include_storage_addons:
+        parseable_addon_limit = True
         if addon_limit == 'unlimited':
             addon_limit = inf
         else:
-            addon_limit = int(addon_limit or 0)
+            try:
+                addon_limit = int(addon_limit)
+            except (ValueError, TypeError):
```
Guitlle, line 413: Should also log the unparseable addon limit

Guitlle (changes requested):
`

Guitlle (approved):

PR #7607, opened 2026-09-17
refactor(formSidebar): typescriptize formViewSideTabs DEV-2749

duvld:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Updates old component to typescript

### 💭 Notes
Generally easy conversion, this component only serves to organize routes and serve up other components. There is a small quirk for just the "Sharing" navigation, it triggers a refresh of the sidebar and the content page. The logic should be the same with an added TS guard that doesn't affect the routing. It doesn't break any behaviour but it is strange that it only affects this route.

### 👀 Preview steps
2. go to project settings
5. 🟢 [on PR] notice that clicking the tabs produces the right content and the content does indeed affect the correct project


jsapp/js/project/formViewSideTabs.tsx
```diff
@@ -42,17 +59,25 @@ export function getFormDataTabs(assetUid) {
   ]
 }
 
-class FormViewSideTabs extends Reflux.Component {
-  constructor(props) {
+class FormViewSideTabs extends Reflux.Component<typeof Reflux.Store, FormViewSideTabsProps, FormViewSideTabsState> {
+  // Needed to reference the reactMixin functions
+  declare currentAssetID: () => string | undefined
+  declare isActiveRoute: (path: string) => boolean
```
magicznyleszek: `declare` causes tests to fail. I think you could use functions from jsapp/js/router/routerUtils.ts instead of those, and then drop all mixins from the file

jsapp/js/project/formViewSideTabs.tsx
```diff
@@ -176,15 +210,15 @@ class FormViewSideTabs extends Reflux.Component {
...
+  render(): React.ReactNode {
     if (!this.props.show) {
       return false
     }
     return this.renderFormSideTabs()
   }
 }
 
-reactMixin(FormViewSideTabs.prototype, Reflux.ListenerMixin)
+reactMixin(FormViewSideTabs.prototype, Reflux.ListenerMixin as unknown as Mixin<any, any>)
```
magicznyleszek: If you don't see `listenTo` being used anywhere in a file, you can simply drop `ListenerMixin`

magicznyleszek (changes requested):

magicznyleszek (approved):

PR #7610, opened 2026-09-17
refactor(frontend): move FormSubScreens to mantine and drop Reflux DEV-2478

duvld:
### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
Updates the frontend logic for the sidebar in the form view

### 💭 Notes
Clean refactor in general. Dropped Reflux in this current component, but the component needs to reuse the asset loaded by `PermProtectedRoute` instead of calling `useAssetsRetieve` independently. This is needed to avoid duplicate requests while the authorization and asset loading is owned by the protected route. Refactoring this is beyond the scope of this PR.

`bem` components easily replaced with `Box`, but the classnames needed to stay. We should be able to easily move the classname style attributes directly to mantine style props, but this will create inconsistent styling with other consumers that still rely on legacy styles. Left a `TODO` for this as well
### 👀 Preview steps
**Copied steps from #7581 as the goal is to check for regressions**.

1. ℹ️ have an account and a deployed project with a handful of submissions
2. Go to Project → Data and click through Table, Gallery, Downloads and Map
3. 🟢 all four open as before
4. On the Map, pick a question from the "Display by" list
5. 🟢 the map redraws for that question, and the question name shows up in the URL
6. Go to Project → Settings and click through General, Media, Sharing, Connect Projects, REST Services and Activity
7. 🟢 all six open as before
8. In REST Services, add a service (any URL will do), then open it to see its logs
9. 🟢 the logs screen opens
10. Switch between a couple of projects and revisit some of the screens above
11. 🟢 each screen shows the project you're actually looking at, not the one you just left


magicznyleszek (approved):
LGTM

PR #7612, opened 2026-09-18
refactor(auth): make more generic AuthAside, and AuthPageFrame DEV-1850

magicznyleszek:
### 🗒️ Checklist

1. [x] run linter locally

### 💭 Notes

Groundwork for a route blocker - it will add a "complete your profile details" screen that has to look like the other authentication screens but is reached after login, outside the `/auth` routes. Two pieces it needs were split out:

- `AuthPageFrame` - the background, logo, language picker and legal links, lifted out of `AuthContainer`.  `AuthContainer` is now just the `/auth` layout route: the frame around an `<Outlet/>`.
- `AuthAside` - `RegisterRoute/RegisterAside` made generic and moved next to `AuthCard`, plus  `shouldRenderAuthAside()` so a screen can decide whether to open the column at all. 

Pure refactor, no behavior change intended.

### 👀 Preview steps

1. ℹ️ needs the `authRedesignEnabled` feature flag on
2. go to `/#/auth/register`
3. 🟢 notice the background, logo, language picker and Terms/Privacy links are unchanged from main
4. ℹ️ in Django admin, clear both the login supporting image and the `welcome_message` sitewide message
5. 🟢 notice the card is one column wide, with no divider
6. ℹ️ set either one back
7. 🟢 notice the supporting column returns, and the card widens as before
8. 🟢 notice tab order through logo → language picker → form → footer links is unchanged


jamesrkiger:
I don't think it's directly caused by this PR, but I notice when loading the registration page I'm hitting the /environment endpoint four times. I think we need to configure the staletime on the hook for that API call. Can you look into that?

jamesrkiger (approved):
Left a comment about us hitting the environment endpoint too much. Let's resolve that in this PR stack, but I'll leave it up to you which PR to do it in. Otherwise LGTM

PR #7621, opened 2026-09-19
fix(frontend): Prevent duplicate alerts and network requests within share form

### LINKED ISSUE #7620: Duplicate alerts can show within Sharing form
**Summary**
Within the Sharing form, such as the Sharing Permissions modal, a user can attempt to add a users name to the list. If that user is not found, a `User not found` alert is shown. If they try to again add that same user, two `User not found` alerts are shown, instead of one. This also runs a network request, even though we already know that the user does not exist.

https://github.com/user-attachments/assets/25d735cc-4404-48e8-8679-2e22805905f8

**Steps to reproduce**
1. Be on the Projects overview page.
2. Select a project via its checkbox.
3. Click the 'Share project' button.
4. Click 'Add user'.
5. Enter a name into the username input that is not part of the project, such as 'John'.
6. Click outside of the input in order for the check to run.
7. Notice the alert message that states 'User not found: John'.
8. Repeat steps 5 and 6.
9. Notice two alert messages are shown that state 'User not found: John'.

**Give sample code if you can**
Within `<UserAssetPermsEditor>`, the function `onUsernameChangeEnd` has an if statement to see if we have previously checked for a username, on line 197. This runs and displays an alert to the user. It then continues on to do a network request, line 204, for that same username that we know does not exist, which then again displays an alert to the user, line 212.

Image

**What you expected would happen**
I expect us to return early, and not display a second alert or run a new network request, since we know that the given user does not exist.

**What actually happens**
When a name that has previously been entered is entered again, we display two alerts for that user and run a network request to check if they exist.

**Notes**
This is occurring because we are not returning early within `onUsernameChangeEnd`, after the point where we have confirmed that a name was entered again and does not exist.

CameronStAmant:
Fixes https://github.com/kobotoolbox/kpi/issues/7620

### 🗒️ Checklist

1. [x] run linter locally

### 📣 Summary
If a username is entered multiple times within the share form, and they do not exist, the form now does not perform additional checks to see if they exist, since we already discovered that during the initial time they were entered.

When adding a username within the share form, if the name has been previously entered and they were not found, entering it again now returns early. This prevents 2 new duplicate alerts from appearing and a network request from being called. I also added a test to prevent future regression.

### 💭 Notes
I noticed while creating the test that the Jest asset mapper was erroring because it was missing the ?react suffix on SVG imports, so Jest tried to resolve real files like alert.svg?react instead of treating them as mocked assets.

### 👀 Preview steps
1. Be on the Projects overview page.
2. Select a project via its checkbox.
3. Click the 'Share project' button.
4. Click 'Add user'.
5. Enter a name into the username input that is not part of the project, such as 'John'.
6. Click outside of the input in order for the check to run.
7. Notice the alert message that states 'User not found: John'.
8. Repeat steps 5 and 6.
9. 🔴 [on main] Notice the two duplicated alert messages that state 'User not found: John' and the network request that runs.
10. 🟢 [on PR] Notice the single alert message that states 'User not found: John' and no network request.


CameronStAmant:
I am unable to do the following checklist items possibly due to read only permissions to this repo, though I did confirm the CI passes locally.

1. [ ] assign yourself, tag PR: at least Front end and/or Back end or workflow
2. [ ] open PR & confirm that CI passes & request reviewers, if needed

PR #7622, opened 2026-09-21
fix(library): transfer uploaded library items to the organization and let admins delete them DEV-2873

platreth:
### 📣 Summary
Library items uploaded by a member of a Teams organization now belong to the organization, and organization admins can delete library items from My Library.

### 📖 Description
In a Teams organization, whatever a member creates belongs to the organization. That was already true for question blocks, templates and collections created from the Library. Items added with the Upload button were the exception: they stayed with the person who uploaded them. They now go to the organization as well, and the uploader can still edit and manage them.

Admins were also missing the Delete option on library items owned by their organization, although they are allowed to delete them. The option is now there, as it is for projects.

### 💭 Notes
- `ImportTask` used to create assets with the requesting user as owner. It now does what the API create path does (`perform_create_override` and `AssetSerializer.create`): owner from `get_real_owner`, `is_excluded_from_projects_list`, `manage_asset` for the uploader on every created item in the same transaction (collections only pass view and change down to their children), plus `created_by` and `last_modified_by`. We need `created_by` because the non-owner delete rule checks it, so without it a plain member could not delete their own upload anymore. Imports into an existing asset are not touched.
- The API already let admins delete, the block was in the frontend. `AssetMoreActions` showed Delete only with `delete_submissions`, which library assets can never have, so in practice only the owner saw it. It now uses the same condition as `projectQuickActions`. The backend still has the last word, so Delete on an asset from outside the org ends in the usual error.
- `owner__username` in the import messages now gives the asset's owner instead of the uploader.
- Dropping an XLSForm on the project list runs the same import code, so those projects now go to the organization too, like projects created from the modal.

### 👀 Preview steps
1. ℹ️ have a Teams organization with an owner and a member with the Admin role
2. log in as the admin and open My Library
3. click New, then Question Block, and save a block
4. hover the block and open its "More actions" menu
5. 🔴 [on release/2.026.37] there is no Delete entry
6. 🟢 [on PR] Delete is there and removes the block
7. click New, then Upload, and upload any XLSForm
8. 🔴 [on release/2.026.37] the Owner column shows "me"
9. 🟢 [on PR] the Owner column shows the organization, same as the block from step 3


platreth:
@greptileai

rajpatel24 (approved):
Backend lgtm!

jsapp/js/components/assetsTable/AssetMoreActions.tsx
```diff
@@ -40,8 +42,14 @@ export default function AssetMoreActions(props: AssetMoreActionsProps) {
   const [isMenuOpen, setIsMenuOpen] = React.useState(false)
   const menuRef = React.useRef<HTMLDivElement>(null)
   const assetType = props.asset.asset_type
+  const [organization] = useOrganizationAssumed()
+  const isAdmin = organization.request_user_role === MemberRoleEnum.admin
```
jamesrkiger: Any reason not to align this with isMmoMember below (i.e. isMmoAdmin and checking on organization.is_mmo)? It's not technically necessary, since I don't think it's possible to be an admin of a non-MMO org, but it feels more precise.

jsapp/js/components/assetsTable/AssetMoreActions.tsx
```diff
@@ -40,8 +42,14 @@ export default function AssetMoreActions(props: AssetMoreActionsProps) {
...
+  const isMmoMember = organization.is_mmo && organization.request_user_role === MemberRoleEnum.member
   const userCanEdit = userCan('change_asset', props.asset)
-  const userCanDelete = userCan('delete_submissions', props.asset)
+  // Org admins can delete org-owned assets (backend enforces the rest). MMO
+  // members are gated on manage_asset; everyone else on delete_asset.
+  const userCanDelete =
```
jamesrkiger: Beyond the comment above, I think `isAdmin` is being used too loosely here. The right side of the `or` is fine, but on the left we're ultimately just checking if the user is admin of *an* org, not the relevant org. Can we check the actual permissions here?

jamesrkiger (changes requested):
Left a couple comments

jsapp/js/components/assetsTable/AssetMoreActions.tsx
platreth: Aligned! Thanks

jsapp/js/components/assetsTable/AssetMoreActions.tsx
platreth: Yes fair point, even though the backend wouldn't let this pass, it was inconsistent, I fixed it let me know

jsapp/js/components/assetsTable/AssetMoreActions.tsx
```diff
@@ -40,8 +42,18 @@ export default function AssetMoreActions(props: AssetMoreActionsProps) {
   const [isMenuOpen, setIsMenuOpen] = React.useState(false)
   const menuRef = React.useRef<HTMLDivElement>(null)
   const assetType = props.asset.asset_type
+  const [organization] = useOrganizationAssumed()
+  const isMmoAdmin = organization.is_mmo && organization.request_user_role === MemberRoleEnum.admin
+  const isMmoMember = organization.is_mmo && organization.request_user_role === MemberRoleEnum.member
+  // `owner_label` is the organization name for org-owned assets, so this
+  // checks the admin belongs to the org that owns the asset.
+  const isOwnedByUserOrg = props.asset.owner_label === organization.name
```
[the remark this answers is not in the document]
platreth: @jamesrkiger What do you think about this one? I think it's fair but it shouldn't be in this PR. Follow-up ticket?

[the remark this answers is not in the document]
jamesrkiger: @platreth Now that I'm thinking about this more, do we actually need this at all? Isn't it covered by `userCan('delete_asset', props.asset)`, since that's pulling from effective permissions?

jsapp/js/components/assetsTable/AssetMoreActions.tsx
```diff
@@ -40,8 +46,28 @@ export default function AssetMoreActions(props: AssetMoreActionsProps) {
...
+  const isMmoAdmin = organization?.is_mmo === true && organization.request_user_role === MemberRoleEnum.admin
+  const isMmoMember = organization?.is_mmo === true && organization.request_user_role === MemberRoleEnum.member
+  // `owner_label` is the organization name for org-owned assets, so this
+  // checks the admin belongs to the org that owns the asset.
+  const isOwnedByUserOrg = organization !== undefined && props.asset.owner_label === organization.name
   const userCanEdit = userCan('change_asset', props.asset)
-  const userCanDelete = userCan('delete_submissions', props.asset)
+  // Org admins can delete their org's assets (backend enforces the rest). MMO
+  // members are gated on manage_asset; everyone else on delete_asset.
+  const userCanDelete =
```
jamesrkiger: I'm actually starting to think this should just be `userCan('manage_asset', props.asset)`.

jamesrkiger (changes requested):
Sorry for the back and forth, but I left a couple more comments. I'm wondering if we're making this more complicated than it needs to be given the effective permissions the backend already provides. If I'm missing something and that's not possible, let me know.

jamesrkiger (approved):
LGTM

PR #7623, opened 2026-09-21
fix(exports): include submissions in XLS (legacy) exports of transferred projects DEV-2862

platreth:
### 📣 Summary
XLS (legacy) exports are no longer empty for projects that were transferred to another owner.

### 📖 Description
Once a project had been transferred to another account, its XLS (legacy) export came out with the column headers and no submissions. The table view and the CSV (legacy) export still showed everything. Projects that moved to an organization when their owner joined it were affected too, since that is also a transfer. The export now contains the submissions again.

### 💭 Notes
- A transfer gives the XForm a `mongo_uuid` and rewrites `_userform_id` on every MongoDB document to it (`rewrite_mongo_userform_id`). Readers get the right id from `get_mongo_userform_id()`. `export_tools.query_mongo` was the one place still building `{username}_{id_string}` by hand, so after a transfer it matched nothing and the workbook was saved with headers only, as a successful export.
- CSV (legacy) was fine because `to_flat_csv_export` ignores that cursor and queries again through `ParsedInstance`, which already uses the helper.
- `query_mongo` now takes the `xform` that `generate_export`, its only caller, has already loaded, so it costs no extra query. For projects that were never transferred the query string is the same as before.
- The new test puts the XForm and MongoDB in the state `test_mongo_uuid_after_transfer` asserts after a real transfer, then exports. On `release/2.026.37` it fails with `1 != 2` (header row only).

### 👀 Preview steps
1. ℹ️ have two accounts, and a deployed project with at least one submission on the first one
2. as the first user, open the project, go to SETTINGS > Sharing and transfer the project to the second user
3. log in as the second user and accept the transfer
4. open the project, go to DATA > Downloads, select "XLS (legacy)" and click "New Export"
5. download the file once it is ready and open it
6. 🔴 [on release/2.026.37] the sheet has the header row and nothing else
7. 🟢 [on PR] the sheet has one row per submission, same as "CSV (legacy)"


platreth:
@greptileai

rajpatel24 (approved):
LGTM

PR #7624, opened 2026-09-21
fix(accounts): stop resending confirmation emails at headless login DEV-2435

rajpatel24:
### 📣 Summary
Signing in with an unverified account through the new authentication API no longer sends another activation email automatically. A new link is only sent when it is explicitly asked for.

### 📖 Description
Until now, every attempt to log in with an account whose email address had not been confirmed quietly triggered a fresh activation email. Users who tried a few times ended up with several near-identical messages in their inbox and no clue which link was the live one.

The redesigned sign-in screen (React SPA + `allauth` headless)  replaces that with a button for requesting a new link,
so the automatic send is no longer needed and has been removed. Creating an account still sends the activation email as before, and the current sign-in page is unchanged; it keeps sending a fresh link on login until it is retired.

`AccountAdapter.should_send_confirmation_mail()` now returns `False` for headless requests unless the send is part of a signup. 

### 👀 Preview steps

1. ℹ️ create a new user and note that you receive a confirmation email. Do not verify the email yet.
4. 🔴 [on main] try to log in with the same user and notice that you receive another confirmation email.
5. 🟢 [on PR] try the headless API `/api/v2/allauth/app/v1/auth/login` from `/api/v2/docs` and notice that it does not send an email on login.

NOTE: The existing templated login will still send the email, and its behavior is unchanged. To test the new behavior, you’ll need to use the headless API.


platreth (approved):
LGTM

PR #7627, opened 2026-09-22
fix(hooks): block redirect-based SSRF bypass in webhook delivery and media download DEV-2911

platreth:
### 📣 Summary
REST Services no longer follow redirects into the server's own network.

### 📖 Description
The address of a REST Service is always checked before a call, so that data is never sent to an address inside the server's own network. Only the address written on the service was checked though: if it answered with a redirect, the call followed it without a second look, so a public endpoint could hand the delivery on to an internal address, and the internal response ended up in the REST Service logs. Every address in the redirect chain is checked now, and a delivery that redirects somewhere forbidden fails with the usual "is not allowed" log entry.

### 💭 Notes
- `requests.post` follows redirects on its own and only the stored endpoint was validated. 301/302/303 turn the POST into a GET on the next hop, 307/308 replay the submission body. The final response body is stored in `HookLog.message` and served by the hook-log API, so the fetch wasn't blind.
- The fix is `SSRFProtectedSession` in `kpi/utils/ssrf.py`, a `requests.Session` that validates in `send()`. `Session.send()` runs once for the initial request and once per hop (`resolve_redirects` calls it), so every verb is covered and the 307/308 replay semantics stay with `requests`. `SSRFProtectException` is not a `RequestException`, so a blocked hop lands in the existing `except SSRFProtectException` branch: the log ends FAILED with "is not allowed" and there is no retry.
- `create_media()` in openrosa's `meta_data.py` had the same pattern (validate once, then `requests.get` with redirects). It now goes through the existing `ssrf_safe_get()`, which validates every hop and keeps `stream=True`. That path is unreachable today (nothing calls `media_resources(download=True)`), so this half is hardening rather than a live fix.
- Tests ran green on both settings axes (`testing` and `testing_no_stripe`).

### 👀 Preview steps
1. ℹ️ backend only, so the preview is scripted
2. in the kpi container, on the PR branch: `pytest kobo/apps/hook/tests/test_ssrf.py kpi/tests/test_utils.py kobo/apps/openrosa/apps/main/tests/test_meta_data.py`
3. 🟢 all pass; `test_send_blocks_redirect_to_private_ip` is the redirect scenario end to end
4. still in the container, run the snippet below (`DJANGO_SETTINGS_MODULE=kobo.settings.testing python repro.py`); it performs the same delivery twice, once through the old code path (kept verbatim in the snippet) and once through the new session
5. 🔴 [old path, what release/2.026.37 does] prints `INTERNAL-SECRET-DATA`, fetched from the internal address
6. 🟢 [new path] prints `is not allowed`; the internal server logs no hit


platreth:
@greptileai

platreth:
@greptileai I don't agree with your comment, I did some test and even with `allow_redirects=False`, `Session.send() `steps the redirect generator once to fill Response.next

platreth:
@greptileai I tried it against a real server. By the time requests.get() gives us the redirect response, requests has already read it and closed it while it was building Response.next. So there's nothing left open for us to close, and adding a close() wouldn't change anything.

platreth:
Deleted comments because it was not in a thread.

rgraber (approved):
LGTM

